PR #20, opened 2022-04-11
Added uninstall target

acly7946:
Adds an uninstall target to the makefile to remove files installed after a 'sudo make install'


vaxerski:
LGTM, thanks!

PR #66, opened 2022-05-08
Add files for nix build

Tanish2002:
Added Nix flakes so NixOS users can build hyprland easily.

Here are the Install Instructions so it can be added to the wiki
1. Nixos with Flakes
```nix
{ inputs, ... }:
let
  hyprland = inputs.hyprland.defaultPackage.x86_64-linux;
in {
  home.packages = [
    hyprland
    # hyprland.override { enableXWayland = false; }) Use this if you don't want xwayland
  ];
}
```

2. Nixos without Flakses
```nix
{ ... }:
let
  hyprland = (builtins.getFlake "github:vaxerski/Hyprland").defaultPackage.x86_64-linux;
in {
  home.packages = [
    hyprland
    # hyprland.override { enableXWayland = false; }) Use this if you don't want xwayland
  ];
}
```


vaxerski:
lgtm, thanks!

Narice:
Hey @Tanish2002! Thanks for the flake!
Would it be possible to add an overlay attribute to the flake for better use for example with devos?
something like that should work?
```nix
overlay = final: prev: {
  hyprland = pkgs.callPackage self {
    src = self;
    inherit (nixpkgs-wayland.packages.${system}) wlroots;
  };
};
```
I'm not sure tho, I'll try to make a fork to test that when I can ;)
But this would comply more with the standard nix way of doing things I guess ;)

Issue #68, opened 2022-05-12
Crash: Underflow in removeBeginEndSpacesTabs(): Assertion '__pos <= size()' failed.


mdartmann:
**Describe the bug**
After compiling the latest commit of Hyprland on my system and trying to start it, Hyprland crashes with the following message:
`/usr/include/c++/12.1.0/bits/basic_string.h:1221: constexpr std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator[](size_type) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&; size_type = long unsigned int]: Assertion '__pos <= size()' failed.`

This happens both with the auto-generated config file and the example provided in the repository.
After running it through gdb, the following loop in the `removeBeginEndSpacesTabs()` function seems to be the issue:
```cxx
while (str[str.length() - 1] == ' ' || str[str.length() - 1] == '\t') {
        str = str.substr(0, str.length() - 1);
    }
```
It looks like `substr()` is underflowing here (the value of `__pos` is `numeric_limits(size_t)::max()`:

```gdb
in std::__cxx11::basic_string, std::allocator >::operator[](unsigned long) (__pos=18446744073709551615, this=0x7fff697f9370) at /usr/include/c++/12.1.0/bits/basic_string.h:1221
```
**Images/videos/anything that would help**

<details><summary>GDB backtrace</summary><pre>
Thread 40 "Hyprland" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fff697fa640 (LWP 5885)]
0x00007ffff7602a6c in pthread_kill () from /usr/lib/libc.so.6
(gdb) set pagination off
(gdb) bt
#0  0x00007ffff7602a6c in pthread_kill () at /usr/lib/libc.so.6
#1  0x00007ffff75ab048 in raise () at /usr/lib/libc.so.6
#2  0x00007ffff758f0f4 in abort () at /usr/lib/libc.so.6
#3  0x00007ffff78e7a0e in  () at /usr/lib/libstdc++.so.6
#4  0x00000000004648c1 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator[](unsigned long) (__pos=18446744073709551615, this=0x7fff697f9370) at /usr/include/c++/12.1.0/bits/basic_string.h:1221
[...]

PR #69, opened 2022-05-12
Fix string underflow in `removeBeginEndSpacesTabs()`

mdartmann:
This PR fixes an underflow in the `removeBeginEndSpacesTabs()` function by sanity-checking the string length before looping over it.

Fixes: #68


vaxerski:
lgtm, thanks

PR #76, opened 2022-05-15
Create aur folder for PKGBUILDS

ThatOneCalculator:
As of right now, there's only the `hyprland-git` package, but once Hyprland gets a release tag, there will probably also be `hyprland` and `hyprland-bin` packages made.


vaxerski:
lgtm, thank you

Issue #77, opened 2022-05-16
Build failed (probably `96cdf8fd`)


fedoranvar:
Good day to you!

i've tried building with `1d4d52af`

but stumbled upon error
![image](https://user-images.githubusercontent.com/52028706/168535786-1e894196-f75c-479e-8d18-887d4a019bda.png)

vaxerski:
no, wlroots-git update. Always keep up to date with wlroots-git.

fedoranvar:
you're right. updated wlroots and its builded

i've been using this repo's flake, it seems it's using "old" wlroots.

PR #78, opened 2022-05-16
Providing an overlay for nix users and small fixes on flake.nix and default.nix

Narice:
This PR provides an overlay that can be used by Nix/OS users. This is more inline with what over nix flake repositories provide.
This PR also fixes other small stuff in flake.nix and default.nix such as the platforms supported or code style consistancy.

I tested this on my setup and it works perfectly ;)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -40,7 +40,7 @@ stdenv.mkDerivation rec {
   installPhase = ''
     cd ../
     mkdir -p $out/share/wayland-sessions
-    cp ./example/hyprland.desktop $out/share/wayland-sessions/
+    cp ./example/hyprland.desktop $out/share/wayland-sessions
     mkdir -p $out/bin
     cp ./build/Hyprland $out/bin
     cp ./hyprctl/hyprctl $out/bin
@@ -49,6 +49,8 @@ stdenv.mkDerivation rec {
     cp ./assets/wall_4K.png $out/share/hyprland
     cp ./assets/wall_8K.png $out/share/hyprland
   '';
+
+  passthru.providedSessions = [ "hyprland" ];
 
   meta = with lib; {
     homepage = "https://github.com/vaxerski/Hyprland";
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "cachix": {
       "locked": {
-        "lastModified": 1651865313,
-        "narHash": "sha256-fLnmcBaVpJukNIzXZETxtJKLSSTf4LnBCWa/DwyGJBU=",
+        "lastModified": 1652530570,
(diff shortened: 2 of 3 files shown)


Narice:
yup, I'm looking at updating that on the nix flake, thanks for the tip!
from the logs it seems to be the same issue as in #77 so I'll fix the flake and drop a dependency since it's caused by nixpkgs-wayland for now being late

vaxerski:
lgtm, thanks!

Ashe:
I wanted to give this a test this morning, however weirdly only `hyprctl` is accessible via the command line, `hyprland` gives `command not found`. 

I've made a gist showing how I've used the overlay and tried to install the program in case I did something wrong. https://gist.github.com/Ashe/02ea1d4505bce49ba424f898059b3a18

When I finish work I'll come back and try to get it working, I really need this window manager in my life as I'm getting bored of sway!

I'd like to point out as well that I don't think there's any `nix` installation guides on the wiki even though some was provided in PRs. Maybe that's deliberate, just wanted to let you know!

vaxerski:
there aren't any nix tutorials because I know nothing about nix. If there is no hyprctl, you can always build it manually. just `make all` in the hyprctl directory and copy the hyprctl executable to wherever your PATH is.

Ashe:
Thanks for the response and I have to congratulate you on making such a wonderful window manager, I'm sure a lot of people will be very grateful for your hard work, myself included (when I can get it installed! :P )

> If there is no hyprctl, you can always build it manually

Sorry, my problem is that `hyperctl` is the ONLY thing I see when installing via this flake, meanwhile `hyprland` isn't available. Could be something I did wrong with regards to using the overlay or maybe its a packaging issue.

> there aren't any nix tutorials because I know nothing about nix.

Sorry I didn't want that to come across as criticism! It was more of a point to the people making changes to the `flake.nix` file. In the old PR #66 there's some installation instructions, so perhaps writing instructions for the overlay-based (or otherwise) installation method on the PR itself would be quite a nice way communicating things to put on the wiki. 

When I initially looked and saw that there were instructions for other distros, I kind of thought nix was missed out because it didn't work or maybe the nix files were there for development only. It was only when I searched PRs for when those files came in that I saw this discussion, and perhaps some people wouldn't think to look for PRs in this way.

vaxerski:
I can add them to the wiki, but I can't provide support for your issue. You can always compile and install manually, if flakes fail.

ghishadow:
I followed gist by Ashe, the command you get after adding flake is not `hyprland` it is `Hyprland`.
image
it is smooth even in VM. great work vaxerski

Narice:
Should that also be in the wiki just in case?
This can be probably added as a note to the change I proposed here #45.
Btw, I'm glad it works for everyone x)

Narice:
> Should that also be in the wiki just in case?
> This can be probably added as a note to the change I proposed here #45.

Oh my bad, that's already there in the `launching` section of the install guide, so nothing to change

PR #86, opened 2022-05-17
Nix: more tweaks to the flake

viperML:
Hello, I made some changes to the flake.nix to further comply with the ecosystem.
Before anything, I want to discuss it with the previous contributors of the flake (pinging @Narice @Tanish2002  #78 )
- `packages`: changed the output to `default`, which is the new default target instead of `#defaultPackage.<system>`. I also added an output for the custom wlroots build, in case some builds fail in the future.
- `apps`: the latest versions of nix automatically convert `packages` to `apps` by using the attribute `meta.mainProgram`, and it picks up `packages.<system>.default` or `apps.<system>.default`. So this output is not needed anymore.
- `inputs`: I removed FU (bloat). `nixpkgs-wayland`'s wlroots was just `nixpkgs`'s wlroots with an overriden `src`, which is exactly what we do in `packages.<system>.wlroots`, so I removed it too. The `nixpkgs` input has been changed to `nixos-unstable`, which I find more reasonable, given the nature of the software.
- `formatters`: a recently added output, which formats all nix files just by running `nix fmt`. I chose `alejandra` as the formatter (and formatted all the nixfiles with it), as I find that it produces the most diff-friendly code. Open to discussion...
- `overlay`: initially I wanted to entirely remove this output, but decided to keep it to not break people's setups. With such a simple flake layout, the downstream consumers can just use `(inputs).Hyprland.packages.x86_64-linux.default`, instead of adding the overlay, and then `pkgs.hyprland`. Overlays are cumbersome to use, compose, and introduce some other problems (read https://zimbatm.com/notes/1000-instances-of-nixpkgs ).

Following this, the wiki should be updated to include information for installation without overlays.

I also added a GitHub Workflow that automatically builds the package, in parallel with the Arch build. You can see it in action here: https://github.com/viperML/Hyprland/actions/runs/2338469084 . Ideally, another GH action would be needed to bump the `flake.lock` on schedule (weekly maybe?)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,19 +44,3 @@ jobs:
       - name: Build Hyprland with LEGACY_RENDERER
         run: |
           make legacyrenderer
-
-  nix:
-    name: "Build Hyprland (Nix)"
-    runs-on: ubuntu-latest
-    steps:
-    - name: Clone repository
-      uses: actions/checkout@v3
-    - name: Install nix
-      uses: cachix/install-nix-action@v17
-      with:
-        install_url: https://releases.nixos.org/nix/nix-2.8.0/install
-        extra_nix_config: |
-          auto-optimise-store = true
-          experimental-features = nix-command flakes
-    - name: Build HyprLand with default settings
-      run: nix build --print-build-logs
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,7 +32,7 @@ jobs:
       - name: Fix permissions for git
         run: |
           git config --global --add safe.directory /__w/Hyprland/Hyprland
-        
+
       - name: Checkout Hyprland
         uses: actions/checkout@v3
 
@@ -45,3 +45,18 @@ jobs:
         run: |
           make legacyrenderer
 
+  nix:
+    name: "Build Hyprland (Nix)"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v3
+    - name: Install nix
+      uses: cachix/install-nix-action@v17
+      with:
+        install_url: https://releases.nixos.org/nix/nix-2.8.0/install
+        extra_nix_config: |
+          auto-optimise-store = true
+          experimental-features = nix-command flakes
+    - name: Build HyprLand with default settings
+      run: nix build --print-build-logs
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 name: Build Hyprland
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 jobs:
   gcc:
     name: "Build Hyprland (Arch)"
--- a/.github/workflows/nix-build.yaml
+++ b/.github/workflows/nix-build.yaml
@@ -1,0 +1,19 @@
+name: Build Hyprland (Nix)
+
+on: [push, pull_request, workflow_dispatch]
+jobs:
+  nix:
+    name: "Build Hyprland (Nix)"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v3
+    - name: Install nix
+      uses: cachix/install-nix-action@v17
+      with:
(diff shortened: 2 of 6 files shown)


Tanish2002:
Looks good to me, didn't know about new default output attribute.

viperML:
@Tanish2002 I should have linked it before, they were introduced in nix 2.7 https://nixos.org/manual/nix/stable/release-notes/rl-2.7.html

Narice:
woaw thanks @viperML, I learned a lot just by reading your diff
I tried to find a way to do it without nixpkgs-wayland for the same reason you invoked but couldn't find anything online, I didn't know nixpkgs had a system attribute, I should really spend some time understanding the basics of nix x)
If you have some pointers, I'd gladly take them!

For the module, we can actually copy the sway one: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/sway.nix
That's actually what I based the wiki on tbh

viperML:
With the included module, the installation instruction in the Wiki should be simplified to:

---

## Nix

### With flakes
```nix
# flake.nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    hyprland = {
      url = "github:vaxerski/Hyprland";
      # build with your own instance of nixpkgs
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = { self, nixpkgs, hyprland }: {
    nixosConfigurations.HOSTNAME = nixpkgs.lib.nixosSystem {
      system = "x86_64-linux";
      modules = [
        hyprland.nixosModules.default 
        { programs.hyprland.enable = true; }
        # ...
      ];
    };
  };
  ```

### Without flakes

```nix
# configuration.nix
{config, pkgs, ...}: let
  flake-compat = builtins.fetchTarball "https://github.com/edolstra/flake-compat/archive/master.tar.gz";
  hyprland = import flake-compat {
    src = builtins.fetchTarball "https://github.com/viperML/Hyprland/archive/master.tar.gz";
  };
in {
  imports = [
    hyprland.defaultNix.nixosModules.default
  ];

  programs.hyprland.enable = true;
}
```

vaxerski:
is this now merge-ready?

vaxerski:
@viperML fix your CI to use wlroots-git before I snap it out of existence because that red cross is annoying

viperML:
The fix is trivial, just `nix flake update`.

But this will break again in X days, if the lockfile is left to rot. So, in the same fashion as many other projects, to set a CI action that updates the lockfile and publishes a PR, for a sensible time interval like 2 weeks (it could also be triggered manually).

Tanish2002:
I think a CI action like this would work..
https://github.com/fortuneteller2k/nixpkgs-f2k/blob/master/.github/workflows/update.yml

vaxerski:
fix your CI again, please

wlroots-git is no longer required (it's bundled) but make sure to have the dependencies if you're gonna remove it. It's still built, but now by Hyprland itself.

vaxerski:
@viperML

viperML:
IMHO using submodules like that is not a good solution. Now you are building against a pinned wlroots that the AUR PKGBUILD is not aware of.

viperML:
If you are installing your pinned wlroots to `/usr`, then you are in conflict with whatever wlroots installation is on the system right?

I haven't used Arch for a while, but I can see a solution like:
- Creating a new wlroots package (e.g. wlroots-hyprland) that pins wlroots and conflicts with it.

From the nix side of things, this is a problem that never existed, because each dependency is isolated (for example, we can have multiple versions of wlroots installed alongside).
The exact commit of the repo is in the lockfile https://github.com/vaxerski/Hyprland/blob/48a3b1c51435af24e12c63547f0a11eabd6ed2f1/flake.lock#L33

I propose to either you bump the lockfile alognside the git submodule, or add a git submodule bump to the existing flake update action (https://github.com/vaxerski/Hyprland/blob/main/.github/workflows/nix-update.yaml)

From my part, I don't have the time to each week rework the flake, and I would like to ask the previous contributors for some help ;) @Narice @Tanish2002 , or whoever wants to help too

viperML:
Yeah, the flake interface wont read the submodules by default.

I think I can come up with a solution later this week, as I don't have much time atm....

vaxerski:
Disabled the workflow for now. Feel free anyone to make a PR to fix it.

PR #109, opened 2022-05-26
nix: update workflow and bump flake

viperML:
Addresses the conversation in #86 

This PR adds an flake update action, so the `flake.lock` is not left for days without update.
This action can be manually triggered to update the dependencies (and the wlroots build) to their latest versions.

The timer is set to run every 2 weeks, which seems reasonable.


vaxerski:
lgtm, thank you!

PR #144, opened 2022-05-31
Fix Nix build and add updater

fufexan:
- move unnecessary toplevel files to nix/
- added patch that ignores the submodule (revert)
- add update script run by a workflow

`nix build` might actually update the input for some reason. I should look into hardcoding the rev in `flake.nix` using `sed`.

Regardless, Hyprland seems to work even with newer versions of `wlroots`.

As stated above.

Pinging @viperML @Narice @Tanish2002 for review.


fufexan:
As you can see, the updater doesn't modify the submodule, only polls its commit (to update the `flake.lock`)

fufexan:
Probably yes, but I think it'd work even without. The update script runs on push anyway, so it'll update in the next commit.

fufexan:
My bad. It doesn't run on push, so we'd need a mechanism to update it as well.

vaxerski:
submodule wont be at the latest wlroots pretty much ever, thats why I went with a submodule.

fufexan:
The last commit might be overkill/useless idk.

fufexan:
Then I'll have to create another mechanism that creates a patch modifying `"../*"` into `<wlroots/*>`

fufexan:
Honestly I wouldn't even bother doing this if submodules were properly implemented.

fufexan:
> Submodules are implemented properly by me, as far as I'm aware. 
By you, yes. By nix, nope.

> Regarding patching the paths, sed -i?
Yes, that's what I thought too.

fufexan:
Now the update workflow will run each push, and only update when the submodule is updated. In case nixpkgs lags behind too much (because of no submodule updates), we can run the workflow manually, which will only update nixpkgs.

With this, I think it can be merged and the nix-build and nix-update workflows re-enabled.

fufexan:
Wait, I think the workflow isn't quite right. It will create an empty commit even if not updating. I'll have to fix it.

vaxerski:
If the workflow makes a lot of commits i'll be pissed.

fufexan:
My bad, I didn't read the auto-commit readme before commenting that. It will not commit anything unless the files have changed. So it'll report everything as fine even when not updating.

vaxerski:
is this merge-ready?

fufexan:
Yes, I think so. Do I have to explicitly enable the workflows anywhere or is that something you have to configure in the repo settings?

vaxerski:
I do that

PR #167, opened 2022-06-06
Fix error: attribute 'lastModifiedDate' missing

JamesReynolds:
The above error appears on NixOS 22.05 when using the flake. It appears that lastModifiedDate has been renamed to lastModified, although I can't find a version at which this change happened:
https://github.com/NixOS/templates/issues/13

This fix only applies to some versions of flakes, on the previous versions the old behaviour is preserved. I've tested on nix 2.8.1.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,11 +33,11 @@
   in {
     packages = genSystems (system: {
       wlroots = pkgsFor.${system}.wlroots.overrideAttrs (prev: {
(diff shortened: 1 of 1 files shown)


fufexan (approved):

vaxerski:
ok, merging, thanks!

JamesReynolds:
Thanks for the quick merge guys!

PR #187, opened 2022-06-11
untested: Add help options

godalming123:

It adds help and version options

This has not been tested even compiling, and I barely know c++

It needs testing but should work

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -22,15 +22,11 @@ int main(int argc, char** argv) {
         ignoreSudo = true;
         if (!strcmp(argv[2], "no-warn-message"))
             ignoreSudoWarnMessage = true;
-    else if (arg1 == "version") {
-        std::cout << "v0.5.0beta\n";
-        return 0;
-    }
     else {
         std::cout << "Available commands:\n";
         std::cout << "  help                 - show this help text\n";
-        std::cout << "  version              - show version\n";
         std::cout << "  --i-am-really-stupid - ignore sudo run check\n";
+        std::cout << "More commands can be found by running hyprctrl\n";
         std::cout << "When using --i-am-really-stupid you can ignore the warning message and startup delay with additional argument no-warn-message\n";
         return 0;
      }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -15,16 +15,26 @@ int main(int argc, char** argv) {
         RIP("XDG_RUNTIME_DIR not set!");
 
     // parse some args
-    for (int i = 1; i < argc; ++i) {
-        if (!strcmp(argv[i], "--i-am-really-stupid"))
-            ignoreSudo = true;
-    }
+    if (!strcmp(argv[1], "i-am-really-stupid"))
+        ignoreSudo = true;
+    else if (!strcmp(argv[1], "help"))
+        std::cout << "Available commands:\n";
+        std::cout << "  help                 - show this help text\n";
+        std::cout << "  version              - show version\n";
+        std::cout << "  --i-am-really-stupid - ignore sudo run check\n";
+        return 0;
+    else if (!strcmp(argv[1], "vesrsion"))
+        std::cout << "V5.0\n";
+        return 0;
+    else
+        std::cout << "That argument does not exist\n";
+        return 1;
 
     system("mkdir -p /tmp/hypr");
 
     if (!ignoreSudo) {
         if (Init::isSudo()) {
-            std::cout << "Hyprland shall not be run as the root user. If you really want to, use the --i-am-really-stupid flag.\n";
+            std::cout << "Hyprland shall not be run as the root user. If you really want to, use the i-am-really-stupid flag.\n";
             return 1;
     	}
     } else {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -8,6 +8,7 @@
 // I am a bad bad boy and have used some global vars here,
 // just for this file
 bool ignoreSudo = false;
+bool ignoreSudoWarnMessage = false;
 
 int main(int argc, char** argv) {
 
@@ -15,30 +16,34 @@ int main(int argc, char** argv) {
         RIP("XDG_RUNTIME_DIR not set!");
 
     // parse some args
-    if (!strcmp(argv[1], "i-am-really-stupid"))
+    std::string arg1 = argv[1];
+
+    if (arg1 == "--i-am-really-stupid")
         ignoreSudo = true;
-    else if (!strcmp(argv[1], "help"))
+        if (!strcmp(argv[2], "no-warn-message"))
+            ignoreSudoWarnMessage = true;
+    else if (arg1 == "version") {
+        std::cout << "v0.5.0beta\n";
+        return 0;
+    }
+    else {
         std::cout << "Available commands:\n";
         std::cout << "  help                 - show this help text\n";
         std::cout << "  version              - show version\n";
         std::cout << "  --i-am-really-stupid - ignore sudo run check\n";
+        std::cout << "When using --i-am-really-stupid you can ignore the warning message and startup delay with additional argument no-warn-message\n";
         return 0;
-    else if (!strcmp(argv[1], "vesrsion"))
-        std::cout << "V5.0\n";
-        return 0;
-    else
-        std::cout << "That argument does not exist\n";
-        std::cout << "For a list of available arguments run hyprland help\n";
-        return 1;
+     }
 
     system("mkdir -p /tmp/hypr");
 
     if (!ignoreSudo) {
         if (Init::isSudo()) {
-            std::cout << "Hyprland shall not be run as the root user. If you really want to, use the i-am-really-stupid flag.\n";
+            std::cout << "Hyprland shall not be run as the root user. If you really want to, use the --i-am-really-stupid flag.\n";
             return 1;
     	}
-    } else {
+    }
+    else if (!ignoreSudoWarnMessage) {
         std::cout << "Running with ignored root checks, I surely hope you know what you're doing.\n";
         sleep(1);
     }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -28,6 +28,7 @@ int main(int argc, char** argv) {
         return 0;
     else
         std::cout << "That argument does not exist\n";
+        std::cout << "For a list of available arguments run hyprland help\n";
         return 1;
 
     system("mkdir -p /tmp/hypr");


src/main.cpp
```diff
@@ -15,16 +15,27 @@ int main(int argc, char** argv) {
         RIP("XDG_RUNTIME_DIR not set!");
 
     // parse some args
-    for (int i = 1; i < argc; ++i) {
-        if (!strcmp(argv[i], "--i-am-really-stupid"))
-            ignoreSudo = true;
-    }
+    if (!strcmp(argv[1], "i-am-really-stupid"))
+        ignoreSudo = true;
+    else if (!strcmp(argv[1], "help"))
```
fufexan, line 20: I think it would be easier to convert `argv` to `string` and use the associated `==` to compare them. This is C++ after all.

src/main.cpp
```diff
@@ -15,16 +15,27 @@ int main(int argc, char** argv) {
...
+        std::cout << "Available commands:\n";
+        std::cout << "  help                 - show this help text\n";
+        std::cout << "  version              - show version\n";
+        std::cout << "  --i-am-really-stupid - ignore sudo run check\n";
```
fufexan, line 24: There seem to be inconsistencies in the code. In one place it's `--i-am-really-stupid`, in another it's `i-am-really-stupid`. I think we should leave it with `--` to maintain backwards compatibility. Also this isn't a usual command so it's fine.

src/main.cpp
```diff
@@ -15,16 +15,27 @@ int main(int argc, char** argv) {
...
+        return 0;
+    else if (!strcmp(argv[1], "vesrsion"))
+        std::cout << "V5.0\n";
```
fufexan, line 27: Not sure how helpful this would be, as `hyprctl version` already does this. If you want pretty versions I think we could look up whether a certain ref has a name, and if it does print that instead.

src/main.cpp
```diff
@@ -15,16 +15,27 @@ int main(int argc, char** argv) {
...
+        return 0;
+    else
+        std::cout << "That argument does not exist\n";
```
fufexan, line 30: Or we could just print the help menu by default and not handle `help` explicitly.

fufexan (changes requested):
I see you used `if/else if/else`. In this case I feel like a `switch` is more appropriate, since we're dealing with only one variable (`argv[1]`). Also, for `if/else if/else`, if there is more than one line to be executed for a condition, you should put them in a block (`{ }`).

vaxerski:
@fufexan I'll leave this to ya, I'm on holiday. Feel free to ping me once you find this merge ready.

fufexan:
```diff
diff --git a/src/main.cpp b/src/main.cpp
index a437430..2653099 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -14,10 +14,21 @@ int main(int argc, char** argv) {
     if (!getenv("XDG_RUNTIME_DIR"))
         RIP("XDG_RUNTIME_DIR not set!");
 
-    // parse some args
-    for (int i = 1; i < argc; ++i) {
-        if (!strcmp(argv[i], "--i-am-really-stupid"))
-            ignoreSudo = true;
+    // parse args
+    std::string arg1 = argv[1];
+
+    if (arg1 == "--i-am-really-stupid")
+        ignoreSudo = true;
+    else if (arg1 == "version") {
+        std::cout << "v0.5.0beta\n";
+        return 0;
+    }
+    else {
+        std::cout << "Available commands:\n";
+        std::cout << "  help                 - show this help text\n";
+        std::cout << "  version              - show version\n";
+        std::cout << "  --i-am-really-stupid - ignore sudo run check\n";
+        return 0;
     }
 
     system("mkdir -p /tmp/hypr");
```
This patch seems to work, but isn't ideal.

vaxerski:
thats not how versioning works, please see hyprctl version and copy that. I don't want to bump it manually.

fufexan:
Indeed, what I wrote was just a PoC. Currently writing the code that uses the tag name if a commit happens to be a tag.

tazimamatovzeka:
I missed something about what I'm talking about

src/main.cpp
tazimamatovzeka: I missed something about what I'm talking about

fufexan:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index a3b3584..8e8a6e0 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,18 @@ execute_process(
     OUTPUT_VARIABLE GIT_COMMIT_HASH
     OUTPUT_STRIP_TRAILING_WHITESPACE)
 
+execute_process(
+    COMMAND bash -c "git show-ref --tags ${GIT_COMMIT_HASH}| tail -n 1 | head -c 7"
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    OUTPUT_VARIABLE GIT_TAG_HASH
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+execute_process(
+    COMMAND bash -c "git show-ref --tags ${GIT_COMMIT_HASH}| tail -n 1 | tail -c +19"
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    OUTPUT_VARIABLE GIT_TAG_NAME
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
 execute_process(
     COMMAND bash -c "git show ${GIT_COMMIT_HASH} | head -n 5 | tail -n 1"
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
@@ -67,6 +79,8 @@ ENDIF(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES DEBUG)
 target_compile_definitions(Hyprland PRIVATE "-DGIT_COMMIT_HASH=\"${GIT_COMMIT_HASH}\"")
 target_compile_definitions(Hyprland PRIVATE "-DGIT_BRANCH=\"${GIT_BRANCH}\"")
 target_compile_definitions(Hyprland PRIVATE "-DGIT_COMMIT_MESSAGE=\"${GIT_COMMIT_MESSAGE}\"")
+target_compile_definitions(Hyprland PRIVATE "-DGIT_TAG_HASH=\"${GIT_TAG_HASH}\"")
+target_compile_definitions(Hyprland PRIVATE "-DGIT_TAG_NAME=\"${GIT_TAG_NAME}\"")
 target_compile_definitions(Hyprland PRIVATE "-DGIT_DIRTY=\"${GIT_DIRTY}\"")
 
 target_link_libraries(Hyprland rt)
diff --git a/src/debug/HyprCtl.cpp b/src/debug/HyprCtl.cpp
index 83656e1..5d8d16f 100644
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1,4 +1,5 @@
 #include "HyprCtl.hpp"
+#include "../helpers/MiscFunctions.hpp"
 
 #include <netinet/in.h>
 #include <stdio.h>
@@ -105,22 +106,6 @@ std::string devicesRequest() {
     return result;
 }
 
-std::string versionRequest() {
-    std::string result = "Hyprland, built from branch " + std::string(GIT_BRANCH) + " at commit " + GIT_COMMIT_HASH + GIT_DIRTY + " (" + GIT_COMMIT_MESSAGE + ").\nflags: (if any)\n";
-
-#ifdef LEGACY_RENDERER
-    result += "legacyrenderer\n";
-#endif
-#ifndef NDEBUG
-    result += "debug\n";
-#endif
-#ifdef NO_XWAYLAND
-    result += "no xwayland\n";
-#endif
-
-    return result;
-}
-
 std::string dispatchRequest(std::string in) {
     // get rid of the dispatch keyword
     in = in.substr(in.find_first_of(' ') + 1);
diff --git a/src/defines.hpp b/src/defines.hpp
index 18db395..00beb52 100644
--- a/src/defines.hpp
+++ b/src/defines.hpp
@@ -62,6 +62,12 @@
 #ifndef GIT_COMMIT_MESSAGE
 #define GIT_COMMIT_MESSAGE "?"
 #endif
+#ifndef GIT_TAG_HASH
+#define GIT_TAG_HASH "?"
+#endif
+#ifndef GIT_TAG_NAME
+#define GIT_TAG_NAME "?"
+#endif
 #ifndef GIT_DIRTY
 #define GIT_DIRTY "?"
 #endif
diff --git a/src/helpers/MiscFunctions.cpp b/src/helpers/MiscFunctions.cpp
index 6f9b465..81da811 100644
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -219,4 +219,23 @@ float vecToRectDistanceSquared(const Vector2D& vec, const Vector2D& p1, const Ve
     const float DX = std::max((double)0, std::max(p1.x - vec.x, vec.x - p2.x));
     const float DY = std::max((double)0, std::max(p1.y - vec.y, vec.y - p2.y));
     return DX * DX + DY * DY;
-}
\ No newline at end of file
+}
+
+std::string versionRequest() {
+    bool isTag = !strcmp(GIT_COMMIT_HASH, GIT_TAG_HASH);
+    std::string result = "Hyprland, built from branch " + std::string(GIT_BRANCH)
+        + " at commit " + GIT_COMMIT_HASH + (isTag ? GIT_TAG_NAME : GIT_DIRTY) // can't be both a tag and dirty
+        + " (" + GIT_COMMIT_MESSAGE + ").\nflags: (if any)\n";
+
+#ifdef LEGACY_RENDERER
+    result += "legacyrenderer\n";
+#endif
+#ifndef NDEBUG
+    result += "debug\n";
+#endif
+#ifdef NO_XWAYLAND
+    result += "no xwayland\n";
+#endif
+
+    return result;
+}
diff --git a/src/helpers/MiscFunctions.hpp b/src/helpers/MiscFunctions.hpp
index eed3670..1d6be60 100644
--- a/src/helpers/MiscFunctions.hpp
+++ b/src/helpers/MiscFunctions.hpp
@@ -12,4 +12,5 @@ bool isDirection(const std::string&);
 int getWorkspaceIDFromString(const std::string&, std::string&);
 float vecToRectDistanceSquared(const Vector2D& vec, const Vector2D& p1, const Vector2D& p2);
 
-float getPlusMinusKeywordResult(std::string in, float relative);
\ No newline at end of file
+float getPlusMinusKeywordResult(std::string in, float relative);
+std::string versionRequest();
diff --git a/src/main.cpp b/src/main.cpp
index a437430..5eb20f5 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2,6 +2,7 @@
 #include "debug/Log.hpp"
 #include "Compositor.hpp"
 #include "config/ConfigManager.hpp"
+#include "helpers/MiscFunctions.hpp"
 #include "init/initHelpers.hpp"
 #include <iostream>
 
@@ -14,10 +15,21 @@ int main(int argc, char** argv) {
     if (!getenv("XDG_RUNTIME_DIR"))
         RIP("XDG_RUNTIME_DIR not set!");
 
-    // parse some args
-    for (int i = 1; i < argc; ++i) {
-        if (!strcmp(argv[i], "--i-am-really-stupid"))
-            ignoreSudo = true;
+    // parse args
+    std::string arg1 = argv[1];
+
+    if (arg1 == "--i-am-really-stupid")
+        ignoreSudo = true;
+    else if (arg1 == "version") {
+        std::cout << versionRequest();
+        return 0;
+    }
+    else {
+        std::cout << "Available commands:\n";
+        std::cout << "  help                 - show this help text\n";
+        std::cout << "  version              - show version\n";
+        std::cout << "  --i-am-really-stupid - ignore sudo run check\n";
+        return 0;
     }
 
     system("mkdir -p /tmp/hypr");
```
Oops, I forgot to comment this. It may not be perfect, but I'm accepting criticism.

tazimamatovzeka (approved):

fufexan:
@godalming123 consider adding my patch to the PR, it's better suited than what you have now.

godalming123:
@fufexan I have removed the version option in hyprland and instead mentioned that hyprctrl has some options does that suffice?

fufexan:
Sure that works, but do you still want the version or only the commit?

vaxerski:
idek what is the point of this PR anymore. It's first of all: wrong, second: whats that other option and why? third: what

godalming123:
The other option disables the warning text when you run with `--i-am-an-idiot`

vaxerski:
why does it have to be the second arg? why is it without --?

code is unsafe, will often crash. You removed a perfectly fine arg parser and implemented an unsafe, incorrect one.

It's clear you have no idea what you are doing.

Furthermore, it has been clearly stated in the contributing guidelines, that your code should be **tested**, which yours clearly isn't because you'd catch that issue beforehand.

I don't even know why we keep going on here.

PR #195, opened 2022-06-11
Add natural_scrolling option for touchpads only

PowerBall253:
This PR makes the existing `input:natural_scroll` setting affect only the mouse, and adds a separate `input:touchpad:natural_scroll` for touchpads.


vaxerski:
tested?

PowerBall253:
Tested on my main laptop, it works. Will try to test on my other laptop real quick.

PowerBall253:
Finished testing, works on both of my laptops :+1:

vaxerski:
ok, thank you!

PR #197, opened 2022-06-12
Add loose follow_mouse option

spectreseven1138:
Allows the `follow_mouse` config option to be set to 2, which makes it so that moving the cursor doesn't focus windows, but still calls `notify_enter` so that the windows interact with the cursor correctly (such as for scrolling and hover effects). Keyboard events like typing and navigation still occur in the focused window. Tested and works on Arch.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -167,14 +167,14 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
     Vector2D surfaceLocal = surfacePos == Vector2D(-1337, -1337) ? surfaceCoords : mouseCoords - surfacePos;
 
     if (pFoundWindow) {
-        const int FOLLOWMOUSE = g_pConfigManager->getInt("input:follow_mouse");
-        if (FOLLOWMOUSE != 1 && !refocus) {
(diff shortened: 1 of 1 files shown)


src/managers/input/InputManager.cpp
```diff
@@ -167,12 +167,16 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
     Vector2D surfaceLocal = surfacePos == Vector2D(-1337, -1337) ? surfaceCoords : mouseCoords - surfacePos;
 
     if (pFoundWindow) {
-        if (g_pConfigManager->getInt("input:follow_mouse") == 0 && !refocus) {
+        const int FOLLOWMOUSE = g_pConfigManager->getInt("input:follow_mouse");
```
vaxerski, line 170: can you make it speedier by making it

```
static auto *const PFOLLOWMOUSE = &g_pConfigManager->getrawptr(name)->intvalue;

if (*PFOLLOWMOUSE)

...
```

please?

spectreseven1138: Done and tested 👍

fufexan (approved):

vaxerski:
thank you!

PR #202, opened 2022-06-12
Add movewindowinv dispatcher

spectreseven1138:
Adds the `movewindowinv` dispatcher, which is similar to `movewindow` but moves all windows within the active workspace _except_ the active window (and takes only a monitor as its argument). Useful for when you want to quickly give space to particular window. 

This also moves most of the code from the `moveActiveToWorkspace` method into a new method so that it can be used by the new dispatcher. I've tested both the `movewindowinv` dispatcher and the affected `movewindow` dispatcher.

Example config:
```
bind=SUPER,left,movewindowinv,l
bind=SUPER,right,movewindowinv,r
```

 
Demo video using the above config (I couldn't get the second monitor to be recorded, but that's where the windows are going):

https://user-images.githubusercontent.com/44363049/173507834-f6742ead-b32b-4622-801f-e3a143ea57e3.mp4

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -996,11 +996,75 @@ void CCompositor::updateWindowBorderColor(CWindow* pWindow) {
         pWindow->m_cRealBorderColor = CColor(pWindow == m_pLastWindow ? *ACTIVECOL : *INACTIVECOL);
 }
 
-void CCompositor::moveWindowToWorkspace(CWindow* pWindow, const std::string& work) {
-    m_pLastWindow = pWindow;
-    g_pKeybindManager->moveActiveToWorkspace(work);
+void CCompositor::moveWindowToWorkspace(CWindow* pWindow, const int pWorkspaceID) {
+
+    if (!g_pCompositor->windowValidMapped(pWindow))
+        return;
+
+    const auto OLDWORKSPACE = g_pCompositor->getWorkspaceByID(pWindow->m_iWorkspaceID);
+
+    if (pWorkspaceID == pWindow->m_iWorkspaceID) {
+        Debug::log(LOG, "Not moving to workspace because it didn't change.");
+        return;
+    }
+
+    g_pLayoutManager->getCurrentLayout()->onWindowRemoved(pWindow);
+
+    g_pKeybindManager->changeworkspace(std::to_string(pWorkspaceID));
+
+    const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(pWorkspaceID);
+
+    if (PWORKSPACE == OLDWORKSPACE) {
+        Debug::log(LOG, "Not moving to workspace because it didn't change.");
+        return;
+    }
+
+    if (!PWORKSPACE) {
+        Debug::log(ERR, "Workspace null in moveWindowToWorkspace?");
+        return;
+    }
+
+    OLDWORKSPACE->m_bHasFullscreenWindow = false;
+
+    pWindow->m_iWorkspaceID = PWORKSPACE->m_iID;
+    pWindow->m_iMonitorID = PWORKSPACE->m_iMonitorID;
+    pWindow->m_bIsFullscreen = false;
+
+    if (PWORKSPACE->m_bHasFullscreenWindow) {
+        g_pCompositor->getFullscreenWindowOnWorkspace(PWORKSPACE->m_iID)->m_bIsFullscreen = false;
+        PWORKSPACE->m_bHasFullscreenWindow = false;
+    }
+
+    // Hack: So that the layout doesnt find our window at the cursor
+    pWindow->m_vPosition = Vector2D(-42069, -42069);
+    
+    // Save the real position and size because the layout might set its own
+    const auto PSAVEDSIZE = pWindow->m_vRealSize.vec();
+    const auto PSAVEDPOS = pWindow->m_vRealPosition.vec();
+    g_pLayoutManager->getCurrentLayout()->onWindowCreated(pWindow);
+    // and restore it
+    pWindow->m_vRealPosition.setValue(PSAVEDPOS);
+    pWindow->m_vRealSize.setValue(PSAVEDSIZE);
+
+    if (pWindow->m_bIsFloating) {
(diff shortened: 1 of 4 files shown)


vaxerski:
waiting for checks to succeed, aka freedesktop gitlab to relive itself

vaxerski:
can someone like @fufexan or @ThatOneCalculator verify this? thanks.

ThatOneCalculator:
Testing now

ThatOneCalculator:
Doesn't work, has no effect.

src/managers/KeybindManager.hpp
```diff
@@ -29,6 +29,7 @@ class CKeybindManager {
     std::list<SKeybind> m_lKeybinds;
 
     bool                handleInternalKeybinds(xkb_keysym_t);
+    static void         moveWindowToWorkspace(CWindow*, int);
```
vaxerski, line 32: ayo sir what in the name of all thats holy is this

spectreseven1138: That's the new method I mentioned with most of the code from `moveActiveToWorkspace`, since most of it is also needed by the `movewindowinv` dispatcher. Should I have put it somewhere else?

spectreseven1138:
> Doesn't work, has no effect.

Strange, are you sure you're using the inv dispatcher and not the original `movewindow` one?

ThatOneCalculator:
I made sure I was using the right option.

src/managers/KeybindManager.hpp
vaxerski: no global methods there, move it to CCompositor

spectreseven1138: Ah, I didn't realise it already existed there. Should I replace the existing CCompositor method (which doesn't seem to be used anywhere) or create a new one?

spectreseven1138:
> I made sure I was using the right option.

Can you check the log to see if it's being triggered at all, or if there's an error?

src/managers/KeybindManager.hpp
vaxerski: do whatever you see fit

vaxerski:
@fufexan maybe you can check this?

src/Compositor.cpp
```diff
@@ -996,11 +996,75 @@ void CCompositor::updateWindowBorderColor(CWindow* pWindow) {
...
+        return;
+
+    const auto OLDWORKSPACE = g_pCompositor->getWorkspaceByID(pWindow->m_iWorkspaceID);
+
+    if (pWorkspaceID == pWindow->m_iWorkspaceID) {
+        Debug::log(LOG, "Not moving to workspace because it didn't change.");
+        return;
+    }
+
+    g_pLayoutManager->getCurrentLayout()->onWindowRemoved(pWindow);
```
vaxerski, line 1011: we should do it after the return checks to avoid just yeeting the window to narnia

src/managers/KeybindManager.cpp
```diff
@@ -562,6 +506,29 @@ void CKeybindManager::moveActiveTo(std::string args) {
     g_pLayoutManager->getCurrentLayout()->switchWindows(PLASTWINDOW, PWINDOWTOCHANGETO);
 }
 
+void CKeybindManager::moveInactiveTo(std::string args) {
+    const auto LASTMONITOR = g_pCompositor->m_pLastMonitor;
+    const auto PACTIVE = g_pCompositor->m_pLastWindow;
+    const auto CURSOR_COORDS = Vector2D(g_pCompositor->m_sWLRCursor->x, g_pCompositor->m_sWLRCursor->y);
```
vaxerski, line 512: naming convention, no snake, also g_pInputManager->getMouseCoordsInternal

src/managers/KeybindManager.cpp
```diff
@@ -562,6 +506,29 @@ void CKeybindManager::moveActiveTo(std::string args) {
...
+
+    for (auto& window : g_pCompositor->m_lWindows) {
+        if (window == *PACTIVE || window.m_iWorkspaceID != PACTIVE->m_iWorkspaceID) {
+            continue;
+        }
+        g_pCompositor->moveWindowToWorkspace(&window, g_pCompositor->m_pLastMonitor->activeWorkspace);
+    }
+
+    g_pCompositor->focusWindow(PACTIVE);
+    wlr_cursor_warp(g_pCompositor->m_sWLRCursor, g_pCompositor->m_sSeat.mouse->mouse, CURSOR_COORDS.x, CURSOR_COORDS.y);
```
vaxerski, line 529: consider just warping and calling refocus()

src/managers/KeybindManager.cpp
```diff
@@ -562,6 +506,29 @@ void CKeybindManager::moveActiveTo(std::string args) {
     g_pLayoutManager->getCurrentLayout()->switchWindows(PLASTWINDOW, PWINDOWTOCHANGETO);
 }
 
+void CKeybindManager::moveInactiveTo(std::string args) {
+    const auto LASTMONITOR = g_pCompositor->m_pLastMonitor;
+    const auto PACTIVE = g_pCompositor->m_pLastWindow;
```
vaxerski, line 511: can be null / invalid

src/managers/KeybindManager.cpp
```diff
@@ -562,6 +506,29 @@ void CKeybindManager::moveActiveTo(std::string args) {
...
+
+    focusMonitor(args);
+
+    if (LASTMONITOR == g_pCompositor->m_pLastMonitor) {
+        Debug::log(ERR, "moveInactiveTo: moving to an invalid mon");
+        return;
+    }
+
+    for (auto& window : g_pCompositor->m_lWindows) {
+        if (window == *PACTIVE || window.m_iWorkspaceID != PACTIVE->m_iWorkspaceID) {
```
vaxerski, line 522: `|| !window.m_bismapped` ?

yavko:
Any progress on this?

VexfulNiko:
Any progress?

yavko:
Any progress v2?

Dickby:
Should i fix this?

yavko:
> Should i fix this?

It seems the developer has abandoned the PR so if you want to go ahead

vaxerski:
I will close this one. If anyone wants to make a similiar one, feel free.

PR #223, opened 2022-06-15
Create config directory if it doesn't exist

PowerBall253:
Creates the config directory (`~/.config/hypr/`) if it doesn't exist, so that the auto-generated config can be written without errors.


vaxerski:
thanks!

PR #239, opened 2022-06-19
Contrain cursor to center of window for games on multi-monitor setups. 

SebOuellette:

In games, I noticed that when the cursor was supposed to be constrained, it would accidentally fly past the like 1 pixel hitbox for the constraint, and onto my second monitor, tabbing out of the game. So every time I looked to the right, I would close focus on the game. This PR changes the constraint to a single point in the center of the window, so the user's mouse can't escape any longer. It not works a lot better in games, from what I've noticed, on par with KDE's system for the same thing. 

I also noticed that angular camera movement was also weird, and that's fixed too, but that might just be because I changed from accelerated mouse movement to forced linear. 

This obviously locks the cursor to the center of the screen while constrained, so if this breaks some workflow I'm not aware of, please let me know. This was created to play games better, and that's what it successfully does. 

Ready for merging!


vaxerski:
okay listen up

this is how it used to work, but it causes MASSIVE issues in some games. It used to work (the system thats currently in place) but it broke some day and idk what day that was.

Thus, if you want to fix the constraints inb4 i get back home from my amazing sacred and holy vacation, make it work just how it's supposed from the code. 

The problem is NOT the cursor flying to the second monitor. 

I repeat: it is NOT.

I think it's because we're moving the cursor with wlr_cursor_move outside the constraint surface. Find a way to move it AFTER adjusting for constraints. (remmeber about refocus() and moveAbsolute!)

You can also try to narrow down the commit that broke this. Search through the issues I believe there was one about this, and I should've mentioned the commit there. If my memory is failing, just scramble through the commit history.

SebOuellette:
Ah okay. I'll work on actually fixing this then

SebOuellette:
Alright so the mouse is properly being constrained now, however there's still a bug where games aren't registering mouse movement unless you move it fast enough. 
It only happens when the mouse is being constrained. 

Same thing happens with and without force linear movement, so that's not the problem. 

Not sure if this is at all relevant, but it was happening in the official version even before I made my PR as well so it wasn't me :P

SebOuellette:
Cancel that nvm, needs some more work. I'm doing more in depth testing so I'll advise when it passes them all

vaxerski:
ping me once you think this PR is ready. Thanks for your work!

SebOuellette:
@vaxerski It's working next to fullscreen applications! 

Also the weird deadzone bug I was talking about only happens when the mouse sensitivity is set to something lower than 1.0, so that might be another issue..? unsure if it's documented but I can make a new issue if you'd like. But yeah it's not related to this, and it's working now that I set it to 1.0. 

I'm happy to help :D I love Hyprland!

vaxerski:
I use 0.34 sens, let me test this on actual metal

vaxerski:
works as expected for me. Thank you!

SebOuellette:
Woo! Thanks for your hints and help!

PR #227, opened 2022-06-16
Nix: use meson for builds

fufexan:
#### Describe your PR, what does it fix/add?
- Use meson in Nix builds.
- Update nixpkgs and wlroots
- Change wlroots url to Kainoa's mirror

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

Depends on #205

#### Is it ready for merging, or does it need work?

Pretty sure it's ready for merging as soon as #205 is merged.

Requesting reviews from @viperML @Tanish2002 @Narice

PR #241, opened 2022-06-20
Nix: refactor packages and overlays

viperML:
The code that I added about the flake assumed that the overlay was deprecated and nobody would use it. But it turns out that more people that I thought are using it, and it is very poorly written.

This can be already merged, and will require a rebase in #227

Pinging @viperML @Tanish2002 @Narice for a second look


Tanish2002:
Looks good to me!

fufexan (approved):
LGTM

PR #242, opened 2022-06-20
Added a multiplier for the width to height ratio

PinkFlufflyLlama:

This resolves my issue here: https://github.com/vaxerski/Hyprland/issues/233

I've named the new setting "dwindle:split_width_multiplier" which I think is pretty descriptive, but can be changed.

Compiles and works


PinkFlufflyLlama:
Hmm, put a pin in that for a moment, it seems like the dragging of windows around also needs to depend on that multiplier for it to be have propperly

vaxerski:
ready?

Caleb-T-Owens:
Yes, sorry for the late response

PR #265, opened 2022-06-26
Scroll workspaces with Super+MouseWheel

xDarksome:
Made it similar to the current implementation in Gnome.

I am not sure whether it should be the default behavior, but as I understand it the current configuration requires considerable changes to support mouse button bindings. And I write c++ on pure intuition, so won't be taking it for a spin :laughing:


fufexan:
So have you tested it? Does it work?

src/managers/input/InputManager.cpp
```diff
@@ -253,6 +253,24 @@ void CInputManager::onMouseButton(wlr_pointer_button_event* e) {
...
 }
 
+void CInputManager::onMouseWheel(wlr_pointer_axis_event* e) {
+    const auto PKEYBOARD = wlr_seat_get_keyboard(g_pCompositor->m_sSeat.seat);
+
+    if (wlr_keyboard_get_modifiers(PKEYBOARD) == (uint32_t)g_pConfigManager->getInt("general:main_mod_internal") && 
+        e->source == WLR_AXIS_SOURCE_WHEEL && e->orientation == WLR_AXIS_ORIENTATION_VERTICAL) {
+            
+        if (e->delta < 0) { 
+            g_pKeybindManager->m_mDispatchers["workspace"]("-1");
```
vaxerski, line 263: why not use the appropriate func?

xDarksome:
> So have you tested it? Does it work?

yep, been running it for a week

src/managers/input/InputManager.cpp
xDarksome: @vaxerski 

> why not use the appropriate func?

Which one do you mean?  `KeybindManager::changeworkspace` is private and I suppose was meant to be used internally as almost everything else in that class.

Why exactly only the map of functions is exposed? While lurking the code I was thinking why there is no class that exposes public functions for wm manipulations + KeybindManager just using them?

Again, my knowledge of c++ is limited, so maybe I get something wrong.

vaxerski: no yeah i forgot this is outside keybindmanager. fine then

they are private to not incentivise their usage outside of keybindmanager

vaxerski:
woops forgor about this, @fufexan u tested?

fufexan:
No, but I will in a moment

fufexan:
It works well, but I have a suggestion.

Currently it will cycle through all workspaces, whether they're empty or not. What I propose is having a `general:scroll_empty_workspaces=0/1` or similar option which would control whether scrolling cycles empty ws or just active ones (with windows inside).

vaxerski:
> Currently it will cycle through all workspaces, whether they're empty or not. What I propose is having a general:scroll_empty_workspaces=0/1 or similar option which would control whether scrolling cycles empty ws or just active ones (with windows inside).

to @xDarksome: replace +1 with m+1 for that :P

vaxerski:
actually m+1 only says "next one on this monitor" if you wanna add a next visible one you gotta code that yourself

xDarksome:
@vaxerski ah, I see. Was going to test it first anyway, but I am having some issues with updating my nixos flake to the latest Hyprland version. I will look into the magic string stuff once I fix the flake.

fufexan:
@xDarksome did you manage to build it properly?

You can point the flake to the latest commit from this PR, and github will make it work even if it's coming from a fork.

xDarksome:
@fufexan @vaxerski 

>What I propose is having a general:scroll_empty_workspaces=0/1

So, do we want to change the behavior of  `changeworkspace("+1")`? It will affect the switching by keyboard as well.

I am not sure it should be done in this PR.

Also, I think it should be allowed to switch to at least one empty workspace (next one). More formally: 
```
if current_workspace.has_windows() {
  switch_to_next_workspace()
} else {
  // noop
}
```

vaxerski:
no, scrolling should not affect the current system at all. You can just write your own little func to find the next populated ws. And no, scrolling should scroll through only opened wspcs

xDarksome:
Ah, @fufexan was looking into the use case in which there are gaps between non-empty workspaces, I got it now.

However, I don't think that "scrolling" should be it's own thing. It's still keys, just fancy ones. Ideally it should be possible to bind any action to `mouse_up`/`mouse_down`.

A more future proof solution is to introduce something like `movetopopulatedworkspace(..)`.

And then temporary allow to bind `mouseup`/`mousedown` only to 

`workspace,{+|-}n`
```
bind=SUPER,mouseup,populatedworkspace,+1
```
`populatedworkspace,{+|-}n`
```
bind=SUPER,mouseup,workspace,+1
```

fufexan:
I think the PR is fine as-is. We can always break compatibility in the future, when we add support for mouse buttons.

vaxerski:
> A more future proof solution is to introduce something like movetopopulatedworkspace(..).

very stupid. Just implement e.g. `e+1` `e-1`

xDarksome:
The name or signature of the function was not the point. Do you prefer `movetoworkspace("e+1")` instead of `movetopopulatedworkspace("+1`)? OK, I just don't share the enthusiasm of using magic strings.

The point is to introduce "Move to non-empty workspace" as a general concept and not a "scrolling" thing. Plus to make a future proof configuration.

vaxerski:
`e+1` then is the best option

fufexan:
> The point is to introduce "Move to non-empty workspace" as a general concept and not a "scrolling" thing. Plus to make a future proof configuration.

It exists

```sh
bind=SUPER,bracketleft,workspace,m-1
bind=SUPER,bracketright,workspace,m+1
```

xDarksome:
@fufexan 

> actually m+1 only says "next one on this monitor"

fufexan:
Ah, true. That's where `e` comes in i guess.

xDarksome:
Tweaking configuration to support mouse wheel turned out to be easier than I thought.

vaxerski:
fix conflicts please, and then we can merge this i think

xDarksome:
@vaxerski done

Mention the "e+{n}" stuff in the wiki, please. It can be used on any keybind not just the mouse wheel.

PR #291, opened 2022-07-01
meson: build with -O3

fufexan:
Adds `-O3` to builds by default.

No issues.

I'm also thinking of setting a lower default warning level, and using level 3 for debug builds. `-Wpedantic` is over-eagerly filling up the entire history of my terminal.
@sp1ritCS thoughts?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,15 @@
 project('Hyprland', 'cpp', 'c',
   version : '0.6.2beta',
-  default_options : ['warning_level=3', 'cpp_std=c++20', 'default_library=static'])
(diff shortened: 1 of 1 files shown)


fufexan:
Based on https://github.com/mesonbuild/meson/issues/3275#issuecomment-641354956 and https://github.com/hyprwm/Hyprland/blob/main/CMakeLists.txt#L41 I think we want to enable level 2 plus a few other warning flags.

sp1ritCS:
`-DWLR_USE_UNSTABLE` is already set in src/meson.build in the executable call. there is no need to have it set project wide (hyprctl doesn't need it)

fufexan:
Thanks for the input!

PR #292, opened 2022-07-01
Changed dispatcher: focusWindowByClass -> focusWindow. Now supports titles & classes

Dakes:
Before, you could only focus a window by a class and not differentiate ones of the same class. Now you can use a the title keyword, the same way as with window rules. 
title: `focuswindow title:^(.*)(WaniKani)(.*)$`
class: `focuswindow dolphin`

This is useful to, for example, sort multiple auto-launching browser windows to the right workspace. (I use it together with BroTab for Firefox to activate the right pinned tab beforehand)

focusWindow is backwards compatible to focusWindowByClass. So even though I removed it, I kept a dispatcher for focuswindowbyclass -> focusWindow to not break existing configs. 

ready


vaxerski:
lgtm, thank you!

PR #299, opened 2022-07-03
Don't use the default nixpkgs.hyprland package in the nix flake

xDarksome:
After hyprland package was merged into nigpkgs, the following code in my NixOS flake no longer works:

```nix
modules = [
  hyprland.nixosModules.default
  { programs.hyprland.enable = true;}
];
```

`nixos-rebuild` fails with the following error:

```
error: Package, 'hyprland-0.6.1beta', did not specify any session names, as strings, in
       'passthru.providedSessions'. This is required when used as a session package.

       The session names can be looked up in:
         /nix/store/mblhy3nhyyycnblyx3038pl9md8xsf1b-hyprland-0.6.1beta/share/xsessions
         /nix/store/mblhy3nhyyycnblyx3038pl9md8xsf1b-hyprland-0.6.1beta/share/wayland-sessions
```

Seems like nixpkgs.hyprland is broken, but that's not the point of this PR.

For some reason the local flake uses that default package instead of the local one. If this was the intended behavior, can some nix guru explain me why?

It seems to me that if I am using custom git flake instead of nixpkgs it is reasonable to choose the package inside the flake by default.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -19,7 +19,7 @@ in {
 
     package = mkOption {
       type = types.package;
-      default = self.packages.${pkgs.system}.default;
+      default = self.packages.${pkgs.system}.default or pkgs.hyprland;
       defaultText = literalExpression "<Hyprland flake>.packages.<system>.default";
       example = literalExpression "<Hyprland flake>.packages.<system>.default.override { }";
       description = ''
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -19,7 +19,7 @@ in {
 
     package = mkOption {
       type = types.package;
-      default = self.packages.${pkgs.system}.default or pkgs.hyprland;
+      default = self.packages.${pkgs.system}.default;
       defaultText = literalExpression "<Hyprland flake>.packages.<system>.default";
       example = literalExpression "<Hyprland flake>.packages.<system>.default.override { }";
       description = ''
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -19,7 +19,7 @@ in {
 
     package = mkOption {
       type = types.package;
-      default = pkgs.hyprland or self.packages.${pkgs.system}.default;
+      default = self.packages.${pkgs.system}.default;
       defaultText = literalExpression "<Hyprland flake>.packages.<system>.default";
       example = literalExpression "<Hyprland flake>.packages.<system>.default.override { }";
       description = ''


fufexan:
Seems like a good change to me. We didn't expect this to break when hyprland was merged in nixpkgs, but it's good that you found it out.

It was `pkgs.hyprland or ...` first in order to accommodate people using overlays instead of the package.

nix/module.nix
```diff
@@ -19,7 +19,7 @@ in {
 
     package = mkOption {
       type = types.package;
-      default = pkgs.hyprland or self.packages.${pkgs.system}.default;
+      default = self.packages.${pkgs.system}.default;
```
fufexan, line 22:
```suggestion
      default = self.packages.${pkgs.system}.default or pkgs.hyprland;
```

This way we don't break compatibility all the way.

xDarksome:
> It was `pkgs.hyprland or ...` first in order to accommodate people using overlays instead of the package.

Got it, thanks, never used overlays myself.

viperML:
Ah what a mess.

`pkgs.hyprland` is needed for non-flake users, that have to use the overlay.

xDarksome:
@viperML so
```
default = self.packages.${pkgs.system}.default or pkgs.hyprland;
```

won't work for overlay users?

viperML:
That will use the package built with the flake's nixpkgs, which is not the same as a non-flake user's nixpkgs.

viperML:
Also, the first argument of that `or` always exists, so it is equivalent to `default = self.packages.${pkgs.system}.default;`

fufexan:
I don't see other good option than leaving the package as the default, and instructing non-flake users to set `package = pkgs.hyprland`.

viperML:
We can also check if the `pkgs` is using an overlay, such as using `builtis.hasAttr "wlroots-hyprland" pkgs`, but I don't think making the module so "intelligent" is good in the long run.

nix/module.nix
```diff
@@ -19,7 +19,7 @@ in {
 
     package = mkOption {
       type = types.package;
-      default = pkgs.hyprland or self.packages.${pkgs.system}.default;
+      default = self.packages.${pkgs.system}.default or pkgs.hyprland;
```
viperML, line 22: Just use `default = self.packages.${pkgs.system}.default;`

viperML (changes requested):
This `or` would not serve any purpose.

xDarksome:
I think the change in overlay usage should be mentioned in the wiki.

viperML:
If @fufexan updates the wiki, this LGTM

PR #309, opened 2022-07-05
Add 1:1 swipe gestures to switch workspaces

RealTheSunCat:
I wrote this to enable one-to-one touchpad gestures to switch workspaces as it is a really nice feature to have on laptops.

I have not tested this with multiple monitors, but presumably it should work fine as it uses the same animations as normal workspace switching.
The gesture is broken if you try to switch between two workspaces that skip an ID, such as going from 3->5. This would have to be fixed before merging.

No: the bug described above needs to be fixed, and this needs testing under more configs than my own. It would also be a good idea to make this feature configurable (number of fingers, on/off, sensitivity).
Additionally, I left some comments marked TODO which indicate bits of code that, though functional, might need rethinking.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -87,6 +87,8 @@ CCompositor::CCompositor() {
     m_sWLRCursor = wlr_cursor_create();
     wlr_cursor_attach_output_layout(m_sWLRCursor, m_sWLROutputLayout);
 
+    m_sWLRPointerGestures = wlr_pointer_gestures_v1_create(m_sWLDisplay);
+
     m_sWLRXCursorMgr = wlr_xcursor_manager_create(nullptr, 24);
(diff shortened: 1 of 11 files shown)


fufexan:
Damn you were fast with it. I'll test and report back.

RealTheSunCat:
Thanks for testing! I didn't expect it to crash with multiple monitors... I'm guessing there's an issue with how it searches for the previous/next workspace to go to when there's multiple monitors.
I'd mark it as a draft if I knew how, but yes it very much needs cleaning up as well as configuration options.
I was thinking of allowing the number of fingers to be configured, sensitivity, and a general toggle for the feature, but not sure what else.

Does it crash at all when using a single monitor? It's been fully stable in my testing.

vaxerski:
closing as swipes have been added

PR #321, opened 2022-07-07
meson: add -DGIT arguments

fufexan:
Adds functionality like [this](https://github.com/hyprwm/Hyprland/blob/main/CMakeLists.txt#L12-L34) and [this](https://github.com/hyprwm/Hyprland/blob/main/CMakeLists.txt#L69-L72) to meson.

Compiling with the official meson instructions and then checking out `build/hyprctl/hyprctl version` yields the wrong result:
```
 › build/hyprctl/hyprctl version
Hyprland, built from branch ? at commit ?? (?).
flags: (if any)
debug
```

Still needs the above problem figuring out.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,12 +4,22 @@ project('Hyprland', 'cpp', 'c',
 
 add_global_arguments('-std=c++23', language: 'cpp')
 
+GIT_BRANCH = run_command('git', 'rev-parse', '--abbrev-ref', 'HEAD', check: false).stdout().strip()
+GIT_COMMIT_HASH = run_command('git', 'rev-parse', 'HEAD', check: false).stdout().strip()
+GIT_COMMIT_MESSAGE = run_command('bash', '-c', 'git show | head -n 5 | tail -n 1', check: false).stdout().strip()
+GIT_DIRTY = run_command('bash', '-c', 'git diff-index --quiet HEAD -- || echo "dirty"', check: false).stdout().strip()
+
 add_project_arguments(
   [
     '-Wno-unused-parameter',
     '-Wno-unused-value',
     '-Wno-missing-field-initializers',
     '-Wno-narrowing',
+
+    f'-DGIT_BRANCH="@GIT_BRANCH@"',
+    f'-DGIT_COMMIT_HASH="@GIT_COMMIT_HASH@"',
+    f'-DGIT_COMMIT_MESSAGE="@GIT_COMMIT_MESSAGE@"',
+    f'-DGIT_DIRTY="@GIT_DIRTY@"',
   ],
   language: 'cpp')
 
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -5,6 +5,7 @@
   pkg-config,
   meson,
   ninja,
+  git,
   libdrm,
   libinput,
   libxcb,
@@ -36,6 +37,7 @@ stdenv.mkDerivation {
 
   buildInputs =
     [
+      git
       libdrm
       libinput
       libxcb


meson.build
```diff
@@ -4,12 +4,22 @@ project('Hyprland', 'cpp', 'c',
 
 add_global_arguments('-std=c++23', language: 'cpp')
 
+GIT_BRANCH = run_command('gitinfo.sh', 'GIT_BRANCH').stdout().strip()
+GIT_COMMIT_HASH = run_command('gitinfo.sh', 'GIT_COMMIT_HASH').stdout().strip()
+GIT_COMMIT_MESSAGE = run_command('gitinfo.sh', 'GIT_COMMIT_MESSAGE').stdout().strip()
+GIT_DIRTY = run_command('gitinfo.sh', 'GIT_DIRTY').stdout().strip()
```
eli-schwartz, line 10: You want the `check: true` kwarg to run_command, because recent versions of Meson emit a deprecation warning if you don't specify some value of `check`.

It's best to use "true", because that would ideally have been the default all along -- if those commands error out for some crazy reason Meson can error out if the information is required.

meson.build
```diff
@@ -4,12 +4,22 @@ project('Hyprland', 'cpp', 'c',
...
   [
     '-Wno-unused-parameter',
     '-Wno-unused-value',
     '-Wno-missing-field-initializers',
     '-Wno-narrowing',
+
+    f'-DGIT_BRANCH="@GIT_BRANCH@"',
+    f'-DGIT_COMMIT_HASH="@GIT_COMMIT_HASH@"',
+    f'-DGIT_COMMIT_MESSAGE="@GIT_COMMIT_MESSAGE@"',
+    f'-DGIT_DIRTY="@GIT_DIRTY@"',
```
eli-schwartz, line 22: If you're going to use Meson's documented support for f-strings, which was added in Meson 0.58, you might want to set that in the `project()` function as `meson_version: '>=0.58.0'` so that Meson can error out on older versions and tell the user to "upgrade Meson".

It's a pretty recent version and e.g. not available in Debian stable's official repositories. It's of course trivial to upgrade it with `pip install`, but it's quite easy to imagine people trying the version from the Debian repositories first, then having it fail, so letting them know the reason is because they have an old version of Meson is useful information.

sp1ritCS:
> can you check this for any dumb mistakes?

what does this do when Hyprland is built from source outside of git? (source release or some kind of source puller script that strips the `.git` dir)

fufexan:
> what does this do when Hyprland is built from source outside of git? (source release or some kind of source puller script that strips the .git dir)

> (What does cmake do in that case?)

CMake sets those macros to "", but I'm not sure what meson does. Will have to test once I get it working.

gitinfo.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+declare -A commands=(
+  ["GIT_BRANCH"]=$(git rev-parse --abbrev-ref HEAD)
+  ["GIT_COMMIT_HASH"]=$(git rev-parse HEAD)
+  ["GIT_COMMIT_MESSAGE"]=$(git show ${GIT_COMMIT_HASH} | head -n 5 | tail -n 1)
+  ["GIT_DIRTY"]=$(git diff-index --quiet HEAD -- || echo 'dirty')
+)
+
+echo ${commands[$1]}
```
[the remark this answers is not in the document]
fufexan: No, the script doesn't have the executable bit set. Meson seemed to build fine even without it, so I didn't include it.

> you can do that in Meson too, via run_command('bash', '-c', '.....')

Thanks, on it.

fufexan:
So it still shows the same output, I don't know why. I see that the CMakeLists file also says something about `PRIVATE`. Don't know what that is, but google led me to believe it's about the same as `add_project_arguments` instead of `add_global_arguments`.

fufexan:
So, does anyone have any idea how I could fix this?

sp1ritCS:
> So, does anyone have any idea how I could fix this?

you need to set check to false from the run_command to avoid the build failing should git not be installed / shallow clone / tar archive, besides that it seems to work

PR #342, opened 2022-07-10
ci: add release workflow

viperML:

This worflow will create a new release whenever a new tag is pushed. The workflow attaches a source tarball including the submodules, which has been requested by some packagers.

The distribution release of arch files still needs to be manually added. In the future another step of this workflow would attach it, but this could do for now.

Ready.

You can see it in action here: https://github.com/viperML/Hyprland/releases/tag/v2.0.13beta


fufexan (approved):

vaxerski:
Again, as Kainoa once wanted to, I do not like this. It doesn't have my name in the release.

Furthermore, manual intervention in the script is required (always) to remove some duds

If you cannot make the bot *edit* a release instead of pushing it (or make it not show "github-actions published") then I won't accept this.

viperML:
I didn't know this was already proposed.

> It doesn't have my name in the release.

The action could be modified to do so

> manual intervention in the script is required (always) to remove some duds

Releases can always be manually edited

Just throwing these options. Feel free to close it if you don't want it at all

vaxerski:
> The action could be modified to do so

if it can be modified to show my name, then please make it like that :-)

> Feel free to close it if you don't want it at all

a system like this is fine for me so long as it looks like the current releases.

viperML:
Rewrote the action.

Now it runs after a release is externally published, and will attach the tarball to it.

Sample run: https://github.com/viperML/Hyprland/releases/tag/v3.0.1

@vaxerski

vaxerski:
and the source-3.0.1 is the wlroots source? hyprland source?

viperML:
> and the source-3.0.1 is the wlroots source? hyprland source?

It is what you get after a `git clone --recursive`.

A common usecase is for distro packages to just fetch that tarball, instead of cloning the repo.

vaxerski:
ok

PR #343, opened 2022-07-10
Remove extra packages in nix module

reptee:
Hello!

That's a super short PR, removing any default extra packages provided via nixos module. While extra packages make sense for full DE, i believe that WM is usually DIY environment.

Nope

Ready

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -29,13 +29,9 @@ in {
 
     extraPackages = mkOption {
       type = with types; listOf package;
-      default = with pkgs; [
-        kitty
-        wofi
-        swaybg
-      ];
+      default = with pkgs; [ ];
       defaultText = literalExpression ''
-        with pkgs; [ kitty wofi swaybg ];
+        with pkgs; [ ];
       '';
       example = literalExpression ''
         with pkgs; [
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -26,6 +26,10 @@ in {
         Hyprland package to use.
       '';
     };
+
+    imports = [
+      (mkRemovedOptionModule ["programs" "hyprland" "extraPackages"] "extraPackages has been removed. Use environment.systemPackages instead.")
+    ];
   };
 
   config = mkIf cfg.enable {
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -26,22 +26,6 @@ in {
         Hyprland package to use.
       '';
     };
-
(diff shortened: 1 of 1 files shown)


viperML:
I added that option thinking it could be useful for people to have a working environment with the default config, that uses kitty.

If you want to remove it, I'd also remove the extraPackages option altogether

reptee:
`extraPackages` explicitly says that bunch of packages are needed with another package(hyprland), and aren't really required without it

viperML:
> `extraPackages` explicitly says that bunch of packages are needed with another package(hyprland), and aren't really required without it
> 
> EDIT: when it comes to system.packages, it isn't really clear, why are those packages installed, are they a part of an environment, or simply a standalone applications. Say notification daemon would make more sense with hyprwm as it is a part of complete hyprwm setup, while say, firefox is just a standalone app

Sorry, but I don't understand the point your are making

reptee:
Separating package and its dependencies -> bad

Grouping package and its dependencies -> good

Even if dependencies aren't really a part of runtime for hyprland.

``` nix
# separated
{
  programs.hyprland.enable = true;
  environment.systemPackages = with pkgs; [ some packages ];
}

# grouped, easy to understand extra packages are a part of hyprland setup
{
  programs.hyprland = {
    enable = true;
    extraPackages = with pkgs; [ some packages ];
  };
}
```

Trying my best to explain, at some point I agree with

> If you want to remove it, I'd also remove the extraPackages option altogether

though having such option is nice

viperML:
You can always split your config to make sure that those packages go with the Hyprland part. Personally I see it unnecesary to provide an alias to `environment.systemPackages` that comes empty by default.

Maybe it is just me, keep the changes I guess 👍

viperML:
@fufexan

fufexan:
I agree with what viper said, we should also remove `extraPackages` while we're at it. I haven't seen anyone use them, and if you're tinkering with compositors you already know how to install the proper packages you'll need inside them.

reptee:
Whatever you prefer, gentlemen. I will be satisfied with removal of default packages in either way

fufexan:
@viperML any last words before merge? xD

viperML:
iirc you could set an option as removed with the nixos module system, if you want to go the extra mile

(Right now I don't have access to my nix machines to test)

otherwise lgtm

fufexan:
Should be fine now?

PR #353, opened 2022-07-12
add config drag_lock for touchpad

zzzsyyy:
Add more config "set drag lock" for touchpad

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -530,6 +530,11 @@ void CInputManager::newMouse(wlr_input_device* mouse, bool virt) {
                 libinput_device_config_middle_emulation_set_enabled(LIBINPUTDEV, LIBINPUT_CONFIG_MIDDLE_EMULATION_DISABLED);
         }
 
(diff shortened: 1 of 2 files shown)


src/managers/input/InputManager.cpp
```diff
@@ -530,6 +530,11 @@ void CInputManager::newMouse(wlr_input_device* mouse, bool virt) {
                 libinput_device_config_middle_emulation_set_enabled(LIBINPUTDEV, LIBINPUT_CONFIG_MIDDLE_EMULATION_DISABLED);
         }
 
+        if ((HASCONFIG ? g_pConfigManager->getDeviceInt(PMOUSE->name, "clickfinger_behavior") : g_pConfigManager->getInt("input:touchpad:drag_lock")) == 0)
```
vaxerski, line 533: wrong

zzzsyyy: ahhhhh, fixed:

vaxerski: not fixed, the config for a device called `drag_lock` doesn't exist. add it. you only added the global one

PR #379, opened 2022-07-16
docs: add man pages

volitank:
This PR adds a man page for both `Hyprland` and `hyprctl`.

The man pages are written in `.rst` format so it is easy to read and edit.

`rst2man` is used during install to convert the pages. They are then piped into `gzip` for compression and finally written to the proper man directory.

There should be no bugs added as there is no code changed except for the two lines added into the Makefile.

This is ready to be merged afaik.

The man pages can be extended later to include a `Hyprland-conf.5` page for configurations.

Let me know of anything you would change. I have built and tested using the cmake method from the documentation.

To test the man pages you do not need to build the whole program. You can use `rst2man ./docs/hyprctl.1.rst | man -l -`

`python-docutils` for Arch should probably be added as a build dependency in the wiki when this is merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,9 @@ PKGS = wlroots wayland-server xcb xkbcommon libinput
 CFLAGS += $(foreach p,$(PKGS),$(shell pkg-config --cflags $(p)))
 LDLIBS += $(foreach p,$(PKGS),$(shell pkg-config --libs $(p)))
 
+VERSION=0.7.1beta
+DATE=$(shell date "+%d %b %Y")
+
 xdg-shell-protocol.h:
 	$(WAYLAND_SCANNER) server-header \
 		$(WAYLAND_PROTOCOLS)/stable/xdg-shell/xdg-shell.xml $@
@@ -124,8 +127,7 @@ install:
 	cp ./assets/wall_4K.png ${PREFIX}/share/hyprland
 	cp ./assets/wall_8K.png ${PREFIX}/share/hyprland
 
-	rst2man ./docs/hyprctl.1.rst | gzip -c > /usr/share/man/man1/hyprctl.1.gz
-	rst2man ./docs/Hyprland.1.rst | gzip -c > /usr/share/man/man1/Hyprland.1.gz
+	make man
 
 uninstall:
 	rm -f ${PREFIX}/share/wayland-sessions/hyprland.desktop
@@ -150,3 +152,22 @@ config:
 	cd subprojects/wlroots && ninja -C build/
 
 	cd subprojects/wlroots && ninja -C build/ install
+
+man:
+	pandoc ./docs/Hyprland.1.rst \
+		--standalone \
+		--variable=header:"Hyprland User Manual" \
+		--variable=footer:${VERSION} \
+		--variable=date:"${DATE}" \
+		--variable=section:1 \
+		--from rst \
+		--to man | gzip -c > /usr/share/man/man1/Hyprland.1.gz
+
+	pandoc ./docs/hyprctl.1.rst \
+		--standalone \
+		--variable=header:"hyprctl User Manual" \
+		--variable=footer:${VERSION} \
+		--variable=date:"${DATE}" \
+		--variable=section:1 \
+		--from rst \
+		--to man | gzip -c > /usr/share/man/man1/hyprctl.1.gz
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ PKGS = wlroots wayland-server xcb xkbcommon libinput
 CFLAGS += $(foreach p,$(PKGS),$(shell pkg-config --cflags $(p)))
 LDLIBS += $(foreach p,$(PKGS),$(shell pkg-config --libs $(p)))
 
-VERSION=0.7.1beta
 DATE=$(shell date "+%d %b %Y")
 
 xdg-shell-protocol.h:
@@ -157,7 +156,6 @@ man:
 	pandoc ./docs/Hyprland.1.rst \
 		--standalone \
 		--variable=header:"Hyprland User Manual" \
-		--variable=footer:${VERSION} \
 		--variable=date:"${DATE}" \
 		--variable=section:1 \
 		--from rst \
@@ -166,7 +164,6 @@ man:
 	pandoc ./docs/hyprctl.1.rst \
 		--standalone \
 		--variable=header:"hyprctl User Manual" \
-		--variable=footer:${VERSION} \
 		--variable=date:"${DATE}" \
 		--variable=section:1 \
 		--from rst \
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,8 @@ install:
 	cp ./assets/wall_4K.png ${PREFIX}/share/hyprland
 	cp ./assets/wall_8K.png ${PREFIX}/share/hyprland
 
+	rst2man ./docs/hyprctl.1.rst | gzip -c > /usr/share/man/man1/hyprctl.1.gz
+	rst2man ./docs/Hyprland.1.rst | gzip -c > /usr/share/man/man1/Hyprland.1.gz
 
 uninstall:
 	rm -f ${PREFIX}/share/wayland-sessions/hyprland.desktop
--- a/docs/Hyprland.1.rst
+++ b/docs/Hyprland.1.rst
@@ -1,30 +1,24 @@
-========
-Hyprland
-========
+:title: Hyprland
+:author: Vaxerski <*https://github.com/vaxerski*>
 
----------------------------------
-Dynamic tiling Wayland compositor
----------------------------------
+NAME
+====
 
-:Date: 15 Jul 2022
-:Copyright: Copyright (c) 2022, vaxerski
-:Version: 0.7.1beta
-:Manual section: 1
-:Manual group: Hyprland
+Hyprland - Dynamic tiling Wayland compositor
 
 SYNOPSIS
 ========
 
-``Hyprland`` [arg [...]].
+**Hyprland** [*arg [...]*].
 
 DESCRIPTION
 ===========
 
-``Hyprland`` is a dynamic tiling Wayland compositor based on
+**Hyprland** is a dynamic tiling Wayland compositor based on
 wlroots that doesn't sacrifice on its looks.
 
 You can launch Hyprland by either going into a TTY and
-executing ``Hyprland``, or with a login manager.
+executing **Hyprland**, or with a login manager.
 
 NOTICE
 ======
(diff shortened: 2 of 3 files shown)


docs/Hyprland.1.rst
```diff
@@ -0,0 +1,75 @@
...
+Hyprland
+========
+
+---------------------------------
+Dynamic tiling Wayland compositor
+---------------------------------
+
+:Date: 15 Jul 2022
+:Copyright: Copyright (c) 2022, vaxerski
+:Version: 0.7.1beta
```
fufexan, line 11: This is yet another location where version is used. Can we set it through something like an import? maybe we could keep it by default `@VERSION@`, and only change it using a script on release.

All the places in this repo where we use the version string could benefit from having a single place to edit it.

docs/Hyprland.1.rst
```diff
@@ -0,0 +1,75 @@
...
+=============
+
+For configuration information please see <`https://github.com/hyprwm/Hyprland/wiki`>.
+
+LAUNCHING
+=========
+
+You can launch Hyprland by either going into a TTY and executing ``Hyprland``, or with a login manager.
+
+`IMPORTANT`: Do `not` launch ``Hyprland`` with `root` permissions (don't `sudo`)
```
fufexan, line 43: I think this is not that important, as most users won't do that. And even if there are users who do, the program will tell them before they do anything stupid.

docs/Hyprland.1.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Login managers are not officially supported, but here's a short compatibility list:
```
fufexan, line 45: This depends on the distro, the user's config, etc. Not sure whether it's needed inside the manpage. Could be more useful as an extra "did you know" in the wiki.

docs/Hyprland.1.rst
```diff
@@ -0,0 +1,75 @@
...
+
+    * SDDM -> Works flawlessly.
+    * GDM -> Works with the caveat of crashing `Hyprland` on the first launch.
+    * ly -> Works with minor issues and/or caveats.
+
+OPTIONS
+=======
+
+-h, --help
+    Show this help message.
```
fufexan, line 55:
```suggestion
    Show command usage.
```

docs/Hyprland.1.rst
```diff
@@ -0,0 +1,75 @@
...
+
+-c, --config
+    Specify config file to use.
+
+BUGS
+====
+
+Submit bug reports and feature requests online at:
```
fufexan, line 63:
```suggestion
Submit bug reports and request features online at:
```

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+SYNOPSIS
+========
+
+``hyprctl`` [(opt)flags] [command] [(opt)args]
+
+DESCRIPTION
+===========
+
+``hyprctl`` is a utility for controlling some parts of the compositor from a CLI or a script.
+If you install with make install, or any package, it should automatically be installed.
```
fufexan, line 24: We don't need this line, as people won't have access to this if hyprctl isn't installed. And for those who have it, it's redundant info.

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+
+COMMANDS
+========
+
+Control
+
+    ``dispatch``
+
+        Call a keybinding dispatcher with an argument.
```
fufexan, line 33:
```suggestion
        Call a dispatcher with an argument.
```

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+
+        Examples:
+
+            ``hyprctl`` `dispatch exec kitty`
+
+            ``hyprctl`` `dispatch pseudo x`
+
+    ``keyword``
+
+        Call a config keyword dynamically.
```
fufexan, line 48:
```suggestion
        Set a config keyword dynamically.
```

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+            ``hyprctl`` `keyword general:border_size 10`
+
+    ``reload``
+
+        Force a reload of the config file.
+
+    ``kill``
+
+        Enter kill mode, where you can kill an app by clicking on it.
+        You can exit with ESCAPE.
```
fufexan, line 65:
```suggestion
        You can exit by pressing ESCAPE.
```

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Info
+
+    ``version``
+
+        Prints the hyprland version, meaning flags, commit and branch of build.
```
fufexan, line 71:
```suggestion
        Prints the Hyprland version, flags, commit and branch of build.
```

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+
+        Lists all workspaces with their properties.
+
+    ``clients``
+
+        Lists all windows with their properties.
+
+    ``devices``
+
+        Lists all connected keyboards and mice.
```
fufexan, line 87:
```suggestion
        Lists all connected input devices.
```
Tablets also count.

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    ``activewindow``
+
+        Gets the active window name.
```
fufexan, line 91:
```suggestion
        Returns the active window name.
```

docs/hyprctl.1.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    ``layers``
+
+        Lists all the layers.
+
+    ``splash``
+
+        Prints the current random splash.
```
fufexan, line 99:
```suggestion
        Returns the current random splash.
```
Debatable, but we should favor consistency in our wording.

fufexan (changes requested):
These are just minor suggestions, but I think they make the manpages look better and more professional.

docs/Hyprland.1.rst
volitank: I like all your suggestions. As for `@Version@` I'm okay with doing that, but not sure of how you want to go about implementing that. I do agree though where ever possible version strings should be interpolated

docs/Hyprland.1.rst
volitank: For this one I removed all the extra stuff that doesn't really make sense in a man page. Since it was only one line I just rolle d the `You can launch` line up into the description block

docs/Hyprland.1.rst
fufexan: That was more of a long-term idea. For this PR you can keep the current version, and we'll figure out the versioning system in another one.

fufexan (approved):

Makefile
```diff
@@ -124,6 +124,8 @@ install:
 	cp ./assets/wall_4K.png ${PREFIX}/share/hyprland
 	cp ./assets/wall_8K.png ${PREFIX}/share/hyprland
 
+	rst2man ./docs/hyprctl.1.rst | gzip -c > /usr/share/man/man1/hyprctl.1.gz
```
vaxerski, line 127: what the fuck is rst2man and what is this a dependency of

volitank: I think it's python-docutils on arch.

Docutils-common on debian

volitank: It just coverts rst to a man page format. Groff is a terrible format to read.

vaxerski: I'd rather want us not to rely on python, if it's possible.

fufexan: We can use `pandoc --from rst --to man <files>`

volitank: I'm okay with using pandoc. I've used it before. Not too many differences. Just some slight format changes.

volitank: I've updated this to use pandoc instead of rst2man.

A notable good thing from the switch is that we can use switches in the Makefile for setting the version and date.

fufexan: Oh, even better! Good work.

docs/Hyprland.1.rst
volitank: Okay so I think this one can be resolved. Having the ability to input the version in the make file will make things way easier later.

Makefile
volitank: Thank you. I'm going to have to switch over to pandoc on my own projects.

This should be good to go.
Eventually I'll write up a man page for the config too, but that's going to take a bit.

Makefile
```diff
@@ -9,6 +9,9 @@ PKGS = wlroots wayland-server xcb xkbcommon libinput
 CFLAGS += $(foreach p,$(PKGS),$(shell pkg-config --cflags $(p)))
 LDLIBS += $(foreach p,$(PKGS),$(shell pkg-config --libs $(p)))
 
+VERSION=0.7.1beta
```
vaxerski, line 12: Please don't, I will not manually update this.

Makefile
fufexan: pandoc is pandoc. haskell but already compiled.

Makefile
fufexan: Let's think of a better versioning mechanism then. Say, a file in the repo root, and an action that changes the versions everywhere.

vaxerski: if you can be arsed to do this, sure

fufexan: All you gotta do is edit that file when you make a release. It would still have to be run *before* the release, so that the changed versions end up inside it. Will have to look at how other projects handle it though.

viperML: You can't create another commit that updates the versions all over the place, because that commit wouldn't correspond to the tagged one.

fufexan: then we'd have to do it all in the release action (forgot we have that now).

viperML: We don't, @vaxerski manually bumps the tag

volitank: @vaxerski How would you like this resolved? I guess the most simple way would be to just not include the version in the Man Page, it's not absolutely necessary, just usually included.

But for the long term it would be nice to have some version information automated. I wouldn't mind working to implement this.

Do you know of any other places that currently need a version string?

I would imagine just something simple like `git describe --tags` or whatever could be used. Would not necessarily need a commit, just have it load the version information where necessary during build.

vaxerski: idk for now just remove it i guess

volitank: Okay I've removed it.

vaxerski:
I've removed make man from the make install due to like 140 deps required for pandoc. Can someone just fucking use a normal utility for that for god's sake

fufexan:
We don't even have man building on Meson :older_man:

eli-schwartz:
> I've removed make man from the make install due to like 140 deps required for pandoc. Can someone just fucking use a normal utility for that for god's sake

Well, python-docutils only depends on python and python alone. :) But you said you'd rather not use python.

volitank:
Can always leave it how it is. Only use `make man` for updating the man pages and have them in the `docs` dir as the groff file. Then you won't need any deps unless you update them, just compress it and g2g.

ErikReider:
Why not use something like scdoc?

PR #395, opened 2022-07-20
nix: add home-manager module

Misterio77:
Hey, folks!

Adds a [home-manager](https://github.com/nix-community/home-manager) module.

I built this module for my own use a few weeks ago, so I decided to polish it up and upstream it :)

For those uninitiated with nix: home-manager has modules (abstractions) for doing things on the user (instead of system, as is for NixOS modules) level. It has a lot more (and more extensive) modules than NixOS when we're talking about "dotfiley" stuff (editors, window-managers, terminals, etc).

[Some example usage on my config repo](https://github.com/Misterio77/nix-config/blob/d19cf7ae7d47d1b9ee7ed7521efa16f64f1d886c/home/misterio/desktop/hyprland/default.nix#L8)

I think I read somewhere someone was working on a home-manager module, sorry if this somehow clashes with what you have possibly already written.

Please feel free to use this code as you folks wish, even if you end up merging another implementation. I'll be glad to contribute in any way I can <3

Should be ready to go:)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -61,6 +61,7 @@
     formatter = genSystems (system: pkgsFor.${system}.alejandra);
 
     nixosModules.default = import ./nix/module.nix self;
+    homeManagerModules.default = import ./nix/hm-module.nix self;
 
     overlay = throw "Hyprland: .overlay output is deprecated, please use the .overlays.default output";
   };
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -1,0 +1,98 @@
+self: {
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  cfg = config.wayland.windowManager.hyprland;
+  defaultHyprlandPackage = self.packages.${pkgs.system}.default.override {
+    enableXWayland = cfg.xwayland;
+  };
+in {
+  options.wayland.windowManager.hyprland = {
+    enable = lib.mkEnableOption "hyprland wayland compositor";
+    package = lib.mkOption {
+      type = with lib.types; nullOr package;
+      default = defaultHyprlandPackage;
+      description = ''
+        Hyprland package to use. Will override the 'xwayland' option.
+
+        Defaults to the one provided by the flake. Set it to
+        <literal>pkgs.hyprland</literal> to use the one provided by nixpkgs or
+        if you have an overlay.
+
+        Set to null to not add any Hyprland package to your path. This should
+        be done if you want to use the NixOS module to install Hyprland.
+      '';
+    };
+    systemdIntegration = lib.mkOption {
+      type = lib.types.bool;
+      default = pkgs.stdenv.isLinux;
+      description = ''
+        Whether to enable <filename>hyprland-session.target</filename> on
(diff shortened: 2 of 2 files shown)


viperML:
Looks nice, but I have some doubts:

---
How should we tell users to use this? From a simple configuration with just this module enabled, users would miss some NixOS settings, like enabling polkit, opengl, etc. https://github.com/hyprwm/Hyprland/blob/d52446135bd46dc0260d9507e82173847d4196c8/nix/module.nix#L51

I can think of some options:

- Enable HM and NixOS module, and set the NixOS's package setting to null.
- Just instruct the users to manually enable the settings that the NixOS module sets.

---
In any case, we need a section in the wiki, so maybe send the raw markdown as a github comment. @fufexan had write perms iirc

---
Please run `nix fmt` to have a consistent style in all nixfiles ;)

fufexan:
Great stuff, thanks for the contribution!

Yes, I wrote a stub of a HM module a while ago and never got to finish it (partly because I wanted to configure everything in Nix, but that's quite a lot of stuff to account for. My implementation is available [here](https://github.com/fufexan/Hyprland/commit/fbdce6cbe00e97845a6ee70cb9d23b29343a7926).

I think we should go with this one instead, as config options are added irregularly and it would be a big maintenance burden to add all that to the Nix config options as well.

Also, I agree with what @viperML said, we'll just create a wiki page on Nix, and tell users how to proceed.

fufexan:
I added the <https://github.com/hyprwm/Hyprland/wiki/Nix> page, I think we're ready to mrge as soon as we can also push docs about HM in there. I'll try to whip something up, let me know if you have suggestions.

fufexan:
Here's the diff I'd add, thoughts?

```diff
diff --git a/Nix.md b/Nix.md
index 030c078..b95d8d5 100644
--- a/Nix.md
+++ b/Nix.md
@@ -31,6 +31,8 @@ Make sure to check out the options of the
   };
 \```
 
+Don't forget to replace `HOSTNAME` with your hostname!
+
 ### Without flakes
 
 **NOTE:** If you're using Hyprland through an overlay, set
@@ -56,3 +58,32 @@ in {
   };
 }
 \```
+
+## Home Manager module
+
+You can use the Home Manager module by adding it to your configuration:
+
+```nix
+{ config, pkgs, inputs, ... }: {
+  imports = [
+    inputs.hyprland.homeManagerModules.default
+  ];
+
+  wayland.windowManager.hyprland.enable = true;
+  # ...
+}
+```
+
+For a list of available options, check the
+[module file](https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix).
+
+## Modules mix'n'match
+
+If you plan on using the HM module alongside the NixOS module, set the NixOS
+`programs.hyprland.package = null;`.
+
+If you don't plan on using the NixOS module, but want to use the HM module, you
+will have to enable all the options the NixOS module enables.
+
+If you don't plan on using any module, manually enable whatever options the
+modules set.
```

viperML:
lgtm

Misterio77:
That looks great! Thanks :)

PR #479, opened 2022-08-05
output valid json when there are no clients

rubyist:
The json formatter pops off a character assuming it's a comma, but if there are no clients it pops off the `[` instead:

```
$ hyprctl -j clients
]
```
This fixes that up:

```
$ hyprctl -j clients
[]
```


vaxerski:
thank you uwu

PR #483, opened 2022-08-06
Have hyprctl -j activewindows return empty json object if there are no activewindows

rubyist:
If there are no activewindows, have `hyprctl -j activewindow` return `{}` instead of `Invalid`


vaxerski:
thank you!

PR #515, opened 2022-08-12
fix: support clang in meson build

SignalWalker:

Fixes an error when building with Clang, which doesn't yet support `-std=c++23`, and instead uses `-std=c++2b`.

Tested with Clang 14.0.6 and GCC 12.1.1.

Ideally, someone will replace this with `default_options : [ ..., 'cpp_std=c++23' ]` once Meson supports that.

:+1:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,8 +1,22 @@
 project('Hyprland', 'cpp', 'c',
   version : '0.10.0beta',
-  default_options : ['warning_level=2', 'default_library=static', 'optimization=3'])
+  default_options : [
+    'warning_level=2',
+    'default_library=static',
+    'optimization=3',
+    # 'cpp_std=c++23' # not yet supported by meson, as of version 0.63.0
+    ])
 
-add_global_arguments('-std=c++23', language: 'cpp')
+# clang v14.0.6 uses C++2b instead of C++23, so we've gotta account for that
+# replace the following with a project default option once meson gets support for C++23
+cpp_compiler = meson.get_compiler('cpp')
+if cpp_compiler.has_argument('-std=c++23')
+  add_global_arguments('-std=c++23', language: 'cpp')
+elif cpp_compiler.has_argument('-std=c++2b')
+  add_global_arguments('-std=c++2b', language: 'cpp')
+else
+  error('Could not configure current C++ compiler (' + cpp_compiler.get_id() + ' ' + cpp_compiler.get_version() + ') with required C++ standard (C++23)')
+endif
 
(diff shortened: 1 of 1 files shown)


viperML:
Maybe we could fix this warning instead:

```
meson.build:14: WARNING: Consider using the built-in option for language standard version instead of using "-std=c++23"
```

vaxerski:
Does clang actually compile Hyprland? it didnt want to iirc a while ago

viperML:
https://github.com/viperML/Hyprland/actions/runs/2848249132

It doesn't seem like at first glance (standard builds failed when trying to push to cache)

SignalWalker:
> Maybe we could fix this warning instead:
> 
> ```
> meson.build:14: WARNING: Consider using the built-in option for language standard version instead of using "-std=c++23"
> ```

Meson only supports that flag with C++ <= 20 right now, and Hyprland seems to require C++23. This is why I wrote it to use `has_argument` instead of `project(default_options)`.

> https://github.com/viperML/Hyprland/actions/runs/2848249132
> 
> It doesn't seem like at first glance (standard builds failed when trying to push to cache)

Clang uses libc++ by default (instead of libstdc++). `std::unreachable()` was added to libc++ in [version 15](https://libcxx.llvm.org/Status/Cxx2b.html). Nixpkgs's LLVM stdenv derivations clamp their libc++ versions to that of their corresponding Clang binary. The latest version of Clang available in Nixpkgs is `v14.0.6`, so the latest version of libc++ in Nixpkgs is `v14.0.6`.

Those builds failed because you ran them with `llvmPackages_12` and `llvmPackages_13`, which don't provide recent enough versions of libc++ to have `std::unreachable()`.

fufexan:
If it does build in clang fine, I'd say this is ready to merge. We'll be using GCC in the Nix builds anyway. Not sure if there's a benefit to using Clang instead but I'll be happy to hear it out if there is.

SignalWalker:
> If it does build in clang fine, I'd say this is ready to merge. We'll be using GCC in the Nix builds anyway. Not sure if there's a benefit to using Clang instead but I'll be happy to hear it out if there is.

SignalWalker:
> If it does build in clang fine, I'd say this is ready to merge. We'll be using GCC in the Nix builds anyway. Not sure if there's a benefit to using Clang instead but I'll be happy to hear it out if there is.

Honestly it's mostly just because I have `CC=clang` and `CXX=clang++` set and it was making `yay -Syu` fail when updating Hyprland.

vaxerski:
ok

PR #532, opened 2022-08-16
meson: add man pages

viperML:

Adds the manpages to be built with meson.

**Disabled by default**, enabled for the nix builds.

Maybe pandoc is too fat, and it would be nice to use a different program.

Also pinging for second look:
- @volitank implemented man pages in #379 
- @sp1ritCS wrote the meson files

If CI succeeds


vaxerski:
yeah using something that isnt as fat, and additionally doesnt depend on 400MB of shit would be nice.

viperML:
Not that much
```
$ nix path-info -sh (nix build --no-link --print-out-paths --inputs-from github:hyprwm/Hyprland nixpkgs#pandoc)
/nix/store/i56xxnck0qjpqpaafk7x30vy0lwyf17d-pandoc-2.17.1.1      167.8M
```

But I will check for alternatives

fufexan (approved):
LGTM, waiting for the other reviews.

sp1ritCS:
since you are not using the normal `install_man` function and instead use the `install_dir` arg of `custom_target`, you'll want to add `install_tag: 'man'` arg to your taget too.

See https://mesonbuild.com/Installing.html#installation-tags

viperML:
python-docutils seems nice, but produces bad formatting. And I have no idea about RST to fix it

volitank:
@viperML python-docutils is what I originally wrote the man pages for, but we didn't want to depend on python.

If it's decided that we want to go this route then I can change the formatting. It should be in the PR how the formatting should be.

Alternatively to keep them light we could use them how they are with pandoc, and have the converted groff version as well, and that one would actually be used during build time, and only run pandoc when they need updating. Just a thought though, this way would require no extra dependencies during build time.

viperML:
> but we didn't want to depend on python

`docutils` is certainly smaller, at just 3M. It depends on python, but the user probably already has python installed (and it is a runtime dependency of meson....)

```
% nix path-info -shr $(nix build --no-link --print-out-paths --inputs-from github:hyprwm/Hyprland nixpkgs#docutils) | tail -n2
/nix/store/rc9cz7z4qlgmsbwvpw2acig5g2rdws46-python3-3.10.5                82.5M
/nix/store/ncnx31fvf8bvy6rfbkl2amhxrrv112wh-python3.10-docutils-0.19       3.2M
```

epsilon-0:
It is possible to just write the man pages in mdoc language (which is an extremely simplified version of the man language). This way there is no dependency on anything to _build_ man pages, they are straight usable by just copying.

The wf-recorder and some other wayfire utilities use this format - https://github.com/ammen99/wf-recorder/blob/master/manpage/wf-recorder.1

I can volunteer to do the change if you think it's a good idea.

vaxerski:
that looks ugly af lol

but I guess it wouldn't be THAT bad of an idea, no?

viperML:
I'd rather have the man pages pre-compiled on repo, and have a github action watch for changes in the man sources to compile them to a commit

vaxerski:
fair solution as well

epsilon-0:
> that looks ugly af lol
> 
> but I guess it wouldn't be THAT bad of an idea, no?

LOL, yea, they take a bit of getting used too the first time. I've kind of become used to writing them, they're quite a lot like latex.



> I'd rather have the man pages pre-compiled on repo, and have a github action watch for changes in the man sources to compile them to a commit

This would be the next best solution.

PR #535, opened 2022-08-17
Fix workspace animation showing on other monitors

Roger-Roger-debug:
Windows could briefly show up on other monitors when switching workspaces, this stops them from being drawn.

My described issues in Discord happen regardless of this patch (moving floating windows halfway between monitors sometimes draws them over the fullscreen window and sometimes not at all. In either case letting them go while on the fullscreen monitor makes them impossible to move without leaving fullscreen first).

Ready for merging.


vaxerski:
thank you!

PR #548, opened 2022-08-19
add input:kb_file option to use a .xkb file.

FlafyDev:
The PR adds an option to add a .xkb file as your layout instead of using the existing 5 xkb options.

When the option is set to a valid file, it doesn't use the other xkb options.
When the option is set to null or fails, it uses the other xkb options.

* I created a function called `absolutePath` in order to reduce duplicated code but wasn't sure where to put it so I kept it in `ConfigManager`. This is probably not smart to do, but because I'm not familiar with c++, I wasn't sure where to put it.

From what I tested it works.
Tested:
* When the option is set to null
* When the option is set to a full path
* When the option starts with ~
* When the option points to a non existing file (silently fails)


vaxerski:
pretty nice cutie, just one thing: move the absolutePath func to miscfunctions.cpp, and add another param for the cwd instead of configCurrentPath

FlafyDev:
Done. I had to make `configCurrentPath` public, though.

vaxerski:
lgtm. Thanks!

Issue #302, opened 2022-07-04
Fallback to using system wlroots development files


JamiKettunen:
To make packaging this for distros a whole lot nicer (or mergeable in the first place) there should be a fallback to locating system wlroots instead of forcefully trying to use the `Bundled wlroots` under `subprojects/wlroots`.

I know this currently `Closely follows wlroots-git` but are there plans to later make releases that build against latest tagged release of wlroots?

vaxerski:
> to later make releases that build against latest tagged release of wlroots

most likely not.

> To make packaging this for distros a whole lot nicer (or mergeable in the first place) there should be a fallback to locating system wlroots instead of forcefully trying to use the Bundled wlroots under subprojects/wlroots.

not possible.

JamiKettunen:
> most likely not.

why not? I get that the current `main` branch should build and work with latest wlroots, I have absolutely no issue with that, but you'd not consider making releases later build against tagged wlroots versions?

> not possible.

because? I don't see why it wouldn't be; if the releases against stable wlroots happen I don't see why something similar to [sway's old FindWlroots.cmake](https://git.sr.ht/~sircmpwn/sway/tree/6edc32848f67fe3ea370e12b64019cef6c137840/CMake/FindWlroots.cmake) wouldn't work

vaxerski:
wlroots head has many breaking changes compared to latest tagged release. If I wanted to keep releases that have a tagged wlroots ver, I could either:

- release every time wlroots makes a tag (like what 2-6 months?)
- keep two hyprland versions supported

as you can see it'd be a pain in the ass.

The most reasonable version if I was forced at gunpoint to support tagged wlroots releases would be to make a script (or pragmas, ugh) that control code based on the wlroots ver. Even then, some features might break on wlroots 0.16-dev, and not 0.15, and vice versa. Pure pain.

vaxerski:
Still, I don't understand your problem. Hyprland builds wlroots and uses a separate .so, making it independent of the system wlroots. If you want a static link, you can use the meson build with a static flag. Where does it "interfere" with anything?

JamiKettunen:
I just don't think it's normal for release versions of software to require an arbitrary git hash of a dependency that hasn't yet had a tag either; there wouldn't need to be any shenanigans regarding supporting two versions of wlroots at the same time if all development was focused around the latest wlroots tagged release instead.

[...]

PR #552, opened 2022-08-20
add options to use the system wide wlroots library

epsilon-0:

a similar option is used in wayfire
this reduces the burden on recompilation of the whole
subproject when recompiling hyprland, and multiple
wayland compositors are installed

Signed-off-by: Aisha Tammy <floss@bsd.ac>

The default values for the new options makes it behave identical to how it is right now.

Should be ready to merge.


vaxerski:
There is a reason #302 got closed as not planned.

This is a terrible idea. Why?

> this reduces the burden on recompilation of the whole

wlroots takes like 10% of the compile time.

> multiple wayland compositors are installed

Hyprland does not require, overwrite or conflict with any other wlroots installation.

epsilon-0:
It conflicts on Gentoo (which I am trying to add hyprland into) and also on Arch. This is evidenced by the last comment on the issue linked.

Also, I made the PR while keeping the original behaviour as the default one. I'm hoping that this just gives an option to others to make it easily packagable for other operating systems, while still trying to make it in line with your vision.

viperML:
You can just patch the sources if you need to

https://github.com/hyprwm/Hyprland/blob/main/nix/meson-build.patch
https://github.com/hyprwm/Hyprland/blob/9513031da30861b6ab593625aebe88b811758427/nix/default.nix#L74-L77

vaxerski:
> It conflicts on Gentoo (which I am trying to add hyprland into) and also on Arch. This is evidenced by the last comment on the issue linked.

The comment referenced a bug that was fixed months ago, it no longer does that.

Hyprland will build wlroots from the subproject and NOT install it. The only file it will install is `libwlroots.so.11032` which will not conflict with any wlroots installation, and if built with meson, it's statically linked, so no `.so` is ever installed.

> Also, I made the PR while keeping the original behaviour as the default one. I'm hoping that this just gives an option to others to make it easily packagable for other operating systems, while still trying to make it in line with your vision.

The solution you're proposing is inane though - Hyprland will NOT build with any wlroots older than ~ 10-20 commits (aka. **any** tagged release). By keeping it a subproject, we eliminate conflicts. To build Hyprland with the system wlroots, they'd have to match the locked commit - which would INTRODUCE and not eliminate conflicts.

The only thing your option would introduce is a wave of people coming to the issues saying "oMg I have wlroots 0.15.1 and no buildy why :( :( :("

epsilon-0:
> You can just patch the sources if you need to
> 
> https://github.com/hyprwm/Hyprland/blob/main/nix/meson-build.patch
> 
> https://github.com/hyprwm/Hyprland/blob/9513031da30861b6ab593625aebe88b811758427/nix/default.nix#L74-L77

Yes, this is what I am doing currently for the gentoo package :smile_cat: 



> The only thing your option would introduce is a wave of people coming to the issues saying "oMg I have wlroots 0.15.1 and no buildy why :( :( :("
> 
> If you are thinking I am a dickhead and lying, see the [wlroots 0.16.0 meta](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3347) and divide the amount of bullet points by the amount of days since the last tag, about 210, and you get a breaking change every other week.

Got it. I'll just keep the patch locally for gentoo for now. Maybe when wlroots is stabler with its API this can be done :)

vaxerski:
If it works for gentoo, feel free. I will not keep it in the repo here.

Every distro is free to patch the meson / cmake / whatever to their needs. With the parent repo, however, I try to keep things as versatile and safe as possible.

Furthermore, tagged hyprland releases exist, so I mean you can always use those.

PR #557, opened 2022-08-21
Add window{open/close/move} IPC events (#549)

### LINKED ISSUE #549: Add window{open/close/move} IPC events
I am making a custom status bar using the `eww` framework. And I need to know when windows open, close and move between workspaces.

Open/close is pretty straightforward. I tried to emit IPC events in `Events::listener_mapWindow`/`Events::listener_unmapWindow` and it worked fine. 

Move, on the other hand, is tricky, because `CWindow.m_iWorkspaceID` is public and is being changed all over the place. There are at least 3 places where I need to emit the "move" IPC event:
1. Move window via keybind
2. Drag window and switch workspace
3. Pick window, switch workspace and drop it (without mouse movement)

I suggest hiding `CWindow.m_iWorkspaceID` behind getter/setter. Then it will be possible to emit the event just in the setter.

Would you be willing to merge that?

xDarksome:
Resolves #549


src/Compositor.cpp
```diff
@@ -884,6 +884,20 @@ void CCompositor::moveWindowToTop(CWindow* pWindow) {
     }
 }
 
+void CCompositor::moveWindowToWorkspace(CWindow* pWindow, int workspaceID) {
```
vaxerski, line 887: can you move it to CWindow plox cutie

xDarksome: @vaxerski done

Btw, seems like there is already a `CCompositor::moveWindowToWorkspace` but with different signature. I couldn't find any usages of it.

I forgot this language had overloading xD

vaxerski: lul thanks removed it

vaxerski:
make a PR to the wiki if you could plox <3

PR #562, opened 2022-08-22
openSUSE patches

sp1ritCS:

These are the patches I apply to the Hyprland package on openSUSE, as described here:
https://github.com/hyprwm/hyprland-wiki/pull/10

> The package currently patches Hyprland twice, @vaxerski you might consider merging them?
> - it drops the deprecated `gethostbyname("localhost")` call in `hyprctl/main.cpp`, as binaries calling this should not be part of openSUSE (should use getaddrinfo instead); but in this case the call seems to be leftover from a previous iteration anyways and of little use. [gethostbyname.patch](https://opi-proxy.opensuse.org/?obs_api_link=https%3A%2F%2Fapi.opensuse.org%2Fsource%2FX11%3AWayland%2Fhyprland%2Fgethostbyname.patch&obs_instance=openSUSE)
> - it patches the <compiler>.get_version() to <compiler>.version() in `meson.build`, as the get_version() method was seemingly only introduced recently and the version() method was there since forever. This would allow building hyprland with older meson verions (but in the case of the package it was useless, since every distro shipping an outdated meson also had outdated libraries). [old-meson.patch](https://opi-proxy.opensuse.org/?obs_api_link=https%3A%2F%2Fapi.opensuse.org%2Fsource%2FX11%3AWayland%2Fhyprland%2Fold-meson.patch&obs_instance=openSUSE)

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -50,12 +50,6 @@ void request(std::string arg) {
         return;
     }
 
-    const auto SERVER = gethostbyname("localhost");
-
-    if (!SERVER) {
-        std::cout << "Couldn't get host (2)";
-        return;
-    }
 
     // get the instance signature
     auto instanceSig = getenv("HYPRLAND_INSTANCE_SIGNATURE");
--- a/meson.build
+++ b/meson.build
@@ -15,7 +15,7 @@ if cpp_compiler.has_argument('-std=c++23')
 elif cpp_compiler.has_argument('-std=c++2b')
   add_global_arguments('-std=c++2b', language: 'cpp')
 else
-  error('Could not configure current C++ compiler (' + cpp_compiler.get_id() + ' ' + cpp_compiler.get_version() + ') with required C++ standard (C++23)')
+  error('Could not configure current C++ compiler (' + cpp_compiler.get_id() + ' ' + cpp_compiler.version() + ') with required C++ standard (C++23)')
 endif
 
 GIT_BRANCH = run_command('git', 'rev-parse', '--abbrev-ref', 'HEAD', check: false).stdout().strip()


eli-schwartz:
> - it patches the .get_version() to .version() in `meson.build`, as the get_version() method was seemingly only introduced recently and the version() method was there since forever. This would allow building hyprland with older meson verions (but in the case of the package it was useless, since every distro shipping an outdated meson also had outdated libraries). [old-meson.patch](https://opi-proxy.opensuse.org/?obs_api_link=https%3A%2F%2Fapi.opensuse.org%2Fsource%2FX11%3AWayland%2Fhyprland%2Fold-meson.patch&obs_instance=openSUSE)

Hmm, which version of Meson *does* have a get_version method? It's not there recently either.

This probably doesn't usually cause issues, simply because it only causes an error message to break and that would already be a failed build.

sp1ritCS:
@eli-schwartz oh I see. I was already confused why get_version wasn't mentioned in the docs.

I think what happened is, that @SignalWalker (according to git blame) pushed [fix: support clang in meson build (38056c0)](https://github.com/hyprwm/Hyprland/commit/38056c0102e4d92edee1cfbd17e1dea62e3757e9) and because nobody ever ran it on an OS with outdated gcc/(or clang) it seemed to work.

I ran across this while packaging the openSUSE package on openSUSE Leap <=15.4, which ships with gcc 7.5, but because I got a meson parsing error, I incorrectly attributed it to an outdated meson version (Leap is locked to meson 0.59, as python is still at version 3.6); and only after patching the meson file to use the correct method I've seen that the compiler is outdated and thus decided not to pursue making package leap compatible.

still good that it's fixed now ._.

PR #573, opened 2022-08-24
Add cleaninstall task

NotAShelf:
Adds a cleaninstall option in the Makefile, this is so that users that launch Hyprland through a wrapper on login managers don't have to update their hyprland.desktop file everytime they build Hyprland from source. It's a clone of the install task without the part that copies the example desktop file.

Maybe cleaninstall could be something else, but nothing breaking.

Ready for merging.


Makefile
```diff
@@ -137,6 +137,25 @@ install:
 	cp ./assets/wall_8K.png ${PREFIX}/share/hyprland
 
 	install -Dm644 -t ${PREFIX}/share/man/man1 ./docs/*.1
+	
+cleaninstall:
```
vaxerski, line 141: why not
```
make clear
make install
```
?

NotAShelf: correct me if I'm wrong, but `make install` still copies the desktop file to /usr/share/wayland-sessions

and also, I wanted this task to be global, functional regardless of it's an old installation or freshly built from source

vaxerski: oh right yea mb

vaxerski:
update the wiki plox

Issue #178, opened 2022-06-10
Link errors when building without xwayland


asdfjkluiop:
Attempting to build Hyprland when wlroots does not have xwayland support enabled results in linker errors even when `-DNO_XWAYLAND=true` is set. I would expect that setting that flag completely ifdefs out the XWayland code but it doesn't. As it currently stands even if you try to disable XWayland support in hyprland you must have wlroots built with XWayland support and therefore you must have the xwayland dependencies installed even if you do not wish to use them.

vaxerski:
NO_XWAYLAND only disables xwayland run-time. It does not ifdef it. True.

PR #587, opened 2022-08-27
fix xwayland config option for nix

nicholascioli:
When building Hyprland in nix using the following home-manager recipe:

```nix
{}:

{
  wayland.windowManager.hyprland = {
    enable = true;
    xwayland = false;
  };
}
```

The default nix configuration still makes reference to a compile argument `NO_XWAYLAND`, but the `meson_options.txt` only makes reference to `xwayland`. This causes the build to fail with:

```
meson.build:1:0: ERROR: Unknown options: "NO_XWAYLAND"
```

This change fixes that issue.

This still doesn't allow you to build Hyprland without having XWayland installed (#178), but it should at least allow for valid nix configurations for when that issue gets resolved.

The change does not touch any of the actual code, but it might require the root `flake.nix` / `flake.lock` to be updated to bump the patch version.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -72,7 +72,7 @@ stdenv.mkDerivation {
     else "release";
 
   mesonFlags = builtins.concatLists [
-    (lib.optional (!enableXWayland) "-DNO_XWAYLAND=true")
+    (lib.optional (!enableXWayland) "-Dxwayland=disabled")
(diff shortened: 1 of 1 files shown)


vaxerski:
@fufexan

viperML (approved):

fufexan:
Thanks!

PR #591, opened 2022-08-28
Nix: add XWayland hidpi option

fufexan:
Adds `hidpiXWayland` flag and the required logic to get it working.
Based on https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/733
The flag will be enabled by default, as I haven't had any issues with it in the past few days.

Users will have to adjust their config files accordingly, possibly having to add
```sh
exec-once=xprop -root -f _XWAYLAND_GLOBAL_OUTPUT_SCALE 32c -set _XWAYLAND_GLOBAL_OUTPUT_SCALE 2
```
in order for XWL to scale. This will make XWL programs look as if they were unscaled. To fix this, users have to export different environment variables to make the specific toolkits render at the proper scaling. For example
```sh
export GDK_SCALE=2
export XCURSOR_SIZE=32
```
The `GDK_SCALE` variable won't conflict with wayland-native GTK programs.

I'm still not sure whether we should go with this approach, or rather having a separate package.

@viperML @Misterio77 @nix guys for reviews

### DIFF
--- a/.github/workflows/nix-build.yaml
+++ b/.github/workflows/nix-build.yaml
@@ -5,6 +5,11 @@ jobs:
   nix:
     name: "Build Hyprland (Nix)"
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package:
+        - default
+        - hyprland-no-hidpi
     steps:
     - name: Clone repository
       uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


viperML:
I would add another job to the CI action to build this

viperML:
And also move the patching into its own wlroots.nix file

fufexan:
@viperML I updated the HM module as well. Do you think I should also add a `mkRenamedModule` entry? Essentially it's `xwayland -> xwayland.enable`.

viperML:
Sure. I don't even think anyone would use that option to disable xwayland

fufexan:
Done. Mergeable?

viperML:
Yes merge

PR #684, opened 2022-09-19
fix some bugs with attaching/disabling monitors

tomahk:

Fixes problems with reenabling monitors and freezes when undocking from dockingstation
works for me, other should test it
yes

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -65,8 +65,8 @@ void CMonitor::onConnect(bool noRule) {
 	}
 
     if (!m_bRenderingInitPassed) {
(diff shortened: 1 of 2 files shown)


src/helpers/Monitor.cpp
```diff
@@ -65,6 +65,8 @@ void CMonitor::onConnect(bool noRule) {
 	}
 
     if (!m_bRenderingInitPassed) {
+        output->allocator = NULL;
+        output->renderer = NULL;
```
vaxerski, line 69: this is C++, use `nullptr`

tomahk: done

PR #715, opened 2022-09-23
fix crash

### LINKED ISSUE #711: New crash!
This is not that easy to reproduce, but i get it to crash eventually if i try long enough.
It  happens when i group all 3 windows of a workspace together.
Than activate fullscreen, changegroupactive a few times and then finally togglegroup makes it crash.
As mentioned this is happening not every time, sometimes it does the first try, sometimes it takes a while.
[coredump.log](https://github.com/hyprwm/Hyprland/files/9624444/coredump.log)

Dickby:
Fixes #711

I'm pretty shure this is not how you would do it!
there is a problem in toggleWindowGroup if that group is in fullscreen mode.
I think the pPrevious- and pNextGroupMember of the first window are set to nullptr, after that 
the nextGroupMember calls getGroupMemberCount, but with a Pointer to a group member that
has it's next prev pointers nulled, getGroupMemberCount will crash.
Since it is a crashing bug, if you don't have time to fix it yourself, use this first?

I think yes.
It's not a big change, look at it, what could possibly go wrong!?

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -819,15 +819,13 @@ void CHyprDwindleLayout::toggleWindowGroup(CWindow* pWindow) {
             PWINDOW->m_bHidden = false;
         }
 	    
-	if(PHEAD->pPreviousGroupMember)
-	{
-	  PHEAD->pPreviousGroupMember->pNextGroupMember = PHEAD->pNextGroupMember;
-	  PHEAD->pPreviousGroupMember = nullptr;
+	if(PHEAD->pPreviousGroupMember){
+	    PHEAD->pPreviousGroupMember->pNextGroupMember = PHEAD->pNextGroupMember;
+	    PHEAD->pPreviousGroupMember = nullptr;
 	}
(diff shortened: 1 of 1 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -818,9 +818,18 @@ void CHyprDwindleLayout::toggleWindowGroup(CWindow* pWindow) {
 
             PWINDOW->m_bHidden = false;
         }
+	    
+	if(PHEAD->pPreviousGroupMember)
+	{
```
vaxerski, line 823: style: indent is 4 and braces are not on newlines

Dickby: done?

vaxerski: ugh

Dickby: what?

vaxerski: fixed it for you. Why cant you just look around how code looks like..

Dickby: well thanks, i guess.
I will try harder next time!

Issue #718, opened 2022-09-23
border_size modifies gap size on monitor scale other than 1


bryceberger:
Steps to reproduce: set monitor scale to a value other than 1 (noticed on 1.75, happens on 0.8, 1.1, 1.5, 2.0, etc). set gaps_in and gaps_out to 0. set border_size to a nonzero value (larger is easier to see)

Expected outcome: No wallpaper visible

Noted outcome: Wallpaper visible (when monitor scale above 1. Below 1 results in strange overlap of borders)

Estebiu:
Also, if you have a multi-monitor setup with different scale settings, there'll be inconsistencies between them.
On one the borders will be too big or too small in comparation to the other one.

PR #722, opened 2022-09-25
Scale the window borders.

Dickby:
Fixes https://github.com/hyprwm/Hyprland/issues/718

Maybe it's possible that the scaling and later casting to an integer (which is basicly truncation) could introduce a value
that is 1 lower than wanted in some cases.
I tried it for reasonable values, and it seems to work fine.

I think so!

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -780,27 +780,29 @@ void CHyprOpenGLImpl::renderBorder(wlr_box* box, const CColor& col, int round) {
     if (*PBORDERSIZE < 1)
         return;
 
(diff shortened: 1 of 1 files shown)


Dickby:
The point could be made, that a 5 pixel border should be 5 pixels when scaled too.
But because the gaps are scaled too, i think this is the right solution.

Dickby:
Thinking about it a bit more, i now prefer the other way!
I will make a new pull request where the bordersize is scaled not in the render part, but 
where it's used to calculate the window sizes, that way the windowsizes and borders stay consistent
when changing the scale!
If you prefer that way too, don't merge this yet

PR #727, opened 2022-09-25
Scale border size in window size pos calculation

Dickby:
Fixes https://github.com/hyprwm/Hyprland/issues/718

Don't see any

I think it's ready.

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -182,15 +182,16 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
     PWINDOW->m_vPosition = pNode->position;
 
     static auto *const PNOGAPSWHENONLY = &g_pConfigManager->getConfigValuePtr("dwindle:no_gaps_when_only")->intValue;
+    float scaledBorderSize = *PBORDERSIZE / PMONITOR->scale;
 
-    auto calcPos = PWINDOW->m_vPosition + Vector2D(*PBORDERSIZE, *PBORDERSIZE);
-    auto calcSize = PWINDOW->m_vSize - Vector2D(2 * *PBORDERSIZE, 2 * *PBORDERSIZE);
+    auto calcPos = PWINDOW->m_vPosition + Vector2D(scaledBorderSize, scaledBorderSize);
+    auto calcSize = PWINDOW->m_vSize - Vector2D(2 * scaledBorderSize, 2 * scaledBorderSize);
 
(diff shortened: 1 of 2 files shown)


fufexan:
I've built this but I see no difference in border size.

Dickby:
The border size shouldn't change, and now the window gaps don't change either (on scale change).
Did you read the issue?

Dickby:
I closed a pull request yesterday that did it the other way around.
By scaling the bordersize inside the renderer.
But i think this solution is nicer because window size and border size now are consistent on scale changes.
@vaxerski : if prefer the first solution (would feel more like actual monitor scaling, this one is more window scaling)
i will change it again.

Estebiu:
How do you build merge requests? 
I'd like to try it.
Unrelated, but cool merge request number (727)

Dickby:
@Estebiu you could clone and build my fork.
Probably there is a nicer way though.

Dickby:
That test commit was not planned to be part of the pull request,
i didn't made a branch for the pr and now i just tried to push from my machine with a new gh token.
But i think it doesn't hurt.

fufexan:
Took a look at the issue and tested again, seems to work fine.

If you could squash that last commit or rename it, that'd be great.

Dickby:
Done!
nobody should give me write access until i really figured out git/github !

fufexan (approved):
LGTM. I'll let Vaxry comment or merge.

src/render/OpenGL.cpp
```diff
@@ -1124,7 +1124,7 @@ void CHyprOpenGLImpl::renderSplash(cairo_t *const CAIRO, cairo_surface_t *const
     cairo_text_extents_t textExtents;
     cairo_text_extents(CAIRO, g_pCompositor->m_szCurrentSplash.c_str(), &textExtents);
 
-    cairo_move_to(CAIRO, m_RenderData.pMonitor->vecPixelSize.x / 2.f - textExtents.width / 2.f, m_RenderData.pMonitor->vecPixelSize.y - textExtents.height - 1);
```
vaxerski, line 1127: wot

Dickby: should i take that out? It was the result of commit to test the gh token i just made.
Didn't know that it will change the pull request.
Shouldn't change the behavior, and if unoptimized, it should be a bit more efficient.

Dickby: I changed the 2.f to 2.0 since all the values are doubles and cairo_move_to takes doubles too.

Dickby: By the way, am i supposed to resolve a conversation like this?

vaxerski: remove this change. it's unrelated to the pr

PR #734, opened 2022-09-25
Scale the window gaps.

Dickby:
Scales the gaps between the windows.

If this is the way to go, shadow range and corner rounding should be scaled too,
probably shadow offset.

If there is a multi monitor problem with the border scaling thing, 
it will probably effect this too!


vaxerski:
I'll test this once I get my PC set up. Probably wednesday. Ping me on Wed

Dickby:
Ok.
I will try to extend it for shadows and rounding.

Issue #679, opened 2022-09-18
Hyprland splash text incorrectly positioned


godalming123:
The text is positioned properly on my monitor but is on the left on my laptop. Will upload photo.

godalming123:
Laptop monitor:
![IMG20220918173438.jpg](https://user-images.githubusercontent.com/68993177/190918029-4136af30-7407-4edc-9780-9925ab544ff8.jpg)
Main monitor:
![IMG20220918173425.jpg](https://user-images.githubusercontent.com/68993177/190918066-28e453c2-e32e-45a9-88b0-b8f083bd9140.jpg)

vaxerski:
what are the monitor settings? could you paste them here in a code block?

godalming123:
```

monitor=HDMI-A-1,1920x1080,0x0,1
monitor=eDP-1,1366x768,300x1080,1
#monitor=eDP-1,disable
workspace=eDP-1,1,1
workspace=HDMI-A-1,2
```

godalming123:
If it helps to know this happens when monitor is plugged in or unplugged.

PR #735, opened 2022-09-26
Splash position fix

Dickby:
This fixes the offcenter splash rendering

Fixes https://github.com/hyprwm/Hyprland/issues/679
On screens with a higher than texture ratio, the splash will be offscreen (below),
this is not introduced by this pr.
The splash has to be moved by the same amount as the origin.y value
in createBGTextureForMonitor in those cases.

I think yes, it doesn't fix the problem totally, but it makes things better.


vaxerski:
I'll wait for when you fix that other issue. Let's just make one PR to end those alignment issues once and for all

Dickby:
@vaxerski  this is done!
I tested it like crazy with different monitor settings.
I eat my shoe if this is not working!

vaxerski:
thanks as usual

PR #737, opened 2022-09-26
Replace clamp with max if there is no upper bound.

Dickby:
prevents an unnecessary upper bound check

No.

Should be good to go.


vaxerski:
rip my 1337420 placeholders

PR #742, opened 2022-09-28
Rework rounding shader

Dickby:
This optimizes the rounding part of the Textures.hpp shaders.
Also the Shaders.h file is now only included by OpenGL.cpp, 
that way only OpenGL.cpp has to be recompiled if a shader was changed.

It shouldn't change the behavior at all.

I tried it, and it is working for me, but should be tested in other configurations/ on other computers.

### DIFF
--- a/src/render/shaders/Textures.hpp
+++ b/src/render/shaders/Textures.hpp
@@ -13,9 +13,6 @@ inline static constexpr auto ROUNDED_SHADER_FUNC = [](const std::string colorVar
 
     if (pixCoord.x + pixCoord.y > radius) {
 
-	if (ignoreCorners == 1)
-	    discard;
-
 	float dist = length(pixCoord);
 
 	if (dist > radius)
@@ -64,7 +61,6 @@ uniform vec2 fullSize;
 uniform float radius;
 
 uniform int primitiveMultisample;
-uniform int ignoreCorners;
 
 void main() {
 
@@ -104,7 +100,6 @@ uniform int applyTint;
 uniform vec3 tint;
 
 uniform int primitiveMultisample;
-uniform int ignoreCorners;
 
 void main() {
 
@@ -141,7 +136,6 @@ uniform int applyTint;
 uniform vec3 tint;
 
 uniform int primitiveMultisample;
-uniform int ignoreCorners;
 
 void main() {
 
@@ -227,7 +221,6 @@ uniform int applyTint;
 uniform vec3 tint;
 
 uniform int primitiveMultisample;
-uniform int ignoreCorners;
 
 void main() {
 
--- a/src/render/shaders/Textures.hpp
+++ b/src/render/shaders/Textures.hpp
@@ -7,11 +7,11 @@ inline static constexpr auto ROUNDED_SHADER_FUNC = [](const std::string colorVar
 
     // branchless baby!
     highp vec2 pixCoord = vec2(gl_FragCoord);
-    pixCoord -= topLeft + fullSize / vec2(2.0);
-    pixCoord *= vec2(lessThan(pixCoord, vec2(0.0))) * vec2(-2.0) + vec2(1.0);
-    pixCoord -= fullSize * vec2(0.5) - vec2(radius);
+    pixCoord -= topLeft + fullSize * 0.5;
+    pixCoord *= vec2(lessThan(pixCoord, vec2(0.0))) * -2.0 + 1.0;
+    pixCoord -= fullSize * 0.5 - radius;
 
-    if (all(greaterThan(pixCoord, vec2(0.0)))) {
+    if (pixCoord.x + pixCoord.y > radius) {
 
 	if (ignoreCorners == 1)
 	    discard;
--- a/src/render/shaders/Textures.hpp
+++ b/src/render/shaders/Textures.hpp
@@ -22,15 +22,15 @@ inline static constexpr auto ROUNDED_SHADER_FUNC = [](const std::string colorVar
 
 	if (primitiveMultisample == 1 && dist > radius - 1.0) {
 	    float distances = 0.0;
-	    if (length(pixCoord + vec2(0.25, 0.25)) < radius) { distances = distances + 1.0; }
-	    if (length(pixCoord + vec2(0.75, 0.25)) < radius) { distances = distances + 1.0; }
-	    if (length(pixCoord + vec2(0.25, 0.75)) < radius) { distances = distances + 1.0; }
-	    if (length(pixCoord + vec2(0.75, 0.75)) < radius) { distances = distances + 1.0; }
+	    distances += float(length(pixCoord + vec2(0.25, 0.25)) < radius);
+	    distances += float(length(pixCoord + vec2(0.75, 0.25)) < radius);
+	    distances += float(length(pixCoord + vec2(0.25, 0.75)) < radius);
+	    distances += float(length(pixCoord + vec2(0.75, 0.75)) < radius);
 
 	    if (distances == 0.0)
 		discard;
 
-	    distances = distances / 4.0;
+	    distances /= 4.0;
 
 	    )#" + colorVarName + R"#( = )#" + colorVarName + R"#( * distances;
         }
--- a/src/render/shaders/Textures.hpp
+++ b/src/render/shaders/Textures.hpp
@@ -5,38 +5,38 @@
 inline static constexpr auto ROUNDED_SHADER_FUNC = [](const std::string colorVarName) -> std::string {
         return R"#(
 
-	// branchless baby!
-	vec2 pixCoord = v_texcoord - vec2(0.5);
-	pixCoord *= (vec2(lessThan(pixCoord, vec2(0.0))) * vec2(-2.0) + vec2(1.0)) * fullSize;
-	pixCoord -= (bottomRight - topLeft) * vec2(0.5);
+    // branchless baby!
+    vec2 pixCoord = v_texcoord - vec2(0.5);
+    pixCoord *= (vec2(lessThan(pixCoord, vec2(0.0))) * vec2(-2.0) + vec2(1.0)) * fullSize;
+    pixCoord -= (bottomRight - topLeft) * vec2(0.5);
 
-	if (all(greaterThan(pixCoord, vec2(0.0)))) {
+    if (all(greaterThan(pixCoord, vec2(0.0)))) {
 
-		if (ignoreCorners == 1)
-			discard;
+	if (ignoreCorners == 1)
+	    discard;
 
-		float dist = length(pixCoord);
+	float dist = length(pixCoord);
 
-		if (dist > radius)
-			discard;
+	if (dist > radius)
+	    discard;
 
-		if (primitiveMultisample == 1 && dist > radius - 1.0) {
-			float distances = 0.0;
-			if (length(pixCoord + vec2(0.25, 0.25)) < radius) { distances = distances + 1.0; }
-			if (length(pixCoord + vec2(0.75, 0.25)) < radius) { distances = distances + 1.0; }
-			if (length(pixCoord + vec2(0.25, 0.75)) < radius) { distances = distances + 1.0; }
-			if (length(pixCoord + vec2(0.75, 0.75)) < radius) { distances = distances + 1.0; }
+	if (primitiveMultisample == 1 && dist > radius - 1.0) {
+	    float distances = 0.0;
+	    if (length(pixCoord + vec2(0.25, 0.25)) < radius) { distances = distances + 1.0; }
+	    if (length(pixCoord + vec2(0.75, 0.25)) < radius) { distances = distances + 1.0; }
+	    if (length(pixCoord + vec2(0.25, 0.75)) < radius) { distances = distances + 1.0; }
+	    if (length(pixCoord + vec2(0.75, 0.75)) < radius) { distances = distances + 1.0; }
 
-			if (distances == 0.0)
-				discard;
+	    if (distances == 0.0)
+		discard;
 
-			distances = distances / 4.0;
+	    distances = distances / 4.0;
 
-			)#" + colorVarName + R"#( = )#" + colorVarName + R"#( * distances;
-		}
+	    )#" + colorVarName + R"#( = )#" + colorVarName + R"#( * distances;
+        }
 
-	}
-	)#";
+    }
+)#";
 };
 
 inline const std::string QUADVERTSRC = R"#(
@@ -67,13 +67,14 @@ uniform int primitiveMultisample;
 uniform int ignoreCorners;
 
 void main() {
-	vec4 pixColor = v_color;
 
-	if (radius > 0.0) {
-                )#" + ROUNDED_SHADER_FUNC("pixColor") + R"#(
-	}
+    vec4 pixColor = v_color;
 
-	gl_FragColor = pixColor;
+    if (radius > 0.0) {
+	)#" + ROUNDED_SHADER_FUNC("pixColor") + R"#(
+    }
+
+    gl_FragColor = pixColor;
 })#";
 
 inline const std::string TEXVERTSRC = R"#(
@@ -83,8 +84,8 @@ attribute vec2 texcoord;
 varying vec2 v_texcoord;
 
 void main() {
-	gl_Position = vec4(proj * vec3(pos, 1.0), 1.0);
-	v_texcoord = texcoord;
+    gl_Position = vec4(proj * vec3(pos, 1.0), 1.0);
+    v_texcoord = texcoord;
 })#";
 
 // this is texture rendering!!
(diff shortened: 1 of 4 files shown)


vaxerski:
have you actually done any benchmarks to back your performance claim?

Dickby:
I didn't, but just counting the number of branches that it has to take for each if the 5 possible paths it can take
(the 4 corners, the rest), and the knowledge that branching will be the bottleneck, this is more efficient.

vaxerski:
branching is a bottleneck? Didn't know that. I am not very proficient with GPUs, but a branch to me is just
```
cmp <reg> <reg2>
jne <addr>
...
jmp <addr2>
```

If you are sure that this is more optimized, (I'll also tag @fufexan to test) please also test that it works under GLES2 (build with LEGACY_RENDERER)

src/render/shaders/Textures.hpp
```diff
@@ -3,138 +3,38 @@
...
-
-					distances = distances / 4.0;
-
-					)#" + colorVarName + R"#( = )#" + colorVarName + R"#( * distances;
-				}
-			}
+
+			distances = distances / 4.0;
+
+			/* )#" + colorVarName + R"#( = )#" + colorVarName + R"#( * distances; */
```
vaxerski, line 35: why is this commented out? Where do you set the multisample alpha? I see it working but am confused

src/render/shaders/Textures.hpp
```diff
@@ -309,17 +197,17 @@ uniform vec2 halfpixel;
 void main() {
 	vec2 uv = v_texcoord / 2.0;
 
-    vec4 sum = texture2D(tex, uv + vec2(-halfpixel.x * 2.0, 0.0) * radius);
+        vec4 sum = texture2D(tex, uv + vec2(-halfpixel.x * 2.0, 0.0) * radius);
```
vaxerski, line 200: this wasnt intended I think

Dickby:
Did you ever hear about prefetching?
A branch is a totally performance killer if it can't be predicted the right way.
That means all the prefetched data is worthless and has to be newly created.

src/render/shaders/Textures.hpp
Dickby: damn, i forgot that.
I'm trying to fix another thing and i had to find out which shader does what.

src/render/shaders/Textures.hpp
Dickby: Shouldn't it be 8 spaces?

Dickby:
It's the same on cpu (even more important because it's less parallel)

src/render/shaders/Textures.hpp
vaxerski: why??? it's the first indent level, so 4.

vaxerski:
me using a ton of ifs:

ok

Dickby:
You know, there are just a few places in a program where it really makes a difference.
I there is a function with lot's of branches that will execute only a few times per frame, it's not worth the effort

Dickby:
By the way the Multisampling thing inside that shader, it was shifting the different positions for each corner in the same direction, was that intended? because the way it's in this pr, it shifts them towards the rounding, of every corner
i think that is the right direction, but you wrote that code so i'm not shure.

vaxerski:
> By the way the Multisampling thing inside that shader, it was shifting the different positions for each corner in the same direction, was that intended? because the way it's in this pr, it shifts them towards the rounding, of every corner
i think that is the right direction, but you wrote that code so i'm not shure.

Yeah thats better.

src/render/shaders/Textures.hpp
Dickby: But look inside that shader, its everywhere 8 inside the string, even the vec2 uv... above that.
I was trying to keep the style.

vaxerski: weird, it looks really off on my end. Defo wasnt intended.

```
void main() {
    // 4 spaces
        // 8 spaces
}
```

Dickby:
Don't merge! I didn't test the multisampling.
I will notify you when i'm shure!

src/render/shaders/Textures.hpp
Dickby: Is this ok though?

Dickby:
Multisamplig was the other way around.
Now it works!
Ready to merge if you want to.

src/render/shaders/Textures.hpp
vaxerski: na make it 4

Dickby: the whole file?

vaxerski: wherever you changed it, the rest seems to be 4 to me

Dickby: One second!

fufexan:
image
Looks fine but there are some artifacts in the bottom left corners of the windows.

src/render/shaders/Textures.hpp
Dickby: just for the record, this file wasn't 4 spaces before.
Next time i find a file like that i let you fix it before i touch it !

Dickby: Your resolve this conversation.

vaxerski: no clue then. In my vscode it was 4 spaces. weird shit.

Dickby: Man i kind of get used to the tone in this repo!

vaxerski: good

Dickby:
@fufexan I don't have that! can you give me your config?

fufexan:
https://github.com/fufexan/dotfiles/blob/main/home/wayland/hyprland/config.nix

Dickby:
Ok @fufexan  check if it's gone if you disable multisampling

Dickby:
And is it in all windows or only that terminal?

Dickby:
And did you build after  [force-pushed the rework_rounding_shader branch from 78637a3 to 168a326 ](https://github.com/hyprwm/Hyprland/pull/742/commits/168a3266095ab014eee1363d8c943f91d461aa40)

fufexan:
> And did you build after [fo](https://github.com/hyprwm/Hyprland/pull/742/commits/168a3266095ab014eee1363d8c943f91d461aa40)

Yes.

> And is it in all windows or only that terminal?

All windows afaict.

> Ok @fufexan check if it's gone if you disable multisampling

Still there.

Dickby:
Me neither!
@fufexan could you please play with your settings (scale, decorations...) and see if you can
find anything that changes the artifacts size, position, color or anything?

Are you totally shure it isn't there without this change?

Dickby:
@fufexan and try to disable all decorations, is it still there without them?

fufexan:
1. Scaling affects it. On scaling 1 it does not show, but on more it does, even on 2.
2. It only happens for rounding > 6 on scale 1.6, rounding > 14 on scale 2. For scale 1.8 it happens on rounding > 3. Really weird and inconsistent.

> Are you totally shure it isn't there without this change?

Yes. I would've noticed it if it were the case.

Dickby:
I will push another Thing today or tomorrow, that fixes lots of animation stuff (e.g. the corner rounding thing on chromium [or was it firefox?]).
It's possible that this is caused by the same problem.

Dickby:
I know i will change a few things in the shaders. it already works, but i have to work on the decoration too to don't have 1 off artifacts.

vaxerski:
maybe make this a draft then for now?

Dickby:
The UV coordinates are not really working there, i tried that days ago.
The problem with that approach is, that during resizing animations thos UV coordinates don't change dynamicly enough.
It depends on the program that is running in that window.
I solve it in another way, once and for all!

Dickby:
what's a draft?

vaxerski:
this

> The problem with that approach is, that during resizing animations thos UV coordinates don't change dynamicly enough.

Noticed that, it's a tiny bit off. No clue why but oh well.

Dickby:
There is a proper solution for that, i have it working with the textures, but during animations white borders will look a bit red. I think i know how to fix that!

vaxerski:
good luck and thanks

Dickby:
@vaxerski what you could try to fix in the meantime:

The clients window has standard uv coordinates as long as the size of the window matches the size of the texture of that client. Now if a client has a small window, and you make it bigger (e.g. by closing another window).
Hyprland will notice that client of the size change, than the client will change it's size, but not immediately,
Hyprland should change the calculated uv's at the point when the internal client texture size changes, not 
when the window resizing action is executed.

Dickby:
That would make a more fluid animation i think!

vaxerski:
probably. But it's not a pressing issue at the current moment, as the error is like a couple pixels at most. I'll take a look if I'll have time and nothing better to do

Dickby:
It's not a couple of pixels, its just that jump of scale. 
Like it is now, before that changed the window looks streched and than it jumps to the right new size.
It's not pressing though, you're right there.

Dickby:
Damn that red looking border is alway there. Without my commit too.
Or maybe it isn't at all, and my eyes are just fucked up!
Will update in a second!

Dickby:
People noticed it with chromium because there it's visible without animation.
I myself noticed first that the corners of opening windows (while growing) wheren't all rounded.
But it was even more, i made it really visible here
before:

https://user-images.githubusercontent.com/10100765/193168154-96eb07bd-ea4e-4dee-b8cd-f7df3680e94c.mp4



after:


https://user-images.githubusercontent.com/10100765/193168272-3caef869-e310-48ad-8048-829494a8e1f0.mp4

Dickby:
By the way @vaxerski, you should look up how floating point numbers work, and why you are really asking for 
trouble when you compare them for equality!

vaxerski:
I know how they work, thats why in a bunch of places I use an epsilon of 1.

Dickby:
Might be, but an operator== and operator!= for Vector2d that work the way they do in Hyprland are just a bit scary to look at.

Dickby:
Is there a way to send pm's on github?

Dickby:
I don't know another place to message you, so i write it here.
There are issues concerning the performance of the program, i didn't read the whole program of course, but
seeing the render part, i know that we could have it orders of magnitude faster (i bet at least 10x faster, if not more).
But we would need to restructure it quite a bit.
How do you think about that? Are you willing to do that?

vaxerski:
> Is there a way to send pm's on github?

no afaik, join the discord

> There are issues concerning the performance of the program, i didn't read the whole program of course, but
seeing the render part, i know that we could have it orders of magnitude faster (i bet at least 10x faster, if not more).
But we would need to restructure it quite a bit.
How do you think about that? Are you willing to do that?

I am if that's the difference we're talking about

Dickby:
It is shurely!
I would take the Renderer.cpp and Opengl.cpp + the shaders if you want me too, but you have to restructure
the data that is send to those.

vaxerski:
What's the issue there?

Dickby:
We need all the data that is needed for rendering of the windows more or less together and ordered in a certain way.
And for the love of god, we need to get rid of virtual draw functions and those things.
We need to do all calculations for preparing the windows in loops, don't call those functions for every window seperately.
The windows that havent moved doesn't need to be recalculated at all. maybe it's possible to don't even redraw that window, the windows have to be drawn in an order that makes the rendering more efficient:
all the windows that are tiled and not animated at the moment first, than the ones floating, in the order they are drawn now.
than the special workspaces of course.
The border, shadow, texture, blur shaders, use same calculations all the time, so we should make it just one shader.
We could even draw all non overlapping windows in the same shader call.
to calculate shadows and borders inside the fragment shader for every window is not very efficient, since it's the same 
values all the time for the same distances, we could render a texture once for those things and then just,
set the vertexes appropriately (don't scale the corners, but scale the edges, it would be enough to have a texture of one rounded corner of the actual shadow to render it (by stretching the edges and rotation the corner, shaders are fucking fast at those things).
The slow thing about a gpu render  system is the communication between CPU and GPU, so we could/should reduce that to some places. We can have texture and vertex arrays for all the windows that can be drawn at the same time, just one call to one shader program. Inside that shader we draw all the window and the decorations at once.
The only reason why we can't do the whole rendering in one, is blur, so we have to think about how to do this the best way.
Of course we would have to try, and find the best way to do it.

Dickby:
@fufexan : are the artifacts still there?

vaxerski:
> And for the love of god, we need to get rid of virtual draw functions and those things.

You mean the decorations' draw()?

> The windows that havent moved doesn't need to be recalculated at all.

True, but shouldRenderWindow() should already get rid of a bunch of those no?

> We could even draw all non overlapping windows in the same shader call.

wot

Generally you bring valid points, but nobody (literally nobody) does this afaik, because it leads to more code clutter and untimately bugs over not *that* much of a performance boost. Can't imagine how this would boost the performance 10x. Simply can't, considering the difference between sway, xorg or wf and Hyprland, on similiar settings, is small.

Dickby:
I could start by writing a big shader that combines all of the window shaders.
The shader for the "wp"/splash should probably be seperated.

vaxerski:
try and benchmark it.

Dickby:
> You mean the decorations' draw()?

Yes, every window should have all those decoration values together, and if you draw a window (should probably be one renderWindow function)

> True, but shouldRenderWindow() should already get rid of a bunch of those no?

Just mark them as has to be recalculated, no need to check every frame for every window. 
Or at least have a function that checks all the windows in a loop.


> wot
> 
> Generally you bring valid points, but nobody (literally nobody) does this afaik, because it leads to more code clutter and untimately bugs over not _that_ much of a performance boost. Can't imagine how this would boost the performance 10x. Simply can't, considering the difference between sway, xorg or wf and Hyprland, on similiar settings, is small.

Those other window managers are not dealing with overlapping or moved windows.
and if you read the issue "idle cpu usage is over 9000" you see that Hyprland is much resource intensive than e.g. sway.
They just have to check if a window has changed, thats all, we are dealing with animations/ dynamic overlapping and so on
(because Hyprland is doing nicer things!).
But we should really try to make it an window manager that doesn't eat up 10 percent of a laptops performance.
In the end a window manager is just a program that let's open other graphical programs.

vaxerski:
> Those other window managers are not dealing with overlapping or moved windows.

Every single one I mentioned does deal with it, especially xorg or wayfire, wayfire has even more effects than hl.

> and if you read the issue "idle cpu usage is over 9000" you see that Hyprland is much resource intensive than e.g. sway.

It's constant. it does not scale up. Hyprland uses around 0.5-2W more on my laptop than sway regardless of what is open. It's prolly some polling or stuff

Dickby:
> try and benchmark it.

Not shure if drawing lot's of windows in the same shader call is what we end up with, but if it's possible, it should be checked.
In general i'm talking of a approach of data oriented design, opposed to object oriented.
I know lot's of programs now today are really not taking that into account, but that's the reason why lot's of programs are fucking slow compared to how fast todays computers are!
If you want to have something to look up, google mike acton or casey muratori (handmade hero), jonathan blow.
I really like this talk: https://www.youtube.com/watch?v=rX0ItVEVjHc .

vaxerski:
Provide a benchmark that shows considerable gains and I'll look into this.

Dickby:
How should i do that without writing the whole thing?

vaxerski:
I think you'd just have to write a part of it :P

Dickby:
I will start writing a general window shader (for all the things one window needs) that should make a difference right away.
But for that we need to combine at least shadow, border, blur and texture rendering functions.

Dickby:
Let's wait for @fufexan to test this draft though!

Dickby:
> I really like this talk: https://www.youtube.com/watch?v=rX0ItVEVjHc .

I would really recomment to watch this!

Dickby:
@vaxerski If @fufexan 's issue is solved, feel free to merge!

Dickby:
Don't merge this yet! this seems to be indeed not faster, i will check exactly what's the problem is,
and pick and choose what should stay!

Dickby:
I'm closing this and will open a new pr when i checked!

Dickby:
I think i saw some ghost's yesterday night. 
Today i tried to fix the thing that causes the chromium rounding bug without this shader change, and i felt 
like i would revert something that was already there before and was taken out!
Or as if someone already prepared to do what i was doing.
Did you write the shaders yourself?
The solution with this change however is much cleaner!
I would like this to be tested, and @fufexan could maybe try if the artifact is still there.
If somebody has a good way of benchmarking shaders that would be helpful, 
the benchmark tests i came up with are very limited.
After some more limited quality testing, i could notice no speed difference to the old shader, but with the new way 
the corner problem is fixed.
Make with it what you think makes sense.

vaxerski:
> Did you write the shaders yourself?

yes

> If somebody has a good way of benchmarking shaders that would be helpful,
the benchmark tests i came up with are very limited.

What did you try? My initial thought would be to just time it, but I am not sure whether that's going to work with CPU-GPU shit. Prolly good idea to google. I'll see.

fufexan:
@Dickby the artifacts are indeed fixed. Thanks!

Dickby:
@vaxerski : Sorry for the late response, 
what i tried was to try to have that shaders running a lot, by opening lot's windows and having as much blur passes that the framerate is not fully 60 and not totally down to 30. than i compared those framerate drops for the 2 shaders.
But because i didn't find a way to disable vsync, this might not be very good information.
It looked like the rounding part of the shaders are not changing that much at all when i tested it against the same setup with no rounding at all.
Since my version of the shader has less branching, but a bit more calculations, i would guess that it performs a bit better if
gpu cores are limited, but might be performing a bit worse on a proper graphics card with lots of cores.
Because a branch on a graphics card with SIMT (single instruction multiple threads) basicly goes both ways by using 2 cores
and deciding later which result to use.
But those cards are quite advanced and i might be wrong here.
If i'm right though, i think it would make sense to prefer a performance gain on slower systems, because for a faster
setup running a few small shaders like this shouldn't be a concern at all.
And after @fufexan 's reported artivacts are gone (thanks for testing), i would consider to merge this.
But as always it's up to you!

Dickby:
@vaxerski What about this change?
Is there something you want me to clarify?
Should i add some comments, or are you not happy with some of the changes?
I use it since days and it fixes the animation/chromium problem, it makes the multisampling more consistent on each corner,
the shader has to test only half the number of fragments against the radius (so computing the distance is used only half as much), and i don't see any downsides.

vaxerski:
sorry, I forgor :skull: about it cuz it was a draft and thought you were still working on it

Dickby:
@vaxerski thanks!
I think this should work flawlessly, but we will know best by just let it be tested by the users.
I will have a close eye on the issues now to see if i'm right, and fix shit if i'm not!
Let's see

Dickby:
Might be the multi sampling that was basicly different for all corners before.

PR #751, opened 2022-09-30
Some questions about pixman damage tracking (could optimize a bit right away).

Dickby:
I'm not familiar with pixman, 
since you know better probably, can you take a look at this

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -320,6 +320,10 @@ void CHyprOpenGLImpl::renderRectWithDamage(wlr_box* box, const CColor& col, pixm
     RASSERT((box->width > 0 && box->height > 0), "Tried to render rect with width/height < 0!");
     RASSERT(m_RenderData.pMonitor, "Tried to render rect without begin()!");
 
+    // TODO: can we early out here?
+    // if (!pixman_region32_not_empty(damage))
+    // 	return;
+
     float matrix[9];
     wlr_matrix_project_box(matrix, box, wlr_output_transform_invert(!m_bEndFrame ? WL_OUTPUT_TRANSFORM_NORMAL : m_RenderData.pMonitor->transform), 0, m_RenderData.pMonitor->output->transform_matrix);  // TODO: write own, don't use WLR here
 
@@ -356,6 +360,7 @@ void CHyprOpenGLImpl::renderRectWithDamage(wlr_box* box, const CColor& col, pixm
     glEnableVertexAttribArray(m_RenderData.pCurrentMonData->m_shQUAD.posAttrib);
     glEnableVertexAttribArray(m_RenderData.pCurrentMonData->m_shQUAD.texAttrib);
 
+    // TODO: or has it to be that late in the function?
     if (pixman_region32_not_empty(damage)) {
         PIXMAN_DAMAGE_FOREACH(damage) {
             const auto RECT = RECTSARR[i];
@@ -455,7 +460,8 @@ void CHyprOpenGLImpl::renderTextureInternalWithDamage(const CTexture& tex, wlr_b
         glUniform1i(shader->applyTint, 0);
     }
 
-    glVertexAttribPointer(shader->posAttrib, 2, GL_FLOAT, GL_FALSE, 0, fullVerts);
+    // TODO: ???
+    // glVertexAttribPointer(shader->posAttrib, 2, GL_FLOAT, GL_FALSE, 0, fullVerts);
 
     const float verts[] = {
         m_RenderData.primarySurfaceUVBottomRight.x, m_RenderData.primarySurfaceUVTopLeft.y,      // top right
@@ -677,6 +683,13 @@ void CHyprOpenGLImpl::preWindowPass() {
 void CHyprOpenGLImpl::renderTextureWithBlur(const CTexture& tex, wlr_box* pBox, float a, wlr_surface* pSurface, int round) {
     RASSERT(m_RenderData.pMonitor, "Tried to render texture with blur without begin()!");
 
+    // TODO: while those static const pointers to the data values look kind of clever,
+    //       it is an indirection for every config value every time for everything.
+    //       why not have it be values send to these functions and just read them?
+    //       And btw these auto keyword thing might make it easier to write, but
+    //       If you want people to contribute, it makes it more hard to read and understand for them!
+    //       I myself would use auto only if there is a real good reason for that (e.g. pointer to function that returns an array of
+    //       function pointers or something like that, or maybe unnamed structs in C)
     static auto *const PBLURENABLED = &g_pConfigManager->getConfigValuePtr("decoration:blur")->intValue;
     static auto *const PNOBLUROVERSIZED = &g_pConfigManager->getConfigValuePtr("decoration:no_blur_on_oversized")->intValue;
     static auto *const PBLURNEWOPTIMIZE = &g_pConfigManager->getConfigValuePtr("decoration:blur_new_optimizations")->intValue;
@@ -690,6 +703,10 @@ void CHyprOpenGLImpl::renderTextureWithBlur(const CTexture& tex, wlr_box* pBox, 
     pixman_region32_t damage;
     pixman_region32_init(&damage);
     pixman_region32_intersect_rect(&damage, m_RenderData.pDamage, pBox->x, pBox->y, pBox->width, pBox->height);  // clip it to the box
+
+    // TODO(Dickby): can we early out here?
+    // if(!pixman_region32_not_empty(&damage))
+    // 	return;
 
     // amazing hack: the surface has an opaque region!
     pixman_region32_t inverseOpaque;
@@ -741,6 +758,7 @@ void CHyprOpenGLImpl::renderTextureWithBlur(const CTexture& tex, wlr_box* pBox, 
 
     // stencil done. Render everything.
     wlr_box MONITORBOX = {0, 0, m_RenderData.pMonitor->vecTransformedSize.x, m_RenderData.pMonitor->vecTransformedSize.y};
+    // TODO: or does it have to be here?
     if (pixman_region32_not_empty(&damage)) {
         // render our great blurred FB
         static auto *const PBLURIGNOREOPACITY = &g_pConfigManager->getConfigValuePtr("decoration:blur_ignore_opacity")->intValue;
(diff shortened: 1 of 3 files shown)


Dickby:
And can't it be checked once for every window, before you call any of those functions?

Dickby:
I have to work on another project for a while now.
Please take a look at this.

src/render/decorations/CHyprDropShadowDecoration.cpp
```diff
@@ -69,6 +69,8 @@ void CHyprDropShadowDecoration::draw(CMonitor* pMonitor, float a) {
         return; // disabled
 
     // get the real offset
+    // TODO: please tell me you didn't meant to do this every frame (in a try catch block???)
```
vaxerski, line 72: classic example of me saying "I'll fix this later".

src/render/OpenGL.cpp
```diff
@@ -320,6 +320,10 @@ void CHyprOpenGLImpl::renderRectWithDamage(wlr_box* box, const CColor& col, pixm
     RASSERT((box->width > 0 && box->height > 0), "Tried to render rect with width/height < 0!");
     RASSERT(m_RenderData.pMonitor, "Tried to render rect without begin()!");
 
+    // TODO: can we early out here?
```
vaxerski, line 323: yeah, but the damage won't be empty since we're rendering a frame.

src/render/OpenGL.cpp
```diff
@@ -455,7 +460,8 @@ void CHyprOpenGLImpl::renderTextureInternalWithDamage(const CTexture& tex, wlr_b
         glUniform1i(shader->applyTint, 0);
     }
 
-    glVertexAttribPointer(shader->posAttrib, 2, GL_FLOAT, GL_FALSE, 0, fullVerts);
+    // TODO: ???
+    // glVertexAttribPointer(shader->posAttrib, 2, GL_FLOAT, GL_FALSE, 0, fullVerts);
```
vaxerski, line 464: dunno. That's how others do it.

src/render/decorations/CHyprDropShadowDecoration.cpp
Dickby: It's easy to fix!
I'll leave that to you.

src/render/OpenGL.cpp
```diff
@@ -691,6 +704,10 @@ void CHyprOpenGLImpl::renderTextureWithBlur(const CTexture& tex, wlr_box* pBox,
     pixman_region32_init(&damage);
     pixman_region32_intersect_rect(&damage, m_RenderData.pDamage, pBox->x, pBox->y, pBox->width, pBox->height);  // clip it to the box
 
+    // TODO(Dickby): can we early out here?
```
vaxerski, line 707: yeah, good idea

src/render/OpenGL.cpp
```diff
@@ -777,6 +795,7 @@ void CHyprOpenGLImpl::renderBorder(wlr_box* box, const CColor& col, int round) {
     static auto *const PBORDERSIZE = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
     static auto *const PMULTISAMPLE = &g_pConfigManager->getConfigValuePtr("decoration:multisample_edges")->intValue;
 
+    // TODO: pixman early out here?
```
vaxerski, line 798: same as above, damage should never be empty here

src/render/decorations/CHyprDropShadowDecoration.cpp
vaxerski: ik sure sec

src/render/OpenGL.cpp
Dickby: Just look 10 lines further below there is an if-else block that sets it anyways

src/render/OpenGL.cpp
Dickby: will it never be empty?
And if so, why are you checking it later in the function?

src/render/OpenGL.cpp
```diff
@@ -677,6 +683,13 @@ void CHyprOpenGLImpl::preWindowPass() {
 void CHyprOpenGLImpl::renderTextureWithBlur(const CTexture& tex, wlr_box* pBox, float a, wlr_surface* pSurface, int round) {
     RASSERT(m_RenderData.pMonitor, "Tried to render texture with blur without begin()!");
 
+    // TODO: while those static const pointers to the data values look kind of clever,
+    //       it is an indirection for every config value every time for everything.
+    //       why not have it be values send to these functions and just read them?
+    //       And btw these auto keyword thing might make it easier to write, but
+    //       If you want people to contribute, it makes it more hard to read and understand for them!
+    //       I myself would use auto only if there is a real good reason for that (e.g. pointer to function that returns an array of
+    //       function pointers or something like that, or maybe unnamed structs in C)
```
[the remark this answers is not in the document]
Dickby: Looks like fucking python or something, i hate that auto thing!
The worst is, there is now the same keyword in C23 standard.
By the way, i would give so much for this project being written in C!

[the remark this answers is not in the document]
vaxerski: I would give so much for this project not being written in C

src/render/decorations/CHyprDropShadowDecoration.cpp
vaxerski: done

src/render/OpenGL.cpp
vaxerski: oh, dunno, maybe a leftover

src/render/OpenGL.cpp
vaxerski: done

Dickby:
I would like someone (could be me or you) to rewrite own (possible slimmer and more suited for the use case)
versions of the STL functions you use!
That could cut the compile time a lot!.
If you would consider a unity build (not that engine the design concept of pound including source files into one single
compilation unit) i'm shure we could make the whole program compile in less than 10 seconds, that would 
developing much faster!

vaxerski:
that is true. But we should focus on one thing at a time.

src/render/Renderer.cpp
```diff
@@ -36,18 +36,15 @@ void renderSurface(struct wlr_surface* surface, int x, int y, void* data) {
...
-        g_pHyprOpenGL->renderTexture(TEXTURE, &windowBox, RDATA->fadeAlpha * RDATA->alpha, rounding, true);
+    // TODO: this is purely cosmetic (the compiler will know to do this).
+    //       I write code like this too: adding stuff trying and end up with something that works,
+    //       but eventually somebody should look over it to make it's kind of readable (so i did!).
+    if (RDATA->surface && surface == RDATA->surface &&
+	!(wlr_surface_is_xwayland_surface(surface) && !wlr_xwayland_surface_from_wlr_surface(surface)->has_alpha && RDATA->fadeAlpha * RDATA->alpha == 255.f) &&
+         RDATA->blur) {
+	g_pHyprOpenGL->renderTextureWithBlur(TEXTURE, &windowBox, RDATA->fadeAlpha * RDATA->alpha, surface, rounding);
+    }else {
+	g_pHyprOpenGL->renderTexture(TEXTURE, &windowBox, RDATA->fadeAlpha * RDATA->alpha, rounding, true);
```
Dickby, line 47: New

src/render/decorations/CHyprDropShadowDecoration.cpp
```diff
@@ -89,7 +91,13 @@ void CHyprDropShadowDecoration::draw(CMonitor* pMonitor, float a) {
...
 
     // draw the shadow
-    wlr_box fullBox = {m_vLastWindowPos.x - m_seExtents.topLeft.x + 2, m_vLastWindowPos.y - m_seExtents.topLeft.y + 2, m_vLastWindowSize.x + m_seExtents.topLeft.x + m_seExtents.bottomRight.x - 4, m_vLastWindowSize.y + m_seExtents.topLeft.y + m_seExtents.bottomRight.y - 4};
+    // TODO: would it be ok if i break long lines like this, if i come across them?
+    //       It's so much more readable, and not everybody is working on a super wide screeen! PLEASE!
+    wlr_box fullBox = {
+	m_vLastWindowPos.x - m_seExtents.topLeft.x + 2,
+	m_vLastWindowPos.y - m_seExtents.topLeft.y + 2,
+	m_vLastWindowSize.x + m_seExtents.topLeft.x + m_seExtents.bottomRight.x - 4,
+	m_vLastWindowSize.y + m_seExtents.topLeft.y + m_seExtents.bottomRight.y - 4};
```
Dickby, line 101: I'm beggin you here

src/render/decorations/CHyprDropShadowDecoration.cpp
```diff
@@ -89,7 +91,13 @@ void CHyprDropShadowDecoration::draw(CMonitor* pMonitor, float a) {
     m_seExtents = {{*PSHADOWSIZE + 2 - offset.x, *PSHADOWSIZE + 2 - offset.y}, {*PSHADOWSIZE + 2 + offset.x, *PSHADOWSIZE + 2 + offset.y}};
 
     // draw the shadow
-    wlr_box fullBox = {m_vLastWindowPos.x - m_seExtents.topLeft.x + 2, m_vLastWindowPos.y - m_seExtents.topLeft.y + 2, m_vLastWindowSize.x + m_seExtents.topLeft.x + m_seExtents.bottomRight.x - 4, m_vLastWindowSize.y + m_seExtents.topLeft.y + m_seExtents.bottomRight.y - 4};
+    // TODO: would it be ok if i break long lines like this, if i come across them?
+    //       It's so much more readable, and not everybody is working on a super wide screeen! PLEASE!
```
vaxerski, line 95: I don't use an ultrawide but am used to java's `FactoryFactoryFactoryFactoryFactoryFactoryFactoryFactoryFactory`

It's ugh. I don't like it a lot, but it is more readable, yeah. Feel free, I guess.

src/render/decorations/CHyprDropShadowDecoration.cpp
vaxerski: commented below

src/render/Renderer.cpp
vaxerski: see below

src/render/decorations/CHyprDropShadowDecoration.cpp
Dickby: Love you for that!!!

src/render/Renderer.cpp
Dickby: Not shure if you noticed but this is not just line breaking.
It was a nested if else thing that, in the end, just did to different things.

Dickby:
@vaxerski could you maybe cherry-pick the part of that "rework rounding shader" 
that moves the `#include "Shaders.hpp"` from OpenGL.hpp to OpenGL.cpp, for now.
It's such a pain to work on Shaders!
By now, all the time i do something new i have to do it manually!

vaxerski:
done

Dickby:
How do you things like that so fast?
Freaking young generation coders!
How old are you btw.

vaxerski:
19 and I have never seen a woman

PR #756, opened 2022-09-30
Improve the control of shadow_render_power & make it less confusing.

Dickby:
I changed the variable `shadow_render_power` to be a float.
further i changed how that works internally.
Now you have finegrained control over the power of the shadow:
At 0.0 no shadow will be visible (but it should not be confused with turning shadows off,
that should still be done by setting the `drop_shadow`)
I think this is less confusing to users who just think its the power of the shadow, and now nothing about `distance^x`.
until a value of 1.0 there is a smooth transition from full to no shadow within the range.
values more than 1.0 will result in a shadow that has it's full opacity for some way, and than starts to fade to 0.0 within the range.

If you really prefer the way it worked before (higher power means stronger fading) i 
can change it to that with the same fine grained control as this one has.
But as mentioned above i think this is more intuitive.

@vaxerski i think it would be cool to have the shadow power be an animated variable.
Than if we later add individual window rules (right now it's just the color or am i wrong here?)
it could make a nice animation when switching windows.

It's ready to be tested by @fufexan (seems to be the tester here!?).

### DIFF
--- a/src/render/shaders/Shadow.hpp
+++ b/src/render/shaders/Shadow.hpp
@@ -14,26 +14,26 @@ uniform float radius;
 uniform float range;
 uniform float shadowPower;
 
-float pixAlphaRoundedDistance(float distanceToCorner) {
-     if (distanceToCorner > radius) {
-        return 0.0;
-    }
+float pixAlphaRoundedDistance(float distanceToRadiusCenter) {
+    float distanceToCorner = distanceToRadiusCenter - radius;
 
-    if (distanceToCorner > radius - range) {
-        return pow((range - (distanceToCorner - radius + range)) / range, shadowPower); // i think?
-    }
+    if (distanceToCorner > range)
+        return 0.0;
+
+    if (distanceToCorner > 0.0)
+        return 1.0 - pow(distanceToCorner / range, shadowPower);
 
     return 1.0;
 }
 
 void main() {
 
-	vec4 pixColor = v_color;
+    vec4 pixColor = v_color;
     float originalAlpha = pixColor[3];
 
     bool done = false;
 
-	vec2 pixCoord = fullSize * v_texcoord;
+    vec2 pixCoord = fullSize * v_texcoord;
 
     // ok, now we check the distance to a border.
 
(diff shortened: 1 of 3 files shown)


src/render/shaders/Shadow.hpp
```diff
@@ -66,17 +66,17 @@ void main() {
...
 
-        // get the smallest
-        float smallest = min(min(distanceT, distanceB), min(distanceL, distanceR));
+        // get the distance to the closest edge
+        float distanceToEdge = range - min(min(distanceT, distanceB), min(distanceL, distanceR));
 
-        if (smallest < range) {
-            pixColor[3] = pixColor[3] * pow((smallest / range), shadowPower);
+        if (distanceToEdge > 0.0) {
+            pixColor[3] *= 1.0 - pow(distanceToEdge / range, shadowPower);
```
vaxerski, line 73: isn't a floating power WAY slower than an integer one?

Dickby: That's fine, a graphics card is expert in FLOPs. But technically it could be a tad slower.

Dickby: I'm not shure to be honest.

vaxerski: a non-integer power requires roots, which are notoriously slow.

Dickby: But test the thing, it works beautifully and you have super nice control over the shadow.

Dickby: How about the distance function that you use everywhere?
Honestly hardware nowadays is quite good at roots.

vaxerski: you're right, maybe I am a bit too paranoid here. I am stuck in the intel 8086 days.

vaxerski:
I'll wait for the drunkard @fufexan to approve this too, in the meantime, you can make a PR to the wiki repo to change the description of the option

src/render/shaders/Shadow.hpp
Dickby: But distance is a squareroot too.
I can optimize those away though, because you don't want the distance but just a distance comparison.
Didn't think about that, i will do that later.
But this should be fine, especially beccause its just done for a small part of the whole texture.

vaxerski: You should focus on that alleged 10x performance boost, as I still cannot believe you it would make a difference over 10% :P

vaxerski: and I should do some profiling to figure out those constant CPU usage issues

Dickby:
I was interested after your question so i looked it up and found this

> once a hardcoded integer pow() would take more than 3 multiplies (ie: pow(x, 10)), the shader gets compiled to the exp(x * (log)) version, so it would seem modern shader compilers assume that's just as fast or faster than 4 multiplies.

vaxerski:
thats cool, yeah so we'll just wait for mr fufexan to approve this

Dickby:
@fufexan No need to apologize, thanks for testing!
@vaxerski i will try the performance of this too!
will notify you when i know exactly if it's worth it!

Dickby:
I think this one is fine!

vaxerski:
The only thing I don't like about this PR is the fact that it introduces a silent breaking change. Silent, because old settings will work and not throw an error, but will function differently.

Any ideas on how to not make this breaking?

Dickby:
You have a point there.
I will think about it, thanks.

PR #773, opened 2022-10-04
focusedmon event check change

### LINKED ISSUE #758: movefocus keybind across monitors doesn't create a focusedmon event
image
in the attached image chromium to the right is on a different monitor to alacritty on the left listening to the socket
movefocus was used twice, to the left and then back to the right, where alacritty was the original focus.

the expected outcome is that a focusedmon event would also be made as a new monitor is being focused when the movefocus command is used to move focus to a window in another workspace on another monitor

hyprctl version output
`Hyprland, built from branch HEAD at commit e6c9e3f81dda723e289fa31883b326c086979c7ddirty (add case for empty strings in isNumber).
flags: (if any)
debug`

lylac-1:
Compare PLASTWINDOW & PWINDOWTOCHANGETO m_iMonitorID's instead of PWINDOWTOCHANGETO->m_iMonitorID & g_pCompositor->m_pLastMonitor->ID

fixes #758, compairing `PWINDOWTOCHANGETO->m_iMonitorID` to `g_pCompositor->m_pLastMonitor->ID` misses conditions

I'm largely unfamiliar with the code and am both unsure this is reliable and unsure it is good practice or fit

yes, its a single line change with as far as i can tell little to no effect on anything else in the project

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -988,7 +988,7 @@ void CKeybindManager::moveFocusTo(std::string args) {
             Vector2D middle = PWINDOWTOCHANGETO->m_vRealPosition.goalv() + PWINDOWTOCHANGETO->m_vRealSize.goalv() / 2.f;
             g_pCompositor->warpCursorTo(middle);
 
-            if (PWINDOWTOCHANGETO->m_iMonitorID != g_pCompositor->m_pLastMonitor->ID) {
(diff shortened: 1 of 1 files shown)


lylac-1:
yeah
the only things i tested for was it creating the event when I expected and checked some situations i thought it might happen but shouldnt (change between windows on the same monitor and change between windows over different workspaces on the same monitor) and found no issues

Issue #778, opened 2022-10-05
left_handed has no effect


mrshmllow:
My mouse has changed to being left handed i _believe_ since https://github.com/hyprwm/Hyprland/commit/60c414ccadd0e2ff6be22bb880f605a164f451d5 & #776, and setting `left_handed = false` has no effect even after restarting

mrshmllow:
Just ran a bisect- 
with `input:left_handed = false`, 60c414c is the first bad commit

EvilBoi123:
yeah i got this problem too, now, by default my mouse were inverted rmb and lmb

PR #781, opened 2022-10-05
Fix getDeviceInt string arg "input:left_handed"

Dickby:
Fixes https://github.com/hyprwm/Hyprland/issues/778
I guess it was intended this way?


mrshmllow:
Can confirm this patch fixes #778 and respects left_handed properly

PR #793, opened 2022-10-06
Fix compiler warnings.

Dickby:
Don't you get those warnings?
I waited for a fix, but nobody seems to see it / feel resposible.


vaxerski:
I did, just said to myself "I'll fix this after x" and never did it

PR #794, opened 2022-10-06
Check earlier if pixman_region is empty in some places.

Dickby:
Checks earlier if pixman_region is empty in some places in OpenGL.cpp

I checked, and all of those early outs get used at times.

I didn't find anything broken by this pr,
but it should be checked before merging!

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -310,7 +310,8 @@ void CHyprOpenGLImpl::scissor(const int x, const int y, const int w, const int h
 }
 
 void CHyprOpenGLImpl::renderRect(wlr_box* box, const CColor& col, int round) {
-    renderRectWithDamage(box, col, m_RenderData.pDamage, round);
+    if(pixman_region32_not_empty(m_RenderData.pDamage))
+	renderRectWithDamage(box, col, m_RenderData.pDamage, round);
 }
 
 void CHyprOpenGLImpl::renderRectWithDamage(wlr_box* box, const CColor& col, pixman_region32_t* damage, int round) {
@@ -364,15 +365,13 @@ void CHyprOpenGLImpl::renderRectWithDamage(wlr_box* box, const CColor& col, pixm
             }
         }
 
-        pixman_region32_fini(&damageClip);
+	pixman_region32_fini(&damageClip);
     } else {
-        if (pixman_region32_not_empty(damage)) {
-            PIXMAN_DAMAGE_FOREACH(damage) {
-                const auto RECT = RECTSARR[i];
-                scissor(&RECT);
-                glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
-            }
-        }
+	PIXMAN_DAMAGE_FOREACH(damage) {
+	    const auto RECT = RECTSARR[i];
+	    scissor(&RECT);
+	    glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
+	}
     }
 
     glDisableVertexAttribArray(m_RenderData.pCurrentMonData->m_shQUAD.posAttrib);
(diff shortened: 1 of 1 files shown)


src/render/OpenGL.cpp
```diff
@@ -467,13 +470,13 @@ void CHyprOpenGLImpl::renderTextureInternalWithDamage(const CTexture& tex, wlr_b
     glEnableVertexAttribArray(shader->posAttrib);
     glEnableVertexAttribArray(shader->texAttrib);
 
-    if (pixman_region32_not_empty(m_RenderData.pDamage)) {
+    /* if (pixman_region32_not_empty(m_RenderData.pDamage)) { */
```
vaxerski, line 473: why comment? delete.

Dickby: I missed that!

vaxerski:
needs a conflicts fix and lgtm

Dickby:
I'm i right that damageclip is empty if the damage region is empty?
It's just used during workspace switching, right!?
I tested it after the change, and it seems to work!

vaxerski:
> Am I right that damageclip is empty if the damage region is empty?

Yes, damageclip is an intersection of the damage and the clip region.

src/render/OpenGL.cpp
```diff
@@ -310,7 +310,8 @@ void CHyprOpenGLImpl::scissor(const int x, const int y, const int w, const int h
 }
 
 void CHyprOpenGLImpl::renderRect(wlr_box* box, const CColor& col, int round) {
-    renderRectWithDamage(box, col, m_RenderData.pDamage, round);
+    if(pixman_region32_not_empty(m_RenderData.pDamage))
+      renderRectWithDamage(box, col, m_RenderData.pDamage, round);
```
vaxerski, line 314: odd indent?

src/render/OpenGL.cpp
```diff
@@ -802,6 +803,9 @@ void CHyprOpenGLImpl::renderBorder(wlr_box* box, const CColor& col, int round) {
     RASSERT((box->width > 0 && box->height > 0), "Tried to render rect with width/height < 0!");
     RASSERT(m_RenderData.pMonitor, "Tried to render rect without begin()!");
 
+    if (!pixman_region32_not_empty(m_RenderData.pDamage))
+      return;
```
vaxerski, line 807: odd indent?

src/render/OpenGL.cpp
```diff
@@ -1083,6 +1085,9 @@ void CHyprOpenGLImpl::renderRoundedShadow(wlr_box* box, int round, int range, fl
     RASSERT((box->width > 0 && box->height > 0), "Tried to render shadow with width/height < 0!");
     RASSERT(m_pCurrentWindow, "Tried to render shadow without a window!");
 
+    if (!pixman_region32_not_empty(m_RenderData.pDamage))
+      return;
```
vaxerski, line 1089: odd indent?

Dickby:
Sorry!
Btw the clipping thing does nothing at all!
just check offset.x /offset.y : it will never be anything but 0.
Should i fix that or are you giving it another try?

vaxerski:
clipping will only be active when .width and .height of the clipBox are set. and it works. dw.

Dickby:
When is that set?

vaxerski:
> When is that set?

Used to be, aint anymore. But might come handy in the future

Dickby:
The code block inside renderWindow that sets `offset` is totally weird, and it never sets an offset.
I had the feeling it should compare BBWIN with PWORKSPACE->m_vRenderOffset.vec and not with PWSMON->vecPosition.
The whole PROGRESS thing doesn't make sense at all to me,  but might be just me!

Dickby:
Another thing, maybe i should make it an issue, but whatever:
I think it would help if we had some more graphical debug functions for:
- rendering a rectanble of lines (to check if boxes (damage ...) are right
- rendering lines points for checking positions.
- scaling the whole screen down ,not like monitor scaling, but actually making the whole thing smaller, so that
  the indicators above, can be checked outside of the monitors dimensions.

vaxerski:
> The whole PROGRESS thing doesn't make sense at all to me, but might be just me!

dw. it works :)

> rendering a rectanble of lines (to check if boxes (damage ...) are right

umm what

> rendering lines points for checking positions.

eh maybe

Dickby:
I set breakpoints on lines : 259,261,269 and 271 of Renderer.cpp and none of them ever hit.
when changing Workspaces.
PROGRESS is the fraction of the screen that has has shifted by workspace switching.
PWSMON->vecPosition.x doesn't change during workspace switching,  WINBB.x doesn't change during workspace switching.
But the difference of those two is multiplied by PROGRESS, this is so weird.

vaxerski:
move a floating window to be partially off screen and use workspace swipe or slide workspace anim

Dickby:
So it's just for floating windows?
Why shouldn't that work for tiled once during ws switching?

Dickby:
And if the floating window is partially of screen, why than only clip if the workspace is sliding?

vaxerski:
to prevent it from "popping" out when the animation finishes.

Dickby:
Ok i misunderstood the functionality then, thanks for clarifying!
How about this pr?

Issue #803, opened 2022-10-07
monitor transform 90 and 270


umop3plsdn:
In the transition from 0.15.1 from 0.15.0 my third monitor which is rotated 90 degrees did this strange skew thing where some of the outline of the window would shade over the top also the corner would completely disapear but you could still see the cursor.  The way the cursor reacts to the rotation is absolutely correct with all the borders but it seems to be just a visual issue. I tested with my other monitors with 90 and 270 and it did the same thing
monitor=DP-1,transform,3

![image](https://user-images.githubusercontent.com/696094/194441796-86c12a78-d3d4-4fa4-8412-36887654918c.jpeg)

![2022-10-06-20:21:45](https://user-images.githubusercontent.com/696094/194441665-5a5af8db-7d60-4529-8210-f709a7836bde.png)

vaxerski:
seems like a shader issue. @Dickby can you take a look at this? Probably a bisect to narrow it down to the faulty commit.

vaxerski:
yep @Dickby your fault, the rounding shader rework 7bdfdaa28a4d107b1a9c9260837f81001000e697

vaxerski:
My guess is you need to transform the box in the `renderTextureInternalWithDamage` func (where you pass pos & size to the shader) but I can't seem to get the balance right.

Dickby:
I will look at that!

KyrillGobber:
Same problem here, thanks for looking into it :+1:

umop3plsdn:
nice works thanks!

Dickby:
@umop3plsdn 
next thing to transform would be your profile picture!

PR #807, opened 2022-10-07
Transform the box data send to texture shaders.

Dickby:
Fixes https://github.com/hyprwm/Hyprland/issues/803

This is a bit ugly:
Because the Texture shaders are working with absolute screen coordinates now, the pBox has to be transformed according to the monitor transform.
for the cases 90 and 270 (without flip) those transformes need to be swapped.
I will try to find a cleaner solution to this, but this one seems to work.

It should be tested on other monitors, but in my case it works!


vaxerski:
invert the transform maybe

Dickby:
@vaxerski 
Much cleaner!
Thanks!

PR #812, opened 2022-10-08
Add debug coredump instructions

KyleUltimate:
Add how to obtain the debug coredump.

Whether should this coredump also be recommended when reporting bugs

100% working

### DIFF
--- a/docs/ISSUE_GUIDELINES.md
+++ b/docs/ISSUE_GUIDELINES.md
@@ -61,6 +61,7 @@ where `[PID]` is the PID you remembered.
 In very rare cases, the normal coredump would not be enough.
 Only then, you could try obtaining the debug coredump.
 1. [Compile Hyprland with debug mode](http://wiki.hyprland.org/Contributing-and-Debugging/#build-in-debug-mode)
+1. Reproduce the crash in debug mode.
 2. `env DEBUGINFOD_URLS="https://debuginfod.archlinux.org/" coredumpctl debug [PID]`(see above)
 3. Wait until the `gdb` appears`, then do:
 4. `set pagination off`
--- a/docs/ISSUE_GUIDELINES.md
+++ b/docs/ISSUE_GUIDELINES.md
@@ -58,6 +58,6 @@ coredumpctl info [PID]
 where `[PID]` is the PID you remembered.
 
 ## Obtaining the debug Hyprland coredump
-1. Compile Hyprland with debug flags turn on
+1. [Compile Hyprland with debug mode](http://wiki.hyprland.org/Contributing-and-Debugging/#build-in-debug-mode)
 2. `coredumpctl debug [PID]`(see above)
 3. Wait until the `gdb` appears, then do `bt -full`
--- a/docs/ISSUE_GUIDELINES.md
+++ b/docs/ISSUE_GUIDELINES.md
@@ -57,3 +57,7 @@ coredumpctl info [PID]
 ```
 where `[PID]` is the PID you remembered.
 
+## Obtaining the debug Hyprland coredump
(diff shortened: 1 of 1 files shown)


vaxerski:
why do we need 2 methods? the first is simpler and in most cases enough

KyleUltimate:
Yeah, most of the time the first method would be enough.
This pr is just a "just in case you need the debug dump* and you can just point the user to this page instead of teaching them how to do it

vaxerski:
make it clearer then that it's "just in case"

Additionally, since most reporters are completely not into debugging, please make additional comments of continuing without paging and enabling debuginfod.

KyleUltimate:
Do you mean this `set pagination off`?

vaxerski:
no, gdb sometimes asks you to "press y to continue without paging"

docs/ISSUE_GUIDELINES.md
```diff
@@ -57,3 +57,12 @@ coredumpctl info [PID]
...
 where `[PID]` is the PID you remembered.
 
+## Obtaining the debug Hyprland coredump
+In very rare cases, the normal coredump would not be enough.
+Only then, you could try obtaining the debug coredump.
+1. [Compile Hyprland with debug mode](http://wiki.hyprland.org/Contributing-and-Debugging/#build-in-debug-mode)
+1. Reproduce the crash in debug mode.
+2. `env DEBUGINFOD_URLS="https://debuginfod.archlinux.org/" coredumpctl debug [PID]`(see above)
+3. Wait until the `gdb` appears
+4. If gdb asks you `press y to continue without paging?` say no.
```
vaxerski, line 67: what??? why?? Just continue without paging.

docs/ISSUE_GUIDELINES.md
```diff
@@ -57,3 +57,12 @@ coredumpctl info [PID]
 ```
 where `[PID]` is the PID you remembered.
 
+## Obtaining the debug Hyprland coredump
+In very rare cases, the normal coredump would not be enough.
+Only then, you could try obtaining the debug coredump.
+1. [Compile Hyprland with debug mode](http://wiki.hyprland.org/Contributing-and-Debugging/#build-in-debug-mode)
```
vaxerski, line 63: please mention that debug hyprland uses `hyprlandd.conf`

docs/ISSUE_GUIDELINES.md
```diff
@@ -57,3 +57,12 @@ coredumpctl info [PID]
...
+1. Reproduce the crash in debug mode.
+2. `env DEBUGINFOD_URLS="https://debuginfod.archlinux.org/" coredumpctl debug [PID]`(see above)
+3. Wait until the `gdb` appears
```
vaxerski, line 66: `(gdb)`

docs/ISSUE_GUIDELINES.md
KyleUltimate: Oh, wait I'm sorry
A brain fart I suppose

docs/ISSUE_GUIDELINES.md
```diff
@@ -57,3 +57,14 @@ coredumpctl info [PID]
 ```
 where `[PID]` is the PID you remembered.
 
+## Obtaining the debug Hyprland coredump
+In very rare cases, the normal coredump would not be enough.
+Only then, you could try obtaining the debug coredump.
+1. [Compile Hyprland with debug mode](http://wiki.hyprland.org/Contributing-and-Debugging/#build-in-debug-mode)
+> Note: the config file will be in `$HOME/.config/hypr/hyprlandd.conf`
```
vaxerski, line 64: *the config file used will be `hyprlandd.conf` instead of `hyprland.conf`*

PR #823, opened 2022-10-09
Simplify matrix calculations

Dickby:
Removes unnecessary matrix calculations. 

No

I think it's ready!

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -178,7 +178,7 @@ bool isNumber(const std::string& str, bool allowfloat) {
     std::string copy = str;
     if (*copy.begin() == '-')
         copy = copy.substr(1);
-    
+
     if (copy.empty())
         return false;
 
@@ -331,7 +331,7 @@ void logSystemInfo() {
     Debug::log(LOG, "Node name: %s", unameInfo.nodename);
     Debug::log(LOG, "Release: %s", unameInfo.release);
     Debug::log(LOG, "Version: %s", unameInfo.version);
-    
+
     Debug::log(NONE, "\n");
 
     const std::string GPUINFO = execAndGet("lspci -vnn | grep VGA");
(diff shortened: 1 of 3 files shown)


vaxerski:
what the fuck

I know absolutely jack shit nothing about matrices here but that was how wlroots' renderer used to do it.

Are you mega sure this is correct? Have you also tested transformed displays?

Dickby:
It's not changing anything really, just taking away the two minuses in matrixProjection infront of t[3] and t[4] ,which are the values that effect the y value of the vectors multiplicated with that matrix. look at the transforms[], 
in every transform the value pairs for 3,4 and 0,1 are :
one is zero and the other one is 1 or -1;
that basicly switches x and y for the resulting vector and multiplicates -1 for some of those transforms.
If you thing about e.g. switching x and y of a vector and multiplying x with -1 would be a rotation of 90 degrees counter clockwise. this is how rotations are done here(it's basicly the mechanism of imaginary numbers).
by multiplying t[3] and t[4] with -1 like you did, all it does is it changes the sign of the resulting y coordinate of the vectors.
With that you would end up with a image that is flipped180, to deal with that you had to multiply that matrix with your flipped180 matrix (which if you look at it does nothing but flipping the sign of the y value of the resulting vector.
I took away those sign flips in your matrixProjection, so the flipped180 is not needed anymore.

The second thing i did was just taking away the wlr_matrix_transpose(matrix,matrix) which "mirrors the matrix across the diagonal, that is because the matrices in the code are ordered the other way around as inside opengl's mechanism.
This is so common, that opengl's glUniformMatrix functions have an boolean value that controls if opengl should to that to the matrix itself (in this case it's basicly for free because it just effects how opengl copies those values).
I tested it with scaling and all 8 available transpositions.

Dickby:
If your unsure, 
@fufexan would you be so kind and try to break this!

fufexan:
I've been running it for half an hour, been playing games. No issues found.

Dickby:
@fufexan thanks.

PR #825, opened 2022-10-10
Replace $(nproc) with $(shell nproc).

Dickby:
At least on my end, just $(nproc) doesn't set the number.

I don't really know why this is, check if it's the same on your setup.

If it has no different behavior elsewhere, it can be merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,19 +93,19 @@ wlr-output-power-management-unstable-v1-protocol.o: wlr-output-power-management-
 
 legacyrenderer:
 	mkdir -p build && cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DLEGACY_RENDERER:STRING=true -H./ -B./build -G Ninja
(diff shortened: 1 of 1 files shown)


vaxerski:
`command not found: shell`

you mean `/bin/sh`?

Dickby:
Hmm, this works for me!
With /bin/sh it doesn't.
I don't have time now, but i'll research later what the issue is.

Dickby:
But is it working for you as it is right now?

Dickby:
Do you get that error when compiling, or did you just test it in a terminal?
In a terminal i get the same error.

vaxerski:
term, but I wouldnt risk it

Dickby:
A small search later, i know that it's the way to use shell command in makefiles:
https://stackoverflow.com/questions/10024279/how-to-use-shell-commands-in-makefile

Dickby:
Or more official:
https://www.gnu.org/software/make/manual/html_node/Shell-Function.html

vaxerski:
alr then

Issue #726, opened 2022-09-25
xwayland windows not opening


patrickaldis:
Currently, I cannot open any xwayland windows, with each one giving some variation of `cannot open display` in command line

For example, when running **spotify** from command line:
Command Output:
```
[spotifywm] attached to spotify
/app/extra/share/spotify/spotify: /app/lib/libcurl-gnutls.so.4: no version information available (required by /app/extra/share/spotify/spotify)
/app/extra/share/spotify/spotify: /app/lib/libcurl-gnutls.so.4: no version information available (required by /app/extra/share/spotify/spotify)
[spotifywm] attached to spotify
[spotifywm] attached to spotify

(spotify --force-device-scale-factor=2:6): Gtk-WARNING **: 10:26:05.703: cannot open display: :99.0
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
```
Hyprland logs:
```
[LOG] Window 262c730 set title to flatpak run com.spot ~
[LOG] Window 262c730 set title to ~
[LOG] New XWayland Surface created (class (null)).
[LOG] Registered signal for owner 25fbaa0: 2657d38 -> 25fbb08 (owner: XWayland Window)
[LOG] Registered signal for owner 25fbaa0: 2657d48 -> 25fbb70 (owner: XWayland Window)
[LOG] Registered signal for owner 25fbaa0: 2657cb8 -> 25fbbd8 (owner: XWayland Window)
[LOG] Window 25fbaa0 destroyed, queueing. (class unmanaged X11)
[LOG] XWayland class raw: (null)
[LOG] Callback 25fbb30 -> 25fbb28, XWayland Window removed.
[LOG] Callback 25fbb98 -> 25fbb90, XWayland Window removed.
[LOG] Callback 25fbc00 -> 25fbbf8, XWayland Window removed.
```

Or when running **emacs**:
Command output:
```
(emacs:111040): Gtk-WARNING **: 10:30:55.807: cannot open display: :0
```

Hyprland Logs:
```
[LOG] Window 262c730 set title to emacs ~
[LOG] Window 262c730 set title to ~
```

Estebiu:
Did you try setting this in your config?
 `exec-once=dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
exec-once=systemctl --user import-environment WAYLAND_DISPLAY XDG_CURRENT_DESKTOP`

[...]

PR #850, opened 2022-10-13
Nix modules: fix default environment variables

Narice:
This PR fixes a problem with the environment variables automatically set by the nix module as seen in #726.

The nixos module was tested but the hm-module was not.
However it should not be a problem as the solution should be the same.

This PR is ready to be merged.

@fufexan @viperML

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -92,7 +92,7 @@ in {
       ++ lib.optional cfg.xwayland.enable pkgs.xwayland;
 
     home.sessionVariables = lib.mkIf cfg.recommendedEnvironment {
-      GDK_BACKEND = "wayland";
+      GDK_BACKEND = "wayland,x11";
       _JAVA_AWT_WM_NONREPARENTING = "1";
       NIXOS_OZONE_WL = "1";
       XCURSOR_SIZE = toString config.home.pointerCursor.size or "24";
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -47,7 +47,7 @@ in {
       systemPackages = lib.optional (cfg.package != null) cfg.package;
 
       sessionVariables = mkIf cfg.recommendedEnvironment {
-        GDK_BACKEND = "wayland";
+        GDK_BACKEND = "wayland,x11";
         _JAVA_AWT_WM_NONREPARENTING = "1";
         NIXOS_OZONE_WL = "1";
         XCURSOR_SIZE = "24";


vaxerski:
@fufexan for nix

fufexan:
What if we don't include that variable at all? Would it be an issue?

fufexan:
Alright, thanks for the info. Will merge as is.

Narice:
I'm very late but thanks for merging!

Narice:
Oops, I was wrong, it's actually the XDG_SESSION_TYPE variable, could it be set when launching Hyprland instead of globally?
If that's the case, maybe we could do the same for the other variables so that they are only set when actually using Hyprland.

fufexan:
I'd like them to be set by Hyprland. Creating a wrapper for such common variables is a pain.

Narice:
Yeah, I just checked what we can do and that sounds like a pain to do.
It's probably a good practice anyway for Hyprland to set this variable itself so it wouldn't hurt (hopefully?).
Do you happen to know where the responsability lies normally?
It makes sense that it would be the responsability of the compositor to set this, but I'd just want to be sure before poking around hyprland's code a bit to see what I can do.
Don't expect me to fix it before @vaxerski does though x'), I'll probably be pretty slow on that one.

fufexan:
Afaik Sway sets `XDG_SESSION_TYPE` itself, it was how I discovered it was the missing variable to make my screensharing work.

vaxerski:
WHOMST'D'VE PINGED ME

Narice:
my bad x')
also, found where to put that in main.cpp, yeah!
can actually do a pr pretty fast, so that's what I'll do :yum:
since I'm there, should I do the same for _JAVA_AWT_WM_NONREPARENTING and other variables? (XCURSOR_SIZE and NIXOS_OZONE_WL)

vaxerski:
java - yes. Cursor - no, it's already done. nixos - ask @fufexan

fufexan:
XCURSOR_SIZE is set already afaik. No idea whether it would make sense to pollute users of other distros' environments with NIXOS_OZONE_WL.

Narice:
what's the purpose of the NIXOS_OZONE_WL var?
and else, got it, I'll do that!

Narice:
also, in the nix modules, should I delete the cursor one then while I'm at it?

fufexan:
`NIXOS_OZONE_WL` is checked by electron apps' wrappers in nixpkgs. If it's set, the wrapper passes `--ozone-platform=wayland` and something else so that electron apps can be Wayland native "out of the box".

I'd say it's fine to remove the XWAYLAND_CURSOR from the modules, yeah.

Narice:
from what I saw, vscode is an app that needs this var.
since I still have this variable set, and that I'm still on x11, I tested to see if that had any effect on the app, but it runs fine.
Maybe there's a specific thing that will not work in x11 because of the `NIXOS_OZONE_WL` variable being set, but that's probably fine for now.
I'll make a pull request then, thanks!

Issue #788, opened 2022-10-05
Window Rule for min-size of floating windows


K1llf0rce:
So we currently have the window rule `size [x] [y]` for resizing a floating window on open.
I would really appreciate a rule, let's call it `min-size [x] [y]`, that only resizes a floating window if the spawned window has an initial height and width that is less than `[x] [y]`. 
Same could be done for `max-size` the other way around, although a minimum setting would be enough imo.

vaxerski:
done in 095688712d2a064bb7bfaded3ffc21580e5d338f

PR #857, opened 2022-10-15
add maxsize window rule

K1llf0rce:

Adds the `max-size [x] [y]` window rule, basically the same as #788 just for limiting the max size of a window on spawn. Initially I thought only a min-size rule would be useful but having the option to also set max-size is quite handy.

Not the best at C++, but I think I did the right thing here by just replacing `std::max` (in the min-size rule) with `std::min` for getting the smaller size of the two. Seems to work as expected after testing.

Should be ready for merge :)


vaxerski:
make a wiki PR too plox for this

K1llf0rce:
done, [PR](https://github.com/hyprwm/hyprland-wiki/pull/30)

vaxerski:
thanks

PR #893, opened 2022-10-22
meson.build,CMakeLists.txt: use sh instead of bash

wael444:
the commands ran under bash do not need bash's features, and bash doesn't exist on many systems.

in meson, theres a better way to get such variables.


fufexan:
> in meson, theres a better way to get such variables.

Please elaborate.

wael444:
> No

PR #912, opened 2022-10-25
Improve several hyprctl commands

riley-martin:
This adds better help for some hyprctl subcommands, particularly `dispatch`, `keyword`,
`hyprpaper` and `setcursor`. It also makes them return a nonzero exit status to the shell
when they do not have enough arguments.

No, not that I am aware of.

Works for me (tm), but yes, it should be ready to merge.

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -158,11 +158,12 @@ void requestHyprpaper(std::string arg) {
     std::cout << std::string(buffer);
 }
 
-void dispatchRequest(int argc, char** argv) {
+int dispatchRequest(int argc, char** argv) {
 
     if (argc < 4) {
-        std::cout << "dispatch requires 2 params";
-        return;
+        std::cout << "Usage: hyprctl dispatch <dispatcher> <arg>\n\
+            Execute a hyprland keybind dispatcher with the given argument";
+        return 1;
     }
 
     std::string rq = "/dispatch";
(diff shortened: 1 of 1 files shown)


fufexan (approved):

NotAShelf:
I'm all in for this PR, but why those commands in those particular?

riley-martin:
Mostly because they were ones that took additional arguments, and gave somewhat unclear 
error messages when not enough arguments were provided. `getoption` also takes another argument
but since it is just one, its a little easier to understand IMO; also changing it would be more
complicated. If you think the others should be improved, though, I can work on it.

hyprctl/main.cpp
```diff
@@ -172,28 +175,33 @@ void keywordRequest(int argc, char** argv) {
...
     request(rq);
+    return 0;
 }
 
-void hyprpaperRequest(int argc, char** argv) {
+int hyprpaperRequest(int argc, char** argv) {
     if (argc < 4) {
-        std::cout << "hyprpaper requires 2 params";
-        return;
+        std::cout << "Usage: hyprctl hyprpaper <commang> <arg>\n\
```
vaxerski, line 183: commang

hyprctl/main.cpp
```diff
@@ -158,12 +159,14 @@ void dispatchRequest(int argc, char** argv) {
...
+    return 0;
 }
 
-void keywordRequest(int argc, char** argv) {
+int keywordRequest(int argc, char** argv) {
     if (argc < 4) {
-        std::cout << "keyword requires 2 params";
-        return;
+        std::cout << "Usage: hyprctl keyword <keyword> <args>\n\
+            Execute a hyprland keyword with the given arguments";
```
vaxerski, line 168: arguments here, argument above. Decide

vaxerski:
can you plox squash

PR #924, opened 2022-10-27
meson: Fallback to 'opengl' when 'GL' is not found.

rakino:
This patch adds 'opengl' as a fallback to 'GL' for dependency lookup, to link with libglvnd configured without X11 support.

For OpenGL, libglvnd provides two pkg-config files: `gl.pc` with GLX support while `opengl.pc` not.  When building without X11 support, the former won't be installed.

N

It's ready for merging.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -18,7 +18,7 @@ executable('Hyprland', src,
     xcb_dep,
 
     dependency('pixman-1'),
-    dependency('GL'),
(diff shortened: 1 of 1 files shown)


rakino:
@fufexan

fufexan (approved):
If you've tested it and it builds fine, LGTM. I've built it with no errors.

PR #961, opened 2022-11-06
Dispatcher to center floating window

abbadanor:
I had some trouble using hyprctl to center my window, doing things like `hyprctl dispatch movewindowpixel 50% 50% active`, which did not work. If there is some way to achieve this with the current version, then please show how.

With this PR, it's as easy as `hyprctl dispatch centerwindow active`

If this PR is accepted, I will make a PR in the wiki.

Btw, although the code is working, I don't know C++, so I can't guarantee the stability/quality of my code.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -16,6 +16,7 @@ CKeybindManager::CKeybindManager() {
     m_mDispatchers["pseudo"]                    = toggleActivePseudo;
     m_mDispatchers["movefocus"]                 = moveFocusTo;
     m_mDispatchers["movewindow"]                = moveActiveTo;
+    m_mDispatchers["centerwindow"]              = centerWindow;
     m_mDispatchers["togglegroup"]               = toggleGroup;
     m_mDispatchers["changegroupactive"]         = changeGroupActive;
     m_mDispatchers["togglesplit"]               = toggleSplit;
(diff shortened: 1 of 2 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -579,6 +580,19 @@ void CKeybindManager::toggleActiveFloating(std::string args) {
     g_pLayoutManager->getCurrentLayout()->changeWindowFloatingMode(PWINDOW);
 }
 
+void CKeybindManager::centerWindow(std::string args) {
+    CWindow* PWINDOW = nullptr;
+
+    PWINDOW = g_pCompositor->m_pLastWindow;
```
vaxerski, line 586: why is this two lines? also, PWINDOW can be made const. That's what ALLCAPS means.

src/managers/KeybindManager.cpp
```diff
@@ -579,6 +580,19 @@ void CKeybindManager::toggleActiveFloating(std::string args) {
...
+
+    if (!PWINDOW || !PWINDOW->m_bIsFloating || PWINDOW->m_bIsFullscreen)
+        return;
+
+    const auto PMONITOR = g_pCompositor->getMonitorFromID(PWINDOW->m_iMonitorID);
+
+    PWINDOW->m_vRealPosition = PMONITOR->vecPosition + PMONITOR->vecSize / 2.f - PWINDOW->m_vRealSize.goalv() / 2.f;
```
vaxerski, line 593: for floating windows, also update PWINDOW->m_vPosition

abbadanor: Ok wait a sec

abbadanor (commented):
it now also updates vPosition

PR #971, opened 2022-11-07
meson: unbreak build with BSD find(1)

jbeich:
According to [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html) *path* argument is not optional. Compare [GNU find(1)](https://www.man7.org/linux/man-pages/man1/find.1.html) with [FreeBSD find(1)](https://man.freebsd.org/find/1).

See also [downstream package](https://freshports.org/x11-wm/hyprland-devel).


vaxerski:
@fufexan

vaxerski:
ok wont wait for fuffie he prolly banging some chicks, lgtm thanks

PR #1000, opened 2022-11-12
Set child stdout and stderr to /dev/null

Mango0x45:

Some scripts or utilities check to see if stdout or stderr are connected
to a TTY or not to determine certain behaviors (for example, you might
want to prompt a user for input with fzf(1) if in a terminal but tofi(1)
or wofi(1) when spawned from your WM).  Since hyprland never closes
these output streams for spawned processes, they end up just spewing
their output onto the TTY while giving the user no real way to have a
script detect if it's being run from the shell or WM.

Instead of just closing stdout and stderr though, we close them and then
proceed to reopen them but connect them to /dev/null.  This allows
scripts and processes to not fail when attempting to write, but for that
writing to simply have no effect.

It is ready to merge.  It doesn't hurt for someone else to test this
patch, but I've done tests of my own and it seems to work flawlessly.  A
quick test you could do would be run the following binding:

```
bind = $modkey, T, exec, test-script.sh
```

Where `test-script.sh` is this:

```sh
#!/bin/sh

{
	[ -t 1 ] && echo "stdout is a terminal";
	[ -t 2 ] && echo "stderr is a terminal";
} > /tmp/test-script.log
```

If all goes well, `/tmp/test-script.log` should be an empty file.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1,5 +1,7 @@
 #include "KeybindManager.hpp"
 
+#include <fcntl.h>
+#include <paths.h>
 #include <regex>
 
 CKeybindManager::CKeybindManager() {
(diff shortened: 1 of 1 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -526,6 +526,24 @@ void CKeybindManager::spawn(std::string args) {
...
+            close(STDERR_FILENO);
+
+            int devnull = open(_PATH_DEVNULL, O_WRONLY);
+            if (devnull == -1) {
+                Debug::log(LOG, "Unable to open /dev/null for writing");
+                return;
+            }
+
+            if (dup2(devnull, STDOUT_FILENO) == -1) {
+                Debug::log(LOG, "Unable to duplicale /dev/null to stdout");
```
vaxerski, line 539: typo in duplicate

Mango0x45:
I've fixed the typo(s) and also made sure to close the `devnull` file
descriptor after duplicating it.

src/includes.hpp
```diff
@@ -7,9 +7,11 @@
 #endif
 
 #include <X11/Xlib.h>
+#include <fcntl.h>
```
vaxerski, line 10: no need to add additional includes here - my original mistake of doing that pays now.

Move them to keybindmanager.cpp to shorten compile time.

vaxerski:
also happy 1k issue!

vaxerski:
Had to revert it due to stuff like #1013

nnyyxxxx:
hmmm

PR #1009, opened 2022-11-13
Fix cmake error when compiling with clang

Maks-s:
## Description

CMake build throws an error when compiling using clang :
`error: invalid value 'c++23' in '-std=c++23'`

This PR aims to fix this

#### How to reproduce

```bash
$ export CXX=clang++
$ cmake . build
$ ninja -C build # ERROR
```

## Improvements

It could be improved to support other compilers, however gcc and clang should be enough for now


bazuin-32:
I'm pretty sure you could just do:

```cmake
set(CMAKE_CXX_STANDARD 23)
```
and let cmake decide the appropriate flags for the compiler.
 https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html

vaxerski:
agreed with @bazuin-32

Maks-s:
All done, thanks for the tip

PR #1020, opened 2022-11-14
apply some rules dynamically when state of window changes

ElJeffe:
When toggling the floating state of a window, some rules that change the appearance of a window did not apply anymore, since rules are only matched when a window is created.
This MR will reevaluate the rules when the floating. pinned and fullscreen state of a window changes, and consider these specific rules that affects the window:
noblur, noborder, noshadow, rounding, opacity, bordercolor.

The MR also adds two additional  rule matching fields:
fullscreen - 0/1
pinned - 0/1

example rules:
windowrulev2 = bordercolor rgb(EE4B55) rgb(880808),fullscreen:1
windowrulev2 = bordercolor rgb(32CD32) rgb(7CFC00),floating:1
windowrulev2 = opacity 0.8 0.8,pinned:1

I took a selection of rules that seemed relevant to me to modify based on a window state. If more rules should be added or some removed, this is possible. (opaque, noanim, animation maybe?)

I mainly created this because these rules allow me to have a visual clue if a window is in fullscreen or not.

I tested bordercolor and opacity in combination with fullscreen, pinned and floating and did not see an issue. Since rules are only reevaluated when the state changes, I also do not think there is a noticeable performance impact.

I will update the wiki when approved

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1732,6 +1732,9 @@ void CCompositor::setWindowFullscreen(CWindow* pWindow, bool on, eFullscreenMode
 
     g_pXWaylandManager->setWindowFullscreen(pWindow, pWindow->m_bIsFullscreen && mode == FULLSCREEN_FULL);
 
+    pWindow->updateDynamicRules();
+    g_pCompositor->updateWindowAnimatedDecorationValues(pWindow);
+
     // make all windows on the same workspace under the fullscreen window
     for (auto& w : g_pCompositor->m_vWindows) {
         if (w->m_iWorkspaceID == pWindow->m_iWorkspaceID) {
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -292,3 +292,63 @@ void CWindow::setHidden(bool hidden) {
 bool CWindow::isHidden() {
     return m_bHidden;
 }
+
+void CWindow::applyDynamicRule(const SWindowRule& r) {
+    if (r.szRule == "noblur") {
+        m_sAdditionalConfigData.forceNoBlur = true;
+    } else if (r.szRule == "noborder") {
+        m_sAdditionalConfigData.forceNoBorder = true;
+    } else if (r.szRule == "noshadow") {
+        m_sAdditionalConfigData.forceNoShadow = true;
+    } else if (r.szRule.find("rounding") == 0) {
+        try {
+            m_sAdditionalConfigData.rounding = std::stoi(r.szRule.substr(r.szRule.find_first_of(' ') + 1));
+        } catch (std::exception& e) {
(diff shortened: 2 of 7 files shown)


vaxerski:
two things:
1) These should update on window title/class change events
2) aren't they a bit too ambiguous? maybe adding a flag like `windowrulev2d` would clear up this one? I'm up to debating this.

I also hope that if you don't specify either of the "dynamic" variables (e.g. `class=kitty` and that's it) that the rule won't magically stop working on some floating change.

ElJeffe:
1. Do you mean ALSO on title/class change events (makes sense) or ONLY on events? bc I do not see events  that will be triggered when needed.
2. I was actually expecting that the floating rule would react dynamically (e.g. opacity rule applies when a  window becomes floating) and not only when a window is created, so to me it does not feel ambiguous. What might be ambiguous is that only a subset of rules will apply on a _dynamic window identifier_ (`floating`, `fullscreen`,`pinned`,`window title` if i read 1. correcly). I choose to only let _visual rules_ be applied on dynamic window modifiers since that made more sense to me, hence only rules `noblur`, `noborder`, `noshadow`, `rounding`, `opacity`, `bordercolor` (can be expanded). I think if this is described on the wiki it is ok. A `windowrulev2d` would clutter the code and make a mix of `windowrulev2` and `windowrulev2d` not so easy to understand for the user. 
3. The static rules will keep on working, unless overridden by a dynamic window identifier. The last rule that matches will take precedence.
ie 

```
windowrulev2 = opacity 0.8 0.8,class:^(foot)$
windowrulev2 = bordercolor rgb(EE4B55),class:^(foot)$
windowrulev2 = opacity 0.5 0.5,floating:1
```
create a foot window -> rule 1 and 2 will apply
make the window floating -> opacity will change to 0.5, bordercolor stays red
make the window non-floating -> opacity will revert to 0.8, bordercolor stays red

but
```
windowrulev2 = opacity 0.5 0.5,floating:1
windowrulev2 = opacity 0.8 0.8,class:^(foot)$
windowrulev2 = bordercolor rgb(EE4B55),class:^(foot)$
```
create a foot window -> rule 2 and 3 will apply
make the window floating -> opacity will stay 0.8
make the window non-floating -> opacity will stay 0.8

-> if we would have a `windowrulev2` and `windowrulev2d`  the logic in code to accomplish this will be a lot harder imo.

vaxerski:
okay, fair point. Just make sure the wiki clearly states that.

For 1, I mean ALSO. e.g. when the app changes its title to "XYZ" I want it to change border color.

ElJeffe:
Added dynamic rule update on title change and added `noanim` `animation` and `opaque` to dynamic rules
```
windowrulev2 = bordercolor rgb(FFFF00),title:^(.*Hyprland.*)$
windowrulev2 = noanim,class:^(foot)$
windowrulev2 = animation slide,title:^(.*Hyprland.*)$
```

PR #1045, opened 2022-11-18
Setting wayland environment variables at startup

Narice:
Answers different issues brought up in #850
- sets `XDG_BACKEND` and `_JAVA_AWT_WM_NONREPARENTING` to `wayland` and `1` at startup, so that the user doesn't have to worry setting these up.
- removed these variables from nix modules recommended environment
- removed XCURSOR_SIZE from nix modules as it is already set

`NIXOS_OZONE_WL` for now stays in the nix modules as it would clutter non-nixos users environment variables and doesn't seem to interact badly with x11 sessions, though it it possible that it could.

it is fully ready to be merged

@fufexan

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -20,7 +20,7 @@ int main(int argc, char** argv) {
         cmd += std::string(i == 0 ? "" : " ") + argv[i];
     setenv("HYPRLAND_CMD", cmd.c_str(), 1);
     setenv("XDG_BACKEND", "wayland", 1);
-    setenv("_JAVA_AWT_WM_NONREPARENTING", "1", 1);
+    setenv("_JAVA_AWT_WM_NONREPARENTING", "1", 0);
 
     // parse some args
     std::string configPath;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -19,6 +19,8 @@ int main(int argc, char** argv) {
     for (auto i = 0; i < argc; ++i)
         cmd += std::string(i == 0 ? "" : " ") + argv[i];
     setenv("HYPRLAND_CMD", cmd.c_str(), 1);
+    setenv("XDG_BACKEND", "wayland", 1);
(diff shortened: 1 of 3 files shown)


vaxerski:
edqhijnoijfweeoi jwef werfoh jnf werikjnlkfwer erf kljnrt ekljnberfg erfg lkb jnerg lkjnb rg e?

src/main.cpp
```diff
@@ -19,6 +19,8 @@ int main(int argc, char** argv) {
     for (auto i = 0; i < argc; ++i)
         cmd += std::string(i == 0 ? "" : " ") + argv[i];
     setenv("HYPRLAND_CMD", cmd.c_str(), 1);
+    setenv("XDG_BACKEND", "wayland", 1);
+    setenv("_JAVA_AWT_WM_NONREPARENTING", "1", 1);
```
vaxerski, line 23: should we overwrite them if the user specified something else?

also wtf is xdg backend

Narice: for the XDG one, probably since it should always be wayland for wayland compositors, the java one I don't know though, it's probably ok to not override that, it's not set by i3 or xfce, don't know about sway, but that seems pretty particular.
Do you know anything about this var @fufexan?

Narice: It's apparently important for java swing apps: https://github.com/xmonad/xmonad/issues/126
it's probably ok to overwrite this as well, it needs to be anything else that not defined: https://bugs.openjdk.org/browse/JDK-8058197
so if a user defined it and we overwrite it with a 1, it shouldn't be a big deal I guess, as you want tbh

Narice: I've changed it to not overwrite to be safe

Narice:
forgot to ping @viperML.

vaxerski (approved):
LGTM for the code side. @fufexan for Nix.

fufexan (approved):
LGTM

Narice:
Thanks!

Issue #1053, opened 2022-11-20
Option to disable the workspace cycling round left


CactiChameleon9:
My description assumes the context of `workspace_swipe` being enabled.

Swiping right creates a blank workspace, and I cannot swipe further. However swiping left on workspace 1 brings me to my final non-empty workspace (workspace -1). It would be nice to have an option to disable this looping behaviour because it doesn't quite make sense with `workspace_swipe`

I made a quick diagram below in case my question didn't make sense:

```
1, [2], 3, empty
(right)
1, 2, [3], empty
(right)
1, 2, 3, [empty]
(right)
1, 2, 3, [empty]
(left)
1, 2, [3], empty
(left)
1, [2], 3, empty
(left)
[1], 2, 3, empty
(left - unwanted, prefer to stay at 1)
1, 2, [3], empty
(right)
1, 2, 3, [empty]
```

`###` Discussed in https://github.com/hyprwm/Hyprland/discussions/1052

PR #1067, opened 2022-11-21
Bugfix unintentional cycling workspace with workspace_swipe

CactiChameleon9:
This PR solves the issue #1053 by adding a new option `workspace_swipe_left_cycle` to `gestures`.
If the option is enabled, it is possible to swipe left on the 1st workspace to go the the -1 workspace (old behaviour)
If the option is disabled, it is not possible to go to the other end of the workspaces by swiping (matching the auto new workspace behaviour)

`workspace_swipe_left_cycle` is disabled by default (changing the default behaviour) but can be enabled if the user wishes

Yes, I would consider this ready for merging.

### DIFF
--- a/src/managers/input/Swipe.cpp
+++ b/src/managers/input/Swipe.cpp
@@ -188,7 +188,10 @@ void CInputManager::onSwipeUpdate(wlr_pointer_swipe_update_event* e) {
 
     m_sActiveSwipe.delta = std::clamp(m_sActiveSwipe.delta, (double)-*PSWIPEDIST, (double)*PSWIPEDIST);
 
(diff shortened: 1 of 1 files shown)


vaxerski:
If you disable creating new with swipe right, iirc it will also cycle when you swipe right on the last workspace. I think unifying that to `workspace_swipe_cycles` would be a better idea.

CactiChameleon9:
Hmmm... OK that is a little odd.
Default behaviour, `create_new = 0`:
- Cannot swipe left or right past the final workspace
Default behaviour, `create_new = 1`:
- Can swipe left past the final workspace, but cannot swipe right past the final workspace

Is this a bugfix then as opposed to a feature??
What should I implement? I can completely redo what I have done in order to enable swipe-around.... (doesn't work for create_new = 0 currently with or without my work). Or I could remove the setting and change this into a consistency bugfix

vaxerski:
Oh, wait, it doesn't wrap. Right. Yeah this is a bugfix then. (Sorry I am a bit retarded) Make it mandatory for now (not configurable)

CactiChameleon9:
Cool. Thats easier for me. No worries btw :)

CactiChameleon9:
Should be done. Feel free to test more, but seems fine to me :+1:

vaxerski:
Thanks!

PR #1085, opened 2022-11-23
Add option for retrieving first empty workspace

julianschuler:

This pull request adds a new option to reference workspaces with: "empty".
It references the first empty workspace, i.e. the workspace with the smallest (positive) workspace ID without any windows.

The new option allows to e.g. focus an empty workspace or move a window to an empty workspace by using something like this:

```
bind = $MOD, $KEY1, workspace, empty
bind = $MOD, $KEY2, movetoworkspace, empty
```

These examples are equivalent to the `ws_empty` and `ws_empty_move` actions in spectrwm.

I will add a pull request for updating the wiki entry soon after.

It is ready for merging.

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -244,6 +244,13 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
             result = WORKSPACE->m_iID;
         }
         outName = WORKSPACENAME;
+    } else if (in.find("empty") == 0) {
+        int id = 0;
+        while (++id < INT_MAX) {
+            const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(id);
+            if (!PWORKSPACE || (g_pCompositor->getWindowsOnWorkspace(id) == 0))
+                return id;
+        }
     } else {
         if ((in[0] == 'm' || in[0] == 'e') && (in[1] == '-' || in[1] == '+') && isNumber(in.substr(2))) {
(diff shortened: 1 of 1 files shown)


vaxerski:
There is a function to retrieve the highest workspace ID in CCompositor iirc. Why not use that?

Also, make a wiki entry prior to merge. I'll merge both at once

julianschuler:
I assume you mean `CCompositor::workspaceIDOutOfBounds()` and to use the workspace ID which is out of bounds as new empty workspace?

I decided against this approach due to three reasons:
1. There may be workspaces without any windows but still within `m_vWorkspaces`. Empty workspaces seem to be only removed in `CCompositor::sanityCheckWorkspaces()`. I am not sure whether this is called often enough to ensure that only non-empty or active empty workspaces are present when `getWorkspaceIDFromString()` is called.
2. If an empty workspace is currently focused, the focus would be switched to a new empty workspace, since the current empty workspace is considered to be in bounds.
3. If we have e.g. two non-empty workspaces with IDs 1 and 3, this approach would result in ID 4 while ID 2 would be the first empty workspace. In general, empty workspaces within the range given by the lowest to highest IDs of non-empty workspaces will never be considered and the IDs potentially becoming larger and larger.

Please correct me, if any of my assumptions are wrong or there is another reason to prefer `CCompositor::workspaceIDOutOfBounds()`. Thank your for all your work and time for Hyprland!

vaxerski:
I see your points. Makes sense, sure. Thanks!

PR #1115, opened 2022-11-27
Fix compile warning

hnboy:
Fix Wnon-virtual-dtor compile warning


src/layout/IHyprLayout.hpp
```diff
@@ -133,6 +133,10 @@ interface IHyprLayout {
     */
     virtual void         onWindowFocusChange(CWindow*);
 
+    /*
+      Add virtual descontruct fiunc
+    */
+    virtual ~IHyprLayout()=0;
```
vaxerski, line 139: destructors should be right below the constructor, or on the top if one is not present. Furthermore, useless comment, and styling: no spaces around the `=`

PR #1155, opened 2022-12-03
nix: add cmake to devshell and create a "wlroots-ready" devshell

FlafyDev:

- fixes `sh: line 1: cmake: command not found` when running `make debug` by adding `cmake` as a nativeBuildInput.

- adds a `wlroots-ready` devshell which puts wlroots in the devshell so nix compiles it.
  Especially useful for NixOS since `make config` can't install `wlroots`: (no `/usr/include`)
```
Installing subdir /mnt/general/repos/flafydev/Hyprland/subprojects/wlroots/include/wlr to /usr/include/wlr
Installation failed due to insufficient permissions.
```

none

ready for review/merge.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -81,17 +81,14 @@
         nativeBuildInputs = with pkgsFor.${system}; [
           cmake
         ];
+        buildInputs = [
+          self.packages.${system}.wlroots-hyprland
+        ];
         inputsFrom = [
           self.packages.${system}.wlroots-hyprland
           self.packages.${system}.hyprland
         ];
       };
-      wlroots-ready = default.overrideAttrs (oldAttrs: {
-        name = "hyprland-shell-wlroots-ready";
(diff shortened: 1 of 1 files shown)


vaxerski:
@fufexan

fufexan:
Why do this instead of building with meson?

FlafyDev:
> Why do this instead of building with meson?

Ah I was not aware building with meson was an option. But I still think the devshell should allow support for compiling with cmake. The wiki says it's the recommended way and it documents more for cmake than meson(like https://wiki.hyprland.org/Contributing-and-Debugging )

fufexan:
Alright. I think the default shell should have wlroots in it already, so no need for another one.

FlafyDev:
> Alright. I think the default shell should have wlroots in it already, so no need for another one.

done

fufexan (approved):
LGTM

PR #1158, opened 2022-12-03
blur xray for layers

FlafyDev:
Makes `blur_xray` xray all surfaces.

This PR makes blur_new_optimizaitons work for every surface and not only windows so would probably require some testing. 

From want I tried it works fine.

It works rn so ready for merging/testing.

### DIFF
--- a/src/helpers/WLClasses.hpp
+++ b/src/helpers/WLClasses.hpp
@@ -70,6 +70,7 @@ struct SRenderData {
 
     // for blurring
     bool blur = false;
+    bool blockBlurOptimization = false;
 
     // only for windows, not popups
     bool squishOversized = true;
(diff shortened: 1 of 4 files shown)


vaxerski:
using blur optimizations for layers below the window will result in graphical artifacts. Furthermore, why?

FlafyDev:
> using blur optimizations for layers below the window layer will result in graphical artifacts. Furthermore, why?

Oh, so should I add a property to `SRenderData` that prevents `USENEWOPTIMIZE` from being true and enable it for background and bottom layers?

And as for the why, same reason that blur_xray exists for windows: optimization
I am animating a layer shell in the top layer and it's slow rn since blur_xray doesn't work for layers. Also it looks bad being able to see the window behind imo. 

https://user-images.githubusercontent.com/44374434/205456806-03c97024-cfac-4fde-be5e-f7822a1cfa5f.mp4

vaxerski:
If you make it disabled for bottom and background layers it's ok.

PR #1165, opened 2022-12-03
Handle multiple keyboards with switchXKBLayoutRequest

leftas:

It fixes multiple keyboards issue (see picture below) when changing xkb layout with hyprctl.

image
 

There is a problem with how CVarList parses things - it ignores quotes("). Hence when you supply the first parameter, keyboard name, with spaces (see in the picture above) it interprets the keyboard name and param wrong. So either this request has to change the separator, add a wiki note regarding this issue, or change how CVarList parses things. 

Yes.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -692,41 +692,48 @@ std::string switchXKBLayoutRequest(std::string request) {
     const auto KB = vars[1];
     const auto CMD = vars[2];
 
-    // get kb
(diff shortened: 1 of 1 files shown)


vaxerski:
you mean duplicated keyboard names? That's a problem in keyboard naming, not switchxkblayout.

leftas:
> you mean duplicated keyboard names? That's a problem in keyboard naming, not switchxkblayout.

Oh yes, I agree, totally. But then either Hyprland should give a unique attribute(name/id or whatever) for all devices or iterate through all - like it was done here (I don't like this approach, tho). An additional argument for a unique attribute would be as I explained about the CVarList problem - that a lot of keyboards have spaces in their name and then it is failing to interpret the correct name and value.

vaxerski:
spaces should not be showing anymore, update hyprland.

vaxerski:
should be fixed with 409ac12f236c06cbb28756239e1d1536078b33bb

leftas:
Nice, thank you:)

PR #1174, opened 2022-12-04
Use internal device naming for Hyprctl devices

leftas:

Due to 409ac12f236c06cbb28756239e1d1536078b33bb commit, Hyprland internally uses a bit different name from the WLR device one. However, that was not reflected in `hyprctl devices`, this commit fixes the issue. 

No, expect that switches are still using the WLR device name.

Yes.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -383,7 +383,7 @@ R"#(    {
         "name": "%s"
     },)#",
                 &d,
-                d.name
+                d.name.c_str()
             );
         }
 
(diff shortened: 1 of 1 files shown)


src/debug/HyprCtl.cpp
```diff
@@ -383,7 +383,7 @@ R"#(    {
         "name": "%s"
     },)#",
                 &d,
-                d.pWlrDevice ? d.pWlrDevice->name : ""
+                d.name
```
vaxerski, line 386: std::string cannot be used directly in C format functions. You need to call `.c_str()` (also applies to all of the below)

leftas: Ah, thanks! Didn't notice this.

jlaunay:
I have the impression that this commit introduces a problem for the [per device input config](https://wiki.hyprland.org/Configuring/Keywords/#per-device-input-configs) .
I have for example `sensitivity = -0.5` for device `Logitech MX Master` but `hyprctl devices` return logitech-mx-master` that I can't use in hyprland.conf.

vaxerski:
why? 
```
device:logitech-mx-master {
    sensitivity = -0.5
}
```

jlaunay:
Because the wiki says "The name can be easily obtained by doing hyprctl devices." and as far as I can remember this is how I got `Logitech MX Master` before

vaxerski:
it is taken into account and that's what the name was internally converted to anyways.

jlaunay:
Ok, great, thank you very much for these clarifications and for all the work done on Hyprland.

jlaunay:
I don't understand why but the following works
```
device:Logitech MX Master {
    enabled = false
}
```
but using the name given by `hyprctl devices` have no effect
```
device:logitech-mx-master {
    enabled = false
}
```

vaxerski:
:P

PR #1179, opened 2022-12-04
Implement window sharing with the hl toplevel export proto

vaxerski:
Generally works.

Depends on:
 - https://github.com/hyprwm/hyprland-protocols/pull/1
 - https://github.com/hyprwm/xdg-desktop-portal-hyprland/pull/2

Blockers:
-  [x] Nix
-  [x] Meson

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "wlroots"]
 	path = subprojects/wlroots
 	url = https://gitlab.freedesktop.org/wlroots/wlroots.git
+[submodule "subprojects/hyprland-protocols"]
+	path = subprojects/hyprland-protocols
+	url = https://github.com/hyprwm/hyprland-protocols
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ target_link_libraries(Hyprland
         pthread
         ${CMAKE_THREAD_LIBS_INIT}
         ${CMAKE_SOURCE_DIR}/ext-workspace-unstable-v1-protocol.o
+        ${CMAKE_SOURCE_DIR}/hyprland-toplevel-export-v1-protocol.o
 )
 
 IF(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES DEBUG)
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,16 @@ wlr-output-power-management-unstable-v1-protocol.c:
 
 wlr-output-power-management-unstable-v1-protocol.o: wlr-output-power-management-unstable-v1-protocol.h
 
+hyprland-toplevel-export-v1-protocol.h:
+	$(WAYLAND_SCANNER) server-header \
+		subprojects/hyprland-protocols/protocols/hyprland-toplevel-export-v1.xml $@
+
+hyprland-toplevel-export-v1-protocol.c:
+	$(WAYLAND_SCANNER) private-code \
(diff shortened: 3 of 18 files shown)


fufexan:
Will fix.

fufexan:
Anyone good at Meson? I'm not sure how to add the protocol to `protocols/meson.build` akin to how it's done in CMake.
@sp1ritCS sorry for pinging, would you happen to know?

sp1ritCS:
Doesn't adding it as `[meson.project_source_root(), 'subprojects/hyprland-protocols/protocols/hyprland-toplevel-export-v1.xml']` to the protocols array work?

fufexan:
> Doesn't adding it as `[meson.project_source_root(), 'subprojects/hyprland-protocols/protocols/hyprland-toplevel-export-v1.xml']` to the protocols array work?

@sp1ritCS nope, it errors with this
```
protocols/meson.build:29:1: ERROR: Sandbox violation: Tried to grab file hyprland-toplevel-export-v1.xml from a nested subproject.
```

fufexan:
@eli-schwartz sorry for the ping, but you helped us in the past. Do you know the proper way of doing this?

Maybe I should try to package `hyprland-protocols` and reference it `wayland-protocols`-style?

eli-schwartz:
I'm always happy to be pinged for Meson questions, no need to apologize. :)

There are two ways to do this, and both of them use the same underlying functionality.
- package `hyprland-protocols` and install it, then acquire those files via pkg-config.
- use hyprland-protocols as a Meson subproject, mark the protocols as installable, and
  - generate a pkg-config file with e.g. a `protocolsdir` variable pointing to the install location
  - create a `declare_dependency()` exposing the same variable name, but pointing to the in-source-tree location

You can then look up the dependency and `dep.get_variable('protocolsdir') / 'hyprland-toplevel-export-v1.xml')` will comply with the sandbox because it uses the dependency interface to find it. This will work even if hyprland-protocols is a subproject.

fufexan:
> But I assume @vaxerski would prefer not to add a meson file to the hyprland protocols repo upside_down_face .

I think it's fine. He lets me add Nix stuff everywhere, so there's even more reason to add Meson stuff :D.

> You might get arround the "sandbox" by symlinking subprojects/hyprland-protocols to somewhere outside subprojects :)

I've tried doing that and it doesn't really work. I think I'll try adding a meson.build like both you and @eli-schwartz suggested.

Thanks for the input!

eli-schwartz:
I'm always happy to be pinged for Meson questions, no need to apologize. :)

There are two ways to do this, and both of them use the same underlying functionality.
- package `hyprland-protocols` and install it, then acquire those files via pkg-config.
- use hyprland-protocols as a Meson subproject, mark the protocols as installable, and
  - generate a pkg-config file with e.g. a `protocolsdir` variable pointing to the install location
  - create a `declare_dependency()` exposing the same variable name, but pointing to the in-source-tree location

You can then look up the dependency and `dep.get_variable('protocolsdir') / 'hyprland-toplevel-export-v1.xml')` will comply with the sandbox because it uses the dependency interface to find it. This will work even if hyprland-protocols is a subproject.

fufexan:
@eli-schwartz @sp1ritCS mind taking a look at https://github.com/hyprwm/hyprland-protocols/commit/0dcff94fc10df2bbb66d3e1b5a1d6cfd3ada5515 ? Maybe I've overcomplicated it. I took it from `wayland-protocols` and adapted.

PR #1191, opened 2022-12-07
Input: Fix unlock mouse warp

scorpion-26:
Unconstraining a locked mouse would result in the mouse warping to (0, 0), since wlroots give us these as the current mouse position. This PR adds a rather hacky fix for this by manually storing the last mouse position and warp the cursor upon unlocking

Tested with GLFW and glfwSetInputMode() (For reference, GLFW uses zwp_relative_pointer and zwp_locked_pointer)

Current behavior 
 - Lock cursor at (x,y) with glfwSetInputMode(GLFW_CURSOR_DISABLED)
 - Move mouse
 - Unlock mouse with glfwSetInputMode(GLFW_CURSOR_NORMAL)
 - Cursor is warped to the topleft of the window (0,0)
 
Behavior with this PR
 - Lock cursor at (x,y) with glfwSetInputMode(GLFW_CURSOR_DISABLED)
 - Move mouse
 - Unlock mouse with glfwSetInputMode(GLFW_CURSOR_NORMAL)
 - Cursor is warped back to (x,y)

Unfortunately I'm not that familiar of wlroots so I don't know if (un)constraining a mouse is the only time newConstraint/destroyConstraint is called.  

This PR is ready for merging, though I'm open for suggestions/improvements

### DIFF
--- a/src/events/Devices.cpp
+++ b/src/events/Devices.cpp
@@ -105,6 +105,10 @@ void Events::listener_newConstraint(wl_listener* listener, void* data) {
     CONSTRAINT->pMouse = g_pCompositor->m_sSeat.mouse;
     CONSTRAINT->constraint = PCONSTRAINT;
 
+    // HACK: Store last mouse pos
+    CONSTRAINT->x = g_pCompositor->m_sWLRCursor->x;
+    CONSTRAINT->y = g_pCompositor->m_sWLRCursor->y;
+
     CONSTRAINT->hyprListener_destroyConstraint.initCallback(&PCONSTRAINT->events.destroy, &Events::listener_destroyConstraint, CONSTRAINT, "Constraint");
     CONSTRAINT->hyprListener_setConstraintRegion.initCallback(&PCONSTRAINT->events.set_region, &Events::listener_setConstraintRegion, CONSTRAINT, "Constraint");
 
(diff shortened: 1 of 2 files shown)


vaxerski:
Not how the protocol works.

If you lock the cursor you should also [send a cursor_hint](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml#L218-236) if you want the cursor to reappear at a specific location.

scorpion-26:
Thanks for the correction on the protocol. 

The protocol suggests however that a "compositor may use this information to warp the pointer upon unlock in order to avoid pointer jumps", meaning that a compositor is allowed to ignore it. 

I understand that it is probably best to honor the cursor hint, but warping the cursor to (0,0) if it isn't set doesn't seem right to me. Sway for example does the same thing as (wrongly) implemented in this PR(meaning not warping to (0,0)). 
Should I open an Issue instead, or is the warping to (0,0) 100% as intended?

vaxerski:
it is as intended. wlroots does not offer a signal for setting the hint. Maybe I should make a PR to wlroots to expose this signal so that we can track whether the surface has actually provided a hint or not.

vaxerski:
actually, nvm. There already is a way to get that information

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_pointer_constraints_v1.c#L103

I'll sort it out tomorrow if I don't forget.

scorpion-26:
Thanks!

vaxerski:
fyi 62f4503f07066762af9546bb4a2acdd58b72d4bf

Issue #523, opened 2022-08-13
Monocle alternative to full screen. (and automatic monocle?)


nosvagor:
I have a large monitor, and full screen is often too much. Other window managers like bspwm have a monocle layout, which I think is just all the windows stacked on top of each other and can be cycled or something like that. 

However, my primary use case: if I have just one or two windows, I'd prefer a different layout, with much larger gaps. Ideally, I could also toggle this layout to get a pseudo full screen effect on workspaces with many windows. 

I realize that I could probably manually with sockets (https://github.com/hyprwm/Hyprland/wiki/Expanding-functionality), but I'm not quite sure how to do so. Perhaps expanding this wiki (https://github.com/hyprwm/Hyprland/wiki/IPC) would be enough to allow me to create such scripts.

vaxerski:
technically, monocle is automatic when you enter fullscreen. Using cyclenext / movefocus will select a new fullscreen window.

> However, my primary use case: if I have just one or two windows, I'd prefer a different layout, with much larger gaps. Ideally, I could also toggle this layout to get a pseudo full screen effect on workspaces with many windows.

"pseudo" fullscreen, you mean fullscreen 1 (maximize)? That's already a thing.

A script - you need to be more precise when requesting a script. Your descriptions are pretty vague.

nosvagor:
Oh, missesd

PR #1213, opened 2022-12-11
Master layout: lose or inherit fullscreen on deliberate window switching

proycon:

In the master layout as implemented up to this point, when you have a master window, some slave windows and you fullscreen any of them, the fullscreen status remained on the old window even if  you switch to another window via `cyclenext`,`cycleprev`, `swapmaster`, etc...   This meant you effectively couldn't see the window you switched to, had to switch back and manually toggle off fullscreen again. I consider this a bug (though surprisingly didn't find an open issue about it). This PR fixes it in one of two ways:

1. Simply automatically unset the fullscreen status when explicitly switching away from it, showing you all the tiles again (new default behaviour)
2. Like one, but additionally inherit the fullscreen status by immediately re-setting it on the new window. This requires that a new property `master:inherit_fullscreen` is set. This results in behaviour like 'monocle' layout in other tiling WMs. May also be somewhat related to #516. I think this is also more in line with how the dwindle layout does things, judging the comment in #523.

In some cases (like the new orientating switching), I implemented option 1 as the sole option which made sense, otherwise the result of the action would be invisible.

I introduced two new helper functions `prepareLoseFocus` and `prepareNewFocus` to keep the amount of repetition in the code lower.

Should be ready to go

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -580,6 +580,8 @@ void CHyprMasterLayout::switchWindows(CWindow* pWindow, CWindow* pWindow2) {
     if (!PNODE2 || !PNODE)
         return;
 
+    prepareLoseFocus(pWindow);
+
     if (PNODE->workspaceID != PNODE2->workspaceID) {
         std::swap(pWindow2->m_iMonitorID, pWindow->m_iMonitorID);
         std::swap(pWindow2->m_iWorkspaceID, pWindow->m_iWorkspaceID);
@@ -681,6 +683,15 @@ CWindow* CHyprMasterLayout::getNextWindow(CWindow* pWindow, bool next) {
     return nullptr;
 }
 
+void CHyprMasterLayout::prepareLoseFocus(CWindow * pWindow) {
+    if (!pWindow)
+        return;
+
+    //if the current window is fullscreen, make it normal again if we are about to lose focus
+    if (pWindow->m_bIsFullscreen)
+        g_pCompositor->setWindowFullscreen(pWindow, false, FULLSCREEN_FULL);
+}
+
 std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::string message) {
     auto switchToWindow = [&](CWindow* PWINDOWTOCHANGETO) {
         if (!g_pCompositor->windowValidMapped(PWINDOWTOCHANGETO))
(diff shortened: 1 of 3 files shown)


src/layout/MasterLayout.cpp
```diff
@@ -681,6 +685,27 @@ CWindow* CHyprMasterLayout::getNextWindow(CWindow* pWindow, bool next) {
...
+    }
+
+    return false;
+}
+
+void CHyprMasterLayout::prepareNewFocus(CWindow * pWindow, bool inherit_fullscreen) {
+    if (!pWindow)
+        return;
+
+    if (inherit_fullscreen)
```
vaxerski, line 705: we use camelCase

src/layout/MasterLayout.cpp
```diff
@@ -681,6 +685,27 @@ CWindow* CHyprMasterLayout::getNextWindow(CWindow* pWindow, bool next) {
...
+    //if the current window is fullscreen, make it normal again if we are about to lose focus
+    if (pWindow->m_bIsFullscreen) {
+        g_pCompositor->setWindowFullscreen(pWindow, false, FULLSCREEN_FULL);
+        return g_pConfigManager->getConfigValuePtr("master:inherit_fullscreen")->intValue == 1;
+    }
+
+    return false;
+}
+
+void CHyprMasterLayout::prepareNewFocus(CWindow * pWindow, bool inherit_fullscreen) {
```
vaxerski, line 701: `CWindow*`

src/layout/MasterLayout.cpp
```diff
@@ -681,6 +685,27 @@ CWindow* CHyprMasterLayout::getNextWindow(CWindow* pWindow, bool next) {
...
 }
 
+bool CHyprMasterLayout::prepareLoseFocus(CWindow * pWindow) {
+    if (!pWindow)
+        return false;
+
+    //if the current window is fullscreen, make it normal again if we are about to lose focus
+    if (pWindow->m_bIsFullscreen) {
+        g_pCompositor->setWindowFullscreen(pWindow, false, FULLSCREEN_FULL);
+        return g_pConfigManager->getConfigValuePtr("master:inherit_fullscreen")->intValue == 1;
```
vaxerski, line 695: you're not caching the ptr

vaxerski:
shouldn't the default behavior be set to 1 to better follow dwindle?

src/layout/MasterLayout.cpp
proycon: fixed now

src/layout/MasterLayout.cpp
proycon: should be fixed now

src/layout/MasterLayout.cpp
proycon: fixed now

Issue #79, opened 2022-05-16
Scaling doesnt work


Magnitudexx:
**Describe the bug**
On a 4K display, I try to run 2x scaling, resulting in the compositor only running in a quarter ofthe screen.

**Images/videos/anything that would help**
I'll try getting one

Please attach a log (it's in /tmp/hypr/hyprland.log) and, if crashing, a crashdump (coredumpctl, then coredumpctl info <pid>)
[hyprland.log](https://github.com/vaxerski/Hyprland/files/8699819/hyprland.log)

vaxerski:
i know. It's on the todo list. It's a real pain in the ass to code in, at least has been for me.

Magnitudexx:
I bet, I realized that I've seen it somewhere else before so I started looking into it but I cannot understand why it does this

vaxerski:
please try the [scaling branch](https://github.com/vaxerski/Hyprland/tree/scaling), notable features not working: damage tracking on full.

vaxerski:
blur fixed

vaxerski:
fadeout fixed

Magnitudexx:
scaling works ok now. Librewolf (fork of firefox) doesn't render fully, leaving a big gap to the border

vaxerski:
yeah, there might be bugs. I need to investigate more.

vaxerski:
fixed (as far as im aware) and recently merged. Report any new bugs separarely.

PR #1228, opened 2022-12-14
do not overwrite existing `hyprland.desktop`

NotAShelf:
Implements a "fix" for the long-standing issue of desktop files being overwritten on each `make install` by copying only if `hyprland.desktop` does not exist. Also removes `cleaninstall` option because it is obsolete with this PR.

N/A

Tested on my own desktop, worked. If it does not work on your machine, your system is haunted.


NotAShelf:
I will make a PR to the wiki if this looks ready to merge.

vaxerski:
firstly, don't remove cleaninstall, deprecate it. Aka, just echo "deprecated blah blah" and run make install

secondly, wiki pr

thirdly, this wont fix the packages (which is most of people), but cool i guess.

NotAShelf:
> firstly, don't remove cleaninstall, deprecate it. Aka, just echo "deprecated blah blah" and run make install
Pretty sure I was the only one using that, but sure. I'll make it echo something and run make install instead

> secondly, wiki pr
The PR is ready to be made, it was just awaiting confirmation

> thirdly, this wont fix the packages (which is most of people), but cool i guess.
it *should* fix those packages, don't those use install in the PKGBUILD? (didn't read the PKGBUILD how could you tell)

vaxerski:
nope they don't cuz they can't. They run in a sandbox.

NotAShelf:
darn, maybe I can patch the PKGBUILDs with some workaround

in the meantime I'll fix this PR and send the wiki PR your way

NotAShelf:
...and I forgot to mention, this PR also does move the copy function for the desktop file below the binaries because we don't want to have a desktop file that doesn't work if binaries don't work for some reason

shouldn't matter anyway

NotAShelf:
Should be good now, wiki PR made over [#79](https://github.com/hyprwm/hyprland-wiki/pull/79)

Issue #528, opened 2022-08-15
Option for `blurls` to look better with rounded corners/fully transparent pixels


Gutawer:
`blurls` is great for my notification and bar styling, but right now I can't really put rounded corners on them as Hyprland still blurs the pixels that aren't really part of the window (corner radius on `dunst` exaggerated to show the effect):

![image](https://user-images.githubusercontent.com/9727338/184680755-f44cc95f-7f22-40a7-82a9-11c51f601f6f.png)

In all of my cases this would be fixable by simply not blurring pixels that are completely transparent, but there might be some better way of achieving this that I'm not thinking of.

Apologies if there already is a way around this!

vaxerski:
we're discussing this as a part of the ext-blur protocol @ wayland-protocols: [https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/43](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/43)

I am not sure whether I should introduce a per-layer option in the meantime before this gets merged and adopted. Lmk.

vars1ty:
Although not affected currently, having the option to exclude certain layers from blur would be nice.

vaxerski:
well that's already done with blurls no?

vars1ty:
Almost forgot about `blurls` smh, mb lmao

amilonop:
Any updates on this issue? Would lead into some *super* cool UIs with EWW and waybar :D

HanoJing:
Yeah, I have the same issue I guess. Layers that are blurred using `blurls=NAMESPACE` have corners pretruding from the rounded corners seen on fuzzel for example (although the corners are very subtle).

vaxerski:
done in e5870d47c74150aa744ed55f9b7e58713d968209

PR #1241, opened 2022-12-16
Add option to round corners on layers

yanisdb:

Adds the keyword `roundls` (similar to `blurls`) to add rounded corner to a specified layer.

This is useful because it enables us to have rounder corner and blur on notifications, Rofi, and similar tools. Indeed, even though we can make those applications rounded without Hyprland (using CSS generally), if we do so and also use blur, we will see some artefacts in the windows corner:

2022-12-16T23:42:44,794219812+01:00

To fix that, I added the option to render with rounded corners in layers.

There is not much change, and it seems to work perfectly, but this is my first time working on a WM, so you might want to check everything.

I hope it is. Let me know if anything isn't right.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -887,6 +887,16 @@ void CConfigManager::handleBlurLS(const std::string& command, const std::string&
     m_dBlurLSNamespaces.emplace_back(value);
 }
 
+void CConfigManager::handleRoundingLS(const std::string& command, const std::string& value) {
+    if (value.find("remove,") == 0) {
+        const auto TOREMOVE = removeBeginEndSpacesTabs(value.substr(7));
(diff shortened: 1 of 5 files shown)


vaxerski:
The consensus to avoid blurred corners has been generally to add an option to ignore a == 0 pixels in blur of ls. No clue why this is needed.

yanisdb:
I didn't know about the option you're talking about. But, I think having a `roundls` option is useful because then, you don't have to figure out how to configure rounded corner for every piece of software. You can just make everything rounded from Hyprland directly, layers and windows. In my case, I want everything to be rounded, no exception.

But I just came to Wayland, maybe I'm completly missing the points and the goal of layers

vaxerski:
layers are supposed to style themselves. That's generally how it works.

> I didn't know about the option you're talking about.

it's not implemented yet, see #528
ignoring a == 0 is a better idea though.

PR #1242, opened 2022-12-17
Fix cycleprev bug introduced in 46891b12cf (#1213)

julianschuler:

This pull request fixes that `cycleprev` always focuses the next window (like `cyclenext`) instead of the previous one.

The reason for this is an incorrect boolean flag for `getNextWindow()` introduced in commit 46891b12cf belonging to #1213.

Nothing here.

It is ready for merging.


vaxerski:
thanks!

PR #1245, opened 2022-12-17
Add 'exact' option for 'splitratio'

julianschuler:

This pull request adds an `exact` option to `splitratio`. It allows to specify an exact value to set the split ratio to instead of a relative one.
This is useful for scripting purposes as well as resetting the split ratio to a known value.

Consider the following example:

```
bind = [...], splitratio, +0.05       # increase the split ratio by 0.05
bind = [...], splitratio, exact 0.6   # set the split ratio to 0.6
```

Additionally, I simplified the function `getPlusMinusKeywordResult()` making it also robust against any amount of whitespaces between `exact` and the provided value. I also replaced all tabs by spaces in the same file according to the style guide.

It is ready for merging.


vaxerski:
please make an accompanying wiki pr

vaxerski:
thanks!

PR #1202, opened 2022-12-10
Implement orientation (placement of master area) for master layout

proycon:
#### Describe your PR, what does it fix/add?

This implements the idea I posted in https://github.com/hyprwm/Hyprland/issues/1059 to make the master layout more friendly for use on vertical monitors. Each workspace now has an `orientation` which states where the master area should be (I opted for the more generic term `orientation` rather than `masterplacement`).  The values can be `left` (default, so nothing changes for existing configurations), `top`, `right` and `bottom`.

For the layoutmsg dispatcher the following commands are added:

* orientationleft
* orientationtop
* orientationright
* orientationbottom
* orientationnext   (cycles clockwise)
* orientationprev  (cycles counter-clockwise)

Here's a screenshot of a vertical monitor using the bottom orientation, with the master area at the bottom, and two slave windows on top:

Screenshot of the bottom orientation

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

Should be good to go.

#### Is it ready for merging, or does it need work?

Tested to the best of my ability and ready for merging

PR #1259, opened 2022-12-20
fix for gap in master layout orientation right

### LINKED ISSUE #1171: Master layout unwanted gap between master and client on vertical bar
### Steps to Reproduce
1. Enable master layout by running `hyprctl keyword general:layout master`, or enabling it in `hyprland.conf`:
```
general {
   layout=master
}
```
2. Set up and enable vertical bar with "exclusive" zone from compositor option ON. In my case, I tried eww and waybar with both resulting the same bug.
3. The vertical bar could be on either left or right side of the screen

4. Open master and stacks clients
5. Observe the gap between master and stacks

### Expected Outcome
1. No gap increase between master and stacks
2. The client not pushed far to the right

### Noted Outcome
1. Extra gap space between master and stacks  (probably Hyprland regard it as an exclusive zone)
2. Stack pushed far to the right
    - When the bar is on the right, the client window is pushed to the bar area
    - When the bar is on the left, the client window is pushed off the screen

Consult the attached video below:

https://user-images.githubusercontent.com/20262272/205470176-2d7ff404-ecd3-49f0-9e4d-ddf6fa484283.mp4

proycon:

Fixes #1171, most of which was already resolved by #1202, but one edge case remained.

Tested to work

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -240,7 +240,7 @@ void CHyprMasterLayout::calculateWorkspace(const int& ws) {
         for (auto& n : m_lMasterNodesData) {
             if (n.workspaceID == PWORKSPACE->m_iID && n.isMaster) {
                 if (PWORKSPACEDATA->orientation == ORIENTATION_RIGHT) {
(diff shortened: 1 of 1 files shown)


proycon:
Oops, no, I found a bug still.. this time on the left.. marking as work in progress and reopening when done

PR #1263, opened 2022-12-21
nix: handle reloading of multiple/dead instances on home manager generation switch

blurgyy:

When trying to switch a home-manager generation from the tty and Hyprland is closed for some reason, the switching process fails because the `onChange` hook tries to use the dead instance's socket.

This PR handles this situation better by trying to reload each instance under `/tmp/hypr` and ignore the dead instances.

No.

It's ready for merging.


fufexan (approved):
LGTM. I didn't think of this edge case.

PR #1296, opened 2022-12-30
s/pkgs.system/pkgs.hostPlatform.system/g

colemickens:

I have aliases disabled ever since a user of one of my repos sent a PR for it... so now I'm paying it forward I guess.

Thanks for hyprland... I think... still getting the initial config going.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,8 +57,8 @@
         mesonFlags = oldAttrs.mesonFlags ++ ["-Dexperimental=true"];
       });
 
-      xdg-desktop-portal-hyprland = inputs.xdph.packages.${prev.system}.default.override {
-        hyprland-share-picker = inputs.xdph.packages.${prev.system}.hyprland-share-picker.override {inherit hyprland;};
+      xdg-desktop-portal-hyprland = inputs.xdph.packages.${prev.hostPlatform.system}.default.override {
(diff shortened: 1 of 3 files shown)


vaxerski:
@fufexan

fufexan:
Thanks for this.

Mind explaining what the benefit of using `hostPlatform` is? It's not directly clear to me.

colemickens:
`pkgs.system` is wrong in some scenarios, like cross-compiling. Also, is just doesn't exist for me, since I have these aliases disabled (forces me to notice things like this, keep up with tech rot/debt, etc).

In this case, when I cross-compile `pkgs.system` is going to be `x86_64-linux`, but `pkgs.hostPlatform.system` is going to be `riscv64-linux`, etc.

fufexan:
Ah, okay. That makes sense. Thanks for the explanation.

fufexan:
Btw `prev.system` exists in the overlay inside flake.nix. Shouldn't that be changed as well?

colemickens:
Okay, I tested, you're right, it was necessary if I forced the use of the overlay with nixpkgs.config.allowAliases = false. Tested with the change, it fixes eval. Should be good to go now.

fufexan:
Thank you!

PR #1340, opened 2023-01-08
Pin dispatcher for a specific window

### LINKED ISSUE #1339: [FR] pinWindow dispatcher
I request a pinWindow dispatcher that takes a `window` rule as an argument

raffaem:
Closes #1339


vaxerski:
why make a new dispatcher? We've done this before.

Make it use the same `pin` dispatcher, but if the arg is empty or `active` just use last window.

raffaem:
> why make a new dispatcher? We've done this before.
> 
> Make it use the same `pin` dispatcher, but if the arg is empty or `active` just use last window.

done

vaxerski:
make a wiki pr plox

Issue #1346, opened 2023-01-08
Window resizing by dragging on border/gaps


horriblename:
I'd like to be able to resize windows by dragging on their border or gaps

PR #1347, opened 2023-01-08
Resize on border

horriblename:
#1346

- The resizing function for tiled windows is kinda wonky for this, if you try to drag the left edge of a window it might resize the right side instead, I might make a separate pr for that

Yes


vaxerski:
if its not ready make it a draft

src/Compositor.cpp
```diff
@@ -560,12 +560,14 @@ void findExtensionForVector2D(wlr_surface* surface, int x, int y, void* data) {
 }
 
 CWindow* CCompositor::vectorToWindowIdeal(const Vector2D& pos) {
-    const auto PMONITOR = getMonitorFromVector(pos);
+    const auto PMONITOR     = getMonitorFromVector(pos);
+    const auto FOCUS_EXTENT = g_pConfigManager->getConfigValuePtr("general:resize_on_borders")->intValue ? g_pConfigManager->getConfigValuePtr("general:border_size")->intValue : 0;
```
vaxerski, line 564: both vars should be static and then FOCUS_EXTENT should be using the static vars

vaxerski (changes requested):
Two more things (apart from the comment I left) I'd like seeing:
 - an option to expand the grab area to eat into the window a bit (especially on floating windows the grab area is unintuitive and annoying
 - an option to enable the cursor image to show the grab possibility (like in KDE/Gnome, the different "resize" cursors) - you can use `setCursorImageUntilUnset` and `unsetCursorImage` for that.

horriblename:
> an option to expand the grab area to eat into the window a bit (especially on floating windows the grab area is unintuitive and annoying

I just hopped on Gnome and it seems that they extend the grab area _outwards_, it made more sense to me since it won't completely deny certain area from the app so I'll go with that 

Also, I just noticed dragging window while fullscreen isn't allowed, why is that? I can't decide if i should disable focusable border + dragging cursor icon in fullscreen

vaxerski:
> Also, I just noticed dragging window while fullscreen isn't allowed, why is that?

why would it be? a fullscreen window should not be dragged to be resized.

horriblename:
sry, I was talking about pinned windows over fullscreen, I could've sworn they did nothing when I tried to resize them last night. It works now so I guess I'm just seeing things

src/Compositor.cpp
horriblename: not sure if this is what you want, pls double check!

vaxerski:
not sure if you've finished with the code yet, but re-request a review from me when you do.

Two things with the icons:
 - moving a mouse over the corner will flash the pointer if you go over the border
 - there are only corner icons, while there should be also non-corner ones
 
 an option to disable those cursors would be nice, as well as maybe a way to configure a modifier that enables this functionality? I should add transparent binds, that would do it.

horriblename:
> re-request a review from me when you do.

ok got it thx

> moving a mouse over the corner will flash the pointer if you go over the border

can you send a screenshot/recording? I'm not sure what u mean

> as well as maybe a way to configure a modifier that enables this functionality? 

If u meant changing the cursor icon only when a modifier key is held down, I think it's somewhat redundant to what we already have with mouse binds, I mean, y would anyone hold down the key then point at the border if they can just use mouse binds?

vaxerski:
> If u meant changing the cursor icon only when a modifier key is held down, I think it's somewhat redundant to what we already have with mouse binds, I mean, y would anyone hold down the key then point at the border if they can just use mouse binds?

right, my bad.

> can you send a screenshot/recording? I'm not sure what u mean

odd, when trying to record it's not visible on the recording. Maybe it's my laptop or the fact I am using a nest.

vaxerski:
hows progress

horriblename:
my system ran out of space and now I have a half corrupted system :facepalm:. I can't build anything atm, can you help try out the latest commit and see if it fix this bug:

right click menus that can't be clicked when the cursor is outside the region of the parent window, like this in dolphin:

image

if this works that'll be everything

horriblename:
is there anyway to detect when a window has a right click menu?

horriblename:
if the mouse is outside of the main window's bounds it's treated as if it's on a border, so subsurfaces (?) that stretch outside the window is partially treated as border

horriblename:
I decided to check for `g_pCompositor->m_vXDGPopups.empty()`, and disabling resizing border when there are any popups present; should I also check if the popup is on the same monitor? idk how popups behave in multi-monitor (does clicking close a popup on another monitor?)

vaxerski:
closing / opening popups is down to the application not hyprland

vaxerski:
can't review until there are conflicts. Rebase against head

horriblename:
done

horriblename:
also the way I did the cursor icon seemed rly bad, I set the cursor icon every single tick of mouse movement. should maybe change that?

vaxerski:
> also the way I did the cursor icon seemed rly bad, I set the cursor icon every single tick of mouse movement. should maybe change that?

if that's still your method, yea

vaxerski:
> the reason being IHyprLayout::onMouseMove calls onEndDragWindow very often which resets all cursor icons 

why not fix the source instead of ugly patches to the method? Isn't onEndDragWindow ignored on no drag anyways?

src/layout/IHyprLayout.cpp
```diff
@@ -195,10 +195,13 @@ void IHyprLayout::onBeginDragWindow() {
 void IHyprLayout::onEndDragWindow() {
     const auto DRAGGINGWINDOW = g_pInputManager->currentlyDraggedWindow;
 
-    g_pInputManager->unsetCursorImage();
-
-    if (!g_pCompositor->windowValidMapped(DRAGGINGWINDOW))
+    if (!g_pCompositor->windowValidMapped(DRAGGINGWINDOW)) {
+        if (DRAGGINGWINDOW)
```
vaxerski, line 199: this is always false

vaxerski (changes requested):
Looks neat. One small peeve.

Thanks for the hasPopupAt. Nice addition.

src/layout/IHyprLayout.cpp
horriblename: getting rid of that line would mean if a window closes while being dragged the icon would be stuck in the dragging one. Although realistically nobody would ever encounter this situation, still yeet?

vaxerski: huh you right. However, in that case, you should reset dragging window to a nullptr.

horriblename: done!

vaxerski (approved):
LGTM

vaxerski:
Ready to merge when checks pass, if you could make a wiki pr with new opts it would be great.

src/Window.hpp
```diff
@@ -7,6 +7,7 @@
 #include "render/decorations/IHyprWindowDecoration.hpp"
 #include <deque>
 #include "config/ConfigDataValues.hpp"
+#include "src/helpers/Vector2D.hpp"
```
vaxerski, line 10: ah yeah. `"helpers/..."`

horriblename:
aight
wiki pr: https://github.com/hyprwm/hyprland-wiki/pull/130

PR #1383, opened 2023-01-14
remove IsSwitchingToPrevious variable and checks

### LINKED ISSUE #1218: Previous workspace is forgotten or cycles too far
This applies to when `binds:allow_workspace_cycles` is set to `false`:

The `workspace previous` command is not working as I'd expect.  I'm using the master layout (though the code seems to generically apply to all) and I would expect `workspace previous` always to return to the previous workspace (the one I was on before switching), effectively allowing me to toggle between my last two workspaces indefinitely. Instead, I find that cycling to the previous workspace is no longer possible after I used it once.

observed behaviour:

```
workspace 1 -> ok
workspace 2 -> ok
workspace 3 -> ok
workspace previous -> takes me to 2, ok
workspace previous -> takes me to 1, not ok, I have allow_workspace_cycles disabled
workspace previous -> takes me to 3 (why now?)
workspace previous -> no action, keeps me at 3
.. stuck at 3..
```

expected behaviour:

```
workspace 1
workspace 2
workspace 3
workspace previous -> takes me to 2
workspace previous -> takes me to 3
workspace previous -> takes me to 2
workspace previous -> takes me to 3
..repeat ad infinitum..
```

Note:

Though all of this is on a single monitor, I do run a multi-monitor setup with workspace 10 (and only 10) on the second monitor. I haven't switched to it in the above tests but can't rule out that multi-monitor is somehow a factor in this bug.

devil-may-c0de:
fixes #1218

it removes the IsSwitchingToPrevious variable from the previous workspace logic. this means that when switching to a previous workspace, switching to the previous workspace again will bring you to the window that switched to previous workspace ie; workspace 1 > previous workspace 2 > previous  workspace 1 > previous  workspace 2

this is in line with it's implementation on other window managers, as well as more consistent with what is expected when switching to the previous workspace.

could break some workflows, although I can't seem to find anyone that wanted this behavior, apart from maybe the original coder of this functionality. using this alone for a few days, I haven't seen anything break, nor have I seen this variable in other parts of the project.

maybe it's better to have a config option to decide this rather than a hardcoded fix. I don't think there's any problems with this patch apart from that.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -641,10 +641,6 @@ void CKeybindManager::changeworkspace(std::string args) {
     int         workspaceToChangeTo = 0;
     std::string workspaceName       = "";
 
-    // Flag needed so that the previous workspace is not recorded when switching
-    // to a previous workspace.
-    bool isSwitchingToPrevious = false;
-
     bool internal = false;
 
     if (args.find("[internal]") == 0) {
@@ -669,10 +665,6 @@ void CKeybindManager::changeworkspace(std::string args) {
             else
                 workspaceName = std::to_string(workspaceToChangeTo);
 
-            isSwitchingToPrevious = true;
-
-            // If the previous workspace ID isn't reset, cycles can form when continually going
-            // to the previous workspace again and again.
             static auto* const PALLOWWORKSPACECYCLES = &g_pConfigManager->getConfigValuePtr("binds:allow_workspace_cycles")->intValue;
             if (!*PALLOWWORKSPACECYCLES)
                 PCURRENTWORKSPACE->m_iPrevWorkspaceID = -1;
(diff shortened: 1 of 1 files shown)


vaxerski:
one thing I'd like you to check is all the dispatchers, especially those that move windows around. stuff like movetoworkspace prev etc

devil-may-c0de:
I was unable to use movetoworkspace prev (or movetoworkspace previous), but that doesn't seem to work on the normal source package either. otherwise, moving the windows between workspaces works fine.

devil-may-c0de:
it appears to change the behavior of allow_workspace_cycles from that documented in the wiki, which may be why the check was there in the first place

vaxerski:
I've not seen really anyone benefitting from this so if there is no resistance I am for changing the wiki entry

devil-may-c0de:
I think it breaks that functionality entirely since it relies on the previous workspace creating chains, which is only possible when it isn't saved when switching to it, although maybe it never functioned properly to begin with.

vaxerski:
do whatever you find adequate. I don't use this and as such you're a better person to judge the best behavior.

Lmk when it's ready.

devil-may-c0de:
I'm going to attatch isSwitchingToPrevious  to allow_workspace_cycles variable instead of this patch. closing.

Issue #1390, opened 2023-01-16
Defullscreening ignores previous maximizing


scorpion-26:
*Everything tested on 589046ecf74965a8e22a9891290f4b289c60a334*

Steps to reproduce
1. Maximize application (e.g. firefox)
2. Send a fullscreen request through the application (e.g. fullscreening a video)
3. Send a defullscreen request through the application (e.g. exiting the video)

Outcome: 
The application window is now back to its normal, unmaximized state.

This behavior feels a bit off to me, since I expect the window to return to its state before the fullscreen was requested (i.e. maximized), and not always to the normal state.

If you also think this is unwanted behavior, I can open a PR with the fixes I made in my fork, if you want.

vaxerski:
you can make a PR to that but it is expected behavior, although maybe not the most intuitive.

PR #1393, opened 2023-01-16
Honor previous maximize on defullscreen request

scorpion-26:
When defullscreening a window by the apps' request, we would return the window to normal mode, even if the window was previously maximized. Now a defullscreening request honors the previous maximized state.

See #1390

It is ready to merge. Since the issue it is fixing, is expected behaviour, the behaviour could probably be controlled by a configuration, which could still be done(I'll have to look into that).

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -767,11 +767,23 @@ void Events::listener_fullscreenWindow(void* owner, void* data) {
         if (REQUESTED->fullscreen && PWINDOW->m_bIsFullscreen) {
             const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(PWINDOW->m_iWorkspaceID);
             if (PWORKSPACE->m_efFullscreenMode != FULLSCREEN_FULL) {
+                // Store that we were maximized
+                PWINDOW->m_bWasMaximized = true;
                 g_pCompositor->setWindowFullscreen(PWINDOW, false, FULLSCREEN_MAXIMIZED);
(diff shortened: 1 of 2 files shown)


src/events/Windows.cpp
```diff
@@ -767,11 +767,23 @@ void Events::listener_fullscreenWindow(void* owner, void* data) {
         if (REQUESTED->fullscreen && PWINDOW->m_bIsFullscreen) {
             const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(PWINDOW->m_iWorkspaceID);
             if (PWORKSPACE->m_efFullscreenMode != FULLSCREEN_FULL) {
+                // Store that we were maximized
+                PWINDOW->m_bWasMaximized = true;
```
vaxerski, line 771: why not plop setting of thise into CCompositor::setWindowFullscreen?

scorpion-26: I'm pretty sure, the only way to trigger a window to switch from maximize -> fullscreen is by the application itself. Trying to fullscreen with e.g. a hyprland keybind just switches back to normal. 
Besides that, then I should probably also put the switching code (from and to fullscreen/maximize) into setWindowFullscreen(Then manual maximize -> fullscreen would work the same way). The problem here is, that many places in the codebase assume, that you cannot switch from one fullscreen mode to the other(This is bypassed here). I'd have to disable all these checks for things to work correctly again (I actually did that before, code is in fullscreen-fixes branch of my fork), which isn't ideal.

vaxerski: ah yeah right. forgor :skull:

PR #1399, opened 2023-01-18
nix: react to another stdenv change

colemickens:

Similar to  #1296. Seems I caught nixpkgs in the middle of a change, and this is the other part of it.

I've been running this for a while, just remembered I needed to upstream and drop my fork.

Should be a straight-forward merge. Thanks!


fufexan:
Thanks!

Issue #1015, opened 2022-11-13
When group focused, return to last focused window in that group


OctarineSourcerer:
I'd like for there to be an option to change which window gets selected when moving back to a previously focused group.

Take this example:
![FocusDemo](https://user-images.githubusercontent.com/6842867/201544703-4ac0ff1a-d647-4ef9-89f4-c40f915b93de.png)

With current behaviour, starting at the big window and focusing right, down, left, right would select the _top-right_ window. Having just come from bspwm and i3, this surprised me; I'd have expected it to have selected the _bottom-right_ window, as that was the window on the right of the big one that was last focused.

As far as I can tell, I can't see a way of setting this in the docs. I'd perhaps call it `focus_prefer_recent` or something?

vaxerski:
yeah the current system.... I don't even remember how it works but iirc prefers the oldest window.

OctarineSourcerer:
> should the system prefer _last_ over _longer shared border_?

I think so: last indicates what the user most recently used, so is more likely to be the intended target when going back again. 

If that'd cause divides between opinions, we could always do an option such as `preferred_focus_window` with options like `recent`, `biggest_shared` or `oldest`

vaxerski:
added in 94b7b6b58475a0e802e390591f95e04c70aa676d

`binds:focus_preferred_method` can be either 0 for history (default) or 1 for longest edge

PR #1402, opened 2023-01-19
feat: add focus to urgent or last window

cyrinux:

Following the PR https://github.com/hyprwm/Hyprland/pull/1379, with @maximbaz we try to implement the dispatcher to jump to urgent window, or last window, if no urgent window - we also need to jump between workspace if needed.

The code to jump to the urgent window works, but we are not able to keep track of the PREVIOUS window.
A urgent window come, I see the workspace blinking in my waybar, I trigger `focusurgentorlastwindow` keybind which lead me to well jump on the urgent one, but then, if I trigger again the keybind, sadly, I can't go back to the previous one to continue my work.

We together try different way, but everytime we loose the track of this previous window.

We try to set `g_pCompositor->m_pPrevWindow;` from `src/managers/XWaylandManager.cpp` in `activate()`, or from `switchToWindow()` in `src/managers/KeybindManager.cpp` but looks like, we always loose the trace of the last window where the user was, before jumping to urgent one. 

Any idea of what is our problem, what make us loose the track of the previous window ?

ps: need `focus_on_activate = false` to reproduce.

No, @vaxerski we need help 🙏

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1084,6 +1084,15 @@ int CCompositor::getWindowsOnWorkspace(const int& id) {
     return no;
 }
 
+CWindow* CCompositor::getUrgentWindow() {
+    for (auto& w : m_vWindows) {
+        if (w->m_bIsMapped && w->m_bIsUrgent)
+            return w.get();
+    }
+
+    return nullptr;
+}
+
 bool CCompositor::hasUrgentWindowOnWorkspace(const int& id) {
     for (auto& w : m_vWindows) {
         if (w->m_iWorkspaceID == id && w->m_bIsMapped && w->m_bIsUrgent)
--- a/src/Compositor.hpp
+++ b/src/Compositor.hpp
@@ -137,6 +137,7 @@ class CCompositor {
     void           sanityCheckWorkspaces();
     void           updateWorkspaceWindowDecos(const int&);
     int            getWindowsOnWorkspace(const int&);
+    CWindow*       getUrgentWindow();
     bool           hasUrgentWindowOnWorkspace(const int&);
     CWindow*       getFirstWindowOnWorkspace(const int&);
     CWindow*       getFullscreenWindowOnWorkspace(const int&);
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -48,7 +48,7 @@ CKeybindManager::CKeybindManager() {
     m_mDispatchers["pin"]                           = pinActive;
     m_mDispatchers["mouse"]                         = mouse;
     m_mDispatchers["bringactivetotop"]              = bringActiveToTop;
-    m_mDispatchers["focusurgentorlastwindow"]       = focusUrgentOrLastWindow;
+    m_mDispatchers["focusurgentorlast"]             = focusUrgentOrLast;
 
     m_tScrollTimer.reset();
 }
@@ -1100,7 +1100,7 @@ void CKeybindManager::moveFocusTo(std::string args) {
     }
 }
 
-void CKeybindManager::focusUrgentOrLastWindow(std::string args) {
+void CKeybindManager::focusUrgentOrLast(std::string args) {
     const auto PWINDOWURGENT = g_pCompositor->getUrgentWindow();
     const auto PWINDOWPREV = g_pCompositor->m_pLastWindow
         ? (g_pCompositor->m_vWindowFocusHistory.size() < 2 ? nullptr : g_pCompositor->m_vWindowFocusHistory[1])
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -48,6 +48,7 @@ CKeybindManager::CKeybindManager() {
     m_mDispatchers["pin"]                           = pinActive;
     m_mDispatchers["mouse"]                         = mouse;
     m_mDispatchers["bringactivetotop"]              = bringActiveToTop;
+    m_mDispatchers["focusurgentorlastwindow"]       = focusUrgentOrLastWindow;
 
     m_tScrollTimer.reset();
 }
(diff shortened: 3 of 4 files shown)


vaxerski:
why track the last window if you can track the last workspace? every workspace has a last window field

cyrinux:
Thanks for your reply, I will explore this 👍🏻

maximbaz:
@vaxerski the reason for tracking previous window is that sometimes we need the focus to jump within the same workspace.

Here's one scenario, which doesn't involve workspace moves at all:

- Start with no workspaces, no windows.
- Open terminal 1, run `sleep 5 && printf '\a'` - this window will get urgent in 5s.
- Open terminal 2.
- Open terminal 3.
- After 5s, terminal 1 is urgent.
- `hyprctl dispatch focusurgentorlastwindow ''` should move your focus to terminal 1.
- `hyprctl dispatch focusurgentorlastwindow ''` should move your focus to terminal 3.

In the absence of any urgent windows, this same dispatcher can also be used as a productivity boost if you simply have lots of windows on the same workspace, and you want to frequently switch between just two of them.

- Start with no workspaces, no windows.
- Open 5 windows.
- Click with a mouse on window 2 and window 5 (thus creating a "connection" between them).
- Now repeatedly executing `hyprctl dispatch focusurgentorlastwindow ''` would toggle between them (perhaps, especially useful if you fullscreen the window).

And for completeness, here's another realistic scenario that we want to achieve with this PR:

- You are currently working in a code editor (currently focused and visible) in workspace 1, and you have a window with docs and a chat app on workspace 2 (and docs were focused last on the workspace 2) .
- You get a desktop notification that you got a message in the chat (window became urgent), and you want to quickly reply to it and get back to coding.
- `hyprctl dispatch focusurgentorlastwindow ''` moves your focus from code editor on workspace 1 to the chat window in workspace 2.
- You type an answer, hit keybinding for `hyprctl dispatch focusurgentorlastwindow ''`, and you are back in the code editor on workspace 1.

And finally, this dispatcher can also be used to gain the functionality of `workspace_back_and_forth`, without having to know a workspace number.

I hope these examples clarify what exactly we want to achieve with one dispatcher / keybinding, and why we seemingly need to track the previous window? Let me know if I can provide more info!

----

I believe we managed to finally achieve the desired behavior, but the tracking of the previous window feels a bit "fragile" (in particular, see the commented block, it took quite a long debugging time to figure that out 😅). If you have any ideas how this could be done less error-prone, or if you have other suggestions, please share! 🙏

vaxerski:
I don't like the `m_pPrevWindow` global var.

Wouldn't a deque with the focus hierarchy per-workspace (in `CWorkspace`) work better? It would also ease doing #1015

maximbaz:
Even if we have a per-workspace deque of focus hierarchy, we still need to know on which workspace was the previously focused window, correct? Would that knowledge still be global state?

vaxerski:
https://github.com/hyprwm/Hyprland/blob/d5913a23acd36eeaa46a99380afe8b2a78492f2e/src/helpers/Workspace.hpp#L26

;)

maximbaz:
I feel like I'm missing something, sorry 😅 How could we distinguish whether the last two focused windows were on the same or different workspaces?

maximbaz:
Say on workspace 1 we have code and docs, and on workspace 2 we have the chat. 

I switch between docs and code, so this dispatcher moves me within the same workspace 1. Then I get a message in chat, which makes it urgent, so next time I trigger dispatcher, I want to go to the workspace 2 into the chat, and then next time I am moved back to workspace 1.

With the previous workspace ID and a deque of focused windows per workspace, I feel like it's not enough info to determine where the dispatcher should move the focus to... Right?...

maximbaz:
I'm sorry I can't understand how we can know whether the focus should go to "prev window on currentworkspace" vs "last window on currentworkspace->m_iPrevWorkspaceID". Do you perhaps mean, that we could store a boolean "prevWindowWasAlsoOnThisWorkspace" in the each of the workspace instances, and keep all those booleans in sync?

vaxerski:
jfc

vaxerski:
done 

now let's reconsider the awful dispatcher name...

vaxerski:
I assume that on a focus change after the dispatcher has been used it should be reset and act as a noop right?

eg. focus 3 urgent 1
use dispatcher n times 
manually switch to 2
now dispatcher is noop

right?

maximbaz:
Not in the way we both envision this feature (which we used to have in sway, via [swayr](https://sr.ht/~tsdh/swayr/)). A big part of the workflow of using such a dispatcher (for both of us, at least) is being able to continuously use it to toggle between two most recently used windows (say docs and code), and IF there's an urgent window, then it takes priority and next dispatcher call would take us there. This way, in some sense, it's just one single keyboard shortcut that takes you to wherever you need to go 😄

vaxerski:
odd that you are considering a dispatcher in general, an "urgent" event in IPC (socket2) would have allowed you to make a script pretty easily, no?

maximbaz:
I think it is possible to make a script, yes. We'd need to track which windows became urgent, which windows were focused (if an urgent window was focused, then it's no longer urgent; + the currently focused window cannot become urgent), which windows were closed (kinda like "not mapped" in C++ code), and probably a lot more caveats that need to be inferred from the events that flow by the socket 😄 But the complexity of such script should not be an argument for writing it directly in Hyprland, if you don't feel like having this feature, let's not have it 😉

vaxerski:
having it is fine, just surprised you didnt take the other route. You don't need to track closed windows, you didnt do it here either.

Anyways, I'll fix this in a sec

vaxerski:
No, wait. That still reqs me to track the window history. Ugh. Sec.

vaxerski:
done. Rebase against master and you now have a focus history vec. Use that to make your dispatcher. Should be trivial.

maximbaz:
YES! Thanks, this is much better! 
I just need your help with one bug, and then we can choose a less awful name for the dispatcher 😁

Here's how to repro:

1. Start from scratch, on `workspace1` open `terminal1`, run `sleep 5 && printf '\a'`, open `terminal2`, go to `workspace2` and open `terminal3`.
2. After 5s, once `terminal1` is urgent, trigger the dispatcher -> `terminal1` correctly gets focused.
3. Now trigger dispatcher again -> `terminal2` gets incorrectly focused, instead of `terminal3`.

I suspect the bug happens because `m_vWindowFocusHistory` gets messed up when `CCompositor::focusWindow()` gets called on a window from another workspace that wasn't the last focused one.

More concretely:

On step 2 above, the dispatcher calls `focusWindow(terminal1)` [\*]. The code [will trigger `changeworkspace(workspace1)`](https://github.com/hyprwm/Hyprland/blob/60b880d9312574c3115d973fd532f2e15d1eafff/src/Compositor.cpp#L810-L811), as `workspace1` needs to become visible first. That function will in turn [execute `focusWindow(terminal2)`](https://github.com/hyprwm/Hyprland/blob/18330dec4e59529c396d3cdadb32fc9bd763f6b2/src/managers/KeybindManager.cpp#L788), as `terminal2` was the last focused window on `workspace1` after step 1. And as `focusWindow(terminal2)` gets called, it would of course [place `terminal2` on top of the history](https://github.com/hyprwm/Hyprland/blob/60b880d9312574c3115d973fd532f2e15d1eafff/src/Compositor.cpp#L869-L870). Then, after `workspace1` is finally visible, the initial `focusWindow` [\*] execution continues, and [places terminal1 on top of the history](https://github.com/hyprwm/Hyprland/blob/60b880d9312574c3115d973fd532f2e15d1eafff/src/Compositor.cpp#L869-L870). So in this example, the history becomes (`terminal1`, `terminal2`, `terminal3`) instead of (`terminal1`, `terminal3`, `terminal2`).

Do you agree with the assessment, and if so, what do you think is the best way to fix it?

vaxerski:
rebase against head and it should be fixed

maximbaz:
Amazing, thanks once again, the current code has precisely the behavior that we wanted in all scenarios! :partying_face: 

As functionality is complete, do you want me to make some aesthetic changes? Besides picking another name for the dispatcher, the `switchToWindow` is currently more or less copy-pasted three times in the same file, we could try to refactor the code and for example extract it in its own method. We can also do it in another PR, or you are more than welcome to push changes. You decide! :rocket:

vaxerski:
the name is awful. What about just focusurgentorlast, dropping the window?

Copy-pasting switchtowindow yeah we should do smth about it but thats for another pr

maximbaz:
Agreed :+1:

vaxerski (approved):
lgtm update the wiki pr with the name

cyrinux:
I can do the easy job for the wiki asap, thanks guy! ❤️

PR #1448, opened 2023-01-29
Unbreak VT switch on FreeBSD

jbeich:
[linsysfs(5)](https://man.freebsd.org/linsysfs/5) is not mounted by default and doesn't support `/sys/devices/virtual/tty/tty0/active`. Switch to ioctl approach which is similar between Linux and BSDs.


vaxerski (approved):
lgtm

PR #1493, opened 2023-02-05
Update blurriness of layersurfaces after hyprctl keyword blurls

eriedaberrie:
It makes `hyprctl keyword blurls` update the blurriness of existing layer surfaces instead of needing them to close and reopen.

I'm not super confident in my understanding of the rendering stack, if there's something I'm neglecting to update or if I'm doing something wrong let me know.

It seems to be fine working for me.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -932,17 +932,16 @@ void CConfigManager::updateBlurredLS(const std::string& name, const bool forceBl
     }
 }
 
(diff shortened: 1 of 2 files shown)


vaxerski:
why are we making it dependent on `dynamic`?

User changes blurls via keyword: changes
User changes blurls via config + save: no change

??

eriedaberrie:
I was worried that something bad might happen if it ran on the initial config loading but in hindsight that doesn't make any sense.

vaxerski (approved):
Nice. Thank you!

PR #1514, opened 2023-02-08
Fix CMake minimum required version

bloodandcoffee:
The CMakeLists.txt file uses string(JSON), which was not introduced until 3.19.

I have attempted to build on 3.18 (Debian Bullseye), it does not work.

https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5159


vaxerski:
thanks!

PR #1517, opened 2023-02-10
Only use true/false in default config to reduce confusion

dbeley:

Hello, 

I think it would be a good idea to standardize the format of all the boolean values in the example configuration file in order to always use `true`/`false`.

I think the different `true`/`false`/`on`/`off`/`yes`/`no` leads to more confusion for the new users even if the values in themselves are perfectly valid.

It is ready if you think the changes are a good idea :smile:

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -30,7 +30,7 @@ input {
     follow_mouse = 1
 
     touchpad {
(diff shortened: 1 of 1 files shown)


vaxerski:
fair, thanks!

PR #1522, opened 2023-02-10
Added support for layoutmsg params and added the param 'newfocus' for 'swapwithmaster' and 'focusmaster'

### LINKED ISSUE #674: Support 'promote' focused window to master pane
In XMonad contrib there's a features called [Promote](https://hackage.haskell.org/package/xmonad-contrib-0.17.1/docs/XMonad-Actions-Promote.html) which I often use(d).

Citing the description:
> Moves the focused window to the master pane. All other windows retain their order. If focus is in the master, swap it with the next window in the stack. Focus stays in the master.

I think this feature could fit well in Hyprland

Philipp-M:

It adds the possibility to add parameters to the `layoutmsg` dispatcher for the Master layout.
It also adds the parameter `newfocus` for the `layoutmsg` dispatchers `swapwithmaster` and `focusmaster`,
which can have the following values for `swapwithmaster`:
* `master` - focus at the new master after swap
* `child` - focus the new child after swap
* `auto` - keep the focus on the previously focused window (default and also current behavior)

For `focusmaster` it can have the following values:
* `master` - focus the master, or keep the focus at master if it is already focused
* `auto` - swap the focus with the first child, if the current focus was master, otherwise focus master (default and also current behavior)

Fixes #674 with for example the following binding:
```
bind = $mod,P,layoutmsg,swapwithmaster master
```

The PR itself doesn't introduce any breaking compatibility.

I think `CVarList` should have `size() == 0` when the given parameter for the constructor is an empty string (while it is `1` and the first parameter is the empty string).

I haven't changed this behavior in fear that anything breaks that depends on the current behavior.

I'm working around it with a check like this right now: `CVarList::size() > 0 && vars[0] != ""`.

Yes it's ready for merge, though we might need to update the hyprland-wiki as well with the new Master layout layoutmsg para, should I add a PR there?

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -747,7 +747,7 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
 
     CVarList vars(message, 0, ' ');
 
-    if (vars.size() < 1 || vars[0] == "") {
+    if (vars.size() < 1 || vars[0].empty()) {
         Debug::log(ERR, "layoutmsg called without params");
         return 0;
     }
@@ -773,10 +773,10 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
         // * child - keep the focus at the new child
         // * auto (default) - swap the focus (keep the focus of the previously selected window)
         if (PMASTER->pWindow != PWINDOW) {
-            auto newChild = PMASTER->pWindow;
+            const auto NEWCHILD = PMASTER->pWindow;
             switchWindows(PWINDOW, PMASTER->pWindow);
             if (vars.size() >= 2 && vars[1] == "child")
-                switchToWindow(newChild);
+                switchToWindow(NEWCHILD);
             else // default switch to new master
                 switchToWindow(PMASTER->pWindow);
         } else {
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -745,7 +745,7 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
         g_pCompositor->warpCursorTo(PWINDOWTOCHANGETO->m_vRealPosition.goalv() + PWINDOWTOCHANGETO->m_vRealSize.goalv() / 2.f);
     };
 
-    CVarList vars(message);
+    CVarList vars(message, 0, ' ');
 
     if (vars.size() < 1 || vars[0] == "") {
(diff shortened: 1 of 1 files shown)


vaxerski:
I'd rather this being separated with ` ` instead of `,` to avoid bugs with parsing.

a la `swapwithmaster master`

Yes, make a wiki pr plox.

Philipp-M:
Changed it to `' '`

But one thing that might be an unwanted effect, is that multiple spaces count as multiple parameters (empty strings again), I doubt that this should be intended behavior?

> Yes, make a wiki pr plox.

Alright doing that now

vaxerski:
> But one thing that might be an unwanted effect, is that multiple spaces count as multiple parameters (empty strings again), I doubt that this should be intended behavior?

well why should something like `swapwithmaster                    master` be valid anyways...?

Philipp-M:
Well just wanted to point that out, if people accidentally use multiple spaces, but I explicitly added a comment about that in the docs right now, I don't have a problem with that.

src/layout/MasterLayout.cpp
```diff
@@ -745,7 +745,16 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
         g_pCompositor->warpCursorTo(PWINDOWTOCHANGETO->m_vRealPosition.goalv() + PWINDOWTOCHANGETO->m_vRealSize.goalv() / 2.f);
     };
 
-    if (message == "swapwithmaster") {
+    CVarList vars(message, 0, ' ');
+
+    if (vars.size() < 1 || vars[0] == "") {
```
vaxerski, line 750: `vars[0].empty()`

src/layout/MasterLayout.cpp
```diff
@@ -759,21 +768,34 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
         if (!PMASTER)
             return 0;
 
+        // first message argument can have the following values:
+        // * master - keep the focus at the new master
+        // * child - keep the focus at the new child
+        // * auto (default) - swap the focus (keep the focus of the previously selected window)
         if (PMASTER->pWindow != PWINDOW) {
+            auto newChild = PMASTER->pWindow;
```
vaxerski, line 776: can be constified. Remember that const vars are all caps, so `NEWCHILD`

vaxerski (changes requested):
some minor code style changes

vaxerski (approved):

vaxerski:
tanks!

PR #1545, opened 2023-02-13
focusCurrentOrLast dispatcher

normaltaro:
+ It adds a `focusCurrentOrLast` dispatcher, which is a slight modification of `focusUrgentOrLast`. 
+ This function allows users to switch back and forth (toggle) between the currently focused and the previously focused window _across workspaces_ or within the same workspace. 
+ It is similar to `dwl`'s `view` function with `{0}` argument (in the config).
+ Users can bind this dispatcher to `Mod+Tab` or to their preference to quickly toggle between two windows.

The entire function is a copy of `focusUrgentOrLast`. I have modified 5 lines in total (3 lines of the function).

It is very simple so it should be good.


vaxerski (approved):
yeah. I should get rid of the local auto switchToWindow in the future...

aloispichler:
Would be nice to have `focusBack` which goes back in history (bounded to alt+tab), 
and  `focusForth` in the reverse direction (bounded to alt+shift+tab)
instead of toggling only current or last.

PR #1560, opened 2023-02-16
Add vrr to hyprctl monitor

scorpion-26:
Since the (re)addition of vrr in 127e80692f78b0d91450194fcb4ce0009c2a98a4, it is very useful to know, whether a given monitor has vrr enabled or not, since not all monitors support it. Now hyprctl reports, whether a monitor has vrr enabled or not.

I'm unsure, whether my approach is the right way to check for vrr. Though through my testing with both a vrr monitor and a non-vrr monitor, it seems to work properly.

Ready for merging.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -60,13 +60,13 @@ std::string monitorsRequest(HyprCtl::eHyprCtlOutputFormat format) {
         for (auto& m : g_pCompositor->m_vMonitors) {
             result += getFormat("Monitor %s (ID %i):\n\t%ix%i@%f at %ix%i\n\tdescription: %s\n\tmake: %s\n\tmodel: %s\n\tserial: %s\n\tactive workspace: %i (%s)\n\treserved: %i "
                                 "%i %i %i\n\tscale: %.2f\n\ttransform: "
-                                "%i\n\tfocused: %s\n\tdpmsStatus: %i\n\tvariable refresh rate: %s\n\n",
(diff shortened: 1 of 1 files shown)


src/debug/HyprCtl.cpp
```diff
@@ -59,12 +60,13 @@ std::string monitorsRequest(HyprCtl::eHyprCtlOutputFormat format) {
         for (auto& m : g_pCompositor->m_vMonitors) {
             result += getFormat("Monitor %s (ID %i):\n\t%ix%i@%f at %ix%i\n\tdescription: %s\n\tmake: %s\n\tmodel: %s\n\tserial: %s\n\tactive workspace: %i (%s)\n\treserved: %i "
                                 "%i %i %i\n\tscale: %.2f\n\ttransform: "
-                                "%i\n\tfocused: %s\n\tdpmsStatus: %i\n\n",
+                                "%i\n\tfocused: %s\n\tdpmsStatus: %i\n\tvariable refresh rate: %s\n\n",
```
vaxerski, line 63: two notes: no spaces in name (vrr better) and it should be `0/1` and not `yes/no`

vaxerski (changes requested):

src/debug/HyprCtl.cpp
scorpion-26: Done in 9bcfb5079a404276d43df6792a559efb2938b4bc

vaxerski:
why do you use "1" instead of %i with 1 and 0? that way it's (int)w->the_prop

vaxerski:
nvm u fixed it

vaxerski (approved):
lgtm tanks

PR #1604, opened 2023-02-21
update license on main branch

o69mar:


vaxerski:
question still stands, shouldnt that be 2022-2023

o69mar:
> question still stands, shouldnt that be 2022-2023

fixed it

PR #1580, opened 2023-02-19
Group/Tab Rework

vaxerski:
Reworks how groups work internally, making them independent of layouts, allows for floating them, etc.

Fixes: #1576 #770 
Closes: #593 (impossible to achieve, superseded)

Breaking changes:
`dwindle:col.group...` -> `general:col.group...`

Behavior changes: 
 - groups are now always created as singular (they don't "eat" neighbors)

PR #1627, opened 2023-02-24
Fix changegroupactive dispatcher

### LINKED ISSUE #1594: changegroupactive dispatcher's param no longer works
The group/tab rework has changed the `changegroupactive` dispatcher to no longer use its parameter, instead always switching to the next tab. This means there is no way to go to the previous tab anymore.

opekope2:

This commit re-adds the option to specify direction (b - backward, anything else - forward), as it was before #1580.
Fixes #1594.

If the active window is head, it automatically switches to the tail window. Otherwise, it starts finding the previous window at the group head instead of the next window. This way, the average time to find the previous window is less than the number of windows.

Yes.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1321,7 +1321,25 @@ void CKeybindManager::changeGroupActive(std::string args) {
     if (PWINDOW->m_sGroupData.pNextWindow == PWINDOW)
         return;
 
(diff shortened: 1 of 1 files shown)


vaxerski:
wtf

vaxerski:
1089e858b49975a4b9f6e49d377726ee4f164f96

PR #1640, opened 2023-02-26
movefocus: Set new workspace as active when focusing new monitor

RyanDwyer:
Prior to this patch, using `movefocus` to change focus to a different monitor wouldn't set the new workspace as active, which meant that status bars were not notified of the workspace change. They would continue to show the previous monitor/workspace as active.

No.

It's ready.


vaxerski (approved):
thanks!

Issue #1250, opened 2022-12-19
Add CenterMaster layout for Ultrawide monitors


etrigan63:
When you have the time, please add a CenterMaster layout for Ultrawide monitors. The layout behaves as follows:
Single window = full screen
Two windows = Split 50/50 Master on the left
Three or more windows = Master in center 50% of screen, stack on left and right using 25% of screen each. As more windows are added to the stack, they appear round-robin in the left and right stacks.

Here is a sample from `dwm`
![image](https://user-images.githubusercontent.com/7347264/208364536-38c903fa-2f5a-4257-bdd4-67cf7b374ad5.png)

proycon:
Just some thoughts: Implementation-wise, I think this could be implemented in the existing master layout by expanding the `orientation` functionality I implemented in https://github.com/hyprwm/Hyprland/pull/1202. Looking there might serve as a good starting point in case anybody wants to give this a try.

Extra orientations could for example be:

* `orientationcenterorleft` - master in horizontal center or on the left in case of two windows, as suggested
* `orientationcenterorright` - master in horizontal center or on the right in case of two windows

And then there's the question if people want something like this in a vertical layout (surely  there must be some crazy people who turn ultrawide monitors in portrait mode? :) ), if that's the case:

* `orientationcenterortop` - master in vertical center or at the top in case of two windows
* `orientationcenterorbottom` - master in vertical center or at the bottom in case of two windows

I'd be a bit careful of feature creep and not suggest the latter two unless there's actual demand for it. @etrigans63's use case on a horizontal wide monitor does make sense to me. (I had also thought of it when doing #1202 but didn't really have a use for it myself)

etrigan63:
I have no issues with those suggestions, @proycon. My c++ is a bit rusty (no pun intended), but I will take a look at that request.

etrigan63:
My skills are rustier than I thought.

ecocode:
With an ultrawide monitor this layout is really awesome

etrigan63:
For clarification, new windows added to the layout are always the latest master with former master being pushed to the left or right stacks in a round-robin fashion.

PR #1642, opened 2023-02-26
Added center orientation to master layout

sharkusk:

This adds support for a new orientation for master layout.  It is very close to the request in #1250 .

For my workflow I prefer my main work window to be centered.  So, this implementation centers the master window immediately when a single slave is created.  This may be a bit awkward for some folks as this leaves empty space on the right side.  If there is a strong preference I can change this to split into halves when only two windows are open.

Only other issue I found is that the mouse resize operation is a bit wonky when in this mode.

The following was added to the master layout config:
 
- orientation: added "center"
- always_center_master: bool - only used for center orientation - if true, while in center orientation the master window will always occupy the center of the screen, otherwise center orientation behaves like left orientation until three windows are open
- layoutmsg: orientationcenter

See above.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -114,6 +114,7 @@ void CConfigManager::setDefaultVars() {
 
     configValues["master:special_scale_factor"].floatValue = 0.8f;
     configValues["master:new_is_master"].intValue          = 1;
+    configValues["master:always_center_master"].intValue   = 0;
     configValues["master:new_on_top"].intValue             = 0;
     configValues["master:no_gaps_when_only"].intValue      = 0;
     configValues["master:orientation"].strValue            = "left";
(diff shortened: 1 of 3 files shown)


vaxerski:
> I can change this to split into halves when only two windows are open.

you can make this a config option in the `master:` cat.

sharkusk:
> > I can change this to split into halves when only two windows are open.
> 
> 
> 
> you can make this a config option in the `master:` cat.

Good idea. I will take a look today and update the pull request.

sharkusk:
Hi @vaxerski, your suggestion has been added.

vaxerski (approved):
Thanks. I assume you've tested it. Can you make a wiki MR as well with the docs for your new config var plox

then I'll merge both

vaxerski:
Thanks!

sharkusk:
@vaxerski You can probably close #1250 as well.

PR #1658, opened 2023-02-28
Add option to disable/enable mouse window dragging animations

DashieTM:
Add a configuration flag for windows being animated when dragged with the mouse.
This doesn't replace animate_manual_resizes, instead it is an addition.

wiki entry: https://github.com/hyprwm/hyprland-wiki/pull/141

Users would need to update their config to keep the floating window animations.

ready to merge.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -272,7 +272,7 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
     const auto DELTA     = Vector2D(mousePos.x - m_vBeginDragXY.x, mousePos.y - m_vBeginDragXY.y);
     const auto TICKDELTA = Vector2D(mousePos.x - m_vLastDragXY.x, mousePos.y - m_vLastDragXY.y);
 
-    const auto PANIMATE = &g_pConfigManager->getConfigValuePtr("misc:animate_manual_resizes")->intValue;
+    const auto PANIMATE = &g_pConfigManager->getConfigValuePtr("misc:animate_mouse_windowdragging")->intValue;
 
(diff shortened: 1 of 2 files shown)


src/layout/IHyprLayout.cpp
```diff
@@ -272,7 +272,7 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
     const auto DELTA     = Vector2D(mousePos.x - m_vBeginDragXY.x, mousePos.y - m_vBeginDragXY.y);
     const auto TICKDELTA = Vector2D(mousePos.x - m_vLastDragXY.x, mousePos.y - m_vLastDragXY.y);
 
-    const auto PANIMATE = &g_pConfigManager->getConfigValuePtr("misc:animate_manual_resizes")->intValue;
+    const auto PANIMATE = &g_pConfigManager->getConfigValuePtr("misc:animate_mouse_windowdragging")->intValue;
```
vaxerski, line 275: what about resizing?

DashieTM: I really wonder what I am doing sometimes

vaxerski: please direct those questions to your therapist and not me :)

vaxerski:
I don't like though what this has created. Ugh. When you drag a tiled window it "snaps". Should be enabled by default TBH.

vaxerski:
it's still IMHO a result of a poor curve choice for windowMove.

DashieTM:
I mean I used the default curve, and it was really laggy. Just like the problem shown in the video. 
From then on I disabled the manual resize animation.
I guess just see it as an additional animation option.

Issue #1661, opened 2023-02-28
Ensure plugin cleanup on exit


Duckonaut:
# Why
Simple plugins that just add visuals or hook onto existing code are fine being dropped unexpectedly, since they are self-contained in the process. More advanced plugins, like a plugin manager I'm working on, potentially interact with outside things like the filesystem.

Being able to clean up on unload is necessary, and currently when shutting down Hyprland ignores unloading plugins.

PR #1662, opened 2023-02-28
Unload plugins on compositor cleanup

Duckonaut:
Unloads plugins (in reverse order of loading) on compositor cleanup i.e. exit.

The current plugins out there don't benefit much from it, and should have clean exits anyway, so it shouldn't break compatibility. As bigger plugins appear I would predict this will be useful enough to warrant being the default, instead of forcing each plugin to hook the exit and do it on its own.

The functionality is there. Potentially it might be useful to be able to mark plugins as "special" and only unload those, but I'm not sure introducing inconsistency like that will do any good really.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -285,6 +285,10 @@ void CCompositor::cleanup() {
 
     m_bIsShuttingDown = true;
 
+    // unload all remaining plugins while the compositor is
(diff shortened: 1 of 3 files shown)


Duckonaut:
Related to #1661

vaxerski (approved):
LGTM

PR #1687, opened 2023-03-03
Fix plugin config breakage

### LINKED ISSUE #1681: Plugin-registered config values never change past initialization
While making plugins I ran into an issue with plugin-registered config values. I can't tell why it would be happening, but they don't seem to ever be read from the config file, staying only at the value defined as the default.

Here's an example of the usage:
1. Definition: https://github.com/Duckonaut/hyprload/blob/96a575dae08ba47124b8533db626669416195099/src/main.cpp#L64
2. Access: https://github.com/Duckonaut/hyprload/blob/96a575dae08ba47124b8533db626669416195099/src/Hyprload.cpp#L31
3. Usage: https://github.com/Duckonaut/hyprload/blob/96a575dae08ba47124b8533db626669416195099/src/HyprloadPlugin.cpp#L90

With the default value set as an empty string, despite having this section in my hyprland.conf, it will always throw the "No hyprland headers" error there.
```
plugin {
    hyprload {
        debug = 0
        quiet = 0
        hyprland_headers=/home/duckonaut/repos/Hyprland
    }
}
```

Duckonaut:
Fixes plugin config value setting

yea

Fixes #1681

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -311,7 +311,7 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
         for (auto& [handle, pMap] : pluginConfigs) {
             auto it = std::find_if(pMap->begin(), pMap->end(), [&](const auto& other) { return other.first == COMMAND; });
             if (it == pMap->end()) {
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
woops

PR #1688, opened 2023-03-03
Use ${PREFIX} rather than /usr directly in Makefile, and force the build to attempt compilation even if the user does not have ${PREFIX}/lib write permissions

Froggo8311:
This uses PREFIX rather than installing to /usr directly, and attempts to build even if the user does not have permission to write to /usr/lib

I built it and it seems to work properly.

It is ready for review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ clear:
 all:
 	make clear
 	make fixwlr
-	cd ./subprojects/wlroots && meson build/ --buildtype=release && ninja -C build/ && cp ./build/libwlroots.so.12032 /usr/lib/ && cd ../..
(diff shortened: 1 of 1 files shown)


vaxerski:
y

primalmotion:
because $PREFIX is not always /usr ? :)

vaxerski:
I mean yeah, but why would anyone want to install somewhere else?

primalmotion:
some people like /usr/local. Also AUR uses a different PREFIX to perform the packaging

vaxerski:
fair enough

vaxerski (approved):
el gee tee em

PR #1697, opened 2023-03-04
Fix clang compilation

scorpion-26:
void* + integral type is not allowed on standard C++ and is a GNU extension. This PR explicitly casts every void* to uint8_t* when used with pointer arithmetic to fix compilation with clang.

Since void doesn't have a size in the C++ spec, one cannot perform pointer arithmetic with it. (https://en.cppreference.com/w/cpp/language/operator_arithmetic#Additive_operators).
According to GNU spec (https://gcc.gnu.org/onlinedocs/gcc/Pointer-Arith.html) void is considered 1 byte, so pointer arithmetic with void* is allowed for GCC. uint8_t is also 1 byte large, so the calculations are the exact same.

Ready for merging

### DIFF
--- a/src/plugins/HookSystem.cpp
+++ b/src/plugins/HookSystem.cpp
@@ -57,7 +57,7 @@ size_t CFunctionHook::probeMinimumJumpSize(void* start, size_t min) {
 
     while (size <= min) {
         // find info about this instruction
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
lgtm thanks

PR #1698, opened 2023-03-04
Use glBlitFramebuffer for filling the wlr FB

Dickby:
Using glBlitFramebuffer for the final copy to wlr's framebuffer should be quite a bit faster.

This is tested with monitor roations and scaling, but further tests should be done before using this.

I made this a draft PR just to show some code that works, and to raise the point that glBlitFramebuffer should be used for this. This is not meant to be the final implementation!


vaxerski:
defo can't be final cuz 
image

so we need to wrap it into ifdefs for legacy renderer.

Dickby:
Good point! 
But it should shurely be used if available!

Dickby:
Do you have a precise way to benchmark yet?

vaxerski:
Check my changes. Screen shaders and GLES2 are blockers for `blit`.

> Do you have a precise way to benchmark yet?

not _precise_, but kind of. Sec.

vaxerski:
interesting, from my tests it's actually made everything significantly worse...?

Dickby:
That's strange! How did you test?

vaxerski:
turned on the debug overlay and noted the timings alongside the var, and pretty much all timings doubled when I switched to this fork.

vaxerski:
https://www.reddit.com/r/opengl/comments/7erxlh/why_is_glblitframebuffer_so_slow/ 

seems to be the same issue. IMHO a passthru shader here would be better.

Dickby:
I'm really surprised by this, i'll try to find some more informations.
From my experience it should be faster.

vaxerski:
b69f40815fb9302526858bbf699bc6942f8bbd3e this has increased performance a bit.

Dickby:
Ok then. At least something good came out of this!
I'll try to look into it anyways.

Issue #1502, opened 2023-02-06
Pass more uniforms to screen shaders


ElSargo:
It would be great if things like the time and maybe even cursor position were available from within the shader.

vaxerski:
no, because it's impossible to track damage when a shader has access to dynamic uniforms

PR #1700, opened 2023-03-04
Optionally expose time to screen shaders

scorpion-26:
This PR adds a way to have access to time in screen shaders. Time is essential for advanced, animated post-process shader effects (i.e. CRT effect, glitch effects, ...). 
Since that requires damage tracking to be disabled, I implemented it as "Pay for what you use". This means time is only added, if it is actually used in the shader; and users that don't use it have no disadvantage. 
If damage tracking is enabled and time is used an error is displayed. 

Implements part of https://github.com/hyprwm/Hyprland/issues/1502

This requires, that "damage_tracking = 0" updates every frame, which seems to be the case.

Ready. Criticisms/suggestions welcome though.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -31,6 +31,8 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() {
     pixman_region32_init(&m_rOriginalDamageRegion);
 
     RASSERT(eglMakeCurrent(wlr_egl_get_display(g_pCompositor->m_sWLREGL), EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT), "Couldn't unset current EGL!");
+
+    m_Timer.reset();
 }
 
 GLuint CHyprOpenGLImpl::createProgram(const std::string& vert, const std::string& frag, bool dynamic) {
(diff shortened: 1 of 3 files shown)


src/render/OpenGL.hpp
```diff
@@ -133,6 +134,7 @@ class CHyprOpenGLImpl {
     bool              m_bApplyFinalShader = false;
 
     CShader           m_sFinalScreenShader;
+    CTimer            m_Timer;
```
vaxerski, line 137: terrible name. I propose `m_tGlobalTime` or sumn

src/render/OpenGL.cpp
```diff
@@ -511,6 +520,14 @@ void CHyprOpenGLImpl::renderTextureInternalWithDamage(const CTexture& tex, wlr_b
     glUniformMatrix3fv(shader->proj, 1, GL_FALSE, glMatrix);
 #endif
     glUniform1i(shader->tex, 0);
+
+    if (usingFinalShader && g_pConfigManager->getInt("debug:damage_tracking") == 0) {
+        glUniform1f(shader->time, m_Timer.getSeconds());
```
vaxerski, line 525: this is not an issue per se - but wouldn't milliseconds with an int make more sense here? Dunno if opengl uses 64 or 32 bit ints though - if 32 then it's not really a good solution imho (overflow after 596h) but if it's 64-bit we can make it have zero inaccuracies.

vaxerski (changes requested):
One comment and one requested change. In general, this looks reasonable to me.

src/render/OpenGL.hpp
scorpion-26: Fixed

src/render/OpenGL.cpp
scorpion-26: Good question. In terms of shader programming, ```float time;``` in seconds would be way less awkward than ```int time;``` in milliseconds and definitely more common. We're limited anyways by the return value of the internal call to ```std::...::duration::count()``` (which is a 32bit long and in milliseconds for ```getSeconds```). ```GLint``` is also typedef'ed to ```int``` on the C++ side.

vaxerski: > We're limited anyways by the return value of the internal call to `std::...::duration::count()` (which is a 32bit long and in milliseconds for getSeconds).

umm...
image

> `GLint` is also typedef'ed to `int` on the C++ side (GLlong doesn't exist and IIRC 64bit integers don't exist for OpenGL).

Right. Forgot.

Well then this is alr.

vaxerski (approved):
lgtm. thanks!

PR #1707, opened 2023-03-05
Buffer over-read in removeBeginEndSpacesTabs()

opsuu:
If a string given to removeBeginEndSpacesTabs() was only tabs/spaces, or had tabs/spaces followed by a comment, it would lead to buffer over-read. 

The reason is that the code `str.length() != 0` , doesn't do anything, since we don't modify the string inside the while loop. Because of this the loop will run forever until `str.length() - countAfter - 1` wraps around and tries to read memory beyond the scope of the string.


vaxerski (approved):
lgtm, thanks

PR #1715, opened 2023-03-06
fix mfact not applying to addmaster

hillyu:
this will fix the issue of newly added mfact setting not being applied to addmaster (i.e. more than one master).


vaxerski (approved):
Thanks!

PR #1723, opened 2023-03-06
Fix plugin argument-less hyprctl calls

Duckonaut:
The current way of internal hyprctl commands adds whitespace at the end of the call. Due to the way `HyprCtl::getReply` handles argument-less commands like "version", the whitespace added by `HyprlandAPI::invokeHyprctlCommand` breaks the hard comparison of eg. `"version" == "version "`. This fixes that by just not adding unnecessary whitespace.

Seems ready to me


vaxerski (approved):
lgtm thanks!

PR #1730, opened 2023-03-07
support for touch events in keybinds

horriblename:

allow binding to touch events for keybinds and also mouse binds

```
bindm=SUPER,touch,movewindow
```

No, it kinda works but sometimes doesn't... until I put it to sleep then turn it back on...?


horriblename:
I no longer need this

PR #1778, opened 2023-03-13
[WIP] text-input-v1 support

vaxerski:
Adds support for tiv1

Doesn't seem to work yet due to chromium being weird...?

@fufexan for meson / nix (protocol added)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,5 +140,6 @@ target_link_libraries(Hyprland
         ${CMAKE_SOURCE_DIR}/wlr-foreign-toplevel-management-unstable-v1-protocol.o
         ${CMAKE_SOURCE_DIR}/hyprland-toplevel-export-v1-protocol.o
         ${CMAKE_SOURCE_DIR}/fractional-scale-v1-protocol.o
+        ${CMAKE_SOURCE_DIR}/text-input-v1-protocol.o
         ${CMAKE_SOURCE_DIR}/subprojects/udis86/build/libudis86/liblibudis86.a
 )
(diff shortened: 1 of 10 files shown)


vaxerski:
hm, so far so good, but something's off. Chromium insta-crashes on any text input from IME lol

vaxerski:
Chromium doesn't crash most of the times now, oddly I can type entire sentences but the moment I try to type ！ after a 乾杯 it dies.

vaxerski:
awright fixed crashes

vaxerski:
@dragove can you test this as the original reporter for the issue

I-Want-ToBelieve:
image

q234rty:
Tested this a bit on arch:

```
electron19 --enable-wayland-ime --ozone-platform=wayland https://github.com # doesn't work, same symptom as the comment above
electron20 --enable-wayland-ime --ozone-platform=wayland https://github.com # ditto
electron21 --enable-wayland-ime --ozone-platform=wayland https://github.com # works
electron22 --enable-wayland-ime --ozone-platform=wayland https://github.com # works
```

VSCode is currently using electron19, so I guess the electron/chromium behavior changed in recent versions of electron/chromium.

q234rty:
However, afaik all of the above works in weston and kwin, so there must be something that could be done by hyprland as well.

I-Want-ToBelieve:
https://github.com/electron/electron/releases/tag/v21.0.0-alpha.1

> Added support for --ozone-platform-hint flag on Linux. https://github.com/electron/electron/pull/34937 (Also in https://github.com/electron/electron/pull/35015, https://github.com/electron/electron/pull/35014)

vaxerski:
> And the positioning of the candidate word panel is always in the upper left corner

Odd, as the client should send the cursor box. If it doesn't, then maybe hl doesn't send some event it expects?

Although if it works with electron >=21 I'd say it's a minor issue.

I'd rather push this (because it generally works) and work on that minor problem as time goes on. Seems stable to me, so no harm in doing that.

Feel free to open another issue for that. 

Thanks for the feedback!

PR #1847, opened 2023-03-21
Added overflow check for blur radius

alyanser:
Added a (temporary) fix to prevent blur radius from overflowing. When the value if decoration::blur_size and/or decoration::blur_passes is large, the expression `*BLUR_SIZE * pow(2, *BLURPASSES)` might return a value that won't fit into an `int` so an overflow occurs when the returned value is implicitly casted to an `int` when it gets passed to `wlr_region_expand`.

`wlr_region_expand` has this assertion which made me find out about the overflow:

    assert(distance >= 0);
This is not supposed to be a permanent (or even temporary) fix, I'm sure someone who has a better understanding might figure out something better as I'm not yet familiar with the codebase at all.

Ready for merging.


vaxerski (approved):
lgtm, thanks!

alyanser:
Your fix just accounts for `blur_passes` but `blur_size` also matters.

    blur_size = 100000000
    blur_passes = 5

That'll cause overflow.

vaxerski:
ah yeah we should cap that too.

vaxerski:
should work now. max is 40 * 1024 = 40960, so way below an overflow

PR #1863, opened 2023-03-23
Introduce render_ahead_of_time

vaxerski:
This PR adds rendering ahead-of-time, akin to sway's `max-render-time`, but a bit better.

`misc:render_ahead_of_time` is the master switch, while `misc:render_ahead_safezone` is an additional safezone around the rendering. Minimum is 1.

While with sway you had to basically "guess" how long your rendering is, with Hyprland, it will automatically average out your last rendertimes and use that as a base. the safezone is additional padding to that.

e.g. if your average rendertime is calculated to be 0.4ms, added safezone that's 1.4ms, floored that's 1ms, and that would be the "sway's" max-render-time.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -86,7 +86,7 @@ void CConfigManager::setDefaultVars() {
     configValues["misc:hide_cursor_on_touch"].intValue         = 1;
     configValues["misc:mouse_move_focuses_monitor"].intValue   = 1;
     configValues["misc:suppress_portal_warnings"].intValue     = 0;
-    configValues["misc:render_ahead_of_time"].intValue         = 1;
+    configValues["misc:render_ahead_of_time"].intValue         = 0;
     configValues["misc:render_ahead_safezone"].intValue        = 1;
 
     configValues["debug:int"].intValue             = 0;
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -86,6 +86,8 @@ void CConfigManager::setDefaultVars() {
     configValues["misc:hide_cursor_on_touch"].intValue         = 1;
     configValues["misc:mouse_move_focuses_monitor"].intValue   = 1;
     configValues["misc:suppress_portal_warnings"].intValue     = 0;
+    configValues["misc:render_ahead_of_time"].intValue         = 1;
+    configValues["misc:render_ahead_safezone"].intValue        = 1;
 
     configValues["debug:int"].intValue             = 0;
     configValues["debug:log_damage"].intValue      = 0;
--- a/src/debug/HyprDebugOverlay.hpp
+++ b/src/debug/HyprDebugOverlay.hpp
@@ -7,6 +7,8 @@
 #include <cairo/cairo.h>
 #include <unordered_map>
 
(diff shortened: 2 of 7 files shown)


FlafyDev:
I have just tried this PR

Before adding `render_ahead_of_time` to the config file: Everything is the same but runs at a very low fps.
Hyprland runs at about 10-30 fps.

Setting `render_ahead_of_time` to true does the same thing as the above.

Setting `render_ahead_of_time` to false makes the screen black.
Keybinds still work so I used SUPER+M to exit Hyprland.

Tested on my laptop with Intel iGPU. Without the external screen.
The iGPU barely does anything at all. It's at under 5% usage in Render/3D.

Using the Nix build:
```
❯ hyprctl version
Hyprland, built from branch  at commit a7e10dc0d0b9d8a05d91c1bf205c16ba6c10178c ().
flags: (if any)
debug
```

commit 463690a27ac9c921d34dad7169a3d2c8cea5b46f runs normally.

FlafyDev:
I tested it some more and by inspecting the overlay fps counter(with vfr disabled), the minimum `render_ahead_safezone` I found that resembled my usual fps is `13`.
Is that a reasonable value to use?

I also noticed that this does not work exactly like sway's `max-render-time`.
I originally wanted this feature because I am experiencing the exact same problem shown here: https://www.reddit.com/r/swaywm/comments/s8gs35/every_game_stuck_30fps_when_fullscreen/.
When I fullscreen Minecraft, it becomes stuck at 30fps in both Hyprland and Sway.
If I set `max-render-time` to anything under around `15` in Sway, Minecraft gets stuck at 60 fps, which is what i am aiming for.
This doesn't seem to be the same for Hyprland's `render_ahead_safezone`. Do you know why?

FlafyDev:
> disable direct scanout. That's unrelated.

I see, i was not aware of this option. Do you think it is worth to add as a section in the wiki's FAQ? Or at least in the nvidia page? "Game lags on fullscreen"


>  the minimum render_ahead_safezone I found that resembled my usual fps is 13.
> Is that a reasonable value to use?
> > No

So would that be a bug? Or is it something wrong on my side

vaxerski:
> So would that be a bug? Or is it something wrong on my side

¯\\\_(ツ)_/¯

works on my side

vaxerski:
I'll merge this cuz it works for me, bugs can be ironed out later. I'll just disable it by default.

JustSimplyKyle:
The safezone value I need is around 3.
But is there a way to set it per monitor? Cause while 3 works for my main, it has some really weird results in my second.
>  the following sentence is with vrr off, with vrr on, it seems to be a little bit more laggy

The fps from what I can tell is lower than 1 if and only if I'm putting my mouse over it.
If I just put the video in my second monitor, it's very smooth. But as soon as I try and move my mouse over it. The entire monitor freezes for about a second

JustSimplyKyle:
The interesting thing is, if I change my main monitor's refersh rate to 60, it also requires a 13 safe zone.(otherwise it has the same issue as my second)
Maybe this has something to do with it?

vaxerski:
it's really buggy on drm for some reason, I'll look into it, but TBH if we get tearing merged this will be useless.

PR #1879, opened 2023-03-25
Update InputManager.cpp

### LINKED ISSUE #1711: Focus changes when dragging window quickly
Quite self explanatory, when dragging a floating window quickly, the window will change focus to windows below for a split second.

https://user-images.githubusercontent.com/15178513/222989622-0bbbe5cf-6f5a-43d6-905b-92b94def9fca.mp4
> Also video seems to appear corrupt on firefox for me, but isn't, works in other browsers

devil-may-c0de:
fixes #1711 by adding a check for dragging when changing focus from moveMouseUnified in InputManager.cpp. This means that whenever a user drags a window, they cannot change focus with their mouse, and the focus stays on its current window unless changed via keybind.

Tested a few times before and after the patch. I was unable to change focus as described in #1711 with my patch, whereas without the patch I could reproduce the issue. slightly deeper testing seems to only produce expected behaviors.

I think this patch is ready for merging.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -303,7 +303,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
         m_pFoundSurfaceToFocus = foundSurface;
     }
 
-    if (pFoundWindow) {
+    if (pFoundWindow && g_pInputManager->currentlyDraggedWindow == nullptr) {
         // change cursor icon if hovering over border
         if (*PRESIZEONBORDER && *PRESIZECURSORICON && !pFoundWindow->m_bIsFullscreen && !pFoundWindow->hasPopupAt(mouseCoords)) {
             setCursorIconOnBorder(pFoundWindow);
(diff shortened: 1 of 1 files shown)


vaxerski (changes requested):
This is pretty bad as you're messing up the pointer coordinates sent to applications.

I'd rather suggest completely bailing out at like like 307 just after
```c
if (pFoundWindow) {
```
since we probably don't need to send pointer updates when dragging anyways.

devil-may-c0de:
ok! made the change.

devil-may-c0de:
i left the old code in i will fix that

vaxerski:
nuh-uh. You are still sending invalid coord events.

```cpp
if (pFoundWindow) {
    if (currently dragged)
        return;
```

there is no point in going further.

devil-may-c0de:
ah i see what you mean!

devil-may-c0de:
returning seems to break focusing. I don't think I'm the right person for this...

vaxerski:
lmao I'll fix it

yavko:
> lmao I'll fix it

Is it fixed?

vaxerski:
ye d6b069458d990d372eaa9a35bb8726605b44a588

PR #1894, opened 2023-03-27
Fix nix build options

oliver-ni:

Some of the build options in the Nix derivative currently don't work. For example, setting `legacyRenderer` to `true` does nothing (it is built with the legacy renderer off), and setting `withSystemd` to `false` does nothing (it is built with systemd enabled). I spent a long time debugging why Hyprland was not running with on Asahi Linux even with the legacy renderer on before I realized that the binary was not being built with legacy renderer.

I am not familiar with Meson and am not sure whether this is the right way to implement this. Please let me know if changes need to be made.

It works in my tests.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -59,12 +59,16 @@ endif
 backtrace_dep = cpp_compiler.find_library('execinfo', required: false)
 systemd_dep = dependency('libsystemd', required: get_option('systemd'))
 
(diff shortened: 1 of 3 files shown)


oliver-ni:
This also seems like the cause for #1352 (same issue I was having).

vaxerski:
@fufexan for nix

PR #1926, opened 2023-03-31
Misc FreeBSD fixes

jbeich:
FreeBSD currently uses `nm` from [ElfToolChain](https://sourceforge.net/projects/elftoolchain/) but plans to [switch to LLVM](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=258872). DragonFly and NetBSD probably still use GNU binutils because it's required by GCC, anyway.

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -6,11 +6,11 @@
 #include <iomanip>
 #include <sstream>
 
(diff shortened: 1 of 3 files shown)


jbeich:
FreeBSD currently uses `nm` from [ElfToolChain](https://sourceforge.net/projects/elftoolchain/) but plans to [switch to LLVM](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=258872). DragonFly and NetBSD probably still use GNU binutils because it's required by GCC, anyway.

vaxerski (approved):
Thanks as usual!

PR #1971, opened 2023-04-07
Fix swiping onto a new workspace with multiple monitors.

awpr:
The previous code could run into issues into the following circumstances:

* The focused monitor is on its rightmost workspace with ID `i`.
* Another monitor has a workspace with ID `i+1`.
* `workspace_swipe_create_new` is enabled.

Then, swiping rightwards attempts to target a new workspace with ID `i+1`: completing the swipe gesture unintentionally focuses that workspace on whichever monitor it's already on while leaving the active monitor in a broken state where it shows no windows but creates new windows on the workspace it was previously on; and cancelling the swipe gesture shifts the entire workspace `i+1` to the right by the width of the active monitor.

By choosing an ID that doesn't exist, this problematic behavior is avoided.  More specifically, it's the smallest ID greater than any existing workspace's ID, because otherwise the new workspace that was seemingly just created to the right of the rightmost workspace could end up somewhere in the middle of the workspace order.

No issues I'm aware of.

Ready for merging as far as I'm concerned.


vaxerski (approved):
lgtm, thanks!

PR #1976, opened 2023-04-07
Better and more secure argument parsing, and code reformatting

nitrogenez:
PR adds a better non-C-style argument parsing based on std::vector and std::container_type::iterator.  

Also, I've reformatted the code a little.  

All changes:
  - Added help() function to display help message
  - Added "-i-am-really-stupid" flag to the help message
  - Changed "root user" term to "superuser privileges"
  - Added user-specified config file path validation
  - Superuser privileges warning now shows up only if the privileges check has found ones
  - std::vector-based argument array
  - std::container_type::iterator-based argument parsing, which is a kind of a standard in modern C++

Didn't find anything unusual on my machine. Compiled with meson & ninja.
Would like to mention that the code is a real mess.
And also that the index-based argument parsing is insecure because of data-breach (out-of-bounds r/w)

Remember polkit?
[CVE-2021-4034](https://nvd.nist.gov/vuln/detail/cve-2021-4034)

I think it's ready. More than that. This patch only affects src/main.cpp, so no conflicts might be possible.

That's all your decision, merge that or not.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -12,18 +12,16 @@
 #include <filesystem>
 
 #ifdef USES_SYSTEMD
-    #include <systemd/sd-daemon.h> // for sd_notify
+#include <systemd/sd-daemon.h> // for sd_notify
 #endif
-
 
 void help() {
     std::cout << "usage: Hyprland [arg [...]].\n";
     std::cout << "\nArguments:\n";
     std::cout << "  --help              -h       - Show this message again\n";
     std::cout << "  --config FILE       -c FILE  - Specify config file to use\n";
-    std::cout << "  --i-am-really-stupid         - Omits root user privileges check (why'd u do that?)\n";
+    std::cout << "  --i-am-really-stupid         - Omits root user privileges check (why would u do that?)\n";
 }
-
 
 int main(int argc, char** argv) {
 
@@ -41,12 +39,11 @@ int main(int argc, char** argv) {
     setenv("MOZ_ENABLE_WAYLAND", "1", 1);
     setenv("XDG_CURRENT_DESKTOP", "Hyprland", 1);
 
-
     // parse some args
-    std::string configPath;
-    bool ignoreSudo = false;
+    std::string              configPath;
+    bool                     ignoreSudo = false;
 
-    std::vector<std::string> args { argv + 1, argv + argc };
+    std::vector<std::string> args{argv + 1, argv + argc};
 
     for (auto it = args.begin(); it != args.end(); it++) {
         if (it->compare("--i-am-really-stupid") == 0 && !ignoreSudo) {
@@ -77,11 +74,10 @@ int main(int argc, char** argv) {
 
     if (!ignoreSudo && Init::isSudo()) {
         std::cerr << "[ ERROR ] Hyprland was launched with superuser priveleges, but the privileges check is not omitted.\n";
-        std::cerr << "          Hint: Use --i-am-really-stupid flag to omit that check.\n";
+        std::cerr << "          Hint: Use the --i-am-really-stupid flag to omit that check.\n";
 
         return 1;
-    }
-    else if (ignoreSudo && Init::isSudo()) {
+    } else if (ignoreSudo && Init::isSudo()) {
         std::cout << "Superuser privileges check is omitted. I hope you know what you're doing.\n";
     }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -17,11 +17,11 @@
 
 
 void help() {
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -17,37 +34,66 @@ int main(int argc, char** argv) {
...
+
+    // Funny thing.
+    // We don't *really* need that check, because in most cases `root` user
+    // doesn't have XDG_RUNTIME_DIR envar set, which causes runtime error.
+
+    // But if set, then yes, that needs to be checked.
+
+    // We don't need to display that message when no superuser privs are found
+    if (!ignoreSudo && Init::isSudo()) {
+        std::cerr << "[ ERROR ] Hyprland was launched with superuser priveleges, but the privileges check is not omitted." << std::endl;
```
vaxerski, line 92: You removed the hint for the user to use `--i-am-really-stupid` to omit it

src/main.cpp
```diff
@@ -17,37 +34,66 @@ int main(int argc, char** argv) {
...
 
+
     // parse some args
     std::string configPath;
-    bool        ignoreSudo = false;
-    for (int i = 1; i < argc; ++i) {
-        if (!strcmp(argv[i], "--i-am-really-stupid"))
+    bool ignoreSudo = false;
+
+    // Opinion: All of this code is a mess. Try to do something with that, folks? Pretty please?
```
vaxerski, line 49: all of what? argument parsing? it is, cuz nobody really cared a lot as we do very few args

src/main.cpp
```diff
@@ -3,11 +3,28 @@
...
+#include <filesystem>
+
 #ifdef USES_SYSTEMD
-#include <systemd/sd-daemon.h> // for sd_notify
+    #include <systemd/sd-daemon.h> // for sd_notify
 #endif
 
+
+void help() {
+    std::cout << "usage: Hyprland [arg [...]]." << std::endl;
```
vaxerski, line 20: why are you using `std::endl`? Why flush after every line...?

vaxerski (changes requested):
```
// Funny thing.
// We don't *really* need that check, because in most cases `root` user
// doesn't have XDG_RUNTIME_DIR envar set, which causes runtime error.
```
The check is in case it does.

src/main.cpp
nitrogenez: If the user needs to omit the check, just see --help and see what arguments are there.

vaxerski: you vastly overestimate the intelligence of an average user.

src/main.cpp
nitrogenez: Good thing to not care about the code, even which is possibly insecure, right?

src/main.cpp
nitrogenez: 'Kay, I'll change that. Not a big deal.

src/main.cpp
vaxerski: You're trying to find a hole in argument parsing, which assumes the attacker already knows you're running Hyprland.

In that case, running `hyprctl dispatch exec "rm -rf ~"` is already miles easier.

Besides, IDEK what your problem here is, as hyprland does not execute any of the args. The CVE you reference does not make sense here.

src/main.cpp
```diff
@@ -17,37 +34,55 @@ int main(int argc, char** argv) {
...
+            continue;
         }
-    } else {
-        std::cout << "Running with ignored root checks, I surely hope you know what you're doing.\n";
-        sleep(1);
+    }
+
+    if (!ignoreSudo && Init::isSudo()) {
+        std::cerr << "[ ERROR ] Hyprland was launched with superuser priveleges, but the privileges check is not omitted.\n";
+        std::cerr << "          Hint: Use --i-am-really-stupid flag to omit that check.\n";
```
vaxerski, line 80: use _the_

src/main.cpp
```diff
@@ -3,11 +3,28 @@
...
+    #include <systemd/sd-daemon.h> // for sd_notify
 #endif
 
+
+void help() {
+    std::cout << "usage: Hyprland [arg [...]].\n";
+    std::cout << "\nArguments:\n";
+    std::cout << "  --help              -h       - Show this message again\n";
+    std::cout << "  --config FILE       -c FILE  - Specify config file to use\n";
+    std::cout << "  --i-am-really-stupid         - Omits root user privileges check (why'd u do that?)\n";
```
vaxerski, line 24: why _would you_ do that

vaxerski (changes requested):
Minor typos. After that, format with the .clang-format and it should be gtg

vaxerski (approved):
lgtm

PR #1983, opened 2023-04-07
Fix crash in CConfigManager::parseKeyword

scorpion-26:
If debug:manual_crash is set on startup, parseKeyword tries to call g_pHyprNotificationOverlay->addNotification, but g_pHyprNotificationOverlay isn't initialized yet (is nullptr)

This commit adds a nullptr check for that.

This fixes the symptom, not the cause. With this fix the notification won't ever be displayed.

Ready, as it is a simple fix


vaxerski:
wouldn't `isFirstLaunch` work better here?

scorpion-26:
AFAIK currently it would work exactly the same. If at some point in the future HyprNotificationOverlay is available at first launch for ConfigManager, the null check would work better however, since then the notification could be shown.

vaxerski:
k

PR #1986, opened 2023-04-07
Added some workspace-specific rules

levnikmyskin:
This PR adds some workspace-specific rules, using the already existing `workspace` config (until now only used for default workspaces for monitors).

More specifically, we are now able to have some custom rules that are applied to workspaces, such as having a different amount of gaps/borders, whether to have rounding, and so on.

The specific syntax on the config would be:
`workspace=monitor,workspace,gaps_in,gaps_out,border_size,drawBorder,doRounding,doDecorate`

where:

1.  `monitor` was the pre-existing argument;
2. `workspace` is a workspace (numbered, named, etc.);
3. `gaps_in` is an int;
4. `gaps_out` is an int;
5. `border_size` is an int;
6. `drawBorder` is a bool, and determines whether borders are drawn or not;
7. `doRounding` is a bool, same as before;
8. `doDecorate` is a bool, same as before.

Windows opened in the specified workspace, will be rendered with those specific rules. Notice, **this takes precedence over** `no_gaps_when_only` (i.e., if `no_gaps_when_only` is `true`, but a workspace rule is found for this specific workspace, `no_gaps_when_only` is ignored).
This basically implements what discussed in #1345.

This should not break any compatibility. I don't know, however, whether we want to keep this syntax for the config file, as having 8 arguments might feel like a bit too much for other people (doesn't look too bad to me tbh).

Theoretically ready for merging, but I guess this might need a little discussion.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1009,41 +1009,52 @@ void CConfigManager::handleBlurLS(const std::string& command, const std::string&
 }
 
 void CConfigManager::handleWorkspaceRules(const std::string& command, const std::string& value) {
-    const auto     MONITOR_DELIM   = value.find_first_of(',');
-    const auto     WORKSPACE_DELIM = value.find_first_of(',', MONITOR_DELIM + 1);
+    // This can either be the monitor or the workspace identifier
+    const auto     FIRST_DELIM = value.find_first_of(',');
 
+    std::string    name        = "";
+    auto           first_ident = removeBeginEndSpacesTabs(value.substr(0, FIRST_DELIM));
+    int            id          = getWorkspaceIDFromString(first_ident, name);
+
+    auto           rules = value.substr(FIRST_DELIM + 1);
     SWorkspaceRule wsRule;
-    wsRule.monitor         = value.substr(0, MONITOR_DELIM);
-    wsRule.workspaceString = value.substr(MONITOR_DELIM + 1, (WORKSPACE_DELIM - MONITOR_DELIM - 1));
-
-    std::string name = "";
-    int         id   = getWorkspaceIDFromString(wsRule.workspaceString, name);
-
+    wsRule.workspaceString = first_ident;
     if (id == INT_MAX) {
-        Debug::log(ERR, "Invalid workspace identifier found: %s", wsRule.workspaceString.c_str());
-        parseError = "Invalid workspace identifier found: " + wsRule.workspaceString;
-        return;
+        // it could be the monitor. If so, second value MUST be
+        // the workspace.
+        const auto WORKSPACE_DELIM = value.find_first_of(',', FIRST_DELIM + 1);
+        auto       wsIdent         = removeBeginEndSpacesTabs(value.substr(FIRST_DELIM + 1, (WORKSPACE_DELIM - FIRST_DELIM - 1)));
+        id                         = getWorkspaceIDFromString(wsIdent, name);
+        if (id == INT_MAX) {
+            Debug::log(ERR, "Invalid workspace identifier found: %s", wsIdent.c_str());
+            parseError = "Invalid workspace identifier found: " + wsIdent;
+            return;
+        }
+        wsRule.monitor         = first_ident;
+        wsRule.workspaceString = wsIdent;
+        rules                  = value.substr(WORKSPACE_DELIM + 1);
     }
-
-    auto        rules = value.substr(WORKSPACE_DELIM + 1);
 
     size_t      pos = 0;
     std::string rule;
     while ((pos = rules.find(',')) != std::string::npos) {
-        rule = rules.substr(0, pos);
-        Debug::log(INFO, "found rule: %s", rule.c_str());
-        if (rule.find("gapsin:") != std::string::npos)
-            wsRule.gapsIn = std::stoi(rule.substr(7));
-        else if (rule.find("gapsout:") != std::string::npos)
-            wsRule.gapsOut = std::stoi(rule.substr(8));
-        else if (rule.find("bordersize:") != std::string::npos)
-            wsRule.borderSize = std::stoi(rule.substr(11));
-        else if (rule.find("border:") != std::string::npos)
-            wsRule.border = configStringToInt(rule.substr(7));
-        else if (rule.find("rounding:") != std::string::npos)
-            wsRule.rounding = configStringToInt(rule.substr(9));
-        else if (rule.find("decorate:") != std::string::npos)
-            wsRule.decorate = configStringToInt(rule.substr(9));
+        rule         = rules.substr(0, pos);
+        size_t delim = std::string::npos;
+        Debug::log(INFO, "found workspacerule: %s", rule.c_str());
+        if ((delim = rule.find("gapsin:")) != std::string::npos)
+            wsRule.gapsIn = std::stoi(rule.substr(delim + 7));
+        else if ((delim = rule.find("gapsout:")) != std::string::npos)
+            wsRule.gapsOut = std::stoi(rule.substr(delim + 8));
+        else if ((delim = rule.find("bordersize:")) != std::string::npos)
+            wsRule.borderSize = std::stoi(rule.substr(delim + 11));
+        else if ((delim = rule.find("border:")) != std::string::npos)
+            wsRule.border = configStringToInt(rule.substr(delim + 7));
+        else if ((delim = rule.find("rounding:")) != std::string::npos)
+            wsRule.rounding = configStringToInt(rule.substr(delim + 9));
+        else if ((delim = rule.find("decorate:")) != std::string::npos)
+            wsRule.decorate = configStringToInt(rule.substr(delim + 9));
+        else if ((delim = rule.find("monitor:")) != std::string::npos)
+            wsRule.monitor = rule.substr(delim + 8);
         rules.erase(0, pos + 1);
     }
 
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -573,7 +573,19 @@ void CConfigManager::handleMonitor(const std::string& command, const std::string
             newrule.transform = (wl_output_transform)std::stoi(ARGS[argno + 1]);
             argno++;
         } else if (ARGS[argno] == "workspace") {
-            m_mDefaultWorkspaces[newrule.name] = ARGS[argno + 1];
+            std::string    name = "";
+            int            wsId = getWorkspaceIDFromString(ARGS[argno + 1], name);
+
+            SWorkspaceRule wsRule;
+            wsRule.monitor         = newrule.name;
+            wsRule.workspaceString = ARGS[argno + 1];
+            wsRule.workspaceName   = name;
+            wsRule.workspaceId     = wsId;
+            wsRule.gapsIn          = configValues["general:gaps_in"].intValue;
+            wsRule.gapsOut         = configValues["general:gaps_out"].intValue;
+            wsRule.borderSize      = configValues["general:border_size"].intValue;
+
+            m_mWorkspaceRules[wsId] = wsRule;
             argno++;
         } else {
             Debug::log(ERR, "Config error: invalid monitor syntax");
@@ -999,10 +1011,23 @@ void CConfigManager::handleBlurLS(const std::string& command, const std::string&
     updateBlurredLS(value, true);
 }
 
-void CConfigManager::handleDefaultWorkspace(const std::string& command, const std::string& value) {
-    const auto ARGS = CVarList(value);
+void CConfigManager::handleWorkspaceRules(const std::string& command, const std::string& value) {
+    const auto     ARGS = CVarList(value);
 
-    m_mDefaultWorkspaces[ARGS[0]] = ARGS[1];
+    std::string    name = "";
+    int            id   = getWorkspaceIDFromString(ARGS[1], name);
+    SWorkspaceRule wsRule;
+    wsRule.monitor         = ARGS[0];
+    wsRule.workspaceString = ARGS[1];
+    wsRule.workspaceName   = name;
+    wsRule.workspaceId     = id;
+    wsRule.gapsIn          = std::stoi(ARGS[2]);
+    wsRule.gapsOut         = std::stoi(ARGS[3]);
+    wsRule.borderSize      = std::stoi(ARGS[4]);
+    wsRule.border          = configStringToInt(ARGS[5]);
+    wsRule.rounding        = configStringToInt(ARGS[6]);
+    wsRule.decorate        = configStringToInt(ARGS[7]);
+    m_mWorkspaceRules[id]  = wsRule;
 }
 
 void CConfigManager::handleSubmap(const std::string& command, const std::string& submap) {
@@ -1130,7 +1155,7 @@ std::string CConfigManager::parseKeyword(const std::string& COMMAND, const std::
     else if (COMMAND == "unbind")
         handleUnbind(COMMAND, VALUE);
     else if (COMMAND == "workspace")
-        handleDefaultWorkspace(COMMAND, VALUE);
+        handleWorkspaceRules(COMMAND, VALUE);
     else if (COMMAND == "windowrule")
         handleWindowRule(COMMAND, VALUE);
     else if (COMMAND == "windowrulev2")
@@ -1302,6 +1327,7 @@ void CConfigManager::loadConfigLoadVars() {
     deviceConfigs.clear();
     m_dBlurLSNamespaces.clear();
     boundWorkspaces.clear();
+    m_mWorkspaceRules.clear();
     setDefaultAnimationVars(); // reset anims
 
     // paths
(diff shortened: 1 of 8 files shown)


vaxerski:
skimmed over the code, two major issues:
 1) the syntax is abhorrent. I'd be for utilizing something along the lines of what `windowrulev2` does.
 2) There absolutely needs to be a mechanism for optionals. You're throwing away hot-reload for props. If a user wants to _just_ change idk gaps_in, they shouldn't effectively lock their workspace's other props. `std::optional` would work here probably

levnikmyskin:
> skimmed over the code, two major issues:
> 
>     1. the syntax is abhorrent. I'd be for utilizing something along the lines of what `windowrulev2` does.
> 
>     2. There absolutely needs to be a mechanism for optionals. You're throwing away hot-reload for props. If a user wants to _just_ change idk gaps_in, they shouldn't effectively lock their workspace's other props. `std::optional` would work here probably

Absolutely agree with point 1, and yes you're right. Didn't think about point 2. Will fix it in the next days :) thanks

levnikmyskin:
Sorry, don't know what happened with github, but it forced-pushed (and then closed) this pull request when I synced the two repos. I'm not familiary with these new github functionalities. If the main branch of my fork is not in sync with the pull request, I'll open a new one, sorry.

Anyways, I changed the syntax to a windowrule-like one (see [here](https://github.com/levnikmyskin/Hyprland/commit/15aca7e0bbd26242f6742111c6b6b2ec5189d953), with a few differences:

1. the first two rules, i.e., the monitor and the workspace identifier still work like before for backward compatibility;
2. the workspace identifier is mandatory, and the parsing will fail if it's not specified (this makes sense imo, as configuring a workspace-specific rule without giving the workspace makes no sense). The first `monitor` rule is also mandatory (can be left blank, as long as we put the comma) for backward compatibility.

The new syntax is thus as follows:
`workspace = monitor, workspace, gapsin:int, gapsout:int, bordersize:int, border:bool, rounding:bool, decorate:bool`

None of the `gapsin`, `gapsout` etc. rules are mandatory.

That said, I have a question regarding `DwindleLayout::applyNodeDataToWindow` function (or its `MasterLayout` equivalent): why do we use pointers to the `PGAPSIN`, `PGAPSOUT`, `PBORDERSIZE` values? If we switched to values, we could simply do 
```cpp
    auto       gapsIn     = WORKSPACE_RULE.gapsIn.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue);
    auto       gapsOut    = WORKSPACE_RULE.gapsOut.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_out")->intValue);
    auto       borderSize = WORKSPACE_RULE.borderSize.value_or(g_pConfigManager->getConfigValuePtr("general:border_size")->intValue);
```
and then avoid the multiple dereferencing further down. However, I don't know if this is done for any specific reason, and that's why I'm asking :)

vaxerski:
monitor should be optional too. (remember about backwards compat tho)

> That said, I have a question regarding DwindleLayout::applyNodeDataToWindow function (or its MasterLayout equivalent): why do we use pointers to the PGAPSIN, PGAPSOUT, PBORDERSIZE values? 

https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/#some-code-faq

levnikmyskin:
> monitor should be optional too. (remember about backwards compat tho)
> 
> > That said, I have a question regarding DwindleLayout::applyNodeDataToWindow function (or its MasterLayout equivalent): why do we use pointers to the PGAPSIN, PGAPSOUT, PBORDERSIZE values?
> 
> https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/#some-code-faq

Damn, I also remember reading that part from the FAQ lol. Btw, maybe I put it badly: users don't have to specify `monitor` value, but the comma is still required (for backwards compatibility), i.e.:
`workspace=,2,gapsin:0`
is perfectly valid. Is this what you meant or do you believe something still needs to change?

levnikmyskin:
right, would the monitor still be the first value specified then?
I mean:
`workspace = DP-1, 2`
is valid but 
`workspace = 2, gapsin:0, DP-1`
is not. I would either do this, or allow the monitor to be either first or second keyword, because it would be a pain to parse otherwise.

vaxerski:
???

```
workspace = DP-1, 2
workspace = 2, gapsin:420
workspace = 2, monitor:DP-1
```

levnikmyskin:
Looks good, I'll add the `monitor:` thing as well

levnikmyskin:
I think it should be fine now. Let me know if there's any other issue we should still address ^^

src/layout/MasterLayout.cpp
```diff
@@ -478,11 +478,18 @@ void CHyprMasterLayout::applyNodeDataToWindow(SMasterNodeData* pNode) {
...
-    const auto PGAPSOUT    = &g_pConfigManager->getConfigValuePtr("general:gaps_out")->intValue;
-
     const auto PWINDOW = pNode->pWindow;
+    // get specific gaps and rules for this workspace,
+    // if user specified them in config
+    const auto WORKSPACE_RULE = g_pConfigManager->getWorkspaceRuleFor(g_pCompositor->getWorkspaceByID(PWINDOW->m_iWorkspaceID));
+
+    auto       gapsIn     = WORKSPACE_RULE.gapsIn.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue);
+    auto       gapsOut    = WORKSPACE_RULE.gapsOut.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_out")->intValue);
+    auto       borderSize = WORKSPACE_RULE.borderSize.value_or(g_pConfigManager->getConfigValuePtr("general:border_size")->intValue);
```
vaxerski, line 488: all three, pointers to config values not static

src/layout/DwindleLayout.cpp
```diff
@@ -108,11 +108,18 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
...
-    const auto PGAPSOUT    = &g_pConfigManager->getConfigValuePtr("general:gaps_out")->intValue;
-
     const auto PWINDOW = pNode->pWindow;
+    // get specific gaps and rules for this workspace,
+    // if user specified them in config
+    const auto WORKSPACE_RULE = g_pConfigManager->getWorkspaceRuleFor(g_pCompositor->getWorkspaceByID(PWINDOW->m_iWorkspaceID));
+
+    auto       gapsIn     = WORKSPACE_RULE.gapsIn.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue);
+    auto       gapsOut    = WORKSPACE_RULE.gapsOut.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_out")->intValue);
+    auto       borderSize = WORKSPACE_RULE.borderSize.value_or(g_pConfigManager->getConfigValuePtr("general:border_size")->intValue);
```
vaxerski, line 118: pointers to config values not static

src/config/ConfigManager.hpp
```diff
@@ -45,6 +46,19 @@ struct SMonitorRule {
     bool                enable10bit = false;
 };
 
+struct SWorkspaceRule {
+    std::string            monitor         = "";
+    std::string            workspaceString = "";
+    std::string            workspaceName   = "";
+    int                    workspaceId     = -1338;
```
vaxerski, line 53: for "unset" we use `-1`. `-1338` is a valid id for a named workspace.

vaxerski (changes requested):

src/layout/DwindleLayout.cpp
levnikmyskin: How would you do that though? I'm not very well versed in C++ unfortunately, and the problem here would be that `WORKSPACE_RULE` is not static (and neither its members). But again, I'm not that expert with C++ and I might be missing something here.

vaxerski:
```cpp
auto       gapsIn     = WORKSPACE_RULE.gapsIn.value_or(g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue);
```

```cpp
static auto* const PGAPSIN = &g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue;
auto                         gapsIn    = WORKSPACE_RULE.gapsIn.value_or(*PGAPSIN);
```

src/config/ConfigManager.hpp
```diff
@@ -45,6 +46,19 @@ struct SMonitorRule {
     bool                enable10bit = false;
 };
 
+struct SWorkspaceRule {
+    std::string            monitor         = "";
+    std::string            workspaceString = "";
+    std::string            workspaceName   = "";
+    int                    workspaceId     = -1;
+    std::optional<int64_t> gapsIn          = 5;
```
vaxerski, line 54: since those are optional, shouldn't they be left empty? You're initializing some defaults and then using `value_or` which means it will always use `value` because it's init, no?

vaxerski (changes requested):
small note. For the wiki MR, please make an accompanying MR, standard part of MRs.

I'll test this once you fix that one issue and let you know if any other issues arise

src/config/ConfigManager.hpp
levnikmyskin: Ups, correct. Forgot to change it. Should be good now.

levnikmyskin:
> small note. For the wiki MR, please make an accompanying MR, standard part of MRs.
> 
> I'll test this once you fix that one issue and let you know if any other issues arise

Sounds good, I'll prepare the wiki MR in the next days

src/config/ConfigManager.cpp
```diff
@@ -997,10 +1006,59 @@ void CConfigManager::handleBlurLS(const std::string& command, const std::string&
...
+            return;
+        }
+        wsRule.monitor         = first_ident;
+        wsRule.workspaceString = wsIdent;
+        rules                  = value.substr(WORKSPACE_DELIM + 1);
+    }
+
+    size_t      pos = 0;
+    std::string rule;
+    while ((pos = rules.find(',')) != std::string::npos) {
```
vaxerski, line 1038: you miss the last argument here

`gapsin:0, gapsout:0`
will not catch gapsout because there is no `,` at the end :P

vaxerski (changes requested):
shit works great, but

src/config/ConfigManager.cpp
levnikmyskin: lol indeed you're right. I changed the code with. I can commit it already, but I'd wait to solve the other issue too.
```cpp
auto assignRule = [&](std::string rule) {
        size_t delim = std::string::npos;
        Debug::log(INFO, "found workspacerule: %s", rule.c_str());
        if ((delim = rule.find("gapsin:")) != std::string::npos)
            wsRule.gapsIn = std::stoi(rule.substr(delim + 7));
        else if ((delim = rule.find("gapsout:")) != std::string::npos)
            wsRule.gapsOut = std::stoi(rule.substr(delim + 8));
        else if ((delim = rule.find("bordersize:")) != std::string::npos)
            wsRule.borderSize = std::stoi(rule.substr(delim + 11));
        else if ((delim = rule.find("border:")) != std::string::npos)
            wsRule.border = configStringToInt(rule.substr(delim + 7));
        else if ((delim = rule.find("rounding:")) != std::string::npos)
            wsRule.rounding = configStringToInt(rule.substr(delim + 9));
        else if ((delim = rule.find("decorate:")) != std::string::npos)
            wsRule.decorate = configStringToInt(rule.substr(delim + 9));
        else if ((delim = rule.find("monitor:")) != std::string::npos)
            wsRule.monitor = rule.substr(delim + 8);
    };
size_t      pos = 0;
std::string rule;
while ((pos = rules.find(',')) != std::string::npos) {
    rule = rules.substr(0, pos);
    assignRule(rule);
    rules.erase(0, pos + 1);
}
assignRule(rules); // match remaining rule
```

src/layout/DwindleLayout.cpp
```diff
@@ -108,11 +108,18 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
...
+    // if user specified them in config
+    const auto         WORKSPACE_RULE = g_pConfigManager->getWorkspaceRuleFor(g_pCompositor->getWorkspaceByID(PWINDOW->m_iWorkspaceID));
+
+    static auto* const PGAPSIN     = &g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue;
+    static auto* const PGAPSOUT    = &g_pConfigManager->getConfigValuePtr("general:gaps_out")->intValue;
+    static auto* const PBORDERSIZE = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
+
+    auto               gapsIn     = WORKSPACE_RULE.gapsIn.value_or(*PGAPSIN);
+    auto               gapsOut    = WORKSPACE_RULE.gapsOut.value_or(*PGAPSOUT);
+    auto               borderSize = WORKSPACE_RULE.borderSize.value_or(*PBORDERSIZE);
```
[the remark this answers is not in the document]
levnikmyskin: wait, not clear to me. The bordersize behaviour should be unaffacted by my code right? was this a pre-existing issue?

Anyways, I need yet again some guidance here. I can simply add a field to the `SWindowSpecialRenderData` struct, but where should this be used then? I assume something has to change in the `CWindow::updateWindowDecos()` somewhere, but wouldn't really know how or what.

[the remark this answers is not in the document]
vaxerski: windows never had bordersize props. The border size here is only used to calculate gaps, because gaps = gaps_in/out + borderSize

> but where should this be used then

CHyprRenderer::renderWindow

levnikmyskin:
Last commit should implement the border size spec on windows. Please let me know if it works, and if this is then ready to be merged.

levnikmyskin:
Hey! I added the documentation in the hyprland wiki, just fyi.

src/layout/DwindleLayout.cpp
```diff
@@ -144,13 +151,14 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
         return;
     }
 
-    PWINDOW->m_sSpecialRenderData.rounding = true;
-    PWINDOW->m_sSpecialRenderData.border   = true;
-    PWINDOW->m_sSpecialRenderData.decorate = true;
+    PWINDOW->m_sSpecialRenderData.rounding   = WORKSPACE_RULE.rounding.value_or(true);
+    PWINDOW->m_sSpecialRenderData.decorate   = WORKSPACE_RULE.decorate.value_or(true);
+    PWINDOW->m_sSpecialRenderData.border     = WORKSPACE_RULE.border.value_or(true);
+    PWINDOW->m_sSpecialRenderData.borderSize = borderSize;
```
vaxerski, line 157: `value_or`? Why are you hard-setting it here? It should be left as `-1` (unset) if it's not overridden.

src/render/OpenGL.cpp
```diff
@@ -1058,17 +1058,21 @@ void CHyprOpenGLImpl::renderBorder(wlr_box* box, const CGradientValueData& grad,
     static auto* const PBORDERSIZE  = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
     static auto* const PMULTISAMPLE = &g_pConfigManager->getConfigValuePtr("decoration:multisample_edges")->intValue;
 
-    wlr_box            newBox = *box;
+    if (borderSize < 0) {
+        borderSize = *PBORDERSIZE;
```
vaxerski, line 1062: I don't like this approach, I'd rather make borderSize required and NOT shadow it here with a config var, instead doing the checks and stuff in the renderer.

src/render/Renderer.cpp
```diff
@@ -342,7 +342,7 @@ void CHyprRenderer::renderWindow(CWindow* pWindow, CMonitor* pMonitor, timespec*
 
             scaleBox(&windowBox, pMonitor->scale);
 
-            g_pHyprOpenGL->renderBorder(&windowBox, grad, rounding, a1);
+            g_pHyprOpenGL->renderBorder(&windowBox, grad, rounding, a1, pWindow->m_sSpecialRenderData.borderSize);
```
vaxerski, line 345: if borderSize is not -1, pass it. Otherwise, pass the global config border size value. Render data contains additional render data, not persistent render data. It's meant for addons, not properties.

vaxerski (changes requested):

vaxerski:
if there's anything amiss after this I reckon I'll just fix it myself instead of doing this ping-pong

src/layout/DwindleLayout.cpp
levnikmyskin: Isn't that what the previous code was doing? I hard coded it just to replicate the previous behaviour, which was:
```cpp
PWINDOW->m_sSpecialRenderData.rounding = true;
PWINDOW->m_sSpecialRenderData.border   = true;
PWINDOW->m_sSpecialRenderData.decorate = true;
```

src/render/OpenGL.cpp
levnikmyskin: fair enough. I hadn't checked where else this code was used so I just went for the variable with default, trying not to break anything else

src/render/Renderer.cpp
levnikmyskin: Yeah, actually since this check is already done in the layout code, we might as well just use the `pWindow->m_sSpecialRenderData.borderSize` (which anyways defaults to -1). Since this function is called by other code though, we might also add the check here, I don't know if it makes sense though. What do you think?

levnikmyskin:
So, I've just pushed a commit where I removed the default value from the `renderBorder` method in the meantime. I'll wait for your advices on the remaining things

src/layout/DwindleLayout.cpp
vaxerski: yes, borderSize. it should be -1 if a workspace rule is not overriding it. Currently, it's always set.

vaxerski (approved):
looks gangsta

vaxerski:
if you could adjust to my comment in the wiki mr we could get this merged

Issue #1999, opened 2023-04-08
Move to empty workspace on next monitor with `movefocus` and `movewindow`


happenslol:
I've just moved to hyprland from sway, and I've realized that my workflow pretty much only consists of moving focus/windows in directions. In sway, it was possible to move to an empty workspace with a direction move, and I could also move a window to the next (empty) monitor by moving it in that direction.

In hyprland, it seems like moving the focus to an empty workspace with a direction is not possible at all, and moving a window is only possible if another window is on that workspace already - and the windows will be swapped instead of actually moving the window to the workspace, next to the window already there.

I'm interested in working on this feature myself. Is this feature something you would be open to?

vaxerski:
It's been on my radar to fixup `movewindow` for a while, because atm it's not a `move`, it's a `swap`

the `movefocus` thing would be trivial to implement yes

happenslol:
Alright, got the development environment set up. Those makefiles need cleaning up sometime, that was a real journey :sweat_smile: Any idea how I might go about testing multi-screen features in a nested window? Restarting my entire WM every time I want to test a change seems like a huge timesink :thinking:

vaxerski:
run `hyprctl output create auto` in a nest

PR #2011, opened 2023-04-10
Allow movefocus for empty workspaces

happenslol:

As discussed in #1999, this PR implements moving to and from empty workspaces using `movefocus`.

I've tried to adapt to your naming schemes, but I'm not sure in which cases you use the `P` prefix for variables, and if I'm correct in assuming that all local window/workspace/monitor variables in functions should be all uppercase (e.g. `PNEWWINDOW`). I'd like having at least underscore to separate words for readability (`P_NEW_WINDOW`), but I'll use the current style if that's what you prefer.

I was also wondering, if a workspace has no current focused window, is it possible that there are still windows? I imagine that could happen if a silent move to a workspace occurred. See my TODO in the code for that.

I was unsure about how this functionality should be activated, since it's potentially breaking the workflow of current users who depend on a fallback to any window if no window is found in the direction. It would also be possible to still do that fallback behavior if no monitor was found in that direction.

My idea would be to add an option to enable the fallback behavior (`getNextWindowOnWorkspace`) and enable it by default.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1153,14 +1153,34 @@ void CKeybindManager::moveFocusTo(std::string args) {
     const auto PWINDOWTOCHANGETO = PLASTWINDOW->m_bIsFullscreen ? g_pCompositor->getNextWindowOnWorkspace(PLASTWINDOW, arg == 'u' || arg == 't' || arg == 'r') :
                                                                   g_pCompositor->getWindowInDirection(PLASTWINDOW, arg);
 
+    // Found window in direction, switch to it
     if (PWINDOWTOCHANGETO) {
         switchToWindow(PWINDOWTOCHANGETO);
-    } else {
-        const auto PWINDOWNEXT = g_pCompositor->getNextWindowOnWorkspace(PLASTWINDOW, true);
-        if (PWINDOWNEXT) {
-            switchToWindow(PWINDOWNEXT);
+        return;
+    }
+
+    Debug::log(LOG, "No window found in direction %c, looking for a monitor", arg);
+
+    // No window found in direction, switch to workspace in direction
+    const auto PNEXTMONITOR = g_pCompositor->getMonitorInDirection(arg);
+    if (PNEXTMONITOR && PNEXTMONITOR->activeWorkspace != -1) {
+        const auto PNEWWORKSPACE = g_pCompositor->getWorkspaceByID(PNEXTMONITOR->activeWorkspace);
+        if (PNEWWORKSPACE) {
+            g_pCompositor->focusWindow(nullptr);
+            g_pCompositor->setActiveMonitor(PNEXTMONITOR);
+            g_pCompositor->deactivateAllWLRWorkspaces(PNEWWORKSPACE->m_pWlrHandle);
+            PNEWWORKSPACE->setActive(true);
+            return;
         }
     }
+
+    Debug::log(LOG, "No monitor found in direction %c", arg);
+
+    // TODO: Add option to disable looping/falling back to any next window?
+    // const auto PWINDOWNEXT = g_pCompositor->getNextWindowOnWorkspace(PLASTWINDOW, true);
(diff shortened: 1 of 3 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -1153,14 +1181,38 @@ void CKeybindManager::moveFocusTo(std::string args) {
...
+            Vector2D middle = PNEWMONITOR->vecPosition + PNEWMONITOR->vecSize / 2.f;
+            g_pCompositor->warpCursorTo(middle);
+
+            return;
         }
     }
+
+    Debug::log(LOG, "No monitor found in direction %c", arg);
+
+    // TODO: Add option to disable looping/falling back to any next window?
```
vaxerski, line 1211: I'd rather not yeet it atm if we don't have the option

src/managers/KeybindManager.cpp
```diff
@@ -1153,14 +1181,38 @@ void CKeybindManager::moveFocusTo(std::string args) {
...
-        if (PWINDOWNEXT) {
-            switchToWindow(PWINDOWNEXT);
+        return;
+    }
+
+    Debug::log(LOG, "No window found in direction %c, looking for a monitor", arg);
+
+    // No window found in direction, switch to workspace in direction
+    const auto PNEWMONITOR = g_pCompositor->getMonitorInDirection(arg);
+    if (PNEWMONITOR && PNEWMONITOR->activeWorkspace != -1) {
```
vaxerski, line 1194: redundant `activeWorkspace` check. it's guaranteed

src/managers/KeybindManager.cpp
```diff
@@ -1113,9 +1113,37 @@ void CKeybindManager::moveFocusTo(std::string args) {
...
     const auto PLASTWINDOW = g_pCompositor->m_pLastWindow;
+    if (!PLASTWINDOW) {
+        // Try to move to monitor in direction
+        const auto PNEWMONITOR = g_pCompositor->getMonitorInDirection(arg);
+        if (!PNEWMONITOR) {
+            return;
+        }
+
+        const auto PNEWWORKSPACE = g_pCompositor->getWorkspaceByID(PNEWMONITOR->activeWorkspace);
+        if (!PNEWWORKSPACE) {
```
vaxerski, line 1124: redundant

src/managers/KeybindManager.cpp
```diff
@@ -1153,14 +1181,38 @@ void CKeybindManager::moveFocusTo(std::string args) {
...
+    }
+
+    Debug::log(LOG, "No window found in direction %c, looking for a monitor", arg);
+
+    // No window found in direction, switch to workspace in direction
+    const auto PNEWMONITOR = g_pCompositor->getMonitorInDirection(arg);
+    if (PNEWMONITOR && PNEWMONITOR->activeWorkspace != -1) {
+        const auto PNEWWORKSPACE = g_pCompositor->getWorkspaceByID(PNEWMONITOR->activeWorkspace);
+        if (PNEWWORKSPACE) {
+            g_pCompositor->focusWindow(nullptr);
```
vaxerski, line 1197: what if there is a floating window? You should do the same as above, getlastfocusedwindow

vaxerski (changes requested):
https://user-images.githubusercontent.com/43317083/230892463-25776994-3bcb-4f7a-a5cd-976f7dd8357c.mp4

src/managers/KeybindManager.cpp
happenslol: Can I just add an option? Cause I personally don't want that at all, so while I'm at it

happenslol:
Ok i pull up

All except the option thingie should be fixed, tell me if you're ok with the helper function like that

src/managers/KeybindManager.cpp
vaxerski: sure

vaxerski (approved):
lgtm

Issue #2002, opened 2023-04-09
Allow exceptions to window swallowing


mrkajetanp:
As far as I can see, under the current implementation window swallowing is pretty much all or nothing - if I add the terminal in the regex, it will swallow any window it spawns. While that's the desired behaviour almost all the time there's occasionally some exceptions (like `wev`) where you want to keep the default behaviour. Could we add something like `swallow_exception_regex` for this use case?

vaxerski:
how would you even except that to work?? Can't you just make a regex that does _not_ target that one thing that you don't want to swallow?

grappas:
I think swallow_regex =^(classes)$ but in reverse (not terminals but child windows like wev for example)

tchofy:
Most terminals put the working application on their title, or can be configured to do so, so what about excluding by title?
I'm not particularly good with regex yet, but I think it'd go something like: swallow_regex=class:kitty,title:^(?!wev)$

mrkajetanp:
> I think swallow_regex =^(classes)$ but in reverse (not terminals but child windows like wev for example)

Yes exactly, other WMs that do swallowing like DWM for instance already have this functionality. Implementation-wise it just boils down to matching against the regex before a window is about to be swallowed to see if it's one of the exceptions or not.

I implemented a PR that does just that, it's the one mentioned right above this comment. I tested it on my machine and now it works as intended.

MightyPlaza:
why this open?

mrkajetanp:
Good question, probably oversight.. I'll close this

PR #2026, opened 2023-04-11
swallow: Add swallow_exception_regex

mrkajetanp:

Currently, if a window class is specified in the swallow_regex (e.g. Kitty) it will swallow every other window spawned by it automatically. Many other WMs implementing this functionality allow for defining exceptions from this rule. For instance, we want Kitty to swallow sxiv or zathura but we do not want Kitty to swallow something like wev.

This commit adds an additional regex - swallow_exception_regex where these exceptions can be defined. This regex is then compared against the title of the window about to be swallowed and if it happens to be a match, aborts the swallowing.

This works because whenever an application that could be swallowed is launched by a terminal, the class of the terminal remains the same while the title changes to whatever the application's name is, thus letting it be matched against a regex.

The PR should be ready to be merged unless the config option name should be changed for some reason. Presumably the documentation will need to be updated as well.

Solves issue #2002

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -81,6 +81,7 @@ void CConfigManager::setDefaultVars() {
     configValues["misc:disable_autoreload"].intValue           = 0;
     configValues["misc:enable_swallow"].intValue               = 0;
     configValues["misc:swallow_regex"].strValue                = STRVAL_EMPTY;
+    configValues["misc:swallow_exception_regex"].strValue      = STRVAL_EMPTY;
     configValues["misc:focus_on_activate"].intValue            = 0;
     configValues["misc:no_direct_scanout"].intValue            = 1;
     configValues["misc:hide_cursor_on_touch"].intValue         = 1;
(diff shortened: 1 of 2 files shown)


vaxerski:
overall ok, make a wiki mr plox

mrkajetanp:
> overall ok, make a wiki mr plox

Wiki PR submitted, it's mentioned above

vaxerski:
discrepancy: in the wiki you say regex for the _child_, while here you match against the _parent_?

mrkajetanp:
Hm I think it probably makes more sense for it to be described this way in the wiki context?

As in we have a parent (Kitty) that spawns a child process (wev). At that point the title of that Kitty instance changes to 'wev' which is why that's what we match against.

I think emphasising 'child' on the wiki makes it easier for people tell which names go into which regex but if you have a better way to phrase this that's fine by me.

vaxerski:
you can add that note, but since you match the parent, you should state you match the parent. Shells/terms might have different methods of handling shit

mrkajetanp:
Fair enough, I updated the PR, how about this?

vaxerski:
yeah lgtm

PR #2029, opened 2023-04-12
Add "tick" event for plugins

Duckonaut:
Add a "tick" event for plugins to use for periodic calls, useful for handling async problems.

Could theoretically pass the delta time between ticks as a param of the event? I personally don't think it would be useful, and if needed can be kept plugin-side anyway.

Ready


vaxerski (approved):
lgtm

Issue #1040, opened 2022-11-18
global config file?


quantenzitrone:
What do you guys think about a global config file directory like `$XDG_CONFIG_DIRS/hypr` and `/etc/xdg/hypr` as fallback for all hypr related stuff?

Having a system-wide configuration file is actually part of the [xdg base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

> `$XDG_CONFIG_DIRS` defines the preference-ordered set of base directories to search for configuration files in addition to the `$XDG_CONFIG_HOME` base directory. The directories in `$XDG_CONFIG_DIRS` should be seperated with a colon ':'. 

Another benefit would that NixOSers wouldn't have to use home-manager to configure Hyprland from their configuration.nix.

vaxerski:
@fufexan kowalski, analysis

fufexan:
You don't *need* HM to configure Hyprland. It'll create a config file if `~/.config/hypr/hyprland.conf` doesn't exist. You can then modify that as usual.

I'm not opposed to this change but using HM and having this be done are orthogonal.

Quantenzitrone:
You need home manager to place config files reproducably in your home directory. For global config files, you can use `environment.etc.`

fufexan:
Can I close this? Hyprland auto-generates a config file, which you do not need to manage with Home Manager.

Quantenzitrone:
I don't see how configuration file auto generation solves this. To change the configuration globally, one would need to recompile the changed nix expression every time and even then Hyprland would only use the new global config, if i delete the old one.
Yes, i can have editable configuration files floating around in my XDG_CONFIG_HOME, which i don't need home-manager to manage.
But if i want to include my Hyprland config in my reproducible NixOS config repository, I would either need home-manager, or a global config file.

Quantenzitrone:
Also, I don't see this as a NixOS only issue, because a global config file is something that could concern people on other distros too.

fufexan:
A solution until this is properly taken care of would be to write your config to `/etc/xdg/hyprland.conf` and starting hyprland with `Hyprland -c /etc/xdg/hyprland.conf`.

I don't know when I will have time, but I'll try to implement this.

Quantenzitrone:
Yeah, i gotta be honest, I've gone back to i3 at the moment, but thanks

romanstingler:
@Quantenzitrone this issue should be closed or?

[...]

PR #2047, opened 2023-04-13
Xdg config home support

edyounis:

The default config path will now first check if XDG_CONFIG_HOME exists and will use it rather than $HOME/.config if it does exist.

I didn't read deep enough into the code to understand what OpenGL.cpp is doing, so I am not 100% sure my change makes sense there.

Tested it locally, seems to work in all cases.

Related to #1040

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1310,6 +1310,7 @@ void CConfigManager::loadConfigLoadVars() {
     // paths
     configPaths.clear();
     std::string mainConfigPath = getMainConfigPath();
+    Debug::log(LOG, "Using config: %s", mainConfigPath.c_str());
     configPaths.push_back(mainConfigPath);
     std::string configPath = mainConfigPath.substr(0, mainConfigPath.find_last_of('/'));
     // find_last_of never returns npos since main_config atleast has /hpyr/
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -21,20 +21,27 @@ CConfigManager::CConfigManager() {
     setDefaultVars();
     setDefaultAnimationVars();
 
-    std::string CONFIGPATH;
-    if (g_pCompositor->explicitConfigPath == "") {
-        static const char* const ENVHOME = getenv("HOME");
-        CONFIGPATH                       = ENVHOME + (ISDEBUG ? (std::string) "/.config/hypr/hyprlandd.conf" : (std::string) "/.config/hypr/hyprland.conf");
-    } else {
-        CONFIGPATH = g_pCompositor->explicitConfigPath;
-    }
-
-    configPaths.emplace_back(CONFIGPATH);
+    configPaths.emplace_back(getMainConfigPath());
 
     Debug::disableLogs = &configValues["debug:disable_logs"].intValue;
     Debug::disableTime = &configValues["debug:disable_time"].intValue;
 
     populateEnvironment();
+}
+
+std::string CConfigManager::getMainConfigPath() {
+    if (!g_pCompositor->explicitConfigPath.empty())
+        return g_pCompositor->explicitConfigPath;
+
+    static const char* xdgConfigHome = getenv("XDG_CONFIG_HOME");
+    std::string        configPath;
+    if (!xdgConfigHome)
+        configPath = getenv("HOME") + std::string("/.config");
+    else
+        configPath = xdgConfigHome;
+    std::string configDebug = "/hypr/hyprlandd.conf";
+    std::string config      = "/hypr/hyprland.conf";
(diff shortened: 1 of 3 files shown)


src/config/ConfigManager.cpp
```diff
@@ -21,22 +21,28 @@ CConfigManager::CConfigManager() {
...
+    configPaths.emplace_back(getMainConfigPath());
 
     Debug::disableLogs = &configValues["debug:disable_logs"].intValue;
     Debug::disableTime = &configValues["debug:disable_time"].intValue;
 
     populateEnvironment();
 }
 
+std::string CConfigManager::getMainConfigPath() {
+    if (g_pCompositor->explicitConfigPath != "")
```
vaxerski, line 33: `.empty()`

src/config/ConfigManager.cpp
```diff
@@ -21,22 +21,28 @@ CConfigManager::CConfigManager() {
...
+        return g_pCompositor->explicitConfigPath;
+    
+    std::string config_home = std::string(getenv("XDG_CONFIG_HOME"));
```
vaxerski, line 36: we use camelCase

+ `std::string(const char*)` requires a nullcheck for the char*

src/config/ConfigManager.cpp
```diff
@@ -21,22 +21,28 @@ CConfigManager::CConfigManager() {
...
+    if (config_home == "")
+        config_home = std::string(getenv("HOME")) + "/.config";
+
+    std::string debug_file_name = std::string("/hypr/hyprlandd.conf");
+    std::string normal_file_name = std::string("/hypr/hyprland.conf");
+    std::string config_file_name = (ISDEBUG ? debug_file_name : normal_file_name);
+    return config_home + config_file_name;
```
vaxerski, line 43: all here as above

src/config/ConfigManager.cpp
```diff
@@ -1297,49 +1301,52 @@ void CConfigManager::loadConfigLoadVars() {
 
     // paths
     configPaths.clear();
+    std::string main_config = getMainConfigPath();
+    Debug::log(LOG, "Using config: %s", main_config.c_str());
+    configPaths.push_back(main_config);
+    std::string config_home = main_config.substr(0, main_config.find_last_of('/'));
+    // find_last_of never returns npos since main_config atleast has /hpyr/
+
+    if (!std::filesystem::is_directory(config_home))
```
vaxerski, line 1310: if () {

Please format your file with the provided `.clang-format`

vaxerski (changes requested):
all of the comments are applicable to the entirety of the mr

src/config/ConfigManager.cpp
Thaodan: Using std::string in that case makes it harder to check if getenv returns nullptr.

edyounis:
Supersedes the superseding: #2107
Implemented the requested changes in the other PR.

src/config/ConfigManager.cpp
```diff
@@ -21,22 +21,31 @@ CConfigManager::CConfigManager() {
...
+    else
+        configPath = xdgConfigHome;
+    return configPath;
+}
+
+std::string CConfigManager::getMainConfigPath() {
+    if (!g_pCompositor->explicitConfigPath.empty())
+        return g_pCompositor->explicitConfigPath;
+
+    return getConfigDir() + (ISDEBUG ? "/hypr/hyprlandd.conf" : "/hypr/hyprland.conf");
```
fufexan, line 46:
```suggestion
    return getConfigDir() + "/hypr/" + (ISDEBUG ? "hyprlandd.conf" : "hyprland.conf");
```

src/config/ConfigManager.cpp
```diff
@@ -1304,49 +1311,49 @@ void CConfigManager::loadConfigLoadVars() {
...
+    configPaths.push_back(mainConfigPath);
+    std::string configPath = mainConfigPath.substr(0, mainConfigPath.find_last_of('/'));
+    // find_last_of never returns npos since main_config at least has /hypr/
+
+    if (!std::filesystem::is_directory(configPath)) {
+        Debug::log(WARN, "Creating config home directory");
+        try {
+            std::filesystem::create_directories(configPath);
+        } catch (...) {
+            parseError = "Broken config file! (Could not create directory)";
```
fufexan, line 1325: Why is the config file broken? It's non-existent. This should probably say `Could not create config directory`.

fufexan (changes requested):

src/config/ConfigManager.cpp
edyounis: This error message existed as-is before this PR, but will be happy to change it.

vaxerski (approved):
seems alright to me, wbu @fufexan

fufexan:
all good

PR #2066, opened 2023-04-15
Unbreak dbus-update-activation-environment calls on non-systemd distros

jbeich:
fea2031bfe3f + 4abc608bc025 added `systemctl` outside of `#ifdef USES_SYSTEMD`. That would be fine if the command didn't use `&&` to suppress `dbus-update-activation-environment` fallback.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -443,7 +443,9 @@ void CCompositor::startCompositor() {
     if (m_sWLRSession /* Session-less Hyprland usually means a nest, don't update the env in that case */ && fork() == 0)
         execl(
             "/bin/sh", "/bin/sh", "-c",
+#ifdef USES_SYSTEMD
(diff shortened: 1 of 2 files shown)


vaxerski:
what about the `--systemd` flag? does it get ignored on non-sysd systems?

jbeich:
`--systemd` is [ignored](https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.14.6/tools/dbus-update-activation-environment.c#L164) on non-Linux:
```
$ dbus-update-activation-environment --verbose --systemd WAYLAND_DISPLAY
dbus-update-activation-environment: not on Linux, ignoring --systemd argument
dbus-update-activation-environment: setting WAYLAND_DISPLAY=wayland-2
$ echo $?
0
```
and [also](https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.14.6/tools/dbus-update-activation-environment.c#L235) on Linux without systemd (didn't test myself).

vaxerski:
alr, thanks!

vaxerski (approved):

PR #2087, opened 2023-04-17
Plugin header overhaul

horriblename:

More of a proof of concept, example flake that uses this pluginenv package to build hyprbars: https://github.com/horriblename/hyprbars-nix

I barely know any meson and nix, there's probably a better way to do this

(thanks to https://github.com/outfoxxed/hyprland-plugin-nix for the cmake and flake template)

yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+PREFIX = /usr/local
 
 legacyrenderer:
 	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DLEGACY_RENDERER:BOOL=true -S . -B ./build -G Ninja
--- a/Makefile
+++ b/Makefile
@@ -1,155 +1,3 @@
-include config.mk
-
-CFLAGS += -I. -DWLR_USE_UNSTABLE -std=c99
-
-WAYLAND_PROTOCOLS=$(shell pkg-config --variable=pkgdatadir wayland-protocols)
-WAYLAND_SCANNER=$(shell pkg-config --variable=wayland_scanner wayland-scanner)
-
-PKGS = wlroots wayland-server xcb xkbcommon libinput
-CFLAGS += $(foreach p,$(PKGS),$(shell pkg-config --cflags $(p)))
-LDLIBS += $(foreach p,$(PKGS),$(shell pkg-config --libs $(p)))
-
-DATE=$(shell date "+%d %b %Y")
-
-xdg-shell-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		$(WAYLAND_PROTOCOLS)/stable/xdg-shell/xdg-shell.xml $@
-
-xdg-shell-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		$(WAYLAND_PROTOCOLS)/stable/xdg-shell/xdg-shell.xml $@
-
-xdg-shell-protocol.o: xdg-shell-protocol.h
-
-wlr-layer-shell-unstable-v1-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/wlr-layer-shell-unstable-v1.xml $@
-
-wlr-layer-shell-unstable-v1-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/wlr-layer-shell-unstable-v1.xml $@
-
-wlr-layer-shell-unstable-v1-protocol.o: wlr-layer-shell-unstable-v1-protocol.h
-
-wlr-screencopy-unstable-v1-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/wlr-screencopy-unstable-v1.xml $@
-
-wlr-screencopy-unstable-v1-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/wlr-screencopy-unstable-v1.xml $@
-
-wlr-screencopy-unstable-v1-protocol.o: wlr-screencopy-unstable-v1-protocol.h
-
-ext-workspace-unstable-v1-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/ext-workspace-unstable-v1.xml $@
-
-ext-workspace-unstable-v1-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/ext-workspace-unstable-v1.xml $@
-
-ext-workspace-unstable-v1-protocol.o: ext-workspace-unstable-v1-protocol.h
-
-pointer-constraints-unstable-v1-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/pointer-constraints-unstable-v1.xml $@
-
-pointer-constraints-unstable-v1-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/pointer-constraints-unstable-v1.xml $@
-
-pointer-constraints-unstable-v1-protocol.o: pointer-constraints-unstable-v1-protocol.h
-
-tablet-unstable-v2-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/tablet-unstable-v2.xml $@
-
-tablet-unstable-v2-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/tablet-unstable-v2.xml $@
-
-tablet-unstable-v2-protocol.o: tablet-unstable-v2-protocol.h
-
-idle-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/idle.xml $@
-
-idle-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/idle.xml $@
-
-idle-protocol.o: idle-protocol.h
-
-wlr-output-power-management-unstable-v1-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		protocols/wlr-output-power-management-unstable-v1.xml $@
-
-wlr-output-power-management-unstable-v1-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		protocols/wlr-output-power-management-unstable-v1.xml $@
-
-wlr-output-power-management-unstable-v1-protocol.o: wlr-output-power-management-unstable-v1-protocol.h
-
-hyprland-toplevel-export-v1-protocol.h:
-	$(WAYLAND_SCANNER) server-header \
-		subprojects/hyprland-protocols/protocols/hyprland-toplevel-export-v1.xml $@
-
-hyprland-toplevel-export-v1-protocol.c:
-	$(WAYLAND_SCANNER) private-code \
-		subprojects/hyprland-protocols/protocols/hyprland-toplevel-export-v1.xml $@
-
-hyprland-toplevel-export-v1-protocol.o: hyprland-toplevel-export-v1-protocol.h
-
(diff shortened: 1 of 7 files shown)


vaxerski:
@fufexan nix XD

fufexan:
@eli-schwartz sorry for the ping, but I know you work on Meson. Is this approach to generating headers and protocols the right one, or should we try a different way?

meson.build
```diff
@@ -75,9 +75,20 @@ if get_option('buildtype') == 'debug'
...
-subdir('assets')
-subdir('example')
-subdir('docs')
+if get_option('pluginenv')
+  fs = import('fs')
+  globber = run_command('find', 'src', '-name', '*.hpp', check: true)
+  headers = globber.stdout().strip().split('\n')
+  foreach file : headers
+    dir = fs.parent(file)
+    fs.copyfile(file, install: true, install_dir: dir)
```
eli-schwartz, line 84: I don't understand what this is even trying to do. You want to install all header files from the src directory to the install directory, so you are copying them to the build directory instead? Why?

Just use install_headers().

You probably also want to install a pkg-config file that plugin env software can check for.

You should double check what files are installed when running "meson install", e.g. by posting the verbose status log of meson itself. Because it's not obvious that these files are ending up in the right place right now.

meson_options.txt
```diff
@@ -1,3 +1,4 @@
 option('xwayland', type: 'feature', value: 'auto', description: 'Enable support for X11 applications')
 option('systemd', type: 'feature', value: 'auto', description: 'Enable systemd integration')
 option('legacy_renderer', type: 'feature', value: 'disabled', description: 'Enable legacy renderer')
+option('pluginenv', type: 'boolean', value: false, description: 'Generate protocol headers and exit')
```
eli-schwartz, line 4: Hmm. Why not produce both as part of the same installation process? Do people typically want only the headers, but not to install hyprland itself?

eli-schwartz (commented):
No worries, I'm fine being pinged. :)

meson_options.txt
fufexan: No, I think it's fine if we provide both when installing.

meson.build
horriblename: can't believe I didn't find it before, thx

fufexan:
@eli-schwartz we've applied your suggestions, and now the `include` dir seems to have the necessary files, and [`hyprland.pc`](https://github.com/hyprwm/Hyprland/files/11264944/hyprland.pc.txt) *seems* alright to me.

However, I think I've missed something. I've tried building the [hyprbars](https://github.com/hyprwm/hyprland-plugins/tree/meson/hyprbars) plugin using a `meson.build` I just wrote, but it's trying to link Hyprland from what I can see.
The log is here [ninja.log](https://github.com/hyprwm/Hyprland/files/11264922/ninja.log).

Could you take a look and tell me where I've messed up? I've been at it for a few hours :')

eli-schwartz:
Since we're talking about plugins, I assume that hyprbars is intended to be `dlopen()`'ed by the Hyprland executable, and it is the job of the Hyprland executable to export symbols for use by plugins.

As such you don't want to build a "library", you want to build a "shared_module". This has a few semantic differences, for example, shared modules do not receive a DT_SONAME. They also are not linked with -Wl,--no-undefined, because it is *expected* that symbols are not known at link time, only at runtime.

Aside: which directory does Hyprland scan for plugins?

fufexan:
> As such you don't want to build a "library", you want to build a "shared_module".

That makes sense, thanks!

> Aside: which directory does Hyprland scan for plugins?

There's no default directory. They can be loaded using `hyprctl plugin load path/to/plugin.so`, and unloaded the same way.
For a persistent configuration, the above command can be added to `hyprland.conf` as an `exec-once` call.

meson.build
```diff
@@ -75,9 +75,26 @@ if get_option('buildtype') == 'debug'
...
+pkgconfig_configuration = configuration_data()
+pkgconfig_configuration.set('PREFIX', get_option('prefix'))
+pkgconfig_configuration.set('HYPRLAND_VERSION', meson.project_version())
+pkg_install_dir = join_paths(get_option('datadir'), 'pkgconfig')
+configure_file(
+	input: 'hyprland.pc.in',
+	output: 'hyprland.pc',
+	configuration: pkgconfig_configuration,
+	install_dir: pkg_install_dir,
+)
```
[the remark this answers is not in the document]
fufexan: I actually did this at first but decided to use `hyprland.pc.in` so that CMake could also build `hyprland.pc`. But I suppose I could let CMake do its thing, while having Meson generate the pc file directly.

fufexan:
If there aren't any more suggestions, I will merge this.

fufexan:
One thing I remembered. CMake doesn't install the headers yet, so we'll have to figure that out before merging.

Thaodan:
I was just looking into the same issue for OpenSUSE, the PR will help with that too.

horriblename:
make file now installs headers + pkgconfig

Unless I'm missing something, we can merge this now

Makefile
```diff
@@ -203,6 +203,11 @@ install:
 
 	install -Dm644 -t ${PREFIX}/share/man/man1 ./docs/*.1
 
+	mkdir -p ${PREFIX}/include/hyprland
```
vaxerski, line 206: doing this essentially deprecates `make pluginenv`, no?

fufexan: Yes. We can remove it and instruct plugin devs to install Hyprland from source again, and then use `#include <hyprland/src/...>`.

vaxerski: sure.

Makefile
```diff
@@ -203,6 +203,11 @@ install:
...
+	find src -name '*.h*' -exec cp --parents '{}' ${PREFIX}/include/hyprland ';'
+	cp ./*-protocol.h ${PREFIX}/include
```
vaxerski, line 208: Not sure whether this is a good idea, we're kinda trashing `/usr/include`.
Wouldn't this be better put in like e.g. `${PREFIX}/include/hyprland/protocols`?

fufexan: Yeah, probably a good idea. I'll see if I can also make that work in meson.

vaxerski: In general, IMHO, a clean solution is to move protocols to e.g. `/protocols/build`

fufexan: I tried this in Meson, but it doesn't work. Maybe someone has a better idea.

```diff
diff --git a/meson.build b/meson.build
index ddf5c97f..d7bc1a52 100644
--- a/meson.build
+++ b/meson.build
@@ -88,6 +88,12 @@ subdir('assets')
 subdir('example')
 subdir('docs')
 
+protocols_globber = run_command('find', '$MESON_BUILD_ROOT/protocols', '-name', '*.h*', check: true)
+protocols_headers = protocols_globber.stdout().strip().split('\n')
+foreach file : protocols_headers
+  install_headers(file, subdir: 'hyprland/protocols')
+endforeach
+
 pkg_install_dir = join_paths(get_option('datadir'), 'pkgconfig')
 
 import('pkgconfig').generate(
diff --git a/protocols/meson.build b/protocols/meson.build
index 193d7aaf..09003b7c 100644
--- a/protocols/meson.build
+++ b/protocols/meson.build
@@ -47,8 +47,6 @@ foreach p : protocols
 	wl_protos_headers += custom_target(
 		xml.underscorify() + '_server_h',
 		input: xml,
-		install: true,
-		install_dir: 'include',
 		output: '@BASENAME@-protocol.h',
 		command: [wayland_scanner, 'server-header', '@INPUT@', '@OUTPUT@'],
 	)
```

```
meson.build:91:0: ERROR: Command "/nix/store/j5wraaxv16fcl10x11566a3807nr4nlr-findutils-4.9.0/bin/find $MESON_BUILD_ROOT/protocols -name *.h*" failed with status 1.
```

eli-schwartz: I'm very confused... Shouldn't you just change the install_dir?

fufexan: That was my first idea as well. Turns out Nix doesn't like that. The way it constructs the `hyprland-dev` package is by `mv`ing the include dir into another. And it fails when I change `install_dir` to any subdir of `include`, because the dir already exists and mv fails.
I will search for a workaround tomorrow.

eli-schwartz: Shouldn't you just update Nix as well? To adapt to the updated install requirements.

eli-schwartz: Actually this is probably a meson.build bug, you should respect -Dincludedir by using get_option() and then nix can use standard rulesets to install without manual mv.

fufexan: Nix relies on Meson & Ninja to do the heavy lifting of building the package and installing it into a directory. I did add a "dev" output which is what causes this behavior. I could leave it out though, and it shouldn't fail anymore.

horriblename: a possible workaround would be to patch meson to install into `include/` in nix only (and install into `include/hyprland` on other systems), that is if no other solution exists ofc

fufexan: @eli-schwartz your idea worked perfectly. Thanks!

protocols/meson.build
```diff
@@ -47,6 +47,8 @@ foreach p : protocols
 	wl_protos_headers += custom_target(
 		xml.underscorify() + '_server_h',
 		input: xml,
+		install: true,
+		install_dir: join_paths(get_option('includedir'), 'hyprland/protocols'),
```
eli-schwartz, line 51: This will install, to e.g.`-I${includedir}/hyprland` -- as provided in the pkg-config file -- a header named `protocols/@BASENAME@-protocol.h`. For example, you'd then use:
```
#include <protocols/idle-protocol.h>
```

To include the file `/usr/include/hyprland/protocols/idle-protocol.h`

Can you confirm that's what you want? Because it's not what the Makefile does in this PR.

eli-schwartz: If so then you don't want to put `subdirs: 'hyprland',` into the pkg-config file. That adds `/path/to/nix-store/include/hyprland` to the include path but if "hyprland/" is part of the `#include` name then you need `/path/to/nix-store/include` on the include path.

vaxerski:
moved protocols to `/protocols` and now CMake builds them

vaxerski (approved):
LGTM

vaxerski:
Now something we completely forgot about were the wlroots headers. That's an oops.

sp1ritCS:
@fufexan @horriblename 
Is there any reason why (at least for meson) the pkg-config files gets installed to <pfx>/share/pkgconfig instead of <pfx>/lib64/pkgconfig?

eli-schwartz:
The difference between the two pkgconfig directories is, in fact, that one is in the Filesystem  Hierarchy Standard "lib" location, and the other one is in the Filesystem Hierarchy Standard "share" location.

This corresponds to the way that wall_8K.png goes in the Filesystem Hierarchy Standard "share" location, while libwlroots.so goes in the Filesystem Hierarchy Standard "lib" location.

...

So, does hyprland.pc match the definition of "architecture-independent data" as defined by https://www.pathname.com/fhs/pub/fhs-2.3.html#USRSHAREARCHITECTUREINDEPENDENTDATA or doesn't it?

eli-schwartz:
In this case I think it's correct, since there's only header includes, no references to an architecture-dependent shared library. So for example on a multi-arch system you could have one set of headers and one pkg-config file that is in a shared data package, and two different arch-specific packages each containing libraries and binaries in split-arch libdir/bindir locations. If the pkg-config file could be stored in the common multi-arch package, then it counts as something which belongs in share/.

(One case where headers cannot be in a common arch-independent package is when the headers are generated and contain different contents depending on architecture. This is more common for e.g. glibc than for high-level libraries.)

PR #2135, opened 2023-04-23
Add follow mouse mode to avoid refocusing under cursor

outfoxxed:
This PR adds a new follow_mouse mode (4), that is the same as 1 (focus follows mouse) except it requires crossing a window border to re-focus the hovered window. I find it useful because my window focus won't change if my mouse gets bumped.

I'm fairly sure adding `m_pLastMouseFocus` at the end of `CInputManager` keeps the plugin abi the same as long as it wasn't accessing private fields, but I'm also fairly sure thats not a concern anyway.

Ready to merge.

### DIFF
--- a/src/managers/input/InputManager.hpp
+++ b/src/managers/input/InputManager.hpp
@@ -175,6 +175,9 @@ class CInputManager {
     // for hiding cursor on touch
     bool m_bLastInputTouch = false;
 
+    // for follow mouse 4
+    CWindow* m_pLastMouseFocus = nullptr;
+
   private:
     bool                 m_bCursorImageOverridden = false;
     eBorderIconDirection m_eBorderIconDirection   = BORDERICON_NONE;
--- a/src/managers/input/InputManager.hpp
+++ b/src/managers/input/InputManager.hpp
@@ -175,7 +175,7 @@ class CInputManager {
     // for hiding cursor on touch
     bool m_bLastInputTouch = false;
 
-    // for follow mouse 4
+    // for tracking mouse refocus
     CWindow* m_pLastMouseFocus = nullptr;
 
   private:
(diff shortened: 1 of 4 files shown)


vaxerski:
witnof

this shouldn't be a follow_mouse opt, rather a separate on/off switch

furthermore I don't see how this changes anything

outfoxxed:
I can change it to a separate switch.

The difference from follow_mouse=1 is that the mouse can't refocus the hovered window if the focus has been moved by the keyboard.


https://user-images.githubusercontent.com/83010835/233858565-de997335-bc14-4c9b-85c1-6e458f83319c.mp4

vaxerski:
oh that thing

ok, yeah, bool, not mode

outfoxxed:
I tested all the modes and it would have no effect on follow_mouse=0,2,3 because those modes don't actively follow the mouse like follow_mouse=1 does. Should it still be a bool?

vaxerski:
yes

outfoxxed:
`follow_mouse=4` is now `mouse_refocus=0` combined with `follow_mouse=1`.

vaxerski:
yes, as usual. wiki pr plox

src/managers/input/InputManager.hpp
```diff
@@ -175,6 +175,9 @@ class CInputManager {
     // for hiding cursor on touch
     bool m_bLastInputTouch = false;
 
+    // for follow mouse 4
```
vaxerski, line 178: comment bug

outfoxxed: Fixed

outfoxxed:
Wiki pr above

vaxerski (approved):
lgtm

PR #2139, opened 2023-04-23
Fixed a crash when waking up monitors in power-saving mode

teschnei:
When waking up my monitors with dpms on after they powered down into power-saving mode, hyprland would crash in the animation manager.  I don't know what specifically in my setup causes it (one monitor takes longer to power on, 3 monitors, etc.), but this lets it wake without crashing.

The workspaces get reordered all incorrectly when waking up (they are all bound to a specific monitor), but for now I can just fix it with a dispatch script.  If there's another way you want it fixed, or more information, let me know.

Ready


vaxerski (approved):
lgtm

PR #2140, opened 2023-04-24
Window resizing for pseudotiled windows

outfoxxed:
Adds resize support to pseudotiled windows. When resizing outside of the window the pseudotiled space is resized, but when resizing inside of the window, the window itself is resized. If the pseudotiled window has a floating size bigger than its pseudotiled area then its floating size is set to the visible pseudotiled size.

Not sure if you use nested structs at all, haven't seen any in hyprland so far. I used one here.

A lot of code is duplicated from `applyNodeDataToWindow`, this could probably be done better.

Yes if the above is not a concern.

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -497,6 +497,12 @@ bool CHyprDwindleLayout::isWindowTiled(CWindow* pWindow) {
     return getNodeFromWindow(pWindow) != nullptr;
 }
 
+void CHyprDwindleLayout::onBeginDragWindow() {
+    m_PseudoDragFlags.started = false;
+    m_PseudoDragFlags.pseudo  = false;
+    IHyprLayout::onBeginDragWindow();
+}
+
 void CHyprDwindleLayout::resizeActiveWindow(const Vector2D& pixResize, CWindow* pWindow) {
 
     const auto PWINDOW = pWindow ? pWindow : g_pCompositor->m_pLastWindow;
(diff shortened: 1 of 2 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -521,6 +527,63 @@ void CHyprDwindleLayout::resizeActiveWindow(const Vector2D& pixResize, CWindow*
     const bool DISPLAYTOP    = STICKS(PWINDOW->m_vPosition.y, PMONITOR->vecPosition.y + PMONITOR->vecReservedTopLeft.y);
     const bool DISPLAYBOTTOM = STICKS(PWINDOW->m_vPosition.y + PWINDOW->m_vSize.y, PMONITOR->vecPosition.y + PMONITOR->vecSize.y - PMONITOR->vecReservedBottomRight.y);
 
+    if (PWINDOW->m_bIsPseudotiled) {
+        if (!m_PseudoDragFlags.started) {
+            m_PseudoDragFlags.started = true;
+
+            const auto PBORDERSIZE = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
```
vaxerski, line 534: 1. missing static
2. any particular reason not to use recalcSizePosRecursive?

outfoxxed: 1. I didn't use it because you didn't seem to use it yourself in most places. (e.g. `applyNodeDataToWindow`)
2. I thought about it for a minute and decided to use `m_vRealSize` since it should already be at the right size before you start dragging it.

vaxerski: 1. did I miss any? cock shit piss fuck

PR #2180, opened 2023-04-28
Declarative plugin management

outfoxxed:
Allows declaring `plugin` entries in the hyprland configuration.
Plugins will be loaded if an entry is added and unloaded if that entry is removed.

Example:
Loading hyprland with
```
plugin = /absolute/path1.so
```
will load the plugin at that path. Removing the entry from the configuration will unload the plugin.
Adding an entry while hyprland is running will cause it to load the plugin.

This is useful for having a nicer config, and decoratively loading plugins.

Nixos example:

You update your system and your hyprland plugin rebuilds. It now has a different path due to the hash of the plugin's sources. After your update your configuration now looks like this

```diff
 # ...
-plugin = /nix/store/jy7lsy0xzyxpkyydk6biql6gpcsg3xh9-hy3-0.1/lib/libhy3.so
+plugin = /nix/store/38a2886e42a9bc982006f6607a983xh9-hy3-0.1/lib/libhy3.so
 # ...
```

Hyprland notices the change and first unloads the old plugin version. then loads the new one.

No

Yes, I'll write the wiki pr shortly.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1107,7 +1107,7 @@ void CConfigManager::handleEnv(const std::string& command, const std::string& va
 
 void CConfigManager::handlePlugin(const std::string& command, const std::string& path) {
     if (std::find(m_vDeclaredPlugins.begin(), m_vDeclaredPlugins.end(), path) != m_vDeclaredPlugins.end()) {
-        parseError = "plugin declared twice";
+        parseError = "plugin '" + path + "' declared twice";
         return;
     }
 
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1974,14 +1974,14 @@ void CConfigManager::handlePluginLoads() {
         return;
 
     bool pluginsChanged = false;
-    auto failedPlugins = g_pPluginSystem->updateConfigPlugins(m_vDeclaredPlugins, pluginsChanged);
+    auto failedPlugins  = g_pPluginSystem->updateConfigPlugins(m_vDeclaredPlugins, pluginsChanged);
 
     if (!failedPlugins.empty()) {
         std::stringstream error;
         error << "Failed to load the following plugins:";
 
         for (auto path : failedPlugins) {
-            error << "\n" << *path;
+            error << "\n" << path;
         }
 
         g_pHyprError->queueCreate(error.str(), CColor(1.0, 50.0 / 255.0, 50.0 / 255.0, 1.0));
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1105,6 +1105,15 @@ void CConfigManager::handleEnv(const std::string& command, const std::string& va
     }
 }
 
+void CConfigManager::handlePlugin(const std::string& command, const std::string& path) {
+    if (std::find(m_vDeclaredPlugins.begin(), m_vDeclaredPlugins.end(), path) != m_vDeclaredPlugins.end()) {
+        parseError = "plugin declared twice";
+        return;
+    }
+
+    m_vDeclaredPlugins.push_back(path);
+}
+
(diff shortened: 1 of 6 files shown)


fufexan (approved):
LGTM

src/plugins/PluginSystem.cpp
```diff
@@ -119,6 +119,35 @@ void CPluginSystem::unloadAllPlugins() {
...
 }
 
+std::vector<std::string*> CPluginSystem::updateConfigPlugins(const std::vector<std::string>& plugins, bool& changed) {
+    std::vector<std::string*> failures;
+
+    // unload all plugins that are no longer present
+    for (auto& p : m_vLoadedPlugins | std::views::reverse) {
+        if (p->m_bLoadedWithConfig && std::find(plugins.begin(), plugins.end(), p->path) == plugins.end()) {
+            Debug::log(LOG, "Unloading plugin %s which is no longer present in config", p->path.c_str());
+            unloadPlugin(p.get(), false);
```
vaxerski, line 129: reallocates m_vLoadedPlugins, segfault.

src/config/ConfigManager.cpp
```diff
@@ -1954,6 +1969,31 @@ void CConfigManager::addExecRule(const SExecRequestedRule& rule) {
...
+        for (auto path : failedPlugins) {
+            error << "\n" << *path;
+        }
+
+        g_pHyprError->queueCreate(error.str(), CColor(1.0, 50.0 / 255.0, 50.0 / 255.0, 1.0));
+    }
+
+    if (pluginsChanged) {
+        g_pHyprError->destroy();
+        m_bForceReload = true;
```
vaxerski, line 1992: y

src/plugins/PluginSystem.cpp
```diff
@@ -119,6 +119,35 @@ void CPluginSystem::unloadAllPlugins() {
         unloadPlugin(p.get(), false); // Unload remaining plugins gracefully
 }
 
+std::vector<std::string*> CPluginSystem::updateConfigPlugins(const std::vector<std::string>& plugins, bool& changed) {
+    std::vector<std::string*> failures;
```
vaxerski, line 123: `std::vector<std::string>`, pointers only incentivise bad stuff later

src/config/ConfigManager.cpp
```diff
@@ -1105,6 +1105,15 @@ void CConfigManager::handleEnv(const std::string& command, const std::string& va
     }
 }
 
+void CConfigManager::handlePlugin(const std::string& command, const std::string& path) {
+    if (std::find(m_vDeclaredPlugins.begin(), m_vDeclaredPlugins.end(), path) != m_vDeclaredPlugins.end()) {
+        parseError = "plugin declared twice";
```
vaxerski, line 1110: would be nice to say what plugin

vaxerski (changes requested):
lgtm my ass

src/plugins/PluginSystem.cpp
outfoxxed: Where is the realloc? If its the `std::erase_if` at the end of `CPluginSystem::unloadPlugin` could you explain how the call in `CPluginSystem::unloadAllPlugins` is not ub? (I copied from there)

src/config/ConfigManager.cpp
outfoxxed: If you're referring to destroying the error, because it won't go away otherwise.
If you're referring to reloading the config, because it won't pick up plugin related settings and dispatchers otherwise.

src/plugins/PluginSystem.cpp
outfoxxed: Replaced with copying

src/config/ConfigManager.cpp
vaxerski: it wont? oh ok

vaxerski (approved):
looks gangsta

PR #2181, opened 2023-04-28
cmake: sync pkg-config --cflags with meson

jbeich:
```
$ git clone https://github.com/hyprwm/hyprland-plugins
$ meson setup /tmp/hyprbars_build hyprland-plugins/hyprbars
$ meson compile -C /tmp/hyprbars_build
c++ -Ilibhyprbars.so.p -I. -Ihyprland-plugins/hyprbars -I/usr/local/include -I/usr/local/include/hyprland/protocols -I/usr/local/include/pixman-1 -I/usr/local/include/libdrm -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -O3 -std=c++2b -fPIC -MD -MQ libhyprbars.so.p/main.cpp.o -MF libhyprbars.so.p/main.cpp.o.d -o libhyprbars.so.p/main.cpp.o -c hyprland-plugins/hyprbars/main.cpp
hyprland-plugins/hyprbars/main.cpp:6:10: fatal error: 'src/Compositor.hpp' file not found
#include <src/Compositor.hpp>
         ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

### DIFF
--- a/hyprland.pc.in
+++ b/hyprland.pc.in
@@ -5,4 +5,4 @@ Name: Hyprland
 URL: https://github.com/hyprwm/Hyprland
 Description: Hyprland header files
 Version: @HYPRLAND_VERSION@
-Cflags: -I"${includedir}/hyprland" -I"${includedir}/hyprland/protocols"
+Cflags: -I"${includedir}" -I"${includedir}/hyprland/protocols"


fufexan:
Is that how it's done in cmake? Paths should always start with `<hyprland/...`, just like any other program's headers.

fufexan (approved):
LGTM

q234rty:
> Can you please remove the `/hyprland` in the first CFlag on this line instead?

Hyprland headers require wlroots headers, which I'm currently installing under `/usr/include/hyprland/wlr` in my aur package. (Otherwise it falls back to the system wlroots headers, which are not necessarily compatible or even exist). Should I move those to `/usr/include/hyprland/protocols/wlr` instead?

PR #2191, opened 2023-04-29
Keep fullscreen mode in moveWindowToWorkspaceSafe

scorpion-26:
Moving a maximized window would always result in the window being fullscreen instead of maximized. This PR keeps the fullscreen mode when transitioning

CWindow doesn't track fullscreen mode, so the mode of the original workspace of the window is used instead. Since only one window per workspace can be fullscreen, this should be fine.

Assumes, that CWindow::m_iWorkspaceID is always valid(Is this a correct assumption?).

Ready


vaxerski (approved):
lgtm

PR #2197, opened 2023-04-30
Honor debug:enable_stdout_logs on startup

scorpion-26:
disableStdout is set via config in CConfigManager::init(), which is called early in CCompositor::initServer(). initServer() always disables stdout logs at the end though, even when stdout is enabled is config. With this commit, the config is respected.

When debug:enable_stdout_logs=0, then nothing from CCompositor::initServer() will be logged to stdout (which it previously did). I can reenable that  behaviour if you want.

Ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1430,6 +1430,8 @@ void CConfigManager::loadConfigLoadVars() {
     }
 
     Debug::disableStdout = !configValues["debug:enable_stdout_logs"].intValue;
+    if (Debug::disableStdout)
+        Debug::log(LOG, "Disabling stdout logs! Check the log for further logs.");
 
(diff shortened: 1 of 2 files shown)


src/config/ConfigManager.cpp
```diff
@@ -1430,6 +1430,8 @@ void CConfigManager::loadConfigLoadVars() {
     }
 
     Debug::disableStdout = !configValues["debug:enable_stdout_logs"].intValue;
+    if (Debug::disableStdout)
```
vaxerski, line 1433: shouldn't it also be && isFirstLaunch?

Otherwise this will be spammed every time the cfg is reloaded.

scorpion-26: Oh, I missed that. Should be fixed now.

vaxerski (approved):
gansta

PR #2198, opened 2023-04-30
Add "next on monitor or empty" workspace parameter

scorpion-26:
Implements the following workspace parameter:
r+x/r-x (i.e. r+1): Behaves similar to the "m" parameter, but can also select empty workspaces and it doesn't wrap around

Example behaviour (Status: x[0-9] -> current workspace on mon [0-9]; o[0-9] -> open workspace on mon [0-9]; e -> empty workspace): 
Initial workspaces: 
|ID | 1| 2| 3| 4 | 5|
|----|---|---|---|---|---|
|Status | o1 | x1 | e | x2 | e |

- with "workspace r+1" on monitor 1:

|ID | 1| 2| 3| 4 | 5 |
|----|---|---|---|---| --- |
|Status | o1 | o1 | x1 | x2 | e |

- with "workspace m+1" on monitor 1 (Notice the difference in "x1"):

|ID | 1| 2| 3| 4 | 5 |
|----|---|---|---|---| --- |
|Status | x1 | o1 | e | x2 | e |

- with "workspace r+2" on monitor 1:

|ID | 1| 2| 3| 4 | 5 |
|----|---|---|---|---|---|
|Status | o1 | o1 | e | o2 | x1 |

The "next or empty" equivalent for "e" is not implemented, since that is the same as +x/-x

Ready. Although I already tested it with multiple scenarios, further testing would probably be good.

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -2,6 +2,7 @@
 #include "../defines.hpp"
 #include <algorithm>
 #include "../Compositor.hpp"
+#include <set>
 #include <sys/utsname.h>
 #include <iomanip>
 #include <sstream>
@@ -305,9 +306,96 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
         outName = PLASTWORKSPACE->m_szName;
         return PLASTWORKSPACE->m_iID;
     } else {
-        if ((in[0] == 'm' || in[0] == 'e' || in[0] == 'r') && (in[1] == '-' || in[1] == '+') && isNumber(in.substr(2))) {
+        if (in[0] == 'r' && (in[1] == '-' || in[1] == '+') && isNumber(in.substr(2))) {
+            if (!g_pCompositor->m_pLastMonitor) {
+                Debug::log(ERR, "Relative monitor workspace on monitor null!");
+                result = INT_MAX;
+                return result;
+            }
+            result = (int)getPlusMinusKeywordResult(in.substr(1), 0);
+
+            int           remains = (int)result;
+
+            std::set<int> invalidWSes;
+
+            // Collect all the workspaces we can't jump to.
+            for (auto& ws : g_pCompositor->m_vWorkspaces) {
+                if (ws->m_bIsSpecialWorkspace || (ws->m_iMonitorID != g_pCompositor->m_pLastMonitor->ID)) {
+                    // Can't jump to this workspace
+                    invalidWSes.insert(ws->m_iID);
+                }
+            }
+            for (auto& [ws, mon] : g_pConfigManager->getAllBoundWorkspaces()) {
+                const auto PMONITOR = g_pCompositor->getMonitorFromString(mon);
+                if (!PMONITOR || PMONITOR->ID == g_pCompositor->m_pLastMonitor->ID) {
+                    // Can't be invalid
+                    continue;
+                }
+                const auto WSNAME     = ws.find("name:") == 0 ? ws.substr(5) : ws;
+                const auto PWORKSPACE = g_pCompositor->getWorkspaceByName(WSNAME);
+                if (PWORKSPACE) {
+                    // WS is bound to another monitor, can't jump to this
+                    invalidWSes.insert(PWORKSPACE->m_iID);
+                }
+            }
+
+            // Just take a blind guess at where we'll probably end up
+            int  predictedWSID = g_pCompositor->m_pLastMonitor->activeWorkspace + remains;
+            int  remainingWSes = 0;
+            char walkDir       = in[1];
+            if (predictedWSID <= 0) {
+                // We overshot, get first valid
+                predictedWSID = 0;
+                walkDir       = '+';
+                remainingWSes = 1;
+            } else {
+                // Count how many invalidWSes are in between (how bad the prediction was)
+                int  beginID = in[1] == '+' ? g_pCompositor->m_pLastMonitor->activeWorkspace + 1 : predictedWSID;
+                int  endID   = in[1] == '+' ? predictedWSID : g_pCompositor->m_pLastMonitor->activeWorkspace;
+                auto begin   = invalidWSes.upper_bound(beginID - 1); // upper_bound is >, we want >=
+                for (auto it = begin; *it <= endID && it != invalidWSes.end(); it++) {
+                    remainingWSes++;
+                }
+            }
+
+            // Go in the search direction for remainingWSes
+            // This takes at most 2 * invalidWSes.size() steps(A workspace cannot be both behind predictedWSID and after predictedWSID.
+            // This is because the loop takes at most ([number of invalidWSes between beginID and endID] + [number of invalidWSes after/before endID]) steps and it is potentially run twice)
+            int finalWSID = predictedWSID;
+            if (walkDir == '-') {
+                int beginID = finalWSID;
+                int curID   = finalWSID;
+                while (--curID > 0 && remainingWSes > 0) {
+                    if (invalidWSes.find(curID) == invalidWSes.end()) {
+                        remainingWSes--;
+                    }
+                    finalWSID = curID;
+                }
+                if (invalidWSes.find(finalWSID) != invalidWSes.end()) {
+                    // Couldn't find valid workspace in negative direction, search last first one back up positive direction
+                    walkDir = '+';
+                    // We know, that everything less than beginID is invalid, so don't bother with that
+                    finalWSID     = beginID;
+                    remainingWSes = 1;
+                }
+            }
+            if (walkDir == '+') {
+                int curID = finalWSID;
+                while (++curID < INT32_MAX && remainingWSes > 0) {
+                    if (invalidWSes.find(curID) == invalidWSes.end()) {
+                        remainingWSes--;
+                    }
+                    finalWSID = curID;
+                }
+            }
+
+            result                = finalWSID;
+            const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(result);
+            if (PWORKSPACE)
+                outName = g_pCompositor->getWorkspaceByID(result)->m_szName;
+
+        } else if ((in[0] == 'm' || in[0] == 'e') && (in[1] == '-' || in[1] == '+') && isNumber(in.substr(2))) {
             bool onAllMonitors = in[0] == 'e';
-            bool includeEmpty  = in[0] == 'r';
 
             if (!g_pCompositor->m_pLastMonitor) {
                 Debug::log(ERR, "Relative monitor workspace on monitor null!");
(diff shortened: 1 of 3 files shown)


vaxerski:
I think I need something strong to read this one

scorpion-26:
@vaxerski Friendly request for review or NAK

src/helpers/MiscFunctions.cpp
```diff
@@ -330,8 +331,28 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+                // Include all empty workspaces in between the known workspaces
+                // We cannot just add all empty workspaces beforehand, since then validWSes would be potentially of size 2^31-1
+                for (auto it = validWSes.begin(); it != validWSes.end() - 1; ++it) {
+                    // add all empty workspaces in between current and last
+                    int curValidWSID  = *it;
+                    int nextValidWSID = *(it + 1);
+                    // ]curValidWSID, nextValidWSID[
+                    for (int id = curValidWSID + 1; id < nextValidWSID; id++) {
+                        const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(id);
+                        if (!PWORKSPACE || (g_pCompositor->getWindowsOnWorkspace(id) == 0 && PWORKSPACE->m_iMonitorID == g_pCompositor->m_pLastMonitor->ID)) {
```
vaxerski, line 344: not checking for whether it's bound to a different monitor

src/helpers/MiscFunctions.cpp
```diff
@@ -346,14 +367,54 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+            } else {
+                if (currentItem >= (int)validWSes.size()) {
+                    // out-of-bounds in positive dir, find the nth empty workspace
+                    int id            = validWSes.back();
+                    int remainingWSes = currentItem - validWSes.size(); // How many we still need to jump
+                    for (int i = 0; i <= remainingWSes; i++) {
+                        // Repeatedly find an empty workspace until we have found enough
+                        while (++id < INT_MAX) {
+                            const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(id);
+                            if (!PWORKSPACE || (g_pCompositor->getWindowsOnWorkspace(id) == 0 && PWORKSPACE->m_iMonitorID == g_pCompositor->m_pLastMonitor->ID)) {
```
vaxerski, line 385: ditto

src/helpers/MiscFunctions.cpp
```diff
@@ -346,14 +367,54 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+                    currentItem = validWSes.size() - 1;
+                } else if (currentItem < 0) {
+                    // out-of-bounds in negative, find the nth empty workspace (or stay on the first valid one, if there are none empty)
+                    int id            = validWSes.front();
+                    int remainingWSes = -currentItem;
+                    for (int i = 0; i <= remainingWSes; i++) {
+                        // Repeatedly find an empty workspace until we have found enough
+                        while (--id > 0) {
+                            const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(id);
+                            if (!PWORKSPACE || (g_pCompositor->getWindowsOnWorkspace(id) == 0 && PWORKSPACE->m_iMonitorID == g_pCompositor->m_pLastMonitor->ID)) {
```
vaxerski, line 402: ditto

src/helpers/MiscFunctions.cpp
```diff
@@ -346,14 +367,54 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
-            } else if (currentItem < 0) {
-                currentItem = validWSes.size() + currentItem;
+            if (!includeEmpty) {
+                if (currentItem >= (int)validWSes.size()) {
+                    currentItem = currentItem % validWSes.size();
+                } else if (currentItem < 0) {
+                    currentItem = validWSes.size() + currentItem;
+                }
+            } else {
+                if (currentItem >= (int)validWSes.size()) {
```
vaxerski, line 377: can't we somehow unify this, and possibly get rid of the while loop? It has caused issues in the past if someone had e.g. workspace `12378723` open

src/helpers/MiscFunctions.cpp
```diff
@@ -330,8 +331,28 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
 
+            if (includeEmpty && validWSes.size() >= 2) {
+                // Include all empty workspaces in between the known workspaces
+                // We cannot just add all empty workspaces beforehand, since then validWSes would be potentially of size 2^31-1
+                for (auto it = validWSes.begin(); it != validWSes.end() - 1; ++it) {
+                    // add all empty workspaces in between current and last
+                    int curValidWSID  = *it;
+                    int nextValidWSID = *(it + 1);
+                    // ]curValidWSID, nextValidWSID[
+                    for (int id = curValidWSID + 1; id < nextValidWSID; id++) {
```
vaxerski, line 342: issue: if someone opens workspace id `3483267` this will literally hang for a couple seconds

vaxerski (changes requested):

src/helpers/MiscFunctions.cpp
scorpion-26: Can a nullptr workspace be bound on another monitor? Otherwise its already being checked.

src/helpers/MiscFunctions.cpp
scorpion-26: Good catch. I was thinking of only searching up to |remains| empty workspaces in the direction we're looking in. This should prevent hangs (unless someone calls hyprctl dispatch r+30000 or so, but I don't think that is realistic)

src/helpers/MiscFunctions.cpp
scorpion-26: For 'r' and 'e' we can just wrap around and call it a day. For 'r' we don't wrap around and need to consider the possibility, that someone calls e.g. r+5 at the last open workspace. So I don't see a way to unify this (Unless you mean the two direction code paths). 

Concerning the while loop. AFAIK, not an issue. For this code to be relevant, we need to be currently on e.g. ws ```123456```, we only need to find at most ```|remains|``` workspaces. As above, this will only hang if someone calls ```r+123456```

src/helpers/MiscFunctions.cpp
vaxerski: yes, a workspace can be closed but bound. Check CConfigManager::getBoundMonitorForWS

scorpion-26: Thanks for tip, I just copied what was in l.325 (I guess that's faulty too...)

src/helpers/MiscFunctions.cpp
scorpion-26: I've thought of another way of implementing that doesn't have the mentioned issues.

scorpion-26:
Implemented a better version, that fixes any performance problems. The performance of the implementation is *only* dependant on the number of active workspaces plus workspace rules. Neither the magnitude of a workspace ID nor the magnitude of the dispatcher argument should impact the performance. Ridiculous parameters like r+300000 work without problems. The implementation now also respects the new(?) workspace rules.

src/helpers/MiscFunctions.cpp
```diff
@@ -305,7 +306,91 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+                }
+                if (invalidWSes.find(finalWSID) != invalidWSes.end()) {
+                    // Couldn't find valid workspace in negative direction, search last first one back up positive direction
+                    walkDir = '+';
+                    // We know, that everything less than beginID is invalid, so don't bother with that
+                    finalWSID     = beginID;
+                    remainingWSes = 1;
+                }
+            }
+            if (walkDir == '+') {
```
vaxerski, line 378: else if

src/helpers/MiscFunctions.cpp
```diff
@@ -305,7 +306,91 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+                        remainingWSes--;
+                    }
+                    finalWSID = curID;
+                }
+            }
+
+            result                = finalWSID;
+            const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(result);
+            if (PWORKSPACE)
+                outName = g_pCompositor->getWorkspaceByID(result)->m_szName;
```
vaxerski, line 391: `else outName = std::to_string(finalWSID);`

we can't leave outName empty.

src/helpers/MiscFunctions.cpp
```diff
@@ -305,7 +306,91 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+                if (!PMONITOR || PMONITOR->ID == g_pCompositor->m_pLastMonitor->ID) {
+                    // Can't be invalid
+                    continue;
+                }
+                // WS is bound to another monitor, can't jump to this
+                invalidWSes.insert(rule.workspaceId);
+            }
+
+            // Just take a blind guess at where we'll probably end up
+            int  predictedWSID = g_pCompositor->m_pLastMonitor->activeWorkspace + remains;
```
vaxerski, line 339: well, the issue here is what about named workspaces? 

Let's say we're on `name:lol` which is the first named workspace and we want to go `r+3`. This will just send us to `-1334` which is completely invalid.

Maybe we should make `r` behave like `m` on named?

vaxerski (changes requested):

src/helpers/MiscFunctions.cpp
scorpion-26: Nope, that's intentional, since we modify walkDir to '+' in l. 372

src/helpers/MiscFunctions.cpp
scorpion-26: Done!

src/helpers/MiscFunctions.cpp
scorpion-26: Sounds reasonable. Can it be assumed, that all negative workspace ids are named?

scorpion-26: Now implemented. Currently only open named workspaces are considered (I don't know the conditions for which are named workspaces are "created" but stay empty) and they are logically "behind" the normal workspaces.

vaxerski:
this is my proposal

check if it works, and if you're happy with it. Cleaner IMHO.

vaxerski:
nit with my method:

r-10 on 5 goes to 1. Probably better this way

scorpion-26:
Your implementation does not work as intended (Bug 1): 
Take this configuration for example: 
Shot_230507_010505

*Green: Active on DP-1, Blue: Active on HDMI-2*
$mainMod-N (r+1) on DP-1 results in no-op. Expected is WS 3

vaxerski:
https://user-images.githubusercontent.com/43317083/236649889-c310f935-92a4-44c0-aa7c-a4c2009de7b7.mp4

gimme sec

scorpion-26:
Wait, I have more: 
Bug 2:
Shot_230507_011513

r+1 on DP-1 goes to WS 5, expected WS 4

If you then call r-1 on DP-1 on WS 5, WS 2 on HDMI-A-2 is focused, expected WS 4

vaxerski:
odd, created 2,3,4 like you, with 2 on a different mon and r+1 on 3 goes to 4 for me

scorpion-26:
> odd, created 2,3,4 like you, with 2 on a different mon and r+1 on 3 goes to 4 for me

Works now on latest commit

But I still found *at least* 2 bugs again. 

Is there anything particularly wrong with my version :disappointed: ?

vaxerski:
not inherently, just bad readability. We can revert to it if mine is so buggy, I guess.

scorpion-26:
I can try to improve the readability, though the poor readability stems mostly from the many edge cases. I've merged my solution into my personal daily-driving branch (https://github.com/scorpion-26/Hyprland/tree/personalfixes) and I've yet to encounter any bugs.

vaxerski:
feel free to revert my commits and I'll try to review this tomorrow

scorpion-26:
FYI: I've pushed your solution to the emptyrelws-vaxry branch on my fork, in case it is ever needed again.

vaxerski (approved):
lgtm. If you could make an accompanying wiki MR plox. Thanks for the work

vaxerski:
splendid

PR #2204, opened 2023-05-01
Prefer bundled wlroots headers to system ones

q234rty:

In the case that the prefix is `/usr`, system wlroots headers installed in `/usr/include/wlr` will be preferred over those bundled by hyprland as `-I` directories are searched [from left to right](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html), which is not great since the system wlroots headers might not be compatible with hyprland. Fix the order of cflags so bundled wlroots headers will be preferred over system ones.

No.

Ready for merge.


vaxerski (approved):
looks gangsta

PR #2209, opened 2023-05-01
Unbreak CMake build on FreeBSD

jbeich:
Most incompatibilities are caused by custom Makefile used to drive CMake.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,20 +23,20 @@ clear:
 	rm -rf ./subprojects/wlroots/build
 
 all:
-	make clear
-	make fixwlr
+	$(MAKE) clear
+	$(MAKE) fixwlr
 	cd ./subprojects/wlroots && meson setup build/ --buildtype=release && ninja -C build/ && cp ./build/libwlroots.so.12032 ${PREFIX}/lib/ || echo "Could not install libwlroots to ${PREFIX}/lib/libwlroots.so.12032"
 	cd subprojects/udis86 && cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -S . -B./build -G Ninja && cmake --build ./build --config Release --target all -j$(shell nproc)
-	make release
-	make -C hyprctl all
+	$(MAKE) release
+	$(MAKE) -C hyprctl all
 
 install:
-	make clear
-	make fixwlr
+	$(MAKE) clear
+	$(MAKE) fixwlr
 	cd ./subprojects/wlroots && meson setup build/ --buildtype=release && ninja -C build/ && cp ./build/libwlroots.so.12032 ${PREFIX}/lib/ || echo "Could not install libwlroots to ${PREFIX}/lib/libwlroots.so.12032"
 	cd subprojects/udis86 && cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -S . -B./build -G Ninja && cmake --build ./build --config Release --target all -j$(shell nproc) && cd ../..
-	make release
-	make -C hyprctl all
+	$(MAKE) release
+	$(MAKE) -C hyprctl all
 
 	mkdir -p ${PREFIX}/share/wayland-sessions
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -104,14 +104,13 @@ pluginenv:
 	mkdir -p ${PREFIX}/include/hyprland/wlroots
 	mkdir -p ${PREFIX}/share/pkgconfig
 	
-	find src -name '*.h*' -exec cp --parents '{}' ${PREFIX}/include/hyprland ';'
-	cd subprojects/wlroots/include && find . -name '*.h*' -exec cp --parents '{}' ${PREFIX}/include/hyprland/wlroots ';' && cd ../../..
+	find src -name '*.h*' -print0 | cpio --quiet -0dump ${PREFIX}/include/hyprland
+	cd subprojects/wlroots/include && find . -name '*.h*' -print0 | cpio --quiet -0dump ${PREFIX}/include/hyprland/wlroots && cd ../../..
 	cp ./protocols/*-protocol.h ${PREFIX}/include/hyprland/protocols
 	cp ./build/hyprland.pc ${PREFIX}/share/pkgconfig
-	cp ./build/hyprland.pc /usr/share/pkgconfig
```
vaxerski, line 111: this is actually quite important because pkg-config will not find the .pc file in `/usr/local/share/pkgconfig`

jbeich: If your pkg-config is broken set `PKG_CONFIG_PATH=/usr/local/share/pkgconfig`.

jbeich: I'm not sure what ArchLinux uses but [pkgconf](https://git.sr.ht/~kaniini/pkgconf) built with `--with-pkg-config-dir=/usr/local/share/pkgconfig` will look there by default.

jbeich: With non-default PREFIX pkg-config may not look there by default. For example, after `make install PREFIX=/tmp/test` pkg-config won't find `hyprland.pc` unless `PKG_CONFIG_PATH=/tmp/test/share/pkgconfig` is set via environ(7).

jbeich: Why not relocate `hyprland.pc` from `share/pkgconfig` to `lib/pkgconfig`? For example, `wayfire.pc` is installed under `lib/pkgconfig`, so it's probably more compatible with various distros.

jbeich: FWIW, Wayfire has header-only API just like Hyprland.

vaxerski: sure, as long as it works.

jbeich: Unfortunately, FreeBSD looks under `libdata/pkgconfig` instead of `lib/pkgconfig` by default. Let's keep 2 copies of `hyprland.pc` instead of special-casing by OS.

jbeich: I can't test runtime on Linux, including whether pkg-config still finds `hyprland.pc`.

vaxerski:
request a review from me when you think this is ready, cuz I see you making a lot of commtis

jbeich:
This was ready from the start. I simply can't test on Linux to check what's **not** covered by CI.

vaxerski:
Alright, we'll need to export the pkg-config environ cuz `/usr/local/lib/pkgconfig` is not used by arch either... ugh

jbeich:
I can make the error non-fatal. It'd create unused directory outside of PREFIX if run as root, so have to be connected to `uninstall` target.

vaxerski (approved):
looks gangsta to me I'll let fuf also take a look if I didn't miss anything.

Note to self: update wiki

vaxerski:
> I can make the error non-fatal. It'd create unused directory outside of PREFIX if run as root, so have to be connected to uninstall target.

error? I mean, copying it automatically to /usr/share/pkgconfig would be nice, do what you need I guess

vaxerski:
ye thats cool. @fufexan review when you got time

fufexan (approved):
Looks alright from a bird's-eye view, but I think squashing some of the commits would be a good idea.

PR #2210, opened 2023-05-01
Fix recursion in package overlays

spikespaz:
@fufexan 

**Edit:**

It started with the first commit but I've pushed more work and the PR needs some explanation.

# Description

Previously, the way the flake constructed its `packages` output was to call the `overlays.default` output with `null` for the fixed-point and `prev` for the instance of `pkgs` to modify. Instead of using packages from `final` to use in the derivations overridden from `prev`, it would use the `rec` keyword to use packages adjacent.

Now, `packages` is constructed with a new instance of `pkgs` for both `final` and `prev`, and the overlays themselves use `final` and `prev` correctly. This was done so that some flake packages could depend on other packages also exported by the flake. However, because `packages` was created through the use of `overlays`, this is incorrect and defies the purpose of the overlay mechanism.

This new instance of `pkgs` that overlays receive will include all other dependency packages exposed by the flake (and its inputs, in the case of `hyprland-protocols`). This means that anywhere a package was previously used to create an overridden derivation, where it was before using `rec`, it now properly uses the package from `final` recursively, properly utilizing fixed-point extension.

# Why?

The dependency graph was incorrect, and overlay packages instantiated incorrectly.

This fixes the behavior of both overlays and usage of packages, as well as usage of a combination of the two. Previously, Hyprland could not be built if the flake's `default` overlay was applied to current `nixpkgs/nixos-unstable`, because of a version conflict with `wayland-server` (needed `1.22`, has `1.21`). Now the `wayland-latest` overlay will be applied to packages from the `packages` output, and additionally the other overlays from the flake can be used separately without rebuilding everything in *nixpkgs* that depends on `wayland`.

# Is it ready to merge?

Yes

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,7 +41,6 @@
           self.overlays.wlroots-hyprland
           self.overlays.wayland-latest
           inputs.hyprland-protocols.overlays.default
-          inputs.xdph.overlays.default
         ];
       });
 
--- a/flake.nix
+++ b/flake.nix
@@ -36,10 +36,10 @@
       import nixpkgs {
         inherit system;
         overlays = [
-          (_: prev: {
+          (final: prev: {
             wayland = prev.wayland.overrideAttrs (old: rec {
               version = "1.22.0";
-              src = prev.fetchurl {
+              src = final.fetchurl {
                 url = "https://gitlab.freedesktop.org/wayland/wayland/-/releases/${version}/downloads/${old.pname}-${version}.tar.xz";
                 hash = "sha256-FUCvHqaYpHHC2OnSiDMsfg/TYMjx0Sk267fny8JCWEI=";
               };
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -56,88 +55,125 @@
...
-          self.packages.${system}.wlroots-hyprland
-        ];
-        inputsFrom = [
-          self.packages.${system}.wlroots-hyprland
-          self.packages.${system}.hyprland
-        ];
-      };
+      default =
+        pkgsFor.${system}.mkShell.override {
+          stdenv = pkgsFor.${system}.gcc12Stdenv;
```
fufexan, line 167: We can remove this line and others referring to gcc12Stdenv, since it's now the default.

fufexan (approved):
Overall looks good to me. Maybe we could separate the overlays into a `overlays.nix` to keep it clean?

spikespaz (commented):
This will break people's configs, but I don't think it should have been here in the first place. I will re-export the inputs' packages in `hyprland-extras`.

spikespaz:
Currently [fd132e5](https://github.com/hyprwm/Hyprland/pull/2210/commits/fd132e588dba9a2dd6839a6075c628d6f9441f85) is causing
```
jacob-thinkpad% nix eval 'path:.#packages'
error: getting status of '/nix/store/3yw6yxcnd2g4nvib8kycqp07c6nkas9g-source/nix/nix': No such file or directory
```
Do not know why

fufexan (approved):
lookin good

PR #2217, opened 2023-05-02
only ignore no_gaps_when_only when the workspace rule specifies a border

jacekpoz:
fixes a bug introduced in c949173bc9c9f9e8276cba098cb82ed51ad4965b that drew the borders with no_gaps_when_only=1, even if the current workspace rule didn't specify a border


vaxerski (approved):
gangsta

PR #2227, opened 2023-05-03
quote hash for the GIT_COMMIT_HASH macro

solopasha:
otherwise compilation fails with smth like
```
../src/debug/../plugins/../defines.hpp:1:25:
error: ‘f27873a6f06dc2f87600edb890f3c38298bfb55f’ was not declared in this scope
```


vaxerski:
dangit, my bad, I'll edit the source tarball

PR #2254, opened 2023-05-05
renderer,config: add custom DRM modeline support

Akaricchi:

This allows specifying custom display resolutions for the DRM backend. This is useful for display overclocking, working around broken EDIDs, etc. To use this feature, specify a modeline instead of a resolution in the config, for example:

```
monitor = DP-1, modeline 1071.101 3840 3848 3880 3920 2160 2263 2271 2277 +hsync -vsync, 0x0, 1
```

This example is a custom 3840x2160@120Hz mode with tightened timings. I use it because the standard timings don't work with my monitor and GPU combination (M28U with RX580).

The syntax is compatible with Sway and Xorg.

I've been daily driving this for a few weeks and haven't ran into any issues.

I think it is, but probably needs a documentation update as well.


src/config/ConfigManager.cpp
```diff
@@ -469,6 +469,80 @@ void CConfigManager::handleRawExec(const std::string& command, const std::string
     g_pKeybindManager->spawn(args);
 }
 
+static bool parseModeLine(const std::string& modeline, drmModeModeInfo& mode) {
+    std::vector<std::string> args;
+    std::stringstream ss;
+
+    for (auto it = modeline.begin();; it++) {
```
vaxerski, line 476: useless, you can use `CVarList(modeline, 0, ' ')`

src/render/Renderer.cpp
```diff
@@ -1515,6 +1515,10 @@ bool CHyprRenderer::applyMonitorRule(CMonitor* pMonitor, SMonitorRule* pMonitorR
         force = true;
     }
 
+    if (pMonitorRule->drmMode.type == DRM_MODE_TYPE_USERDEF) {
```
vaxerski, line 1518: why? Just store the last modeline

vaxerski (changes requested):
Also please format the file with the .clang-format

src/config/ConfigManager.cpp
Akaricchi: I thought about it, but the problem with CVarList is that only one character can be a separator. Someone trying to use tabs might get an unexpected surprise (which is also inconsistent with the parser's behavior in other contexts, e.g. even CVarList itself strips both tabs and spaces *around* the separators).

If you are fine with that I can remove the complicated loop, though in my personal opinion it's poor UX.

vaxerski: or you could've just added the functionality to cvarlist. Anyways, I've done that. Rebase against master and use `'s'` as the separator

src/render/Renderer.cpp
Akaricchi: I didn't want to touch the scary monster if-condition right below. I thought it not important because re-applying the same mode doesn't seem to do anything anyway (that I know of). What do you think of simply removing that whole already-applied check for simplicity?

vaxerski: it's there for a reason, it sets frame skips, monitor reloads, layer reloads, damage and stuff. The "monster if" is not that scary... store the modeline in a string and compare it

Akaricchi:
Changed `parseModeLine` to use `CVarList`, implemented same-rule check for custom modelines, and ran clang-format.

Instead of comparing modeline strings, I opted to store `drmModeModeInfo` in `CMonitor` and `memcmp` it with the version from the rule. It's slightly more technically correct, because multiple modeline spellings can parse into identical parameters. It also doesn't require keeping the unparsed modeline in the rule.

Clang-format also caught some stuff not related to my changes, but I filtered them out in the interest of keeping the diff clean.

src/config/ConfigManager.cpp
```diff
@@ -470,6 +470,60 @@ void CConfigManager::handleRawExec(const std::string& command, const std::string
     g_pKeybindManager->spawn(args);
 }
 
+static bool parseModeLine(const std::string& modeline, drmModeModeInfo& mode) {
+    auto args = CVarList(modeline, 0, 's');
+
+    auto keyword = args[0];
+    std::transform(keyword.begin(), keyword.end(), keyword.begin(), ::tolower);
+
+    if (keyword != "modeline") {
```
vaxerski, line 479: style: no braces around single ifs

src/config/ConfigManager.cpp
```diff
@@ -470,6 +470,60 @@ void CConfigManager::handleRawExec(const std::string& command, const std::string
...
+
+    if (args.size() < 10) {
+        Debug::log(ERR, "modeline parse error: expected at least 9 arguments, got %i", args.size() - 1);
+        return false;
+    }
+
+    int argno = 1;
+
+    mode.type        = DRM_MODE_TYPE_USERDEF;
+    mode.clock       = stof(args[argno++]) * 1000;
```
vaxerski, line 491: `std::`

vaxerski (changes requested):
minor style changes, rest lgtm

Akaricchi:
Done, though I have to say this same file is inconsistent about both of those things.

vaxerski:
if it is you can blame my small brane

vaxerski (approved):
gangsta, thanks!

vaxerski:
forgot, could you make a wiki mr? plox

Akaricchi:
I'm a little busy currently, but I'll get to that soon-ish.

PR #2292, opened 2023-05-13
Fix not finding function symbols for hooking

### LINKED ISSUE #2177: Plugin api will not find function to hook if running from tty
When run from a tty, as the root compositor, hyprland fails to find functions to hook. It works as expected in a nested hyprland session (using the same version and plugin .so)

### Details
The function `HyprlandAPI::findFunctionsByName` returns 0 functions in the root instance and 1 function (as expected) in a nested instance.

### Reproducing
Call `HyprlandAPI::findFunctionsByName(PHANDLE, "updateWindowAnimatedDecorationValues")` from both a tty and nested inside another hyprland instance. One function will be returned from the nested instance and zero from the tty.

I have a plugin that demonstrates this behavior here: https://github.com/outfoxxed/hy3/tree/e41c3cb2562456f30d406d2ebe6217c41f30a615 (linked at current commit).
To reproduce with the plugin set `general:layout = hy3`. Load both from a tty and a nested instance.

outfoxxed:
Fixes #2177.

Adds a new `execAndGet` function that puts its STDERR into the log to make it more discoverable when child processes fail. This also fixes not being able to run `nm` on the hyprland binary if the path contains spaces.

Wraps the nix binary to include `binutils` (`nm`) as function symbols were missing because it was not in path.

There may or may not be a C++ stdlib function that actually implements child process pipes. I was not able to find it if there is.
I'm fairly sure every case in `execAndGet` is handled, including exceptions, but there may be something I missed.

I have tested it, and it's ready if you have no issues with the implementation.


nix/default.nix
```diff
@@ -113,6 +116,11 @@ in
         }'
       '';
 
+      postInstall = ''
+        wrapProgram $out/bin/Hyprland \
```
FlafyDev, line 120: Maybe it would be better to separate the wrapping into a different derivation. So:

`hyprland-unwrapped` - regular hyprland
`hyprland` - wrapped hyprland

This way people who don't want to use the wrapper or want to provide their own, will be able to opt out of it.

many packages do that in nixpkgs: https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=-unwrapped

nix/default.nix
```diff
@@ -113,6 +116,11 @@ in
...
+          --suffix PATH : ${lib.makeBinPath [ binutils ]}
```
FlafyDev, line 121: I think this would also need `pciutils`

FlafyDev (changes requested):
not sure about the c++ part, there seem to be some formatting changes in ead61aa, is that intentional?

nix/default.nix
outfoxxed: Will do

nix/default.nix
outfoxxed: Wasn't sure if i should add it since its less of a hard dependency than binutils. Only used for printing stats on launch and in the crash report

outfoxxed:
The formatting changes were just the result of letting clang-format do its thing

nix/default.nix
FlafyDev: pciutils is currentlyin bulidinputs, not sure why it's there, don't think it's doing anything..

outfoxxed: Probably lspci

FlafyDev: it needs lspci in PATH at runtime(like nm). adding it to buildinputs doesn't do that

vaxerski:
I am against the trash that is the new command. We don't need stderr.

outfoxxed:
Alright, that seemed like the "right" way to do it. Here's the easy way.

vaxerski (approved):
gangsta from me, dk about nix

nix/default.nix
fufexan: The current impl looks good to me. Users can still override hyprland as usual, and only those who want to wrap it themselves have to use an extra override option.

nix/default.nix
```diff
@@ -13,6 +14,7 @@
   libxcb,
   libxkbcommon,
   mesa,
+  makeWrapper,
```
fufexan, line 17: I'd put this between `pkg-config` and `meson`, since it's a `nativeBuildInput`. I tend to put those higher in the set.

nix/default.nix
```diff
@@ -113,6 +117,11 @@ in
         }'
       '';
 
+      postInstall = lib.optional wrapRuntimeDeps ''
```
fufexan, line 120: I think this needs a rewrite now that I've merged https://github.com/hyprwm/Hyprland/commit/78826c6d18ee587a28cf39479f9a015888d2e096.

fufexan (changes requested):

nix/default.nix
outfoxxed: alright, wasn't sure how it was organized so I just put them in where it seemed correct

nix/default.nix
outfoxxed: Yep looks like it

fufexan (approved):

PR #2299, opened 2023-05-14
make split preselect available in config file

cassii2:
A little over a week ago split preselect was added, but the only way to use it was using the "layoutmsg" dispatcher in the config file, so I've added a normal "preselect" dispatcher

I did change DwindleLayout.cpp so that it will split arguments on ',' instead of ' ', so you can use the "layoutmsg" dispatcher with comma separated arguments instead of space separated ones. This can break your config file if it still uses the "layoutmsg" dispatcher with space-separated arguments.

Ready for merging

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -841,7 +841,7 @@ void CHyprDwindleLayout::alterSplitRatio(CWindow* pWindow, float ratio, bool exa
 }
 
 std::any CHyprDwindleLayout::layoutMessage(SLayoutMessageHeader header, std::string message) {
-    const auto ARGS = CVarList(message, 0, ',');
+    const auto ARGS = CVarList(message, 0, ' ');
     if (ARGS[0] == "togglesplit") {
         toggleSplit(header.pWindow);
     } else if (ARGS[0] == "preselect") {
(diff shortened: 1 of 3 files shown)


vaxerski:
> I did change DwindleLayout.cpp so that it will split arguments on ',' instead of ' ', so you can use the "layoutmsg" dispatcher with comma separated arguments instead of space separated ones. This can break your config file if it still uses the "layoutmsg" dispatcher with space-separated arguments.

dont

cassii2:
alright, changed it back

vaxerski:
but why would you? It was clearly stated in the mr afaik that since it's a layout thing, it should not have a global dispatcher.

cassii2:
honestly i just downloaded hyprland today and was trying to set up my config file and it was listed on the wiki here https://wiki.hyprland.org/hyprland-wiki/pages/Configuring/Dwindle-Layout/#bind-dispatchers where the other two in the list are global dispatchers but it didn’t actually work when i tried it as one, so i made this pull request for it

vaxerski:
it's a doc bug. The other two are dispatchers for historic reasons.

vaxerski:
https://github.com/hyprwm/hyprland-wiki/commit/4d8b0f7f5bc74798154babe297e18f30d9c1ffae

PR #2313, opened 2023-05-16
Improvements to CMakeLists

m00nwtchr:
This PR brings (as far as i can tell) all of the functionality of the Makefile into CMake, which makes it more automated and friendly to package maintainers, and it follows the typical CMake project convention.

Example build:
```sh
$ git clone <Hyprland>
$ cmake -B build -S Hyprland
$ cmake --build build
```
If desired, can be installed with `cmake --install build`

~~It should be ready to go as is.~~
Nvm, i found one issue

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,9 +79,9 @@ ExternalProject_Add(wlroots_build
     PREFIX wlroots
     SOURCE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots
     PATCH_COMMAND ${SED_EXE} -E -i -e "s/(soversion = 12)([^032]|$$)/soversion = ${WLROOTS_SOVERSION}/g" ${CMAKE_SOURCE_DIR}/subprojects/wlroots/meson.build
-    CONFIGURE_COMMAND ${MESON_EXE} setup --prefix ${CMAKE_CURRENT_BINARY_DIR}/wlroots . ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    CONFIGURE_COMMAND ${MESON_EXE} setup --prefix "/" . ${CMAKE_SOURCE_DIR}/subprojects/wlroots
     BUILD_COMMAND ${NINJA_EXE} -C .
-    INSTALL_COMMAND ${NINJA_EXE} -C . install
+    INSTALL_COMMAND ${MESON_EXE} install --destdir ${CMAKE_CURRENT_BINARY_DIR}/wlroots
 	BUILD_BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
 
 set(wlroots_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots/include/)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,8 @@ ExternalProject_Add(wlroots_build
     PATCH_COMMAND ${SED_EXE} -E -i -e "s/(soversion = 12)([^032]|$$)/soversion = ${WLROOTS_SOVERSION}/g" ${CMAKE_SOURCE_DIR}/subprojects/wlroots/meson.build
     CONFIGURE_COMMAND ${MESON_EXE} setup --prefix ${CMAKE_CURRENT_BINARY_DIR}/wlroots . ${CMAKE_SOURCE_DIR}/subprojects/wlroots
     BUILD_COMMAND ${NINJA_EXE} -C .
-    INSTALL_COMMAND ${NINJA_EXE} -C . install)
+    INSTALL_COMMAND ${NINJA_EXE} -C . install
+	BUILD_BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
 
 set(wlroots_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots/include/)
 set(wlroots_LIBRARY ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,12 +83,12 @@ ExternalProject_Add(wlroots
     BUILD_COMMAND ${NINJA_EXE} -C .
     INSTALL_COMMAND ${NINJA_EXE} -C . install)
 
-set(wlroots_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/wlroots/include/)
-set(wlroots_LIB_PATH ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
+set(wlroots_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots/include/)
+set(wlroots_LIBRARY ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
 
 add_library(wlroots_lib SHARED IMPORTED)
 set_target_properties(wlroots_lib PROPERTIES
-    IMPORTED_LOCATION ${wlroots_LIB_PATH}
+    IMPORTED_LOCATION ${wlroots_LIBRARY}
     INTERFACE_INCLUDE_DIRECTORIES ${wlroots_INCLUDE_DIR})
 
 add_subdirectory("subprojects/udis86" udis86)
@@ -200,7 +200,7 @@ target_link_libraries(Hyprland
 )
 
 install(TARGETS Hyprland hyprctl DESTINATION bin)
-install(FILES ${wlroots_LIB_PATH} DESTINATION lib)
+install(FILES ${wlroots_LIBRARY} DESTINATION lib)
 install(DIRECTORY assets/ DESTINATION share/hyprland PATTERN "meson.build" EXCLUDE)
 install(FILES example/hyprland.desktop DESTINATION share/wayland-sessions)
 install(FILES example/hyprland.conf DESTINATION share/hyprland)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.19)
 include(CheckIncludeFile)
+include(ExternalProject)
 
 # Get version
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/props.json PROPS)
@@ -12,7 +13,7 @@ project(Hyprland
 
 set(HYPRLAND_VERSION ${VER})
 set(PREFIX ${CMAKE_INSTALL_PREFIX})
-configure_file(hyprland.pc.in hyprland.pc @ONLY) 
+configure_file(hyprland.pc.in hyprland.pc @ONLY)
 
 set(CMAKE_MESSAGE_LOG_LEVEL "STATUS")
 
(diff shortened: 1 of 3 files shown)


m00nwtchr:
So, the issue is that while this works fine on its own, for some reason under Arch's makepkg I get a build error: `make[2]: *** No rule to make target 'wlroots/lib/libwlroots.so.12032', needed by 'Hyprland'.  Stop.`

m00nwtchr:
Ok, that fixed it

CMakeLists.txt
```diff
@@ -153,35 +171,48 @@ include(CPack)
...
 
+install(TARGETS Hyprland hyprctl DESTINATION bin)
+install(FILES ${wlroots_LIBRARY} DESTINATION lib)
+install(DIRECTORY assets/ DESTINATION share/hyprland PATTERN "meson.build" EXCLUDE)
+install(FILES example/hyprland.desktop DESTINATION share/wayland-sessions)
+install(FILES example/hyprland.conf DESTINATION share/hyprland)
+install(FILES docs/hyprctl.1 docs/Hyprland.1 DESTINATION share/man/man1)
+
+install(DIRECTORY ${wlroots_INCLUDE_DIR}/ DESTINATION include/hyprland/wlroots)
+install(DIRECTORY protocols/ DESTINATION include/hyprland/protocols PATTERN "meson.build" EXCLUDE)
```
vaxerski, line 211: `.xml`s and `.c`s should also be excluded. Ideally it should be pattern `.h` only

CMakeLists.txt
```diff
@@ -153,35 +171,48 @@ include(CPack)
...
+
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/hyprland.pc
```
vaxerski, line 213: what is the prefix here? generally hyprland installs itself into `/usr/local`, but this has to be `/usr` (cuz arch)

CMakeLists.txt
```diff
@@ -69,12 +70,31 @@ else()
...
+find_program(MESON_EXE NAMES meson)
+find_program(NINJA_EXE NAMES ninja)
+find_program(SED_EXE NAMES sed)
+ExternalProject_Add(wlroots_build
+    PREFIX wlroots
+    SOURCE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    PATCH_COMMAND ${SED_EXE} -E -i -e "s/(soversion = 12)([^032]|$$)/soversion = ${WLROOTS_SOVERSION}/g" ${CMAKE_SOURCE_DIR}/subprojects/wlroots/meson.build
+    CONFIGURE_COMMAND ${MESON_EXE} setup --prefix ${CMAKE_CURRENT_BINARY_DIR}/wlroots . ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    BUILD_COMMAND ${NINJA_EXE} -C .
+    INSTALL_COMMAND ${NINJA_EXE} -C . install)
```
vaxerski, line 84: we should not install wlroots. We only copy the `.so`.

CMakeLists.txt
```diff
@@ -69,12 +70,31 @@ else()
...
+set(wlroots_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots/include/)
+set(wlroots_LIBRARY ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
+
+add_library(wlroots SHARED IMPORTED)
+add_dependencies(wlroots wlroots_build)
+set_target_properties(wlroots PROPERTIES
+    IMPORTED_LOCATION ${wlroots_LIBRARY}
+    INTERFACE_INCLUDE_DIRECTORIES ${wlroots_INCLUDE_DIR})
+
+add_subdirectory("subprojects/udis86" udis86)
```
vaxerski, line 95: udis needs to be built, no? or does this build?

https://github.com/hyprwm/Hyprland/blob/642030f959efe7bd2692bf6a4236d4c1b39ee51a/Makefile#L92

vaxerski (changes requested):

CMakeLists.txt
m00nwtchr: The install command is only used to copy wlroots (headers and lib) files to another location, it doesn't get installed in the system.

CMakeLists.txt
m00nwtchr: It adds udis' cmake project to the main project, and then later its added as a dependency (`target_link_libraries`) which causes it to be built automatically by cmake.

CMakeLists.txt
m00nwtchr: It's `/usr/local/` still, but that's fine, because directories under there have (almost) identical meanings to `/usr`, and packagers can override the prefix.

CMakeLists.txt
vaxerski: uhh, `ninja -C build install` would install the headers to `/usr` which we don't want, no?

CMakeLists.txt
vaxerski: pkg-config is the issue here, it doesn't include /usr/local for some reason on arch, so we want to install this one to /usr.

m00nwtchr: That will be handled in the PKGBUILD, see: https://github.com/hyprwm/aur/blob/5e0fce6d9454d8ffe7e3e9ac50b37854ccec1b55/PKGBUILD-git#L84

CMakeLists.txt
vaxerski: I don't believe this is resolved

m00nwtchr: The setup command sets the prefix to a location within the cmake build folder.

vaxerski: right, my bad

CMakeLists.txt
vaxerski: still don't believe this is a good idea, we should do both

CMakeLists.txt
```diff
@@ -153,35 +171,53 @@ include(CPack)
...
+install(FILES docs/hyprctl.1 docs/Hyprland.1 DESTINATION share/man/man1)
+
+if(NO_DEV)
+    message(STATUS "Development headers disabled")
+else()
+    message(STATUS "Development headers enabled")
+    install(DIRECTORY ${wlroots_INCLUDE_DIR}/ DESTINATION include/hyprland/wlroots)
+    install(DIRECTORY protocols/ DESTINATION include/hyprland/protocols FILES_MATCHING PATTERN "*.h")    
+    install(DIRECTORY src/ DESTINATION include/hyprland/src FILES_MATCHING PATTERN "*.h*")
+    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/hyprland.pc
```
vaxerski, line 217: all is good when this gets installed to `/usr` too. we might just execute a command here, the one that was in the makefile

vaxerski (changes requested):

CMakeLists.txt
m00nwtchr: This already does the same thing as: ` cp ./build/hyprland.pc /usr/share/pkgconfig`

m00nwtchr: It doesn't matter. CMake shouldn't install anything to `/usr` by default, and packagers can change the location. (As can users, if they do want to do this, for some reason)

vaxerski: the thing is, I don't care whether it should or not. It has to, otherwise pkg-config will not pick up the .pc file.

m00nwtchr: That's what package managers are for.

CMakeLists.txt
m00nwtchr: 1. Then those packages need to be fixed.
2. Installing manually is a bad idea (and if someone wants to do it, they can just add the correct arguments to CMake)

vaxerski:
Would be nice if it actually succeeded through the pipelines. If you do `make clear` and then `make release` it fails because ninja can't find libwlroots

vaxerski:
Reverting my commits will not get you anywhere as this will not be merged without .pc installing to /usr/share.

m00nwtchr:
The Makefile will install to /usr/share now

Makefile
```diff
@@ -24,48 +24,12 @@ clear:
...
-	cd subprojects/wlroots/include && find . -name '*.h*' -print0 | cpio --quiet -0dump ${PREFIX}/include/hyprland/wlroots && cd ../../..
-	cd subprojects/wlroots/build/include && find . -name '*.h*' -print0 | cpio --quiet -0dump ${PREFIX}/include/hyprland/wlroots && cd ../../../..
-	cp ./protocols/*-protocol.h ${PREFIX}/include/hyprland/protocols
-	cp ./build/hyprland.pc ${PREFIX}/share/pkgconfig
-	if [ -d /usr/share/pkgconfig ]; then cp ./build/hyprland.pc /usr/share/pkgconfig 2>/dev/null || true; fi
-
-cleaninstall:
-	echo -en "$(MAKE) cleaninstall has been DEPRECATED, you should avoid using it in the future.\nRunning $(MAKE) install instead...\n"
-	$(MAKE) install
+	cmake --install build --prefix "/usr"
```
vaxerski, line 32: fun of you to disallow commits from me on my own project.

Hyprland installs to /usr/local, .pc to /usr.

```
cmake --install build
if [ -d /usr/share/pkgconfig ]; then cp ./build/hyprland.pc /usr/share/pkgconfig 2>/dev/null || true; fi
```
if anything.

vaxerski (changes requested):

stephan-cr (commented):
@m00nwtchr I had the same idea and then I found this PR. :-) Left some comments and hope they're helpful.

CMakeLists.txt
```diff
@@ -69,12 +70,32 @@ else()
...
+ExternalProject_Add(wlroots_build
+    PREFIX wlroots
+    SOURCE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    PATCH_COMMAND ${SED_EXE} -E -i -e "s/(soversion = 12)([^032]|$$)/soversion = ${WLROOTS_SOVERSION}/g" ${CMAKE_SOURCE_DIR}/subprojects/wlroots/meson.build
+    CONFIGURE_COMMAND ${MESON_EXE} setup --prefix "/" . ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    BUILD_COMMAND ${NINJA_EXE} -C .
+    INSTALL_COMMAND ${MESON_EXE} install --destdir ${CMAKE_CURRENT_BINARY_DIR}/wlroots
+	BUILD_BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
+
+set(wlroots_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots/include/)
```
stephan-cr, line 87: I guess it should be something like:

```cmake
ExternalProject_Get_property(wlroots_build INSTALL_DIR)
set(wlroots_INCLUDE_DIR ${INSTALL_DIR}/include/)
```
I think that is reason why the build currently fails. `config.h` is a generated file and is either in the build directory or in the final "installation" directory.

CMakeLists.txt
```diff
@@ -69,12 +70,32 @@ else()
...
+find_program(NINJA_EXE NAMES ninja)
+find_program(SED_EXE NAMES sed)
+ExternalProject_Add(wlroots_build
+    PREFIX wlroots
+    SOURCE_DIR ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    PATCH_COMMAND ${SED_EXE} -E -i -e "s/(soversion = 12)([^032]|$$)/soversion = ${WLROOTS_SOVERSION}/g" ${CMAKE_SOURCE_DIR}/subprojects/wlroots/meson.build
+    CONFIGURE_COMMAND ${MESON_EXE} setup --prefix "/" . ${CMAKE_SOURCE_DIR}/subprojects/wlroots
+    BUILD_COMMAND ${NINJA_EXE} -C .
+    INSTALL_COMMAND ${MESON_EXE} install --destdir ${CMAKE_CURRENT_BINARY_DIR}/wlroots
+	BUILD_BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/wlroots/lib/libwlroots.so.${WLROOTS_SOVERSION})
```
stephan-cr, line 85: Nit: wrong indention?

stephan-cr (changes requested):
I think I know why the build fails currently.

CMakeLists.txt
```diff
@@ -153,35 +172,53 @@ include(CPack)
 message(STATUS "Setting link libraries")
 
 function(protocol protoPath protoName external)
-    if (external)
+    if(external)
         execute_process(
```
majestrate, line 176: using `add_custom_command()` here to generate these files into `CMAKE_CURRENT_BINARY_DIR` would be nice. this way we could add them to an interface library target that any dependents can `target_link_libraries()` to.

m00nwtchr:
Note to anyone still looking at this: I gave up on having this upstreamed due to the above conversations. But if you still want to continue this you're welcome to fork my fork.

majestrate:
> Note to anyone still looking at this: I gave up on having this upstreamed due to the above conversations. But if you still want to continue this you're welcome to fork my fork.

i think the underlying pain point comes from the lack of separation in the dev branches and any packaging fixups. when packaging logic gets woven so deeply into the main development branch 100% of the time package maintainers have their fixups for their own distros mucking up things where it does not have to.
debian has spinoff branches for packages for this very reason, instead of mucking about in the dev process of everything they package they periodically rebase spinoff branches for applying small commits for JUST packaging. packaging is a full time job and making the life of the package maintainer is just as important as the people working on the bulk of the meat. when package maintainers and core software developers have to care about divergences from any distro getting in their way something is wrong with how packaging is done. if hyprland could do something more manageable like spinoff branches for nix flakes, RPMs, ebuilds, debs or whatever ever packaging is required i think that this problem would go away and everyone ends up happy as a result. 
i would personally like to feel out how people in hyprland feel about that as i think that everyone who wants to contribute but doesnt is influenced by the build system being so convoluted by being intertwined with packaging right now.

m00nwtchr:
> > Note to anyone still looking at this: I gave up on having this upstreamed due to the above conversations. But if you still want to continue this you're welcome to fork my fork.
> 
> i think the underlying pain point comes from the lack of separation in the dev branches and any packaging fixups. when packaging logic gets woven so deeply into the main development branch 100% of the time package maintainers have their fixups for their own distros mucking up things where it does not have to.
> debian has spinoff branches for packages for this very reason, instead of mucking about in the dev process of everything they package they periodically rebase spinoff branches for applying small commits for JUST packaging. packaging is a full time job and making the life of the package maintainer is just as important as the people working on the bulk of the meat. when package maintainers and core software developers have to care about divergences from any distro getting in their way something is wrong with how packaging is done. if hyprland could do something more manageable like spinoff branches for nix flakes, RPMs, ebuilds, debs or whatever ever packaging is required i think that this problem would go away and everyone ends up happy as a result. 
> i would personally like to feel out how people in hyprland feel about that as i think that everyone who wants to contribute but doesnt is influenced by the build system being so convoluted by being intertwined with packaging right now.
> 

That's what I tried to do, the upstreame shouldn't care about the target distribution at all, and it should be left to the packagers, but they insisted on supporting distributions when installing cmake, which is just moronic

majestrate:
> > > Note to anyone still looking at this: I gave up on having this upstreamed due to the above conversations. But if you still want to continue this you're welcome to fork my fork.
> > 
> > 
> > i think the underlying pain point comes from the lack of separation in the dev branches and any packaging fixups. when packaging logic gets woven so deeply into the main development branch 100% of the time package maintainers have their fixups for their own distros mucking up things where it does not have to.
> > debian has spinoff branches for packages for this very reason, instead of mucking about in the dev process of everything they package they periodically rebase spinoff branches for applying small commits for JUST packaging. packaging is a full time job and making the life of the package maintainer is just as important as the people working on the bulk of the meat. when package maintainers and core software developers have to care about divergences from any distro getting in their way something is wrong with how packaging is done. if hyprland could do something more manageable like spinoff branches for nix flakes, RPMs, ebuilds, debs or whatever ever packaging is required i think that this problem would go away and everyone ends up happy as a result.
> > i would personally like to feel out how people in hyprland feel about that as i think that everyone who wants to contribute but doesnt is influenced by the build system being so convoluted by being intertwined with packaging right now.
> 
> That's what I tried to do, the upstream shouldn't care about the target distribution at all, and it should be left to the packagers, but they insisted on supporting distributions when installing to system with cmake, which is just moronic (and makes downstream packaging more difficult)

i personally would like to reduce the friction on this somehow so hyprland can eventually be packaged in debian.

PR #2322, opened 2023-05-17
Don't swap workspaces if monitors are the same

aruhier:
Return swapActiveWorkspaces early if MON1 and MON2 are equals, to avoid buggy behavior.


vaxerski (approved):
fair enoguh

PR #2323, opened 2023-05-17
input: don't refocus when moving window to wp

aruhier:
Fixes a behavior when moving a window to another workspace will sometimes (if the window is the last one of the current workspace) trigger a refocus following the cursor, even when disabled.


vaxerski:
this is intended, you're moving silent. If we skip this, focus might remain on a window that has been silently moved to an invisible workspace.

aruhier:
In which scenarios would it happen? The logic around the focus when sending a window to a workspace, in silent, should be the same as unmapping a window from the workspace, no?

I tried this patch, if I have a workspace WS1 with only one window W1 and I send it to another workspace, the focus stays on the WS1, which is expected.

PR #2339, opened 2023-05-21
Fix broken pipe crash when event listener terminates

maqrrr:

Fixes a case where hyprland can crash due to a broken IPC pipe.

Here's how to reproduce it:
```bash
socat -d -d -d -U - "UNIX-CONNECT:/tmp/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket2.sock" | (while read -r line; do echo "$line"; kill -SIGINT 0; done)
```

Open two windows.  Run that in one.  Change focus (or generate any other IPC event).

This is a race between https://github.com/hyprwm/Hyprland/blob/5627b70981730cf1839ae4477f9fd086d4fc7a6c/src/managers/EventManager.cpp#L27 and https://github.com/hyprwm/Hyprland/blob/5627b70981730cf1839ae4477f9fd086d4fc7a6c/src/managers/EventManager.cpp#L111

The array that keeps track of the FDs does eventually get cleaned up, and maybe this catch block should clean it up immediately, but ignoring it seems ok too.

Ready for merging.

### DIFF
--- a/src/managers/EventManager.cpp
+++ b/src/managers/EventManager.cpp
@@ -108,7 +108,9 @@ void CEventManager::flushEvents() {
     for (auto& ev : m_dQueuedEvents) {
         std::string eventString = (ev.event + ">>" + ev.data).substr(0, 1022) + "\n";
         for (auto& fd : m_dAcceptedSocketFDs) {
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
gangsta, thanks

PR #2373, opened 2023-05-27
Add hyprland to waybar-hyprland path as fallback for hyprctl

Sorixelle:
In some environments, such as using the [home-manager waybar module's systemd integration](https://github.com/nix-community/home-manager/blob/master/modules/programs/waybar.nix#L198), hyprctl may not be in `PATH`. As a result, waybar crashes when attempting to call out to hyprctl to get information for eg. the currently active workspace.

This PR adds a wrapper to the waybar binary, appending hyprctl to `PATH`. This allows waybar to use the current environment's hyprctl, if one exists, and falls back to the default hyprland package if there is none.

This might not be the most efficient way to do this, as it always pulls the `hyprland` derivation no matter what the user is actually using. People using patched versions of hyprland (eg. `hyprland-nvidia`) would end up pulling two copies of Hyprland into the store. While (to my knowledge) this shouldn't break anything in waybar, this does result in a potentially unnecessary copy of Hyprland being added to the store. I'm open to suggestions on improving this.

Assuming the caveat I described above is fine, this should be ready to merge.

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -56,6 +56,10 @@ in {
           # use hyprctl to switch workspaces
           sed -i 's/zext_workspace_handle_v1_activate(workspace_handle_);/const std::string command = "hyprctl dispatch workspace " + name_;\n\tsystem(command.c_str());/g' src/modules/wlr/workspace_manager.cpp
         '';
+        postFixup = ''
+          wrapProgram $out/bin/waybar \
+            --suffix PATH : ${lib.makeBinPath [ prev.hyprland ]}
+        '';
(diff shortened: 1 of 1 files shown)


vaxerski:
@fufexan for nix uwu

fufexan:
I agree having multiple copies of hyprland is unnecessary. This could be easily solved by wrapping waybar inside a HM/NixOS module for example. However including such an option in the modules isn't ideal, since only few of the users would actually use it.

fufexan:
I'll merge this. Better to have a working bar, despite the double copy of Hyprland.

PR #2445, opened 2023-06-05
wip: FEATURE Scaling for rounded borders

kleinschrader:
My PR (once it is done) plans on adding a scaling option to the radius of rounded corners based, for example, on the location of the corner (top left etc...) or if its a corner on the outside or inside.

The reason i went with an approach of scaling rather then overriding specific corners is that one can simply multiply multiple corner modifiers together for a simpler and non compatibility breaking  configuration.

So far my code is horrendous due to the fact that i'm having trouble understanding the current code base. 

It needs a lot of work, im just opening the PR now if someone wants to help me in this venture.


vaxerski:
closing due to inactivity

Issue #1794, opened 2023-03-15
Wayland windows created on 100% scale doesn't scale on it's own to 200%


grappas:
https://cdn.discordapp.com/attachments/995047159542063154/1085673198827356210/2023-03-15_21-49-35.mkv

^^^ note when I grab window with mouse and put back it sharpens.

Steps to reproduce:
1. Have two monitors with different scaling.
2. Create window on lower scaled monitor.
3. Use shortcut (movewindow, movetoworkspace) to higher dpi monitor.

Expected behaviour:
Window is rescaled to 200%

Noted outcome:
Window is blurry like a xwayland app. It rescales only when is dragged with mouse. Window created in higher dpi monitor isn't affected.

vaxerski:
[patch.txt](https://github.com/hyprwm/Hyprland/files/10985548/patch.txt)

can you check this

grappas:
> 

do you mean compile recent git repo?

fberg:
I think I'm running into the same bug with Alacritty, see https://github.com/alacritty/alacritty/issues/6973. (Also happens if the scales are 1.0 and 2.0 as in the OP.)

Moving the window with the mouse works as expected, but `movecurrentworkspacetomonitor` using a keyboard shortcuts leads to blurry fonts. Changing the scale on the monitor with the terminal window (by editing the config file) also has the same effect.

This is on Hyprland 0.25.0 and also on the newest revision (9cf72a30fc040f0bd2fb4651ec7f8e82df962278).

PR #2447, opened 2023-06-05
Fix fractional scale updates in some cases

fberg:

The fractional scale of windows was not updated on "keyboard" moves (`movecurrentworkspacetomonitor` and `movetoworkspace` if the target workspace is on a different monitor), while it worked as expected when dragging a window with the mouse. Similar to https://github.com/hyprwm/Hyprland/issues/1794. Although that issue was already fixed, I am observing the same behavior on current `main`.

This is my first time doing anything with Hyprland (or even Wayland for that matter), so not sure if I did it right.

In `CWindow::moveToWorkspace()`, the scale used appears to be the one from the previous monitor (where the window used to be), but I'm not sure if there are any unintended side-effects from setting `m_iMonitorID`. If that is bad, the scaling behavior was also fixed by only creating `PMONITOR` from `PWORKSPACE->m_iMonitorID`.

Apart from the above disclaimer, I think it's ready (for what it's worth). I tested it with two monitors of different scale (2.0 and 1.6), and alacritty as client.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1929,6 +1929,8 @@ void CCompositor::moveWorkspaceToMonitor(CWorkspace* pWorkspace, CMonitor* pMoni
             }
 
             w->updateToplevel();
+
+            g_pProtocolManager->m_pFractionalScaleProtocolManager->setPreferredScaleForSurface(w->m_pWLSurface.wlr(), pMonitor->scale);
         }
     }
 
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -291,8 +291,11 @@ void CWindow::moveToWorkspace(int workspaceID) {
 
     m_iWorkspaceID = workspaceID;
 
-    const auto PMONITOR   = g_pCompositor->getMonitorFromID(m_iMonitorID);
     const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(m_iWorkspaceID);
+
+    m_iMonitorID = PWORKSPACE->m_iMonitorID;
+
+    const auto PMONITOR   = g_pCompositor->getMonitorFromID(m_iMonitorID);
 
     if (PWORKSPACE) {
         g_pEventManager->postEvent(SHyprIPCEvent{"movewindow", getFormat("%lx,%s", this, PWORKSPACE->m_szName.c_str())});


src/Compositor.cpp
```diff
@@ -1929,6 +1929,8 @@ void CCompositor::moveWorkspaceToMonitor(CWorkspace* pWorkspace, CMonitor* pMoni
             }
 
             w->updateToplevel();
+
+            g_pProtocolManager->m_pFractionalScaleProtocolManager->setPreferredScaleForSurface(w->m_pWLSurface.wlr(), pMonitor->scale);
```
vaxerski, line 1933: this should be done in updateToplevel, it already has a func for this, but ignores the fractional scale protocol

vaxerski (changes requested):

src/Compositor.cpp
fberg: That makes sense.

>it already has a func for this

Do you mean `updateSurfaceOutputs()`?

Since `updateTopLevel` is also called with the `movetoworkspace` dispatcher, I have moved all of the fractional scale code from `moveToWorkspace` (including the part from 70eb74c356d291f4b5e048abef7772864a4f95f1) to `updateSurfaceOutputs`.

vaxerski: setpreferredscaleforsurface is a noop if nothing changed so it doesnt matter if it gets called more often than needed tbh

src/Window.cpp
```diff
@@ -283,6 +283,19 @@ void CWindow::updateSurfaceOutputs() {
         wlr_surface_for_each_surface(m_pWLSurface.wlr(), sendLeaveIter, PLASTMONITOR->output);
 
     wlr_surface_for_each_surface(m_pWLSurface.wlr(), sendEnterIter, PNEWMONITOR->output);
+
+    g_pProtocolManager->m_pFractionalScaleProtocolManager->setPreferredScaleForSurface(m_pWLSurface.wlr(), PNEWMONITOR->scale);
+
+    if (!m_bIsMapped)
```
vaxerski, line 289: why is this here, its already checked on l273

src/Window.cpp
```diff
@@ -283,6 +283,19 @@ void CWindow::updateSurfaceOutputs() {
         wlr_surface_for_each_surface(m_pWLSurface.wlr(), sendLeaveIter, PLASTMONITOR->output);
 
     wlr_surface_for_each_surface(m_pWLSurface.wlr(), sendEnterIter, PNEWMONITOR->output);
+
+    g_pProtocolManager->m_pFractionalScaleProtocolManager->setPreferredScaleForSurface(m_pWLSurface.wlr(), PNEWMONITOR->scale);
```
vaxerski, line 287: you do for each surface later this is usels

vaxerski (changes requested):

vaxerski (approved):
gangsta

PR #2470, opened 2023-06-09
main: fix segfault when -c is given with no other arguments

eriedaberrie:
Has the proper check to prevent `Hyprland -c` from segfaulting.

I'm not gonna lie I'm not totally sure what the preexisting code was trying to do but I don't think it was working so I fixed it.

Should be good to go.


vaxerski (approved):
oops

PR #2478, opened 2023-06-09
feat: add lockactivegroup dispatcher

memchr:

The `lockactivewindow` dispatcher takes `lock`, `toggle` or `unlock` as arguments. When a group is locked, no window or group can be added to it, nor can it be added to another group, but the `moveintogroup` and `moveoutofgroup` dispatches are not affected. Has no effect on non-group.

#### Implementation details:

the lock is implement via `SGroupData.locked` flag (defaults to false).

The flag is only relevant to the group head, and upon the group head's succession, the flag will be passed down to the new head. Meanwhile, the old head's flag will be set to false.

The flag is set to false when a group is dismissed.

New condition checks have been added to the dwindle and master layout to check if target group is unlocked (and if the source is also a group and unlocked) before adding windows to the target group.

tested locally.

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -645,6 +645,7 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
             curr                            = curr->m_sGroupData.pNextWindow;
             PLAST->m_sGroupData.pNextWindow = nullptr;
             PLAST->m_sGroupData.head        = false;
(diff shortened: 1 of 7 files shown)


vaxerski (approved):
fair enough, make a wiki mr pls doe

memchr:
Submitted

PR #2495, opened 2023-06-10
zoom: multiply by scale only on mouseZoomUseMouse

nik012003:
Sorry, I didn't realize my previous commit skewed the zoom when mouseZoomUseMouse was false. 
Now everything should work right :3


vaxerski (approved):
tard

PR #2523, opened 2023-06-15
Fix crash when screen size is 0x0

shemig:
Fix crash when screen size is 0x0 (When booting into laptop clam mode in dell XPS)
and also ignore any screen with size 0x0 in the first place

# Crash fix
When I boot into clam mode (lid is closed) on my laptop, the laptop screen is recognized as size 0x0.
Hyprland starts normally and treats it as a regular screen (which i cannot use of course).
Then I can crash it in two ways:
1. Try to disable the monitor (to move all workspaces to the external monitor)
2. Switch to a workspace on the laptop monitor, and close a window.

I get the following error:
```
ASSERTION FAILED! 

cannot alloc a FB with negative / zero size! (attempted 0x0)

at: line 6 in Framebuffer.cpp
```

## Additional work
Since there is no point in using a 0x0 screen,  listener_newOutput will ignore such screens

### DIFF
--- a/src/events/Monitors.cpp
+++ b/src/events/Monitors.cpp
@@ -57,6 +57,11 @@ void Events::listener_newOutput(wl_listener* listener, void* data) {
         return;
     }
 
+    if (OUTPUT->width <= 0 || OUTPUT->height <= 0) {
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
extra

PR #2524, opened 2023-06-15
fix revision getting updated to the previous one, not current

solopasha:


vaxerski:
@fufexan for nix

fufexan:
Lol I wonder how it kept working without errors for so long.

Issue #2537, opened 2023-06-17
hyprland doesn't recognise monitors anymore in docking station setup. 


MangoIV:
This must've broken some time in the last two weeks. I sadly cannot give you the exact commit that broke it due to me having commited my setup often enough. 

I just wanted to leave this here although I don't have much information, I will try to add on some info at some later point in time. 

302ec1372c40cd4130101920c2f6c38cad6fc9dc broke it

MangoIV:
@shemig, @gilad-sx your commit introduced a breaking change to me. My monitors don't start up anymore. 

Just wanted to leave a notice about it :)

shemig:
Oh no,
Can you post the hyprland.log ?
In any case, I will revert the screen selection change and leave only the crash defense, until I understand why

skulwicki:
Just an fyi @shemig, this issue is also noted in #2525 and #2535. It’s happening in general at startup it seems. My setup is a desktop with two monitors and was not getting a graphical environment on startup.

MangoIV:
fixed by 302ec13 , thank you @shemig! 

I will see whether I post the log, I have restarted my PC often enough today (issue doesn't occur if one doesn't restart)

Ahwxorg:
I have something somewhat similar. I boot up my laptop with HDMI connected, works fine, no issues. I disconnect HDMI, use my laptop, works fine, automatically puts all the workspaces on one display, no issues. I connect HDMI again, and then nothing happens. I use `wdisplays` to change display stuff, and it does not detect any other screen than my laptop. Is this what you were talking about? How can I fix this?

PR #2539, opened 2023-06-17
Partial revert of Commit 302ec13:

shemig:
Fix crash when screen size is 0x0 (#2523)

Reason: The disable of a monitor with 0x0 size
is causing issuses with some users.

https://github.com/hyprwm/Hyprland/issues/2537

Left the defensive code to resolve the crash.
Will continue to investigate and find a solution for the dell xps disabled monitor


vaxerski (approved):
ha

PR #2594, opened 2023-06-27
Reloads animated decoration values set on window rules 

MightyPlaza:
Reloads animated decoration values like border color and opacity set on window rule on config reload.
didn't notice it wasn't working last time lol


vaxerski (approved):
k

PR #2624, opened 2023-07-01
make "togglefloating" functional on windows in special workspaces

CoolElectronics:
Currently, if you have the special workspace active and you attempt to switch it from floating to tiled or vice versa with the "togglefloating" dispatch, it will do nothing. In order to fix a window you sent to a special workspace that was floating, you would have to send it back to a normal workspace, toggle it there, then send it back to the special workspace, possibly messing up the existing layout. 
This adds the much more more intuitive behavior of allowing you to switch a window between floating and tiled within the special workspace.

All changes appear to work as expected, though this is my first time submitting a patch so I'm not sure if there are code style issues that need to be resolved.


vaxerski:
floating windows on special would require way more work than this. It is planned, though.

dranull:
This seems to be implemented now.

Issue #1595, opened 2023-02-20
Grouping and ungrouping window causes weird state


Merlin04:
If you do these steps:
- open two windows
- group the focused one
- switch focus to the other one, group and ungroup it
- move it into the group with the first window
Then you'll have a group with two windows in it, but the little tab indicator at the top doesn't update when you switch between them, and when you create a window in the group with window 1 focused, it will make it so you can only switch between and access window 1 and that new window.

MightyPlaza:
Never mind was testing with the mouse instead of the keyboard. Yep this is an issue.

MightyPlaza:
also fixed @vaxerski

PR #2630, opened 2023-07-03
groupbar fixes

MightyPlaza:
Fixes multiple groupbar decoration issues:
- togglegroup removes fullscreen to avoid to avoid weird state
- fixes issue where a group had multiple windows with head = true
- fixes issue where merging 2 groups would cause a window to have 2 groupbar decorations
- fixes issue where merging a group with more than 1 window into another group would make windows have no groupbar decoration
- fixes issue where ungrouping windows could just move them into another group on the same workspace  

probably more but I don't remember

Dwindle Layout as been tested and seems stable and working as intended.
Master Layout seem to work as well, but has been less tested and I'm less familiar with how it works. 

Known issues:
- Still haven't found the cause of the focused window groupbar not getting correctly updated when there are multiple groups open side by side (text gets streched) (was already on main branch). Probably related to multiple groupbar damage issues open here.

- There is also an unusual animation the first time opening a window that was in a group after merging it with another group (only on Master Layout). Still haven't found the cause. (was already on main branch as well)

Probably ready to merge, but a bit of testing would be good

Closes:
- https://github.com/hyprwm/Hyprland/issues/1595

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -670,8 +670,9 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
     const auto PTAIL = getGroupTail();
 
     if (pWindow->m_sGroupData.pNextWindow) {
+        const auto SHEAD = pWindow->getGroupHead();
         std::vector<CWindow*> members;
-        CWindow*              curr = pWindow;
+        CWindow*              curr = SHEAD;
         do {
             const auto PLAST = curr;
             members.push_back(curr);
@@ -679,7 +680,7 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
             PLAST->m_sGroupData.pNextWindow = nullptr;
             PLAST->m_sGroupData.head        = false;
             PLAST->m_sGroupData.locked      = false;
-        } while (curr != pWindow);
+        } while (curr != SHEAD);
 
         for (auto& w : members) {
             insertWindowToGroup(w);
@@ -690,8 +691,6 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
 
     PTAIL->m_sGroupData.pNextWindow   = pWindow;
     pWindow->m_sGroupData.pNextWindow = PHEAD;
-
-    setGroupCurrent(pWindow);
 }
 
 void CWindow::updateGroupOutputs() {
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -670,9 +670,9 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
     const auto PTAIL = getGroupTail();
 
     if (pWindow->m_sGroupData.pNextWindow) {
-        const auto SHEAD = pWindow->getGroupHead();
+        const auto            PHEAD = pWindow->getGroupHead();
         std::vector<CWindow*> members;
-        CWindow*              curr = SHEAD;
+        CWindow*              curr = PHEAD;
         do {
             const auto PLAST = curr;
             members.push_back(curr);
@@ -680,7 +680,7 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
             PLAST->m_sGroupData.pNextWindow = nullptr;
             PLAST->m_sGroupData.head        = false;
             PLAST->m_sGroupData.locked      = false;
-        } while (curr != SHEAD);
+        } while (curr != PHEAD);
 
         for (auto& w : members) {
             insertWindowToGroup(w);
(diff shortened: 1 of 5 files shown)


src/Window.cpp
```diff
@@ -670,16 +670,17 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
     const auto PTAIL = getGroupTail();
 
     if (pWindow->m_sGroupData.pNextWindow) {
+        const auto SHEAD = pWindow->getGroupHead();
```
vaxerski, line 673: style: `PHEAD`

src/layout/MasterLayout.cpp
```diff
@@ -91,16 +91,34 @@ void CHyprMasterLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
+            pWindow->m_dWindowDecorations.emplace_back(std::make_unique<CHyprGroupBarDecoration>(pWindow));
+            pWindow->updateWindowDecos();
+            recalculateWindow(pWindow);
 
-        return;
+            g_pCompositor->focusWindow(pWindow);
+            return;
+        }
+
+        if (!pWindow->getGroupHead()->m_sGroupData.locked) { // source is a unlocked group
```
vaxerski, line 111: an

src/layout/DwindleLayout.cpp
```diff
@@ -311,18 +311,33 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
+            pWindow->m_dWindowDecorations.emplace_back(std::make_unique<CHyprGroupBarDecoration>(pWindow));
+            pWindow->updateWindowDecos();
+            recalculateWindow(pWindow);
 
-        return;
+            g_pCompositor->focusWindow(pWindow);
+            return;
+        }
+
+        if (!pWindow->getGroupHead()->m_sGroupData.locked) { // source is a unlocked group
```
vaxerski, line 330: an

src/layout/DwindleLayout.cpp
```diff
@@ -311,18 +311,33 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
     // if it's a group, add the window
-    if (OPENINGON->pWindow->m_sGroupData.pNextWindow && !OPENINGON->pWindow->getGroupHead()->m_sGroupData.locked && // target is an unlocked group
-        (!pWindow->m_sGroupData.pNextWindow || !pWindow->getGroupHead()->m_sGroupData.locked)                       // source is not group or is a unlocked group
-        && !g_pKeybindManager->m_bGroupsLocked) {
-        m_lDwindleNodesData.remove(*PNODE);
+    if (OPENINGON->pWindow->m_sGroupData.pNextWindow && !OPENINGON->pWindow->getGroupHead()->m_sGroupData.locked &&
+        !g_pKeybindManager->m_bGroupsLocked) { // target is an unlocked group
 
-        OPENINGON->pWindow->insertWindowToGroup(pWindow);
+        if (!pWindow->m_sGroupData.pNextWindow) { // source is not group
```
vaxerski, line 317: not a group

src/layout/MasterLayout.cpp
```diff
@@ -91,16 +91,34 @@ void CHyprMasterLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
-    if (OPENINGON && OPENINGON->pWindow->m_sGroupData.pNextWindow && !OPENINGON->pWindow->getGroupHead()->m_sGroupData.locked && // target is an unlocked group
-        (!pWindow->m_sGroupData.pNextWindow || !pWindow->getGroupHead()->m_sGroupData.locked)                                    // source is not group or is an unlocked group
-        && OPENINGON != PNODE && !g_pKeybindManager->m_bGroupsLocked) {
-        m_lMasterNodesData.remove(*PNODE);
+    // if it's a group, add the window
+    if (OPENINGON && OPENINGON->pWindow->m_sGroupData.pNextWindow && !OPENINGON->pWindow->getGroupHead()->m_sGroupData.locked && !g_pKeybindManager->m_bGroupsLocked &&
+        OPENINGON != PNODE) { // target is an unlocked group
 
-        OPENINGON->pWindow->insertWindowToGroup(pWindow);
+        if (!pWindow->m_sGroupData.pNextWindow) { // source is not group
```
vaxerski, line 98: not a group

vaxerski (commented):
also format with clang-format pls

vaxerski:
ready to merge to me, the stretching issue is to be fixed separately, I can already see why it happens

MightyPlaza:
PHEAD has already defined in that function as the HEAD of the target group, SHEAD was the the head of the source group.
Altough it doesn't cause a conflict for now, might cause confusion in the future.

PR #2657, opened 2023-07-06
Cleanup compositor deadcode

nexec:

1. Removes redundant code from main, as it is a part of Compositor::cleanup
2. Unificates systemd notifications along main/Compositor

No

Ready

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -11,10 +11,6 @@
 #include <string>
 #include <filesystem>
 
-#ifdef USES_SYSTEMD
-#include <systemd/sd-daemon.h> // for sd_notify
-#endif
-
 void help() {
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -102,17 +98,5 @@ int main(int argc, char** argv) {
     // If we are here it means we got yote.
     Debug::log(LOG, "Hyprland reached the end.");
 
-#ifdef USES_SYSTEMD
```
vaxerski, line 105: why is all of this removed?

vaxerski: ah, right it's in the destructor my bad

vaxerski (approved):
gang gang

PR #2676, opened 2023-07-10
Add support for smart splitting

DanielAdolfsson:

Imagine a square or rectangular box. Picture two diagonal lines crossing each other to form an 'X' that divides the box into four equal sections or triangles. Each triangle is an equal quarter of the box.

This PR adds support for splitting depending on which section the mouse cursor is currently over. Personally, I find this slightly more intuitive.

A new configuration option has been added `dwindle:smart_split`, which has to be set to `1` in order for this functionality to be enabled.

Not that I know- but I'm new to the codebase so feel free to check!

Maybe give it a different name. "smart_split" isn't very smart ;)

Also, maybe a second option to only enabling this when *moving* a window?

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -157,7 +157,7 @@ void CConfigManager::setDefaultVars() {
     configValues["dwindle:no_gaps_when_only"].intValue            = 0;
     configValues["dwindle:use_active_for_splits"].intValue        = 1;
     configValues["dwindle:default_split_ratio"].floatValue        = 1.f;
-    configValues["dwindle:smart_split"].intValue                  = 1;
+    configValues["dwindle:smart_split"].intValue                  = 0;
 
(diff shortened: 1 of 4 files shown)


vaxerski (commented):
not formatted with clang-format

also, wouldn't this require preserve_split?

DanielAdolfsson:
> not formatted with clang-format
Alright!

> also, wouldn't this require preserve_split?
Yeah I believe so. It will probably feel weird with it off. Maybe preserve_split should be implied when smart_split is set?

vaxerski:
> Yeah I believe so. It will probably feel weird with it off. Maybe preserve_split should be implied when smart_split is set?

Yeah, we can do that.

DanielAdolfsson:
This should work?

DanielAdolfsson:
Created PR hyprwm/hyprland-wiki#266 for the option. 
Been testing this for a few hours at this point as I'm running this code live. Haven't ran into any issues yet!

PR #2678, opened 2023-07-11
monitor desc default workspace but working now

MightyPlaza:
same has previous PR but now works outside of embedded sessions and doesn't crash anymore on nullptr


vaxerski (approved):
I need to start testing MRs myslef

nwg-piotr:
I could only test on a very ancient machine at the office so far, but it looks all right. Many thanks!

Issue #2112, opened 2023-04-20
Resizing with mouse unintuitive if you grab the 'wrong' side of the window


Ashe:
Hello, I was wondering if this is feasible - currently, resizing a tiled window is really unintuitive in comparison to a floating window, as pointed out by PR #1347 .

When floating, the position of the cursor vs the container matters - if you click on the right half of the window it'll grow the window if you drag right, and shrink if you go left. If you grab the left half, it does the opposite.

When tiles however, you can grab the 'wrong' side of a window and the resize mechanism is the opposite. It also makes it hard to change the 'splits', since it means you have to figure out which window controls said split.

If you enable `resize_on_border` the issue is worsened - depending on which part of the gap you drag, you will either get the right or wrong functionality. I'd expect dragging the gap to always allow you to move the split, rather than disturbing some other split. This is just me considering horizontal splits, but as soon as you include vertical the problem gets worse as it could be the correct window for one and not the other.

I believe this is the case in `sway` and even though I hate to do direct comparisons, I do think it would make this a little nicer to daily drive.

Thank you for your hard work :slightly_smiling_face: 

---

### Summary

* Please make it so that the mouse's position is taken into account when resizing a tiled window such that you are stretching the window with the mouse:
  * Dragging right on the right side of the window will grow, and left would shrink.
  * This will mimic the way it's done for floating windows
* Visualised another way, it should look like you're manipulating the splits between windows:
  * Find the closest window corner to where the cursor clicks and change the position of the splits such that the window's corner follows the mouse

Thanks!

Ashe:
Another weird thing I've encountered is with `hyprctl`. 

On a floating window, `hyprctl dispatch resizeactive exact 0 0` makes it tiny, as expected. Likewise, `hyprctl dispatch resizeactive exact 2560 1440` sets it to that size. Perfect!

On a tiled window however, these commands do the *opposite*. `hyprctl dispatch resizeactive exact 2560 1440` shrinks the window into the bottom right of my screen, and `hyprctl dispatch resizeactive exact 0 0` makes the window take up almost half my screen. I know tiled windows may not be able to reach the values you input, but it would be nice to try and grow and shrink other windows in available directions to reach it. For instance, if there's only a row of windows, then `exact` can't really change the vertical height as it'll be clamped to monitor size.

vaxerski:
[...]

PR #2681, opened 2023-07-11
Dwindle: Make resize more intuitive

DanielAdolfsson:

This improves resizing to make it a little more intuitive. It allows you to to change the size by dragging any corner of the window while anchoring the opposite side.

I think this issue is related:
https://github.com/hyprwm/Hyprland/issues/2112

Not currently.

Couple of questions:

Should this behavior be opt-in? 
When using keyboard resizing, should we use the old behavior?

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -672,41 +672,43 @@ void CHyprDwindleLayout::resizeActiveWindow(const Vector2D& pixResize, eRectCorn
         const auto PPARENT = PCURRENT->pParent;
 
         if (!PVOUTER && PPARENT->splitTop && ((TOP && PPARENT->children[1] == PCURRENT) || (BOTTOM && PPARENT->children[0] == PCURRENT)))
-            PVOUTER = PPARENT;
-        else if (!PVOUTER && !PVINNER && PPARENT->splitTop && ((TOP && PPARENT->children[0] == PCURRENT) || (BOTTOM && PPARENT->children[1] == PCURRENT)))
-            PVINNER = PPARENT;
+            PVOUTER = PCURRENT;
+        else if (!PVOUTER && !PVINNER && PPARENT->splitTop)
+            PVINNER = PCURRENT;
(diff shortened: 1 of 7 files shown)


vaxerski:
with keyboard resizing we should keep how it used to be imo, you can achieve that by adding a CORNER_NONE to the enum and using that as default

DanielAdolfsson:
Sure! The original code only allows two axis of freedom if the parent's immediate parent has another orientation. My proposal is we keep traversing the tree (as in the new behavior) to find a suitable parent.

Not exactly what you asked for, but maybe good anyway. Up to you!

vaxerski (approved):
looks alright

Ashe:
This works really well!! Thank you so much! 

I wanted to make a note of some behaviour I found though. Given this layout:

```
   |   |   |
 1 |   | 3 |
   |   |   |
---| 2 |---|
   |   |   |
 4 |   | 5 |
   |   |   |
```
Dragging any corners of window `2` works as expected, only the the vertical split between `2` and the side splits (`1` and `4`, or `3` and `5`) moves, leaving the other split and the opposite windows in the exact the same positions and size.

However, dragging on the 'inner' corners of the side windows (`1`, `4`, `3` or `5`) *can* result in the opposite windows being affected. The corner that triggers this is decided by the order you spawn and the way you position windows, but for example right now if I drag window `4`, not only do the `1-4` and `1-2/4-2` splits move, but so does the opposite vertical split (`2-3/2-5`). Subtle but noticeable. Things still works great and this feature makes it easy to rectify things, since only one side of window `2` gets this bug you can easily resize things how they should be on the side that moved without you asking.

Does this help?

DanielAdolfsson:
Yes, it's a limitation with the implementation now. When you resize the window - in order to guarantee that it doesn't move around, it need to update the split ratios of a secondary parent. In certain situations that will make the other elements shrink/grow. The behavior is technically correct, but it might look weird. 

It's possible to make this update recursively and make stuff move around less, but I think this is good enough for a first version. But you're welcome to open a new issue to improve this. :)

Ashe:
Yes, agreed, this is more than I could ask for as a first pass, the main things are all there! I just wanted to raise it in case you weren't aware :) Thank you so much for your hard work

PR #2686, opened 2023-07-12
Fix new LAYER_TOP layer surface rendering above fullscreen windows

zakk4223:

Any newly created LAYER_TOP layer surface renders above fullscreen windows because the alpha for that depth is only set when the window initially goes fullscreen. This PR sets new TOP layer surfaces' alpha value to zero if there is a fullscreen window in the active workspace.

No?

It should be ready, although it feels a bit weird to have the layer map event go poking around in the workspace state? Maybe there's a better way to do this?


vaxerski (approved):
activeWorkspace is always valid so this is alr

PR #2694, opened 2023-07-13
Switch tabs

burningserenity:
Adds the ability to move windows within a group.
I frequently mess up the order I want when I add windows to groups; this lets me work with myself a little easier.
This is added as a new command, as I felt this was best since all of the group commands are separate from other movement commands.

I welcome comments. It should work, but there are probably things that can be cleaned up or improved.


burningserenity:
Indeed, it would have been better not to write my own shitty algorithm, like the way you did it. Thanks for adding this in, though.

burningserenity:
> This is a horrendous implementation.
> 
> Added in [9f7382b](https://github.com/hyprwm/Hyprland/commit/9f7382bca4ddc6617b0524645335723763c55302)

I respect your programming and design skills, Hyprland truly has potential. However, your soft skills, particularly the way you deliver criticism, could use some work.

vaxerski:
that's usually why programmers in corpos don't work directly with clients

burningserenity:
Yeah, community is hard. Goodbye.

PR #2705, opened 2023-07-13
update group decos

MightyPlaza:
updates group decos after moveGroupWindow


vaxerski (approved):
fair, worked for me, but it is a good idea

Issue #2684, opened 2023-07-11
movewindow and resizewindow do not bring floating windows on top.


Lassebq:
Neither movewindow nor resizewindow bring **floating** window on top when used.
Instead, the window stays under other floating windows.
You need to left click window once, and then perform action and the window will be on top.

## Expected behavior:
The window will become topmost in the floating window stack, and then it'll start moving/resizing

### hyprland.conf:
```bash
# Move/resize windows with mainMod + LMB/RMB and dragging
bindm = $mainMod, mouse:272, movewindow
bindm = $mainMod, mouse:273, resizewindow
```

PR #2706, opened 2023-07-13
move/resize window visual improvments

MightyPlaza:
Moves window to top when moving/resizing starts
Also moves focus to window when moving/resizing starts instead of when moving/resizing ends

didn't find any issues/reason to do it at the end instead of beggining
closes https://github.com/hyprwm/Hyprland/issues/2684

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -241,6 +241,9 @@ void IHyprLayout::onBeginDragWindow() {
     g_pHyprRenderer->damageWindow(DRAGGINGWINDOW);
 
     g_pKeybindManager->shadowKeybinds();
+
+    g_pCompositor->focusWindow(DRAGGINGWINDOW);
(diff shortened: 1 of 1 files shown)


src/layout/IHyprLayout.cpp
```diff
@@ -241,6 +241,9 @@ void IHyprLayout::onBeginDragWindow() {
     g_pHyprRenderer->damageWindow(DRAGGINGWINDOW);
 
     g_pKeybindManager->shadowKeybinds();
+
+    g_pCompositor->focusWindow(DRAGGINGWINDOW);
```
vaxerski, line 245: don't think this is necessary it's already focused

MightyPlaza: you can currently resize without focusing , the focus is only given in the end.
this way you can see which window you're resizing

vaxerski: w8 wat you can? okay nvm then

vaxerski (approved):
k

PR #2714, opened 2023-07-14
Make bind modifiers case-insensitive

Ordoviz:
Hyprland is very lenient when it comes to modifiers: `SUPERSHIFT`, `SUPER + SHIFT`, and `_xSUPERfoobarSHIFTx_` are all treated equally, but  `SUPER + shift` is treated as `SUPER`.

Thus, the config line

    bind = $mod + shift, T,       exec, notify-send Hello

unexpectedly binds notify-send to `$mod` not `$mod + SHIFT`. No warning is printed.

This PR makes modifiers case-insensitive. It has been tested and is ready to be merged.

Feel free to close this, if you don't like the change.


vaxerski (approved):
octopus

PR #2722, opened 2023-07-16
nix: use final.hyprland instead of prev.hyprland in waybar-hyprland

psentee:
Currently, waybar-hyprland package adds `prev.hyprland` to `$PATH`. This is nixpkgs' default hyprland, not hyprland injected by this overlay. I'd expect we want waybar-hyprland to depend on what's in `pkgs.hyprland` after this (and possibly user's other overlays) are applied – which is `final.hyprland`.

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -58,7 +58,7 @@ in {
         '';
         postFixup = ''
           wrapProgram $out/bin/waybar \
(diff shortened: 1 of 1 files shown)


fufexan (approved):
Sounds about right. I don't use overlays so it slipped my mind.

PR #2724, opened 2023-07-16
hyprctl empty dispatcher

MightyPlaza:
passes empty str as arg when in (var) doesn't have any empty spaces

should probably be added to the other hyprctl options, but only causes problems when argument can be none or not, which only happens in dispatchers afaik


vaxerski (approved):
fair

PR #2725, opened 2023-07-16
special workspace id fix

MightyPlaza:
sets highest as -99 instead of -100, so named special workspaces can't have id = -99


vaxerski (approved):
I do not want to know how you found this.

Issue #1802, opened 2023-03-16
Fullscreen window focusing is incomplete.


leg7:
When in fullscreen mode it's not possible to go backwards through the stack of windows, you can only go forwards.

`movefocus l` and `movefocus r` both go forwards in the window stack when they should go opposite directions.
Same thing for `movefocus u` and `movefocus d`.

`cyclenext prev` starts cycling in the opposite direction of `cyclenext` but after the 2nd call it will go back to the original window, so if we have windows 1-2-3 and we are on window 1 it will do this:

```
-> window 1
cyclenext prev -> window 3
cyclenext prev -> window 1 (should be 2)
cyclenext prev -> window 3 (should be 1)
cyclenext prev -> window 1 ...
```

vaxerski:
movefocus improved in 6fec5bfbeb80a5b2231af2ac5710a8b9c8bbc08a

cant seem to repro the cyclenext thing tho

romanstingler:
can you check if the issue is resolved for you

leg7:
still bugged altough I'm using the nixos stable version right now and cannot verify on latest commit

romanstingler:
@leg7 this should be resolved

PR #2728, opened 2023-07-17
fix: handle window change directions in fullscreen

Isti115:

It fixes the first part of #1802 by picking a window in the direction based on the parameter of the `movefocus` dispatcher:

> When in fullscreen mode it's not possible to go backwards through the stack of windows, you can only go forwards.
> 
> `movefocus l` and `movefocus r` both go forwards in the window stack when they should go opposite directions. Same thing for `movefocus u` and `movefocus d`.

But it runs into the same problem as mentioned in the second part of the issue by @leg7, which still also happens with `cyclenext prev`:

> `cyclenext prev` starts cycling in the opposite direction of `cyclenext` but after the 2nd call it will go back to the original window, so if we have windows 1-2-3 and we are on window 1 it will do this:
> 
> ```
> -> window 1
> cyclenext prev -> window 3
> cyclenext prev -> window 1 (should be 2)
> cyclenext prev -> window 3 (should be 1)
> cyclenext prev -> window 1 ...
> ```

(It just keeps switching between two windows. I guess that this could be due to the fact that the window focus history still keeps changing while going backwards, which makes `getPrevWindowOnWorkspace` misbehave.)

I might try figuring out the issue with `cyclenext prev` as well, so this could be considered a draft, but the first part is already useful, as you can at least go backwards one window if you accidentally skip over the one you're looking for while switching, and you don't need to flip through all of them again just to get back to the previous one.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1013,8 +1013,11 @@ void CKeybindManager::moveFocusTo(std::string args) {
         }
     };
 
-    const auto PWINDOWTOCHANGETO = PLASTWINDOW->m_bIsFullscreen ? g_pCompositor->getNextWindowOnWorkspace(PLASTWINDOW, arg == 'u' || arg == 't' || arg == 'r') :
-                                                                  g_pCompositor->getWindowInDirection(PLASTWINDOW, arg);
+    const auto PWINDOWTOCHANGETO = PLASTWINDOW->m_bIsFullscreen
+      ? (arg == 'd' || arg == 'b' || arg == 'r'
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
why change the directions? it was u t r for next, now u and t will be prev...?

vaxerski:
fair

vaxerski (approved):
g

leg7:
Going up the stack is broken on my end now just like the cyclenext bug that I described in #1802. This happens when going 'up' or 'left'. I'm not sure that this is the commit that broke it but since this is the only pull request that tagged my issue and I haven't updated in a long time.

PR #2731, opened 2023-07-17
Avoid connected monitor reusing unavailable ID

zjeffer:

@ti-mo noticed a bug in my previous PR: https://github.com/hyprwm/Hyprland/pull/2666#issuecomment-1637994116, where two monitors might get the same ID. This PR aims to fix this bug. 

The problem seems to happen when you have 2 monitors connected, and the monitor ID map looks something like this:

```jsonc
{
  DP-2 = 0 // connected
  DP-3 = 1 // connected
}
```
Disconnecting DP-2 and plugging it into DP-1 would give it ID=0 and update the map like this:
```jsonc
{
  DP-1 = 0 // connected
  DP-2 = 0 // disconnected
  DP-3 = 1 // connected
}
```
Then, disconnecting DP-3 and plugging it into DP-2 would still give it ID=0:
```jsonc
{
  DP-1 = 0 // connected
  DP-2 = 0 // connected
  DP-3 = 1 // disconnected
}
```

This PR fixes this, by checking whether no other currently connected monitor is using this ID. If it is, it will pick the next available ID. The last situtation would then look like this instead:
```jsonc
{
  DP-1 = 0 // connected
  DP-2 = 1 // connected
  DP-3 = 1 // disconnected
}
```

Maybe there's a better way to fix this, but I can't think of anything right now.

I couldn't reproduce the bug but I believe this will fix it. @ti-mo Can you verify this?

Needs to be tested just to be sure. I saw no problems with my own tests, but couldn't reproduce the issue before my changes.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1707,17 +1707,17 @@ void CCompositor::updateWindowAnimatedDecorationValues(CWindow* pWindow) {
 }
 
 int CCompositor::getNextAvailableMonitorID(std::string const& name) {
-    // reuse ID if it's already in the map
-    if (m_mMonitorIDMap.contains(name))
+    // reuse ID if it's already in the map, and the monitor with that ID is not being used by another monitor
+    if (m_mMonitorIDMap.contains(name) && !std::any_of(m_vRealMonitors.begin(), m_vRealMonitors.end(), [&](auto m) { return m->ID == m_mMonitorIDMap[name]; }))
         return m_mMonitorIDMap[name];
 
     // otherwise, find minimum available ID that is not in the map
-    std::unordered_set<int> usedIDs;
+    std::unordered_set<uint64_t> usedIDs;
     for (auto const& monitor : m_vRealMonitors) {
         usedIDs.insert(monitor->ID);
     }
 
-    int nextID = 0;
+    uint64_t nextID = 0;
     while (usedIDs.count(nextID) > 0) {
         nextID++;
     }
--- a/src/Compositor.hpp
+++ b/src/Compositor.hpp
@@ -102,7 +102,7 @@ class CCompositor {
     std::vector<CWindow*>                     m_vWindowsFadingOut;
     std::vector<SLayerSurface*>               m_vSurfacesFadingOut;
 
-    std::unordered_map<std::string, int64_t>  m_mMonitorIDMap;
+    std::unordered_map<std::string, uint64_t>  m_mMonitorIDMap;
 
     void                                      initServer();
     void                                      startCompositor();


src/Compositor.cpp
```diff
@@ -1707,17 +1707,18 @@ void CCompositor::updateWindowAnimatedDecorationValues(CWindow* pWindow) {
 }
 
 int CCompositor::getNextAvailableMonitorID(std::string const& name) {
-    // reuse ID if it's already in the map
-    if (m_mMonitorIDMap.contains(name))
+    // reuse ID if it's already in the map, and the monitor with that ID is not being used by another monitor
+    if (m_mMonitorIDMap.contains(name) &&
+        std::find_if(m_vRealMonitors.begin(), m_vRealMonitors.end(), [&](auto monitor) { return monitor->ID == m_mMonitorIDMap[name]; }) == m_vRealMonitors.end())
```
vaxerski, line 1712: you can use any_of from ranges

vaxerski:
I will wait for that guy to test it. Request my review when he approves the fix plox

ti-mo:
Thanks, this seems to work!

Another side effect that I've noticed is that the total desktop size keeps growing. After re-plugging both monitors, they are at coordinates `5120x0` and `7680x0` respectively:

```
λ  ~  hyprctl monitors
Monitor DP-1 (ID 0):
	2560x1440@165.080002 at 5120x0
	description: Dell Inc. DELL S2721DGF DZD6Q83 (DP-1)
	make: Dell Inc.
	model: DELL S2721DGF
	serial: DZD6Q83
	active workspace: 1 (1)
	special workspace: 0 ()
	reserved: 0 38 0 0
	scale: 1.00
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: 1

Monitor DP-2 (ID 1):
	2560x1440@165.080002 at 7680x0
	description: Dell Inc. DELL S2721DGF DKNCP83 (DP-2)
	make: Dell Inc.
	model: DELL S2721DGF
	serial: DKNCP83
	active workspace: 2 (2)
	special workspace: 0 ()
	reserved: 0 38 0 0
	scale: 1.00
	transform: 0
	focused: no
	dpmsStatus: 1
	vrr: 1
```

Not sure if this can be easily solved, though.

Also, I've noticed the monitor name is being used for looking up/assigning IDs. Shouldn't a combination of name and serial be used instead? Maybe this is already the case if the `name` parameter is the equivalent of the 'description' field in `hyprctl monitors`. Maybe you need identical monitors to reproduce this.

vaxerski:
> Another side effect that I've noticed is that the total desktop size keeps growing. After re-plugging both monitors, they are at coordinates 5120x0 and 7680x0 respectively:

thats a different bug with `auto` pos iirc

ti-mo:
Thanks!

PR #2740, opened 2023-07-18
movegroupwindow-improvement

MightyPlaza:
do not swap on group head and tail, instead move the window from one side to the other.
was kinda strange before so don't thing it needs a variable to turn on this behaviour

the other part of the diff is clang-format's doing

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -2086,14 +2086,9 @@ void CKeybindManager::moveGroupWindow(std::string args) {
     if (!g_pCompositor->m_pLastWindow || !g_pCompositor->m_pLastWindow->m_sGroupData.pNextWindow)
         return;
 
-    if (!BACK && g_pCompositor->m_pLastWindow->m_sGroupData.pNextWindow->m_sGroupData.head) {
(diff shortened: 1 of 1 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -2088,6 +2086,14 @@ void CKeybindManager::moveGroupWindow(std::string args) {
     if (!g_pCompositor->m_pLastWindow || !g_pCompositor->m_pLastWindow->m_sGroupData.pNextWindow)
         return;
 
-    g_pCompositor->m_pLastWindow->switchWithWindowInGroup(BACK ? g_pCompositor->m_pLastWindow->getGroupPrevious() : g_pCompositor->m_pLastWindow->m_sGroupData.pNextWindow);
+    if (!BACK && g_pCompositor->m_pLastWindow->m_sGroupData.pNextWindow->m_sGroupData.head) {
+        g_pCompositor->m_pLastWindow->m_sGroupData.pNextWindow->m_sGroupData.head = false;
```
vaxerski, line 2090: can't this be `std::swap`?

MightyPlaza: done

vaxerski (approved):
gucci

PR #2759, opened 2023-07-20
adds use_current_group_pos

MightyPlaza:
adds use_current_group_pos (bool) to decide whether to put new windows at current group position or at the tail.
this might need a better name

also simplifies the insertWindowToGroup and rewrites getGroupPrevious (the last condition seems to work but didn't test it a lot because there are no dispatchers for it)
also fixes locked groups in some cases

didn't find any bugs or broken stuff

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -664,7 +664,7 @@ void CWindow::setGroupCurrent(CWindow* pWindow) {
 }
 
 void CWindow::insertWindowToGroup(CWindow* pWindow) {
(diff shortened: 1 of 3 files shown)


vaxerski:
a wiki mr would be nice for the new opt

better name? `group_insert_after_current`?

MightyPlaza:
both done
https://github.com/hyprwm/hyprland-wiki/pull/270

vaxerski (approved):
soulja boy

PR #2776, opened 2023-07-22
xwayland: fix vertical pointer movement bug

### LINKED ISSUE #2376: Mouse not locking in Wine/Proton programs in fullscreen or windowed
It seems like I've found a major bug that could push away a lot of people playing games on Hyprland, It's got to do with mouse locking. Every time I run a game under Wine or Proton, (might not be exclusive to Wine or Proton) after tabbing out for a while and tab back in, the mouse locking is broken. The cursor seems to be still moving. I have tested this under Sway to see if the same behaviour could be observed there and after my testing, I can say that this is exclusively Hyprland behaviour. This doesn't look like It's a Proton or Wine bug, I've tested a lot of games extensively under Sway and Hyprland and It looks like Hyprland has some problems to iron out 

Test games were:
Deep Rock Galactic (Proton 7.0-6)
Overwatch 2 (Latest Wine-GE as of 27-05-2023) (DD//MM//YEAR)

Steps to Reproduce:
>Open a game run with Proton or Wine
>Switch to another workspace for a few minutes, maybe 10 if you want to be sure
>Switch back to the game and have your mouse lock ruined

Expected Behaviour
>For the mouse not to lose it's mouse locking properties after 

Noted Outcome:
>Mouse locking breaking after switching to a different workspace

Possible related bugs on the issue tracker:
https://github.com/hyprwm/Hyprland/issues/1732 (This issue is not the same as my mouse doesn't escape into my other monitor all the time but rather it looks like I am locked between certain regions while in-game which is caused by the mouse getting stuck on the edges of the screen.

It is also worth noting that a similar thing happens on Waydroid while trying to run Roblox on it, though it might be a function on Waydroid not working, since there are issues about that open the Waydroid's Github page too. (Can't test because this stopped working for me, will update if I can test)

rkanati:
This should _partially_ fix #2376, preventing the strange vertical-axis twitching behaviour seen in some XWayland applications and games.

As far as I can tell, #2376 is likely describing more than one issue. Perhaps it should be split up?

This is a single-line commit fixing an obvious inconsistency between X- and Y-axes in a relevant part of the code, so I don't think there's much to say.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1175,7 +1175,7 @@ void CInputManager::constrainMouse(SMouse* pMouse, wlr_pointer_constraint_v1* co
                 if (PWINDOW) {
                     if (PWINDOW->m_bIsX11) {
                         wlr_cursor_warp(g_pCompositor->m_sWLRCursor, nullptr, constraint->current.cursor_hint.x + PWINDOW->m_uSurface.xwayland->x,
-                                        PWINDOW->m_uSurface.xwayland->y + PWINDOW->m_vRealPosition.vec().y);
+                                        constraint->current.cursor_hint.y + PWINDOW->m_uSurface.xwayland->y);
 
                         wlr_seat_pointer_warp(constraint->seat, constraint->current.cursor_hint.x, constraint->current.cursor_hint.y);
                     } else {


vaxerski (approved):
OwO I made a wittwe fucky-wucky! Thankies fow spotting wis <3

PR #2781, opened 2023-07-23
Border fixes

MightyPlaza:
fixes some border issues (most of the time)

wrong spacing
20230723_02h56m32s_grim

damaging issues
20230723_02h57m48s_grim

noborder not removing the border spacing
20230723_03h00m51s_grim


vaxerski (approved):
all good, we should make a func just like `rounding()` probably in the long run.

PR #2785, opened 2023-07-23
implements nomaximizerequest

MightyPlaza:
https://github.com/hyprwm/hyprland-wiki/pull/275

also why does clang-format keep putting the bracket in the same line? (in Window.hpp changed from https://github.com/hyprwm/Hyprland/commit/18f9fb5e0fdb3fdc1aebe61b3c589bc23aed1e9f)

### DIFF
--- a/src/Window.hpp
+++ b/src/Window.hpp
@@ -10,8 +10,7 @@
 #include "helpers/Vector2D.hpp"
 #include "helpers/WLSurface.hpp"
 
-enum eIdleInhibitMode
-{
+enum eIdleInhibitMode {
     IDLEINHIBIT_NONE = 0,
     IDLEINHIBIT_ALWAYS,
     IDLEINHIBIT_FULLSCREEN,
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
nofullscreenrequest should also block maximize (see events/windows.cpp@1128)

So I think we should just check for that in mapWindow too and ditch a new rule for consistency.

MightyPlaza:
that's my fault, was using both rules at the same time for testing, will fix soon.
the idea is to separate maximize and fullscreen since a lot of programs are misusing the maximize request

MightyPlaza:
Now actually works as intended

vaxerski (approved):
fair

vaxerski:
> also why does clang-format keep putting the bracket in the same line? (in Window.hpp changed from https://github.com/hyprwm/Hyprland/commit/18f9fb5e0fdb3fdc1aebe61b3c589bc23aed1e9f)

my desktop's clang-format is fucked and it insists on ignoring the same line opening brace for enums rule

Issue #2576, opened 2023-06-23
Show groupbar when no_gaps_when_only is true


RedShiningFish:
When `no_gaps_when_only` is set to true and only a single window is on a workspace, the groupbar is also hidden. However, this makes it visually impossible to tell if a window is a group. The groupbar should be shown regardless of `no_gaps_when_only`.

MightyPlaza:
Mostly done, looks kinda strange.
not sure what spacing to put on top, maybe the normal out space, maybe the inner space, maybe a new var
![groupbar_noborder](https://github.com/hyprwm/Hyprland/assets/123664421/2df05325-6250-429a-a689-aeb90d913b4c)

MightyPlaza:
this has been fixed already @vaxerski

Issue #2286, opened 2023-05-11
indicate grouped windows with tabs when no_gaps_when_only is on


hazycora:
Window groups in hyprland have tabs, but when `no_gaps_when_only` is enabled they're turned off along with regular window borders, so there is no indication a window is a group when this setting is on. I think the tab lines at the top of window groups should remain visible with this setting on

wilecoyote2015:
This is also related to #520

hazycora:
I'm not sure it is? While it was mentioned in the comments of that issue, I think it's a separate feature, that's specific to the text being included and I'm referring to just the border being hidden with or without text

MightyPlaza:
also implemented now @vaxerski

PR #2791, opened 2023-07-24
no_gaps_when_only-improvements

MightyPlaza:
adds no_border_when_only=2 to show the border (not the same as no_border_when_only, since that would take floating into account)
https://github.com/hyprwm/hyprland-wiki/pull/276

also adds the group bar in no_gaps_when_only (maybe needs a bit extra space on top, but idk)

Closes:
https://github.com/hyprwm/Hyprland/issues/2576
https://github.com/hyprwm/Hyprland/issues/2286

don't think introduces any bugs


vaxerski (approved):
k

PR #2795, opened 2023-07-24
removes group locked check

MightyPlaza:
ignores group locked state when using dispatchers , since that seems to be the intended behaviour.


MightyPlaza:
closing since dupe of above

PR #2796, opened 2023-07-24
Make moveintogroup locking check configurable

memchr:
A new `moveintogroup_lock_check` configuration variable has been added to enable/disable lock checking in moveintogroup dispatch. 

This variable is set to false by default because the original behaviour of moveintogroup is to ignore `m_sGroupData.locked` as the call to the dispatcher is considered explicit.


vaxerski (approved):
oke

Issue #2797, opened 2023-07-24
Recent change broke auto-login launching Hyprland


ferrreo:
### Hyprland Version

Hyprland, built from branch main at commit 6db3c4ef5ec21f415577a26b0cdf57c98a778eb5 dirty (layout: no_border_when_only-improvements (2791)). Tag: v0.27.2-45-g6db3c4ef  flags: (if any)

### Bug or Regression?

Bug

### Description

Since https://github.com/hyprwm/Hyprland/commit/9fc5f4c48b20875337b7a1eddd274f191edd504d using greetd autologin to launch Hyprland directly gives a [CRITICAL] m_sWLRBackend was NULL! error.

If I remove this commit it works fine. 
If I launch from TTY it works fine.
If I launch via manual login via greetd it works fine.

### How to reproduce

Install + enable greetd
Set /etc/greetd/config.toml to have initial_session as running Hyprland as your user
reboot
check logs after getting kicked straight to login screen

### Crash reports, logs, images, videos

Crash Log:

--------------------------------------------
   Hyprland Crash Report
--------------------------------------------
Vaxry is going to be upset.

Hyprland received signal 6 (Aborted)

Version: 2bbe3aa122f242f43ac31fb85a39da48db69ca79
Tag: v0.27.2-44-g2bbe3aa1

System info:
	System name: Linux
	Node name: freya
	Release: 6.4.5-1-cachyos
	Version: #1 SMP PREEMPT_DYNAMIC Mon, 24 Jul 2023 03:13:53 +0000

GPU:
	03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 31 [Radeon RX 7900 XT/7900 XTX] [1002:744c] (rev c8) (prog-if 00 [VGA controller])
1b:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Raphael [1002:164e] (rev c9) (prog-if 00 [VGA controller])

os-release:
	NAME="CachyOS Linux"
	PRETTY_NAME="CachyOS"
	ID=cachyos
	ID_LIKE=arch
	BUILD_ID=rolling
	ANSI_COLOR="38;2;23;147;209"
	HOME_URL="https://cachyos.org/"
	DOCUMENTATION_URL="https://wiki.cachyos.org/"
	SUPPORT_URL="https://forum.cachyos.org/"
	BUG_REPORT_URL="https://github.com/cachyos"
[...]

PR #2799, opened 2023-07-24
Fix for issue #2797

ferrreo:
Fixes issue https://github.com/hyprwm/Hyprland/issues/2797

There was some weird issue with setting the priority to realtime before calling g_pCompositor initserver that was making wlroots very unhappy resulting in a crash in some instances.

Also some params were being lost by not reading the current params then modifying the priority and instead creating new params with only a priority.


vaxerski (approved):
spongebob squarepants

Issue #2809, opened 2023-07-25
No session for cookie, weird polkit error since gainRealTime was added


ferrreo:
### Hyprland Version

Hyprland, built from branch main at commit a9b8e2159c2813a6951d0b0186fbb7f2f8554d3e dirty (renderer: reset scissor after renderTexturePrimitive). Tag: v0.27.2-54-ga9b8e215  flags: (if any)

### Bug or Regression?

Bug

### Description

Another weird issue related to https://github.com/hyprwm/Hyprland/commit/9fc5f4c48b20875337b7a1eddd274f191edd504d

Since that was added and even with the follow up fix polkit just doesn't work at all. Removing the code from that commit it works perfectly fine as normal.

I've tried various different ways of doing the realtime priority change in the code and just can't get it to work without just removing it.

Also testing with and without it I see no meaningful difference in behaviour when system is under full load. Cursor still lags when CPU is 100% maxed etc.

### How to reproduce

Have Hyprland auto-launch on boot
Have polkitd run via systemd
Have an exec-once to launch polkit-kde or whatever in hypr config
Watch polkit shit itself, running pkexec gives a GDBus.Error:org.freedesktop.PolicyKit1.Error.Failed: No session for cookie error

vaxerski:
@milyth as the author of the patch, I don't know what the issue is. Maybe we should make it configurable (opt-in)?

ferrreo:
I'll create a pr tomorrow putting it behind a config variable.

milyth:
Are you using either `ananicy-cpp` or `system76-scheduler`?

ferrreo:
ananicy-cpp yeah

ferrreo:
Submitted a PR to do this if you could check you agree it is the correct solution and that it works @milyth and that you'd be happy with that fix too @vaxerski

milyth:
> ananicy-cpp yeah, several distros come with it preconfigured out of the box. Same for system76-scheduler.
> 
> So config option or maybe check if the priority has been changed first and skip if it has?
I really don't like "a broken config", so checking it first will be better imo.

jbeich:
Can you reproduce on Sway? The affected Hyprland code is identical to Sway (except style).

jbeich:
> Looks like setting SCHED_RR is just not compatible with external things that try and alter the nice value, which does make sense but is quite annoying.

[...]

PR #2814, opened 2023-07-26
Fix for issue #2809

ferrreo:

It fixes https://github.com/hyprwm/Hyprland/issues/2809 and from what I can tell when the priority is the default of 0 (so not modified externally) it seems to work as intended still.

Worth someone else double checking it still works as intended when not running ananicy-cpp

Ready pending the double check

### DIFF
--- a/src/init/initHelpers.cpp
+++ b/src/init/initHelpers.cpp
@@ -1,20 +1,18 @@
 #include "initHelpers.hpp"
+#include <sys/resource.h>
 
 bool Init::isSudo() {
     return getuid() != geteuid() || !geteuid();
 }
 
 void Init::gainRealTime() {
-    const int                minPrio = sched_get_priority_min(SCHED_RR);
-    int old_policy;
-    struct sched_param param;
-
-    if (pthread_getschedparam(pthread_self(), &old_policy, &param)) {
-        Debug::log(WARN, "Failed to get old pthread scheduling priority");
+    // If something else is adjusting priority, don't interfere
+    if (getpriority(PRIO_PROCESS, 0) != 0) {
         return;
     }
 
-    param.sched_priority = minPrio;
+    const int                minPrio = sched_get_priority_min(SCHED_RR);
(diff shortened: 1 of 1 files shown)


milyth (approved):

vaxerski (approved):
lgtm, @jbeich for bsd

src/init/initHelpers.cpp
```diff
@@ -1,20 +1,18 @@
...
 
 void Init::gainRealTime() {
-    const int                minPrio = sched_get_priority_min(SCHED_RR);
-    int old_policy;
-    struct sched_param param;
-
-    if (pthread_getschedparam(pthread_self(), &old_policy, &param)) {
-        Debug::log(WARN, "Failed to get old pthread scheduling priority");
+    // If something else is adjusting priority, don't interfere
+    if (getpriority(PRIO_PROCESS, 0) != 0) {
```
jbeich, line 10: `getpriority` checks `nice` value which is only valid for `SCHED_OTHER` (default, time-sharing). If realtime priority is set externally `nice` value may remain unchanged (zero). For example, on FreeBSD this line detects `nice Hyprland` but **not** `rtprio 31 Hyprland` and `idprio 31 Hyprland`.

Maybe instead try comparing `old_policy` and `param.sched_priority` values against defaults.

jbeich (changes requested):
Builds fine on FreeBSD but doesn't look correct.

src/init/initHelpers.cpp
ferrreo: The issue this pr is fixing is a conflict between this code setting Hyprland to realtime and external programs that try to modify the nice value e.g ananicy-cpp. 

That combination causes polkit and some other things to stop working correctly. Hence why we are looking for a chance in nice value.

The same issue shouldn't happen with externally set realtime priorities as this will just overwrite those, so I'm not sure such a change is warranted or needed.

ferrreo:
As per the issue closing this as looks to be an issue with ananicy-cpp/linux kernel

PR #2819, opened 2023-07-26
Renamed deprecated `fonts.enableDefaultFonts` value in nix config

Okirshen:
Changed the deprecated `fonts.enableDefaultFonts` to `fonts.enableDefaultPackages` in nix/module.nix.
See https://github.com/NixOS/nixpkgs/commit/83793ca8980283a6f62c028ebed336b9d1bbf80f for more info.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -72,7 +72,7 @@ in {
       };
     };
 
-    fonts.enableDefaultFonts = mkDefault true;
+    fonts.enableDefaultPackages = mkDefault true;
     hardware.opengl.enable = mkDefault true;
 
     programs = {


NotAShelf (approved):

NotAShelf:
@fufexan

fufexan (approved):
Thanks

crazyminecuber:
This breaks my system build on NixOS 23.05 since the new option is only available in unstable. Will hyprland only support NixOS unstable now? Or can we figure out a better way to make it work on both the latest stable NixOS and on unstable? I could move to the upstreamed module instead of using the flake input module as I do now, but then I still could not build latest hyprland on stable NixOS :(

NotAShelf:
> This breaks my system build on NixOS 23.05 since the new option is only available in unstable. Will hyprland only support NixOS unstable now? Or can we figure out a better way to make it work on both the latest stable NixOS and on unstable? I could move to the upstreamed module instead of using the flake input module as I do now, but then I still could not build latest hyprland on stable NixOS :(

You should consider using the native module (the one in nixpkgs) if you are on nixos stable. Hyprland, as far as I'm aware, has always targeted nixos unstable.

crazyminecuber:
Looks like the wiki has been updated with a recommended way to run the development version of hyprland on NixOS stable. So I guess I should just follow that everything is good :smiley:

PR #2820, opened 2023-07-26
assets/meson.build: install new wallpapers

parona-source:
Previous changes added new wallpapers and rng in which is used. Meson got left out so those who use it might silently get either the normal og wallpaper or nothing. And the cause wouldnt be obvious.

-

Ready to merge unless the solution is overdone and just an array with the names plainly is preferred.


vaxerski (approved):
lgtm but @fufexan for meson

fufexan (approved):

Issue #2808, opened 2023-07-25
Segmentation fault when opening a single client


smb374:
### Hyprland Version

b08b72358ad549fd066e5be0fc3aa4c9df367607

### Bug or Regression?

Bug

### Description

After the wallpaper is set (either `hyprpaper` or `swaybg`), spawning a client will cause segmentation fault with no obvious error message in the crash report.

The problem also occurred on all v0.27.* versions, while everything works after I downgraded to v0.26.0.

### How to reproduce

1. Start hyprland, either from TTY or display manager.
2. Spawn any client (e.g. spawn a terminal from keybind)

### Crash reports, logs, images, videos

v0.27.2: [hyprlandCrashReport34879.txt](https://github.com/hyprwm/Hyprland/files/12161211/hyprlandCrashReport34879.txt)
v0.27.1: [hyprlandCrashReport35968.txt](https://github.com/hyprwm/Hyprland/files/12161215/hyprlandCrashReport35968.txt)
v0.27.0: [hyprlandCrashReport36507.txt](https://github.com/hyprwm/Hyprland/files/12161219/hyprlandCrashReport36507.txt)

vaxerski:
it's a regression then, do a git bisect please

smb374:
Took me some time doing git bisect

Result first bad commit: [01f85a09a981eb6c8784d459f729d418c0dd19e3](https://github.com/hyprwm/Hyprland/commit/01f85a09a981eb6c8784d459f729d418c0dd19e3)

Config used for testing: [hyprland.conf](https://github.com/hyprwm/Hyprland/files/12161810/hyprland.conf.txt) (It comes with a `.txt` extension as GitHub doesn't support `.conf` files.)

vaxerski:
can you build in debug and provide the stacktrace? this is weird, I can't repro and you seem to be the only one affected.

when it crashes, to obtain the stacktrace, on systemd, do `coredumpctl`, find the last occurrence of hyprland (press END to go to the end) note its PID and then `coredumpctl debug <pid>`. When asked about symbols, say `y` when asked about pagers, do `c`. When you get to `(gdb)` do `bt -full` and paste the output here

smb374:
Sorry for the late reply as it was near midnight in my region, here is the backtrace:
[gdb_hypr.log](https://github.com/hyprwm/Hyprland/files/12167744/gdb_hypr.log)

I've been using `pwndbg` so the output may be different.
Hyprland is built by `make debug`.

[...]

PR #2829, opened 2023-07-28
Add temporary fix for #2808 by adding nullptr check.

smb374:
It is a temporary fix for #2808, the fix makes Hyprland not to segfault when starting a client.

There might be potential cases like #2808 where `nullptr` check is not performed properly and can lead to someone's running instance break after an update under rare circumstances.

It just adds `nullptr` checks in a if statement, should be able to merge without problem.


vaxerski (commented):
why `!= nullptr`? redundant. `if (g_pXWaylandManager->m_sWLRXwayland)` means the same

smb374:
> why `!= nullptr`? redundant. `if (g_pXWaylandManager->m_sWLRXwayland)` means the same

I tend to explicitly write `nullptr` comparison explicitly for clarity, just a personal preference.

vaxerski (approved):
top g

vaxerski:
> I tend to explicitly write nullptr comparison explicitly for clarity, just a personal preference.

we dont

PR #2852, opened 2023-07-31
Add setfloating command

Kasper24:
This adds a new dispatcher - setfloating. The another dispatcher togglefloating like it's name suggests only toggles the floating state, but what if a user only wants to set it to a specific state? an example use case could that we want to float a window when dragging it, but it doesn't make much sense to unfloat when dragging if it's already floated.
Example bind:
```
  bind = $mainMod, mouse:272, setfloating, 1
```

It's my first time here, so I obviously don't have much experience with Hyprland so I'm hoping I did not fuck things up.

### DIFF
--- a/src/events/Layers.cpp
+++ b/src/events/Layers.cpp
@@ -148,6 +148,7 @@ void Events::listener_mapLayerSurface(void* owner, void* data) {
         (!g_pCompositor->m_sSeat.mouse || !g_pInputManager->isConstrained());
 
     if (GRABSFOCUS) {
+        g_pInputManager->releaseAllMouseButtons();
         g_pCompositor->focusSurface(layersurface->layerSurface->surface);
 
         const auto LOCAL =


src/managers/KeybindManager.cpp
```diff
@@ -709,6 +710,49 @@ void CKeybindManager::toggleActiveFloating(std::string args) {
     }
 }
 
+void CKeybindManager::setActiveFloating(std::string args) {
+    CWindow* PWINDOW = nullptr;
+
+    bool floating_state = args.substr(0, args.find_first_of(' ')) == "1" ? true : false;
```
vaxerski, line 716: snake is illegal

src/managers/KeybindManager.cpp
```diff
@@ -709,6 +710,49 @@ void CKeybindManager::toggleActiveFloating(std::string args) {
...
+    std::string window_str   = args.substr(args.find_first_of(' ') + 1);
+
+    if (window_str != "" && window_str != "active" && window_str.length() > 1) {
```
vaxerski, line 719: no {} around short ifs

src/managers/KeybindManager.cpp
```diff
@@ -709,6 +710,49 @@ void CKeybindManager::toggleActiveFloating(std::string args) {
...
+        PWINDOW = g_pCompositor->getWindowByRegex(window_str);
+    } else {
+        PWINDOW = g_pCompositor->m_pLastWindow;
+    }
+
+    if (!PWINDOW)
+        return;
+
+    // remove drag status
+    g_pInputManager->currentlyDraggedWindow = nullptr;
```
vaxerski, line 729: this is horrible, copy+pasting code is borderline illegal. Just execute togglefloating if requested != current

vaxerski (changes requested):

src/managers/KeybindManager.cpp
Kasper24: What about merging these 2 dispatchers into 1 called 'floating'? I just wasn't sure if you guys are ok with breaking people configs or not

vaxerski: "you guys" lol thanks for acknowledging my 19 personalities.

I don't think I want to break people's configs here, but we could make a new dispatcher simply called "floating" and deprecate the old one

Kasper24: Oh, wasn't aware you were the only one working on this. 
Should togglefloating internally make use of the new floating function then?

vaxerski: yes, if that's what we are doing, obviously. De-dup code.

Kasper24: @vaxerski I've made the changes. Hopefully it's ok.

vaxerski: I don't understand why your code is still so unnecessairly convoluted

```cpp
const bool FLOATINGMODE = arg == "1";
if (FLOATINGMODE == pWindow->m_bIsFloating)
   return;
togglefloating(pwindow);
```

???

Kasper24: You are more expert in the Hyprland code than me, but from what I can understand from the codebase, you need to check take into account if it's a window group or not. That's what the current code does.

first case:
```cpp
if (PWINDOW->m_sGroupData.pNextWindow && PWINDOW->m_sGroupData.pNextWindow != PWINDOW)
{
   const auto PCURRENT     = PWINDOW->getGroupCurrent();
   PCURRENT->m_bIsFloating = !PCURRENT->m_bIsFloating;
}
````

second case:
```cpp
else {
   PWINDOW->m_bIsFloating = !PWINDOW->m_bIsFloating;
}
```

 That's why I couldn't just do
```cpp
FLOATINGMODE == pWindow->m_bIsFloating
```
I would need to add the if else from above, which again dupes the code and makes it uglier. If we turn the floatingState arguement to an int rather than a string, the if  (floatingStateStr = x) and enum will become redundant. It's also not necessary now too.

You are also doing it the other way around. I made the toggleFloating use the new setFloating method, you are making setFloating use toggleFloating. 

 In your example it also doesn't take care of the 'toggle' string option for setfloating.

Kasper24: Why was one checking the group floating state and the other was checking the window floating state before if you say it's unnecessary? That is what was confusing me lol

vaxerski: I dont understand you

Kasper24:
```cpp
    if (PWINDOW->m_sGroupData.pNextWindow && PWINDOW->m_sGroupData.pNextWindow != PWINDOW) {

        const auto PCURRENT     = PWINDOW->getGroupCurrent();
        PCURRENT->m_bIsFloating = !PCURRENT->m_bIsFloating;
        g_pLayoutManager->getCurrentLayout()->changeWindowFloatingMode(PCURRENT);

        CWindow* curr = PCURRENT->m_sGroupData.pNextWindow;
        while (curr != PCURRENT) {
            curr->m_bIsFloating = PCURRENT->m_bIsFloating;
            curr->updateDynamicRules();
            curr->updateSpecialRenderData();
            curr = curr->m_sGroupData.pNextWindow;
        }
    } else {
        PWINDOW->m_bIsFloating = !PWINDOW->m_bIsFloating;

        PWINDOW->updateDynamicRules();

        g_pLayoutManager->getCurrentLayout()->changeWindowFloatingMode(PWINDOW);
    }
```

Why was the check necessary in the old toggleActiveFloating, couldn't you have done the below in both cases like you are suggesting to do now? 
```cpp
PWINDOW->m_bIsFloating = !PWINDOW->m_bIsFloating;
```

vaxerski: thats not my point, my point is getting rid of the enum

Kasper24: This is what your code snippet suggested.

vaxerski: well the difference is just that you need to flip the floating state for all group members, thats it

PR #2871, opened 2023-08-01
Turinized Hyprland.

Alenygam:
Sei stato [Torinato](https://www.youtube.com/watch?v=gm3pDBTsO-Y).

(Skill Issue)

image


vaxerski:
this is trolling and spam, I'd suggest against it if you don't want to be banned from here.

PR #2874, opened 2023-08-02
Nix: fix nvidia patch for dual gpu system

MrDuartePT:
- Update ```wlroots-nvidia.patch```  patch in order fix the random flickering on the external output connected to Dgpu
- Updated patch to maintain distro packages downstream.
- Also update the wlroots.nix since  ```wlroots-nvidia.patch``` now patch renderer.c

This PR was created since last change to wlroots include with Hyprland 0.27.2 broke the patch use by gentoo.
Bug report here: https://bugs.gentoo.org/910576

Ready to merge

Patch tested on my Lenovo Legion 5, both in Hybrid and Discrete mode and the random flickering is not present anymore.
Also ran the workflow and can confirm the Hyprland still build fine on Nix.

image

```wlroots-nvidia.patch``` logs:
image

### DIFF
--- a/nix/patches/wlroots-nvidia.patch
+++ b/nix/patches/wlroots-nvidia.patch
@@ -1,3 +1,16 @@
+diff --git a/render/gles2/renderer.c b/render/gles2/renderer.c
+index 9fe934f7..9662d4ee 100644
+--- a/render/gles2/renderer.c
++++ b/render/gles2/renderer.c
+@@ -176,7 +176,7 @@ static bool gles2_bind_buffer(struct wlr_renderer *wlr_renderer,
+ 		assert(wlr_egl_is_current(renderer->egl));
+ 
+ 		push_gles2_debug(renderer);
+-		glFlush();
++		glFinish();
+ 		glBindFramebuffer(GL_FRAMEBUFFER, 0);
+ 		pop_gles2_debug(renderer);
+ 
 diff --git a/types/output/render.c b/types/output/render.c
 index 2e38919a..97f78608 100644
 --- a/types/output/render.c
--- a/nix/wlroots.nix
+++ b/nix/wlroots.nix
@@ -46,15 +46,6 @@ assert (lib.assertMsg (hidpiXWayland -> enableXWayland) ''
         ++ (lib.optionals nvidiaPatches [
           ./patches/wlroots-nvidia.patch
         ]);
-      postPatch =
-        (old.postPatch or "")
-        + (
-          if nvidiaPatches
-          then ''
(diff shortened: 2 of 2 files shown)


MrDuartePT:
I also remove lines from wlroots.nix file. If you think is not right i revert this change

MrDuartePT:
> The Gentoo ebuild isn't unique in using the patch as is, so there would benefit in having the nvidia patches in one place and kept up to date with the bundled wlroots whenever its updated.
> 
> https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hyprland-nvidia-git
> 
> The modification of the patch wouldn't change the the current result as this change was applied on Nix already. https://github.com/hyprwm/Hyprland/blob/main/nix/wlroots.nix#L54

I think that part need to be remove (i do it in my commit) because the nvidia patch it added before the modification in nix but let see.

jchhuang-tech:
> The Gentoo ebuild isn't unique in using the patch as is, so there would benefit in having the nvidia patches in one place and kept up to date with the bundled wlroots whenever its updated.
> 
> https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hyprland-nvidia-git (hyprland-nvidia and wlroots-nvidia in the aur use this patch but they carry it separately)
> 
> The modification of the patch wouldn't change the the current result as this change was applied on Nix already. https://github.com/hyprwm/Hyprland/blob/main/nix/wlroots.nix#L54

currently packages like `hyprland-nvidia-git` only apply `wlroots-nvidia.patch` without applying/knowing about the patch in `wlroots.nix`, leaving the result half patched and causing issues. both should be kept in one place

MrDuartePT:
> > The Gentoo ebuild isn't unique in using the patch as is, so there would benefit in having the nvidia patches in one place and kept up to date with the bundled wlroots whenever its updated.
> > https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hyprland-nvidia-git (hyprland-nvidia and wlroots-nvidia in the aur use this patch but they carry it separately)
> > The modification of the patch wouldn't change the the current result as this change was applied on Nix already. https://github.com/hyprwm/Hyprland/blob/main/nix/wlroots.nix#L54
> 
> currently packages like `hyprland-nvidia-git` only apply `wlroots-nvidia.patch` without applying/knowing about the patch in `wlroots.nix`, leaving the result half patched and causing issues. both should be kept in one place

But you think i should create a separate nvidia patch for Gentoo and AUR packages or the change I done to the wlroots-nvidia.patch and wlroots.nix is enough.

fufexan:
This looks fine. I'll merge after CI passes.

fufexan (approved):

PR #2884, opened 2023-08-03
xwayland: drop wlr_xwayland_surface.events.set_pid

romanstingler:
The PID of an X11 window cannot change.
This is a remnant from the days when wlroots queried the PID with a
window property, instead of using XRes.
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4306


vaxerski:
although I am grateful for pointing it out, I'd rather not receive wlr version bump MRs.

Please keep that in mind for the future.

vaxerski (approved):
looks alright, but yeah, please refrain from wlr bump MRs in the future.

PR #2897, opened 2023-08-06
Dwindle: smart resizing fallback

MightyPlaza:
add smart resize fallback in case the resize direction is a monitor border 
not sure if a variable to toggle this behaviour is needed (so will ping @DanielAdolfsson)

Issues:
- cursor image is wrong (but that deserves it's own PR imo)


vaxerski (approved):
top g from me

vaxerski:
nothing from adolfsson so I will merge

Issue #2849, opened 2023-07-30
Jetbrains IDE [Goland] completely crashes/freezes upon attempt to `open` new project


rbroggi:
### Hyprland Version

v0.27.2

### Bug or Regression?

Bug

### Description

Hello,

First of all, thank you a lot for this nice product! I've been using it for a couple of days and thought that doing some bug-reports could be an initial way of contributing.

Here is my running version:

```shell
❯ hyprctl version
Hyprland, built from branch  at commit b08b72358ad549fd066e5be0fc3aa4c9df367607 dirty ().
Tag: v0.27.2
```

```
GoLand 2023.2
Build #GO-232.8660.185, built on July 26, 2023
Runtime version: 17.0.7+7-b1000.6 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Linux 6.4.6-arch1-1
GC: G1 Young Generation, G1 Old Generation
Memory: 3970M
Cores: 12
Registry:
    suggest.all.run.configurations.from.context=true
    ide.completion.variant.limit=500
    editor.distraction.free.mode=true
    ide.experimental.ui=true

Non-Bundled Plugins:
    org.toml.lang (232.8660.88)
    IdeaVIM (2.4.1)
    com.intellij.ml.llm (232.9353)

Current Desktop: Hyprland
```

When working on a project on goland, if I attempt to `open` a new project, the whole IDE crashes and no longer responds. I cannot even close the window using `killactive`.  My only option is to open a shell and issue a kill command on the goland process. 

I'm not sure if this issue is a hyprland specific or Goland but I would be glad to provide more troubleshoot-information if instructed how. 

### How to reproduce

1. open goland;
2. file -> Open... 

### Crash reports, logs, images, videos

I have inspected those files and I cannot spot crash reports or interesting things in the `/tmp` log.  Everything seems to be pretty business as usual, e.g.:

```
[LOG] Window 55ce998bf480 set title to New Issue · hyprwm/Hyprland - Brave
[LOG] Searching for matching rules for Brave-browser (title: New Issue · hyprwm/Hyprland - Brave)
[LOG] Window 55ce998bf480 set title to New Issue · hyprwm/Hyprland - Brave
[LOG] Keybind triggered, calling dispatcher (64, , 52)
[LOG] Set keyboard focus to surface 55ce99358e50, with window name: tmux
[LOG] Activate request for surface at 55ce99358e50
```

rbroggi:
[...]

Issue #2880, opened 2023-08-03
[force_zero_scaling] Context menus get weirdly scaled mouse position and damage until refocused


TheSunCat:
### Hyprland Version

Hyprland, built from branch main at commit 51cda87fe47efec25265b3b9f6bebf9a0882aa08 dirty (hyprctl: rerender on cursor_zoom_factor dynamic calls). Tag: v0.27.2-64-g51cda87f

### Bug or Regression?

Bug

### Description

Most context menus of XWayland apps will have wrong mouse position and damage until they get refocused (f.e by entering and exiting a submenu).

### How to reproduce

- Open XWayland app (such as Android Studio)
- Open any context menu (such as top menu in Android Studio)
- Move the mouse down
- Observe how, the further down it gets, the more wrong the position is, and the worse the damage tracking is

### Crash reports, logs, images, videos

https://github.com/hyprwm/Hyprland/assets/44881120/3544cd14-8d9d-405a-85f3-09df817e25e3

wowblvck:
I have a similar problem in WebStorm.

https://github.com/hyprwm/Hyprland/assets/73607681/69714da1-7c9a-4803-b31c-7790d4788cae

TheSunCat:
This might be an issue specific to Java Swing applications, since both WebStorm and Android Studio use it. I remember encountering this in other software, but it might have been Swing as well.

TheSunCat:
This appears fixed since the new submenu rewrite! I cannot repro in Android Studio or JetBrains Rider.

PR #2906, opened 2023-08-07
Fix Jetbrains IDE dialogs

PatrickTaibel:
Fixes the following Jetbrains IDE dialog issues:
- Wrong position / size of dialogs due to not correctly scaled `WM_SIZE_HINTS` with hidpi patch
- Floating detection ignores that size hints can be unset which is not the same as both min and max sizes being equal
- "Find in Files" dialog jumping around weirdly due to being forced to position 0/0 initially sometimes

- This is the first time I'm touching a window manager and looking into X11
- I only use the Jetbrains IDEs via XWayland so I can't assure that I didn't break any other X11 apps functionality
- I'm not too sure about using `goalv` in `CHyprXWaylandManager::setWindowSize`  for the position. Without it the "Find in Files" dialog gets forced to position 0/0 sometimes which it really dislikes (and also does not make sense). As far as I can see most of the calls also use the goal vector for the size, so I guess that should be fine.

Ready for merging

### DIFF
--- a/nix/patches/wlroots-hidpi.patch
+++ b/nix/patches/wlroots-hidpi.patch
@@ -19,7 +19,7 @@ index 3d540522..1c5a2e37 100644
  	xcb_atom_t atoms[ATOM_LAST];
  	xcb_connection_t *xcb_conn;
 diff --git a/xwayland/xwm.c b/xwayland/xwm.c
-index 5f857f24..21584ebd 100644
+index 0c30d8f6..3d49695e 100644
 --- a/xwayland/xwm.c
 +++ b/xwayland/xwm.c
 @@ -19,6 +19,14 @@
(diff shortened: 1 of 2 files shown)


vaxerski:
the wlroots hidpi patches are being staged for removal, so I am against any changes to them. If the other changes help in cases without them (either unscaled or xwayland:force_zero_scaling) then I am okay with the mr as long as the patch changes are removed

PatrickTaibel:
Ah, I missed the new `xwayland:force_zero_scaling` option.
I've now tried it out. Problem is the issue with the dialogs is still there / made worse. First, I hoped this was again due to the size hints being scaled wrong in the geometry fetch. But, as far as I can see, the `void IHyprLayout::onWindowCreatedFloating(CWindow* pWindow) {` mixes scaled and unscaled properties (e.g. `desiredGeometry` being from XWayland scale and `PMONITOR->vecPosition` from wlroots). So that's not so easy to fix without a lot of conditions.
Wouldn't it be better to have some central place / functions for this instead of handling this on each occurence of a probably not correctly scaled XWayland property all over the code base?

vaxerski:
it probably would be better.

PatrickTaibel:
Fixed quite a few issues with the `xwayland:force_zero_scaling` option. Looks quite promising. I need to clean it up and then I will update this PR. This should fix most of the open issues with the new option as far as I can tell.

BrannonKing:
Does this fix https://github.com/hyprwm/Hyprland/issues/2849 ?

PatrickTaibel:
@BrannonKing Yes, it shoud fix this and basically all dialog size/position issues that happen in the Jetbrains IDEs with active scaling and multiple monitors / offset values. 
The current status of the PR patches the wlroots patch for this and has a few minor adjustments on top. This will never be merged, as the wlroots patch has been phased out. 
I started working on getting the `xwayland:force_zero_scaling` option working which is in https://github.com/PatrickTaibel/Hyprland/tree/fix_force_zero_scaling. That works too (but will break if you have different scale values set and is code wise pretty hacky). Problem here is that it won't be merged liked this either as @vaxerski changed quite a bit there the last two days.

wowblvck:
Does this fix the problem #2880 ?

h0tc0d3:
I got wlroots patches from [wlroots-hidpi](https://aur.archlinux.org/packages/wlroots-hidpi-xprop-git) because the nix patch didn't work https://github.com/hyprwm/Hyprland/issues/3055

I also applied the patch from this PR, it didn't apply completely, I looked at the code, edited it, and I got this patch.

Wayland apps run in the center and have the right size. Tested on: wine battle.net client, jetbrains ide, kicad. Dialogs and floating windows display correctly in the center.

```
diff --git a/src/managers/XWaylandManager.cpp b/src/managers/XWaylandManager.cpp
index 76f979da4b43..c1ce2287a955 100644
--- a/src/managers/XWaylandManager.cpp
+++ b/src/managers/XWaylandManager.cpp
@@ -151,7 +151,7 @@ void CHyprXWaylandManager::setWindowSize(CWindow* pWindow, Vector2D size, bool f
 
     // calculate pos
     // TODO: this should be decoupled from setWindowSize IMO
-    Vector2D windowPos = pWindow->m_vRealPosition.vec();
+    Vector2D windowPos = pWindow->m_vRealPosition.goalv();
 
     if (pWindow->m_bIsX11) {
         windowPos = windowPos - PMONITOR->vecPosition; // normalize to monitor
@@ -235,7 +235,9 @@ bool CHyprXWaylandManager::shouldBeFloated(CWindow* pWindow) {
             return true; // override_redirect
 
         const auto SIZEHINTS = pWindow->m_uSurface.xwayland->size_hints;
-        if (SIZEHINTS && (pWindow->m_uSurface.xwayland->parent || ((SIZEHINTS->min_width == SIZEHINTS->max_width) && (SIZEHINTS->min_height == SIZEHINTS->max_height))))
+        if (SIZEHINTS &&
+            (pWindow->m_uSurface.xwayland->parent ||
+             (SIZEHINTS->min_width > 0 && SIZEHINTS->min_height > 0 && SIZEHINTS->min_width == SIZEHINTS->max_width && SIZEHINTS->min_height == SIZEHINTS->max_height)))
             return true;
     } else {
         const auto PSTATE = &pWindow->m_uSurface.xdg->toplevel->current;
```

PatrickTaibel:
Closing this as I won't continue here. There are bugs in the JBR (Jetbrains fork of the openJDK) which cause quite some issues (mostly with dialogs) opened on different screens.

Issue #2008, opened 2023-04-09
keepaspect windowrule


bvr-yr:
**precondition**

- mpv || any other aspect_ratio sensible window (i.e. obs preview)

**str**

- float this window
- resize with mouse

**exp**

- window (with mouse resize) is resizing keeping initial aspect_ratio 

**act**

- whatever black px background u wish

PR #2907, opened 2023-08-07
Keep aspect ratio

MightyPlaza:
Allows resizing of windows without changing their aspect ratio

Implements window rule, prop and resize mode argument to mouse bind `resizewindow`

didn't check max size constrain
shouldn't cause any issues

will make wiki PR soon

Closes issue: https://github.com/hyprwm/Hyprland/issues/2008

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -235,7 +235,7 @@ void IHyprLayout::onBeginDragWindow() {
         }
     }
 
-    if (g_pInputManager->dragMode != MBIND_RESIZE)
+    if (g_pInputManager->dragMode != MBIND_RESIZE && g_pInputManager->dragMode != MBIND_RESIZE_FORCE_RATIO && g_pInputManager->dragMode != MBIND_RESIZE_BLOCK_RATIO)
         g_pInputManager->setCursorImageUntilUnset("grab");
 
     g_pHyprRenderer->damageWindow(DRAGGINGWINDOW);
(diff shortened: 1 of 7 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -1941,7 +1941,8 @@ void CKeybindManager::pinActive(std::string args) {
 }
 
 void CKeybindManager::mouse(std::string args) {
-    const auto TRUEARG = args.substr(1);
+    const auto TRUEARG = args.substr(1, args.find(' ') - 1);
+    const auto EXTRARG = args.substr(args.find(' ') + 1);
```
vaxerski, line 1945: this might act wonky if we get a `std::string::npos`. Might be worth checking if there exists a space, or even better, using a CVarList

src/managers/KeybindManager.cpp
```diff
@@ -1966,8 +1967,16 @@ void CKeybindManager::mouse(std::string args) {
             g_pKeybindManager->m_bIsMouseBindActive = true;
 
             g_pInputManager->currentlyDraggedWindow = g_pCompositor->vectorToWindowIdeal(g_pInputManager->getMouseCoordsInternal());
-            g_pInputManager->dragMode               = MBIND_RESIZE;
 
+            if (isNumber(EXTRARG)) {
+                switch (stoi(EXTRARG)) {
```
vaxerski, line 1972: a) `std::stoi`
b) all `sto`s have to be wrapped in a try{} to prevent bad data crashes. I recommend doing this in the beginning and ignoring it if it's invalid.

src/layout/IHyprLayout.cpp
```diff
@@ -314,27 +314,58 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
...
-                newSize = (newSize - DELTA).clamp(Vector2D(20, 20), MAXSIZE);
-                newPos  = newPos - newSize + m_vBeginDragSizeXY;
+                newSize = newSize - DELTA;
+            } else if (m_eGrabbedCorner == CORNER_TOPRIGHT) {
+                newSize = newSize + Vector2D(DELTA.x, -DELTA.y);
+            } else if (m_eGrabbedCorner == CORNER_BOTTOMLEFT) {
+                newSize = newSize + Vector2D(-DELTA.x, DELTA.y);
+            }
+
+            if ((m_vBeginDragSizeXY.x >= 1 && m_vBeginDragSizeXY.y >= 1) &&
```
vaxerski, line 336: style on all of this: no braces around single-line ifs / elses

MightyPlaza:
all done
should I also add a bind to the default config?

vaxerski (approved):
rollin' on the block

vaxerski:
> should I also add a bind to the default config?

na

MightyPlaza:
#2008 can be closed as well

bvr-yr:
clap clap! <3

PR #2913, opened 2023-08-08
Fix background not being cleared

outfoxxed:
Previously no image would be rendered if hyprland could not find one to use, leaving the last frame as the background. This generates a grey image if hl can't find a background (with the splash overlayed this time).

Additionally it adds `assets/` as an image search path for use in dev. (nix does not store assets in /usr paths which is why I noticed this at all)

no

ready to merge

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1723,19 +1723,15 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
         texName += "2K.png";
     }
 
-    // goto is banned? ok have a do while.
-    std::string texPath = "/usr/share/hyprland/" + texName;
-    do {
-        if (std::filesystem::exists(texPath))
-            break;
-        texPath = "/usr/local/share/hyprland/" + texName;
-        if (std::filesystem::exists(texPath))
-            break;
-        texPath = "assets/" + texName;
+    std::string texPath;
+    auto        paths = {"/usr/share/hyprland/", "/usr/local/share/hyprland/", "assets/"};
+
+    for (auto& path : paths) {
+        texPath = path + texName;
         if (std::filesystem::exists(texPath))
             break;
         texPath = "";
-    } while (false);
+    }
 
     PTEX->m_vSize = textureSize;
 
@@ -1771,7 +1767,7 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
 
     if (texPath.empty()) {
         cairo_set_source_rgba(CAIRO, 0.106, 0.106, 0.106, 1.0);
-        cairo_rectangle(CAIRO, 0, 0, box.width, box.height);
+        cairo_rectangle(CAIRO, 0, 0, textureSize.x, textureSize.y);
         cairo_fill(CAIRO);
     }
 
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1723,17 +1723,19 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
         texName += "2K.png";
     }
 
+    // goto is banned? ok have a do while.
     std::string texPath = "/usr/share/hyprland/" + texName;
-    if (std::filesystem::exists(texPath))
-        goto hastex;
-    texPath = "/usr/local/share/hyprland/" + texName;
-    if (std::filesystem::exists(texPath))
-        goto hastex;
-    texPath = "assets/" + texName;
-    if (std::filesystem::exists(texPath))
-        goto hastex;
-    texPath = "";
-hastex:
+    do {
+        if (std::filesystem::exists(texPath))
+            break;
+        texPath = "/usr/local/share/hyprland/" + texName;
+        if (std::filesystem::exists(texPath))
+            break;
+        texPath = "assets/" + texName;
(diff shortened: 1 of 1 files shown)


src/render/OpenGL.cpp
```diff
@@ -1686,28 +1686,32 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
...
-    if (!std::filesystem::exists(texPath)) {
-        // try local
-        texPath = texPath.substr(0, 5) + "local/" + texPath.substr(5);
-
-        if (!std::filesystem::exists(texPath))
-            return; // the texture will be empty, oh well. We'll clear with a solid color anyways.
-    }
+    std::string texPath = "/usr/share/hyprland/" + texName;
+    if (std::filesystem::exists(texPath))
+        goto hastex;
```
vaxerski, line 1706: goto is banned

src/render/OpenGL.cpp
```diff
@@ -1686,28 +1686,34 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
...
-    if (!std::filesystem::exists(texPath)) {
-        // try local
-        texPath = texPath.substr(0, 5) + "local/" + texPath.substr(5);
-
-        if (!std::filesystem::exists(texPath))
-            return; // the texture will be empty, oh well. We'll clear with a solid color anyways.
-    }
+    // goto is banned? ok have a do while.
+    std::string texPath = "/usr/share/hyprland/" + texName;
+    do {
```
vaxerski, line 1706: wtf?

if ()
 path =
else if ()
 path = 
 
 ??

outfoxxed: That won't actually work. Its has to be a bunch of nested ifs, a do while (false), or a goto.

vaxerski: oh really?

```cpp
auto findPath = [&] (std::string& path) -> void {
   if (std::filesystem::exists(path)
        return;
   path = "/usr/local/share/hyprland/" + texName;
   if (std::filesystem::exists(path))
       return;
   path = "assets/" + texName;
   if (std::filesystem::exists(path))
       return;
    path = "";
};

findPath(texPath);
```

outfoxxed: That also works I guess

outfoxxed: Fixed

outfoxxed:
Removed the do while and fixed the fallback cairo texture having the wrong size

src/render/OpenGL.cpp
```diff
@@ -1755,8 +1757,19 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
...
+    cairo_surface_t* CAIROSURFACE;
+    if (texPath.empty())
+        // if we couldn't find a texture, use a grey filler image instead.
+        CAIROSURFACE = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, textureSize.x, textureSize.y);
+    else
+        CAIROSURFACE = cairo_image_surface_create_from_png(texPath.c_str());
+    const auto CAIRO = cairo_create(CAIROSURFACE);
+
+    if (texPath.empty()) {
+        cairo_set_source_rgba(CAIRO, 0.106, 0.106, 0.106, 1.0);
```
vaxerski, line 1769: there is a setting for a default bg col. We should leave this empty anyways and not render it if textpath is missing to save on a draw call.

outfoxxed: The image is cached so the draw call is kind of a moot point. As for the bg color, have you considered making the logo images transparent so they can be overlayed? I just took the background color from the images.

vaxerski: rendering of the texture is another heavy drawcall especially on lower-end hardware. We also disable blending to save on some compute time.

Currently, last time I was profiling, rendering the background texture was one of the heaviest calls in the entire draw stack.

outfoxxed: That sounds somewhat concerning, but I think fixing that should be a seperate change based on occlusion of the background (e.g. hyprpaper), and not randomly disabling the background image if we can't find it, because that means the previous frame is never cleared.

vaxerski: if the image is not found we should release the tex and just check for that here:
https://github.com/hyprwm/Hyprland/blob/19c4855afce7e4dfa316cbd40f2c991fd3f87e0b/src/render/Renderer.cpp#L1021-L1024

outfoxxed: Unless that's new since this pr, it dosen't actually work. Also that bit looks like it wouldn't use the background image if the texture exists which seems wrong. You also lose the splash, but I see your point.

What do you think of always doing the bkg color clear and using transparent bkg images?
Also not sure why the bkg clear would be so expensive unless it triggers unconditionally every frame.

outfoxxed: IMO if you want to make not clearing the background an option, it should be a clearly defined configuration option and not removing an undocumented system file. In my opinion attempting to match the behavior with the image if its missing is the best solution.

outfoxxed: Oh sorry I was confusing it with the old behavior. I'll look at it again tomorrow. Using clear seems fine but imo the background color with an image should match for consistency.

vaxerski:
closing due to inactivity

PR #2929, opened 2023-08-10
Nix: corrections for overlays, overrideable systems

spikespaz:
Nearly identical to <https://github.com/hyprwm/xdg-desktop-portal-hyprland/pull/75> and <https://github.com/hyprwm/hyprland-protocols/pull/6>. Copied message and edited below.

1. The `nix fmt` command did not work for me so I have reverted it to use `legacyPackages` for simplicity. It works now.
   - Ran `nix fmt` on the whole flake, however the `overlays` in the `import nixpkgs` in `flake.nix` has some comments that get messed up when Alejandra has its way. I dislike Alejandra.
2. The overlay `hyprland-packages` used to include some dependencies (`udis86`) but not all of them. That was the job of the `default` overlay. I have changed it so that `hyprland-packages` will work correctly on its own, as now it includes all dependency overlays.
   - Now each package has its own overlay, while `default` combines the ones an end-user is expected to need.
   - ```nix
     $ nix eval 'path:.#overlays' --apply 'builtins.attrNames'
     [ "default" "hyprland-extras" "hyprland-packages" "udis86" "waybar-hyprland" "wlroots-hyprland" ]
     ```
3. Versions now use string interpolation rather than concatenation, because it looks nicer than when Alejandra tries to wrap concatenated strings.
   - ~~Perhaps we should adopt my (unpublished) changes from `Alexays/Waybar` and pull the first version fragment from `meson.build`.~~ The `props.json` file takes care of that. Maybe the other flakes could learn a thing or two?
4. Added another input `nix-systems` so that the `systems` is overrideable. I like this much better than using `flake-utils` or `flake-parts`.
   - Using `default-linux`, which is `[ "x86_64-linux" "aarch64-linux" ]`.
   - Updated usage when importing `nixpkgs` to correct for undocumented deprecation of `system`.
   - https://github.com/nix-systems/nix-systems
   - **Blocker:** This PR depends on <https://github.com/hyprwm/hyprland-protocols/pull/6> and for now you will see an error when evaluating flake outputs.
5. The attributes of the `packages` output are inherited from internal `pkgsFor`, which contains all of the overlays pre-applied.
   - ```nix
     $ nix eval 'path:.#packages.x86_64-linux' --apply 'builtins.attrNames'
     [ "default" "hyprland" "hyprland-debug" "hyprland-hidpi" "hyprland-no-hidpi" "hyprland-nvidia" "hyprland-protocols" "hyprland-share-picker" "hyprland-unwrapped" "udis86" "waybar-hyprland" "wlroots-hyprland" "xdg-desktop-portal-hyprland" ]
     ```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -44,6 +44,21 @@
         "xdph": "xdph"
       }
     },
+    "systems": {
+      "locked": {
+        "lastModified": 1689347949,
+        "narHash": "sha256-12tWmuL2zgBgZkdoB6qXZsgJEH9LR3oUgpaQq2RbI80=",
+        "owner": "nix-systems",
+        "repo": "default-linux",
+        "rev": "31732fcf5e8fea42e59c2488ad31a0e651500f68",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default-linux",
+        "type": "github"
+      }
+    },
     "wlroots": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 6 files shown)


spikespaz:
@fufexan

fufexan (approved):

PR #2931, opened 2023-08-10
patches group toplevel-activated

MightyPlaza:
fixes hidden group windows always having toplevel activated status

not ideal, but doing it the right way seems to require messing with other functions, which will likely break stuff

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -929,9 +929,8 @@ void CCompositor::focusWindow(CWindow* pWindow, wlr_surface* pSurface) {
         auto curr = pWindow;
         do {
             curr = curr->m_sGroupData.pNextWindow;
-            if (curr->m_phForeignToplevel) {
+            if (curr->m_phForeignToplevel)
                 wlr_foreign_toplevel_handle_v1_set_activated(curr->m_phForeignToplevel, false);
-            }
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
cant you use the deactivate all func in ccompositor iirc?

MightyPlaza:
did only find done for workspaces, and don't think that's it
if there is one that does that for all the windows, it's even worse than doing just for the group

vaxerski:
fair point

src/Compositor.cpp
```diff
@@ -924,6 +924,17 @@ void CCompositor::focusWindow(CWindow* pWindow, wlr_surface* pSurface) {
 
     g_pLayoutManager->getCurrentLayout()->onWindowFocusChange(pWindow);
 
+    // TODO: implement this better
+    if (!PLASTWINDOW && pWindow->m_sGroupData.pNextWindow && pWindow->m_sGroupData.pNextWindow != pWindow) {
+        auto curr = pWindow;
+        do {
+            curr = curr->m_sGroupData.pNextWindow;
+            if (curr->m_phForeignToplevel) {
```
vaxerski, line 932: style: no braces around short ifs

MightyPlaza: done

vaxerski (approved):
top g gangsta full

PR #2951, opened 2023-08-12
CI: add reusable actions

fufexan:

Allows reusing actions across the org, meaning less copied code and less chances of errors slipping in when a change is made.
- Nix: define strategy, which builds in parallel, faster than doing `nix flake check` (which does not guarantee the builds are run as the dependency graph dictates.)

See my comment below.

I also want to create a conditional job:
- when wlroots gets updated, don't run Nix builds
- when there's no wlroots update, run the builds

The above conditional was replaced by simply running the wlroots update first, and re-cloning HEAD from the branch on the subsequent package build job.


memchr:
FYI, there is a tool called `act` that lets you run and debug github actions locally, it's in nixpkgs I think.

fufexan:
Damn, didn't know that. Thanks. However I think I've mostly finished this PR.

fufexan:
Ok, so if there are both a `pull_request` and a `push` trigger, the wlroots update action will fail for one of them. Doesn't affect builds at all though. This should happen very rarely, only if vax decides to bump wlroots inside a PR. I think it's good enough for now.

Alternatively I could make it so that `pull_request` is never run if the PR ref is from this repo, but it's not worth the complication.

memchr:
Maybe add a condition to skip this job in PR?
```
wlroots / wlroots
Secret PAT is required, but not provided while calling.
```

fufexan:
Oops. Okay, I'll fix it.

memchr:
```yaml
if: ${{ !startsWith(github.ref, "refs/pull/") && !endsWith(github.ref, "/merge") }}
```

fufexan:
I came up with something like this
```yml
name: Nix

on: [push, pull_request, workflow_dispatch]

jobs:
  wlroots:
    if: github.event_name == 'push'
    uses: ./.github/workflows/nix-update-wlroots.yml
    secrets: inherit

  build:
    if: always() && !cancelled() && !contains(needs.*.result, 'failure')
    needs: wlroots
    uses: ./.github/workflows/nix-build.yml
    secrets: inherit
```
since we still want the packages to be built even if wlroots updating was skipped.

Issue #2953, opened 2023-08-12
No shadows when only


YingchangLiu:
### Description

There are two gaps between the bar and my left screen edge and the  window behind it. 

When the bar is not fully aligned with the window, the window shadow near the bar is visible, but not in the other three edges. It looks strange. 

Request a setting option ``` no_shadows_when_only ``` or adding it into ```no_gaps_when_only``` to eliminate shadows when there is only one window.

MightyPlaza:
I think I enabled this because otherwise group bar would not be visible which, in my opinion, is even worse than only having shadows at the top.

I'll see if I can disable it again without totally disabling decorations

PR #2956, opened 2023-08-12
remove shadow no_gaps_when_only

MightyPlaza:
re-disables shadows when only and using no_gaps_when_only (https://github.com/hyprwm/Hyprland/issues/2953)

as a side effect adds shadow as a workspace rule


vaxerski (approved):
alr

MightyPlaza:
forgot to update master as well

PR #2961, opened 2023-08-13
master: fix `always_center_master`

### LINKED ISSUE #2108: Master layout: `always_center_master` no longer works
Commits between [4bc3f9ad and b1580351](https://github.com/hyprwm/hyprland/compare/4bc3f9ad...a4330fe3) broke something so `always_center_master` don't work anymore, in master layout.

kbtz:
Fixes #2108

It is mostly toughening up the conditionals to handle the various combinations of orientations and window count, along with some refactoring for better consistency/simplicity.

Unfortunelly the diff went nuts due to the reordering of some code branches, so I advise to review it by reading the `calculateWorkspace` function which is the only thing I've touched.

Tested with combinations of `allow_small_split` and `always_center_master` on different orientations and it seems to be ok. Also it is worth noting that with `always_center_master` turned off, orientation=center behaves as orientation=left until it gets at least two slave windows (this behavior made sense to me so I kept it).

~~Later I'd like to try to allow resizing the master area even when there is no slave windows if that's ok..~~ Done

Thanks!

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -605,19 +605,20 @@ void CHyprMasterLayout::resizeActiveWindow(const Vector2D& pixResize, eRectCorne
         return;
     }
 
-    // get monitor
-    const auto PMONITOR = g_pCompositor->getMonitorFromID(PWINDOW->m_iMonitorID);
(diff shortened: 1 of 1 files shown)


vaxerski:
@Amatrelan for the original issue

vaxerski (approved):
let it be

Issue #2957, opened 2023-08-13
window wrong size when opening second window on master layout


uoou:
### Hyprland Version

Hyprland, built from branch main at commit 739598717bbff79aad8599dcd0ca7a423682482a dirty (remove shadow no_gaps_when_only (2956)). Tag: v0.28.0-36-g73959871

### Bug or Regression?

Bug

### Description

This only happens on master layout and when `no_gaps_when_only = 1` and gaps are present.

When an existing window is open, opening a second window causes the first window to be incorrectly sized. The original window seems to have the border width added to its dimensions.

This happens with most applications I've tested with, for example qutebrowser, chromium, firefox and foot all exhibit this. It does not happen if the second window is kitty (regardless of what the other window is).

The problem is corrected if I switch workspaces, open a third window or, as I just discovered, take a screenshot with grim.

I'm attaching a screenshot where I've set gaps and borders big to make the problem more obvious.

I hope this makes sense, it's a little tricky to explain.

I'll gladly provide more info if required, there's nothing exotic in my configs, and I'm using hyprland-git from the AUR.

### How to reproduce

- use master layout
- set `no_gaps_when_only = 1`
- set gaps and borders to something chunky to make the problem obvious
- open (for example) foot twice

### Crash reports, logs, images, videos

![2023-08-13_005216_904400879](https://github.com/hyprwm/Hyprland/assets/5243056/f21090c3-8853-43ff-925a-425858b1c42d)

MightyPlaza:
cannot repro with foot, are you by any chance setting a border workspace rule?

uoou:
Okay, thanks. I just tried with a default config with just the changes I mentioned and it doesn't happen. So it must be something else in my config. I had assumed not, since I'm not doing anything weird.

Thanks for your help, I'll find the culprit then close the issue.

uoou:
Right, it happens on this config for me.

The only things I've changed from default are:

- changed layout to master
- increased border size to 10 to make it more noticeable
- added `no_gaps_when_only = 1` inside master layout
- changed default terminal keybinding from kitty to foot

Can you confirm that it doesn't exhibit on this config for you?

[hyprland.conf.txt](https://github.com/hyprwm/Hyprland/files/12328598/hyprland.conf.txt)

MightyPlaza:
my bad was setting no_gaps in dwindle and master was using no_gaps = 2

MightyPlaza:
[...]

PR #2986, opened 2023-08-15
Border improvements

MightyPlaza:
fixes issues with window and workspace border related rules not working as intended
also simplifies getting window border size and setting the workspace rules

closes: https://github.com/hyprwm/Hyprland/issues/2957
shouldn't case any issues

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -581,28 +581,27 @@ void CWindow::updateDynamicRules() {
 // it is assumed that the point is within the real window box (m_vRealPosition, m_vRealSize)
 // otherwise behaviour is undefined
 bool CWindow::isInCurvedCorner(double x, double y) {
-    static auto* const ROUNDING = &g_pConfigManager->getConfigValuePtr("decoration:rounding")->intValue;
-
-    if (getRealBorderSize() >= *ROUNDING || *ROUNDING == 0)
+    const int ROUNDING = rounding();
+    if (getRealBorderSize() >= ROUNDING)
         return false;
 
     // (x0, y0), (x0, y1), ... are the center point of rounding at each corner
-    double x0 = m_vRealPosition.vec().x + *ROUNDING;
-    double y0 = m_vRealPosition.vec().y + *ROUNDING;
-    double x1 = m_vRealPosition.vec().x + m_vRealSize.vec().x - *ROUNDING;
-    double y1 = m_vRealPosition.vec().y + m_vRealSize.vec().y - *ROUNDING;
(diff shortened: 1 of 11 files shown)


src/Window.cpp
```diff
@@ -589,10 +581,9 @@ void CWindow::updateDynamicRules() {
 // it is assumed that the point is within the real window box (m_vRealPosition, m_vRealSize)
 // otherwise behaviour is undefined
 bool CWindow::isInCurvedCorner(double x, double y) {
-    static auto* const ROUNDING   = &g_pConfigManager->getConfigValuePtr("decoration:rounding")->intValue;
-    static auto* const BORDERSIZE = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
+    static auto* const ROUNDING = &g_pConfigManager->getConfigValuePtr("decoration:rounding")->intValue;
 
-    if (BORDERSIZE >= ROUNDING || ROUNDING == 0)
+    if (getRealBorderSize() >= *ROUNDING || *ROUNDING == 0)
```
MightyPlaza, line 586: not sure if this correct, but doesn't make sense otherwise

vaxerski: we could switch this over to `CWindow::rounding()` while we're at it. But yeah, looks alright.

vaxerski (dismissed):
the specialrenderdata / additionalconfigdata situation is a mess, it needs to be cleaned up. Note to self...

MightyPlaza:
> the specialrenderdata / additionalconfigdata situation is a mess, it needs to be cleaned up. Note to self...

I though about making a windowProps object that could be set using hyprctl or window rules and workspace rules (maybe combine them into something new to reduce the number of workspace and window rules and the need to update each one every time we want to add something new).
each thing that could set the property would have a certain priority (hyprctl > window rules > workspace rules > layout requests > global config)
but that was more work, and would be storing a lot of redundant -1 most of the time (no idea if there is a better way to do this)

using rounding() now so should be ready for merge @vaxerski

vaxerski:
> I though about making a windowProps object that could be set using hyprctl or window rules and workspace rules (maybe combine them into something new to reduce the number of workspace and window rules and the need to update each one every time we want to add something new).
each thing that could set the property would have a certain priority (hyprctl > window rules > workspace rules > layout requests > global config)
but that was more work, and would be storing a lot of redundant -1 most of the time (no idea if there is a better way to do this)

I will probably look into it when I am back from holidays. There are a bunch of pain points I want to address, but it's way nicer to code on your desktop than on a small laptop in the middle of nowhere.

PR #2988, opened 2023-08-15
fix: Use monitor refreshrate tick for resizing and/or moving windows

DashieTM:
when using the following 2 settings: animate_mouse_windowdragging=false and animate_manual_resizes=true
The animations for resizing are currently not using monitor refreshrate ticks due to this being locked behind PANIMATEMOUSE, which is only active with animate_mouse_windowdragging.

This results in animations being "laggy" again. (at least for me)

This simply activates the feature for both, meaning both can benefit from the improved tick.

ready


vaxerski (approved):
my bad 🚶‍♂️

Issue #2634, opened 2023-07-03
Waybar remains hidden when only fullscreen window in workspace is moved with `movetoworkspacesilent`


gekoke:
Reproduction:
- have Waybar running
- have exactly one fullscreen (`fullscreen, 0`) window in workspace
- have no other windows in workspace
- ensure Waybar is hidden due to fullscreen
- move said fullscreen window to another workspace using `movetoworkspacesilent`

Expected:
- Waybar is now unhidden, since no windows remain in workspace

Actual:
- Waybar remains hidden, even though no windows remain in workspace

gekoke:
Running a different Waybar config seems to resolve the issue - weird. Not sure if this is an issue with Hyprland.

gekoke:
The issue disappears when using `"mode": "dock"` in Waybar config - however, this is not a complete solution as the bar is visible through transparent windows while in fullscreen mode when that setting is enabled.

memchr:
Here is a possible explanation for the bug

## The `fullscreen` dispatcher

Here is the callstack when using the `fullscreen` dispatcher

- [ls->alpha](https://github.com/memchr/Hyprland/blob/99b8a768cfe8741a0685faad8792bbf923293805/src/Compositor.cpp#L2070)
- [updateFullscreenFadeOnWorkspace](https://github.com/memchr/Hyprland/blob/99b8a768cfe8741a0685faad8792bbf923293805/src/Compositor.cpp#L2104)
- [setWindowFullscreen](https://github.com/memchr/Hyprland/blob/99b8a768cfe8741a0685faad8792bbf923293805/src/managers/KeybindManager.cpp#L875)
- [fullscreenActive](https://github.com/memchr/Hyprland/blob/99b8a768cfe8741a0685faad8792bbf923293805/src/managers/KeybindManager.cpp#L866)

https://github.com/memchr/Hyprland/blob/99b8a768cfe8741a0685faad8792bbf923293805/src/Compositor.cpp#L2068-L2071

The alpha value of top layer surfaces (such as the waybar) in the current monitor is set to 0.f when the workspace has a fullscreen window.
```cpp
    const auto FULLSCREEN = pWorkspace->m_bHasFullscreenWindow;
```

```cpp
    for (auto& ls : PMONITOR->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]) {
        if (!ls->fadingOut)
[...]

PR #2994, opened 2023-08-16
fix: update `alpha` of ZWLR_LAYER_SHELL_V1_LAYER_TOP only when Workspaces is active on the monitor.

memchr:
Previously, the alpha of `ZWLR_LAYER_SHELL_V1_LAYER_TOP` was set based on whether the target workspaces (pWorkspace) of `CCompositor::updateFullscreenFadeOnWorkspace` had fullscreen windows open(m_bHasFullscreenWindow). This could be inconsistent if the target workspace wasn't the active workspace of the monitor, as the active workspace may not have had a fullscreen window open at all.

After this patch, the alpha of `ZWLR_LAYER_SHELL_V1_LAYER_TOP` will only be influenced by `pWorkspace->m_bHasFullscreenWindow` if pWorkspace is the active workspace of the monitor.

Fix: #2634

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2065,9 +2065,11 @@ void CCompositor::updateFullscreenFadeOnWorkspace(CWorkspace* pWorkspace) {
 
     const auto PMONITOR = getMonitorFromID(pWorkspace->m_iMonitorID);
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
right

PR #3001, opened 2023-08-16
fix some partly opaque windows getting treated as fully opaque

fobekyo:
This fixes some partly opaque windows (e.g. firefox with a transparent sidebar) getting treated as fully opaque when Hyprland is launched from the tty.

This error only occurs when launching hyprland from the tty because the opaque region covers the whole application. If Hyperland is launched as a wayland client, the bounding box for the opaque area has the correct size. 

Ready

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -799,10 +799,6 @@ bool CWindow::opaque() {
     if (m_uSurface.xdg->surface->opaque)
         return true;
 
-    const auto EXTENTS = pixman_region32_extents(&m_uSurface.xdg->surface->opaque_region);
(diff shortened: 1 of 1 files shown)


vaxerski:
???

would you mind explaining the change?

opaque_region is in surface-local coords.

fobekyo:
As you can see in the 2nd screenshot, the sidebar of my browser is supposed to be transparent. If i check m_uSurface.xdg->surface->opaque_region though its size is 1920x1080, covering the complete window and not just the right part so background occlusion gets enabled. This is because the transparency is achieved by setting the background to an transparent color instead of the client setting the opaque_region correctly so technically this behavior is correct according to the wayland specification and should not be fixed.

The reason I didn't get this problem when launching hyperland as a wayland client is because for some reason it enabled dpi scaling which resulted in the opaque_region being only 960x540.
what I currently get:
firefox without the change
how its supposed to look:
firefox with change

fobekyo:
Yes, I closed it after realizing that.

PR #3032, opened 2023-08-20
feat(debug): add debug:suppress_errors to suppress errors.

memchr:
If `debug:suppress_errors` is set to true, then `parseErrors` will not be sent to `CHyprError`.

Available in debug build only, no-op in release build


src/config/ConfigManager.cpp
```diff
@@ -1566,7 +1567,11 @@ void CConfigManager::loadConfigLoadVars() {
         g_pHyprOpenGL->m_bReloadScreenShader = true;
 
     // parseError will be displayed next frame
+#ifdef HYPRLAND_DEBUG
+    if (parseError != "" && !configValues["debug:suppress_errors"].intValue)
```
vaxerski, line 1571: no reason to guard this for only debug. `debug:` options are generally available.

memchr: My reasoning is that if some users used this option in the hyprland release build and forgot about it. then reported a issue when they found their configuration did not work...

Well, that would be very time-consuming.

memchr: So, should I remove the preprocessor guard?

vaxerski: yes

memchr: Removed the preprocessor guard, thus marked as resolved.

memchr (commented):
As with the requirement, the preprocessor guard has been removed.

However, with the utmost respect, may I humbly suggest that consideration be given to adding a warning within the issue template, **advising users to check for and disable `debug:suppress_errors`** before submitting an issue report?

vaxerski (approved):
g

vaxerski:
> However, with the utmost respect, may I humbly suggest that consideration be given to adding a warning within the issue template, **advising users to check for and disable `debug:suppress_errors`** before submitting an issue report?

We can. Although strict mentions in the wiki should suffice imo

Issue #2451, opened 2023-06-06
Autofocus first window when no window is focused in a workspace


ItsProfessional:
I have firefox, wezterm, and discord autostart in my hyprland config, which are also assigned workspaces 1, 2 and 3 using window rules.

When hyprland starts, workspace 3 becomes the current workspace with discord is focused in it (I assume it's because discord is the window that starts at last)

All of that is fine, but when I switch to workspace 1 or 2 using keybinds, I assume that firefox and wezterm will be FOCUSED in those workspaces, but that is not the case.

So I have to manually focus it by clicking on windows (because I have `follow-mouse` set to `2` since I don't like hover selection). This means I can't type anything in the terminal without focusing it manually.

An option in hyprland to automatically focus the top-left-most window if there is no window focused in a workspace would be appreciated. I am coming from Sway, which has this feature built in, and I don't know why I should have to manually script such a basic feature like this

memchr:
Hi, would you mind attaching your Hyprland configurations? Thanks.

woojiq:
I'm not issue author but also encounted this. My config:
```conf
exec-once = hyprctl keyword windowrule "workspace 1 silent, Google Chrome"
exec-once = hyprctl keyword windowrule "workspace 2 silent, wezterm"
exec-once = google-chrome-stable
exec-once = wezterm
exec-once = /nix/store/bnx8z8asxh86ghxb7469k460zrzix2bl-cleanup-script
```
And `cleanup-script`:
```sh
sleep 4
hyprctl keyword windowrule "workspace unset, google-chrome-stable"
hyprctl keyword windowrule "workspace unset, wezterm"
```
Shortcuts for switching workspace:
```
bind = $mainMod, 1, workspace, 1
bind = $mainMod, 2, workspace, 2
bind = $mainMod, 3, workspace, 3
...
```
My full config [home-manager module](https://github.com/woojiq/nixos-config/blob/main/modules/desktop/hyprland/home.nix)

I noticed that switching the workspace does not focus a windows, but I have `bind = $altMod, Tab, focuscurrentorlast`. And on launch i press `alt-tab` and it opens focused second workspace. A little workaround.

memchr:
@woojiq Can you try this patch?

[0001-focus-first-window.patch.txt](https://github.com/hyprwm/Hyprland/files/12391958/0001-focus-first-window.patch.txt)

woojiq:
[...]

PR #3041, opened 2023-08-21
fix: improved focus behaviour on workspace switching for follow_mouse!=1

memchr:
## Abstract

When `input:follow_mouse` is not set to 1 (`Cursor movement will always change focus to the window under the cursor`), focus the top-left-most window on a workspace if there is no last focused window when switching workspaces.

Fix: #2451

## Motivation

Currently, if there is no last focused window on a workspace, switching to that workspace results in no focused window. This is particularly inconvenient for users when `input:follow_mouse != 1`, as they will be forced to use the mouse to focus a window, thus degrading the experience. 

## Implementation

This PR updates the `CMonitor::changeWorkspace()` method to check if there is no last focused window, and if so, focus the top-left-most window found for that workspace instead.

The top-left-most window is determined by calling the new `CCompositor::getTopLeftWindowOnWorkspace()` method.

If `CCompositor::getTopLeftWindowOnWorkspace` returns `nullptr`, `CCompositor::getFirstWindowOnWorkspace` will be used as a fallback to find a window to focus.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1195,6 +1195,17 @@ CWindow* CCompositor::getFirstWindowOnWorkspace(const int& id) {
     return nullptr;
 }
 
+CWindow* CCompositor::getTopLeftWindowOnWorkspace(const int& id) {
+    for (auto& w : m_vWindows) {
+        if (w->m_iWorkspaceID != id || !w->m_bIsMapped || w->isHidden())
+            continue;
+        const auto WINDOWIDEALBB = w->getWindowIdealBoundingBoxIgnoreReserved();
+        if (WINDOWIDEALBB.x == 1 && WINDOWIDEALBB.y == 1)
+            return w.get();
+    }
+    return nullptr;
+}
+
 bool CCompositor::doesSeatAcceptInput(wlr_surface* surface) {
     if (g_pSessionLockManager->isSessionLocked()) {
         if (g_pSessionLockManager->isSurfaceSessionLock(surface))
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1196,11 +1196,20 @@ CWindow* CCompositor::getFirstWindowOnWorkspace(const int& id) {
 }
 
 CWindow* CCompositor::getTopLeftWindowOnWorkspace(const int& id) {
+    const auto PWORKSPACE = getWorkspaceByID(id);
+
+    if (!PWORKSPACE)
+        return nullptr;
+
+    const auto PMONITOR = getMonitorFromID(PWORKSPACE->m_iMonitorID);
+
     for (auto& w : m_vWindows) {
         if (w->m_iWorkspaceID != id || !w->m_bIsMapped || w->isHidden())
             continue;
+
         const auto WINDOWIDEALBB = w->getWindowIdealBoundingBoxIgnoreReserved();
-        if (WINDOWIDEALBB.x == 1 && WINDOWIDEALBB.y == 1)
+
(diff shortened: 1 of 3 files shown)


src/helpers/Monitor.cpp
```diff
@@ -553,7 +553,12 @@ void CMonitor::changeWorkspace(CWorkspace* const pWorkspace, bool internal) {
 
         if (const auto PLASTWINDOW = pWorkspace->getLastFocusedWindow(); PLASTWINDOW)
             g_pCompositor->focusWindow(PLASTWINDOW);
-        else {
+        else if (g_pConfigManager->getConfigValuePtr("input:follow_mouse")->intValue != 1) {
+            CWindow* pWindow = g_pCompositor->getTopLeftWindowOnWorkspace(pWorkspace->m_iID);
```
vaxerski, line 557: two things:
first, why not make this default?

second, we still need a simulateMouseMovement at the end to avoid wonky cursor coords if the mouse isnt moved

memchr: > first, why not make this default?

This is because the window under curouse would be focused if follow_mouse were set to 1.

> second, we still need a simulateMouseMovement at the end to avoid wonky cursor coords if the mouse isnt moved

Will update

memchr: > This is because the window under cursor would be focused if follow_mouse were set to 1.

This is probably going to sound a bit silly, but should it be put to a vote?

vaxerski: > This is because the window under cursor would be focused if follow_mouse were set to 1.

then focus it (if there is one) and use the rest of the logic if there isn't...?

memchr: Sorry, I don't understand. Are you talking about the conditions where `follow_mouse == 1`?

memchr: Well, the current main branch behaves like this when `follow_mouse == 1`.

memchr: I will be AFK, there is another branch [fix_2451.alt](https://github.com/memchr/Hyprland/tree/fix_2451.alt) which contains an alternative where this is the default (no follow_mouse check). If this behaviour is desired as the default, you can force push this branch to the current pull request branch. 

- IIRC GitHub already allows maintainers to push to the pull request branch on the fork by default.

Thanks for your time

vaxerski: check my commit out

memchr: Splendid, that will do.

src/Compositor.cpp
```diff
@@ -1195,6 +1195,17 @@ CWindow* CCompositor::getFirstWindowOnWorkspace(const int& id) {
     return nullptr;
 }
 
+CWindow* CCompositor::getTopLeftWindowOnWorkspace(const int& id) {
+    for (auto& w : m_vWindows) {
+        if (w->m_iWorkspaceID != id || !w->m_bIsMapped || w->isHidden())
+            continue;
+        const auto WINDOWIDEALBB = w->getWindowIdealBoundingBoxIgnoreReserved();
+        if (WINDOWIDEALBB.x == 1 && WINDOWIDEALBB.y == 1)
```
vaxerski, line 1203: spotted this: this will break on multimon. We should check against the workspace's monitor's pos, not 1

vaxerski: fixd

vaxerski:
I made some changes, but since I didn't have the issue I will let you test and verify everything's still in order.

memchr:
> I made some changes, but since I didn't have the issue I will let you test and verify everything's still in order.

Yes, it passed the test on a VM with multi-head outputs.

vaxerski:
top g

vaxerski (approved):
I'm blue

PR #3052, opened 2023-08-22
Swipe direction lock

end-4:
- Adds the config option `gestures:workspace_swipe_direction_lock`, which forces the destination to be in one direction only once the user starts swiping. 
  - True by default
  - Let's say the user wants to switch to workspace on the right. They swipe halfway and change their mind and swipe back to cancel that. Setting this to true prevents over-swiping to the workspace on the left.

- Since the option is true by default, it introduces a change in default behavior. But I guess it makes sense (Windows is like that).

- Ready

### DIFF
--- a/src/managers/input/Swipe.cpp
+++ b/src/managers/input/Swipe.cpp
@@ -228,11 +228,12 @@ void CInputManager::onSwipeUpdate(wlr_pointer_swipe_update_event* e) {
         return;
     }
 
-    if (*PSWIPEDIRLOCK)
+    if (*PSWIPEDIRLOCK) {
         if (m_sActiveSwipe.initialDirection != 0 && m_sActiveSwipe.initialDirection != (m_sActiveSwipe.delta < 0 ? -1 : 1))
             m_sActiveSwipe.delta = 0;
         else if (m_sActiveSwipe.initialDirection == 0 && abs(m_sActiveSwipe.delta) > *PSWIPEDIRLOCKTHRESHOLD)
             m_sActiveSwipe.initialDirection = m_sActiveSwipe.delta < 0 ? -1 : 1;
+    }
 
     if (m_sActiveSwipe.delta < 0) {
         const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(workspaceIDLeft);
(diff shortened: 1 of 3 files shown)


src/managers/input/Swipe.cpp
```diff
@@ -225,6 +227,14 @@ void CInputManager::onSwipeUpdate(wlr_pointer_swipe_update_event* e) {
         return;
     }
 
+    if (*PSWIPEDIRLOCK) {
+        if (m_sActiveSwipe.initialDirection != 0 && m_sActiveSwipe.initialDirection != (m_sActiveSwipe.delta < 0 ? -1 : 1)) {
```
vaxerski, line 231: I'd rather have a minimum threshold for the thing to initiate the lock. When you first put your fingers, without moving them, it might jump a few pixels around left and right

src/managers/input/Swipe.cpp
```diff
@@ -225,6 +227,14 @@ void CInputManager::onSwipeUpdate(wlr_pointer_swipe_update_event* e) {
...
+            m_sActiveSwipe.delta = 0;
+        } else if (m_sActiveSwipe.initialDirection == 0) {
```
vaxerski, line 233: style no {} around short ifs

end-4: https://github.com/hyprwm/Hyprland/pull/3052/commits/89a055a45aa47d0a5a3c8b41a3ef098ce8de77da

end-4:
Done
I *have* thought about threshold but there's already some kind of threshold for me. Guess that was a hardware-specific thing.
Now, is 10px enough for fingers of an average linux user, however fat they might be?

vaxerski:
unless someone is as fat as nicocado avocado then it should be enough, and in the case they are, we recommend the gym and not hyprland, so this should be enough

src/managers/input/Swipe.cpp
```diff
@@ -225,6 +228,14 @@ void CInputManager::onSwipeUpdate(wlr_pointer_swipe_update_event* e) {
         return;
     }
 
+    if (*PSWIPEDIRLOCK) {
+        if (m_sActiveSwipe.initialDirection != 0 && m_sActiveSwipe.initialDirection != (m_sActiveSwipe.delta < 0 ? -1 : 1))
+            m_sActiveSwipe.delta = 0;
+        else if (m_sActiveSwipe.initialDirection == 0 && abs(m_sActiveSwipe.delta) > *PSWIPEDIRLOCKTHRESHOLD) {
```
vaxerski, line 234: no {}

end-4: https://github.com/hyprwm/Hyprland/pull/3052/commits/da2eb3eaf99ea3436adc45a2a9181952cf4c312b

vaxerski (approved):
top g

PR #3056, opened 2023-08-23
nix module: Fix attempted workaround for option name change

TLATER:

This commit fixes an erroneous use of `system.stateVersion`, which causes warnings for NixOS-unstable users.

`system.stateVersion` is about *system state*, i.e., stuff that was created by old software versions on user's machines, like SQL databases that have moved to a different directory or such.

It will not help figure out whether an option has been renamed, because a NixOS 23.11 system may have *any* version set as `system.stateVersion`, and should match the version on which the user's NixOS was installed, which will often be much older.

Don't believe so; I double checked against stable and unstable to be sure this evaluates, didn't actually deploy though. @fufexan should probably review :)

Think it's ready!

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -2,6 +2,7 @@ inputs: {
   config,
   lib,
   pkgs,
+  options,
   ...
 }:
 with lib; let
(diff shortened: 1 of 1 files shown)


fufexan:
Thank you!

It didn't occur to me this wouldn't properly fix the issue, as I've bumped my `stateVersion` when the warning about `enableDefaultFonts` being deprecated started appearing, and that made the warning go away.

fufexan (approved):

fufexan:
Built my system using this change, no errors.

Issue #3060, opened 2023-08-24
Segfault when using tilde in path for input.kb_file


t-veor:
### Hyprland Version

Hyprland, built from branch main at commit 41d9b6f0d72947de1b0871795f8d8af32d0e9f98 dirty (dwindle: use smart on display borders (2897)). Tag: v0.28.0-81-g41d9b6f0  flags: (if any)

### Bug or Regression?

Bug

### Description

Hyprland segfaults if you use `~` in your path specification for `input.kb_file`. I didn't dig too deep into it, but it looks like the issue is here:
https://github.com/hyprwm/Hyprland/blob/17ea7db23a21a3150343b3ab0c9404dff8d198b8/src/managers/input/InputManager.cpp#L784-L794

`FILEPATH` is resolved into an absolute path, checked to see if the file exists... and then the unresolved `FILEPATH` is passed into `fopen`. So, if `FILEPATH` was `~/.config/hyprland/mykeymap.xkb`, it gets expanded into `/home/username/.config/hyprland/mykeymap.xkb`, which `std::filesystem::exists` reports exists. Then, `~/.config/hyprland/mykeymap.xkb` is passed into `fopen`, which returns `NULL`. There's no null pointer check on `KEYMAPFILE`, so `xkb_keymap_new_from_file` blows up.

(In this case, wouldn't it be better to just try and go ahead and open the file first... and then check if it's `NULL`, instead of trying to use `std::filesystem::exists`? Since something else on the system could delete/move/rename the file in between the `std::filesystem::exists` check and actually opening it with `fopen`.)

When doing a search through the repository I also found this code which is vulnerable to this issue, although it doesn't appear to be the cause of my segfault: https://github.com/hyprwm/Hyprland/blob/17ea7db23a21a3150343b3ab0c9404dff8d198b8/src/managers/KeybindManager.cpp#L145-L148

### How to reproduce

Ensure a file called `existing_file.xkb` exists in your home directory:
```bash
touch ~/existing_file.xkb
```

Put this in your `hyprland.conf`:
```
input {
  kb_file = ~/existing_file.xkb
}
```

launch Hyprland.

### Crash reports, logs, images, videos

[crash.txt](https://github.com/hyprwm/Hyprland/files/12424218/crash.txt)

memchr:
A grave mistake indeed. tlide(`~`) is interpreted by the shell, not the c library function.

[...]

PR #3061, opened 2023-08-24
fix: filepath handling and null check return value of fopen

memchr:
## Abstract
Improved file path handling in the KeybindManager and InputManager classes by adding null checking for the return value of fopen and resolving the absolute path of `input:kb_file'.

## Motivation

The original code did not properly handle tildes in file paths or check for null return values from fopen. These oversights could lead to runtime errors.

Also, the previous use of `std::filesystem::exists` only checked whether the file existed, not whether it could be read.

Fixes: #3060

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -139,13 +139,12 @@ void CKeybindManager::updateXKBTranslationState() {
     const auto     VARIANT  = g_pConfigManager->getString("input:kb_variant");
     const auto     OPTIONS  = g_pConfigManager->getString("input:kb_options");
 
-    xkb_rule_names rules = {.rules = RULES.c_str(), .model = MODEL.c_str(), .layout = LAYOUT.c_str(), .variant = VARIANT.c_str(), .options = OPTIONS.c_str()};
-
+    xkb_rule_names rules      = {.rules = RULES.c_str(), .model = MODEL.c_str(), .layout = LAYOUT.c_str(), .variant = VARIANT.c_str(), .options = OPTIONS.c_str()};
     const auto     PCONTEXT   = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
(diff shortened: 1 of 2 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -142,7 +142,7 @@ void CKeybindManager::updateXKBTranslationState() {
     xkb_rule_names rules = {.rules = RULES.c_str(), .model = MODEL.c_str(), .layout = LAYOUT.c_str(), .variant = VARIANT.c_str(), .options = OPTIONS.c_str()};
 
     const auto     PCONTEXT   = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
-    FILE* const    KEYMAPFILE = FILEPATH == "" ? NULL : fopen(FILEPATH.c_str(), "r");
+    FILE* const    KEYMAPFILE = FILEPATH == "" ? NULL : fopen(absolutePath(FILEPATH, g_pConfigManager->configCurrentPath).c_str(), "r");
```
t-veor, line 145: This change doesn't quite fix the issue, right? Now `FILEPATH` is correctly expanded, but if `FILEPATH` is non-empty but refers to a nonexistent file, this will still attempt to call `xkb_keymap_new_from_file` on a null pointer.

src/managers/input/InputManager.cpp
```diff
@@ -784,10 +784,9 @@ void CInputManager::applyConfigToKeyboard(SKeyboard* pKeyboard) {
     if (!FILEPATH.empty()) {
         auto path = absolutePath(FILEPATH, g_pConfigManager->configCurrentPath);
 
-        if (!std::filesystem::exists(path))
+        if (FILE* const KEYMAPFILE = fopen(path.c_str(), "r"); !KEYMAPFILE)
             Debug::log(ERR, "input:kb_file= file doesnt exist");
```
t-veor, line 788: It may be more accurate now to say `cannot open file for reading`, because this check is true when the file cannot be opened for reading. It might exist but we might lack the permissions to read it, etc.

vaxerski (approved):
ok

PR #3095, opened 2023-08-27
build: use PCH to reduce compile time

memchr:
Using precompiled headers can consistently reduce initial build time by approximately 26%, and incremental build time by 30% to 40% in relative terms. However, this comes at the cost of increased disk usage during compilation.

For example, on a single zen2 core, compilation time, not including the linking process:

| PCH | initial build | incremental build | total object files generated |
| --- | ------------- | ----------------- | ---------------------------- |
| no  | 238.22s       |    36.839s        | 244.8 MiB                    |
| yes | 186.16s        |     22.202s       | 505.6 MiB                    |

During meson setup PCH can be disabled with `b_pch=false`
```bash
meson setup build -Db_pch=false
```
During cmake setup PCH can be disabled with `CMAKE_DISABLE_PRECOMPILE_HEADERS=ON`
```bash
cmake -S . -B ./build -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON
```

### Caveats

Using precompiled headers makes compilation of source files in src/ uncacheable by ccache

However, this is unlikely to have a significant impact, as the use of global variables and frequent header changes often invalidate the previous cache, leading to a full rebuild.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,7 @@ endif()
 
 include_directories(
   .
+  "src/"
   "subprojects/wlroots/include/"
   "subprojects/wlroots/build/include/"
   "subprojects/udis86/"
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
ok from cmake side

vaxerski:
fuffie got drafted and he should be back next week I believe

fufexan (approved):
LGTM. Reduced my build time from 42s to 33s. Considering Nix builds are always "initial", this is a great reduction.

PR #3102, opened 2023-08-29
groupbar mouse interaction

MightyPlaza:
allows mouse to better interact with groupbar:

- clicking on window bar to set it as active
- drag specific window from groupbar (movewindow)
- drag window into group's groupbar to move it into the group at a specific position (movewindow)

dragging out of floating groups is not allowed, because of crashes (same that happens with moveoutofgroup in main branch)
I'll look into that later

future improvements:
- allow constricted moving window inside group (hold mouse down normal)
- visual hints for what's happening
- fix cursor icon

didn't find any major issues
should be ready for merge

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -170,6 +170,19 @@ wlr_box CWindow::getWindowInputBox() {
     return finalBox;
 }
 
+wlr_box CWindow::getWindowGroupBarBox() {
+    const auto PWORKSPACE      = g_pCompositor->getWorkspaceByID(m_iWorkspaceID);
+    const auto WORKSPACEOFFSET = PWORKSPACE && !m_bPinned ? PWORKSPACE->m_vRenderOffset.vec() : Vector2D();
+    const int  BORDERSIZE      = getRealBorderSize();
+    for (auto& wd : m_dWindowDecorations) {
+        if (wd->getDecorationType() == DECORATION_GROUPBAR) {
+            const SWindowDecorationExtents EXTENTS = wd->getWindowDecorationReservedArea();
+            return {m_vRealPosition.vec().x, m_vRealPosition.vec().y - BORDERSIZE - EXTENTS.topLeft.y, m_vRealSize.vec().x, EXTENTS.topLeft.y};
+        }
+    }
+    return {0, 0, 0, 0};
+}
+
 SWindowDecorationExtents CWindow::getFullWindowReservedArea() {
     SWindowDecorationExtents extents;
 
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -170,19 +170,6 @@ wlr_box CWindow::getWindowInputBox() {
     return finalBox;
 }
 
-wlr_box CWindow::getWindowGroupBarBox() {
-    const auto PWORKSPACE      = g_pCompositor->getWorkspaceByID(m_iWorkspaceID);
-    const auto WORKSPACEOFFSET = PWORKSPACE && !m_bPinned ? PWORKSPACE->m_vRenderOffset.vec() : Vector2D();
-    const int  BORDERSIZE      = getRealBorderSize();
-    for (auto& wd : m_dWindowDecorations) {
-        if (wd->getDecorationType() == DECORATION_GROUPBAR) {
-            const SWindowDecorationExtents EXTENTS = wd->getWindowDecorationReservedArea();
-            return {m_vRealPosition.vec().x, m_vRealPosition.vec().y - BORDERSIZE - EXTENTS.topLeft.y, m_vRealSize.vec().x, EXTENTS.topLeft.y};
-        }
-    }
-    return {0, 0, 0, 0};
-}
-
 SWindowDecorationExtents CWindow::getFullWindowReservedArea() {
     SWindowDecorationExtents extents;
 
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -714,10 +714,8 @@ void CWindow::setGroupCurrent(CWindow* pWindow) {
 }
 
 void CWindow::insertWindowToGroup(CWindow* pWindow) {
-    static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("misc:group_insert_after_current")->intValue;
-
-    const auto         BEGINAT = *USECURRPOS ? this : getGroupTail();
-    const auto         ENDAT   = *USECURRPOS ? m_sGroupData.pNextWindow : getGroupHead();
+    const auto BEGINAT = this;
+    const auto ENDAT   = m_sGroupData.pNextWindow;
 
     if (!pWindow->m_sGroupData.pNextWindow) {
         BEGINAT->m_sGroupData.pNextWindow = pWindow;
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -678,7 +678,7 @@ int CWindow::getGroupSize() {
 
 CWindow* CWindow::getGroupWindowByIndex(int index) {
     const int SIZE = getGroupSize();
-    index          = index % SIZE;
+    index          = ((index % SIZE) + SIZE) % SIZE;
     CWindow* curr  = getGroupHead();
     while (index > 0) {
         curr = curr->m_sGroupData.pNextWindow;
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -666,6 +666,27 @@ CWindow* CWindow::getGroupCurrent() {
     return curr;
 }
 
+int CWindow::getGroupSize() {
+    int      size = 1;
+    CWindow* curr = this;
+    while (curr->m_sGroupData.pNextWindow != this) {
+        curr = curr->m_sGroupData.pNextWindow;
+        size++;
+    }
+    return size;
+}
+
+CWindow* CWindow::getGroupWindowByIndex(int index) {
+    const int SIZE = getGroupSize();
+    index          = index % SIZE;
+    CWindow* curr  = getGroupHead();
+    while (index > 0) {
+        curr = curr->m_sGroupData.pNextWindow;
+        index--;
+    }
+    return curr;
+}
+
 void CWindow::setGroupCurrent(CWindow* pWindow) {
     CWindow* curr     = this->m_sGroupData.pNextWindow;
     bool     isMember = false;
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -312,6 +312,8 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow) {
         return;
     }
 
+    const auto MOUSECOORDS = g_pInputManager->getMouseCoordsInternal();
+
     // if it's a group, add the window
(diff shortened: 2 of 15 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -312,28 +312,33 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
-            g_pCompositor->focusWindow(pWindow);
-            return;
-        }
+            const wlr_box box = OPENINGON->pWindow->getWindowGroupBarBox();
+            if (wlr_box_contains_point(&box, MOUSECOORDS.x, MOUSECOORDS.y)) { // TODO: Deny when not using mouse
+                const int SIZE               = OPENINGON->pWindow->getGroupSize();
+                const int INDEX              = (int)((MOUSECOORDS.x - box.x) * 2 * SIZE / box.width + 1) / 2 - 1;
+                CWindow*  pWindowInsertAfter = OPENINGON->pWindow->getGroupWindowByIndex(INDEX);
+                pWindowInsertAfter->insertWindowToGroup(pWindow);
+                if (INDEX == -1) {
```
vaxerski, line 333: no {}

src/Window.cpp
```diff
@@ -170,6 +170,19 @@ wlr_box CWindow::getWindowInputBox() {
     return finalBox;
 }
 
+wlr_box CWindow::getWindowGroupBarBox() {
+    const auto PWORKSPACE      = g_pCompositor->getWorkspaceByID(m_iWorkspaceID);
```
vaxerski, line 174: wouldn't this be better as a generic?

getWindowDecorationBox(type)

src/layout/DwindleLayout.cpp
MightyPlaza: done

src/Window.cpp
vaxerski: you could make another func then `isPointInDecoration` with an additional `bool` parameter of `ignoreWindow`

MightyPlaza: I would have to either pass a decoration type (wouldn't allow duplicated decorations) or the pointer to the decoration (would create redundant code)

maybe a solution is adding functions for generalized decorations and basing the groupbar functions on those (which doesn't remove the need for the function)

vaxerski: why would we allow duplicate decorations?

vaxerski: idk, for now I don't believe we want duplicate decos. `getDecorationByType` and `removeDecorationByType` already exist

MightyPlaza: seems like decorations don't depend on a window
should each decoration implement it's own `getDecorationBox()` function?
I was trying to put that in the class `IHyprWindowDecoration`

vaxerski: > seems like decorations don't depend on a window

they do...?

> should each decoration implement it's own getDecorationBox() function?

probably

MightyPlaza: the m_pWindow isn't implemented in `IHyprWindowDecoration`, but in each of the decorations
that would have to be moved to `IHyprWindowDecoration` to implement a generalized `getDecorationBox()` function
I think for now I'm just gonna put it in `CHyprGroupBarDecoration`

vaxerski: `m_pWindow` can be moved to the deco iface

vaxerski: generally looks good but the switch in the box get func is an odd method for sure

MightyPlaza: need to separate when when decoration is just in one side of the window, since the box won't include the window itself
no idea how to do that easier with bool algebra

should I add in PR for review?

vaxerski: you can commit yea, easier to read.

MightyPlaza:
i found the issue with the crashing floating grouped windows, might commit here as well and remove the dragging limitation

src/Window.cpp
MightyPlaza: done

src/render/decorations/IHyprWindowDecoration.cpp
```diff
@@ -2,12 +2,37 @@
...
+    m_pWindow = pWindow;
+}
+
 IHyprWindowDecoration::~IHyprWindowDecoration() {}
 
 SWindowDecorationExtents IHyprWindowDecoration::getWindowDecorationReservedArea() {
     return SWindowDecorationExtents{};
 }
 
+wlr_box IHyprWindowDecoration::getWindowDecorationBox() {
```
vaxerski, line 15: we could use a CRegion here. Would also be easier to check against later.

Probably can make a simple CRegion::pointInside(const Vector2D&) as a helper

solves also the issue with the box of the window not being a part of the deco

MightyPlaza: done

MightyPlaza:
might be a good idea to create a group related section in the config to reduce `misc` size

vaxerski:
probably, but that's for a separate mr

src/render/decorations/IHyprWindowDecoration.cpp
```diff
@@ -2,12 +2,29 @@
...
+    m_pWindow = pWindow;
+}
+
 IHyprWindowDecoration::~IHyprWindowDecoration() {}
 
 SWindowDecorationExtents IHyprWindowDecoration::getWindowDecorationReservedArea() {
     return SWindowDecorationExtents{};
 }
 
+CRegion IHyprWindowDecoration::getWindowDecorationBox() {
```
vaxerski, line 15: since it returns a region, this should not be called a box.

src/managers/input/InputManager.cpp
```diff
@@ -545,11 +545,25 @@ void CInputManager::processMouseDownNormal(wlr_pointer_button_event* e) {
     if (!PASS && !*PPASSMOUSE)
         return;
 
+    const auto mouseCoords = g_pInputManager->getMouseCoordsInternal();
+    const auto w           = g_pCompositor->vectorToWindowIdeal(mouseCoords);
+
+    if (w && !w->m_bIsFullscreen && !w->hasPopupAt(mouseCoords) && w->m_sGroupData.pNextWindow) {
+        const wlr_box box = w->getDecorationByType(DECORATION_GROUPBAR)->getWindowDecorationBox().getExtents();
```
vaxerski, line 552: you made a CRegion::containsPoint for this exact reason no?

src/managers/input/InputManager.cpp
```diff
@@ -613,14 +627,29 @@ void CInputManager::processMouseDownKill(wlr_pointer_button_event* e) {
...
     bool               passEvent = g_pKeybindManager->onAxisEvent(e);
 
     g_pCompositor->notifyIdleActivity();
 
+    const auto MOUSECOORDS = g_pInputManager->getMouseCoordsInternal();
+    const auto pWindow     = g_pCompositor->vectorToWindowIdeal(MOUSECOORDS);
+
+    if (*PGROUPBARSCROLLING && pWindow && !pWindow->m_bIsFullscreen && !pWindow->hasPopupAt(MOUSECOORDS) && pWindow->m_sGroupData.pNextWindow) {
+        const wlr_box box = pWindow->getDecorationByType(DECORATION_GROUPBAR)->getWindowDecorationBox().getExtents();
+        if (wlr_box_contains_point(&box, MOUSECOORDS.x, MOUSECOORDS.y)) {
```
vaxerski, line 644: ditto

src/managers/KeybindManager.cpp
```diff
@@ -1951,9 +1951,26 @@ void CKeybindManager::mouse(std::string args) {
...
             g_pKeybindManager->m_bIsMouseBindActive = true;
 
-            g_pInputManager->currentlyDraggedWindow = g_pCompositor->vectorToWindowIdeal(g_pInputManager->getMouseCoordsInternal());
-            g_pInputManager->dragMode               = MBIND_MOVE;
+            const auto mouseCoords = g_pInputManager->getMouseCoordsInternal();
+            CWindow*   pWindow     = g_pCompositor->vectorToWindowIdeal(mouseCoords);
+
+            if (pWindow && !pWindow->m_bIsFullscreen && !pWindow->hasPopupAt(mouseCoords) && pWindow->m_sGroupData.pNextWindow) {
+                const wlr_box box = pWindow->getDecorationByType(DECORATION_GROUPBAR)->getWindowDecorationBox().getExtents();
+                if (wlr_box_contains_point(&box, mouseCoords.x, mouseCoords.y)) {
```
vaxerski, line 1959: ditto

src/layout/MasterLayout.cpp
```diff
@@ -91,28 +91,32 @@ void CHyprMasterLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
-            pWindow->m_dWindowDecorations.emplace_back(std::make_unique<CHyprGroupBarDecoration>(pWindow));
-            pWindow->updateWindowDecos();
-            recalculateWindow(pWindow);
+            m_lMasterNodesData.remove(*PNODE);
 
-            g_pCompositor->focusWindow(pWindow);
-            return;
-        }
+            const wlr_box box = OPENINGON->pWindow->getDecorationByType(DECORATION_GROUPBAR)->getWindowDecorationBox().getExtents();
+            if (wlr_box_contains_point(&box, MOUSECOORDS.x, MOUSECOORDS.y)) { // TODO: Deny when not using mouse
```
vaxerski, line 107: ditto

src/layout/DwindleLayout.cpp
```diff
@@ -312,28 +312,32 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow) {
...
-            pWindow->m_dWindowDecorations.emplace_back(std::make_unique<CHyprGroupBarDecoration>(pWindow));
-            pWindow->updateWindowDecos();
-            recalculateWindow(pWindow);
+            m_lDwindleNodesData.remove(*PNODE);
 
-            g_pCompositor->focusWindow(pWindow);
-            return;
-        }
+            const wlr_box box = OPENINGON->pWindow->getDecorationByType(DECORATION_GROUPBAR)->getWindowDecorationBox().getExtents();
+            if (wlr_box_contains_point(&box, MOUSECOORDS.x, MOUSECOORDS.y)) { // TODO: Deny when not using mouse
```
vaxerski, line 328: ditto

vaxerski (commented):
other than that, this looks alright to me

src/managers/input/InputManager.cpp
MightyPlaza: I need the box info to calculate what window should be selected (for example)
no point in using the CRegion::containsPoint since I would still need to get the `box.x` and `box.width`

ideal solution would be for each window's tab to be it's own pixman box all making up the pixman region
that way something like `CRegion::boxIndex(const Vector2D&)` would give the needed group window index

other solution is each window would be responsible for its own group tab (wouldn't be hidden), which all together would make up the full bar

vaxerski: hm, fair point. Alright.

vaxerski (approved):
lgtm

Issue #2368, opened 2023-05-26
Window rule doesn't allow `workspace empty` to open in empty workspace


foonathan:
I have the following rule:

```
windowrule=workspace empty silent,^firefox$
```

What I expect to happen: every firefox window opens in a new workspace.

What actually happens: I get an error message in the log

```
[ERR] Invalid workspace requested in workspace silent rule!
```

BowlBird:
Same problem. 
```
windowrule=workspace empty, title:Sharing indicator
```
successfully puts the workspace on a empty workspace, but
```
windowrule=workspace empty silent, title:Sharing Indicator
```
causes the error
xdg-desktop-portal-hyprland[53792]: sh: line 1: workspace: command not found
in journalctl

MightyPlaza:
@vaxerski

PR #3116, opened 2023-08-31
mapWindow use getWorkspaceIDFromString()

MightyPlaza:
allows non `silent` on special workspaces and uses `getWorkspaceIDFromString()`

Fixes:
https://github.com/hyprwm/Hyprland/issues/2368

didn't get any crashes or bad behaviour when testing:
- `workspace normal/named/special/named special (silent)`
- `workspace empty (silent)`

probably ready for merge, but its not a bad idea to test a it bit more


vaxerski (approved):
top g from my g

PR #3124, opened 2023-08-31
refactor: raise SIGABRT instead of nullptr write

memchr:
Use `raise(SIBABRT)` instead of `*((int*)nullptr) = 1337` because writing to the null address may not guarantee a reliable crash, and Hyprland registers a signal handler which will eventually call `abort()`.


vaxerski (approved):
k

PR #3128, opened 2023-09-01
Add orientationcycle command on MasterLayout

shadowmax31:
This allows to cycle through the provided list of orientation. The main goal of this is to be able to skip orientations that I don't want to use. I only use 2 orientations and orientationnext cycles through all five which is a bit annoying for my use.

I also could have set 2  keys binds (one per orientation), but I would rather have one that I can use to toggle/cycle through the orientations I want to use.

Any value not present in the the parameters will not be used to cycle.
Allowed values: top, right, bottom, left, center

e.g.
`bind = $mainMod, X, layoutmsg, orientationcycle left bottom top`
Would cycle from left to bottom to top to left...

- It's been a while since I wrote C++, so let me know if you want me to change anything
- If this is accepted, I'll gladly make a PR for the documentation too


vaxerski (changes requested):
this impl kinda stinks. Why not just read the current orientation and call `layoutMessage` with `orientation(...)`?

shadowmax31:
Sure I can do it like this, but I would still need to figure out what the next orientation should be from the list.

`buildOrientationCycle` would still be needed and this too:
```
        size_t     next = 0;
        for (size_t i = 0; i < cycle.size(); ++i) {
            if (PWORKSPACEDATA->orientation == cycle[i]) {
                next = i + 1;
                break;
            }
        }
```

Unless I am not seeing something, this does not change much

vaxerski:
it de-dups code

>  but I would still need to figure out what the next orientation should be from the list

PWORKSPACEDATA->orientation + 1
if ( > _center)
 PWORKSPACEDATA->orientation = 0
 
 ?

shadowmax31:
I agree, but I cannot use `PWORKSPACEDATA->orientation + 1`, because the next orientation is not always the correct one. 

What I could do is rewrite `orientationnext` and `orientationprevious` to use `orientationcyle`. It would add more stuff to `orientationnext` and `orientationprevious`, but would remove duplication.

Does that make sense?

vaxerski:
> I agree, but I cannot use PWORKSPACEDATA->orientation + 1, because the next orientation is not always the correct one.

thats why I said to check against _CENTER which is last and if it's over that loop back to 0.

Yes, that solution you posted is also fine.

shadowmax31:
> thats why I said to check against _CENTER which is last and if it's over that loop back to 0.

What I meant was that `PWORKSPACEDATA->orientation + 1` will not allow me to skip an orientation LEFT, SKIP TOP, RIGHT, ...

Anyways, I'll make the change later today.

src/layout/MasterLayout.hpp
```diff
@@ -31,7 +32,7 @@ struct SMasterNodeData {
     int      workspaceID = -1;
 
     bool     operator==(const SMasterNodeData& rhs) const {
-        return pWindow == rhs.pWindow;
+            return pWindow == rhs.pWindow;
```
shadowmax31, line 35: clang-format changed this. I figured I would leave it in

src/layout/MasterLayout.hpp
```diff
@@ -40,7 +41,7 @@ struct SMasterWorkspaceData {
     eOrientation orientation = ORIENTATION_LEFT;
 
     bool         operator==(const SMasterWorkspaceData& rhs) const {
-        return workspaceID == rhs.workspaceID;
+                return workspaceID == rhs.workspaceID;
```
shadowmax31, line 44: same here

shadowmax31:
Should be good to go

src/layout/MasterLayout.cpp
```diff
@@ -1071,42 +1071,71 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
+    if (vars != nullptr)
+        buildOrientationCycleVectorFromVars(&cycle, vars);
 
-        if (PWORKSPACEDATA->orientation == ORIENTATION_CENTER) {
-            PWORKSPACEDATA->orientation = ORIENTATION_LEFT;
-        } else {
-            PWORKSPACEDATA->orientation = (eOrientation)(PWORKSPACEDATA->orientation + 1);
-        }
+    if (cycle.size() == 0)
+        buildOrientationCycleVectorFromEOperation(&cycle);
```
vaxerski, line 1091: I still fail to understand why these two functions are even here.

```cpp
const auto PWORKSPACEDATA = getMasterWorkspaceData(PWINDOW->m_iWorkspaceID);

const auto NEWORIENT = (PWORKSPACEDATA->orientation + (direction % (ORIENTATION_CENTER + 1))) % (ORIENTATION_CENTER + 1);

```
?

vaxerski: okay I get you

src/layout/MasterLayout.cpp
```diff
@@ -1071,42 +1071,71 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
+}
+
+void CHyprMasterLayout::buildOrientationCycleVectorFromEOperation(std::vector<eOrientation>* cycle) {
+    for (int i = 0; i <= ORIENTATION_CENTER; ++i) {
+        cycle->push_back((eOrientation)i);
     }
+}
 
-    return 0;
+void CHyprMasterLayout::buildOrientationCycleVectorFromVars(std::vector<eOrientation>* cycle, CVarList* vars) {
```
vaxerski, line 1125: ref both

& not *

src/layout/MasterLayout.cpp
```diff
@@ -1071,42 +1071,71 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
+    if (nextOrPrev >= (int)cycle.size())
+        nextOrPrev = 0;
+    else if (nextOrPrev < 0)
+        nextOrPrev = cycle.size() - 1;
+
+    PWORKSPACEDATA->orientation = cycle.at(nextOrPrev);
+    recalculateMonitor(header->pWindow->m_iMonitorID);
+}
+
+void CHyprMasterLayout::buildOrientationCycleVectorFromEOperation(std::vector<eOrientation>* cycle) {
```
vaxerski, line 1119: ditto

src/layout/MasterLayout.cpp
```diff
@@ -1071,42 +1071,71 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
+        runOrientationCycle(&header, &vars, 1);
+    }
 
-        prepareLoseFocus(PWINDOW);
+    return 0;
+}
 
-        const auto PWORKSPACEDATA = getMasterWorkspaceData(PWINDOW->m_iWorkspaceID);
+// If vars is null, we use the default list
+void CHyprMasterLayout::runOrientationCycle(SLayoutMessageHeader* header, CVarList* vars, int direction) {
```
vaxerski, line 1085: ditto

src/layout/MasterLayout.cpp
```diff
@@ -1071,42 +1071,71 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
+    int        nextOrPrev = 0;
+    for (size_t i = 0; i < cycle.size(); ++i) {
+        if (PWORKSPACEDATA->orientation == cycle.at(i)) {
+            nextOrPrev = i + direction;
+            break;
         }
+    }
 
-        recalculateMonitor(header.pWindow->m_iMonitorID);
+    if (nextOrPrev >= (int)cycle.size())
```
vaxerski, line 1110: future-proofing this would be better. Since you already take `direction` as an int, who knows if in the future someone wont pass -3 or +1337?

```cpp
    if (nextOrPrev >= (int)cycle.size())
        nextOrPrev = nextOrPrev % (int)cycle.size();
    else if (nextOrPrev < 0)
        nextOrPrev = cycle.size() + (nextOrPrev % (int)cycle.size());
```

src/layout/MasterLayout.cpp
shadowmax31: Done

src/layout/MasterLayout.cpp
shadowmax31: Done

src/layout/MasterLayout.cpp
shadowmax31: Done

src/layout/MasterLayout.cpp
shadowmax31: Here I left CVarList as a pointer because I need to allow null to be passed for orientationnext and orientationprev:
```
std::vector<eOrientation> cycle;
if (vars != nullptr)
    buildOrientationCycleVectorFromVars(cycle, *vars);
```

What I could actually do is to remove the new command (orientationcycle) and just pass my new list to orientationext and orientationprev. That would remove the need to pass CVarList as null.

Like this:
```
bind = $mainMod, J, layoutmsg, orientationnext left right center
bind = $mainMod, K, layoutmsg, orientationprev
```
**Current configs would keep working.**

vaxerski (approved):
oke doke

vaxerski:
waiting for a wiki MR, then will be merged

Issue #1746, opened 2023-03-09
Add text-input-v1 support 


dragove:
support text-input-v1 can help to make fcitx5 work in electron apps on native wayland.

support in kde is here: https://invent.kde.org/plasma/kwin/-/merge_requests/3403

vaxerski:
chromium patchset for TI3: https://chromium-review.googlesource.com/c/chromium/src/+/3220698

vaxerski:
Do you have any other apps that use v1 other than chromium? For some weird reason chromium doesn't wanna send any events even though it binds properly. Maybe it also expects ime v1? tho that would be fucking weird.

I-Want-ToBelieve:
Did you pass the parameters that chromium needs? --ozone-platform=wayland --enable-wayland-ime --gtk-version=4

I-Want-ToBelieve:
I found something interesting. Fcitx5 running on hyprland can work in Microsoft’s edge browser.

vaxerski:
--gtk-version=4 forces TI3 instead of TI1 that I am implementing.

vaxerski:
oh. `--enable-wayland-ime` did it. Thanks.

vaxerski:
#1778 btw

dragove:
I've tested #1778 with this app: https://github.com/Rocket1184/electron-netease-cloud-music and got the same result as [q234rty](https://github.com/q234rty) tested. only electron 21 and above version works.
There is a bug that I can't delete my first charater in IME pre-edit region. And I also wonder why the pre-edit region has a yellow background.
btw, It also works on vscode if I use my system electron(22) to launch it. 

![swappy-20230314-221047](https://user-images.githubusercontent.com/24190403/225028138-2782215c-3467-41aa-a628-6657a6aca3ee.png)

I can't delete the last character `a`
![swappy-20230314-221107](https://user-images.githubusercontent.com/24190403/225028119-e8a96223-d14f-438b-ad10-ed06826fa196.png)

vaxerski:
Removing first character fixed in f960f72785141639c2f68f05a27edbebeac2210f

regarding the yellow, idk. Should be `ZWP_TEXT_INPUT_V1_PREEDIT_STYLE_NONE` which should indicate no styling.

vaxerski:
I will close this as tiv1 is supported now, feel free to open issues for other bugs.

PR #3131, opened 2023-09-01
Fix preedit styling in chromium

q234rty:

`ZWP_TEXT_INPUT_V1_PREEDIT_STYLE_NONE` seems to be [ignored](https://source.chromium.org/chromium/chromium/src/+/main:out/Debug/gen/third_party/wayland-protocols/src/unstable/text-input/text-input-unstable-v1-client-protocol.h;drc=6f4c64436342c818aa41e6a5c55034e74ec9c6b6;l=283) in chromium, causing chromium to fallback on the yellow background mentioned in https://github.com/hyprwm/Hyprland/issues/1746#issuecomment-1468179122, which unfortunately doesn't look great under dark themes.

Change it to `ZWP_TEXT_INPUT_V1_PREEDIT_STYLE_HIGHLIGHT` which causes chromium to draw a thick underline under the preedit.

This PR also fixes the [length](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/1.32/unstable/text-input/text-input-unstable-v1.xml?ref_type=tags#L273) of the preedit styling so the last character gets underlined as well.

Alternatives are:
`ZWP_TEXT_INPUT_V1_PREEDIT_STYLE_DEFAULT`, which in chromium means no styling at all and preedit text will look like regular text, which is not great IMO.
`ZWP_TEXT_INPUT_V1_PREEDIT_STYLE_UNDERLINE`, which gives a thinner underline than `ZWP_TEXT_INPUT_V1_PREEDIT_STYLE_HIGHLIGHT` in chromium.
Ready.

### DIFF
--- a/src/managers/input/InputMethodRelay.cpp
+++ b/src/managers/input/InputMethodRelay.cpp
@@ -46,12 +46,12 @@ void CInputMethodRelay::onNewIME(wlr_input_method_v2* pIME) {
             } else {
                 if (PIMR->m_pWLRIME->current.preedit.text) {
                     zwp_text_input_v1_send_preedit_cursor(PTI->pV1Input->resourceImpl, PIMR->m_pWLRIME->current.preedit.cursor_begin);
-                    zwp_text_input_v1_send_preedit_styling(PTI->pV1Input->resourceImpl, 0, std::string(PIMR->m_pWLRIME->current.preedit.text).length() - 1,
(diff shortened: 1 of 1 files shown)


q234rty:
Before this PR:
图片
After this PR:
swappy-20230901_185454

vaxerski (approved):
thanks!

Issue #3144, opened 2023-09-02
Crash on automatic closing of no focused window in groupbar


0xjmz:
### Hyprland Version

f69c5469d79f31f77f4662731ac6255e285febed

### Bug or Regression?

Bug

### Description

As in the title after grouping some windows if some that is not focused closes in background, then Hyprland crashes.

### How to reproduce

Open terminal window and 'group toggle'. Then open next terminal in that group. In first terminal it suffice to run something like "sleep 5; exit" and we change focus to second terminal. After 5 second Hyprland crashes.

### Crash reports, logs, images, videos

<pre>
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff728e8a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007ffff723e668 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff72264b8 in __GI_abort () at abort.c:79
#4  0x00007ffff769ca6f in __gnu_cxx::__verbose_terminate_handler() () at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/vterminate.cc:95
#5  0x00007ffff76b011c in __cxxabiv1::__terminate(void (*)()) (handler=<optimized out>) at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:48
#6  0x00007ffff76b0189 in std::terminate() () at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:58
#7  0x00007ffff76b03ed in __cxxabiv1::__cxa_throw(void*, std::type_info*, void (*)(void*))
    (obj=<optimized out>, tinfo=0x7ffff786ab28 <typeinfo for std::bad_alloc>, dest=0x7ffff76ae680 <std::bad_alloc::~bad_alloc()>)
    at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/eh_throw.cc:98
#8  0x00007ffff769c4f3 in operator new(unsigned long) (sz=651135464502347093) at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/new_op.cc:54
[...]

PR #3146, opened 2023-09-02
fix: use-after-free of CWindow* in CHyprGroupBarDecoration::draw

memchr:
The dangling CWindow* pointer in `CHyprGroupBarDecoration.m_dwGroupMembers` triggered a use-after-free bug in `CHyprGroupBarDecoration::draw()`, leading to a crash in Hyprland when operating on `CTitleTex.szContent`.

https://github.com/hyprwm/Hyprland/blob/06d860c489b667cabe29d180b28321377f4df3d1/src/render/decorations/CHyprGroupBarDecoration.cpp#L134-L138

This patch includes a call to `CWindow::updateWindowDecos()` for the surviving group member when a window is destroyed, ensuring the removal of the dangling pointer from `CHyprGroupBarDecoration.m_dwGroupMembers`.

fixes: #3144

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -57,6 +57,7 @@ void IHyprLayout::onWindowRemoved(CWindow* pWindow) {
             pWindow->setHidden(false);
 
             pWindow->updateWindowDecos();
+            curr->getGroupCurrent()->updateWindowDecos();
             g_pCompositor->updateWindowAnimatedDecorationValues(pWindow);
 
             return;


MightyPlaza:
lol, got there 1st by a minute or so
[group_crash.txt](https://github.com/hyprwm/Hyprland/files/12504413/group_crash.txt)
will it still work with group of size > 2?

memchr:
I have to admit that I was confused by the name `curr`. It's probably less confusing to call this variable `prev`, or to use `active` instead of `current` to describe the active window in a group, e.g. `setGroupActive()`.

MightyPlaza:
`curr` is usually used to iterate through groups

also since you're already messing with the code, this section can use `CWindow::getGroupPrevious()` and `std::swap` in the `if (head))` section
since iterating is done by `CWindow::getGroupPrevious()` just call the window `const CWindow* PWINDOWPREV` or something like that

memchr:
Nice idea, but do you think it's appropriate to refactor unrelated code in a PR, since all commits would supposedly be squashed?

MightyPlaza:
since it doesn't change any behavior I don't see a reason to create another PR, but maybe ask varxy

vaxerski:
I will merge this and yall do any improvements in another pr

vaxerski (approved):
top g

PR #3157, opened 2023-09-03
Monitor rule fix

MightyPlaza:
fixes monitor exec rule being ignored when workspace isn't specified (not from https://github.com/hyprwm/Hyprland/pull/3116, just never worked somehow)

also dispatches as silent to normal workspaces if the monitor has a special workspace open


vaxerski (approved):
oki doki

Issue #2596, opened 2023-06-28
Enhancement for special workspace


A7R7:
Hi, thank you for all the amazing works! 
I have some suggestions for special workspace. 

1. Add a config option to make a special workspace to have only 1 window (or a limited number of windows)

   This is useful when you have a file manager inside a special workspace. When you open a file in the file manager, you generally hope that the file will fallback into the normal workspace, but not tiled together with the file manager in the special workspace.

2. Make "togglespecialworkspace" avaliable to toggle on special workspace that have no windows yet. 

   This makes it convenient to first toggle the special workspace on then launch something inside it.

These can all be achieved with bash scripts, but I found them to be rather tricky or have strange bugs. So I suggest them to be built-in features.

pwnfooj716:
> Hi, thank you for all the amazing works! I have some suggestions for special workspace.
> 
>     1. Add a config option to make a special workspace to have only 1 window (or a limited number of windows)
>        This is useful when you have a file manager inside a special workspace. When you open a file in the file manager, you generally hope that the file will fallback into the normal workspace, but not tiled together with the file manager in the special workspace.
> 
>     2. Make "togglespecialworkspace" avaliable to toggle on special workspace that have no windows yet.
>        This makes it convenient to first toggle the special workspace on then launch something inside it.
> 
> 
> These can all be achieved with bash scripts, but I found them to be rather tricky or have strange bugs. So I suggest them to be built-in features.

I especially am in favor of number 2. I had to write a script to move all windows on a workspace to another one so that i could launch something (that spawns multiple windows) and have them all moved to a special workspace. Otherwise, with just default hyprctl dispatchers you would have to manually move each window to the special workspace.

vaxerski:
point 2 done in 870471dd96724474869d2e648b623c17df12fb57

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #3164, opened 2023-09-03
Set limit on how many windows in special workspace

LamprosPitsillos:
Fixing point 1 of #2596

Added config option `misc:isolate_special INT` 
For windows in special workspace > INT the new window is send to the active workspace (behind special)

Possible bug : Assign negative number to config 

Tested on a nested Hyprland Session and works as expected.

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -417,6 +417,8 @@ void CWindow::onUnmap() {
 
 void CWindow::onMap() {
 
+    static auto* const PISOLATESPECIAL = &g_pConfigManager->getConfigValuePtr("misc:isolate_special")->intValue;
+
     m_pWLSurface.assign(g_pXWaylandManager->getWindowSurface(this));
 
     // JIC, reset the callbacks. If any are set, we'll make sure they are cleared so we don't accidentally unset them. (In case a window got remapped)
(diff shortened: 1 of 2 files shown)


MightyPlaza:
probably better to implement and use `getVisibleWindowsOnWorkspace()`
also this as a workspace rule would be more flexible (not just for special workspaces)

LamprosPitsillos:
> probably better to implement and use getVisibleWindowsOnWorkspace()

Can you give some more details ? Isn't this just 
```cpp
 g_pCompositor->getWindowsOnWorkspace(m_iWorkspaceID)
```
> also this as a workspace rule would be more flexible (not just for special workspaces

I sort of agree with you on this but on the other hand , what would happen to the windows that **dont fit** on the workspace any more , that would also need configuring  ( direction maybe) .
I think this feature is more suited for special workspaces.

Another concern is that this isn't as simple to do anymore 
>> Future PR if all goes well:
Toggle workspace when isolate limit reached. ( ex. isolate_special=1 Open file manager, open file toggles automatically)

MightyPlaza:
> > probably better to implement and use getVisibleWindowsOnWorkspace()
> 
> Can you give some more details ? Isn't this just

check if window isn't hidden, things like groups and swallowed windows would count as only 1

LamprosPitsillos:
I do think this is a feature though , for example i have my terminal in a special workspace open mpv with swallowing , i wouldnt want mpv to spawn behind the special. 
Now for groups i would imagine the same usability ( haven't used groups yet)
I think of it as , If i explicitly went out of my way to group/swallow windows , i clearly want them together.

MightyPlaza:
that's what I meant, the current behavior does the opposite

LamprosPitsillos:
> that's what I meant, the current behavior does the opposite

 Oh, you are right , i was testing my setup with `isolate_special=2`
 
 Will try tomorrow to create the `g_pCompositor->getVisibleWindowsOnWorkspace()` method

MightyPlaza:
there isn't any, you have to copy the `getWindowsOnWorkspace()` and check if window isn't hidden as well to create it

MightyPlaza:
Did the workspace rule implementation in https://github.com/LamprosPitsillos/Hyprland/pull/1
Also did the VisibleWindows on Workspace but that doesn't work since window gets inside group later on
Will see how that can be handled

LamprosPitsillos:
>  but that doesn't work since window gets inside group later on Will see how that can be handled
This is what i noticed last night, and was thinking of ways to solve it.

LamprosPitsillos:
Great ! Only one small request , can "maxsize" be changed to "maxclients" ?

MightyPlaza:
implementation is kinda meh but should be fully working after https://github.com/LamprosPitsillos/Hyprland/pull/2

PR #3180, opened 2023-09-05
feat: add renameWorkspace event

cyrinux:
This should help dev of new waybar hyprland module https://github.com/Alexays/Waybar/issues/2271

And be able to make it receive and display new name generate by `hyprland-autoname-workspace` since its broke with the removale of wlr workspace api https://github.com/hyprwm/Hyprland/commit/bb0933437f4602080a66cc877fee36dce8dcb3ff.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1430,8 +1430,10 @@ void CKeybindManager::renameWorkspace(std::string args) {
             int         workspace = std::stoi(args.substr(0, FIRSTSPACEPOS));
             std::string name      = args.substr(FIRSTSPACEPOS + 1);
             g_pCompositor->renameWorkspace(workspace, name);
+            EMIT_HOOK_EVENT("rename", (std::vector<void*>{workspace, name}));
         } else {
             g_pCompositor->renameWorkspace(std::stoi(args), "");
+            EMIT_HOOK_EVENT("rename", (std::vector<void*>{std::stoi(args), ""}));
         }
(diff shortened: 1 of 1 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -1430,8 +1430,10 @@ void CKeybindManager::renameWorkspace(std::string args) {
             int         workspace = std::stoi(args.substr(0, FIRSTSPACEPOS));
             std::string name      = args.substr(FIRSTSPACEPOS + 1);
             g_pCompositor->renameWorkspace(workspace, name);
+            EMIT_HOOK_EVENT("rename", (std::vector<void*>{workspace, name}));
```
vaxerski, line 1433: first, instead of duping this, it'd be better to put this inside renameWorkspace.

second, `rename` is a terrible name. Rename what? `renameWorkspace` is a better one.

vaxerski (changes requested):

src/managers/KeybindManager.cpp
cyrinux: I will modify this, thanks

vaxerski:
what the hell happened here? I asked for it to be moved in CCompositor::renameWorkspace??

Besides, `EMIT_HOOK_EVENT` is for plugins. If you're talking about waybar, it should be done via socket2

I'll just do this myself

vaxerski:
2d100bf57e0c0184d307bd4173b1a8d9fc7130e1

PR #3181, opened 2023-09-05
Fix blur for passes set to 0.

Dickby:

we can get rid of the first blur pass outside of the loops, 1 << 1 is 2, so if passes is set to anything > 0 the the behaviour wont be changed.
If blur.passes is set to zero and blur is enabled, before this change it did produce a 1/4 size image of the background.
After this change, if passes is zero, no blur is applied, but brightness contrast and noice still work as expected.

I think it's fine

I think it's fine

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -975,13 +975,10 @@ CFramebuffer* CHyprOpenGLImpl::blurMainFramebufferWithDamage(float a, CRegion* o
 
     // damage region will be scaled, make a temp
     CRegion tempDamage{damage};
-    wlr_region_scale(tempDamage.pixman(), damage.pixman(), 1.f / 2.f); // when DOWNscaling, we make the region twice as small because it's the TARGET
-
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
I do believe we should just outright deny blurring if passes are at <= 0

Dickby:
But one might want to have noise without blur.
Anyways, the changes can be used even if passes=0 is not possible, in that case it only slims down the code a bit.

vaxerski:
hm, wait, with this scenario, blurring is just skipped right? never mind then.

Right, people might want the effects.

Maybe it's worth documenting that?

Dickby:
Yes, blur will be skipped.
Not shure how to document that.
It won't break any existing config, and set blur.passes to zero should be quite intuitive.

Dickby:
I mean the .blur section of decoration could be renamed to something else since it now works for other effects without blur, but i wouldn't say its worth breaking existing configs

vaxerski (approved):
I'll allow this, idk how to document this either. Maybe just a note in passes? (sumn like "setting passes to 0 allows for disabling blur but preserving blur effects like noise or contrast")

PR #3184, opened 2023-09-05
Improve fallback behavior for unset device config values

pdamianik:

This should improve the fallback process for device specific, unset config values. It replaces the touchpad bool parameter for `getDevice...` functions with an optional fallback configuration key that describes the global configuration key to fall back to if the device specific key is unset. The previous implementation didn't quite work for me, as it would reset `natural_scroll` to disabled even when just enabling/disabling my touchpad. This should also handle conflicts between the fallback configs `input:touchdevice:transform` and `input:tablet:transform` and `input:touchdevice:output` and `input:tablet:output` better as the current implementation has a hardcoded order which prefers `touchdevice:*` keys over `tablet:*` keys, even for tablet devices.

#### Alternative solution

An alternative solution would be to use the same key structure as the normal keys for device config keys. This could improve intuitiveness for users (the keys aren't just the last section and potentially conflicting) and it would make the extra fallback argument unnecessary.

`CConfigManager::getConfigValueSafeDevice` is overloaded and spreads to `CConfigManager::getDevice...`, so I originally tried to use `std::optional` (I have very little C++ experience). Is there a better way to describe an optional string reference?

If it looks good and the alternative solution is not something to be implemented, it's ready :smiley: .

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1685,7 +1685,7 @@ SConfigValue CConfigManager::getConfigValueSafe(const std::string& val) {
     return copy;
 }
 
-SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, const std::string& val) {
+SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, const std::string& val, const std::string& fallback = "") {
     std::lock_guard<std::mutex> lg(configmtx);
 
     const auto                  it = deviceConfigs.find(dev);
@@ -1695,18 +1695,14 @@ SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, co
         return SConfigValue();
     }
 
-    return it->second[val];
-}
-
-SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, const std::string& val, const std::string& fallback) {
-    auto device_config = getConfigValueSafeDevice(dev, val);
+    const SConfigValue deviceConfig = it->second[val];
 
     // fallback if not set explicitly
-    if (!device_config.set) {
+    if (!deviceConfig.set && !fallback.empty()) {
         return configValues[fallback];
     }
 
-    return device_config;
+    return deviceConfig;
 }
 
 int CConfigManager::getInt(const std::string& v) {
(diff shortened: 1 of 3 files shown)


src/config/ConfigManager.cpp
```diff
@@ -1691,23 +1691,18 @@ SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, co
...
+}
 
-    // fallback if not set explicitly
-    if (!copy.set) {
-        for (auto& cv : configValues) {
-            auto foundIt = cv.first.find(val);
-            if (foundIt == std::string::npos)
-                continue;
+SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, const std::string& val, const std::string& fallback) {
+    auto device_config = getConfigValueSafeDevice(dev, val);
```
vaxerski, line 1698: style: no snake

src/config/ConfigManager.hpp
```diff
@@ -156,9 +156,12 @@ class CConfigManager {
     void                                                            setInt(const std::string&, int);
     void                                                            setString(const std::string&, const std::string&);
 
-    int                                                             getDeviceInt(const std::string&, const std::string&, std::optional<bool> touchpad = {});
-    float                                                           getDeviceFloat(const std::string&, const std::string&, std::optional<bool> touchpad = {});
-    std::string                                                     getDeviceString(const std::string&, const std::string&, std::optional<bool> touchpad = {});
+    int                                                             getDeviceInt(const std::string&, const std::string&);
```
vaxerski, line 159: makes no sense to dupe those. Just make fallback be default to "" and if it's `.empty()` then behave like it wasnt specified

src/config/ConfigManager.hpp
```diff
@@ -255,7 +258,8 @@ class CConfigManager {
     void         applyUserDefinedVars(std::string&, const size_t);
     void         loadConfigLoadVars();
     SConfigValue getConfigValueSafe(const std::string&);
-    SConfigValue getConfigValueSafeDevice(const std::string&, const std::string&, std::optional<bool> touchpad);
+    SConfigValue getConfigValueSafeDevice(const std::string&, const std::string&);
+    SConfigValue getConfigValueSafeDevice(const std::string&, const std::string&, const std::string& fallback);
```
vaxerski, line 262: ditto

vaxerski (commented):
oh, and conflict

pdamianik:
Fixed, what about using the same key structure for device configs as the global config?

vaxerski:
not sure what you mean by that

pdamianik:
Nevermind, pr is ready from my side.

vaxerski:
I don't think this will compile xD

vaxerski:
also #3189 got merged in the meantime so if you could rebase against master and fix that up

vaxerski:
told you it wont compile :P

header:
`const std::string& param = ""`
source:
`const std::string& param`

you can't have `= ""` in the source

pdamianik:
Fixed, compiles, starts (tested for the latest changes this time), fixes my touchpad issue and ready for merging.

src/managers/input/InputManager.cpp
```diff
@@ -1391,12 +1400,13 @@ void CInputManager::setTouchDeviceConfigs(STouchDevice* dev) {
         if (wlr_input_device_is_libinput(PTOUCHDEV->pWlrDevice)) {
             const auto LIBINPUTDEV = (libinput_device*)wlr_libinput_get_device_handle(PTOUCHDEV->pWlrDevice);
 
-            const int  ROTATION =
-                std::clamp(HASCONFIG ? g_pConfigManager->getDeviceInt(PTOUCHDEV->name, "transform") : g_pConfigManager->getInt("input:touchdevice:transform"), 0, 7);
+            const int  ROTATION = std::clamp(
+                HASCONFIG ? g_pConfigManager->getDeviceInt(PTOUCHDEV->name, "transform", "input:touchdevice") : g_pConfigManager->getInt("input:touchdevice:transform"), 0, 7);
```
vaxerski, line 1404: oopsie

src/managers/input/InputManager.cpp
```diff
@@ -1423,11 +1433,12 @@ void CInputManager::setTabletConfigs() {
         if (wlr_input_device_is_libinput(t.wlrDevice)) {
             const auto LIBINPUTDEV = (libinput_device*)wlr_libinput_get_device_handle(t.wlrDevice);
 
-            const int  ROTATION = std::clamp(HASCONFIG ? g_pConfigManager->getDeviceInt(t.name, "transform") : g_pConfigManager->getInt("input:tablet:transform"), 0, 7);
+            const int  ROTATION =
+                std::clamp(HASCONFIG ? g_pConfigManager->getDeviceInt(t.name, "transform", "input:tablet:tranform") : g_pConfigManager->getInt("input:tablet:transform"), 0, 7);
```
vaxerski, line 1437: oopsie

src/config/ConfigManager.cpp
```diff
@@ -1695,23 +1695,14 @@ SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, co
         return SConfigValue();
     }
 
-    SConfigValue copy = it->second[val];
+    const SConfigValue deviceConfig = it->second[val];
```
vaxerski, line 1698: style nit: const values are `CAPS` so `DEVICECONFIG`

vaxerski (commented):
some typos.

As a next step (for the future) since we have fallbacks, we could drop the HASCONFIG and move it to CConfigManager::getConfigValueSafeDevice

pdamianik:
> some typos.
> 
> As a next step (for the future) since we have fallbacks, we could drop the HASCONFIG and move it to CConfigManager::getConfigValueSafeDevice

I'll make another branch ready.

vaxerski (approved):
h

vaxerski:
ya

PR #3195, opened 2023-09-06
Integrate HASCONFIG into CConfigManager::getConfigValueSafeDevice

pdamianik:

With the new fallback argument for `CConfigManager::getConfigValueSafeDevice`, the `HASCONFIG` constant has become unnecessary in most cases. This PR adds the ability to fall back when there is no config for a device to `CConfigManager::getConfigValueSafeDevice` and removes most uses of `HASCONFIG`.

All ready.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1691,13 +1691,15 @@ SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, co
     const auto                  it = deviceConfigs.find(dev);
 
     if (it == deviceConfigs.end()) {
-        Debug::log(ERR, "getConfigValueSafeDevice: No device config for {} found???", dev.c_str());
-        return SConfigValue();
+        if (fallback.empty()) {
+            Debug::log(ERR, "getConfigValueSafeDevice: No device config for {} found???", dev.c_str());
+            return SConfigValue();
+        }
+        return configValues[fallback];
     }
 
     const SConfigValue DEVICECONFIG = it->second[val];
 
-    // fallback if not set explicitly
     if (!DEVICECONFIG.set && !fallback.empty()) {
         return configValues[fallback];
     }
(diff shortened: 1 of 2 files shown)


src/config/ConfigManager.cpp
```diff
@@ -1691,13 +1691,16 @@ SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, co
     const auto                  it = deviceConfigs.find(dev);
 
     if (it == deviceConfigs.end()) {
-        Debug::log(ERR, "getConfigValueSafeDevice: No device config for {} found???", dev.c_str());
-        return SConfigValue();
+        if (fallback.empty()) {
+            Debug::log(ERR, "getConfigValueSafeDevice: No device config for {} found???", dev.c_str());
```
vaxerski, line 1695: no need for .c_str() since format mr

src/config/ConfigManager.cpp
```diff
@@ -1691,13 +1691,16 @@ SConfigValue CConfigManager::getConfigValueSafeDevice(const std::string& dev, co
...
+            return SConfigValue();
+        } else {
```
vaxerski, line 1697: else statement redundant
```
if (something)  {
 // ...
 return xyz;
}

return xyz2; // works like else, because if above returns
```

vaxerski (commented):
some nits

src/config/ConfigManager.cpp
pdamianik: A lot of the formats in https://github.com/hyprwm/Hyprland/commit/61a71c65ace5124cb26a074715280537e6e34d00 use `.c_str()` so if that isn't the proper way to format strings it should be changed in a single commit.

vaxerski: they do that cuz I was lazy

pdamianik: I could write another pr for all of those, but I wouldn't put that in this pr. Just to keep it consistent.

vaxerski: ye ok

vaxerski (approved):
yummy

src/config/ConfigManager.cpp
pdamianik: https://github.com/hyprwm/Hyprland/pull/3198#issue-1884627024

PR #3216, opened 2023-09-08
fix: focusWindow on hidden workspace triggers another focusWindow

memchr:
## Abstract
This commit address an issue where focusing a window on a hidden
workspace inadvertently triggered a second `focusWindow` call due to
simulated mouse movement. This behaviour led to the incorrect focus on
the window under the cursor instead of target window of method
`focusWindow()`, disrupting `focusurgentorlast` and `focuscurrentorlast`
dispatchers. 

## Implementation
Introduced a flag to the `CMonitor::changeWorkspace()`
method to prevent simulated mouse movements. This flag is set to false
by default. Changed the `focusWindow()` method accordingly to set this
flag to true when the target window is in a hidden workspace.


vaxerski (approved):
k

PR #3217, opened 2023-09-08
implement pushactivetobottom dispacher

alaricljs:
Adds pushactivetobottom dispacher to manage Z-order of floating windows, opposite of bringactivetotop.
Makes both bringactivetotop and pushactivetobottom available to mouse binds.

I had issues with Compositor.cpp:1255 - making this generic with less code via:

```
auto begin = top ? m_vWindows.begin() : (m_vWindows.rbegin() + 1).base();
auto end = top ? m_vWindows.end() : m_vWindows.rend();
```

Never worked for `top=false` regardless of variations to m_vWindows.rbegin().  Being new to C++ I don't know if it's  because of the uses of begin & end inside the algo, or just my botching the cast of the reverse iterator.

In KeybindManager.cpp:1983 I kludged in mouse re-focus behavior without checking follow_mouse or mouse_refocus.  Is this the right place for it?  It doesn't refocus without the simulateMouseMovement() call, but with that call it doesn't obey those 2 options which I would have thought handled in InputManager.cpp around ln 400.

Need to resolve the above with your input, thanks.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1245,14 +1245,27 @@ bool CCompositor::isWindowActive(CWindow* pWindow) {
 }
 
 void CCompositor::moveWindowToTop(CWindow* pWindow) {
+    changeWindowZOrder(pWindow, true);
+}
+
+void CCompositor::changeWindowZOrder(CWindow* pWindow, bool top) {
     if (!windowValidMapped(pWindow))
         return;
 
-    auto moveToTop = [&](CWindow* pw) -> void {
-        for (auto it = m_vWindows.begin(); it != m_vWindows.end(); ++it) {
-            if (it->get() == pw) {
-                std::rotate(it, it + 1, m_vWindows.end());
-                break;
+    auto moveToZ = [&](CWindow* pw, bool top) -> void {
+        if (top) {
+            for (auto it = m_vWindows.begin(); it != m_vWindows.end(); ++it) {
+                if (it->get() == pw) {
+                    std::rotate(it, it + 1, m_vWindows.end());
+                    break;
+                }
+            }
+        } else {
+            for (auto it = m_vWindows.rbegin(); it != m_vWindows.rend(); ++it) {
+                if (it->get() == pw) {
+                    std::rotate(it, it + 1, m_vWindows.rend());
+                    break;
+                }
             }
         }
 
@@ -1260,9 +1273,10 @@ void CCompositor::moveWindowToTop(CWindow* pWindow) {
             g_pHyprRenderer->damageMonitor(getMonitorFromID(pw->m_iMonitorID));
     };
 
-    moveToTop(pWindow);
+    moveToZ(pWindow, top);
 
-    pWindow->m_bCreatedOverFullscreen = true;
+    if (top)
+        pWindow->m_bCreatedOverFullscreen = true;
 
     if (!pWindow->m_bIsX11)
         return;
@@ -1278,9 +1292,9 @@ void CCompositor::moveWindowToTop(CWindow* pWindow) {
     }
 
     for (auto& pw : toMove) {
-        moveToTop(pw);
+        moveToZ(pw, top);
 
-        moveWindowToTop(pw);
+        changeWindowZOrder(pw, top);
     }
 }
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1288,9 +1288,9 @@ void CCompositor::changeWindowZOrder(CWindow* pWindow, bool top) {
     }
 
     for (auto& pw : toMove) {
-        moveToZ(pw, top);
+        moveToZ(pw, true);
 
-        changeWindowZOrder(pw, top);
+        changeWindowZOrder(pw, true);
     }
 }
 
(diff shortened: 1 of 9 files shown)


vaxerski (commented):
hm. Maybe we could unify both into one dispatcher?

Like, "alterzorder" with either "+x" "-x" or "top" "bottom"

We'd deprecate bringactivetotop by that, too.

alaricljs:
I am a fan of good unified design and deprecating the outdated.  One thing though, I can't think of a reason for +/- N changes, even when it was available I only ever used Top/Bottom.  No issue with implementing it if you want.

vaxerski:
maybe we can put it off then for now, and if someone requests it and it's desired, then add it.

alaricljs:
Implemented alterzorder which allows for a window filter but defaults to the active window.  Also removed my adds to the mouse bind.  I managed to miss that short blurb that bind can handle mouse buttons for clicks.

src/managers/KeybindManager.cpp
```diff
@@ -1980,12 +1972,29 @@ void CKeybindManager::bringActiveToTop(std::string args) {
...
+    if (POSITION == "top")
+        g_pCompositor->changeWindowZOrder(PWINDOW, 1);
+    else if (POSITION == "bottom")
+        g_pCompositor->changeWindowZOrder(PWINDOW, 0);
+    else {
+        Debug::log(ERR, "alterZOrder: bad position: %s", POSITION);
+        return;
     }
+
+    // if follow_mouse && mouse_refocus ?
```
vaxerski, line 1996: nope, this is to unstuck the cursor. comment can be removed

src/managers/KeybindManager.cpp
```diff
@@ -1980,12 +1972,29 @@ void CKeybindManager::bringActiveToTop(std::string args) {
...
-    if (g_pCompositor->m_pLastWindow && g_pCompositor->m_pLastWindow->m_bIsFloating) {
-        g_pCompositor->changeWindowZOrder(g_pCompositor->m_pLastWindow, 0);
-        // if follow_mouse && mouse_refocus ?
-        g_pInputManager->simulateMouseMovement();
+void CKeybindManager::alterZOrder(std::string args) {
+    const auto WINDOWREGEX = args.substr(args.find_first_of(',') + 1);
+    const auto POSITION    = args.substr(0, args.find_first_of(','));
+    auto       PWINDOW     = g_pCompositor->getWindowByRegex(WINDOWREGEX);
+
+    if (!PWINDOW && g_pCompositor->m_pLastWindow && g_pCompositor->m_pLastWindow->m_bIsFloating)
```
vaxerski, line 1980: this will fail if provided with a window, and crash if m_pLastWindow is null.

```cpp
if (!PWINDOW && g_pCompositor->m_pLastWindow && g_pCompositor->m_pLastWindow->m_bIsFloating)
    PWINDOW = g_pCompositor->m_pLastWindow;

if (!PWINDOW) {
    Debug::log(ERR, "alterZOrder: no window");
    return;
}
```

src/Compositor.cpp
```diff
@@ -1278,9 +1292,9 @@ void CCompositor::moveWindowToTop(CWindow* pWindow) {
     }
 
     for (auto& pw : toMove) {
-        moveToTop(pw);
+        moveToZ(pw, top);
 
-        moveWindowToTop(pw);
+        changeWindowZOrder(pw, top);
```
vaxerski, line 1297: both of these should be true, not top

src/Compositor.cpp
```diff
@@ -1245,24 +1245,38 @@ bool CCompositor::isWindowActive(CWindow* pWindow) {
 }
 
 void CCompositor::moveWindowToTop(CWindow* pWindow) {
+    changeWindowZOrder(pWindow, true);
```
vaxerski, line 1248: why not rid of this? internal func, we can just yeet it and replace the refs with changewindowzorder

src/Compositor.cpp
alaricljs: Nope, it's uglier than that... 
For the X11 windows I'll need to figure out what order toMove has the windows in and then move them to the bottom Z from the top of the stack down.   I'm thinking the original is just doing a 'top' form the bottom of the stack up.  It'll take me more than the time I have tonight to fix that and test it.

vaxerski: parent -> child -> child


you move the parent to the bottom, but children should still be preserved on top of it.

You're right it shouldn't be `true`, my bad. it should be children from the bottom first to the bottom, parent last, lol

alaricljs:
It's solved... but it's ugly.  I'll think on it for a bit.

alaricljs:
Ok, that's less ugly

alaricljs:
https://github.com/hyprwm/hyprland-wiki/pull/337

vaxerski:
my bad

MightyPlaza:
> I am a fan of good unified design and deprecating the outdated. One thing though, I can't think of a reason for +/- N changes, even when it was available I only ever used Top/Bottom. No issue with implementing it if you want.

when having grouped or other hidden windows it would be best to have to option to swap/set "the same" z order when changing the visible window
otherwise each window will have it's own z order or changing visible window will require moving it to the top

Issue #3208, opened 2023-09-07
Pass bind not releasing


AaronMillward:
### Hyprland Version

Hyprland, built from branch  at commit 96d555e8e794627bfc561e294e148ab8a9961fcc dirty (). Tag: v0.29.1  flags: (if any)

### Bug or Regression?

Regression

### Description

I use a mouse button as my push to talk key in discord. after upgrading from 0.28 to 0.29.1 it seems like only the initial press of the button is sent to discord, my PTT stays on even after releasing the button. if discord is focused the press and release will work correctly unless the button was pressed outside of discord first in which case it still keeps the PTT active.

Testing with `bind=,mouse:276,pass,^(discord)$` on `Hyprland, built from branch  at commit 9654749244117f7f150c6f2a2ce4dede6e8cbb25 dirty ().  v0.28.0 (if any)` I get this `wev` output when pressing the button
```
[13:      wl_pointer] leave: surface: 3
[13:      wl_pointer] frame
[14:     wl_keyboard] modifiers: serial: 0; group: 0
                      depressed: 00000000
                      latched: 00000000
                      locked: 00000000
[14:     wl_keyboard] modifiers: serial: 0; group: 0
                      depressed: 00000000
                      latched: 00000000
                      locked: 00000000
```

Then in `Hyprland, built from branch  at commit 96d555e8e794627bfc561e294e148ab8a9961fcc dirty (). Tag: v0.29.1 flags: (if any)` this is the ouput
```
[13:      wl_pointer] leave: surface: 3
[13:      wl_pointer] frame
[14:     wl_keyboard] modifiers: serial: 0; group: 0
                      depressed: 00000000
                      latched: 00000000
                      locked: 00000000
```

The same happens when using an actual keyboard key `o` in this case. ``bind=,o,pass,^(discord)$``
0.28
```
[14:     wl_keyboard] leave: serial: 244; surface: 3
[14:     wl_keyboard] enter: serial: 249; surface: 3
                      sym: o            (111), utf8: 'o'
[14:     wl_keyboard] modifiers: serial: 0; group: 0
                      depressed: 00000000
                      latched: 00000000
                      locked: 00000000
[11:  wl_data_device] selection: (cleared)
[14:     wl_keyboard] leave: serial: 251; surface: 3
[14:     wl_keyboard] enter: serial: 256; surface: 3
[14:     wl_keyboard] modifiers: serial: 0; group: 0
                      depressed: 00000000
                      latched: 00000000
[...]

PR #3219, opened 2023-09-09
Do not suppress pass and mouse release

shadowmax31:
This fixes the release issue mentionned here: https://github.com/hyprwm/Hyprland/issues/3208

I don't know if it I needed to add `k.handler != "mouse"` too, but I figured it might be needed since the commit that broke this removed a check like this https://github.com/hyprwm/Hyprland/commit/f242f94

Should be ready to merge


vaxerski (approved):
right thanks

PR #3229, opened 2023-09-09
refactor: replace lambdas with CKeybindManager::switchToWindow

### LINKED ISSUE #3227: `focuscurrentorlast`: scroll affects previous window
### Hyprland Version

91e28bbe9df85e2e94fbcc0137106362aea14ab5

### Bug or Regression?

Bug

### Description

```
$altMod = ALT
bind = $altMod, Tab, focuscurrentorlast
```
If I press Alttab and don't move the mouse afterwards and try to scroll, it scrolls the previous window and not the current window. I need to move cursor a little (with touchpad or mouse) to make it work as expected. Watch the video below to understand better.
I've looked at the code a bit and I can guess that it can be fixed here: https://github.com/hyprwm/Hyprland/blob/main/src/managers/input/InputManager.cpp#L1256C1-L1276C2

### How to reproduce

https://github.com/hyprwm/Hyprland/assets/122799969/11679e7e-dc00-4191-84d8-b2a21c09fee0

_No response_

memchr:
## Abstract

There are 3 byte to byte identical and one spiritually identical lambdas `switchToWindow` in `KeybindManager.cpp`. They will surely come back to haunt us one day as bugs often breed in repetitive code.

## Changes

- moves `switchToWindow` lambdas to a single private method `CKeybindManager::switchToWindow()`.

- fixes #3227, as a by-product of the mouse motion simulation at the end of `CKeybindManager::switchToWindow()`.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -214,6 +214,41 @@ bool CKeybindManager::tryMoveFocusToMonitor(CMonitor* monitor) {
     return true;
 }
 
+void CKeybindManager::switchToWindow(CWindow* PWINDOWTOCHANGETO) {
+    const auto PLASTWINDOW = g_pCompositor->m_pLastWindow;
+
+    if (PWINDOWTOCHANGETO == PLASTWINDOW || !PWINDOWTOCHANGETO)
+        return;
+
+    if (PLASTWINDOW && PLASTWINDOW->m_iWorkspaceID == PWINDOWTOCHANGETO->m_iWorkspaceID && PLASTWINDOW->m_bIsFullscreen) {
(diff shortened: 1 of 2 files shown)


memchr:
## Other notes

There is currently a bug similar to #3227 in `listener_mapWindow()`, where scrolling affects the previous window instead of the mapped window if the mapped window was focused. `simulateMouseMovement()` may not be sufficient to fix this bug as there is an edge case: If the cursor is exactly in the middle of the borders of the previous window and the mapped window, the previous window will get the focus when `simulateMouseMovement()` is called.

So that bit remains untouched, along with the fact that this patch is primarily a refactor, and the fix for #3227 is a side effect.

woojiq:
I can confirm that this solves my problem.
@memchr thanks for the quick response :heart:

vaxerski (approved):
thanks!

Issue #829, opened 2022-10-11
Rotation using external app (kanshi)


etrnal70:
Hello, thank you for this really cool project ! Is rotation using external app support planned in the future ? My usecase is that i'm using kanshi for automatic screen configuration for my home and work profile. It's literally the only bug that stopping me from using hyprland daily 😄

vaxerski:
it's maybe planned? never thought of this really, as you can set the transforms in the config.

vaxerski:
should be done in c47581fc5aedcff172b76b658593a1e7c712e448

etrnal70:
Still not working, tried using both kanshi and wdisplays. Currently on latest commit `b4c45aa2e3`

pdamianik:
Should become possible with #3234 (not tested)

PR #3234, opened 2023-09-09
Fix wlr_output_management_unstable_v1 protocol

### LINKED ISSUE #3220: wlr_output_management_v1 protocol not implemented properly
### Hyprland Version

v0.27.0

### Bug or Regression?

Bug

### Description

I am trying to update [rot8](https://github.com/efernau/rot8) to support Hyprland, by making it use the wlroots output_management protocol to apply output rotations. I have it working in Sway, but in Hyprland I'm seeing some unusual behavior.

I can find the display and apply a new configuration with the transform modified, but the output doesn't actually rotate. I get an OutputConfiguration::Success event, but no new configuration serial is advertised via OutputManager::Done event. On future rotation attempts, the configuration change is cancelled, likely because an incorrect serial number was provided. Also, new transforms aren't reported via the OutputHead::Transform event when rotating an output using `hyprctl keyword monitor eDP-1,preferred,auto,1,transform,1`.

I could just add a hyprland-specific backend which uses hyprctl to communicate, but would love to get to a single implementation for all wlroots compositors.

### How to reproduce

Clone [my branch](https://github.com/Smona/rot8/tree/wlroots-support) and `cargo run` with rustc and cargo installed, both on sway and hyprland.

_No response_

pdamianik:

Fixes #3220 and #829. The issue is split into two separate bugs:

1.  The `transform` property does not get applied because the function `CHyprRenderer::outputMgrApplyTest` checks for a difference between `OUTPUT->transform` and `PMONITOR->transform` and only applies the new property when they differ. This is a mistake since these properties both represent the active configration (they are the same and don't change until the new rules are applied).
2. Hyprland does not send notifications for monitor configuration changes via the [`head`](https://wayland.app/protocols/wlr-output-management-unstable-v1#zwlr_output_manager_v1:event:head) event. The function `wlr_output_manager_v1_set_configuration` internally sends this event and gets called from `Events::listener_change` which should get called when the properties of a monitor get changed. Instead it only got called in `CMonitor::onConnect()` when a new monitor got added. Additionally the configuration properties sent by `Events::listener_change` were missing the `custom_mode`, `transform` and `adaptive_sync_enabled` property.

Not including the `transform` property in `CHyprRenderer::outputMgrApplyTest` is not possible since the other properties set in `commandForCfg` variable reset the monitor rule for the applicable monitor.

Ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -580,12 +580,12 @@ void CConfigManager::handleMonitor(const std::string& command, const std::string
                 return;
             }
 
-            wl_output_transform transform = (wl_output_transform)std::stoi(ARGS[2]);
+            const auto TRANSFORM = (wl_output_transform)TSF;
 
             // overwrite if exists
             for (auto& r : m_dMonitorRules) {
                 if (r.name == newrule.name) {
-                    r.transform = transform;
+                    r.transform = TRANSFORM;
                     return;
                 }
             }
--- a/src/events/Monitors.cpp
+++ b/src/events/Monitors.cpp
@@ -38,8 +38,18 @@ void Events::listener_change(wl_listener* listener, void* data) {
 
         CONFIGHEAD->state.enabled = m->output->enabled;
         CONFIGHEAD->state.mode    = m->output->current_mode;
-        CONFIGHEAD->state.x       = m->vecPosition.x;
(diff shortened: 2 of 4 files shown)


vaxerski (commented):
everything looks alright, but please slap it with a `clang-format`.

vaxerski (approved):
thanks!

PR #3235, opened 2023-09-09
fix: add missing groupbar decoration to the window

memchr:
Add groupbar decoration to the operand window of `CWindow::insertWindowToGroup` if it does not exist, to prevent segmentation faults when mouse events are triggered after moving the window to a group, where `getDecorationByType(DECORATION_GROUPBAR)` unexpectedly returns nullptr.

Also fixed a bug where the group bar disappeared when the moveIntoGroup dispatcher's operand window was in a group.

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -733,9 +733,8 @@ void CWindow::insertWindowToGroup(CWindow* pWindow) {
     const auto BEGINAT = this;
     const auto ENDAT   = m_sGroupData.pNextWindow;
 
-    if (!pWindow->getDecorationByType(DECORATION_GROUPBAR)) {
+    if (!pWindow->getDecorationByType(DECORATION_GROUPBAR))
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
style nit: no {} around short ifs

memchr:
Why doesn't clang-format do this? IIRC it has a ruleset for brace around short statements.

vaxerski:
havent seen one

vaxerski (approved):
top g

memchr:
> havent seen one

```yaml
RemoveBracesLLVM: true
```

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#removebracesllvm

vaxerski:
it's not what we want tho, as we keep braces around `for` `while` etc

memchr:
That's a very specific style configuration. I don't think clang-format supports this yet.

People tend to forget specific styles that are not enforced by formatters, especially when working with several different styles.

vaxerski:
true, that's why I remind them. It's not a huge deal anyways

PR #3247, opened 2023-09-10
add on-screen constraint to wrv2 'move'

alaricljs:
Implements an `onscreen` directive to windowsrulev2 `move` that causes the window to be fully on the screen (if it is not too large). 

My usage for this is:
`windowrulev2 = move onscreen cursor -50% -50%,floating:1,class:^(xfce4-terminal|kitty)$`
This will place my terminals centered under the mouse unless it's toward the edge of the screen where they will be nudged to remain fully within the screen.

For too-large windows, the bottom right corner is forced onto the screen.  It could be top left with additional code.

Ready for merge.

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -377,10 +377,7 @@ void Events::listener_mapWindow(void* owner, void* data) {
                     }
 
                     if (ONSCREEN) {
-                        static auto* const PBORDERSIZE = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
-                        int borderSize = PWINDOW->m_sSpecialRenderData.borderSize.toUnderlying() == -1 ? *PBORDERSIZE : PWINDOW->m_sSpecialRenderData.borderSize.toUnderlying();
-                        if (PWINDOW->m_sAdditionalConfigData.borderSize.toUnderlying() != -1)
-                            borderSize = PWINDOW->m_sAdditionalConfigData.borderSize.toUnderlying();
+                        int borderSize = PWINDOW->getRealBorderSize();
 
                         // left
                         if (posX < PMONITOR->vecReservedTopLeft.x + borderSize)
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -322,6 +322,11 @@ void Events::listener_mapWindow(void* owner, void* data) {
                 try {
                     auto       value = r.szRule.substr(r.szRule.find(' ') + 1);
 
+                    const bool ONSCREEN = value.find("onscreen") == 0;
+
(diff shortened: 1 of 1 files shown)


src/events/Windows.cpp
```diff
@@ -373,6 +376,27 @@ void Events::listener_mapWindow(void* owner, void* data) {
                         }
                     }
 
+                    if (ONSCREEN) {
+                        static auto* const PBORDERSIZE = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
+                        int borderSize = PWINDOW->m_sSpecialRenderData.borderSize.toUnderlying() == -1 ? *PBORDERSIZE : PWINDOW->m_sSpecialRenderData.borderSize.toUnderlying();
```
vaxerski, line 381: `CWindow::getRealBorderSize()`

src/events/Windows.cpp
```diff
@@ -373,6 +376,24 @@ void Events::listener_mapWindow(void* owner, void* data) {
                         }
                     }
 
+                    if (ONSCREEN) {
+                        int borderSize = PWINDOW->getRealBorderSize();
+
+                        // left
+                        if (posX < PMONITOR->vecReservedTopLeft.x + borderSize)
```
vaxerski, line 383: wait, why not use `std::clamp` ?

alaricljs: I don't expect you to remember, but I did mention I'm new to C++... Guess I'll start looking in the stdlib for everything and the kitchen sink.   Is this appropriate?

```
posX = std::clamp(posX, int(PMONITOR->vecReservedTopLeft.x + borderSize),
    int(PMONITOR->vecSize.x - PMONITOR->vecReservedBottomRight.x - PWINDOW->m_vRealSize.goalv().x - borderSize));

posY = std::clamp(posY, int(PMONITOR->vecReservedTopLeft.y + borderSize),
    int(PMONITOR->vecSize.y - PMONITOR->vecReservedBottomRight.y - PWINDOW->m_vRealSize.goalv().y - borderSize));

```

vaxerski: yep, looks about right.

just `(int)(something)`, because `int(something)` hurts in the eyes.

> I don't expect you to remember, but I did mention I'm new to C++...

No shame in that, I am just giving tips. In the case of you not using `std::clamp` because of something, I asked the question the way I did.

vaxerski (approved):
gangsta

alaricljs:
@vaxerski anything I need to do to get the last check (wlroots buld) to pass so this actually lands?

vaxerski:
this is merged

PR #3250, opened 2023-09-11
refactor: unify changeworkspace dispatcher

memchr:
Refactor the workspace dispatcher to reduce repetitive code. 

Minor logic changes.

### DIFF
--- a/src/helpers/Workspace.cpp
+++ b/src/helpers/Workspace.cpp
@@ -152,7 +152,7 @@ void CWorkspace::rememberPrevWorkspace(const CWorkspace* prev) {
         return;
     }
 
-    if (prev->m_sPrevWorkspace.iID == m_sPrevWorkspace.iID) {
+    if (prev->m_iID == m_iID) {
         Debug::log(LOG, "Tried to set prev workspace to the same as current one");
         return;
     }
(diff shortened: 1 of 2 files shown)


memchr:
I think the following can also be removed as `changeWorkspace()` does the same thing
```cpp
        if (const auto PLASTWINDOW = pWorkspaceToChangeTo->getLastFocusedWindow(); PLASTWINDOW)
            g_pCompositor->focusWindow(PLASTWINDOW);
        else if (const auto PFIRSTWINDOW = g_pCompositor->getFirstWindowOnWorkspace(pWorkspaceToChangeTo->m_iID); PFIRSTWINDOW)
            g_pCompositor->focusWindow(PFIRSTWINDOW);
        else
            g_pCompositor->focusWindow(nullptr);
            ```

memchr:
In addition, the current behaviour of `CMonitor::changeWorkspace()` is to focus on the last window. and the call to `CInputManager::simulateMouseMovement()` at the end of the dispatcher may betray this, so, shall we wrap the cursor to the middle of `pWorkspace->getLastFocusedWindow()` if `input:follow_mouse == 3` ?

vaxerski:
yeah sounds like an okay idea

vaxerski (approved):
lgtm

Issue #3266, opened 2023-09-12
df51c45 effectively breaks waybar workspace scrolling


Ghosthree3:
### Hyprland Version

Hyprland, built from branch HEAD at commit df51c45d7f9e4e280b959f790f7cde6c88b9b451 dirty (keybinds: unify changeworkspace dispatcher (3250)).

### Bug or Regression?

Regression

### Description

With the introduction of commit https://github.com/hyprwm/Hyprland/commit/df51c45d7f9e4e280b959f790f7cde6c88b9b451 it has become impractical to use scrolling to flick through workspaces with waybar as suggested by the wiki here, https://wiki.hyprland.org/hyprland-wiki/pages/Useful-Utilities/Status-Bars/#scrolling-through-workspaces
This is due to the cursor always being teleported to the center of the screen/new window every time the workspace is changed, requiring you to move the mouse back to the status bar between every scroll/workspace change.
This is not a bug if you consider the above wiki suggestion unsupported.

It is somewhat worth noting that workspace scrolling in this way was already somewhat unusable on multi-monitor setups because if e+1 or e-1 involved changing to a workspace on a different monitor this same issue would already occur.
However I considered this somewhat out of scope for Hyprland as the actual issue here is Waybar's current inability to scroll through workspaces present only on the current status bar (ie. the workspaces on the current monitor only) like it does natively under Sway, instead relying on the hacky e+1 e-1 method.

This change is also quite frustrating when switching workspaces by clicking on them, as the cursor is also teleported center, which makes changing your mind and choosing to click another workspace annoying and is just jarring in general.
Perhaps the 'fix' here is just to have some kind of option to disable mouse warping on workspace swapping? I don't _believe_ this is currently an option? Or better yet, a bool supplied to workspace change commands that optionally disables the cursor warping, as you would only really want this behavior under the above mentioned use cases, and I could see disabling cursor warping entirely when swapping workspaces as also being annoying when you use more traditional keybind methods of doing so.

### How to reproduce

Use https://wiki.hyprland.org/hyprland-wiki/pages/Useful-Utilities/Status-Bars/#scrolling-through-workspaces on any build after commit https://github.com/hyprwm/Hyprland/commit/df51c45d7f9e4e280b959f790f7cde6c88b9b451.

tchofy:
[...]

Issue #3205, opened 2023-09-07
Remember focus per-workspace


uoou:
### Description

Current behaviour seems to be that, when switching workspaces, the focused window on the workspace you're switching to matches (as best it can) the focus of the workspace you're switching from.

i.e. if I'm on workspace 1 with three windows open and focus on the first window, then switch to workspace 2, also with three windows open and switch focus to the third window, then switch back to workspace 1 and focus is on the third window.

It'd be nice if workspaces 'remembered' which window is focused. *Or* just focus the first window when switching (I did the latter with keybinds but that breaks monocled workspaces out of monocle mode when switching to them).

Apologies if this is mentioned elsewhere - I did search - or possible in configs - I did try!

uoou:
I just tried a blank config and it definitely doesn't behave that way for me (and never has).

MightyPlaza:
works for me, do you have `follow_mouse` != 1

uoou:
Ahh okay, so it works how I expect with `follow_mouse = 0`. **But**... I also want focus to follow mouse. Just not on workspace switching. i.e. after switching it should only follow the mouse after the mouse moves.

MightyPlaza:
use 2 to change focus when you click the window

uoou:
Yeah that's absolutely fair enough - but not what I want.

MightyPlaza:
having to move mouse isn't a solution either since any slight vibration would change the focused window out of nowhere
warping mouse to center of focused window on workspace change doesn't seem like a good idea as well

uoou:
> any slight vibration would change the focused window

I've literally never had that happen in all my years of using WMs!

You've given me a clue though. What I was trying to do before was adding a binding (on the same keys) after switching workspace to focus the first window. But, as mentioned, this popped monocled workspaces out of monocle.

Instead I could just add a binding which move the mouse to top left (or, ideally centre of first window if I can figure that out) after switching.

uoou:
On digging a bit deeper into the wiki:

```
mouse_refocus

If disabled and follow_mouse=1 then mouse focus will not switch to the hovered window unless the mouse crosses a window boundary.
```

Sounds like exactly what I want. But on setting `mouse_refocus = false` either it's doing nothing or I don't understand what it's doing and it's not what I expected.

andresilva:
After https://github.com/hyprwm/Hyprland/commit/b884544ee63b4c3b3b150ff69cf24c90058c2777 this seems to be fixed for me.

[...]

PR #3268, opened 2023-09-12
feat: add options to skip cursor warps

memchr:
## Abstruct

Adds a config `general:no_cursor_warps_if_focused` which, if true,
allows the cursor to the cursor to stay in place if it is already in the
focused window. This can help reduce unnecessary cursor movement.

Affects all cursor warps.

doc: hyprwm/hyprland-wiki#340

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2290,10 +2290,27 @@ void CCompositor::warpCursorTo(const Vector2D& pos, bool force) {
     // warpCursorTo should only be used for warps that
     // should be disabled with no_cursor_warps
 
-    static auto* const PNOWARPS = &g_pConfigManager->getConfigValuePtr("general:no_cursor_warps")->intValue;
+    static auto* const PNOWARPS        = &g_pConfigManager->getConfigValuePtr("general:no_cursor_warps")->intValue;
+    static auto* const PNOWARPSFOCUSED = &g_pConfigManager->getConfigValuePtr("general:no_cursor_warps_if_focused")->intValue;
+    static auto* const PGAPSIN         = &g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue;
 
-    if (*PNOWARPS && !force)
-        return;
+    if (!force) {
+        if (*PNOWARPS)
+            return;
+
+        auto mouseCoords = g_pInputManager->getMouseCoordsInternal();
+        if (*PNOWARPSFOCUSED) {
+            auto pWindowFocused     = m_pLastWindow;
+            auto pWindowUnderCursor = vectorToWindowIdeal(mouseCoords);
+            if (pWindowFocused && pWindowUnderCursor && pWindowFocused == pWindowUnderCursor && !pWindowFocused->isNearEdge(mouseCoords, *PGAPSIN + 1))
+                return;
+        }
+
+        Vector2D       surfaceCoords;
+        SLayerSurface* pFoundLayerSurface = nullptr;
+        if (vectorToLayerSurface(mouseCoords, &m_pLastMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_TOP], &surfaceCoords, &pFoundLayerSurface))
+            return;
+    }
 
     if (!m_sSeat.mouse)
         return;
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -627,6 +627,16 @@ bool CWindow::isInCurvedCorner(double x, double y) {
     return false;
 }
 
+bool CWindow::isNearEdge(const Vector2D& pos, double threshold) {
+    const auto wbox = getWindowIdealBoundingBoxIgnoreReserved();
+    const auto ibox = wlr_box{wbox.x + threshold, wbox.y + threshold, wbox.width - 2 * threshold, wbox.height - 2 * threshold};
+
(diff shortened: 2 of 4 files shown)


src/Compositor.cpp
```diff
@@ -2212,10 +2212,27 @@ void CCompositor::warpCursorTo(const Vector2D& pos, bool force) {
     // warpCursorTo should only be used for warps that
     // should be disabled with no_cursor_warps
 
-    static auto* const PNOWARPS = &g_pConfigManager->getConfigValuePtr("general:no_cursor_warps")->intValue;
+    static auto* const PNOWARPS        = &g_pConfigManager->getConfigValuePtr("general:no_cursor_warps")->intValue;
+    static auto* const PNOWARPSFOCUSED = &g_pConfigManager->getConfigValuePtr("general:no_cursor_warps_if_focused")->intValue;
+    static auto* const PGAPSIN         = &g_pConfigManager->getConfigValuePtr("general:gaps_in")->intValue;
 
-    if (*PNOWARPS && !force)
```
vaxerski, line 2217: I dont think this belongs here. I'd rather just not warp to anything if monitor stays the same in changeWorkspace for now, and figure out a better method later.

thejch:
imo a better way to deal with #3266 would be to have a config option to not force focus on the "last active window" or warp the cursor at all, and just focus on the window under the cursor (which is just the previous behavior, and honestly nobody was having an issue with it at all)
a lot of people dont like their cursor being warped around when switching workspaces at all, especially when there's also a popular keybind of mod+scroll to switch workspaces (https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.conf#L158). this just makes the cursor jump around all the time when scrolling

memchr:
> and just focus on the window under the cursor

This will then break `focusurgentorlast` and `focuscurrentorlast`.

memchr:
df51c45d7f9e4e280b959f790f7cde6c88b9b451 warpcursor targets follow_mouse == 1 only

The older iterations of changeWorkspace (commit 5a6d0e9963eaca457a1f91c5ab4c3142c145b646) still focus the window under the cursor because they call simulateMouseMovement directly or indirectly.

- https://github.com/hyprwm/Hyprland/blob/5a6d0e9963eaca457a1f91c5ab4c3142c145b646/src/managers/KeybindManager.cpp#L910
- https://github.com/hyprwm/Hyprland/blob/5a6d0e9963eaca457a1f91c5ab4c3142c145b646/src/managers/KeybindManager.cpp#L895C40-L895C40

If we were to remove these calls, the window under the cursor would still be focused whenever the user moves the mouse (follow_mouse == 1).

So I don't think it can be avoided, because that's what (follow_mouse == 1) does.

At best I think we could introduce an option to disable any mousewarp in changeWorkspace, but its not possible to not change focus at all if the user wants to use follow_mouse == 1.

Per wiki
> 0 - Cursor movement will not change focus.
> 1 - Cursor movement will always change focus to the window under the cursor.
> 2 - Cursor focus will be detached from keyboard focus. Clicking on a window will move keyboard focus to that window.
> 3 - Cursor focus will be completely separate from keyboard focus. Clicking on a window will not change keyboard focus.

thejch:
> > and just focus on the window under the cursor
> 
> This will then break `focusurgentorlast` and `focuscurrentorlast`.

i am not sure how the old behavior would break those

src/Compositor.cpp
memchr: It is from https://github.com/hyprwm/Hyprland/commit/61c9e50bcd6c51e026203f194bc186eda3c44a38

memchr:
> don't warp it

Done

MightyPlaza:
I think the best way to handle this is to add an extra arg to the `workspace` dispatcher to allow disabling the warping on same monitor changes with `follow_mouse=1`. (probably just disable for cursor all settings for consistency)
Should allow user to pick the intended behavior for each individual change

vaxerski:
I don't like this approach. I'd rather deal away with the warps entirely (no config)

it was a bad idea from the start, tbh.

A better option would be to make a `simulateMouseMovement` that cannot change focus, only sends motion events to the current window, use that.

MightyPlaza:
> I don't like this approach. I'd rather deal away with the warps entirely (no config)

same thing I said here: https://github.com/hyprwm/Hyprland/issues/3205
there is no sane way to restore focus and respect `follow_mouse=1`

vaxerski:
the only sane way I see is to restore focus, and just change it if the mouse moves according to `follow_mouse`

if you already moved the mouse you can hover on whatever you want focused :)

memchr:
> it was a bad idea from the start, tbh.

Removing all the warpCursor() calls would be a pretty thorough change. Hyprland has been moving the cursor to the centre of the (keyboard) focused window for some time.

---

Anyway the behaviour should now be the same as 5a6d0e9963eaca457a1f91c5ab4c3142c145b646 by default.

This just adds two extra config values.

vaxerski:
I have pushed what I believe to be the best solution to this.

41e5f401c570eccc0546dcff8bfbe09af16e3ff4
b884544ee63b4c3b3b150ff69cf24c90058c2777

memchr:
Updated. The rest of this MR is unrelated to workspaceChange. And on main, the mouse warp could still be triggered on currently focused window,  e.g. if a window is removed from a group but the cursor is on the rest of the group, in this case, a warp would be considered unnecessary by some.

 So I will leave this open.

Issue #1505, opened 2023-02-07
Option to open windows grouped by default.


Aqa-Ib:
An option to open windows grouped by default would be nice.
Cheers!

memchr:
I think there is a way to implement this by modifying the 
More specifically:
- check for configuration variable, e.g.`misc:new_window_in_group`
- if the value of the configuration variable is `true`, then create new group if possible'

PR #3277, opened 2023-09-12
config: option to open new window as group

memchr:
Add `general:group_new_window`, if true, newly mapped window will open as group if

- no focused window
- focused window is not group
- focused window is locked group

As usual, `ignore_group_lock` has no effect on this. (as noted in the wiki, it currently only affects `moveintogroup`, `moveoutofgroup` and `movewindoworgroup`).

## Other changes

- Moved some general logic governing the transition of a window into a group to `CWindow::Group`.
- Rather wild **clang-format** shenanigans on `MasterLayout.cpp`, mayhap it's time to set up a git hook or CI to automatically format all modified files?

---

closes: #1505
wiki: hyprwm/hyprland-wiki#341

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -640,6 +640,20 @@ bool CWindow::hasPopupAt(const Vector2D& pos) {
     return resultSurf;
 }
 
+void CWindow::Group() {
+    if (!m_sGroupData.pNextWindow) {
+        m_sGroupData.pNextWindow = this;
+        m_sGroupData.head        = true;
+        m_sGroupData.locked      = false;
+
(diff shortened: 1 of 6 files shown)


MightyPlaza:
I think this would be much better off as a window rule, instead of a global setting. (also allowing ignoring groups as an option and normal behaviour)
Can't agree more with the automation of clang-format

MightyPlaza:
a window rule you can make global, no point in adding a value in config to set it for all windows

vaxerski:
I am also in favor of a windowrule

vaxerski:
> Probably more time complexity, but it doesn't matter.

maybe we lose 71ns but we gain more flexibility. You can bet your ass if you make it global one dude is gonna be like "but I want it only on workspace 7"

Issue #2987, opened 2023-08-15
Add window rule for grouping windows


iguanajuice:
### Description

I would like to add something like this to my config:
`windowrule = group, ^virt-manager$`

This will group Virtual Machine Manager by default if this window rule is implemented.

KaspervdHeijden:
This would be a nice. feature. I would be able to spawn my IDE's in a tabbed layout, just like in I3 and Sway :+1:

PR #3279, opened 2023-09-12
windows: add rule `group` to map windows grouped

memchr:
### `group` window rule
```
RULE = "group" {" " GROUP RULE}, WINDOW .
GROUP RULE = 
	| "set" [" always"] 
	| "new" 
	| "lock" [" always"] 
	| "barred" 
	| "invade"
	| "deny"
	| "override " {GROUP RULE}
	| "unset"
```

By default, all rules except `deny` are only effective for new windows (first card).
The `always` qualifier can be used to make some rules effective during the lifetime of a window.

- [x] **set**: Map window grouped if focused window isn't a group.
- [x] **new**: Shorthand of `set barred`. Can be used to open a new group event if the focused window is a group.
- [x] **lock**: Lock the group.
- [x] **barred**: Don't add the window to the focused group.
- [x] **invade**: Force open window in the locked group.
- [x] **deny**: Do not allow window to be toggled as or added to group. This sets `m_sGroupData.deny`
- [x] **override**: Override rules.
- [x] **unset**: Clear all rules.
- [x] **always** qualifier for **new** and **lock**, to make them effective outside first map

A naked `group` rule is a shorthand for `group set`.

### Ban a window from being toggled as a group or added to a group

- [x] `m_sGroupData.deny`: if true, window can't be added to groups or toggled as group.
- [x] **movewindoworgroup**, **moveintogroup**, **moveoutofgroup** respects `m_sGroupData.deny`.
- [x] `m_sGroupData.deny` can be toggled with dispatcher **denywindowfromgroup**.
- [x] window with `m_sGroupData.deny` can be colorized differently.

### Other

- refactor movewindoworgroup
- refactor: Move group creation and destruction logic from KeybindingManager to CWindow.

---
close: #1505 
close: #2987

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -640,6 +640,69 @@ bool CWindow::hasPopupAt(const Vector2D& pos) {
     return resultSurf;
 }
 
+void CWindow::applyGroupRules() {
+    if ((m_eGroupRules & GROUP_SET && m_bFirstMap) || m_eGroupRules & GROUP_SET_ALWAYS)
+        createGroup();
+
+    if (m_sGroupData.pNextWindow && ((m_eGroupRules & GROUP_LOCK && m_bFirstMap) || m_eGroupRules & GROUP_LOCK_ALWAYS))
+        getGroupHead()->m_sGroupData.locked = true;
+}
+
+void CWindow::createGroup() {
+    if (m_sGroupData.deny) {
+        Debug::log(LOG, "createGroup: window:{:x},title:{} is denied as a group, ignored", (uintptr_t)this, this->m_szTitle);
+        return;
+    }
+    if (!m_sGroupData.pNextWindow) {
+        m_sGroupData.pNextWindow = this;
+        m_sGroupData.head        = true;
+        m_sGroupData.locked      = false;
+        m_sGroupData.deny        = false;
+
+        m_dWindowDecorations.emplace_back(std::make_unique<CHyprGroupBarDecoration>(this));
+        updateWindowDecos();
+
+        g_pLayoutManager->getCurrentLayout()->recalculateWindow(this);
+        g_pCompositor->updateAllWindowsAnimatedDecorationValues();
+    }
+}
+
+void CWindow::destroyGroup() {
+    if (m_sGroupData.pNextWindow == this) {
+        if (m_eGroupRules & GROUP_SET_ALWAYS) {
+            Debug::log(LOG, "destoryGroup: window:{:x},title:{} has rule [group set always], ignored", (uintptr_t)this, this->m_szTitle);
+            return;
+        }
+        m_sGroupData.pNextWindow = nullptr;
+        updateWindowDecos();
+        return;
+    }
+
+    CWindow*              curr = this;
+    std::vector<CWindow*> members;
+    do {
+        const auto PLASTWIN                = curr;
+        curr                               = curr->m_sGroupData.pNextWindow;
+        PLASTWIN->m_sGroupData.pNextWindow = nullptr;
+        curr->setHidden(false);
+        members.push_back(curr);
+    } while (curr != this);
+
+    for (auto& w : members) {
+        if (w->m_sGroupData.head)
+            g_pLayoutManager->getCurrentLayout()->onWindowRemoved(curr);
+        w->m_sGroupData.head = false;
+    }
+
+    const bool GROUPSLOCKEDPREV        = g_pKeybindManager->m_bGroupsLocked;
+    g_pKeybindManager->m_bGroupsLocked = true;
+    for (auto& w : members) {
+        g_pLayoutManager->getCurrentLayout()->onWindowCreated(w);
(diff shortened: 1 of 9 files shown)


MightyPlaza:
should allow for args, ex: `group 1` for current behavior, `group 0` to allow removing the rule
(and maybe `group 2` to allow denying going into existing groups)

memchr:
So, now the `group` rule accepts parameters,

- set
- unset
- deny
- locked

See the source for what they do

- https://github.com/hyprwm/Hyprland/pull/3279/files#diff-0ce906c3bba58ce4d8728fb70aa25ece925fd446fdd44e3ea5a846da8399be08R20-R31

some examples
```
exec = [group set locked] vice
exec = [group] foot
exec = [group deny] zynaddsubfx

windowrulev2 = group set locked, workspace:2
windowrulev2 = group unset, blender
```

memchr:
Currently group rules do not survive after the first map. An **alternative approach** is to apply `group` rules not only to the first mapped window, but to all window changes. In this case the `unset` parameter has its own significance and can't be replaced by `override` (ðey are semantically different anyway).

This means that it also affects keybindings, for example :
- `windowrulev2 = group deny, class:blender` will prevent **blender** from ever being added to a group. 
- `windowrulev2 = group unset, class:foot` will prevent **foot** from ever being toggled as a group.

src/events/Windows.cpp
```diff
@@ -198,6 +198,46 @@ void Events::listener_mapWindow(void* owner, void* data) {
...
+                else if (v == "deny")
+                    hasDeny = true;
+                else if (hasDeny && v == "entry")
+                    PWINDOW->m_eGroupRules |= GROUP_DENY_ENTRY;
+                else if (hasDeny && v == "exit")
+                    PWINDOW->m_eGroupRules |= GROUP_DENY_EXIT;
+                else if (v == "override")
+                    // Clear existing rules
+                    PWINDOW->m_eGroupRules = GROUP_OVERRIDE;
+                else if (v == "unset") {
```
memchr, line 232: `group unset` is shorthand for `group override`, but also immediately stops processing other group rules.

src/Window.cpp
```diff
@@ -640,6 +640,59 @@ bool CWindow::hasPopupAt(const Vector2D& pos) {
     return resultSurf;
 }
 
+void CWindow::applyGroupRules() {
+    if ((m_eGroupRules & GROUP_SET && m_bFirstMap) || m_eGroupRules & GROUP_ALWAYS)
+        createGroup();
+    if (m_eGroupRules & GROUP_LOCK && m_sGroupData.pNextWindow)
+        this->getGroupHead()->m_sGroupData.locked = true;
```
memchr, line 647: Bug: should lock new group, not the existing one, or only during the first map.

MightyPlaza:
most uses of this will likely be used 1 time only
ever seems like a strong word
- you should be able to change their initial status with dispatchers (if it seems like a feature to have it permanently it should be opt-in)
- mouse locked and dispatcher locked should be separate things (probably replace the current global setting)
- `deny entry` should probably just be a locked non grouped window (this would allow reusing existing logic and dispatchers)

memchr:
> you should be able to change their initial status with dispatchers (if it seems like a feature to have it permanently it should be opt-in)

`set` is **not** permanent, `always` is.

> mouse locked and dispatcher locked should be separate things (probably replace the current global setting)

I honestly don't know what you're talking about, this patch has nothing to do with the mouse.

> `deny entry` should probably just be a locked non grouped window

lock is a group property.

MightyPlaza:
> `set` is **not** permanent, `always` is.

what about `deny`, can that be removed without window rules?

> I honestly don't know what you're talking about, this patch has nothing to do with the mouse.

I was thinking about https://github.com/hyprwm/Hyprland/pull/2796 (no idea it was you until now)

> lock is a group property, and isn't that exactly what CWindow::m_eGroupRule & GROUP_DENY_ENTRY is?

yes but, if you can lock or deny non grouped windows, the dispatcher `lockativegroup` or some other should remove or set this property has well

memchr:
> I was thinking about https://github.com/hyprwm/Hyprland/pull/2796 (no idea it was you until now)

Eh, now I am even more confused.

> what about deny, can that be removed without window rules?
> yes but, if you can lock or deny non grouped windows, the dispatcher lockativegroup or some other should remove or set this property has well

A `deny entry` window can't be inserted to any group. A `deny` window also can't be made a group.

lockativegroup is about the group, not the window, which is what `m_eGroupRules` is for. (No good variable name)

Mayhap there shall be some dedicated dispatchers, colours around a window banned from clubs, I agree.

memchr:
Should be stable, would you like to review? 
@vaxerski

vaxerski:
is there a wiki mr? I don't understand the `nogroup` options.

memchr:
hyprwm/hyprland-wiki#349

vaxerski (approved):
lgtm, thanks for the work!

PR #3289, opened 2023-09-13
Fix the wrong resize speed in master layout center orientation and remove duplicate code 

thejch:
Fix the issue that in the center orientation of master layout, when you resize with the mouse it moves much slower than the cursor. And also in some cases the direction is wrong.
And remove some duplicate code.

no

ready


vaxerski (approved):
oka

PR #3290, opened 2023-09-13
fix: moveWindowOutOfGroup moves cursor to wrong window

memchr:
In https://github.com/hyprwm/Hyprland/blob/2ad429dfe0687b34c1a4d950e7715f0c8550bed5/src/managers/KeybindManager.cpp#L1999 I mistakenly used `pWindow` instead of `PWINDOWPREV`.

Sorry for the inconvenience.


vaxerski (approved):
unforgivable mistake

Issue #2768, opened 2023-07-21
Dispatcher `lockgroups` does not work


spikespaz:
### Hyprland Version

Branch `main` at commit `1ec0b7b59a84a07f97c3ebee9f6694bc4a1b1dcd`

### Bug or Regression?

Maybe a regression? I do not know, this is the first time I have used it.

### Description

No matter the circumstances, dispatcher `lockgroups` (with any value) does not work for me. This could be `lock`, `unlock`, or `toggle`; none of them work and nothing changes.

**Question:**
What is the defined behavior of `lockgroups toggle` when some groups are already locked and others are not?

**Suggestion:**
The default behavior should be `lock` for `toggle` when some groups are unlocked.

**Suggestion:**
That behavior should have an option, either `lock` or `unlock`. The name could be `lockgroups_toggle_default` or similar. Needs a description because the name is not sufficient to clarify intent. Additionally the option may have a third value `toggle` where it just toggles the lock state on all groups, but this is silly, I can't see a use for it.

### How to reproduce

Reproduction steps:
1. Open a terminal
2. Run `hyprctl dispatch togglegroup`
3. Open a terminal
4. Run `hyprctl dispatch lockactivegroup lock`
5. Open a terminal
6. Run `hyprctl dispatch togglegroup`
7. Open a terminal
8. Run `hyprctl dispatch lockgroups lock`
9. Observe that neither group changes lock state

MightyPlaza:
(iirc)
the lockgroups is a compositor wide variable that disables the ability for groups to get merged using mouse (see https://github.com/hyprwm/Hyprland/pull/2796)

it does not iterate through all groups and change their state

memchr:
`lockgroup` is for a **global** compositor group lock that prevents any window from
being added to a group. `lockactivegroup` is for a **local** group lock that only
prevents windows from being added to itself.

I did not wish to introduce a breaking change when I added the `lockactivegroup`
dispatcher. So the local group locks are unrelated to or modified by
`lockgroup`, an old dispatcher that predates the local lock. They do not
intermingle.

And, in my humble opinion, global and local locks should be kept separate, like
global and local variables in C++.

---

One improvement I think can be made to the `lockgroups` dispatcher is to fire
an event when the global lock changes, which can be picked up by other programs
such as waybar (so that you would know when the global lock is engaged).

spikespaz:
I would close this issue, but I think it warrants: a change to the wiki, and an event.

PR #3317, opened 2023-09-15
binds: emit lockgroups event

memchr:
Have the lockgroups dispatcher fire a `lockgroups` event when the global lock changes, which can be picked up by external programs.

---
doc: hyprwm/hyprland-wiki#346

close: #2768


vaxerski (approved):
the soviet union collapsed in 1991

PR #3321, opened 2023-09-15
Optimize CBezierCurve::getYForPoint

Dickby:
As the title sais, it optimizes getYForPoint

I tested this against the former solution, 
without some tricks the benchmark part of the bezier setup is optimized out.
There is no functional difference to the replaced code other than performance.

Someone should check the performance on other computers.

### DIFF
--- a/src/helpers/BezierCurve.cpp
+++ b/src/helpers/BezierCurve.cpp
@@ -48,27 +48,25 @@ float CBezierCurve::getXForT(float t) {
 
 // Todo: this probably can be done better and faster
 float CBezierCurve::getYForPoint(float x) {
-    if (x >= 1.0)
-        return 1.0;
+    if (x >= 1.f)
+        return 1.f;
 
-    // binary search for the range UPDOWN X
-    int upperT = BAKEDPOINTS - 1;
-    int lowerT = 0;
-    int mid    = upperT / 2;
+    int  index = 0;
+    bool below = true;
+    for (int step = (BAKEDPOINTS + 1) / 2; step > 0; step /= 2) {
+        if (below)
+            index += step;
+        else
+            index -= step;
 
-    while (std::abs(upperT - lowerT) > 1) {
-        if (m_aPointsBaked[mid].x > x) {
-            upperT = mid;
-        } else {
-            lowerT = mid;
-        }
-
-        mid = (upperT + lowerT) / 2;
+        below = m_aPointsBaked[index].x < x;
     }
 
+    int lowerIndex = index - (!below || index == BAKEDPOINTS - 1);
+
     // in the name of performance i shall make a hack
-    const auto LOWERPOINT = &m_aPointsBaked[std::clamp(lowerT, 0, BAKEDPOINTS - 1)];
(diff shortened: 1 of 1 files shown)


Dickby:
What's up with this Nix shit?
I am not responsible, am I?

Dickby:
Playing around with this, i reallized that the old implementation could lead to lowerT == upperT after clamping.
Of course, otherwise clamping would not be needed.
That could either happen for lowerT = upperT = 0, or lowerT = upperT = BAKEDPOINTS -1.
In Both cases PERCINDELTA would become nan i think.
After the latest change to my PR this should not happen anymore.

Dickby:
A thing to think about:
For our use case, bezier curves that have more than one y value for a given x value don't make sense.
They could lead to jumps in animations.
It would be nice to restrict the possible bezier curves so that every getXForT(t) < getXForT(t+epsylon).

src/helpers/BezierCurve.cpp
```diff
@@ -46,34 +46,34 @@ float CBezierCurve::getXForT(float t) {
...
-        mid = (upperT + lowerT) / 2;
+        below = m_aPointsBaked[index].x < x;
     }
 
-    // in the name of performance i shall make a hack
-    const auto LOWERPOINT = &m_aPointsBaked[std::clamp(lowerT, 0, BAKEDPOINTS - 1)];
-    const auto UPPERPOINT = &m_aPointsBaked[std::clamp(upperT, 0, BAKEDPOINTS - 1)];
+    int lowerIndex = index - (!below || index == BAKEDPOINTS - 1);
+
+    // in the name of performance i shall make a hack (is the hack still here or can this comment be deleted?)
```
vaxerski, line 68: it is, don't remove the comment :)

src/helpers/BezierCurve.cpp
```diff
@@ -46,34 +46,34 @@ float CBezierCurve::getXForT(float t) {
...
-
-    while (std::abs(upperT - lowerT) > 1) {
-        if (m_aPointsBaked[mid].x > x) {
-            upperT = mid;
+    if (x >= 1.f)
+        return 1.f;
+
+    int  index = 0;
+    bool below = true;
+    for (int step = (BAKEDPOINTS + 1) / 2; step; step /= 2) {
```
vaxerski, line 57: I'd rather have some more sane condition for step like e.g. step > 0

src/helpers/BezierCurve.cpp
```diff
@@ -46,34 +46,34 @@ float CBezierCurve::getXForT(float t) {
...
+        if (below) {
```
vaxerski, line 58: I know old code had the issue too, but no {} around short ifs

vaxerski:
> For our use case, bezier curves that have more than one y value for a given x value don't make sense.
They could lead to jumps in animations.
It would be nice to restrict the possible bezier curves so that every getXForT(t) < getXForT(t+epsylon).

Indeed.

vaxerski:
I will review this tomorrow, as I need to do some analysis

vaxerski (approved):
lgtm :) thanks!

PR #3329, opened 2023-09-16
binds(changegroupactive): support number as arg

memchr:
Add support for using number as argument to `changegroupactive` dispatcher to select group by index

- Index starts with `1`.
- Use `0` selects the last group member.


vaxerski (commented):
a) this won't work with e.g. 17
b) why not use `isNumber` and then `std::stoi`?

memchr:
> a) this won't work with e.g. 17

I have no idea if any keyboard has a <kbd>17</kbd> key.
But still, modified

vaxerski (approved):
in the future we could also allow relative

vaxerski:
wiki mr?

memchr:
hyprwm/hyprland-wiki#351

Issue #1426, opened 2023-01-24
Add option to save layouts


SimonBrandner:
Add option to save layouts similarly to what [i3wm allows you to do it](https://i3wm.org/docs/layout-saving.html)

yogeshkumar98103:
It'll be very useful to have an option to automatically restore last layout on reboot.

magicBeast:
When doing development I often have the same four or more windows open, in the same layout. It would be useful to be able to save a session and quickly restore any required layout for a given task or set of tasks. Obviously this can be hacked together via scripts, but it does seem like natural extension of functionality for a tiling compositor that handles window layout.

witnes0417:
yes this would be very useful, even though tmux complement in its way, but why need to configure another window navigating feature in it?

memchr:
How would some applications that open multiple windows, so that they all have the same PID, be distinguished? They may also have the same criteria.

ScuttleSE:
I just assumed this was available when I installed Hyprland and I must say...disappointed that it is not available, and a pull request to actually add it was rejected...

EnderSpy29:
yo i'd love this, having all my windows opening with certain sizes in certain positions in certain workspaces would be amazing. i'd be great if we could either take a snapshot and bind that to a key or something. or have hyprland load the last used layout on startup

tonysvo:
agreed, I want to have few workspace pre-setup every time I login

izmyname:
As a workaround, you can autostart desired programs. Here is an example from my config.

` exec-once=[workspace 1 silent] $browser & sleep 4; $clock & sleep 1; $filemanager & $telegram &`

izmyname:
Also, I see several pull requests, relevant to this issue, were declined, so I guess this function isn't going to be implemented. Should the issue be closed?

EnderSpy29:
> As a workaround, you can autostart desired programs. Here is an example from my config.
> 
> ` exec-once=[workspace 1 silent] $browser & sleep 4; $clock & sleep 1; $filemanager & $telegram &`

i have programs executed on startup. what we want is the ability to set all applications to load in a predefined location, size and layout

zleeepy:
i need this

Fork3r:
oh god this is musthave feature

Cool-Game-Dev:
I want to save my screenshot layout because its a bit annoying to set up

agluck91:
[...]

PR #3337, opened 2023-09-17
save and restore layout

memchr:
## Features

Window floating, fullscreen, group states are retained.
For dwindle, the layout should be restored as is, preserving size and orientation. 
Any windows that do not match will be moved as a group to the last workspace. 

### Dispatcher

use the dispatcher `necromancy` to save and restore layout.

- `necromancy save`: save layout (on disk)
- `necromancy restore`: restore layout (on disk)

The dispatcher can take the path to the save file as an optional argument.

### Hyprctl / IPC

Added hyprctl subcommand `necromancy` to support the following syntax

- **save [path]**: save layout
- **restore [path]**: restore layout
- **inspect [path]**: inspect layout file, support JSON output

### Notes

Currently,

- layout restoration is only implemented with 'dwindle'.
- multiple monitor support has not been tested.

## TODO

- [x] add a dispatcher
  - [x] specific save/restore location in args
- [x] persistent layout save across session (save file)
- [x] configuration save location
- [x] Better error handling

### Restore window state

- [x] group, including local lock
- [x] fake fullscreen
- [x] fullscreen
- [x] pseudo tilling
- [x] floating
  - [x] size, position
  - [x] pin

### Restore workspace state

- [x] workspace id, name
- [x] previous workspace
- [x] special workspace

### Restore layout

- [x] dwindle layout
- ~~master layout~~ I am  unfamiliar with it, so no interest at present
- [x] deal with stray windows (not in layout save file)

### Other

- ~~comply to name convention, advNamely adjHungarian nNotation~~
- [x] remove testing code
- [x] hyprctl command for inspecting current saved layout

## Implementation

Third party dependencies need to be introduced as subprojects and build scripts need to be modified when using popular formats such as JSON, **so this implementation uses simple binary serialisation for ease of development**.

The layout file is located at `~/.local/share/hypr/layout.bin`, or `~/.local/share/hypr/layoutd.bin` for debug builds.

### Serialization format

| Section     |             |                                                  |
| ----------- | ----------- | ------------------------------------------------ |
| Header      | signature   | `"HYPR NECROMANCY"`, for ident                   |
|             | version     | `int`, for verification                          |
|             | layout name | `std::string`                                    |
| Window      | count       | `size_t`                                         |
|             | entries     | `map<(uintptr_t)CWindow*,SWindowState>`          |
| Workspaces  | count       | `size_t`                                         |
|             | entries     | `SWorkspaceState[]`                              |
| LayoutNodes | count       | `size_t`                                         |
|             | entries     | `map<SNodeData*, SNodeData>` (depends on layout) |

- Variable length structures are stored with their length

---

close: #1426

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -416,7 +416,9 @@ int main(int argc, char** argv) {
         request(fullRequest, 2);
     else if (fullRequest.contains("/hyprpaper"))
         requestHyprpaper(fullRequest);
-    else if (fullRequest.contains("/--help"))
+    else if (fullRequest.contains("/necromancy")) {
+        request(fullRequest + " --cwd " + std::filesystem::current_path().string(), 1);
+    } else if (fullRequest.contains("/--help"))
         printf("%s", USAGE.c_str());
     else {
         printf("%s\n", USAGE.c_str());
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -45,6 +45,14 @@ std::string CConfigManager::getConfigDir() {
     return configPath;
 }
 
+std::string CConfigManager::getDataDir() {
+    static const char* xdgDataHome = getenv("XDG_DATA_HOME");
+    if (!xdgDataHome)
+        return std::filesystem::path(getenv("HOME")) / ".local" / "share" / "hypr/";
+    else
+        return std::filesystem::path(xdgDataHome) / "hypr/";
+}
+
 std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
@@ -122,6 +130,7 @@ void CConfigManager::setDefaultVars() {
     configValues["misc:close_special_on_empty"].intValue       = 1;
     configValues["misc:groupbar_text_color"].intValue          = 0xffffffff;
     configValues["misc:background_color"].intValue             = 0xff111111;
+    configValues["misc:layout_save_file"].strValue             = std::filesystem::path(getDataDir()) / (ISDEBUG ? "layoutd.bin" : "layout.bin");
 
     configValues["debug:int"].intValue                = 0;
     configValues["debug:log_damage"].intValue         = 0;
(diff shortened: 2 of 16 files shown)


memchr:
[functional test](https://github.com/hyprwm/Hyprland/assets/118117622/566280e6-1fd2-4240-8836-2234319464b2)

memchr:
Identification is based on the current `class` and `title`, but I think should the identification fails, `initialClass` and `initialTitle` should be used as a fallback, This would be useful for windows that frequently change their title, such as browser windows

memchr:
How should strays (windows that are not tracked by the layout save) be dealt with? 
- Move them all to special
- Just add them to the layout

memchr:
https://github.com/hyprwm/Hyprland/assets/118117622/d849bb54-6ad7-48be-95d7-c2b3e8e83f84

vaxerski:
I appreciate the work but am unsure if this should be in core.

Remembering what apps were open maybe would make sense, but layouts, I don't think so.

memchr:
> unsure if this should be in core.

I have tried to write a plugin, but it is tedious to restore the layout nodes as some important members are private. So IMO the most straightforward way to do this is to just implement it in core.

> Remembering what apps were open maybe would make sense, but layouts, I don't think so.

Can you elaborate? I don't understand this.

vaxerski:
well, it's not about remembering state, but what windows does. Just remember what apps were open on what workspace and reopen them

memchr:
Well, that's a nail in the coffin, but may I keep this open?

vaxerski:
feel free.

memchr:
> > Well, that's a nail in the coffin, but may I keep this open?
> 
> Hey @memchr, if you would like you can send the PR to my fork, where I'm also keeping another rejected PR about persistent desktops. I think it's a good fit for people with our workflow. I'm trying to keep merging from upstream about once a week, of course at some point there could be too much conflict making it not worth to mantain anymore, but for the time being... I also packaged the fork in the AUR, so it's easy to install in Arch.

You can just merge it yourself, you know, with `refs/pull`

justyn:
If the work itself should not be in core due to its size and complexity, but it is difficult to implement as a plugin due to insufficient public interfaces, could the answer be to adjust the public interfaces to make plugins like this more workable?

This feature seems very useful, and actually being implemented as a plugin might allow more alternative implementations to spring up and therefore innovation/experimentation.

Just a thought.

vaxerski:
if you have an idea on how to make proper access modifiers for the plugins, feel free to make a MR :)

Issue #2930, opened 2023-08-10
Completely disable hypr-chan


REALERvolker1:
### Description

I suggest a new bool config line, `disable_hypr_chan`. This will disable the random hyprchan that is displayed on the default background.

Personally I find it very strange how this wasn't added when hypr-chan was introduced.

NotAShelf:
nah

vaxerski:
hypr-chan is very sad from reading this.

pechano:
There is probably an easier way, but if you build from source, you can edit the top line of this file:
`Hyprland/assets/meson.build`
 to not include the anime wallpapers, like this:
`wallpaper_types = ['']`

jlkn:
```patch
diff --git a/assets/meson.build b/assets/meson.build
index ba2c77bd..2c37dbac 100644
--- a/assets/meson.build
+++ b/assets/meson.build
@@ -1,4 +1,4 @@
-wallpaper_types = ['', 'anime_', 'anime2_']
+wallpaper_types = ['']
 
 foreach type : wallpaper_types
   foreach size : [2, 4, 8]
```

kojq:
What's wrong with `disable_hyprland_logo`?

pechano:
> What's wrong with `disable_hyprland_logo`?

That also disables the normal logo. So you need something like hyprpaper to set the non-amine background as your wallpaper. Which is what I do now.

burein-ita:
> > What's wrong with `disable_hyprland_logo`?
> 
> That also disables the normal logo. So you need something like hyprpaper to set the non-amine background as your wallpaper. Which is what I do now.

That closes the issue. No?

pechano:
Yes, it was closed five days ago.

pechano:
kek indeed

meedstrom:
> (for the record I also support the option to disable hyprchan, but I am fine with this solution for now)

I am not fine with compiling from source.

Maybe the documentation can indicate that `disable_hyprland_logo` does the trick.

fufexan:
I've made a PR that lets you disable Hyprchan, please test it if you're interested.

janithcooray:
such betryal, hypr-chan will be sad

Sbatushe:
> What's wrong with `disable_hyprland_logo`?

Because it's cringe to have a girl anime wallpaper bug appearing from nowhere while at work or (worse) work calls and meet

Sbatushe:
[...]

PR #3343, opened 2023-09-17
hyprchan: add disable options

fufexan:

A lot of people complained about how hard it is to disable Hyprchan while still keeping the default Hyprland logo wallpaper. This fixes the issue by allowing users to disable Hyprchan wallpapers from being selected when randomising the wallpaper, or completely remove the wallpaper from the Hyprland installation (if installed with Meson).

- meson: add option (not) to include hyprchan
- misc: add disable_hypr_chan option

Addresses https://github.com/hyprwm/Hyprland/issues/2930.

No.

Ready.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1775,8 +1775,9 @@ void CHyprOpenGLImpl::renderSplash(cairo_t* const CAIRO, cairo_surface_t* const 
 void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
     RASSERT(m_RenderData.pMonitor, "Tried to createBGTex without begin()!");
 
-    static auto* const              PNOSPLASH      = &g_pConfigManager->getConfigValuePtr("misc:disable_splash_rendering")->intValue;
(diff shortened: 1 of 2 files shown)


src/render/OpenGL.cpp
```diff
@@ -1797,7 +1798,10 @@ void CHyprOpenGLImpl::createBGTextureForMonitor(CMonitor* pMonitor) {
     // or configure the paths at build time
 
     // get the adequate tex
-    std::string texPath = "/usr/share/hyprland/wall_" + std::string(USEANIME ? (distribution2(engine) == 0 ? "anime_" : "anime2_") : "");
+    std::string texPath = "/usr/share/hyprland/wall_";
+    if (!*PDISABLEHYPRCHAN) {
```
vaxerski, line 1802: style: no {} around short ifs

vaxerski (commented):
idk why a meson opt is needed if we disable it via the config though. What if someone builds without and then enables in the config?

vaxerski (approved):
oka

fufexan:
> idk why a meson opt is needed if we disable it via the config though. What if someone builds without and then enables in the config?

Good question. Dropped meson option.

vaxerski:
make a commit to the wiki too please

PR #3369, opened 2023-09-19
refactor(varlist): replace unintuitive string splitting

memchr:
Tested with asserts, result should be identical to old logic.

Also added `removeEmpty` as an optional arg to remove empty args from argv, defaults to false,


vaxerski (approved):
modern C++ nice

PR #3371, opened 2023-09-19
xwayland: allow to disable xwayland via config option

vnepogodin:
Adds ability to disable xwayland without recompilation with the flag


vaxerski:
so it just adds an option but that option aint doing nothin'?

vnepogodin:
> so it just adds an option but that option aint doing nothin'?

By default yes nothing will happen, because by default xwayland is enabled.

If you add:
```
xwayland {
  enabled = false
}
```
Then it wont spawn xwayland.

vaxerski:
based on the diff I'd like to contest that statement

vnepogodin:
lol the other file hasnt been added

vaxerski:
well, add it then maybe? :)

vnepogodin:
> well, add it then maybe? :)

yeah already pushed. sry haha

src/managers/XWaylandManager.cpp
```diff
@@ -9,6 +9,10 @@
 
 CHyprXWaylandManager::CHyprXWaylandManager() {
 #ifndef NO_XWAYLAND
+    static auto* const XWAYLAND_ENABLED = &g_pConfigManager->getConfigValuePtr("xwayland:enabled")->intValue;
+    if (!*XWAYLAND_ENABLED) {
```
vaxerski, line 13: style nit: no {} around short ifs

src/managers/XWaylandManager.cpp
```diff
@@ -9,6 +9,10 @@
 
 CHyprXWaylandManager::CHyprXWaylandManager() {
 #ifndef NO_XWAYLAND
+    static auto* const XWAYLAND_ENABLED = &g_pConfigManager->getConfigValuePtr("xwayland:enabled")->intValue;
```
vaxerski, line 12: style nit: no_snake

vaxerski (commented):
outside of this, since xwayland runs in parallel, you'll need to add a function (probably in the config manager) to check the status of xwayland on config reloads.

if someone edits the option, we might want to kill / start xwayland

vnepogodin:
> outside of this, since xwayland runs in parallel, you'll need to add a function (probably in the config manager) to check the status of xwayland on config reloads.
> 
> if someone edits the option, we might want to kill / start xwayland

well I'd rather restrict editing that option. It should be initiated just once in the lifetime of the program

vaxerski:
this is absolutely wrong, as xwayland will naturally be shut down if no xwayland apps are open, anyways (lazy mode in hyprland is always on)

src/managers/XWaylandManager.cpp
vnepogodin: that's screaming snake case, which is even used above )

all capitals just doesn't exist. thats little offtopic for a future

src/managers/XWaylandManager.cpp
vnepogodin: alright

vnepogodin:
> this is absolutely wrong, as xwayland will naturally be shut down if no xwayland apps are open, anyways (lazy mode in hyprland is always on)

the intention for it is to force everything to native wayland. in lazy mode it will still launch as x11 app if broken detection with happened for it to fallback to xorg, etc

src/managers/XWaylandManager.cpp
vaxerski: > all capitals just doesn't exist. thats little offtopic for a future

I'd like to contest that information

image

vaxerski:
yes, but what I want is for people to be able to dynamically toggle xwayland on / off.

As in, if xwayland is enabled and there are xwayland apps open, and someone disables it, kill xwayland.

> force everything to native wayland.

that is done by `setenv` and `unsetenv` of `$DISPLAY`

src/managers/XWaylandManager.cpp
vnepogodin: > > all capitals just doesn't exist. thats little offtopic for a future
> 
> I'd like to contest that information
> 
> image

thanks for showing me terrible code)

vnepogodin:
> yes, but what I want is for people to be able to dynamically toggle xwayland on / off.
> 
> As in, if xwayland is enabled and there are xwayland apps open, and someone disables it, kill xwayland.
> 
> > force everything to native wayland.
> 
> that is done by `setenv` and `unsetenv` of `$DISPLAY`

well yeah I can look into that

h0tc0d3:
You don’t need any extra parameters, you can do this and xwayland won’t start.
```
env = WLR_XWAYLAND, /dev/null
```

PR #3377, opened 2023-09-20
fix(log): use std::format_string to catch formatting string errors at compile time 

memchr:
## Abstract

This patch changed signature of `Debug::log` to
```cpp
void log(LogLevel level, std::format_string<Args...> fmt, Args&&... args);
```
`std::format_string<Args...>` would ensure that the format string and args is valid a compile time, eliminated any possibility of `std::vformat()` throwing `std::format_error`

Actually, `std::format()`, which does not throw `std::format_error` as specified in the c++20 standard in stdlib, is defined as follows in **libstdc++**

```cpp
  template<typename... _Args>
    [[nodiscard]]
    inline string
    format(format_string<_Args...> __fmt, _Args&&... __args)
    { return std::vformat(__fmt.get(), std::make_format_args(__args...)); }
```

Thus, `Debug::log` will now call `std::vformat()` without any exception handling for the above reason.

## Rational

The old signature of `Debug::log(LogLevel level, const std::string& fmt, Args&&... args)` is unsafe,
as the erroneous expressions are listed at compile time, causing `std::vformat()` to throw `std::format_error`.

For example, the current codebase contains the following types of mistakes.

#### Format string does not have enough args

https://github.com/hyprwm/Hyprland/blob/d8d0cd75c2db16a14086e1d5119ed937a60c701e/src/render/Renderer.cpp#L1614-L1615
https://github.com/hyprwm/Hyprland/blob/d8d0cd75c2db16a14086e1d5119ed937a60c701e/src/render/Renderer.cpp#L1727-L1728

#### Invalid format specifier

https://github.com/hyprwm/Hyprland/blob/d8d0cd75c2db16a14086e1d5119ed937a60c701e/src/debug/HyprCtl.cpp#L413-L419

Although `Debug::log` will catch and log such errors, this is suboptimal because:

- invalid formatting is handled at runtime (exception path is slower), 
- leads to subtle bugs, because we would not find out that a format string is invalid without triggering the exception and going through the logs.

## Other changes

- deprecate `getFormat` [1], replaced existing usage with `std::format` [2]
- remove `getFormat` in **hyprctl/main.cpp**

---

- [1] Some plugins or PR may use this.
- [2] `std::format` only throws `std::bad_alloc` as defined in standard, however, almost any stdlib types could potentially throw this if the host runs out of memory, and we are not writing code for nuclear plains or Mars rovers, so no recovery is necessary.

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -12,6 +12,7 @@
 #include <ranges>
 #include <algorithm>
 #include <signal.h>
+#include <format>
 
 #include <iostream>
 #include <string>
@@ -104,20 +105,6 @@ std::vector<SInstanceData> instances() {
     std::sort(result.begin(), result.end(), [&](const auto& a, const auto& b) { return a.time < b.time; });
 
     return result;
-}
-
-std::string getFormat(const char* fmt, ...) {
-    char*   outputStr = nullptr;
-
-    va_list args;
-    va_start(args, fmt);
-    vasprintf(&outputStr, fmt, args);
-    va_end(args);
-
-    std::string output = std::string(outputStr);
-    free(outputStr);
-
-    return output;
 }
 
 void request(std::string arg, int minArgs = 0) {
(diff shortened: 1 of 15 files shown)


vaxerski (approved):
lgtm, thanks as usual!

PR #3395, opened 2023-09-22
refactor: manually resets singletons

memchr:
Manually destroy global static objects, as some of them are interdependent and the wrong order of destruction can cause segfaults or other funny problems, such as libc++ build creating  files like `'$'\0x32\0x324\0xF'<'` during hyprland shutdown process.

---

Static objects are destroyed in the reverse order of construction, but for Hyprland they are not in the same translation unit, so the order of destruction could be rather chaotic, if I am not mistaken.


vaxerski (approved):
steh auf

PR #3400, opened 2023-09-23
Unbreak build without precompiled headers

jbeich:
Fixes build with `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON` (CMake) or `-Db_pch=false` (Meson).


vaxerski (approved):
fair, thanks!

Issue #3403, opened 2023-09-23
closing-opening lid with external monitor connected crashes hyprland


Seme4eg:
### Hyprland Version

Hyprland, built from branch main at commit fb80cbe415835a7e0c64e2038d0fc56e8aba0ab4 dirty

### Bug or Regression?

Bug

### Description

Failing to setup a laptop to work with external monitor connected via HDMI. To move all workspaces on external monitor by disabling eDP-1 on lid close and on lid open move them back by turning it back on. For now to achieve it i use a script that i call on lid off and on events, screens below.

### How to reproduce

And all the times i try to open a lid after closing it with external monitor plugged in i am having either a black screen, like eDP-1 can't be enabled or straight a hyprland crash (logs attached).

### Crash reports, logs, images, videos

Crash report:
[hyprlandCrashReport10414.txt](https://github.com/hyprwm/Hyprland/files/12706372/hyprlandCrashReport10414.txt)

script that i run on open/close lid:
```bash
#!/bin/bash

state=$1

if [ "$state" == "on" ]; then
  systemctl --user restart wlsunset
  change-theme
  sync_tz_and_loc
  hyprctl keyword monitor "eDP-1,highrr,0x0,1.7"
else
  count_monitors=$(hyprctl monitors | grep -c '^Monitor')
  hyprctl keyword monitor "eDP-1, disable"
  # in order for this to work gotta have 'HandleLidSwitch=ignore' set in
  # /etc/systemd/logind.conf
  [ "$count_monitors" = 1 ] && systemctl suspend
fi
```
my bindl switch bindings:
```conf
bindl=,switch:on:Lid Switch,exec,lidhandle on
bindl=,switch:off:Lid Switch,exec,lidhandle off
```

My hyprland config directory in case that can help: https://github.com/Seme4eg/dotfiles/tree/master/.config/hypr

vaxerski:
[patch.txt](https://github.com/hyprwm/Hyprland/files/12706439/patch.txt)

try this

Seme4eg:
that did fix it, thanks! Should i close the issue?

vaxerski:
yeah. Sorry, didn't reply in time. Glad it helped

Issue #3175, opened 2023-09-05
Enabling laptop monitor leads to crash.


rszyma:
### Hyprland Version

Hyprland, built from branch main at commit 4eecb8bffc829ab2222b4acea1942ef8ee7b8904 dirty (config: don't enable vrr 2 for maximized workspaces). Tag: v0.29.1-11-g4eecb8bf  flags: (if any)

### Bug or Regression?

Bug

### Description

I have an external hdmi monitor. Enabling laptop monitor while the hdmi one is plugged with the config below leads to crash. It only happens when live reloading. When running Hyprland fresh from terminal it works normally.

### How to reproduce

1. Use this config for monitors:
```
monitor=eDP-1,disable
monitor=HDMI-A-1,1920x1080@180,0x0,1
```
2. Change the monitors config to this:
```
monitor=HDMI-A-1,1920x1080@180,0x0,1
```
3. Hyprland crashes

### Crash reports, logs, images, videos

[hyprlandCrashReport15979.txt](https://github.com/hyprwm/Hyprland/files/12517762/hyprlandCrashReport15979.txt)

jakubreron:
Looks like your issue is similar to this one: https://github.com/hyprwm/Hyprland/issues/2770

rszyma:
It's a different issue. I've tried today's patches and the crash is still reproductible.

I haven't looked at the code (and neither I will, because is my eyes start to bleed when I see c++), but just from the last log messages in crash log, I can tell that the laptop monitor is being initialized with wrong position (2147483647,-2147483647) and size (0x0).

vaxerski:
somehow that rule doesn't get picked up. Wuuttt

rszyma:
Trying again in fa3de9b70e95ea8f0ebac16ae02f8f0eba0d5b92 (v0.29.1-17-gfa3de9b7) and I can't reproduce there. Maybe db48f97 fixed this? I'll also make sure if there's no regression on master just in case, and close this issue after confirming.

rszyma:
I've noticed that the crash only happens when starting Hyprland with the monitors config in `1.` (laptop monitor disabled). When started with monitors config like in `2.` I can comment/uncommment `monitor=eDP-1,disable` line freely without any crashes.

rszyma:
no longer can repro on main f10996b

Issue #3067, opened 2023-08-24
segfault when unplugging monitor


primalmotion:
### Hyprland Version

Hyprland, built from branch main at commit 90c03e5bd2204ba6d1a0167c68f65b7a9231bef4 dirty (master: layout resize makes the wrong size changes (3064)). Tag: v0.28.0-89-g90c03e5b

### Bug or Regression?

Bug

### Description

Segfault when unplugging an external monitor

### How to reproduce

Unplug the monitor

### Crash reports, logs, images, videos

I can't attach it, github says something went really wrong.. pasting

```
--------------------------------------------
   Hyprland Crash Report
--------------------------------------------
Well this is awkward.

Hyprland received signal 11 (Segmentation fault)

Version: 90c03e5bd2204ba6d1a0167c68f65b7a9231bef4
Tag: v0.28.0-89-g90c03e5b

System info:
	System name: Linux
	Node name: axion
	Release: 6.4.11-arch2-1
	Version: #1 SMP PREEMPT_DYNAMIC Sat, 19 Aug 2023 15:38:34 +0000

GPU:
	00:02.0 VGA compatible controller [0300]: Intel Corporation Comet Lake UHD Graphics [8086:9bca] (rev 04) (prog-if 00 [VGA controller])
	DeviceName: VGA compatible controller


os-release:
	NAME="Arch Linux"
	PRETTY_NAME="Arch Linux"
	ID=arch
	BUILD_ID=rolling
	ANSI_COLOR="38;2;23;147;209"
	HOME_URL="https://archlinux.org/"
	DOCUMENTATION_URL="https://wiki.archlinux.org/"
	SUPPORT_URL="https://bbs.archlinux.org/"
	BUG_REPORT_URL="https://bugs.archlinux.org/"
	PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
	LOGO=archlinux-logo
	


Backtrace:
	#0 | Hyprland(_Z12getBacktracev+0x62) [0x55af1cbfa4d2]
		??
		??:0
	#1 | Hyprland(_ZN13CrashReporter18createAndSaveCrashEi+0x791) [0x55af1cbca651]
		??
		??:0
	#2 | Hyprland(_Z25handleUnrecoverableSignali+0x44) [0x55af1cb5bf94]
		??
		??:0
	#3 | /usr/lib/libc.so.6(+0x3e710) [0x7f675d63e710]
		??
		??:0
[...]

PR #3404, opened 2023-09-23
Further unsafe state improvements

vaxerski:
Instead of allowing Hyprland to sit in a state where there are no monitors, which various parts of the code don't like, we create a fake headless output on all monitor disconnect, and then remove it when a monitor appears.

cc @primalmotion @rszyma 

possibly fixes stuff like #3175 #3403 #3067


fufexan (approved):

primalmotion:
nice! I'll give a try soon

PR #3432, opened 2023-09-27
Make `hyprctl activeworkspace` use specialworkspace if available

### LINKED ISSUE #3406: hyprctl activeworkspace outputs only focussed regular workspace even on special workspaces
### Hyprland Version

Hyprland, built from branch  at commit v0.29.1  (). Tag: ?  flags: (if any) debug

### Bug or Regression?

Bug

### Description

When a special workspace is focused `hyprctl activeworkspace` returns the workspace that would be active if the special workspace is disabled

### How to reproduce

1. Open a terminal emulator in a special workspace
2. run `hyprctl activeworkspace`

_No response_

PiotrekB416:

This PR fixes special workspace not showing in `hyprctl activeworkspace`
Fixes #3406 

Ready for merging

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -232,7 +232,14 @@ std::string activeWorkspaceRequest(HyprCtl::eHyprCtlOutputFormat format) {
         return "unsafe state";
 
     std::string result = "";
-    auto        w      = g_pCompositor->getWorkspaceByID(g_pCompositor->m_pLastMonitor->activeWorkspace);
+
+    int id = g_pCompositor->m_pLastMonitor->specialWorkspaceID;
+
+    if(!id) {
+        id = g_pCompositor->m_pLastMonitor->activeWorkspace;
+    }
+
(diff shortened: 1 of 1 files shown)


vaxerski:
I don't get the point. There is a special workspace field in `hyprctl monitors`

fabian-thomas:
@vaxerski Could you elaborate? I understand that one can just do the following to achieve what is requested here:
`hyprctl activewindow -j | jq -r -e '.workspace.name'`
So going over the active window to get the active workspace, even if it is special. But why does hyprland not return the special workspace on calling `activeworkspace` (when a window on the special workspace is focused)?
Is it that `activeworkspace` should always return the current *normal* workspace? It would make sense to always have that available. But then there should be two commands:
1. `activeworkspace`: return special or normal workspace that is in focus
2. `activenormalworkspace`: return the underlying normal workspace in any case

vaxerski:
hyprctl monitors -> find one with `focused: yes`, inspect its `special workspace:` field.

PR #3456, opened 2023-09-30
Remove redundant clamp of smoothsteps return value.

Dickby:
It removes the clamp around smoothstep.
Smoothstep already clamps the return value.
Reference: https://registry.khronos.org/OpenGL-Refpages/gl4/html/smoothstep.xhtml
In the description t is clamped to 0.0 ,1.0,
the equation `t * t * (3.0 - 2.0 * t )` will always be between 0.0 and 1.0 for that clamped t.
As proof plot the function `3*x^2 - 2*x^3` which is equal to `t * t * (3.0 -2.0 * t)`
Ready

### DIFF
--- a/src/render/shaders/Border.hpp
+++ b/src/render/shaders/Border.hpp
@@ -71,11 +71,11 @@ void main() {
 
 	    if (dist < radius - h) {
             // lower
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
nice, thankies

Issue #3258, opened 2023-09-11
Flatpak apps aren't able to open links in Browsers. 


RoccoRakete:
### Hyprland Version

v0.29.1-2-g8fefb180

### Bug or Regression?

Bug

### Description

Flatpak apps aren't able to open links in Browsers. According to #https://github.com/emersion/xdg-desktop-portal-wlr/issues/193 this seems to be an issue with xdg-desktop-portal.
Does anyone has the same issue or even a fix for that? Is it a bug?  

xdg-desktop-portal-hyprland-git 1:r296.aca5160-1

### How to reproduce

Try to open a link in a flatpak-app

### Crash reports, logs, images, videos

Output from Iotas: 

 Failed to read portal settings: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface
»org.freedesktop.portal.Settings« n object at path /org/freedesktop/portal/desktop

hoywu:
For anyone who encounters this problem:
Edit `/usr/share/xdg-desktop-portal/hyprland-portals.conf` as follows helped me solve the issue.
```
[preferred]
default=hyprland;gtk
```

Reference:
https://github.com/flatpak/flatpak/issues/1508#issuecomment-1702714679
https://github.com/hyprwm/Hyprland/pull/3469

vaxerski:
this is not a hyprland issue, so closing

PR #3469, opened 2023-10-02
Add GTK portal as fallback

gnusenpai:
On systems with xdg-desktop-portal 1.17+, the GTK portal is not started by default and cannot be used (even if started manually). This should fix issues on such systems like: theming/cursors being incorrect, not being able to open the file picker, and being unable to open links from other applications (#3258).

Users who wish to use an alternative portal (e.g. for QT-based dialogs) should still be able to do so by overriding this with their own preferences in `~/.config/xdg-desktop-portal/hyprland-portals.conf`. This PR just makes the portals act like they did before xdg-desktop-portal 1.17.

In my brief testing, this doesn't cause any regressions for those on xdg-desktop-portal 1.16.

### DIFF
--- a/assets/hyprland-portals.conf
+++ b/assets/hyprland-portals.conf
@@ -1,2 +1,2 @@
 [preferred]
-default=hyprland
+default=hyprland;gtk


vaxerski (approved):
lgtm

Issue #3519, opened 2023-10-08
Movewindow doesn't account for monitor offset when moving floating windows


leonardvdj:
### Hyprland Version

Hyprland, built from branch main at commit 728a8bb48e0f7de1cbe1ad13fb469754c3d0bc97 dirty (config: mention force_default_wallpaper in default configs). Tag: v0.30.0-67-g728a8bb4

### Bug or Regression?

Bug

### Description

When using a monitor config that isn't at 0x0 offset, movewindow l/r seems to offset the window's y coordinate by the monitor's y offset. Same happens with u/d and the monitor's x offset.

### How to reproduce

1. Use a config with a monitor set to an offset other than 0x0.
2. Dispatch a movewindow event in a direction perpendicular to a direction the monitor has an offset.

### Crash reports, logs, images, videos

The video below is of a monitor with 1440x200 offset.
https://github.com/hyprwm/Hyprland/assets/1360103/cdce0212-6317-4120-b035-393c94fabdb5

PR #3520, opened 2023-10-08
fix extra monitor offset when snapping floating

MightyPlaza:
removes extra monitor offset being added in the non changed `vPos`

closes: https://github.com/hyprwm/Hyprland/issues/3519


vaxerski (approved):
thansk!

PR #3197, opened 2023-09-06
Groupbar aesthetic

MightyPlaza:
Improves / adds more customization to groupbar:

- adds `group` and `groupbar` sections to free up `misc`
- add bottom as bar position
- separates border and bar colors
- allows setting groupbar heigh (bar height/gradient height depending on mode)
- adds `internal_bar`, `font` and background color options
- fixes gradients not updating with config reload

Mostly ready now, just needs some minors fixes

example of new config:

```
group {
    col.border_active = rgba(bb66ffee)
    #col.border_inactive = 0xffff0000 
    #col.border_locked_active = 0xff00ffff 
    #col.border_locked_inactive = 0xff0000ff 
    #--moveintogroup_lock_check = false-- #removed due to another commit, maybe should be moved here
    #insert_after_current = true
    groupbar {
        enabled = true
        col.active = rgba(cdd6f4cc)
        col.inactive = rgb(666699)
        #col.locked_active = 0xfff00f0f 
        #col.locked_inactive = 0xff0ff00f 
        col.background = rgb(bb66ff)
        #render_titles = false 
        #scrolling = true
        max_size = 200
        titles_font_size = 8
        mode = 0 # 0 - bar, 1 - gradient 
        height = 18
        #internal_bar = false 
        font = JetBrainsMono Nerd Font Mono
        top = no
        text_color = rgb(333333)
        #unified_border = 1
    }
}
```

PR #3522, opened 2023-10-08
create group and groupbar config sections

MightyPlaza:
Moves group related variables to `group` and `group:groupbar` sections

haven't messed with `nogroup` related options but they probably should be moved in to here as well
(gradient hot reloading still doesn't work, but that is for another PR)

begins splitting commits from https://github.com/hyprwm/Hyprland/pull/3197

didn't find any missing instances with `grep -nr` and all changed options seem to work, so should cause any issues

wiki: https://github.com/hyprwm/hyprland-wiki/pull/366


vaxerski (approved):
lgtm

PR #3524, opened 2023-10-08
Avoid rounding errors in window position on dwindle

### LINKED ISSUE #3511: Uncommon resolution - bottom row of pixels not used for windows
### Hyprland Version

v0.30.0

### Bug or Regression?

Bug

### Description

On my laptops internal screen Hyprland doesn't use the bottom row of pixels to display windows. Instead only the wallpaper is shown there. If your wallpaper is black this doesn't really hurt (except that you loose a row of pixels), but if there is something colorful on your wallpaper (like our beloved hypr-chan) it gets annoying really fast, especially for fullscreen applications (e.g. when watching a movie, playing a game, etc.)

I am using Hyprland on a Framework Laptop 13 whose internal display has a quite uncommon resolution:
```
> hyprctl monitors
Monitor eDP-1 (ID 0):
	2256x1504@59.99900 at 0x0
	description: BOE 0x095F (eDP-1)
	make: BOE
	model: 0x095F
	serial:
	active workspace: 2 (2)
	special workspace: 0 ()
	reserved: 0 30 0 0
	scale: 1.50
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: 0
```

### How to reproduce

Use Hyprland on a similar display. Fractional scale doesn't matter (I have tried 1 and 1.5), neither does the type of application. Nothing gets displayed on the bottom row except the desktop wallpaper and the mouse (if I move it to the bottom). 
This bug isn't new (it exists since at least v0.27.2) and I was able to reproduce it on both Arch and NixOS.

I attached some pictures that I took with my smartphone (taking screenshots was not possible since grim doesn't capture those pixels either). As an example I just used alacritty 

IMG_20231006_201744_063
IMG_20231006_201758_102
IMG_20231006_201821_660
IMG_20231006_201827_269
IMG_20231006_201836_299
I took this last image just that you see that this line of pixels is from the hypr-chan background.

FakeMichau:
Avoids rounding errors in window position and size which can result in wallpaper being visible between windows when it shouldn't. Simply gives second children the rest of the space instead of calculating again.

Should resolve #3511 

It's ready

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -21,20 +21,18 @@ void SDwindleNodeData::recalcSizePosRecursive(bool force, bool horizontalOverrid
 
         if (SPLITSIDE) {
             // split left/right
-            int space_0 = size.x / 2 * splitRatio;
-            int space_1 = size.x - space_0;
+            int const FIRSTSIZE = size.x / 2 * splitRatio;
             children[0]->position = position;
-            children[0]->size     = Vector2D(space_0, size.y);
-            children[1]->position = Vector2D(position.x + space_0, position.y);
-            children[1]->size     = Vector2D(space_1, size.y);
+            children[0]->size     = Vector2D(FIRSTSIZE, size.y);
+            children[1]->position = Vector2D(position.x + FIRSTSIZE, position.y);
+            children[1]->size     = Vector2D(size.x - FIRSTSIZE, size.y);
         } else {
             // split top/bottom
-            int space_0 = size.y / 2 * splitRatio;
-            int space_1 = size.y - space_0;
+            int const FIRSTSIZE = size.y / 2 * splitRatio;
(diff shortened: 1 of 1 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -24,16 +21,20 @@ void SDwindleNodeData::recalcSizePosRecursive(bool force, bool horizontalOverrid
 
         if (SPLITSIDE) {
             // split left/right
+            int space_0 = size.x / 2 * splitRatio;
```
vaxerski, line 24: no snake pls, also can be const, in which case CAPS too

FakeMichau:
Just realized it's only for the dwindle layout but I've seen this issue on master as well. Will make a commit soon hopefully

vaxerski (approved):
lgtm

Issue #3491, opened 2023-10-04
Move Cursor to Active Window When Changing Workspaces


austinliuigi:
### Description

When changing focus through the `movefocus` dispatcher, the cursor moves to the center of the active window. I would like this to happen when changing workspaces as well, and in general whenever focus shifts from one window to another. Is this currently possible to set?

Thanks

pdamianik:
Sounds like a bug, because moving the cursor to a switched to workspace this is the behavior for me (and I don't remember changing anything related to moving the cursor).

<details><summary>My config</summary>
<p>

```
# This is an example Hyprland config file.
#
# Refer to the wiki for more information.

#
# Please note not all available settings / options are set here.
# For a full list, see the wiki
#

# See https://wiki.hyprland.org/Configuring/Monitors/
# laptop display
monitor=eDP-1,preferred,0x1050,1
# secondary 1680x1050 display
monitor=desc:N-Vision GM2200 100315290001,preferred,120x0,1
# any other hdmi external display (assumed to be the same size as the main display)
monitor=HDMI-A-1,preferred,0x0,auto
# any external display
monitor=,preferred,auto,1


# See https://wiki.hyprland.org/Configuring/Keywords/ for more

# Execute your favorite apps at launch
# exec-once = waybar & hyprpaper & firefox
exec-once = kanshi
exec-once = sleep 1 && swww init
exec-once = /usr/lib/kdeconnectd
exec-once = /usr/bin/kidex
exec-once = amberol
exec-once = swayosd-server
exec-once = udiskie

# Source a file (multi-file configs)
# source = ~/.config/hypr/myColors.conf

# Some default env vars.
env = XCURSOR_SIZE,24
env = QT_QPA_PLATFORMTHEME,wayland

# For all categories, see https://wiki.hyprland.org/Configuring/Variables/
input {
    kb_layout = us,at
    kb_variant = ,nodeadkeys
    kb_model =
    kb_options = grp:alt_altgr_toggle
    kb_rules =

    numlock_by_default = true

    repeat_delay = 250
    repeat_rate = 50

    follow_mouse = 2

    touchpad {
        natural_scroll = true
    }

    sensitivity = 0 # -1.0 - 1.0, 0 means no modification.
    accel_profile = flat,
}

general {
    # See https://wiki.hyprland.org/Configuring/Variables/ for more

    gaps_in = 8
    gaps_out = 15
    border_size = 1
[...]

PR #3528, opened 2023-10-08
Add option to disable centering on workspace when switching workspace

pdamianik:
Adds an option to disable centering on workspace when switching between workspaces with the `workspace` dispatcher (non breaking - default value is current behavior). See #3491.

I also switched the calculation for the window center in `swapActive()` to use the `middle()` method, as it seems like using the current position (I assume this includes animations) seems to be a mistake.

I have to test it and add a wiki entry.


pdamianik:
Tested

pdamianik:
Wiki pr: https://github.com/hyprwm/hyprland-wiki/pull/368

vaxerski:
shouldn't that go into binds:?

Issue #3529, opened 2023-10-08
movewindoworgroup bypasses lockgroups


BluewyDiamond:
### Hyprland Version

Hyprland, built from branch main at commit 728a8bb48e0f7de1cbe1ad13fb469754c3d0bc97 dirty (config: mention force_default_wallpaper in default configs). Tag: v0.30.0-67-g728a8bb4

### Bug or Regression?

Bug

### Description

Toggling `lockgroups` does not make it immune to movewindoworgroup.

### How to reproduce

Toggling `lockgroups -> on` and then move a window using `movewindoworgroup` to the direction where the group is, will place the window into the group.

### Crash reports, logs, images, videos

![2023-10-09T01:55:28,553402787+02:00](https://github.com/hyprwm/Hyprland/assets/115411576/27492baf-4b9d-4500-a680-aba1a92f0ebd)
![2023-10-09T01:55:31,726617328+02:00](https://github.com/hyprwm/Hyprland/assets/115411576/bef3416f-ac77-430e-abb6-5c1340d0a907)

BluewyDiamond:
`hyprctl dispatch lockgroups lock` instead of `on` also has the same behaviour.

memchr:
It doesn't check the global lock (should it check or not?)

https://github.com/hyprwm/Hyprland/blob/499df49f7b28f9e3be9b1c53843fd5c465dec60b/src/managers/KeybindManager.cpp#L2054-L2069

Try lockactivegroup?

BluewyDiamond:
I just tried `lockactivegroup` and it works, but since the wiki says "Locks the groups (all groups will not accept new windows)", I feel like `lockgroups` should also do the same for all groups, otherwise it would be confusing I think.

memchr:
moveintogroup doesn't check global locks either, this has been the case for a long time. I assume the original intent for this lock is just to prevent new windows from being mapped into a group.

Does it need to be changed?

BluewyDiamond:
I recommend implementing changes to enhance consistency in the functionality. Specifically, it should be modified to verify global locks. Additionally, it's worth noting that 'lockgroups' lacks the visual indicator that 'lockactivegroup' provides when activated, and it does not restrict moving the window out of the group, a behavior that 'lockactivegroup' does exhibit. Therefore, I propose making adjustments to ensure a more consistent behavior across these features.

BluewyDiamond:
summarizing:
[...]

PR #3531, opened 2023-10-09
fix(group): ensure consistency in dispatcher behavior with global group lock

memchr:
The following dispatchers will more consistently respect the global group lock and `binds:ignore_group_lock`. 

- moveintogroup
- moveoutofgroup
- movewindoworgroup

---

close: #3529


vaxerski (approved):
lgtm

PR #2734, opened 2023-07-18
Update subprojects/wlroots

jbeich:
For wlroots debugging by comparing with `sway-git`. Briefly tested [downstream](https://github.com/freebsd/freebsd-ports/blob/2117c1cce9fc/x11-wm/hyprland/files/patch-wlroots-0.17).

Based on:
- https://github.com/swaywm/sway/pull/7637
- https://github.com/swaywm/sway/pull/7675

PR #3532, opened 2023-10-09
Fix build with latest wlroots

h0tc0d3:
Fix buils with latest wlroots.

- Drop KDE idle protocol support
- output_event_commit: Replace committed and buffer with state struct

Ready.


vaxerski:
https://github.com/hyprwm/Hyprland/pull/2734#pullrequestreview-1534706147

PR #3543, opened 2023-10-10
Better versioning

vaxerski:
moves versioning to its own script and file

see `scripts/generateVersion.sh` and `src/version.h.in`

@fufexan for nix/meson

Makes the version propagate to the headers installed with pluginenv

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ gmon.out
 *.tar.gz
 
 PKGBUILD
+
+src/version.h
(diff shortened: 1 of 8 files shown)


NotAShelf:
looks good

outfoxxed:
Looks good as long as GIT_COMMIT_HASH is also stored in a constant so it can be compared. I haven't checked recently but I think this was the case.

vaxerski:
> Looks good as long as GIT_COMMIT_HASH is also stored in a constant so it can be compared. I haven't checked recently but I think this was the case.

it isn't, we'll have to add that to the API.

Issue #3538, opened 2023-10-09
Request: match rotation between outputs and mapped inputs


Smona:
### Description

After the acceptance of #996 and #3220, Hyprland is _almost_ a first class experience on 2-in-1s! At this point, there is only one thing keeping me from switching entirely to Hyprland on my 2-in-1 laptop.

In Sway, it is possible to map inputs to outputs with the `map_to_output` message. This causes inputs to automatically rotate when their mapped output does, ensuring touch inputs always map correctly to dynamically rotated outputs (e.g. when using [rot8](https://github.com/efernau/rot8)). Hyprland also supports mapping inputs to outputs via the `inputs.touchdevice.output` keyword (and apparently maps inputs automatically when it can), but in my experience this does not lock the transform of the input to the output.

It's important for hyprland to support this in my opinion, because AFAICT there is no way for third-party programs to control the transform of existing input devices via a wayland protocol, so at the moment users will always have to implement their own, hyprland specific scripts to achieve what should be a universally desirable behavior: touchscreens inputting where the user's finger is, regardless of dynamic rotation.

Since there is already a `transform` option for inputs, perhaps the right way to implement this would be to _add_ any output rotation changes to the user-configured input rotation.

pdamianik:
Same as last time, I have no 2in1 so I can't test #3544, but this should be all to get it working. As for the remaining items, nothing major is open, just a implementation detail so this should be the final state of the pr.

Smona:
Thank you @pdamianik! I'm traveling at the moment so I don't have my 2-in-1 with me, but I will test this as soon as I get back :slightly_smiling_face:

pdamianik:
@Smona I assume you aren't subscribed to #3544 and didn't get a notification for my latest changes. If you did and just didn't have time to check them out then feel free to ignore this reminder.

Smona:
Thanks for the reminder! This is still on my radar but life got very busy recently. I should be able to re-test this weekend.

pdamianik:
If you are occupied it is always ok to leave this issue open. I was just worried you might be waiting for further changes from my side.

PR #3544, opened 2023-10-10
Map touch devices and tablets bound to an output

pdamianik:

tries to fix: #3538.

This transfers the rotation for an output to any bound touch devices if the `transform` configuration hasn't been set explicitly.

Not tested yet (I don't have a 2in1).

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -64,7 +64,7 @@ void CMonitor::onConnect(bool noRule) {
         createdByUser = true; // should be true. WL, X11 and Headless backends should be addable / removable
 
     // get monitor rule that matches
-    SMonitorRule monitorRule = g_pConfigManager->getMonitorRuleFor(output->name, output->description ? output->description : "");
+    SMonitorRule monitorRule = g_pConfigManager->getMonitorRuleFor(*this);
 
     // if it's disabled, disable and ignore
     if (monitorRule.disabled) {
@@ -449,7 +449,7 @@ void CMonitor::setMirror(const std::string& mirrorOf) {
         pMirrorOf = nullptr;
 
         // set rule
-        const auto RULE = g_pConfigManager->getMonitorRuleFor(this->szName, this->output->description ? this->output->description : "");
+        const auto RULE = g_pConfigManager->getMonitorRuleFor(*this);
 
         vecPosition = RULE.offset;
 
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -142,6 +142,20 @@ void CMonitor::onConnect(bool noRule) {
     if (!noRule)
         g_pHyprRenderer->applyMonitorRule(this, &monitorRule, true);
 
+    for (const auto& PTOUCHDEV : g_pInputManager->m_lTouchDevices) {
+        if (matchesStaticSelector(PTOUCHDEV.boundOutput)) {
+            Debug::log(LOG, "Binding touch device {} to output {}", PTOUCHDEV.name, szName);
+            wlr_cursor_map_input_to_output(g_pCompositor->m_sWLRCursor, PTOUCHDEV.pWlrDevice, output);
+        }
+    }
+
+    for (const auto& PTABLET : g_pInputManager->m_lTablets) {
+        if (matchesStaticSelector(PTABLET.boundOutput)) {
+            Debug::log(LOG, "Binding tablet {} to output {}", PTABLET.name, szName);
+            wlr_cursor_map_input_to_output(g_pCompositor->m_sWLRCursor, PTABLET.wlrDevice, output);
+        }
+    }
+
     if (!state.commit())
         Debug::log(WARN, "wlr_output_commit_state failed in CMonitor::onCommit");
 
@@ -341,6 +355,19 @@ void CMonitor::addDamage(const CBox* box) {
 
 bool CMonitor::isMirror() {
     return pMirrorOf != nullptr;
+}
+
+bool CMonitor::matchesStaticSelector(const std::string& selector) const {
+    if (selector.starts_with("desc:")) {
+        // match by description
+        const auto DESCRIPTIONSELECTOR = selector.substr(5);
+        const auto DESCRIPTION         = removeBeginEndSpacesTabs(szDescription.substr(0, szDescription.find_first_of('(')));
+
+        return DESCRIPTIONSELECTOR == szDescription || DESCRIPTIONSELECTOR == DESCRIPTION;
+    } else {
+        // match by selector
+        return szName == selector;
+    }
 }
 
 int CMonitor::findAvailableDefaultWS() {
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1495,9 +1495,10 @@ void CInputManager::setTouchDeviceConfigs(STouchDevice* dev) {
             if (libinput_device_config_calibration_has_matrix(LIBINPUTDEV))
                 libinput_device_config_calibration_set_matrix(LIBINPUTDEV, MATRICES[ROTATION]);
 
-            auto output = g_pConfigManager->getDeviceString(PTOUCHDEV->name, "output", "input:touchdevice:output");
-            bool bound  = !output.empty() && output != STRVAL_EMPTY;
-            if (!bound) {
+            auto       output     = g_pConfigManager->getDeviceString(PTOUCHDEV->name, "output", "input:touchdevice:output");
+            bool       bound      = !output.empty() && output != STRVAL_EMPTY;
+            const bool AUTODETECT = output == "[[Auto]]";
+            if (!bound && AUTODETECT) {
                 const auto DEFAULTOUTPUT = wlr_touch_from_input_device(PTOUCHDEV->pWlrDevice)->output_name;
                 if (DEFAULTOUTPUT) {
                     output = DEFAULTOUTPUT;
@@ -1509,7 +1510,7 @@ void CInputManager::setTouchDeviceConfigs(STouchDevice* dev) {
             if (PMONITOR) {
                 Debug::log(LOG, "Binding touch device {} to output {}", PTOUCHDEV->name, PMONITOR->szName);
                 wlr_cursor_map_input_to_output(g_pCompositor->m_sWLRCursor, PTOUCHDEV->pWlrDevice, PMONITOR->output);
-            } else
+            } else if (bound)
                 Debug::log(ERR, "Failed to bind touch device {} to output '{}': monitor not found", PTOUCHDEV->name, output);
         }
     };
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1491,14 +1491,26 @@ void CInputManager::setTouchDeviceConfigs(STouchDevice* dev) {
                 libinput_device_config_send_events_set_mode(LIBINPUTDEV, mode);
 
             const int ROTATION = std::clamp(g_pConfigManager->getDeviceInt(PTOUCHDEV->name, "transform", "input:touchdevice:transform"), 0, 7);
+            Debug::log(LOG, "Setting calibration matrix for device {}", PTOUCHDEV->name);
             if (libinput_device_config_calibration_has_matrix(LIBINPUTDEV))
                 libinput_device_config_calibration_set_matrix(LIBINPUTDEV, MATRICES[ROTATION]);
 
-            const auto OUTPUT = g_pConfigManager->getDeviceString(PTOUCHDEV->name, "output", "input:touchdevice:output");
-            if (!OUTPUT.empty() && OUTPUT != STRVAL_EMPTY)
-                PTOUCHDEV->boundOutput = OUTPUT;
-            else
-                PTOUCHDEV->boundOutput = "";
+            auto output = g_pConfigManager->getDeviceString(PTOUCHDEV->name, "output", "input:touchdevice:output");
+            bool bound  = !output.empty() && output != STRVAL_EMPTY;
+            if (!bound) {
+                const auto DEFAULTOUTPUT = wlr_touch_from_input_device(PTOUCHDEV->pWlrDevice)->output_name;
+                if (DEFAULTOUTPUT) {
+                    output = DEFAULTOUTPUT;
+                    bound  = true;
+                }
+            }
+            PTOUCHDEV->boundOutput = bound ? output : "";
+            const auto PMONITOR    = bound ? g_pCompositor->getMonitorFromName(output) : nullptr;
+            if (PMONITOR) {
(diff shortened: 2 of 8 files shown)


vaxerski:
now we are restricting the user's freedom to override the monitor transform, though.

pdamianik:
> now we are restricting the user's freedom to override the monitor transform, though.

What specifically has changed? Also this doesn't make it possible to change the monitor transform, just copies the transform value to any bound touch devices.

vaxerski:
but if a user wants to have transform 2 on a monitor with transform 3, they cant

pdamianik:
If you mean transform 2 for a touchdevice mapped to a monitor with transform 3 enabled, then that's doable, you just have to explicitly set the `transform` configuration value for the touch device. This pr only copies the monitors rotational value when the device specific configuration value is **not** set and the touch device is bound to an output (https://github.com/hyprwm/Hyprland/pull/3544/files#diff-2ba0a4b3fea7c686b5bcf66318fa5912cb5d78a313318bf7ce2d51faf3b7a88aR1397). Also the monitors transform value can still be set without any restrictions.

pdamianik:
Now this also considers whether the global `input:touchdevice:transform` was set explicitly (https://github.com/hyprwm/Hyprland/pull/3544/files#diff-2ba0a4b3fea7c686b5bcf66318fa5912cb5d78a313318bf7ce2d51faf3b7a88aR1400).

src/managers/input/InputManager.cpp
```diff
@@ -1390,15 +1390,27 @@ void CInputManager::setTouchDeviceConfigs(STouchDevice* dev) {
...
-            if (libinput_device_config_calibration_has_matrix(LIBINPUTDEV))
-                libinput_device_config_calibration_set_matrix(LIBINPUTDEV, MATRICES[ROTATION]);
-
-            const auto OUTPUT = g_pConfigManager->getDeviceString(PTOUCHDEV->name, "output", "input:touchdevice:output");
-            if (!OUTPUT.empty() && OUTPUT != STRVAL_EMPTY)
-                PTOUCHDEV->boundOutput = OUTPUT;
+            const auto OUTPUT             = g_pConfigManager->getDeviceString(PTOUCHDEV->name, "output", "input:touchdevice:output");
+            const auto GTRANSFORMSET      = g_pConfigManager->getConfigValuePtr("input:touchdevice:transform")->set;
+            const auto DEVICEOUTPUTSET    = g_pConfigManager->deviceConfigExists(PTOUCHDEV->name, "output");
+            const auto DEVICETRANSFORMSET = g_pConfigManager->deviceConfigExists(PTOUCHDEV->name, "transform");
```
vaxerski, line 1396: why not use the same `->set` flag here? instead of the entire change to how `deviceConfigExists` works?

pdamianik: 1. There is no `getConfigValuePtr()` equivalent for device configurations
2. I think it would be useful to make `deviceConfigExists()` optionally be able to check for specific configuration keys.

vaxerski: 1. getDeviceConfig()->set
2. I don't think so. If a device config exists, all vars belonging to it exist

pdamianik: 1. I cannot find a function named `getDeviceConfig()`, there are `getConfigValueSafeDevice()` which returns the struct or not a pointer and `getConfigValuePtrSafe()` which first checks whether the key is for a device and then has to parse the device identifier first. As an alternative to changing the `deviceConfigExists()` function a  `getDeviceConfigValuePtr()` function could be extracted from `getConfigValuePtr()`.
2. Yes, but it is not clear whether a config key has been set by the user.

vaxerski: Again, you can get the struct. We cannot receive ptrs to device configs as they are not static IIRC.

What is your problem? It returns the struct. The struct has a `set` property.

pdamianik: My bad, can't use `getConfigValueSafeDevice()` as it isn't even public and `getConfigValuePtrSafe()` can return a pointer to a device config value. So another alternative besides creating `getDeviceConfigValuePtr()` would be making `getConfigValueSafeDevice()` public. I will add two more commits to give you the choice between all proposed ways to check if a device config value exists and then you can reset to the one you wish to use and merge this pr.

pdamianik: Now there are three commits:
 1. cec75926ad24d571457d01bf9a478ada2d1c39b2 still uses the modified `deviceConfigExists()`
 2. 0fef48b71c022d3e829c138776727d99f3f0f17f extracts the function `getDeviceConfigValuePtr()` from `getConfigValuePtrSafe()`
 3. 4be9260c8dc09373a55e8476e64b7d632ed22102 makes `getConfigValueSafeDevice()` public

vaxerski (approved):
lgtm, thanks!

src/managers/input/InputManager.cpp
vaxerski: final print of the MR looks good to me.

I squash everything anyways

If there isnt anything else, I am alright with merging this as-is

pdamianik: Just a functionality test by @Smona

Smona:
Hey @pdamianik, unfortunately this doesn't appear to work for me. I've tried 4be9260c8dc09373a55e8476e64b7d632ed22102 without any explicit input mapping, and with both `input.touchdevice.output = "eDP-1"` and `device:wacom-hid-48eb-finger.output = "eDP-1"`. I'm not sure how to confirm that the input is being mapped correctly, but for reference here is my relevant input/output config:

```
$ hyprctl monitors

Monitor eDP-1 (ID 0):
	1920x1200@59.95000 at 0x0
	description: Sharp Corporation 0x14AF (eDP-1)
	make: Sharp Corporation
	model: 0x14AF
	serial: 
	active workspace: 2 (2)
	special workspace: 0 ()
	reserved: 0 0 0 27
	scale: 1.00
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: 0
	activelyTearing: false

$ hyprctl devices

...

Tablets:
	Tablet at 282da10:
		wacom-hid-48eb-pen

Touch:
	Touch Device at 26e7b10:
		wacom-hid-48eb-finger

...
```

pdamianik:
Does the touchscreen behave correctly when you set the `input:tablet:transform` property?

pdamianik:
@Smona just let me know if you wanna switch to [`v0.31.0`](https://github.com/hyprwm/Hyprland/tree/v0.31.0), I have the rebase ready.

Smona:
No luck with the pen either, with or without `device:wacom-hid-48eb-pen:output = "eDP-1"`. Interestingly, `input:tablet:output = "eDP-1"` and `input:tablet:transform = 1` both give me configuration errors that the field `tablet:output/transform` doesn't exist. that section seems to mess up parsing my config no matter where i put it on this commit :shrug:.

Setting the transform manually does work for both pen and finger input via `input:touchdevice:transform` and `device:wacom-hid-48eb-pen:transform`.

As for the rebase, go for it as long as it won't muddy up your debugging! I've found hyprland is pretty quick to compile and i'm on nixos so it's easy to switch ^ ^

pdamianik:
> As for the rebase, go for it as long as it won't muddy up your debugging! I've found hyprland is pretty quick to compile and i'm on nixos so it's easy to switch ^ ^

It's not really a problem, I just wanted to give you the option to update your `hyprlandd.conf` because `v0.31.0` has breaking config changes.

Theoretically the current changes should solve your problem, but since they don't I will have to add some `log` statements and have you capture a log of rotating the monitor. The fact that the setting the config value manually solves the issue and the monitor rotates when rotating the device tells me that everything around this pr works as expected. Also I am guessing that the mouse doesn't do anything special about rotated outputs (you wouldn't want your mouse to be flipped when switching to a rotated monitor :wink:).

pdamianik:
Allright, this version should now do effectively the same as https://github.com/swaywm/sway/blob/master/sway/input/seat.c#L708 (minus selecting the builtin monitor for builtin devices by default plus mapping to monitor AND region at the same time).

Smona:
@pdamianik still no luck with [688e463](https://github.com/hyprwm/Hyprland/pull/3544/commits/688e4637b6440291e1c047e512a5482968681e60) unfortunately. it seems to be behaving exactly the same as before, both with and without explicit output mapping. I would post logs but I'm having trouble tracking them down.

pdamianik:
Alright, that means I will have to deep dive into wl_roots. That could take some time, but if it works on sway and just calls some wl_roots functions (from what I've seen so far), ideally we'd want to match what works (just let wl_roots handle it)

> I would post logs but I'm having trouble tracking them down.

I haven't added logs for now, because the current version calls the same wl_roots functions as sway does, which just tells wl_roots that an input should be mapped to an output, so I can't add any log statements for any events.

pdamianik:
Sorry for the long wait. I now know how libinput events go through wlroots and finally get to Hyprland. Wlr should handle input transforms related to outputs when the input has been bound to an output with `wlr_cursor_map_input_to_output`, as this pr adds in [src/managers/input/InputManager.cpp#1495](https://github.com/hyprwm/Hyprland/pull/3544/files#diff-2ba0a4b3fea7c686b5bcf66318fa5912cb5d78a313318bf7ce2d51faf3b7a88aR1495). This should work similarly to the previous changes in 688e463, so I am not sure if this will work, but I have added some log statements and hope for more info from the logs. Also I discovered [`evemu-record`](https://man.archlinux.org/man/extra/evemu/evemu-record.1.en), which can be used to record input events from a device and this capture can be replayed on any device, regardless of whether it has the same device or not, with [`evemu-play`](https://man.archlinux.org/man/extra/evemu/evemu-play.1.en). So please a capture some touch inputs, so that I can debug the issue on my device.

pdamianik:
Also for getting the logs there is [this](http://wiki.hyprland.org/Contributing-and-Debugging/#logs-dumps-etc) wiki page. TLDR you can find the log file at `/tmp/hypr/$HYPRLAND_INSTANCE_SIGNATURE/hyprland.log` (or `/tmp/hypr/$HYPRLAND_INSTANCE_SIGNATURE/hyprlandd.log` when running a dev build).

Smona:
@pdamianik this one's a winner! I was able to get both finger and stylus input rotating with the output, although I did have to make sure `input:touchdevice:output` and `input:tablet:output` were explicitly set for each, respectively. I had to do the same in my sway config so this is more than acceptable :partying_face:

It didn't work at first, but after figuring out how to enable stdout logs i was able to see that the quotations I had put around the monitor name in my config when setting the output were being included in the name hyprland was looking for :sweat_smile:. once I removed the quotes everything works great. That very well may have been impacting our previous tests FWIW.... just goes to show the importance of good logs!

pdamianik:
@vaxerski
Currently setting both a bound `output` and a `transform` value interact as follows: a bound `output` settings gets applied in wlroots, whereas a transformation matrix gets applied in libinput, which means they should stack (not tested), e.g. when setting a 90° rotated output as the bound output and setting a 90° transformation matrix, the input should get rotated by 180°.
Also, any unbound touchdevice will try to bind to its default output as assigned by wlroots. At the moment, the only option for ignoring a default outputs rotation would be to bind the device to an unknown output. I could add a specific `[[unbound]]` value for this case, because otherwise users would get error logs about unknown outputs. Either way, I will add a comment for "unbinding" default bound touchdevices to the wiki but I want to hear your thoughts first.

src/helpers/Monitor.cpp
```diff
@@ -343,6 +357,30 @@ bool CMonitor::isMirror() {
     return pMirrorOf != nullptr;
 }
 
+bool CMonitor::matchesString(const std::string& name) {
```
vaxerski, line 360: if you're adding this helper here, it'd be nice to use it in appropriate places like e.g. `CCompositor::getMonitorFromString`, otherwise they can easily fall apart later

src/managers/input/InputManager.cpp
```diff
@@ -1529,8 +1542,12 @@ void CInputManager::setTabletConfigs() {
             const auto OUTPUT   = g_pConfigManager->getDeviceString(t.name, "output", "input:tablet:output");
             const auto PMONITOR = g_pCompositor->getMonitorFromString(OUTPUT);
             if (!OUTPUT.empty() && OUTPUT != STRVAL_EMPTY && PMONITOR) {
+                Debug::log(LOG, "Binding tablet {} to output {}", t.name, PMONITOR->szName);
                 wlr_cursor_map_input_to_output(g_pCompositor->m_sWLRCursor, t.wlrDevice, PMONITOR->output);
                 wlr_cursor_map_input_to_region(g_pCompositor->m_sWLRCursor, t.wlrDevice, nullptr);
+                t.boundOutput = OUTPUT;
+            } else {
```
vaxerski, line 1549: style: no {} around short ifs

src/managers/input/InputManager.cpp
```diff
@@ -1491,14 +1491,27 @@ void CInputManager::setTouchDeviceConfigs(STouchDevice* dev) {
...
+                    output = DEFAULTOUTPUT;
+                    bound  = true;
+                }
+            }
+            PTOUCHDEV->boundOutput = bound ? output : "";
+            const auto PMONITOR    = bound ? g_pCompositor->getMonitorFromName(output) : nullptr;
+            if (PMONITOR) {
+                Debug::log(LOG, "Binding touch device {} to output {}", PTOUCHDEV->name, PMONITOR->szName);
+                wlr_cursor_map_input_to_output(g_pCompositor->m_sWLRCursor, PTOUCHDEV->pWlrDevice, PMONITOR->output);
+            } else {
```
vaxerski, line 1512: style: no {}

src/helpers/Monitor.cpp
```diff
@@ -142,6 +142,20 @@ void CMonitor::onConnect(bool noRule) {
     if (!noRule)
         g_pHyprRenderer->applyMonitorRule(this, &monitorRule, true);
 
+    for (const auto& PTOUCHDEV : g_pInputManager->m_lTouchDevices) {
+        if (this->matchesString(PTOUCHDEV.boundOutput)) {
```
vaxerski, line 146: style: we don't use `this->`

pdamianik:
`input:touchdevice:output` now gets set to `[[Auto]]` per default which configures autodetection from wlroots. Unbinding default outputs can therefore be done by setting the output to an empty string or the `[[Empty]]` value.

vaxerski (approved):
:car:

pdamianik:
The related wiki PR hyprwm/hyprland-wiki#492 wasn't merged before the `0.36.0` release.

vaxerski:
my bad

vaxerski:
backported

Smona:
Thank you so much @pdamianik for bearing with me through the testing and landing this seemingly niche feature!! I know artists will appreciate it, and i think this also unblocks experimenting with hyprland on pure touch devices like tablets. hyprland ftw!

PR #3545, opened 2023-10-10
Plugin API: Add version query

vaxerski:
I know `hyprctl version` exists but this is better.

cc @fufexan @outfoxxed @Duckonaut RFC

### DIFF
--- a/src/plugins/PluginAPI.cpp
+++ b/src/plugins/PluginAPI.cpp
@@ -277,7 +277,7 @@ APICALL std::vector<SFunctionMatch> HyprlandAPI::findFunctionsByName(HANDLE hand
     const auto FPATH = std::filesystem::canonical(exe);
 #elif defined(__OpenBSD__)
     // Neither KERN_PROC_PATHNAME nor /proc are supported
-    const auto FPATH = std::filesystem::canonical("/usr/local/bin/Hyprland");
+    const auto        FPATH            = std::filesystem::canonical("/usr/local/bin/Hyprland");
 #else
     const auto FPATH = std::filesystem::canonical("/proc/self/exe");
 #endif
@@ -337,4 +337,13 @@ APICALL std::vector<SFunctionMatch> HyprlandAPI::findFunctionsByName(HANDLE hand
     }
 
     return matches;
+}
+
+APICALL SVersionInfo HyprlandAPI::getHyprlandVersion(HANDLE handle) {
+    auto* const PLUGIN = g_pPluginSystem->getPluginByHandle(handle);
+
+    if (!PLUGIN)
+        return {};
+
+    return {GIT_COMMIT_HASH, GIT_TAG, GIT_DIRTY != std::string(""), GIT_BRANCH, GIT_COMMIT_MESSAGE};
 }
--- a/src/plugins/PluginAPI.hpp
+++ b/src/plugins/PluginAPI.hpp
@@ -265,3 +265,13 @@ namespace HyprlandAPI {
     */
     APICALL SVersionInfo getHyprlandVersion(HANDLE handle);
 };
+
+/*
+    Get the hash this plugin/server was compiled with.
+
+    This function will end up in both hyprland and any/all plugins,
+    and can be found by a simple dlsym()
+*/
+APICALL inline EXPORT std::string __hyprland_api_get_hash() {
+    return GIT_COMMIT_HASH;
+}
--- a/src/plugins/PluginAPI.hpp
+++ b/src/plugins/PluginAPI.hpp
@@ -272,6 +272,6 @@ namespace HyprlandAPI {
     This function will end up in both hyprland and any/all plugins,
     and can be found by a simple dlsym()
 */
-APICALL inline EXPORT std::string __hyprland_api_get_hash() {
+APICALL inline EXPORT const char* __hyprland_api_get_hash() {
     return GIT_COMMIT_HASH;
 }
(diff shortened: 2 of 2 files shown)


fufexan (dismissed):

Duckonaut:
Very very good to have. I won't need to parse shell command output querying the state of the program I'm using, was a really roundabout way to do it. Info seems well organized, happy that ignoring the `dirty` tag is easier here as well. I'll make hyprload rely on this as soon as a versioned release with this exists, don't want to break compatibility for non-git users

vaxerski:
> I'll make hyprload rely on this as soon as a versioned release with this exists, don't want to break compatibility for non-git users

I'd still gatekeep it under some #ifdef for people on < 0.31

outfoxxed:
Looks good to me. It might also be a good idea to have a standard check in place you can opt out of that would show a hyprland notification along the lines of "plugin {} was compiled for a different hyprland version and could not be loaded." (implemented in the header so its compiled on the plugin side)

vaxerski:
Plugins have to do that themselves, I'll gatekeep all the official plugins with that once this is merged and versioned like duckonaut.

outfoxxed:
Do you have any reason not to implement a standard check for it in the header?

vaxerski:
if you have an idea on how to do that, feel free

outfoxxed:
You implement the check in the header so it gets compiled plugin side, and uses the GIT_COMMIT_HASH of the headers its compiling against. I can implement it later if you want.

vaxerski:
how do I make it not compile hl-side though?

outfoxxed:
have a plugin side api header, in that header define a function to get the commit it was compiled against. it dosen't matter if its compiled hyprland side because you're only checking the result from the function in the plugin.

Duckonaut:
> I'd still gatekeep it under some #ifdef for people on < 0.31

Yeah, I could do that as well, i'll just need to tweak the build system to handle that.

vaxerski:
fair point, I'll see tomorrow if I can do it

vaxerski:
maybe not very tomorrow but have a look guys :)

Duckonaut:
For maximum compatibility, maybe it could be useful to return this as a C str just using `.c_str()` for a stable ABI? Since it will be called across plugin boundaries. Not strictly necessary since i mean, most of hyprland doesn't care for stable ABI and returns `std::string`s in lots of places. Either way is cool tho, good to have

vaxerski:
good point.

outfoxxed:
I think it would be useful to have hyprland compare the result of __hyprland_api_get_hash by default (maybe some kind of opt out) and show an error via the notification system that the plugin couldn't be loaded if it mismatches.

vaxerski:
that can be done in a follow-up

outfoxxed:
yeah otherwise LGTM

vaxerski:
although I am opposed to that

outfoxxed:
> although I am opposed to that

I feel like an opt-out makes sense because most plugins have to touch a lot more than the "HyprlandAPI" header surface to do anything useful. What's your reasoning?

vaxerski:
many don't break between rebuilds, like e.g. csgo-vulkan-fix that had been working at one point for 4 months without an issue for me.

it'd be better to make it a _warning_ rather than error

outfoxxed:
Yeah I'm not opposed to that, I just want there to be a visual indication something is/may be wrong by default, because a ton of people get ABI breakage all the time with hy3.

vaxerski:
why don't they just use hyprload :)

outfoxxed:
its been broken with hy3 for quite a while (since the pkg config change). I think echoduck fixed it like a couple days ago but I haven't checked yet.

Duckonaut:
Yeah it wasn't properly serving its own headers ever since the pkgconfig change, in the meantime I think literally every plugin moved to pkg-config so now that I made it serve headers via that, it works well again. It still always worked for downloading and updating plugins, just didn't actually help with building them correctly since May

vaxerski:
since fuffie is dead I will just merge it

memchr:
Meson breaks down magnificently. 
Did you forget to include `version.h` in `src/plugins/PluginAPI.hpp`?

vaxerski:
all checks passed tho? wot

memchr:
```
../src/plugins/PluginAPI.hpp: In function ‘const char* __hyprland_api_get_hash()’:
../src/plugins/PluginAPI.hpp:276:12: error: ‘GIT_COMMIT_HASH’ was not declared in this scope
  276 |     return GIT_COMMIT_HASH;
      |            ^~~~~~~~~~~~~~~
../src/plugins/PluginAPI.hpp:276:12: note: the macro ‘GIT_COMMIT_HASH’ had not yet been defined
In file included from ../src/plugins/../helpers/../macros.hpp:17,
                 from ../src/plugins/../helpers/Vector2D.hpp:5,
                 from ../src/plugins/../includes.hpp:148,
                 from ../src/plugins/../defines.hpp:1,
                 from ../src/plugins/../Compositor.hpp:7,
                 from ../src/plugins/PluginAPI.cpp:2:
../src/plugins/../helpers/../version.h:2: note: it was later defined here
    2 | #define GIT_COMMIT_HASH    "21b5cf402afb8a72ffbf12efdd0a0d46849d86dd"
      |
```
From gcc, `meson setup build` then `ninja -C build`

memchr:
This will also cause plugin build to fail, so I think it's safe to say that include version.h is needed.

Issue #2595, opened 2023-06-27
Add background color options for groupbar and gradients


hmanhng:
I use swww dynamic wallpaper, so the groupbar is relatively hard to see. i don't see any option to change background color. Can add them?

MightyPlaza:
Last ones
http://wiki.hyprland.org/Configuring/Variables/#misc

hmanhng:
> Last ones http://wiki.hyprland.org/Configuring/Variables/#misc

what i want is background color, not text color.

MightyPlaza:
The options beginning in col.group then
http://wiki.hyprland.org/Configuring/Variables/#general

hmanhng:
it's just the border, not the background sir

MightyPlaza:
Hyprland restart is required for some reason, but it changes the gradient.

Aqa-Ib:
> Hyprland restart is required for some reason, but it changes the gradient.

What is the name of the option that changes the group gradient color? thanks.

MightyPlaza:
col.group_...

Aqa-Ib:
> col.group_...

Alright thanks.
I think this issue might be renamed to: recolor the groupbar gradient without having to restart hyprland.
Cheers.

Aqa-Ib:
Is this really fixed? I can not find the way to put a background to the group bar. Thank you all.

Aqa-Ib:
This one is not working at the moment, consider reopening.

Aqa-Ib:
I would gladly git bisect it, but I am having problems trying to build past commits because of subprojects.

vaxerski:
`g reset --hard --recurse-submodules`

MightyPlaza:
> I would gladly git bisect it, but I am having problems trying to build past commits because of subprojects.

https://github.com/hyprwm/Hyprland/issues/4081#issuecomment-1847201757
I will try to fix it when I get the time

MightyPlaza:
@vaxerski fixed now

PR #3546, opened 2023-10-10
allow reload and fix locked groupbar gradient

MightyPlaza:
allows groupbar gradient hot reloading
also fixes gradients for locked groups

closes: https://github.com/hyprwm/Hyprland/issues/2595

not sure if the best way to implement this but shouldn't cause any other issues

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1,6 +1,8 @@
 #include "ConfigManager.hpp"
 #include "../managers/KeybindManager.hpp"
 
+#include "../render/decorations/CHyprGroupBarDecoration.hpp"
+
 #include <string.h>
 #include <string>
 #include <sys/stat.h>
@@ -79,7 +81,7 @@ void CConfigManager::setDefaultVars() {
     configValues["general:no_border_on_floating"].intValue = 0;
     configValues["general:gaps_in"].intValue               = 5;
     configValues["general:gaps_out"].intValue              = 20;
-    configValues["general:gaps_workspaces"].intValue        = 0;
+    configValues["general:gaps_workspaces"].intValue       = 0;
     ((CGradientValueData*)configValues["general:col.active_border"].data.get())->reset(0xffffffff);
     ((CGradientValueData*)configValues["general:col.inactive_border"].data.get())->reset(0xff444444);
     ((CGradientValueData*)configValues["general:col.nogroup_border"].data.get())->reset(0xff444444);
@@ -1667,6 +1669,9 @@ void CConfigManager::loadConfigLoadVars() {
         ensureMonitorStatus();
         ensureVRR();
     }
+
+    if (!isFirstLaunch && !g_pCompositor->m_bUnsafeState)
+        refreshGroupBarGradients();
 
     // Updates dynamic window and workspace rules
     for (auto& w : g_pCompositor->m_vWindows) {
--- a/src/render/decorations/CHyprGroupBarDecoration.cpp
+++ b/src/render/decorations/CHyprGroupBarDecoration.cpp
@@ -6,6 +6,9 @@
 // shared things to conserve VRAM
 static CTexture m_tGradientActive;
 static CTexture m_tGradientInactive;
+static CTexture m_tGradientLockedActive;
+static CTexture m_tGradientLockedInactive;
+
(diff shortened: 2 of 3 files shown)


src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -137,10 +141,14 @@ void CHyprGroupBarDecoration::draw(CMonitor* pMonitor, float a, const Vector2D&
             rect.y -= rect.height;
             rect.width = BARW * pMonitor->scale;
 
-            refreshGradients();
+            if (*PGRADIENTS == 1) {
+                if (m_tGradientActive.m_iTexID == 0) // TODO: find why texture doesn't exist
```
MightyPlaza, line 145: like I said in original issue, despite texture being created previously on the create call they somehow don't exist here, being created 2 times instead of just 1

src/config/ConfigManager.cpp
```diff
@@ -1644,6 +1646,9 @@ void CConfigManager::loadConfigLoadVars() {
         ensureVRR();
     }
 
+    if (!isFirstLaunch)
```
MightyPlaza, line 1649: this crashes if run on 1st launch, could cause problems when run on weird states, but I'm not sure

vaxerski: `&& !g_pCompositor->m_bUnsafeState`

src/render/decorations/CHyprGroupBarDecoration.cpp
vaxerski: try to gdb it maybe? I can try later too.

src/config/ConfigManager.cpp
MightyPlaza: should I log it as a reload error?

src/render/decorations/CHyprGroupBarDecoration.cpp
MightyPlaza: on a closer looks seems like `glGenTextures(1, &m_iTexID);` gets run but the value of `m_iTexID` does not change

vaxerski: thats bad

MightyPlaza: output is 0

vaxerski: output of the funtion? if gles logs any errors I believe it'd go to stdout/stderr. Idk what return codes there are tho

MightyPlaza: using `glGetError()` I get an output of 0 which corresponds to `GL_NO_ERROR`
https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGetError.xhtml
nothing is written to stdout/stderr

vaxerski: ah, this. No clue then.

> some guy on a forum said it might be caused by it running asynchronously from the thread where the context was created, but I don't think that is the cause here

entire pipeline is single-threaded

MightyPlaza: ok, seems like that was right
`glXGetCurrentContext()` is `0x0` in those 4 call while a different value everywhere else

MightyPlaza: are you sure there are no threads, seems like everywhere that is called by events doesn't have an `eglGetCurrentContext()`

vaxerski: only threads are for socket2 iirc. Main renderer is definitely single-threaded.

MightyPlaza:
rebased since issue fixed itself lol
no issues found now

vaxerski (approved):
lgtm, thanks!

src/render/decorations/CHyprGroupBarDecoration.cpp
MightyPlaza: this issue is now back since d2c3b23ace745d3f79926c71d4cb3c481a394841
the gradients are never getting rendered on main
adding the extra rendering in the main loop fixes it, like before

PR #3547, opened 2023-10-11
Guard execinfo.h via build systems

### LINKED ISSUE #3539: Support for musl based systems
### Description

For musl systems execinfo.h and its functions are not able to be used and it fails to compile. You may suggest installing libexecinfo to mitigate this problem however for musl systems libexecinfo has been dropped from alpine as well as gentoo for a long time now because it really wasn't a solution. I have created a patch that will allow every other system apart from musl to retain the execinfo stuff but on musl systems it removes that feature so it can compile. I am not a C developer so there are maybe a few mistakes in this patch file but I would love to hear your thoughts about this and hopefully we can work on incorporating this.

```
diff --git a/src/debug/CrashReporter.cpp b/src/debug/CrashReporter.cpp
--- a/src/debug/CrashReporter.cpp
+++ b/src/debug/CrashReporter.cpp
@@ -1,7 +1,9 @@
 #include "CrashReporter.hpp"
 #include <random>
 #include <sys/utsname.h>
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__GLIBC__) || defined(__NetBSD__) || defined(__OpenBSD__)
 #include <execinfo.h>
+#endif
 #include <fstream>
 #include <signal.h>
 
@@ -78,8 +80,9 @@
     finalCrashReport += std::format("\n\nos-release:\n\t{}\n\n\n", replaceInString(execAndGet("cat /etc/os-release"), "\n", "\n\t"));
 
     finalCrashReport += "Backtrace:\n";

+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__GLIBC__) || defined(__NetBSD__) || defined(__OpenBSD__)
     const auto CALLSTACK = getBacktrace();
+#endif
 
 #if defined(KERN_PROC_PATHNAME)
     int mib[] = {
@@ -106,6 +109,7 @@
     const auto FPATH = std::filesystem::canonical("/proc/self/exe");
 #endif
 
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__GLIBC__) || defined(__NetBSD__) || defined(__OpenBSD__)
     for (size_t i = 0; i < CALLSTACK.size(); ++i) {
         finalCrashReport += std::format("\t#{} | {}\n", i, CALLSTACK[i].desc);
 
@@ -117,6 +121,7 @@
         const auto ADDR2LINE = replaceInString(execAndGet(CMD.c_str()), "\n", "\n\t\t");
         finalCrashReport += "\t\t" + ADDR2LINE.substr(0, ADDR2LINE.length() - 2);
     }
+#endif
 
     finalCrashReport += "\n\nLog tail:\n";

diff --git a/src/helpers/MiscFunctions.cpp b/src/helpers/MiscFunctions.cpp
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -6,7 +6,9 @@
 #include <sys/utsname.h>
 #include <iomanip>
 #include <sstream>
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__GLIBC__) || defined(__NetBSD__) || defined(__OpenBSD__)
 #include <execinfo.h>
+#endif
 
 #if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
 #include <sys/sysctl.h>
@@ -686,6 +688,7 @@
     return subject;
 }
 
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__GLIBC__) || defined(__NetBSD__) || defined(__OpenBSD__)
 std::vector<SCallstackFrameInfo> getBacktrace() {
     std::vector<SCallstackFrameInfo> callstack;
 
@@ -702,8 +705,9 @@
 
     return callstack;
 }
+#endif
 
 void throwError(const std::string& err) {
     Debug::log(CRIT, "Critical error thrown: {}", err);
     throw std::runtime_error(err);
```

vaxerski:
Guards `execinfo.h` and backtrace stuff via a macro

@fufexan for meson / nix

cc @jbeich 

closes #3539


fufexan (dismissed):

src/meson.build
```diff
@@ -24,7 +24,8 @@ executable('Hyprland', src,
     dependency('gl', 'opengl'),
     dependency('threads'),
     dependency('pango'),
-    dependency('pangocairo')
+    dependency('pangocairo'),
+    dependency('libexecinfo', required: false),
```
jbeich, line 28: Redundant with the following:
https://github.com/hyprwm/Hyprland/blob/607c77428be95f1b4caf0cdfd299bed5ce9b21c8/meson.build#L53
https://github.com/hyprwm/Hyprland/blob/607c77428be95f1b4caf0cdfd299bed5ce9b21c8/src/meson.build#L19

Also, libexecinfo as shipped on BSDs and Alpine 3.16 doesn't support pkg-config i.e., `libexecinfo.pc` is missing thus `dependency()` won't work.

meson.build
```diff
@@ -29,6 +29,10 @@ add_project_arguments(
   ],
   language: 'cpp')
 
+if cpp_compiler.check_header('execinfo.h')
+  add_project_arguments('-DHAS_LIBEXECINFO', language: 'cpp')
```
jbeich, line 33: Obsolete macro name after 81115d5f0e7e.
```suggestion
  add_project_arguments('-DHAS_EXECINFO', language: 'cpp')
```

CMakeLists.txt
```diff
@@ -99,6 +99,11 @@ endif()
 
 include(CheckLibraryExists)
 check_library_exists(execinfo backtrace "" HAVE_LIBEXECINFO)
+check_include_file("execinfo.h" EXECINFOH)
```
jbeich, line 102: Can you move the new code above `check_library_exists` block i.e., just above `include`? This one is dedicated to adding `-lexecinfo` on BSDs or nothing on Linux/glibc.

jbeich (changes requested):
CMake and Meson work fine on FreeBSD.

src/meson.build
fufexan: So should it be removed entirely, including `backtrace_dep`?

jbeich: `backtrace_dep` is still required to conditionally add `-lexecinfo` on BSDs or nothing on Linux/glibc.

jbeich:
Modern GCC/Clang support `__has_include`, so touching CMake/Meson isn't necessary:
```c++
#if __has_include(<execinfo.h>)
#include <execinfo.h>
#endif

#if __has_include(<execinfo.h>)
...
#endif
```

vaxerski:
meh, such things I'd rather have the build system check tbh

jbeich (approved):

fufexan (approved):

PR #3549, opened 2023-10-11
Unbreak Meson build on FreeBSD

jbeich:
Regressed by #3543

```c++
$ meson setup /tmp/hyprland_build
$ meson compile -C /tmp/hyprland_build
[...]
In file included from src/pch/pch.hpp:1:
In file included from src/Compositor.hpp:7:
In file included from src/defines.hpp:1:
In file included from src/includes.hpp:148:
In file included from src/helpers/Vector2D.hpp:5:
src/helpers/../macros.hpp:17:10: fatal error: 'version.h' file not found
   17 | #include "version.h"
      |          ^~~~~~~~~~~
```

### DIFF
--- a/scripts/generateVersion.sh
+++ b/scripts/generateVersion.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-cp ./src/version.h.in ./src/version.h -fr
+cp -fr ./src/version.h.in ./src/version.h
 
 HASH=$(git rev-parse HEAD)
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
--- a/scripts/generateVersion.sh
+++ b/scripts/generateVersion.sh
@@ -4,11 +4,11 @@ cp -fr ./src/version.h.in ./src/version.h
 HASH=$(git rev-parse HEAD)
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
 MESSAGE=$(git show ${GIT_COMMIT_HASH} | head -n 5 | tail -n 1 | sed -e 's/#//g' -e 's/\"//g')
-DIRTY=$(git diff-index --quiet HEAD -- || echo \"dirty\")
+DIRTY=$(git diff-index --quiet HEAD -- || echo dirty)
 TAG=$(git describe --tags)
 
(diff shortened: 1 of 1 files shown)


scripts/generateVersion.sh
```diff
@@ -1,14 +1,14 @@
...
 
-awk -i inplace "{sub(/@HASH@/,\"${HASH}\")}1" ./src/version.h
-awk -i inplace "{sub(/@BRANCH@/,\"${BRANCH}\")}1" ./src/version.h
-awk -i inplace "{sub(/@MESSAGE@/,\"${MESSAGE}\")}1" ./src/version.h
-awk -i inplace "{sub(/@DIRTY@/,\"${DIRTY}\")}1" ./src/version.h
-awk -i inplace "{sub(/@TAG@/,\"${TAG}\")}1" ./src/version.h
\ No newline at end of file
+sed -i -e "s/@HASH@/${HASH}/" ./src/version.h
+sed -i -e "s/@BRANCH@/${BRANCH}/" ./src/version.h
+sed -i -e "s/@MESSAGE@/${MESSAGE}/" ./src/version.h
```
vaxerski, line 12: message can contain weird characters like `/` which will destroy this command. We only sanitize `#` and `"`

vaxerski:
and it wouldn't work here? why?

jbeich:
`-i inplace` isn't supported as documented in e1e9c98 commit message. Besides, I don't think `awk` is any better here given it consumers double-quotes in `"dirty"`.

scripts/generateVersion.sh
jbeich: Fixed by using one of the sanitized characters as a the delimited for `s` sed command.

Also, why `#` is sanitized? It doesn't start a comment in C/C++ and has no impact within quotes in Bourne-style shells.

jbeich: Applied `\"` consistently to avoid surprises from tags or (local) branches containing `/` slash.

jbeich: > why `#` is sanitized?

Ah, it's for GNU awk. For best compat with the previous behavior I'll switch to `#` as delimiter.

PR #3550, opened 2023-10-11
meson: Sort input file list

bmwiedemann:

Sort cpp input file list
so that `hyprland` builds in a reproducible way
in spite of non-deterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

This patch was done while working on reproducible builds for openSUSE.

requires a POSIX shell. It should be possible to replace the `find` with a python script, but I did not find a builtin sort function in meson.

I tested that it builds, so it should still work the same as before.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,4 +1,4 @@
-globber = run_command('find', '.', '-name', '*.cpp', check: true)
+globber = run_command('sh', '-c', 'find . -name "*.cpp" | sort', check: true)
 src = globber.stdout().strip().split('\n')
 
 executable('Hyprland', src,


vaxerski:
how does the order of the files change anything?

bmwiedemann:
It influences the link order of .o files into the `hyprland` binary, which makes a difference to how functions are laid out and optimizations are applied in the machine code.

vaxerski (approved):
https://github.com/hyprwm/Hyprland/assets/43317083/2e33f15b-7e34-406e-b994-d67eae9aa963

PR #3557, opened 2023-10-12
Delete SJW

Werenter:
I have fixed SJW CoC.


memchr:
There exist more profound and enduring pursuits than spamming.

vaxerski:
indeed

PR #3559, opened 2023-10-12
Feat: Add new optional "default command" for auto-launching apps on new workspaces

Syndelis:
#### Update 13/10/2023

The PR has been updated to introduce a workspace rule as suggested below, rather than an additional parameter to the `workspace` and `togglespecialworkspace` dispatchers.

#### About this PR

This PR introduces a new [workspace rule](https://wiki.hyprland.org/Configuring/Workspace-Rules/) that allows the user to define a command to be run when switching to a *new* workspace.

The command gets passed to `spawn()`, which is virtually equivalent to calling `hyprctl dispatch exec "<your_command>"`. This means that you can also use the same [execution rules](http://wiki.hyprland.org/Configuring/Dispatchers/#executing-with-rules) that `exec` supports. It's worth noting that the execution rule `workspace <workspacename>` is explicitly overwritten to the workspace the rule has been attributed to. Even if you didn't set it, it will be set automatically.

This required two new methods to be added to `VarList`: `append` and `map`.

#### The Motivation

My biggest use case for special workspaces is to quickly use and hide a terminal, and I absolutely adore Hyprland for having those. However, the setup requires me to 1) launch a terminal and 2) move it to the special workspace. With this feature, I can always press the special workspace shortcut and launch a terminal if there wasn't one already.

I can also see ~more organized~ other people use this to auto-launch their favorite browser when switching to their designated "Browser" workspace.

#### "But you can achieve the same with a script/other shortcut"

Indeed, but the beauty of Hyprland *(to me, at least)* is how much convenience it brings out-of-the-box, with key manager, device settings and (of course) animations. I believe this is one more convenience most would find useful.

Nothing that comes to mind.

Ready for merging.

### Demo Video

https://github.com/hyprwm/Hyprland/assets/25804378/19835f36-d8df-48ba-844b-fe70cc1b64e1

#### Configuration used in the video

```
# Opens 'foot' if workspace doesn't already exist
workspace = special:scratchpad, on-created-empty:foot

# Opens *floating* foot if workspace doesn't already exist
workspace = 5, on-created-empty:[float] foot
```


MightyPlaza:
you can already open programs on a workspace at start with `exec = [workspace special silent] foot` and simply not close them

that being said, the option of doing this as a workspace rule, despite giving less freedom, might be worth considering

Syndelis:
> you can already open programs on a workspace at start with `exec = [workspace special silent] foot` and simply not close them

I don't think that's a realitic comparison to this feature. I personally work with low-memory devices and can't afford to keep apps open just for the convenience of access. Not to mention even idle apps can be hogs to a laptop's battery.

> that being said, the option of doing this as a workspace rule, despite giving less freedom, might be worth considering

I'm curious about what you mean by "giving less freedom". The implementation specifically levarages from the same function that handles the `exec` dispatcher, so you can launch windows with all the same rules supported. Of course, being implemented in both dispatchers that "move" you to another workspace does means that the "silent" rule doesn't appear to make a difference.

MightyPlaza:
> I'm curious about what you mean by "giving less freedom"

With current implementation you can chose to open the workspace with, without or even with different commands
Doing it as a workspace rule would remove this possibility, but would make it more uniform (ex: `workspace +1`) and script friendly
the question is which behavior is better

Syndelis:
> With current implementation you can chose to open the workspace with, without or even with different commands Doing it as a workspace rule would remove this possibility

I'm sorry, I still didn't get it :sweat_smile: ! I'm not sure if this is the adequate answer to your point, but if it helps: the new parameter is optional, so you can still move to workspaces without triggering a command, like the following:

```
# Moves to workspace 1, nothing much
bind = SUPER, 1, workspace, 1

# Moves to workspace 2 and, if it didn't exist, launches firefox on it
bind = SUPER, 1, workspace, 2, firefox
```

FakeMichau:
To me it sounds like something niche enough that it would be better off as a plugin or a script, not part of the main project.

MightyPlaza:
i mean something like 
```
workspace=2, on_create: [float] firefox
```

instead of

> ```
> # Moves to workspace 1, nothing much
> bind = SUPER, 1, workspace, 1
> 
> # Moves to workspace 2 and, if it didn't exist, launches firefox on it
> bind = SUPER, 1, workspace, 2, firefox
> ```
the workspace rule implementation would not allow for this separate behavior

Syndelis:
> To me it sounds like something niche enough that it would be better off as a plugin or a script, not part of the main project.

I hope it doesn't come off as me being rude or overprotective of this PR, but I do think this could be said about a lot of features Hyprland offers when compared to other compositors such as Sway. However, I believe that it's exactly because Hyprland does a lot that it's such a blast to use; it magically seems to fit to everyone's needs.

With that being said, while you believe this is niche enough to be a plugin, I think it would be quite the overkill to make a plugin specifically for this functionality. After all, the changes only add a couple extra operations to the process of moving to a new workspace, which can pretty much be considered negligible both in performance and maintenance costs.

Syndelis:
> i mean something like
> 
> ```
> workspace=2, on_create: [float] firefox
> ```
> 
> instead of
> 
> > ```
> > # Moves to workspace 1, nothing much
> > bind = SUPER, 1, workspace, 1
> > 
> > # Moves to workspace 2 and, if it didn't exist, launches firefox on it
> > bind = SUPER, 1, workspace, 2, firefox
> > ```
> 
> the workspace rule implementation would not allow for this separate behavior

Oh, gotcha! I initially misunderstood it when you said "workspace rule"; I had completely forgotten about these. That said, I'm also unsure of where this feature would be better implemented at. I'm still leaning towards the behavior implemented in this PR because the syntax the new parameter follows is the same as `exec`, which is more likely to be used by key bindings. Nothing against moving it though!

GabrielFoxDev:
I'd just like to point out that it is pretty common to see in r/unixpornm, for instance, setups in which each workspace is represented as an icon to a specific purpose, such as web browsing, as suggested by @Syndelis, or coding/programming, instant messaging, reading, content creation, so forth and so on.

This feature would allow for a more out-of-the-box configuration experience, in which one would could lean on a highly efficient program and not have to write their own scripts for quality of life functionality.

I agree with @MightyPlaza on whether it is worth considering to implement this feature as a workspace rule, as it does make sense to couple this kind of behavior to the workspace itself. On the other hand, I'd also take into consideration that by coupling it to a workspace switch bind instead, one could use different binds to open different applications with the same overall purpose, but in the same specific workspace to that purpose, if so desired.

One use case I have in mind is having mutiple browsers to open on a workspace specific for web app testing purposes.
Another one is for Krita, Gimp and Kdenlive.

vaxerski:
I am in favor of this being in a workspace rule.

Also I was thinking, on a tangent, maybe we should allow mixing workspace rules? :thinking:

Syndelis:
Based on the feedback initially given by @MightyPlaza and supported by @GabrielFoxDev and @vaxerski, commit 0bbad21d7b64bed58674ce1cfbbdbe6369ab6091 removes the optional parameters from the dispatchers and introduces a new workspace rule that serves the same purpose.

The PR's description has been slightly altered to match the new implementation, and a new demo video reflecting the change has been linked.

As a sidenote, a cool side effect this change introduced is that switching to a workspace using relative IDs also trigger the effect, which would previously not be possible by the feature. Thanks again for the feedback, @MightyPlaza!

src/Compositor.cpp
```diff
@@ -2431,7 +2431,7 @@ bool CCompositor::cursorOnReservedArea() {
     return !VECINRECT(CURSORPOS, XY1.x, XY1.y, XY2.x, XY2.y);
 }
 
-CWorkspace* CCompositor::createNewWorkspace(const int& id, const int& monid, const std::string& name) {
+CWorkspace* CCompositor::createNewWorkspace(const int& id, const int& monid, const std::string& name, bool isEmpty) {
```
vaxerski, line 2434: I don't like this approach. Very easy in the future to break the behavior.

Why not just set a flag in the compositor (needsRecheck-ish or something) and then verify this in sanityCheckWorkspaces?

if flag set, and there is an empty workspace that did not have the cmd ran yet (again, boolean for every workspace) then run it

src/config/ConfigManager.cpp
```diff
@@ -1172,6 +1173,9 @@ void CConfigManager::handleWorkspaceRules(const std::string& command, const std:
         rules                  = value.substr(WORKSPACE_DELIM + 1);
     }
 
+    const static std::string ruleOnCreatedEmtpy = "on-created-empty:";
```
vaxerski, line 1176: why?

src/helpers/MiscFunctions.cpp
```diff
@@ -503,6 +503,45 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+
+            bool hadWorkspaceRule = false;
+            rulesList.map([&](std::string& rule) {
+                if (rule.find("workspace") == 0) {
+                    rule             = workspaceRule;
+                    hadWorkspaceRule = true;
+                }
+            });
+
+            if (!hadWorkspaceRule) {
```
vaxerski, line 530: style: no {} around short ifs

src/helpers/MiscFunctions.cpp
```diff
@@ -503,6 +503,45 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
...
+
+            rules = "[" + rulesList.join(";") + "]";
+        } else {
+            rules = "[" + workspaceRule + "]";
+        }
+
+        return rules + " " + cmd;
+    }
+
+    return {};
```
vaxerski, line 542: return "";

src/managers/KeybindManager.cpp
```diff
@@ -810,7 +812,7 @@ void CKeybindManager::changeworkspace(std::string args) {
     const auto         PCURRENTWORKSPACE = g_pCompositor->getWorkspaceByID(PMONITOR->activeWorkspace);
     const bool         EXPLICITPREVIOUS  = args.starts_with("previous");
 
-    if (args.starts_with("previous")) {
+    if (EXPLICITPREVIOUS) {
```
vaxerski, line 815: this doesn't belong in this MR

src/managers/KeybindManager.cpp
```diff
@@ -842,9 +844,10 @@ void CKeybindManager::changeworkspace(std::string args) {
     g_pInputManager->m_bEmptyFocusCursorSet = false;
 
     auto pWorkspaceToChangeTo = g_pCompositor->getWorkspaceByID(BISWORKSPACECURRENT ? PCURRENTWORKSPACE->m_sPrevWorkspace.iID : workspaceToChangeTo);
-    if (!pWorkspaceToChangeTo)
+    if (!pWorkspaceToChangeTo) {
```
vaxerski, line 847: ditto

src/managers/KeybindManager.cpp
```diff
@@ -1467,8 +1470,9 @@ void CKeybindManager::toggleSpecialWorkspace(std::string args) {
         Debug::log(LOG, "Toggling special workspace {} to open", workspaceID);
         auto PSPECIALWORKSPACE = g_pCompositor->getWorkspaceByID(workspaceID);
 
-        if (!PSPECIALWORKSPACE)
+        if (!PSPECIALWORKSPACE) {
```
vaxerski, line 1473: ditto

src/config/ConfigManager.cpp
Syndelis: Just thought using a magic number seemed pointless. There should be no performace penalty on using a const static string, and using it does mean we can use the compiler to count its length and avoid the need of reviewers to have to double check string sizes. As a rule of thumb, I try to avoid magic numbers as much as possible, even if used only once.

With that being said, if you really think this doesn't belong, I'll be glad to remove it. In the meanwhile, I'll work on addressing your other comments in the next commit.

src/helpers/MiscFunctions.cpp
Syndelis: Addressed in 1f9b570

src/managers/KeybindManager.cpp
Syndelis: Addressed in 1f9b570

src/managers/KeybindManager.cpp
Syndelis: Addressed in 1f9b570

src/managers/KeybindManager.cpp
Syndelis: Addressed in 1f9b570

src/Compositor.cpp
Syndelis: Addressed in 41a0734 and 1834dbc

yavko:
Wouldn't it be better to take a dispatcher, so you could do other things separate from starting a new program?
ex:
```
workspace = special:scratchpad, on-created-empty:exec foot
```

vaxerski:
`hyprctl dispatch ...`?

src/Compositor.cpp
```diff
@@ -2044,7 +2053,7 @@ void CCompositor::moveWorkspaceToMonitor(CWorkspace* pWorkspace, CMonitor* pMoni
 
             Debug::log(LOG, "moveWorkspaceToMonitor: Plugging gap with new {}", nextWorkspaceOnMonitorID);
 
-            g_pCompositor->createNewWorkspace(nextWorkspaceOnMonitorID, POLDMON->ID);
+            g_pCompositor->createNewWorkspace(nextWorkspaceOnMonitorID, POLDMON->ID, "");
```
vaxerski, line 2056: ?

src/config/ConfigManager.hpp
```diff
@@ -49,6 +49,7 @@ struct SWorkspaceRule {
     std::optional<int>     rounding;
     std::optional<int>     decorate;
     std::optional<int>     shadow;
+    std::string            onCreatedEmptyRunCmd;
```
vaxerski, line 52: why not make it an `std::optional` too? Keeps consistency.

src/helpers/MiscFunctions.cpp
```diff
@@ -503,6 +503,44 @@ int getWorkspaceIDFromString(const std::string& in, std::string& outName) {
     return result;
 }
 
+std::string cleanCmdForWorkspace(const std::string& inWorkspaceName, std::string dirtyCmd) {
+
```
vaxerski, line 507: pointless double newline. One or none.

vaxerski (commented):
rest is lgtm, except it needs a clang-format

src/Compositor.cpp
Syndelis: Leftover from the removal of 0bbad21d7b64bed58674ce1cfbbdbe6369ab6091, my bad. Addressed in d8d258ea3d152cf831f152feaa668ff8f001fad9

src/config/ConfigManager.hpp
Syndelis: Addressed in 0f0711f

src/helpers/MiscFunctions.cpp
Syndelis: Addressed in 11eddd9

Syndelis:
> rest is lgtm, except it needs a clang-format

I'd hesitated to use it at first because it touched a few lines that I hadn't changed myself. Eitherway, I've ran it and committed the changes in 1c3746f

vaxerski:
I know, for some reason clang-format on my desktop behaves a bit weirdly.

src/helpers/MiscFunctions.cpp
vaxerski: I will disregard this. Short ifs are single-line. What you showed is not. I wish clang-format would support that. Maybe I should write a hyprformat.

memchr:
```yml
# clang-format
RemoveBracesLLVM: true
```

Even though it removes braces from for and while, it is still effective.

vaxerski (approved):
lgtm, thanks!

https://github.com/hyprwm/Hyprland/assets/43317083/00ff0807-fca0-4cee-a8c6-4653a78d31b6

vaxerski:
we just need a wiki mr

Syndelis:
> we just need a wiki mr

Sure thing, will open it and link it here! Thanks

PR #3560, opened 2023-10-12
handle mouse on decorations

MightyPlaza:
handles mouse actions inside decorations, to allow further decoration integration and avoid future duplicated code

don't think will cause any new issues since it mostly moves the code

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -312,6 +312,19 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
         return;
     }
 
+    if (!m_vOverrideFocalPoint && g_pInputManager->m_bWasDraggingWindow) {
+        for (auto& wd : OPENINGON->pWindow->m_dWindowDecorations) {
+            if (!wd->allowsInput())
+                continue;
+
+            if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) {
+                if (!wd->dragWindowToDecoration(pWindow, MOUSECOORDS))
+                    return;
+                break;
+            }
+        }
+    }
+
     // if it's a group, add the window
     if (OPENINGON->pWindow->m_sGroupData.pNextWindow                                                      // target is group
         && !g_pKeybindManager->m_bGroupsLocked                                                            // global group lock disengaged
@@ -327,25 +340,8 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
 
         m_lDwindleNodesData.remove(*PNODE);
 
-        bool handled = true;
-
-        if (g_pInputManager->m_bWasDraggingWindow) {
-            for (auto& wd : OPENINGON->pWindow->m_dWindowDecorations) {
-                if (!wd->allowsInput())
-                    continue;
-
-                if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) {
-                    wd->dragWindowToDecoration(pWindow, MOUSECOORDS);
-                    handled = false;
-                    break;
-                }
-            }
-        }
-
-        if (handled) {
-            static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("group:insert_after_current")->intValue;
-            (*USECURRPOS ? OPENINGON->pWindow : OPENINGON->pWindow->getGroupTail())->insertWindowToGroup(pWindow);
-        }
+        static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("group:insert_after_current")->intValue;
+        (*USECURRPOS ? OPENINGON->pWindow : OPENINGON->pWindow->getGroupTail())->insertWindowToGroup(pWindow);
 
         OPENINGON->pWindow->setGroupCurrent(pWindow);
         pWindow->applyGroupRules();
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -329,14 +329,16 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
 
         bool handled = true;
 
-        for (auto& wd : OPENINGON->pWindow->m_dWindowDecorations) {
-            if (!wd->allowsInput())
-                continue;
+        if (g_pInputManager->m_bWasDraggingWindow) {
+            for (auto& wd : OPENINGON->pWindow->m_dWindowDecorations) {
+                if (!wd->allowsInput())
+                    continue;
 
-            if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) { // TODO: Deny when not using mouse
-                wd->dragWindowToDecoration(pWindow, MOUSECOORDS);
-                handled = false;
-                break;
+                if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) {
+                    wd->dragWindowToDecoration(pWindow, MOUSECOORDS);
+                    handled = false;
+                    break;
+                }
             }
         }
 
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -318,7 +318,7 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
                 continue;
 
             if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) {
-                if (!wd->onEndDragOnDeco(pWindow, MOUSECOORDS))
+                if (!wd->onEndWindowDragOnDeco(pWindow, MOUSECOORDS))
                     return;
                 break;
             }
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -318,7 +318,7 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
                 continue;
 
             if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) {
-                if (!wd->dragWindowToDecoration(pWindow, MOUSECOORDS))
+                if (!wd->onEndDragOnDeco(pWindow, MOUSECOORDS))
                     return;
                 break;
             }
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -327,15 +327,20 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
 
         m_lDwindleNodesData.remove(*PNODE);
 
-        const wlr_box box = OPENINGON->pWindow->getDecorationByType(DECORATION_GROUPBAR)->getWindowDecorationRegion().getExtents();
-        if (wlr_box_contains_point(&box, MOUSECOORDS.x, MOUSECOORDS.y)) { // TODO: Deny when not using mouse
-            const int SIZE               = OPENINGON->pWindow->getGroupSize();
-            const int INDEX              = (int)((MOUSECOORDS.x - box.x) * 2 * SIZE / box.width + 1) / 2 - 1;
-            CWindow*  pWindowInsertAfter = OPENINGON->pWindow->getGroupWindowByIndex(INDEX);
-            pWindowInsertAfter->insertWindowToGroup(pWindow);
-            if (INDEX == -1)
-                std::swap(pWindow->m_sGroupData.pNextWindow->m_sGroupData.head, pWindow->m_sGroupData.head);
-        } else {
+        bool handled = true;
+
+        for (auto& wd : OPENINGON->pWindow->m_dWindowDecorations) {
+            if (!wd->allowsInput())
+                continue;
+
+            if (wd->getWindowDecorationRegion().containsPoint(MOUSECOORDS)) { // TODO: Deny when not using mouse
+                wd->dragWindowToDecoration(pWindow, MOUSECOORDS);
+                handled = false;
+                break;
+            }
+        }
+
+        if (handled) {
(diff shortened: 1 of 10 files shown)


src/render/decorations/IHyprWindowDecoration.hpp
```diff
@@ -39,6 +39,12 @@ class IHyprWindowDecoration {
 
     virtual bool                     allowsInput();
 
+    virtual void                     dragWindowToDecoration(CWindow*, const Vector2D&);
+
+    virtual void                     clickDecoration(const Vector2D&);
+
+    virtual void                     dragFromDecoration(const Vector2D&);
```
vaxerski, line 46: begin? end?

src/render/decorations/IHyprWindowDecoration.hpp
```diff
@@ -39,6 +39,12 @@ class IHyprWindowDecoration {
 
     virtual bool                     allowsInput();
 
+    virtual void                     dragWindowToDecoration(CWindow*, const Vector2D&);
```
vaxerski, line 42: same. begin? end?

And generally all 3 should be prefixed with `on` as they are callbacks

src/render/decorations/IHyprWindowDecoration.hpp
MightyPlaza: what about?
```
virtual void                     onEndDragWindow(CWindow*, const Vector2D&);
virtual void                     onClickDecoration(const Vector2D&);
virtual void                     onBeginDragWindow(const Vector2D&);

```

vaxerski: hm. A bit non-descriptive. 

What is `onBeginDragWindow`? on begin drag of the parent? the deco?

if we want to handle drags on the decoration, why not just `onMouseDown` `onMouseUp`?

The names should generally make it clear what the function is about.

MightyPlaza: `onEndDragWindow` should be called when a window gets dragged onto a decorations (`movewindow` mouse dispatcher) (not what currently happens but TODO should get fixed before that is done)

`onBeginDragWindow` is called when `movewindow` is used but mouse position is inside a window decoration region

not really `onMouseDown` and `onMouseUp`

vaxerski: We can make both into `onBeginDragOnDeco` `onEndDragOnDeco` with an argument like an enum `eDragDecoType` with stuff like `DRAG_TYPE_WINDOW` or stuff, and an `std::any` for data. In the enum, we can use comments to doc what data is passed via the any.

MightyPlaza: > We can make both into `onBeginDragOnDeco` `onEndDragOnDeco` with an argument like an enum `eDragDecoType` with stuff like `DRAG_TYPE_WINDOW` or stuff, and an `std::any` for data. In the enum, we can use comments to doc what data is passed via the any.
> 
> Though we can also add a `DRAG_DATA_NONE`, and then document it could be a click as well, but at that point it's a `onDecoClickBegin` `onDecoClickEnd`
> 
> Some decos might want to respond to clicks in general, and currently they have to hook into events for that.

not sure what you mean, are you suggesting turning the 3 functions into 1 that handles all the events?

vaxerski: no, more like
```
void onDecoClickBegin(const Vector2D& pos, const eClickDataType type, std::any data);
```

where `eClickDataType` could be `CLICK_DATA_NONE`, `CLICK_DATA_WINDOW`, etc. First would imply no data, second would imply `std::any data` has a `CWindow*`

MightyPlaza: still don't get why/how would a click pass a `Cwindow*`, would this be used instead of the current `onBeginDragWindow`?

vaxerski: generally my idea is to pass clicks to the decoration, so that in the future, for example, you can drag tabs around.

Unifying this with dropping a window would be nice.

MightyPlaza: > generally my idea is to pass clicks to the decoration, so that in the future, for example, you can drag tabs around.
> 
> Unifying this with dropping a window would be nice.

wouldn't this be done by adding another drag mode in input manager?
that wouldn't need any extra arguments, except the mouse event type

vaxerski: well no, these are more like "attachments" while drag modes are like resizes

MightyPlaza: I'm pretty sure drag mode would have to be involved for dragging tabs from one group to another.

same group dragging should be doable by adding an mouse event param to `onClickDecoration`, and since it would be fully handled by the deco there would be no need to pass window as a parameter

vaxerski: dont understand your point here.

MightyPlaza: if you're clicking down on a decoration and releasing on mouse up or when leaving deco, that can fully done just by sending the mouse up/down and the mouse move events, no need for extra args

my point is, I still don't get what 
> ```
> void onDecoClickBegin(const Vector2D& pos, const eClickDataType type, std::any data);
> ```
would be used for

vaxerski: we can do that too, thats fine

MightyPlaza: then what should I name the functions?

vaxerski: onMouseDown? I guess?

MightyPlaza:
decided to remove `onEndDragWindow` final nesting with grouped windows
now each decoration can do what it wishes when a window gets dragged onto it

haven't found any issues yet, but probably needs more testing

src/render/decorations/IHyprWindowDecoration.hpp
```diff
@@ -39,6 +39,12 @@ class IHyprWindowDecoration {
 
     virtual bool                     allowsInput();
 
+    virtual void                     onBeginDragOnDeco(const Vector2D&);
+
+    virtual bool                     onEndDragOnDeco(CWindow*, const Vector2D&);
```
vaxerski, line 44: needs a comment what CWindow* and the return value is. For CWindow* a descriptive name would suffice.

src/render/decorations/IHyprWindowDecoration.hpp
```diff
@@ -39,6 +39,12 @@ class IHyprWindowDecoration {
...
+
+    virtual void                     onMouseDown(const Vector2D&, wlr_pointer_button_event*);
```
vaxerski, line 46: onMouseDownOnDeco. This is not onMouseDown. Can be alternatively clarified by a comment.

src/managers/KeybindManager.cpp
```diff
@@ -1836,25 +1836,22 @@ void CKeybindManager::mouse(std::string args) {
...
-                        g_pKeybindManager->m_bGroupsLocked = true;
-                        g_pLayoutManager->getCurrentLayout()->onWindowCreated(pWindow);
-                        g_pKeybindManager->m_bGroupsLocked = GROUPSLOCKEDPREV;
+            if (pWindow && !pWindow->m_bIsFullscreen && !pWindow->hasPopupAt(mouseCoords)) {
+                for (auto& wd : pWindow->m_dWindowDecorations) {
+                    if (!wd->allowsInput())
+                        continue;
+
+                    if (wd->getWindowDecorationRegion().containsPoint(mouseCoords)) {
+                        wd->onBeginDragOnDeco(mouseCoords);
```
vaxerski, line 1845: okay, I finally understand what this means.

Needs a comment then, something like
```
// called when the user calls the "movewindow" mouse dispatcher on the deco
```

And I'd name them `onBeginWindowDragOnDeco`

Since the end function has an override return value, maybe this should also be allowed to "consume" the event?

src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -305,3 +304,79 @@ void CHyprGroupBarDecoration::refreshGradients() {
 bool CHyprGroupBarDecoration::allowsInput() {
     return true;
 }
+
+bool CHyprGroupBarDecoration::onEndDragOnDeco(CWindow* m_pDraggedWindow, const Vector2D& pos) {
+
+    if (!(!g_pKeybindManager->m_bGroupsLocked                                                                                 // global group lock disengaged
```
vaxerski, line 310: note to self: this entire if probably should be a function, it's used many times.

src/managers/KeybindManager.cpp
MightyPlaza: > Since the end function has an override return value, maybe this should also be allowed to "consume" the event?

not sure what you mean

src/layout/MasterLayout.cpp
```diff
@@ -109,18 +122,8 @@ void CHyprMasterLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection direc
...
-            const int INDEX              = (int)((MOUSECOORDS.x - box.x) * 2 * SIZE / box.width + 1) / 2 - 1;
-            CWindow*  pWindowInsertAfter = OPENINGON->pWindow->getGroupWindowByIndex(INDEX);
-            pWindowInsertAfter->insertWindowToGroup(pWindow);
-            if (INDEX == -1)
-                std::swap(pWindow->m_sGroupData.pNextWindow->m_sGroupData.head, pWindow->m_sGroupData.head);
-        } else {
-            static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("group:insert_after_current")->intValue;
-            (*USECURRPOS ? OPENINGON->pWindow : OPENINGON->pWindow->getGroupTail())->insertWindowToGroup(pWindow);
-        }
+        static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("group:insert_after_current")->intValue;
```
vaxerski, line 125: ditto

src/managers/KeybindManager.cpp
```diff
@@ -1836,25 +1836,22 @@ void CKeybindManager::mouse(std::string args) {
...
+                    if (wd->getWindowDecorationRegion().containsPoint(mouseCoords)) {
+                        wd->onBeginWindowDragOnDeco(mouseCoords);
+                        break;
                     }
                 }
             }
 
-            g_pInputManager->currentlyDraggedWindow = pWindow;
-            g_pInputManager->dragMode               = MBIND_MOVE;
+            if (g_pInputManager->currentlyDraggedWindow == nullptr)
```
vaxerski, line 1851: prefer `!var`

src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -305,3 +304,79 @@ void CHyprGroupBarDecoration::refreshGradients() {
...
+
+bool CHyprGroupBarDecoration::onEndWindowDragOnDeco(CWindow* pDraggedWindow, const Vector2D& pos) {
+
+    if (!(!g_pKeybindManager->m_bGroupsLocked                                                                             // global group lock disengaged
+          && ((pDraggedWindow->m_eGroupRules & GROUP_INVADE && pDraggedWindow->m_bFirstMap)                               // window ignore local group locks, or
+              || (!m_pWindow->getGroupHead()->m_sGroupData.locked                                                         //    target unlocked
+                  && !(pDraggedWindow->m_sGroupData.pNextWindow && pDraggedWindow->getGroupHead()->m_sGroupData.locked))) //    source unlocked or isn't group
+          && !pDraggedWindow->m_sGroupData.deny                                                                           // source is not denied entry
+          && !(pDraggedWindow->m_eGroupRules & GROUP_BARRED && pDraggedWindow->m_bFirstMap)))                             // group rule doesn't prevent adding window
+        return 1;
```
vaxerski, line 316: bools are `true` and `false`

src/render/decorations/IHyprWindowDecoration.cpp
```diff
@@ -28,3 +28,11 @@ CRegion IHyprWindowDecoration::getWindowDecorationRegion() {
 bool IHyprWindowDecoration::allowsInput() {
     return false;
 }
+
+void IHyprWindowDecoration::onBeginWindowDragOnDeco(const Vector2D&) {}
+
+bool IHyprWindowDecoration::onEndWindowDragOnDeco(CWindow* pDraggedWindow, const Vector2D&) {
+    return 1;
```
vaxerski, line 35: bools are `true` or `false`

src/layout/DwindleLayout.cpp
```diff
@@ -327,18 +340,8 @@ void CHyprDwindleLayout::onWindowCreatedTiling(CWindow* pWindow, eDirection dire
...
-            const int INDEX              = (int)((MOUSECOORDS.x - box.x) * 2 * SIZE / box.width + 1) / 2 - 1;
-            CWindow*  pWindowInsertAfter = OPENINGON->pWindow->getGroupWindowByIndex(INDEX);
-            pWindowInsertAfter->insertWindowToGroup(pWindow);
-            if (INDEX == -1)
-                std::swap(pWindow->m_sGroupData.pNextWindow->m_sGroupData.head, pWindow->m_sGroupData.head);
-        } else {
-            static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("group:insert_after_current")->intValue;
-            (*USECURRPOS ? OPENINGON->pWindow : OPENINGON->pWindow->getGroupTail())->insertWindowToGroup(pWindow);
-        }
+        static const auto* USECURRPOS = &g_pConfigManager->getConfigValuePtr("group:insert_after_current")->intValue;
```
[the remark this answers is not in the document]
MightyPlaza: not sure what you mean
code was moved above, here it would be restricted to grouped windows only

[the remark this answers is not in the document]
vaxerski: nvm then

vaxerski (approved):
lgtm

PR #3568, opened 2023-10-14
fix(meson): use `ln` with force flag

memchr:
Or a subsequent meson setup command will fail because ln will return a non-zero value if there is an existing symlink.


vaxerski (approved):
nvm I'll review it cuz fuf dead

PR #3570, opened 2023-10-14
Fix makefile using `-d` test for a binary file AND re-add missing portals conf

Frontear:
The current changed Makefile uses `-d` to test if `make all` had been run. The problem is that `-d` is to test for the existence of a directory, while the Makefile uses it to test for the Hyprland binary (via `-d ./build/Hyprland`). This causes the Makefile to instantly fail when running either `make install` or `make installheaders`.

Additionally, since I noticed `./assets/hyprland-portals.conf` was not being copied in the Makefile, I added a declaration to do that as well, since it was part of the original Makefile.

None

Ready for merging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ all:
 	$(MAKE) release
 
 install:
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
slaay gurl

PR #3571, opened 2023-10-14
fix(build): include version.h in PluginAPI.hpp

memchr:
- Include "version.h" in PluginAPI.hpp, otherwise meson or plugin build will fail.
- Also, use `[ -f ]` to test files in Makefile instead of `-d`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,6 @@ all:
 	$(MAKE) release
 
 install:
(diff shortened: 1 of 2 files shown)


vaxerski:
cnfl

vaxerski (approved):
lgtm

PR #3584, opened 2023-10-15
Debug: respect logging settings for wlr

### LINKED ISSUE #3583: Logging causes freezes
### Hyprland Version

Latest main

### Bug or Regression?

Bug

### Description

Hyprland enables logs by default and writes them to disk in the worst way imaginable (`open` + `write` + `close` for each event), without any async/direct stuff.
That is the cause of random freezes when the disk becomes a bit busy.

Also, it's currently not possible to disable logging completely - cause `wlr` logging does not respect ENV nor `debug.disable_logs` (I will attach a fix for that).

My proposal is to either disable logging by-default or in the example config, or to improve it (or better - use any ready-to-use logging library).

### How to reproduce

Reproducible with, for example, `stress --hdd 20 --timeout 30`

```
[ayles@ayles-pc:~]$ sudo perf trace -p `pidof Hyprland` --duration 20 2>&1 | egrep "futex|epoll_wait|poll" -v
  2900.967 (1977.109 ms): .Hyprland-wrap/147332 write(fd: 117, buf: 0x18077c0, count: 76)                             = 76
 10371.868 (6853.721 ms): .Hyprland-wrap/147332 write(fd: 117, buf: 0x180ab60, count: 79)                             = 79
```

_No response_

ayles:
Fixes #3583

### DIFF
--- a/src/debug/Log.cpp
+++ b/src/debug/Log.cpp
@@ -10,6 +10,12 @@ void Debug::init(const std::string& IS) {
 }
 
 void Debug::wlrLog(wlr_log_importance level, const char* fmt, va_list args) {
+    if (disableLogs && *disableLogs)
+        return;
+
+    if (level > wlr_log_get_verbosity())
+        return;
+
     char*         outputStr = nullptr;
 
     std::ofstream ofs;


vaxerski (approved):
my bad, thanks!

PR #3588, opened 2023-10-16
Make new windows taking over fullscreen keep the existing mode

eriedaberrie:

I thought it was strange that, with `misc:new_window_takes_over_fullscreen` set to 1, creating a new window while a window was currently maximized would cause the new window to open in fullscreen. This PR makes new windows match the fullscreen mode of the existing fullscreened window.

No

It should be good to go


vaxerski (approved):
lgtm, thanks!

Strength realigned

PR #3459, opened 2023-09-30
improved hypr_chan config options

DaniD3v:
#### Describe your PR, what does it fix/add?
removed disable_hypr_chan and replaced it with force_hypr_chan = 0.

This removes 1 unnecessary config variable and adds the ability to select which hypr_chan you want to enforce.

Ok ik this is kinda useless but I really wanted to be able to select my favorite hypr_chan.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Removed disable_hypr_chan (but force_hypr_chan=true still works)

I can also add some compatibility code so that force_hypr_chan is automatically set to 0 when disable_hypr_chan is true

#### Is it ready for merging, or does it need work?
Ready

PR #3589, opened 2023-10-16
renderer: unbreak on 32-bit archs via static_cast

jbeich:
Regressed by #3459
`intValue` returns `int64_t` often defined via `long long` but `L` literal suffix is `long`


vaxerski (approved):
ok but who the fuck is running a 32-bit arch

PR #3590, opened 2023-10-17
Fix error building without CMAKE_BUILD_TYPE being set

ideasman42:
This resolves the error building without a CMAKE_BUILD_TYPE.
```
CMake Error at CMakeLists.txt:36 (string):                                                                                                                                                          
  string no output variable specified
```

Resolve a build error with CMake.

It's ready.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,11 +5,6 @@ include(CheckIncludeFile)
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/props.json PROPS)
 string(JSON VER GET ${PROPS} version)
 
-# Avoid having an empty `CMAKE_BUILD_TYPE`.
-if(NOT DEFINED CMAKE_BUILD_TYPE_INIT)
-  set(CMAKE_BUILD_TYPE_INIT "Release")
-endif()
-
 project(Hyprland
     DESCRIPTION "A Modern C++ Wayland Compositor"
     VERSION ${VER}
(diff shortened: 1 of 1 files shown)


q234rty:
It would be better to also fix other valid `CMAKE_BUILD_TYPE` s as well, i.e. translate `None` to meson `empty`, `RelWithDebInfo` to meson `debugoptimized`, `MinSizeRel` to meson `minsize` .

vaxerski:
true that. Also, why is it `_INIT`?

ideasman42:
> true that. Also, why is it `_INIT`?

This is used to set the initial build type, setting `CMAKE_BUILD_TYPE` would overrides the build type set, preventing other kinds of builds, debug and so on.

Strangely `CMAKE_BUILD_TYPE_INIT` is undocumented, but it is used. See:

- https://gitlab.kitware.com/cmake/cmake/-/blob/ee5f31ba727e392e1ebc9f5be54c74ac1868d029/Modules/CMakeInitializeConfigs.cmake#L22

- https://stackoverflow.com/questions/46820155/should-i-cache-cmake-build-type/46835792#46835792

----

If you wanted to avoid relying on an undocumented variable you could always do something like this:

```
if(NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE "Release")
endif()
```

vaxerski:
latter seems more reasonable

ideasman42:
Updated to convert the CMake build type to meson build types (with support for empty build type).

q234rty:
cmake `None` i.e. meson's `empty` seems to be missing

PR #3595, opened 2023-10-17
input: don't reset cursor movement timer on simulateMouseMovement

### LINKED ISSUE #2570: `cursor_inactive_timeout` doesn't hide cursor after `movefocus`
When I set `cursor_inactive_timeout = 1` I expect the cursor to hide after not touching my mouse for 1 second. However, when I change the focused window via `movefocus` the cursor not only reappears and moves to the centre of the new focused window but it also won't hide (ever) unless I use my keyboard on that window (using only my mouse won't make it hide again). This issue appears in any app I've tried.

I don't know what is the expected behaviour here but I would expect the cursor to not appear when changing window focus or at the very least to disappear 1 second after the focus has changed.

andresilva:

Fixes #2570.

Yes.


vaxerski (approved):
thanks

PR #3612, opened 2023-10-18
meson: generate version.h before install_headers

memchr:
Otherwise, `meson install` command would not install version.h as a header in a clean build.


vaxerski (approved):
thanks!

PR #3631, opened 2023-10-21
Change the value of STRVAL_EMPTY to empty string

diniamo:
See the title. This was causing unpredictable behavior with screen shaders (and maybe some other variables too?).

This might be considered a breaking change for some, including me.

Ready


vaxerski:
this breaks every single string config value. Please read the codebase before making such changes.

Issue #3617, opened 2023-10-19
"dimaround" window rule perpetually dims windows with each new instance of target application


slothsh:
### Hyprland Version

Hyprland, built from branch main at commit 1a6f961de2eff84e4725255f09d33395c4a52fcc dirty (hyprctl: Make device configs queryable (3226)). Tag: v0.29.1-35-g1a6f961d  flags: (if any)

### Bug or Regression?

Bug

### Description

Using the `windowrule = dimaround,WINDOW` will dim all other applications around `WINDOW`, as expected. However, spawning more than *one* instance of `WINDOW` causes the dimming boxes on other applications to compound. It causes the dimming on other applications to become darker with each new instance of `WINDOW` that is created.

The bug is present on both `windowrule` & `windowrulev2`.

Platform: Arch Linux x86_64

![hyprland-dimaround-bug](https://github.com/hyprwm/Hyprland/assets/54421807/4dfa9e4e-c441-4df5-af5a-8b1c6fa96cfd)

### How to reproduce

1. Create a `dimaround` window rule for any application in your `hyprland.conf` configuration file.
2. Open more than one instance of the `dimaround` target application.

vaxerski:
this aint a bug tho

SoulXP:
It seems like unexpected behavior, though. If it *is* indeed intended behavior, would you suggest classifying it as a feature-request to have windows only dim once?

vaxerski:
it now is a feature request

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #3633, opened 2023-10-21
add `decoration:dim_once` variable to dim only once during renderWindow (hyprwm#3617)

slothsh:
- Added `decoration:dim_once` variable
- When `dim_once` is set, `renderWindow()` only dims if the current window is the same as the compositor's last window

As described in #3617, the `dimaround` window rule's default behavior is to draw a new transparent rectangle for each open instance of the target application.

This PR adds a `dim_once` variable in `decorations` to allow the user to specify whether dimming should only be applied once instead of being compounded.

At the moment, the `dim_once` variable is defaulted to false - which is the current dimming behavior. I'd recommend, though, that we default it to true. Dimming windows only once feels like the natural approach, or at least that's my opinion.

Ready for merging.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -552,10 +552,10 @@ void IHyprLayout::changeWindowFloatingMode(PHLWINDOW pWindow) {
 
         pWindow->m_vLastFloatingSize = PSAVEDSIZE;
 
-        // move to narnia because we don't wanna find our own node. onWindowCreatedTiling should apply the coords back.
+        // move to narnia because we don't wanna find our own node. onWindowCreated should apply the coords back.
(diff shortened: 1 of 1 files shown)


vaxerski:
what if there is one `dimaround` window but it's not focused? It will not be dimmed.

slothsh:
@vaxerski, sharp eye you've got there, thank you for spotting that. I've amended accordingly.

However, with the latest commit, c253d79, I'm in two minds about inlining directly in `renderWindow()` or having `markDimTarget()` as a method on `CCompositor`.

What would you say is the best approach?

vaxerski:
I am contemplating something else.

What if there are two dimaround windows, neither being in focus?

Issue #3573, opened 2023-10-15
Tablet Issue with fractional scaling and xwayland apps (force_zero_scaling=true)


ooyjian:
### Hyprland Version

Hyprland, built from branch  at commit 1b48642fd15c433c53876f1b933dcd46265caf8f dirty (). Tag: v0.30.0

### Bug or Regression?

Bug

### Description

Hi all, 

I was trying a wacom intuos medium tablet on Hyprland (specifically with Aseprite and Krita, two apps that can only run with xwayland). On my display I have fractional scaling with 1.5, and on xwayland I enabled the force_zero_scaling setting. My problem is that the tablet can only reach 2/3 of the screen in those apps. If I set the 1.5 setting back to zero, this problem disappears. 

Couldn't find someone else with similar problems, so I figure this is the best place to ask. Not sure if this is the issue with Hyprland or issues with xwayland apps. 

Thanks!

Related Configs:
```
monitor=HDMI-A-1, 2560x1440,1920x0,1.5
monitor=eDP-1, preferred, auto, 1.75

xwayland {
	force_zero_scaling = true
}

...

device:wacom-intuos-bt-m-pen { 
	transform=2560x1440
	output=HDMI-A-1
}
```

### How to reproduce

1. Set scaling to be 1.5 on monitor
2. Set force_zero_scaling=true for xwayland
3. Use tablet in Krita

ooyjian:
Any suggestions?

Ant01n3:
Hello,

Same problem here with v 0.30.0 (and later built from git). It seems mouse events are correctly scaled back, but not tablet and touch events. This option would indeed be so great for Krita that is far from having Wayland support actually.

hanspaerna:
Having the same problem.

vaxerski:
MRs welcome. I do not use tablets.

Ant01n3:
> MRs welcome. I do not use tablets. Typing "I have the same problem" will not speed this up.

Sorry, I just wanted to add that touch was also affected. In the meantime, I looked at the problem and I think I may have a solution: mouse events coordinates are scaled by `m_fX11SurfaceScaledBy` but tablet or touch events are not (in managers/input/Tablets.cpp and Touch.cpp). 

I tried to run my modified version of Hyrpland this afternoon and had no problems, however I do not know the code base at all, and I may also have forgotten a lot of things... Should I try to make a pull request? 

In any case, I'm available to test things if needed with my own tablet.

[...]

PR #3644, opened 2023-10-23
Attempt to fix force_zero_scaling for tablet and touch

Ant01n3:

Tries to fix tablet and touch events not scaled back (mouse is OK) when using `force_zero_scaling` on XWayland windows and output scale is not 1. (Fixing #3573 ) 

My knowledge of the code base and event propagation mechanisms of Hyprland is really limited. I hope I haven't forgotten anything.

It needs to be thoroughly tested. I'm currently using it with Krita (which doesn't support Wayland) and I've done some tests with xev to check that mouse, tablet and touch positions match.

### DIFF
--- a/src/managers/input/Touch.cpp
+++ b/src/managers/input/Touch.cpp
@@ -33,7 +33,7 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
     if (m_sTouchData.touchFocusWindow) {
         if (m_sTouchData.touchFocusWindow->m_bIsX11) {
             local = (g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv()) * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy;
-            m_sTouchData.touchSurfaceOrigin = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - local;
+            m_sTouchData.touchSurfaceOrigin = m_sTouchData.touchFocusWindow->m_vRealPosition.goalv();
         } else {
             g_pCompositor->vectorWindowToSurface(g_pInputManager->getMouseCoordsInternal(), m_sTouchData.touchFocusWindow, local);
             m_sTouchData.touchSurfaceOrigin = g_pInputManager->getMouseCoordsInternal() - local;
@@ -64,9 +64,8 @@ void CInputManager::onTouchMove(wlr_touch_motion_event* e) {
         wlr_cursor_warp(g_pCompositor->m_sWLRCursor, nullptr, PMONITOR->vecPosition.x + e->x * PMONITOR->vecSize.x, PMONITOR->vecPosition.y + e->y * PMONITOR->vecSize.y);
 
         auto local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchSurfaceOrigin;
-        if (m_sTouchData.touchFocusWindow->m_bIsX11) {
-            local = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - m_sTouchData.touchSurfaceOrigin;
-        }
+        if (m_sTouchData.touchFocusWindow->m_bIsX11)
+            local = local * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy;
 
         wlr_seat_touch_notify_motion(g_pCompositor->m_sSeat.seat, e->time_msec, e->touch_id, local.x, local.y);
         wlr_seat_pointer_notify_motion(g_pCompositor->m_sSeat.seat, e->time_msec, local.x, local.y);
--- a/src/managers/input/Touch.cpp
+++ b/src/managers/input/Touch.cpp
@@ -64,7 +64,7 @@ void CInputManager::onTouchMove(wlr_touch_motion_event* e) {
         wlr_cursor_warp(g_pCompositor->m_sWLRCursor, nullptr, PMONITOR->vecPosition.x + e->x * PMONITOR->vecSize.x, PMONITOR->vecPosition.y + e->y * PMONITOR->vecSize.y);
 
         auto local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchSurfaceOrigin;
-        if(m_sTouchData.touchFocusWindow->m_bIsX11) {
+        if (m_sTouchData.touchFocusWindow->m_bIsX11) {
             local = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - m_sTouchData.touchSurfaceOrigin;
         }
 
--- a/src/managers/input/Touch.cpp
+++ b/src/managers/input/Touch.cpp
@@ -32,12 +32,12 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
 
     if (m_sTouchData.touchFocusWindow) {
         if (m_sTouchData.touchFocusWindow->m_bIsX11) {
-            local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv();
+            local = (g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv()) * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy;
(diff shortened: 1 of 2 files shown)


src/managers/input/Touch.cpp
```diff
@@ -32,12 +32,12 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
 
     if (m_sTouchData.touchFocusWindow) {
         if (m_sTouchData.touchFocusWindow->m_bIsX11) {
-            local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv();
+            local = (g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv()) * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy;
+            m_sTouchData.touchSurfaceOrigin = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - local;
```
vaxerski, line 36: (coords - windowpos) * scale

src/managers/input/Touch.cpp
```diff
@@ -63,7 +63,10 @@ void CInputManager::onTouchMove(wlr_touch_motion_event* e) {
 
         wlr_cursor_warp(g_pCompositor->m_sWLRCursor, nullptr, PMONITOR->vecPosition.x + e->x * PMONITOR->vecSize.x, PMONITOR->vecPosition.y + e->y * PMONITOR->vecSize.y);
 
-        const auto local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchSurfaceOrigin;
+        auto local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchSurfaceOrigin;
+        if(m_sTouchData.touchFocusWindow->m_bIsX11) {
+            local = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - m_sTouchData.touchSurfaceOrigin;
```
vaxerski, line 68: (coords - windowpos) * scale - origin

vaxerski:
also, needs a clang-format

Ant01n3:
> also, needs a clang-format

Done.

vaxerski:
you resolved conversations without addressing them

Ant01n3:
> you resolved conversations without addressing them

Sorry, I do not understand what you want me to do. Be kind.

vaxerski:
in both cases your calculations are wrong and I commented the correct formulas

Ant01n3:
> in both cases your calculations are wrong and I commented the correct formulas

`(g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv()) * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy`

corresponds, for me, to `(coords - windowPos) * scale` ? But maybe I do not understand well the meaning of these functions/fields ?

vaxerski:
that is correct

Ant01n3:
Ok, so should I consider the code is now ok and resolve the conversations ?

vaxerski:
well, no, you are still calculating it differently and I am pretty sure it is wrong

vaxerski:
I'll test it on my laptop tomorrow and edit myself. But it definitely is wrong

src/managers/input/Touch.cpp
Ant01n3: I'll try to explain my changes and how I think they relate to the formula above. But all my changes revolve around : add a scaling factor where needed in already existing code / computation formulas.

The original code is (if the window is X11) :

```
local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv();
```

I understand `g_pInputManager->getMouseCoordsInternal()` as `coords`.
I understand `m_sTouchData.touchFocusWindow->m_vRealPosition.goalv()` as `windowpos`.

I scaled this whole expression by `m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy` which I understand as `scale`.

The result is : 

```
local = (g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchFocusWindow->m_vRealPosition.goalv()) * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy;
```

Which corresponds to `(coords - windowpos) * scale` in my understanding. After several tests is works for the initial touch down, which is what this function is all about. 

However this function also prepares another, for touch moves, and stores a `touchSurfaceOrigin`. 

The original code uses two distincts paths if the window is X11 or not to compute the `local` variable. In the X11 case this variable units are X11 scaled units, in the other case the units are global screen scale units. What is important is that they differ.

The original code modify this variable this way : 

```
m_sTouchData.touchSurfaceOrigin = g_pInputManager->getMouseCoordsInternal() - local;
```

As `getMouseCoordsInternal()` is in global screen scale units and local may be in one or the other unit, I was forced to compute `touchSurfaceOrigin` differently in the two cases.

The modified code in the case if an X11 window becomes : 

```
m_sTouchData.touchSurfaceOrigin = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - local;
```

`local` is already in X11 units, I have to change the coords again accordingly. (or use a variable and scale it once and for all, but I chose to modify the original code as few as possible). 

All this code prepares what will happen next in `onTouchMove()` that I describe in the other conversation.

src/managers/input/Touch.cpp
Ant01n3: The original code is : 

```
const auto local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchSurfaceOrigin;
```

I changed it to take into account the case where the window is X11 : 

```
 auto local = g_pInputManager->getMouseCoordsInternal() - m_sTouchData.touchSurfaceOrigin;
 if (m_sTouchData.touchFocusWindow->m_bIsX11) {
      local = (g_pInputManager->getMouseCoordsInternal() * m_sTouchData.touchFocusWindow->m_fX11SurfaceScaledBy) - m_sTouchData.touchSurfaceOrigin;
}
```

All I do is to scale anew `getMouseCoordsInternal()` to X11 units. The origin has already been scaled to these units in `onTouchDown`. This could be refactored to avoid computing `local` twice, but once again I wanted to introduce as few changes as possible.

The formula would therefore pass from `coords - origin` to `coords * scale - origin`.

I tested this also and it works as expected, correcting the bug.

Should I introduce the `- windowpos` to have the formula `(coords - windowpos) * scale - origin` ? The original code does not do it, so I'm unsure, but I do not think so. This is already "included" in the `origin` calculation, and indeed not done in the original code. 

Furthermore, when introducing it, the tests fail as soon as the X11 window is not in the top left corner of the screen. My tests are done using Krita and zooming/moving the drawing with touch, and also by running `xev` that report X events and coordinates. 

I am running Hyprland with these corrections for three days now and have encountered no problems with X11 apps.

vaxerski: There are no X11 units here. Everything is surface-local. This will completely implode when your monitor is not at (0,0)

src/managers/input/Touch.cpp
vaxerski: you absolutely cannot scale the getMouseCoordsInternal by any scale. Those are logical wayland coordinates.

Sorry, I also forgot what touchSurfaceOrigin was.

Fixed formula:
`m_sTouchData.touchSurfaceOrigin = m_sTouchData.touchFocusWindow->m_vRealPosition.goalv();`
using the fact that X11 windows are single-surface.

vaxerski:
[patch.txt](https://github.com/hyprwm/Hyprland/files/13183207/patch.txt)

this fixes both my comments, and works properly.

src/managers/input/Touch.cpp
Ant01n3: Well, I did not mean scale them, I mean doing a computation to output the correct event value. 

Ok, simpler way to compute origin.

src/managers/input/Touch.cpp
Ant01n3: Again, x11 "units" was just a way to express the fact there are, in mathematical terms, two coordinate spaces with different metrics. Sorry for being unclear.

Ant01n3:
Done. The last commit contains your patch.

src/managers/input/Touch.cpp
vaxerski: but your conversions were incorrect :)

vaxerski (approved):
lgtm, thanks!

src/managers/input/Touch.cpp
Ant01n3: Ha ha! NO! :P 

They are equivalent to your last code and good from the start. Your new patch code is strictly what I wrote if the rules of algebra are still valid. Substitute and solve to be convinced :)

vaxerski: trust me, they aren't :P

Issue #3602, opened 2023-10-18
bug: special workspaces don't close with `misc:close_special_on_empty` enabled


ozwaldorf:
### Hyprland Version

Hyprland, built from branch main at commit 54e1c2ccbdfa6dfade63221b5df02bf8578d96a6 dirty (fix(build): do not include wlr headers directly, use includes.hpp instead. (3587)). Tag: v0.31.0-1-g54e1c2cc  flags: (if any) debug

### Bug or Regression?

Bug

### Description

The newly added config `misc:close_special_on_empty` does not have any effect when moving the last window out of the special workspace via a dispatch keybind or moving the window with the mouse. Previously, before the config was added, either of these actions would close the special workspace.

### How to reproduce

1. Move a window into a special workspace and open the workspace. 
2. Move the window out of the workspace with a dispatch keybind.
3. Pick up the window with the mouse and release it
4. The window will be moved into the special workspace, showing that it did not close correctly.

OR

1. Move a window into a special workspace and open the workspace.
2. Pick up the window with the mouse and release it.
3. The window will move back into the special workspace, instead of it being closed and the window moving to the normal workspace behind it.

CDTay:
Still having this issue. Moving with the mouse doesn't close the special workspace despite the implication of close_special_on_empty being that if the last window is removed the special workspace will close (silent or not).

Regardless, moving with the mouse seems inherently not "silent", so it feels like it should close the special workspace when moving a window out of it.

PR #3649, opened 2023-10-24
Close special workspace after moving windows out of it

thejch:
Fix the issue that currently when you use `movetoworkspace` to move a window out of the special workspace into a regular WS, it doesn't close the special WS. This also causes weird behavior like being able to move windows in regular workspaces around when in the special WS. 
Seems to me that the correct behavior should be getting out of the special workspace and showing the regular workspace. If I wanted to stay in the current special WS, I would use `movetoworkspacesilent` instead.

No

pending review

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -949,6 +949,12 @@ void CKeybindManager::moveActiveToWorkspace(std::string args) {
 
     POLDWS->m_pLastFocusedWindow = g_pCompositor->getFirstWindowOnWorkspace(POLDWS->m_iID);
 
+    CMonitor* pOldMonitor = g_pCompositor->getMonitorFromID(POLDWS->m_iMonitorID);
+    if (pWorkspace->m_bIsSpecialWorkspace)
+        pOldMonitor->setSpecialWorkspace(pWorkspace);
+    else if (POLDWS->m_bIsSpecialWorkspace)
+        pOldMonitor->setSpecialWorkspace(nullptr);
+
     pMonitor->changeWorkspace(pWorkspace);
 
     g_pCompositor->focusWindow(PWINDOW);
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -949,11 +949,10 @@ void CKeybindManager::moveActiveToWorkspace(std::string args) {
 
     POLDWS->m_pLastFocusedWindow = g_pCompositor->getFirstWindowOnWorkspace(POLDWS->m_iID);
 
-    CMonitor* pOldMonitor = g_pCompositor->getMonitorFromID(POLDWS->m_iMonitorID);
     if (pWorkspace->m_bIsSpecialWorkspace)
-        pOldMonitor->setSpecialWorkspace(pWorkspace);
+        pMonitor->setSpecialWorkspace(pWorkspace);
     else if (POLDWS->m_bIsSpecialWorkspace)
-        pOldMonitor->setSpecialWorkspace(nullptr);
+        g_pCompositor->getMonitorFromID(POLDWS->m_iMonitorID)->setSpecialWorkspace(nullptr);
 
     pMonitor->changeWorkspace(pWorkspace);
 


TangerineGameDev:
I feel like that's not always going to be the requested behavior...

Can we at least make it a config option, if we want to merge this?

thejch:
It's up for discussion. However, I do think this is the bahavior that matches the description of the dispatchers.

vaxerski:
but this has already been fixed?

thejch:
> but this has already been fixed?

If you mean aeb8c8fc7087879c85397f39c565246bc2e193b6, that's only for when the special workspace becomes empty, so not the same thing. This is for always closing it when moving a window out. I think it makes sense because it always opens it when moving a window in. But I also realize this will make the `close_special_on_empty` option kind of useless. Not sure what's the best approach here, but in my opinion I'd like `movetoworkspace` to actually show the new workspace instead of staying in that old special workspace,

vaxerski:
> closing it when moving a window out.

The commit causes that.

>  I think it makes sense because it always opens it when moving a window in

Incorrect, if you use silent, it will not open.

thejch:
> Incorrect, if you use silent, it will not open.

Yes, that's why I didn't change the behavior of silent. But the `movetoworkspace` without silent is behaving just like silent when theres more than 1 windows in special ws. I'm testing on latest git, and if you have 2 windows in special, moving 1 out doesn't close it.
current bahavior: 
move into special: opens special
move into special silent: doesn't open special
move out of special (unless last window): doesn't close special
move out of special silent (unless last window): doesn't close special

vaxerski:
I still don't know what this is fixing. Can I get a reproducible (on head) bug that this will fix?

TangerineGameDev:
> I still don't know what this is fixing. Can I get a reproducible (on head) bug that this will fix?

They are talking about how the behaviour on special workspaces is inconsistent!

1.
When moving a window into the special workspace, it opens the workspace.
That doesn't happen when the window is moved silently into the workspace.

Here's the inconsistency!
2.
When moving the **last** window **out** of a special workspace, it doesn't close the workspace. (It theoretically SHOULD though, following the design philosophy of 1.)
Moving the last window silently out of a special workspace, should still (as it is currently working) NOT close the special workspace. (That's why it's silent)

MightyPlaza:
> 2\. When moving the **last** window **out** of a special workspace, it doesn't close the workspace. (It theoretically SHOULD though, following the design philosophy of 1.)

seems to work for me on git (https://github.com/hyprwm/Hyprland/issues/3602)

TangerineGameDev:
> > 2. When moving the **last** window **out** of a special workspace, it doesn't close the workspace. (It theoretically SHOULD though, following the design philosophy of 1.)
> 
> seems to work for me on git (#3602)

?
Using movetoworkspace doesn't work for me... Even if that variable is set... Only when closing the last window...

Wait, I'll update to the latest git and test again... (I normally keep a relatively up-to-date version of the git repo, but this is a newer commit...)

What is the problem then?

thejch:
> What is the problem then?

The problem is "move out of special" and "move out of special silent" do the same thing, and I'd like them to have different functionalities.

MightyPlaza:
> The problem is "move out of special" and "move out of special silent" do the same thing, and I'd like them to have different functionalities.

agreed, however current implementation doesn't seem ideal (different parts of code fighting for workspace control) neither does it seem to respect `misc:close_special_on_empty` (simply add an if before committed code)

thejch:
> agreed, however current implementation doesn't seem ideal (different parts of code fighting for workspace control) neither does it seem to respect `misc:close_special_on_empty` (simply add an if before committed code)

That is true. After quickly drafting the PR I realized that. Main reason is I also realized the PR needs to be discussed, so I didn't want to put in more effort before we say we actually want this.

TangerineGameDev:
> > What is the problem then?
> 
> The problem is "move out of special" and "move out of special silent" do the same thing, and I'd like them to have different functionalities.

Then do you want the opposite of what I said?
When it's silent, it exits out the special workspace?
That's weird no?

TangerineGameDev:
Oh so you want it to NOT close the special workspace when silent?
Or just in general, they should not have the same behaviour...

thejch:
> Oh so you want it to NOT close the special workspace when silent? Or just in general, they should not have the same behaviour...

silent -> don't close
not silent -> close

CDTay:
> > neither does it seem to respect `misc:close_special_on_empty` (simply add an if before committed code)

`close_special_on_empty` doesn't even respect itself in regards to moving windows out with the mouse cursor, the special workspace stays open while the window and cursor drop into the workspace behind it. An increase in consistency/specificity in variable descriptions may be in order

Dickby:
@vaxerski 
Compare it with non-special workspaces:
if you move a window from one workspace to another unsilently, the moved to workspace is shown afterwards, of course the moved from workspace is not visible anymore.
At the moment moving windows form special workspace to a non special one, if it's silent its equivalent to non special moves (the window disapears from the active special workspace.
If it's moved non-silently, the special workspace stays visible, but not focused.
I think it makes sense to hide the special workspace if a window is moved out non silently so that the moved to workspace is active after moving, that would be the same as for non special workspaces.
Btw, i didn't test this PR, just want to clarify what it is meant to do.

vaxerski:
oooh, okay. I getchu.

src/managers/KeybindManager.cpp
```diff
@@ -949,6 +949,11 @@ void CKeybindManager::moveActiveToWorkspace(std::string args) {
 
     POLDWS->m_pLastFocusedWindow = g_pCompositor->getFirstWindowOnWorkspace(POLDWS->m_iID);
 
+    if (pWorkspace->m_bIsSpecialWorkspace)
+        pMonitor->setSpecialWorkspace(pWorkspace);
+    else if (POLDWS->m_bIsSpecialWorkspace)
+        pMonitor->setSpecialWorkspace(nullptr);
```
vaxerski, line 955: has to be old ws' monitor. They can be different.

thejch:
Messed up the rebase lmao

src/managers/KeybindManager.cpp
```diff
@@ -949,6 +949,12 @@ void CKeybindManager::moveActiveToWorkspace(std::string args) {
 
     POLDWS->m_pLastFocusedWindow = g_pCompositor->getFirstWindowOnWorkspace(POLDWS->m_iID);
 
+    CMonitor* pOldMonitor = g_pCompositor->getMonitorFromID(POLDWS->m_iMonitorID);
+    if (pWorkspace->m_bIsSpecialWorkspace)
+        pOldMonitor->setSpecialWorkspace(pWorkspace);
```
vaxerski, line 954: I believe in this case this would be correct, but this should be pMonitor.

ref above, `pMonitor = g_pCompositor->getMonitorFromID(pWorkspace->m_iMonitorID);`

So pMonitor has pWorkspace

no?

thejch: Yeah indeed. Fixed.

PR #3650, opened 2023-10-24
Blur shader new customization options

Visual-Dawg:
This PR adds new and changes existing customization options for `decoration.blur`.

The new `blur` settings are:
```bash
    # Clamp the brightness after the blurring
    # Allows for a more precise intuitive control than `brightness`
    min_brightness = 0.0
    max_brightness = 1.0    

    # The new color boosting
    boost_colors = true
    saturation_boost = 0.1    
    brightness_boost = 0.0 
    
    base_brightness = 1 # Like previous brightness
```

Other changes:
- Replaced `brightness` with `base_brightness`
- `contrast` and `base_brightness` now get applied only before the blur.

The clamping part ( `softClamp` ) in `Textures.hpp` is currently not optimized.
It should be possible to rewrite it without if statement, but right now it is nice and readable.
I dont know if removing the ifs would increase performance.

~~I could not find where the default config gets generated and add the new setting there.
This still needs to get implemented.~~
Yes. It could be more optimized, but for now its better than before and the code is clean.


Visual-Dawg:
Based on received feedback I will adjust the blur parameters more to get more pleasant results, especially on higher saturation_boost levels

vaxerski:
I'd rather not replace brightness with base_brightness.

TangerineGameDev:
Hmmm weird...

Saturated it should technically look like this:
(Ignore outside the borders...)
image

Also weird photo to pick out 💀

Visual-Dawg:
I havnt picked that photo, thats your wallpaper xDD

It looks like less saturated but bright colors should receive more of a saturation boost.

The reason why its not just saturating everything equally is that perceptually dark colors like deep red or blue would then become way to strong relative to other colors in the image.

In your image the normal saturation method works, because it is solely composed of warm colors. However, if you increase the saturation the reddish pinks will probably start to look kinda ugly.

So yes, I will play around with the function which determines which colors get a boost and keep `brightness` as `brightness`.

Visual-Dawg:
image
Here you can already see how the red starts to clip in an ugly way.

TangerineGameDev:
> I havnt picked that photo, thats your wallpaper xDD
> 
> It looks like less saturated but bright colors should receive more of a saturation boost.
> 
> The reason why its not just saturating everything equally is that perceptually dark colors like deep red or blue would then become way to strong relative to other colors in the image.
> 
> In your image the normal saturation method works, because it is solely composed of warm colors. However, if you increase the saturation the reddish pinks will probably start to look kinda ugly.
> 
> So yes, I will play around with the function which determines which colors get a boost and keep `brightness` as `brightness`.

It's not mine, it's Vaxry's 😆

Visual-Dawg:
@TangerineGameDev didnt realize that your comment wasnt Vaxry, sorry ^^

TangerineGameDev:
> So yes, I will play around with the function which determines which colors get a boost and keep `brightness` as `brightness`.

Honestly I think as something called "saturation boost" it *should* clip, no?
I feel like the user expects, exactly that behaviour I just showed...

If you could somehow keep the values in an HDR color space and then normalize them, I guess maybe a better result could come up? (I don't know if you've already done that...)

Maybe normalize according to an exponential scale?
Colors that are off the charts, receive a way smaller boost in color, than colors that don't get affected as much by the saturation? (But honestly this might just make a big red everything blob, in hindsight...)

Visual-Dawg:
Decreasing the contrast could work.
HSV and HSL are different. HSL is good, but much more expensive. 
HSV is okayish :D

Maybe I should try out HSL, it should work much better.

TangerineGameDev:
Well I was only changing the saturation component here... So I think it shouldn't matter, no?
But yeah no, I was using HSL originally...

Visual-Dawg:
Check out the HSV color picker image I posted. When you increase the saturation you move the X to the right, which creates orange.
And the maxed out orange actually has maximum brightness in HSV. In HSL maximum brightness is white. So HSL is a lot nicer for color this stuff.

So here I used HSL. Much better. But more expensive.
image

@vaxerski What do you think?

TangerineGameDev:
This GLSL shader does conversion with no `if`'s if that's why you're mentioning performance issues:
https://gist.github.com/983/e170a24ae8eba2cd174f

Visual-Dawg:
I think I used this one before, but it turned a lot of color gray. 
But Ill try out, thank you :)

TangerineGameDev:
This shader does conversions between HSL and RGB without any `if`'s for performance.
https://www.shadertoy.com/view/4dKcWK

TangerineGameDev:
> I think I used this one before, but it turned a lot of color gray. But Ill try out, thank you :)

Sorry posted the wrong one, now it's correct...

Visual-Dawg:
This one does turn weak colors into gray, already tried it :/

TangerineGameDev:
How about this site?
I use it quite often... But you have to do the pseudo-code to code yourself...

https://www.easyrgb.com/en/math.php

Theirs has way less branches...
Though it seems like branches are required, for it to look good...

Visual-Dawg:
I will try it out tomorrow and thank you for sharing :)

TangerineGameDev:
I've tried personally to optimize it... (rgb2hsl, I didn't do the other one...)

```
vec3 rgb2hsl_v2( in vec3 c ) {
    float h = 0.0;
    float s = 0.0;
    float l = 0.0;
    
    float r = clamp(c.r, 0.0, 1.0);
    float g = clamp(c.g, 0.0, 1.0);
    float b = clamp(c.b, 0.0, 1.0);

    float cMin = min( r, min( g, b ) );
    float cMax = max( r, max( g, b ) );

    l = ( cMax + cMin ) / 2.0;
    
    if ( cMax > cMin ) {
	float cDelta = cMax - cMin;
        
        //s = l < .05 ? cDelta / ( cMax + cMin ) : cDelta / ( 2.0 - ( cMax + cMin ) ); //Original
        //s = l < .0 ? cDelta / ( cMax + cMin ) : cDelta / ( 2.0 - ( cMax + cMin ) );
        // l can't be under 0.0...
	s = cDelta / ( 2.0 - ( cMax + cMin ) );
        
        float rg = float(r == cMax) * (g - b) / cDelta;
        float gb = float(g == cMax) * (2.0 + (b - r) / cDelta);
        float br = float(b == cMax) * (4.0 + (r - g) / cDelta);
        
        h = rg + gb + br;
        
	h += step(h, 0.0) * 6.0;
	h = h / 6.0;
    }
    
	return vec3( h, s, l );
}
```

I've kept the first huge if, because it probably is more performant to exit out early of all those calculations, instead of always calculating them.

Also quick question, is float(bool == bool) considered branching? No, right?

Anyways, the color conversion still works like the original function.
Test it out! https://www.shadertoy.com/view/msVfzG
hsl2rgb_v2() is the new version, hsl2rgb() is the old one...

Visual-Dawg:
Awesome, thank you, Ill update test it and update the PR later :)

TangerineGameDev:
> Awesome, thank you, Ill update test it and update the PR later :)
> 
> 
> 
> And your `float(bool == bool)` does not seem like branching as its just a multiplication.
> 
> Im not very knowledgeable with shaders yet though :D 

I've also heard multiple times that certain functions like ’step’ actually get compiled to branches (don't know if that's true) and that branches nowadays aren't *as* costly as they originally were...

In any case, I made it a little less branching lol.

src/render/shaders/Textures.hpp
```diff
@@ -167,12 +167,90 @@ void main() {
...
+        y = a + sqrt(pow(1. - a, 2.) - pow(x - 1., 2.));
+    }
+    return y;
+}
 
-uniform float radius;
-uniform vec2 halfpixel;
+// https://www.shadertoy.com/view/wt23Rt
+vec3 saturate(vec3 v) {
+    return clamp(v, 0., 1.);
```
vaxerski, line 212: ??

src/render/shaders/Textures.hpp
```diff
@@ -213,35 +313,87 @@ void main() {
...
+}
+
+
 void main() {
-    vec4 pixColor = texture2D(tex, v_texcoord);
+    vec4  pixColor = texture2D(tex, v_texcoord);
 
-    // contrast
-    pixColor.rgb = (pixColor.rgb - 0.5) * contrast + 0.5;
+    vec3  hsv = rgb2hsv(pixColor.rgb);
```
vaxerski, line 382: these need a nope if no clamping is to be done

src/config/ConfigManager.cpp
```diff
@@ -151,34 +151,39 @@ void CConfigManager::setDefaultVars() {
...
+    configValues["decoration:blur:enabled"].intValue            = 1;
+    configValues["decoration:blur:size"].intValue               = 8;
+    configValues["decoration:blur:passes"].intValue             = 1;
+    configValues["decoration:blur:ignore_opacity"].intValue     = 0;
+    configValues["decoration:blur:new_optimizations"].intValue  = 1;
+    configValues["decoration:blur:xray"].intValue               = 0;
+    configValues["decoration:blur:contrast"].floatValue         = 0.8916;
+    configValues["decoration:blur:brightness"].floatValue       = 1.0;
+    configValues["decoration:blur:boost_colors"].intValue       = 1;
+    configValues["decoration:blur:saturation_boost"].floatValue = 0.1;
```
vaxerski, line 164: default should be 0. Needs more work to be default on

src/render/shaders/Textures.hpp
```diff
@@ -167,12 +167,90 @@ void main() {
...
-uniform vec2 halfpixel;
+// https://www.shadertoy.com/view/wt23Rt
+vec3 saturate(vec3 v) {
+    return clamp(v, 0., 1.);
+}
+vec3 hue2rgb(float hue) {
+    hue = fract(hue);
+    return saturate(vec3(abs(hue * 6. - 3.) - 1., 2. - abs(hue * 6. - 2.), 2. - abs(hue * 6. - 4.)));
+}
+vec3 rgb2hsl(vec3 c) {
```
vaxerski, line 218: https://stackoverflow.com/questions/4728581/hsl-image-adjustements-on-gpu

? seems like more performant code. This has tons of ifs.

Visual-Dawg: Added in [2cb73d6](https://github.com/hyprwm/Hyprland/pull/3650/commits/2cb73d603c659dabd4195e28d50ef5235c01d18e). Looks good

src/config/ConfigManager.cpp
Visual-Dawg: Added in [932dea8](https://github.com/hyprwm/Hyprland/pull/3650/commits/932dea8d5cb783a85a84986e1ec7bcf4a8533bb6)

src/render/shaders/Textures.hpp
Visual-Dawg: Added in [b8fe17f](https://github.com/hyprwm/Hyprland/pull/3650/commits/b8fe17f0eb03f1763be1dceae3e21ab6537ca32a)

src/render/shaders/Textures.hpp
Visual-Dawg: That is now removed with the new HSL function in [2cb73d6](https://github.com/hyprwm/Hyprland/pull/3650/commits/2cb73d603c659dabd4195e28d50ef5235c01d18e)

TangerineGameDev:
> Sadly this version also turns weak colors into fully desaturated grey :/
> 
> image image

I now understand! I am actually seeing it! The hair *should* be blue huh...
Well I fixed it. Seems like the `s` value has to have a range...

Check the ShaderToy shader now out if you wanna test it!

TangerineGameDev:
> Sadly this version also turns weak colors into fully desaturated grey :/
> 
> image image

I now understand! I am actually seeing it! The hair *should* be blue huh...
Well I fixed it. Seems like the `s` value has to have a range...

Check the ShaderToy shader now out if you wanna test it!

Visual-Dawg:
@TangerineGameDev Thank you for sharing!
Vax found a conversion function her https://stackoverflow.com/a/17309861 where the author even takes a look at the generated assembly code to verify that its not branching and it produces good colors, so thats basically perfect -- but still thank you :)

TangerineGameDev:
> @TangerineGameDev Thank you for sharing! Vax found a conversion function her https://stackoverflow.com/a/17309861 where the author even takes a look at the generated assembly code to verify that its not branching and it produces good colors, so thats basically perfect -- but still thank you :)

Are we sure ternaries aren't just another way of branching? I am fine with our current solution (as I said, branching isn't really bad nowadays), but afaik we are still branching this way...

Visual-Dawg:
I noticed that `min` and `max_brightness` create ugly banding on nearly white or black surfaces. 
I will remove those two settings again.
I think putting `brightness` now in `blurfinish` would make sense

Visual-Dawg:
I am not sure about what would be the best approach. 
If `brightness` gets applied after the blurring with a value above 1, it can quickly lead to clipping.
If it gets applied before the blurring with a negative value, it will negatively affect the color boosts.

TangerineGameDev:
> I am not sure about what would be the best approach. 
> 
> If `brightness` gets applied after the blurring with a value above 1, it can quickly lead to clipping.
> 
> If it gets applied before the blurring with a negative value, it will negatively affect the color boosts.
> 
> 
> 
> Also the current implementation of `brightness` will not affect pure black as it is a multiplication and black is 0. 

Should it not be additive?
Also if the value is *above* 1 then clipping is of course expected (We are multiplying after all...)

Visual-Dawg:
The current implementation of `brightness` only adjust the white point as its multiplying the RGB values. 
So black which has a value of `0,0,0` always stays black.
----
Brightness `1.0`
image
---- 
Brightness `2.0`
image
-----
Brightness `0.5`
image
-----
However, what would work is to change the behavior of `brightness` if its `>1.0` or `<1.0`.
If its bigger than `1.0` it could instead affect the black point, like this:
image

I am not sure how to implement that though :D But for sure there is a way. Probably just the `mix` function.

This would probably look nicer, but I dislike how negative values affect the `color_boost` as it wont boost dark that much. 
Thats why I originally added `max_brightness`, which smoothly clamps the brightness, so you dont have to have such a high opacity on translucent surfaces to maintain readability on very bright surfaces (like a light-theme website and your app runner is displayed on top of it), as mixing a solid color with the blur makes the blur look more dull.

What are your thoughts? :) @TangerineGameDev

Visual-Dawg:
Furthermore the current implementation of contrast, which is a simple S-curve, quickly start to clip values, which doesnt look very nice as you end up with a lot of pure black and white.
image

I am not sure how to do it better though, in a way which protects colors better.
Again, your thoughts? :D

TangerineGameDev:
> The current implementation of `brightness` only adjust the white point as its multiplying the RGB values. So black which has a value of `0,0,0` always stays black.
> ## Brightness `1.0`
> image
> ## Brightness `2.0`
> image
> ## Brightness `0.5`
> image
> 
> However, what would work is to change the behavior of `brightness` if its `>1.0` or `<1.0`. If its bigger than `1.0` it could instead affect the black point, like this: image
> 
> I am not sure how to implement that though :D But for sure there is a way. Probably just the `mix` function.
> 
> This would probably look nicer, but I dislike how negative values affect the `color_boost` as it wont boost dark that much. Thats why I originally added `max_brightness`, which smoothly clamps the brightness, so you dont have to have such a high opacity on translucent surfaces to maintain readability on very bright surfaces (like a light-theme website and your app runner is displayed on top of it), as mixing a solid color with the blur makes the blur look more dull.
> 
> What are your thoughts? :) @TangerineGameDev

For the brightness the solution is quite easy actually. Use a power of N!
(A quick example)
https://www.shadertoy.com/view/ddKfz3

If the `pow` value is set to 0.3 for example, the values are stretched to never clip, but decrease the brightness overall. Same goes for values above 1.0. (See `if` case. Guarantees no clipping is happening)

TangerineGameDev:
> Furthermore the current implementation of contrast, which is a simple S-curve, quickly start to clip values, which doesnt look very nice as you end up with a lot of pure black and white. image
> 
> I am not sure how to do it better though, in a way which protects colors better. Again, your thoughts? :D

So about the contrast... I really don't know how we could approach this problem with it clipping...
Thing is I don't understand how you want to get it to work without clipping?

Contrast literally kind of is designed to "clip" (Contrast either increases or decreases the mid-ranges influence causing others to "disappear").

We could try out different smoothing functions to "reduce" the clipping, but it kind of has to happen...

Visual-Dawg:
@TangerineGameDev Thank you for the improved brightness! :)
How would this code look for rgba though? Idk how to write it in a nice way without 3 if statements

Visual-Dawg:
Yes the contrast is a tricky one. 
Iirc DavinciResolve does it by decreasing the amount of contrast applied to a color if it would clip. Similar to your brightness function.

TangerineGameDev:
> @TangerineGameDev Thank you for the improved brightness! :)
> 
> How would this code look for rgba though? Idk how to write it in a nice way without 3 if statements

Wdym if statements? The main piece of code is on the 2nd line. The if statement there is only to visualize that there is no clipping going on...

TangerineGameDev:
> Yes the contrast is a tricky one. 
> 
> Iirc DavinciResolve does it by decreasing the amount of contrast applied to a color if it would clip. Similar to your brightness function. 

So maybe a logarithmic approach?
I'm currently outside when I get home, I'll try something similar out...

TangerineGameDev:
@Visual-Dawg Yea no I don't think there is a way to do what you want...
Thing is you want to spread the colors across the whole range (0, 1), but that's exactly what contrast shouldn't do...
This is paradoxical man...

I did find a function that is REALLY good for mapping contrast though!
Values between 0.0 and 1.0 cause low contrast while higher values cause high contrasty colors!
(PS: This is a function, that doesn't "clip" because it doesn't outright clamp values (so it is reversible), but the values do get VERY small, so it seems like they're clipping to the naked eye...
(If the contrast value is high, the differences are like 0.00001 so yea...)
https://www.shadertoy.com/view/msyfzt

Visual-Dawg:
@TangerineGameDev This looks really promising. Thank you!
I will try it out (might become Monday)
Ah, I have not understood the if statement there. I see, nice

Visual-Dawg:
@TangerineGameDev the new contrast looks really good!
I dont get how to implement `brightness` with `pow` though, so that 1 is normal, +1 is brighter and <1 is darker
```glsl
    pixColor.r =  pow(pixColor.r, brightness);
    pixColor.g =  pow(pixColor.g, brightness);
    pixColor.b =  pow(pixColor.b, brightness);
```

TangerineGameDev:
> @TangerineGameDev the new contrast looks really good!
> 
> I dont get how to implement `brightness` with `pow` though, so that 1 is normal, +1 is brighter and <1 is darker
> 
> ```glsl
> 
>     pixColor.r =  pow(pixColor.r, brightness);
> 
>     pixColor.g =  pow(pixColor.g, brightness);
> 
>     pixColor.b =  pow(pixColor.b, brightness);
> 
> ```

```glsl

    pixColor.r =  1.0 - pow(pixColor.r, brightness);

    pixColor.g =  1.0 - pow(pixColor.g, brightness);

    pixColor.b =  1.0 - pow(pixColor.b, brightness);

```

Visual-Dawg:
I think I tried that already and it inverted the colors. But gimme a sec

TangerineGameDev:
> I think I tried that already and it inverted the colors. But gimme a sec
> 
> 
> 
> Edit: Yes, the colors get inverted like that

Wait, doesn't the original work the way that you described then?

Visual-Dawg:
True, I think brightness is fine as it is

TangerineGameDev:
> I think I tried that already and it inverted the colors. But gimme a sec
> 
> 
> 
> Edit: Yes, the colors get inverted like that

```glsl

    pixColor.r =  pow(pixColor.r, 1.0 / brightness);

    pixColor.g =  pow(pixColor.g, 1.0 / brightness);

    pixColor.b =  pow(pixColor.b, 1.0 / brightness);

```

Visual-Dawg:
I updated the contrast based on @TangerineGameDev formular (which looks really nice) and adjusted the brightness.
Now increasing the brightness happens before the blur, but decreasing happens after the blur.
The clipping from increasing it gets blurred, and the color boost still gets good image data for boosting colors.

Visual-Dawg:
I removed all config options except `saturation_boost`, which by default is `0.0` and wont do anything. 
If increased, the blurred colors will get stronger.

`brightness_boost` has been removed as it does not make much sense with the new `hsl` mode.

I consider this to be ready form merging now.

Visual-Dawg:
@TangerineGameDev Sorry, I phrased it badly :D 
All the previous options like `contrast` and `brightness` are still there, I just removed `brightness_boost` and `boost_colors`. 
`brightness_boost` did not do much with `hsl`, as it quickly just turns white. 
`boost_colors` is then not nessecary anymore, as you can also set `saturation_boost` to something other than `0` to enable it :)

vaxerski:
finally this doesn't look like shit.

image

Actually not bad :)

vaxerski:
skimming over the shader code - definitely needs some cleanup though, indentation errors, no spacing in some places in formulas.

vaxerski:
another comment is could we make the threshold configurable? on some darker wallpapers like this one sat_boost even on 3 doesnt do shit.

a228da8724649dbb5b4a66116de34d8c

vaxerski:
we could make this default on now, as it doesn't mess up anything from my testing, adds a nice small touch.

I'd propose a default for saturation_boost to `0.1696` which is `sqrt(2pi) / (2(e^2))`

Visual-Dawg:
Thank you! 
I might not get around it today but this are going to be my next steps:
- Add the new default value for `saturation_boost`.
- Run the shader code through Clang.
- Figure out a good way to create a `saturation_threshold` as a float (and come up with a good name for it, as it is more of a `lightness_threshold`.

Visual-Dawg:
I added the requested changes. The new `boost_threshold` takes in a float between `0` and `1`. 
`1` is the default value. The lower the more the boost starts to effect darker areas.


-----
`boost_threshold`: `0.3`
`saturation_boost`: `0.3`

image
-------

No boost

image

Visual-Dawg:
@TangerineGameDev Im not sure if I understand. `saturation_boost 0.0` means no boosting. Everthing higher will increase the colors.
And for the threshold, the default is `1`. If you lower the threshold, you start effecting darker tones too. 
Have I understood you correctly? :D

TangerineGameDev:
> @TangerineGameDev Im not sure if I understand. `saturation_boost 0.0` means no boosting. Everthing higher will increase the colors.
> 
> And for the threshold, the default is `1`. If you lower the threshold, you start effecting darker tones too. 
> 
> Have I understood you correctly? :D

Oh I think I got it now...
So the `boost_threshold` affects the darker colors while `saturation_boost` is just in general?
The way I understood you was that `1.0` is the default and a value of `0.3` boosts the color values...

Visual-Dawg:
@TangerineGameDev Yes, kinda. By default dark colors will be left untouched from the saturation boost. However the threshold for those can now be adjusted `boost_threshold 0.0` will allow `saturation_boost` to affect nearly all colors.
So if `saturation_boost` is `0`, `boost_threshold` wont do anything, as it only adjust the behavior of the former.

I think the variable name for `boost_threshold` might not be very good though. 
Do you have an idea for a better one? :D

TangerineGameDev:
Maybe `darkness_chroma_factor`?
Or just `darkness_chroma`

Visual-Dawg:
I renamed to `vibrancy` and `vibrancy_darkness`. 

"Vibrancy" is nice and short and commonly used for similar color operations. 

`vibrancy_darkness` now does not affect anything when set to `0` and "officially" goes up to `1`. 
In that way the user does not need to invert the logic for setting the threshold (decreasing the value to increase the effect on darker areas). 
Also it has the added side-effect that inserting anything higher than `1` increases the strength even more (it doesnt look very good though - but might be desirable for some setups)

vaxerski (approved):
lgtm, looks pretty nice on my hentai wallpapers.

If there isn't anything left I am happy to merge, lmk

Visual-Dawg:
Haha, good to hear xDD
From my side I am happy with the current state. The api is nice and the defaults look good.

PR #3664, opened 2023-10-26
input: Apply scaling to cursorPosOnActivate position

teschnei:
When the cursor is reset in a constraint, the hint given to us is scaled by the monitor's scaling value.  If we don't get a hint, the backup position (cursorPosOnActivate) should also be scaled so that the position is correct when we unscale it later.


vaxerski (approved):
my bad I did an ooga booga

PR #3682, opened 2023-10-28
hyprctl: add handler to report all connected monitors

TheAngusMcFire:
I added a way to show all monitors, not just the active ones.
I use this to see which monitor is connected to which output before changing the config. 

nope

ready for merge


vaxerski:
I'd rather add a flag to hyprctl and a field called `enabled` or something. This just adds tons of duplicate code.

TheAngusMcFire:
> I'd rather add a flag to hyprctl and a field called `enabled` or something. This just adds tons of duplicate code.

Would the code not be almost the same or even more? we iterate a different collection and also print other information (because disconnected monitors have less info)

I mean I can change it to use a field, should I then try to reuse the current print function and just add fallbacks for the not existing code?

vaxerski:
yes

PR #3413, opened 2023-09-24
add movegroupfocus command

svensp:
#### Describe your PR, what does it fix/add?
new command: movegroupfocus

like movefocus but moves through focused groups before continuing to the next window

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
I'm really rusty with C++ so probably not the most optimal code out there but I hope it is readable enough.

#### Is it ready for merging, or does it need work?
I just started using it myself and I haven't found any problems yet.

PR #3683, opened 2023-10-28
hyprctl: return group list in correct order

topisani:
The list of clients in the group returned by hyprctl always starts from the current client. This fixes it to start from the first client in the group.

Context:
I wanted to use hyprctl to implement a keybind to move focus within a group, or out of the group if already at either end (similar to i3/sway tabs). This could probably be implemented as a dispatcher at a later stage.

Tested without issues, should be ready to merge.


MightyPlaza:
unrelated to PR but isn't that already implemented as `movewindoworgroup` ?

topisani:
> unrelated to PR but isn't that already implemented as `movewindoworgroup` ?

That will always move the window out of the group - i want to first move the window to the left in the group, and if it doesnt go further left, move it out

Not the pretiest script, but this is what i do:
```
#! /bin/sh

action=$1; shift
dir=$1; shift

is_first() {
    hyprctl activewindow -j | jq -e ".address == .grouped[0] or .grouped == []" 
}

is_last() {
    hyprctl activewindow -j | jq -e ".address == .grouped[-1] or .grouped == []" 
}

case "$action $dir" in
    "move l")
        if ! is_first; then
            hyprctl dispatch movegroupwindow b
        else
            hyprctl dispatch movewindoworgroup l
        fi
        ;;
    "move r")
        if ! is_last; then
            hyprctl dispatch movegroupwindow f
        else
            hyprctl dispatch movewindoworgroup r
        fi
        ;;
    "focus l")
        if ! is_first; then
            hyprctl dispatch changegroupactive b
        else
            hyprctl dispatch movefocus l
        fi
        ;;
    "focus r")
        if ! is_last; then
            hyprctl dispatch changegroupactive f
        else
            hyprctl dispatch movefocus r
        fi
        ;;
esac
```

Additionally, i think i will use this to essentially move my groupbar into waybar. That would also require me to know the position of the client in the group

andresilva:
I guess this would make #3413 unnecessary, since it can be implemented as a script (as you have shown).

topisani:
I think it might make sense to restructure a lot of the dispatchers to take an optional group handling flag. While scripting is nice, I think some of this navigation with(in) groups is important enough to have in.

For instance, the next thing I may want to do is a movetoworkspace dispatcher that doesn't send the whole group, but just the current window

andresilva:
> For instance, the next thing I may want to do is a movetoworkspace dispatcher that doesn't send the whole group, but just the current window

This would be great indeed! I assume, like me, you are coming to Hyprland from i3 (or sway).

vaxerski (approved):
lgtm

PR #3693, opened 2023-10-30
refactor: create a function for grouping conditions

MightyPlaza:
creates `canBeGroupedInto(CWindow*)` to check whether a window can be grouped or not
simply moves duplicated code
didn't find any and shouldn't cause any issues

function might need a better name


vaxerski (approved):
much needed

PR #3700, opened 2023-10-30
meson: enable ndebug on release builds

andresilva:

Disables asserts on release builds. Hyprland uses it's own assert macro that is unaffected by this, but it could apply to dependencies.

No.

Yes.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,6 +5,7 @@ project('Hyprland', 'cpp', 'c',
     'default_library=static',
     'optimization=3',
     'buildtype=release',
+    'b_ndebug=if-release',
     'debug=false'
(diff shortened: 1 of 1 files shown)


andresilva:
Build fails due to `Werror=unused-variable` because of this: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/render/pass.c#L32-35. `box` is only used for an assert which is now a no-op 🤷.

andresilva:
wlroots [recommends](https://gitlab.freedesktop.org/wlroots/wlroots/-/wikis/Packaging-recommendations) against enabling `ndebug` (there's also this: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3338). So I'll close this and if @fufexan disagrees we can reopen it.

PR #3708, opened 2023-10-31
fixes for dragging groups onto groupbars

MightyPlaza:
fixes issues when dragging multiple windows group into a groupbar

there is an issue when dragging a window out a group
seems to be caused by https://github.com/hyprwm/Hyprland/commit/7f35f33b4ca16509e127976772db6abd9abf100b
window does gets stretched (likely from the hacky way it's being handled) and does not get redrawn
not really related to this PR, but apart from that didn't find any other issues

### DIFF
--- a/src/render/decorations/CHyprGroupBarDecoration.cpp
+++ b/src/render/decorations/CHyprGroupBarDecoration.cpp
@@ -314,13 +314,14 @@ bool CHyprGroupBarDecoration::onEndWindowDragOnDeco(CWindow* pDraggedWindow, con
     const int   WINDOWINDEX  = BARRELATIVEX < 0 ? -1 : (BARRELATIVEX) / (m_fBarWidth + BAR_HORIZONTAL_PADDING);
 
     CWindow*    pWindowInsertAfter = m_pWindow->getGroupWindowByIndex(WINDOWINDEX);
+    CWindow*    pDraggedHead       = pDraggedWindow->m_sGroupData.pNextWindow ? pDraggedWindow->getGroupHead() : pDraggedWindow;
 
-    g_pLayoutManager->getCurrentLayout()->onWindowRemoved(pDraggedWindow);
+    g_pLayoutManager->getCurrentLayout()->onWindowRemovedTiling(pDraggedWindow);
 
     pWindowInsertAfter->insertWindowToGroup(pDraggedWindow);
 
     if (WINDOWINDEX == -1)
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
What's broken is what leads us all astray.

src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -314,13 +314,14 @@ bool CHyprGroupBarDecoration::onEndWindowDragOnDeco(CWindow* pDraggedWindow, con
     const int   WINDOWINDEX  = BARRELATIVEX < 0 ? -1 : (BARRELATIVEX) / (m_fBarWidth + BAR_HORIZONTAL_PADDING);
 
     CWindow*    pWindowInsertAfter = m_pWindow->getGroupWindowByIndex(WINDOWINDEX);
+    CWindow*    pDraggedHead       = pDraggedWindow->m_sGroupData.pNextWindow ? pDraggedWindow->getGroupHead() : pDraggedWindow;
 
-    g_pLayoutManager->getCurrentLayout()->onWindowRemoved(pDraggedWindow);
+    g_pLayoutManager->getCurrentLayout()->onWindowRemovedTiling(pDraggedWindow);
```
vaxerski, line 319: wait a second, dragged windows are floating?

MightyPlaza: that's used after window gets placed by the layout to remove the node
`onWindowRemoved` was removing grouped properties and messing stuff up
`onWindowRemovedTiling` does everything the same except removing the grouped properties
so now everything seems to work properly

vaxerski: I don't like this. This may silently break very easily in the future.

MightyPlaza: I don't like it either, but we can't access the layout from the outside to delete the nodes

vaxerski: we can save the group state tho

MightyPlaza: we could but we would still be using the same function (wrapped inside another one)
not sure if you mean the unnecessary placing and removing of windows or bypassing `onWindowRemoved()`

vaxerski: I'd want to call `onWindowRemoved` instead

vaxerski (approved):
lgtm

PR #3713, opened 2023-10-31
Sigh unsafe stuff take 3

vaxerski:
Stuff

re: #3637

PR #3718, opened 2023-11-01
Fix sending wl "finished" events on monitor disabling. Fixes #1274

### LINKED ISSUE #1274: Disabling a monitor causes it to no longer be detected
Disabling through either kanshi, wlr-randr, or hyperctl makes the monitor undetectable through either.
So for example if I run `hyprctl keyword monitor LVDS-1, disable` , the monitor `LVDS-1` can no longer be found through: `hyprctl monitors`, or `wlr-randr`. Thus making it impossible to re-enable. Perhaps the issue that is being alluded to all the way at the end of #453. Perhaps this is even introduced by the fix?

running latest git version.

levnikmyskin:
This PR should fix the issues in #1274, and in general should allow the usage of tools like `kanshi` or `shikane` for monitor management.  
Basically, now Hyprland does not send `finished` events when a monitor is disabled, allowing `kanshi`-like tools to hold references to the monitor resource.

This is still in draft because more things should probably change. Notably:

- Disabled but not really removed monitors (like laptop built-in screens) are not deleted anymore from `realMonitors` and `monitors`. Does this cause any problem?
- I believe it would make more sense to avoid entering an unsafe state at all, if we have a disabled but still phisically attached monitor. For instance, we could simply re-enable the monitor (despite the user previously disabled it) and avoid unsafe states. This would also probably fix a lot of issues with unsafe state crashing on laptops.
- At the moment, unsafe states are still created, but without a headless monitor. This is because otherwise, in shikane/kanshi config, we would need to have profiles considering both the laptop screen and the headless monitor, which is undesirable imo.

As mentioned, it needs work. Also, I only tested with shikane but not with kanshi.

@vaxerski Sorry to tag you directly, but since this PR is changing core stuff, I would really appreciate your help :)


vaxerski:
we're changing some stuff with this in #3713

levnikmyskin:
Ok noticed. The main thing changing now is that the headless output is created at the beginning, so tools like `shikane` will probably see them. I'll check whether this causes "unexpected" behaviours when monitors are enabled with these tools (i.e., if you have 2 monitors, they'll see 3 and profile matching will fail).

That said, regarding monitor disabling, I think the best option would be to not go into unsafe state at all, and just enable whatever monitor is still in `realMonitors`. What do you think?

levnikmyskin:
I confirm that with the new code tools like kanshi/shikane will not match any profile, unless one explicitly adds the headless output in their config files :/ Notice this is not related to monitor disabling only, but also to a "normal" situation where all monitors are connected and enabled.

leiserfg:
No idea if it's related to this change of something else near it, but a now a big portion of my screen started to flicker

levnikmyskin:
> No idea if it's related to this change of something else near it, but a now a big portion of my screen started to flicker

Is this happening with this PR or with Hyprland's main branch in general?

leiserfg:
> > No idea if it's related to this change of something else near it, but a now a big portion of my screen started to flicker
> 
> Is this happening with this PR or with Hyprland's main branch in general?

Sorry but I had to close hyprland,  the blinking was making me feel bad. Now I'm back in x11 (where there is no blinking, but tearing). I can't say for real if the issue was this branch only or master as it happened only after I tried modifying the screens with `wdisplay`.

levnikmyskin:
A couple of comments on the status of this:

1. Since disabled (but still physically connected monitors, like laptop's) monitors are not removed from the vector structures, this now re-enables automatically laptop's screen when there's no other backup monitor;
   - this means that laptops never go into unsafe states;
   - a new check should actually be inserted, to first use enabled monitors as backup, and disabled as last resort (I'll do this the moment Vaxry agrees with the changes);
 3. several small changes were made in order to deal with this. In general, I believe it would be better if the code now considered the possibility of a disabled (but not physically detached monitor): i.e., dealing better with monitor re-enabling (and with monitors present in the vector structures, that are actually disabled);
 4. regarding the new changes to the headless backup monitor...that breaks compatibility with other wl-roots tools that manage monitors, so if possible I would go back to the previous behaviour (I know this sucks as a proposal).

What are your feelings about this @vaxerski ?

vaxerski:
1. no
2. disabled = not in m_vMonitors but in m_vRealMonitors
3. no. Sway does the same, what's your issue?

levnikmyskin:
So, I tried to understand what Sway was doing different than Hyprland (considering the changes already done in this branch), and the thing mainly was that Hyprland was calling `wlr_output_layout_remove` on both monitor destroy and disable.  
Sway instead does not call this when the monitor is destroyed. The difference is that, if `wlr_output_layout_remove` is called, then the monitor is removed from `wlr_output_manager_v1` resources.  
The consequence is then that, when calling `wlr_output_manager_v1_set_configuration` after a monitor is destroyed, the [done](https://wayland.app/protocols/wlr-output-management-unstable-v1#zwlr_output_manager_v1:event:done) event is not sent for the removed monitor. This caused kanshi (but not shikane) to not update the profile.  
  
Sorry for the wall of text, but I think this might be helpful in the future. 
Coming back to your reply @vaxerski :  

1. fair enough, I removed that;
2. indeed, my bad;
3. sway does the same indeed. For me it was causing issues, but it was a "bug" on my branch. This is fixed now.

Given all this, the PR should be ready, I'll promote it to a real one. Please let me know if there's anything wrong here

src/helpers/Monitor.cpp
```diff
@@ -276,7 +276,8 @@ void CMonitor::onDisconnect() {
 
     activeWorkspace = -1;
 
-    wlr_output_layout_remove(g_pCompositor->m_sWLROutputLayout, output);
+    if (!destroy)
```
vaxerski, line 279: are you sure this shouldn't be flipped?

vaxerski: my bad misunderstood your comment

vaxerski (approved):
yea this is fine

vaxerski:
thanks!

leiserfg:
@vaxerski  don't you think this merits a patch release?

vaxerski:
?

leiserfg:
This fix I mean, it's not included in the last stable release.

vaxerski:
it is

leiserfg:
Sorry to bother you, and thank you for all your work, but  https://github.com/hyprwm/Hyprland/compare/v0.32.3...main  says different :wink:

vaxerski:
my bad, this will land in the next release then, when it's time for the next release.

PR #3734, opened 2023-11-03
Master layout: fix drop_at_cursor when there are only two windows

thejch:
drop_at_cursor currently doesn't work when there are only two windows in the workspace (it always makes the dropped window the master even when it's dropped on the stack's side of the screen).
The pr fixes this so that when the window is dropped on the stack's side, it becomes a stack window instead. If it's dropped on the master's side, it becomes the master.

Also fixes drop_at_cursor when new_is_master is enabled. Previously didn't work.

No

Ready


vaxerski (approved):
to show you the power of flex paste

PR #3771, opened 2023-11-05
fix box headers

brettalcox:
#### Describe your PR, what does it fix/add?
Fixes the undefined reference to wlr_box_transform in Box.cpp

#### Is it ready for merging, or does it need work?

Ready for merge

PR #3772, opened 2023-11-05
Fix Compile error on non PCH system

darkstack:
By Default on Gentoo, PCH is masked for GCC.

this PR fix the compilation issue on these systems.


darkstack:
Fixed in #3771

PR #3800, opened 2023-11-08
Decoration Positioner

vaxerski:
Adds a decoration positioner.

Positioning can be absolute (like right now) or sticky, aka it sticks to an edge (like hyprbars / groupbar)

Sticky decos get nicely positioned:
20231108_12h11m59s

CC @MightyPlaza (deco mrs) @outfoxxed (hy3) @fufexan (review) 

I also thought about relative decos but they are kinda like absolute ones and I cba to do collision detection and shit for them. Seems pointless anyways.

Needs some battle testing.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -81,6 +81,7 @@ CCompositor::CCompositor() {
 
 CCompositor::~CCompositor() {
     cleanup();
+    g_pDecorationPositioner.reset();
     g_pPluginSystem.reset();
     g_pHyprNotificationOverlay.reset();
     g_pDebugOverlay.reset();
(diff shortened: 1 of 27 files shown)


outfoxxed:
In hy3's case I don't use window decorations because hy3's tabs operate on a node basis, not a window basis, where a node can be either a window or collection of other nodes. After quickly skimming the diff I don't think this does anything for my usecase.

Example of a bar spanning a node that contains multiple windows:
image

Relevant code:
https://github.com/outfoxxed/hy3/blob/4db9b1f37a5f429aa5d103143176ba2f5dd20f16/src/Hy3Layout.cpp#L1292-L1333

MightyPlaza:
should we have a function for specifically reloading the config in `CDecorationPositioner`?
or maybe an extra arg on `CDecorationPositioner::onWindowUpdate`
similar to what was done in https://github.com/hyprwm/Hyprland/pull/3546/files with `refreshGroupBarGradients`
this could be used to do a harder than `onWindowUpdate` reload and avoid doing config dependent-only calculations every `updateWindow`

vaxerski:
why would we? config values can be sourced every render pass.

MightyPlaza:
> why would we? config values can be sourced every render pass.

yes but there are some stuff that only needs to be run once, each time the config changes
for example creating the textures in the groupbars

vaxerski:
that can be checked and called inside render, or on `configReloaded`

vaxerski:
image

Everything seems good to me.

vaxerski:
I'll merge if checks pass, no major comments have been raised

PR #3831, opened 2023-11-11
Small optimization in rgb2hsl.

Dickby:
In the old code, if `delta == 0.0` then deltaGtz would be 0.0 and hue gets multiplied with it.
That makes all the other operations (including `dot()`) useless.
So let's do the whole thing only if delta has a value > 0.0

Ready i think.


Dickby:
What am i doing wrong for those checks to fail?

vaxerski:
nothing. Arch container from docker is shitting itself.

vaxerski (approved):
lgtm

PR #3834, opened 2023-11-11
Some more changes in rgb2hsl.

Dickby:
1. also calculate `marks` and `adds` only if they are used.
2. use glsl vector functions to build the masks vector and avoid multiple branches.
3. change the `adds` initialization to divide the Vector by `delta` at once
4. add missing new line between functions

I think it's ready.


Eclextic:
What the f is the difference between `equal` and `==`?
Doesn't the compiler transform them anyways?

Dickby:
@Eclextic 
equal works on vectors e.g
equal(vec3(1.0,2.0,3.0), vec3(2.0,2.0,2.0))  produces a bool vector (0 , 1 , 0)

Eclextic:
ooooh! I didn't know that!
So we can get a bool for every element in the vector separately!
Thanks for that info, that's awesome!

vaxerski (approved):
coolio

Issue #3851, opened 2023-11-14
Incorrect rendering of groupbar titles with a scaled display


q234rty:
### Hyprland Version

Hyprland, built from branch main at commit 5edb4e4a304bbef1de30a36f7b921ea328d7efc6 dirty (decorations: recalc layout and positioner on add/remove).

### Bug or Regression?

Bug

### Description

Rendering of groupbar titles seems to be incorrect under a scaled display.
 

### How to reproduce

1. Use a scale > 1, such as 2 or 1.75
2. Write `group:groupbar:render_titles = true` in hyprland.conf
3. `hyprctl dispatch togglegroup`
4. Observe that rendering of groupbar titles is broken

### Crash reports, logs, images, videos

![图片](https://github.com/hyprwm/Hyprland/assets/11485439/5669c8b7-3115-41ae-a907-11c2c6cee5cb)

PR #3856, opened 2023-11-14
Fix position of groupbar titles on monitor scales != 1.0

Dickby:
Fixes https://github.com/hyprwm/Hyprland/issues/3851

Test it before merging, but i think it's ready to go

### DIFF
--- a/src/render/decorations/CHyprGroupBarDecoration.cpp
+++ b/src/render/decorations/CHyprGroupBarDecoration.cpp
@@ -118,8 +118,9 @@ void CHyprGroupBarDecoration::draw(CMonitor* pMonitor, float a, const Vector2D& 
 
         // render title if necessary
         if (*PRENDERTITLES) {
-            CBox       rect = {ASSIGNEDBOX.x + xoff - pMonitor->vecPosition.x + offset.x, ASSIGNEDBOX.y - pMonitor->vecPosition.y + offset.y + BAR_PADDING_OUTER_VERT, m_fBarWidth,
+            CBox rect = {ASSIGNEDBOX.x + xoff - pMonitor->vecPosition.x + offset.x, ASSIGNEDBOX.y - pMonitor->vecPosition.y + offset.y + BAR_PADDING_OUTER_VERT, m_fBarWidth,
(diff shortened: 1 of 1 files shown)


Dickby:
The groupbar titles seem a bit lower on non scaled monitor.
I'll look into that

Dickby:
Btw. is there a way to `make debug` without the `-B` flag, so that not everything is recompiled every time?

Dickby:
It looks ok for me like this.
The groupbar text seems to be not scaled though.
But without this PR the groupbar titles on scale are totally broken

src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -131,13 +132,10 @@ void CHyprGroupBarDecoration::draw(CMonitor* pMonitor, float a, const Vector2D&
...
-            if (*PGRADIENTS) {
-                CBox rect2 = rect;
-                rect2.scale(pMonitor->scale);
-                g_pHyprOpenGL->renderTexture((m_dwGroupMembers[i] == g_pCompositor->m_pLastWindow ? m_tGradientActive : m_tGradientInactive), &rect2, 1.0);
-            }
+            if (*PGRADIENTS)
+                g_pHyprOpenGL->renderTexture((m_dwGroupMembers[i] == g_pCompositor->m_pLastWindow ? m_tGradientActive : m_tGradientInactive), &rect, 1.0);
 
-            rect.y      = ASSIGNEDBOX.y + ASSIGNEDBOX.h / 2.0 - (*PTITLEFONTSIZE + 2 * BAR_TEXT_PAD) / 2.0;
+            rect.y      = (ASSIGNEDBOX.y + BAR_TEXT_PAD) * pMonitor->scale;
```
Dickby, line 138: I'm not shure about this line, because i'm not really shure what the replaced line was doing.
ASSIGNEDBOX.h / 2.0   
is just the half size of the box,
(*PTITLEFONTSIZE + 2 * BAR_TEXT_PAD) / 2.0
is half the font size + the bar text padding, isn't that just the the half size of the box again ?

Dickby:
Back to my first approach.
The bug came from not scaling the rect for the groupbar title.
Since rect2 and rect have to be scaled, get rid of rect2 and use rect for both.

vaxerski (approved):
lgtm thanks!

PR #3867, opened 2023-11-16
Don't use wrong matrix just because it's GLES

Dickby:
I don't use GLES, so it's untested!


vaxerski (approved):
pff lmao someone missed this

damn it @Visual-Dawg

Issue #827, opened 2022-10-10
Floating window with special workspace


thomascft:
When moving a floating window to the special workspace it floats above other clients and doesn't tile. This could be fixed by switching the window to tiling somewhere through the process, but having a check before moving to the special workspace would be good.

vaxerski:
I was thinking about allowing floats in special in general.

thomascft:
Ok, I think the windows in the special need to be resizable somehow.

ttmx:
This would be really nice, it would allow for easy "dropdown" windows, which I would say are very useful.
Thank you

thomascft:
> This would be really nice, it would allow for easy "dropdown" windows, which I would say are very useful.
> Thank you

I didn't do the best job at phrasing my initial issue but the special workspace essentially creates a second workspace overlaying the current. My issue was when sending a floating window to the special workspace. Upon toggling the special you'll find the floating window above all the other special windows often covering them as if on a third layer. When trying to resize or move the special window it's placed on the active normal workspace

TheSunCat:
I am experiencing this issue, except moving/resizing the floating window in the special workspace does not tile it, nor does executing the `toggletiling` bind. The only way I've found to get it out of that state is to move it to another workspace.
I'm not opposed to having floating windows in the special workspace, though I think the default behavior being tiled works quite well (if this bug didn't exist).
I can trigger it by making any window floating, then moving it to the special workspace using `movetoworkspace,special`.

MightyPlaza:
@vaxerski

PR #3872, opened 2023-11-17
Allow floating windows on special

### LINKED ISSUE #2271: Feature request: Provide a dispatcher/variable to make floating windows tiled while on the special workspace
## What should hapen
When a floating window is sent to the special workspace, it should become tiled instead of float. When the same window is send out of the special workspace, it should recover its float state.

## What actually hapen
When a float window is sent to the special workspace, it stays float. You cannot make it tiled unless you send it back to a workspace, manually make it tiled, and then send it again to the special workspace. When you want to send it out of the special workspace, you will have to manually undo the same operation. Which kinda cuts the flow and is prone to mistakes.

## More info
Related to #2161. We said we want to keep the current behavior for users who want to manually fill their special workspace with floating windows. But we also want to provide an alternative for people who want float windows to go 'unfloat' while on the special workspace. Which is likely the most common case of use.

vaxerski:
Allows floating windows on the special workspace.

Supersedes #2624
Fixes #2271 #827 

cc @fufexan review


vaxerski:
next up: allowing fullscreen / maximized

vaxerski:
fuck it we ball

Issue #3574, opened 2023-10-15
pseudo window get resized on every drag


imxyy1soope1:
### Hyprland Version

Hyprland, built from branch main at commit 424c9a7e704590db5c823557e5e388e366f7b1cd dirty (render: stop locking SWC on tearing). Tag: v0.30.0-86-g424c9a7e  flags: (if any)

### Bug or Regression?

Bug

### Description

see the video

### How to reproduce

see the video

### Crash reports, logs, images, videos

https://github.com/hyprwm/Hyprland/assets/103114856/6618e166-e23f-4869-afd0-6a1dd7d55c80

PR #3873, opened 2023-11-17
Don't update pseudoSize after window moved by mouse. 

Dickby:
Fixes https://github.com/hyprwm/Hyprland/issues/3574
Prevents a pseudoTiled window size gets changed after moving it with the mouse. 
That size change makes sense when a floating window is turned to a pseudoTiled, but not after a mouse move.

I tested it, and it seems to work.
I think it's ready, but double check please.


vaxerski (approved):
lgtm

imxyy1soope1:
thx!

PR #3877, opened 2023-11-18
Workspace gaps

end-4:
- Space between workspaces in the animation
- Config variable: `animations:workspace_gap` [int]
image

- nope

- ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -208,7 +208,7 @@ void CConfigManager::setDefaultVars() {
     configValues["master:smart_resizing"].intValue         = 1;
     configValues["master:drop_at_cursor"].intValue         = 1;
 
-    configValues["animations:enabled"].intValue = 1;
+    configValues["animations:enabled"].intValue       = 1;
     configValues["animations:workspace_gap"].intValue = 0;
 
     configValues["input:follow_mouse"].intValue                     = 1;
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -79,6 +79,7 @@ void CConfigManager::setDefaultVars() {
     configValues["general:no_border_on_floating"].intValue = 0;
     configValues["general:gaps_in"].intValue               = 5;
     configValues["general:gaps_out"].intValue              = 20;
+    configValues["general:gaps_workspace"].intValue        = 0;
     ((CGradientValueData*)configValues["general:col.active_border"].data.get())->reset(0xffffffff);
     ((CGradientValueData*)configValues["general:col.inactive_border"].data.get())->reset(0xff444444);
     ((CGradientValueData*)configValues["general:col.nogroup_border"].data.get())->reset(0xff444444);
@@ -208,8 +209,7 @@ void CConfigManager::setDefaultVars() {
     configValues["master:smart_resizing"].intValue         = 1;
     configValues["master:drop_at_cursor"].intValue         = 1;
 
-    configValues["animations:enabled"].intValue       = 1;
-    configValues["animations:workspace_gap"].intValue = 0;
+    configValues["animations:enabled"].intValue = 1;
 
     configValues["input:follow_mouse"].intValue                     = 1;
     configValues["input:mouse_refocus"].intValue                    = 1;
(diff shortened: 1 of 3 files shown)


src/config/ConfigManager.cpp
```diff
@@ -208,7 +208,8 @@ void CConfigManager::setDefaultVars() {
     configValues["master:smart_resizing"].intValue         = 1;
     configValues["master:drop_at_cursor"].intValue         = 1;
 
-    configValues["animations:enabled"].intValue = 1;
+    configValues["animations:enabled"].intValue       = 1;
+    configValues["animations:workspace_gap"].intValue = 0;
```
vaxerski, line 212: does this really belong to animations: ?

end-4: Yes, because the gap is only shown when switching workspaces. I only modified workspace change anims.
An animation option, technically at least. 
Do you think we should have this in general or something?

Eclextic:
Does this PR work additively with gaps_out?
I imagine so as it would complicate things a lot if that's not the case, but would be appreciated!

end-4:
> Does this PR work additively with gaps_out?
- workspace_gap and gaps_out are separate
image

Eclextic:
@end-4 So it does work additively? Guess I expected it...

end-4:
- yes
> would be appreciated if it wasn't!

- about this, it adds a calculation which i don't think will very likely be meaningful. can you explain why this could be a good idea?

Eclextic:
> * yes
> 
> 
> > would be appreciated if it wasn't!
> 
>     * about this, it adds a calculation which i don't think will very likely be meaningful. can you explain why this could be a good idea?

I could imagine, if someone has `gaps_out` set to something big like 20, then if I don't want it to be visible

end-4:
> if someone has gaps_out set to something big like 20, then if I don't want it to be visible

just set workspace gap to 0? that's why there's a config value

Eclextic:
If gaps_out is set to something big like 20, then if I want a smaller gap when animating that gets bigger after the animation has finished, that wouldn't be possible...

If that behaviour is necessarily needed is debatable, but I suppose one could use a `gaps_out` with 5 and still want the animations to be squished next to each other...

So yea, like I said, it would complicated things, but is a nice-to-have 👍🏼

end-4:
> one could use a gaps_out with 5 and still want the animations to be squished next to each other...

i see... sounds like the swipe up overview on mobile devices.

Eclextic:
I'm just making a suggestion btw...
If this is too hard to accomplish it's fine!

end-4:
Not really something I'm interested to implement right now. I might look at  this later and figure out if it can be even more polished than that (I mean, I'd really like that effect to come with some zooming too)

Visual-Dawg:
Thats quite cool

Visual-Dawg:
Personally, I dont like that this is in `animations` - while it is only visible during an animation, it is not part of the "animation mental model", which adjusts animation parameters, but more of general layout, like window gaps. 

Also who knows, maybe in the future this will not only affect animations

end-4:
vaxry still hasn't responded in the review lol. but i guess that does make sense. i'll change it

src/config/ConfigManager.cpp
end-4: moved to general:gaps_workspace in the latest commit

Eclextic:
> Personally, I dont like that this is in `animations` - while it is only visible during an animation, it is not part of the "animation mental model", which adjusts animation parameters, but more of general layout, like window gaps.

Honestly we should probably have a separate category for workspaces... especially because we have so many of these config options for them.

But this is for another PR to do XD.
Maybe I'll do it? Been thinking about contributing for some tome lol... Could be a good first contribution.

MightyPlaza:
> Honestly we should probably have a separate category for workspaces... especially because we have so many of these config options for them.
workspace rules

vaxerski (approved):
lgtm

Visual-Dawg:
> workspace rules
Those are for individual workspaces though. 
For now, having them in general makes the most sense, imo.

But yes @Eclextic is right, there is probably a better way

PR #3895, opened 2023-11-19
Add hyprctl 'layouts' command

zakk4223:
Adds a 'layouts' command to hyprctl for listing available layouts

This makes m_vLayouts public (obvious from commit just thought I'd mention it)
It also doesn't show the value of getLayoutName() since nothing uses that anyways.

Should be ready

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -52,6 +52,7 @@ commands:
     notify
     globalshortcuts
     instances
+		layouts
 
 flags:
     -j -> output in JSON
@@ -420,6 +421,8 @@ int main(int argc, char** argv) {
         request(fullRequest, 2);
     else if (fullRequest.contains("/hyprpaper"))
         requestHyprpaper(fullRequest);
+    else if (fullRequest.contains("/layouts"))
+        request(fullRequest);
     else if (fullRequest.contains("/--help"))
         printf("%s", USAGE.c_str());
     else {
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -52,7 +52,7 @@ commands:
     notify
     globalshortcuts
     instances
-		layouts
+    layouts
 
 flags:
     -j -> output in JSON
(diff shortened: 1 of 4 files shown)


hyprctl/main.cpp
```diff
@@ -52,6 +52,7 @@ const std::string USAGE = R"#(usage: hyprctl [(opt)flags] [command] [(opt)args]
     notify
     globalshortcuts
     instances
+		layouts
```
vaxerski, line 55: we use spaces for indent

src/managers/LayoutManager.hpp
```diff
@@ -13,6 +13,7 @@ class CLayoutManager {
 
     bool         addLayout(const std::string& name, IHyprLayout* layout);
     bool         removeLayout(IHyprLayout* layout);
+    std::vector<std::pair<std::string, IHyprLayout*>> m_vLayouts;
```
vaxerski, line 16: I'd rather have a function that returns data about loaded layouts. m_vLayouts should stay privat

hyprctl/main.cpp
zakk4223: Betrayed by clang-format

src/managers/LayoutManager.hpp
zakk4223: I'll just do whatever the plugin shit does. That uses unique_ptr and returns the raw pointers. I have some layout plugins that need to hold references to layouts (because doing name lookups constantly is dumb) and I'm pretty sure unique_ptr+raw is unsafe. Can we used shared_ptr here and for stuff that holds layout ptrs they can just make weak_ptrs? Or is there a better way to deal with that?

vaxerski: return just the names

vaxerski (commented):
all good, just needs a clang-format :)

vaxerski (approved):
lgtm, thanks!

PR #3898, opened 2023-11-19
Use thunar as default file manager (with respect to archinstall)

kulothunganug:
## Problem

I find the file-manager `dolphin` which is set on the keybindings section in the config ([default](https://github.com/hyprwm/Hyprland/blob/646f4bc63861e03b01fed3833bb65d1d014e1bf3/example/hyprland.conf#L132)) to be bloated.
If someone is installing hyprland on minimal arch, it's installation can be decided to the user's preference.

But if someone like me installing/setuping arch using [archinstall](https://github.com/archlinux/archinstall/) with hyprland profile, it automatically installs `dolphin` which depends on too many kde dependencies...
It's a pain to remove `dolphin` and finding and removing it's dependencies, removing dolphin's deps using pacman leads to the warning of the dep is a dep of another packages and goes on...

Dependencies I need to find when removing `dolphin`:-

image

Dependencies of `dolphin` when installing it:-

image

## Fix/Improvement
We could request there ([archinstall](https://github.com/archlinux/archinstall/)  repo) to change dolphin -> other, But I think it would be inconsistent...
So I planned first requesting here to use `thunar` as default on hyprland, then opening a PR on [archinstall](https://github.com/archlinux/archinstall/) repo

Dependencies of `thunar` is minimum & can be uninstalled easily by [removing orphaned packages](https://www.reddit.com/r/archlinux/comments/kc4zq3/removing_orphans/):-

image


Eclextic:
First of all, I find that you will only install arch once anyways and this is kinda the struggle of arch...

Second of all, if we decide on an alternative file manager why not choose Nemo?
It's the most minimal GTK you can possibly get, while also being native to Wayland!

I've tried literally every graphical file manager and found Nemo to be the least bloated while staying the most functional!

Eclextic:
First of all, I find that you will only install arch once anyways and this is kinda the struggle of arch...
Meaning I am kind of against this...

Second of all, if we decide on an alternative file manager why not choose Nemo?
It's the most minimal GTK you can possibly get, while also being native to Wayland!

I've tried literally every graphical file manager and found Nemo to be the least bloated while staying the most functional!

kulothunganug:
Having experimented with Nemo, I find it minimal and fast and am now convinced. If the reviewer/maintainer consider it superior to Thunar, let's make the switch to Nemo.

PS: I haven't had much experience with PRs in the past, so I would welcome assistance or mentorship from others to successfully complete this one.

vaxerski:
if archinstall perfers a different file manager they can edit their profile.

You don't need to install a file manager at all for hyprland to run. 

Dolphin is there because I use it and guarantee it to work.

"Bloat" is a stupid buzzword that doesn't carry any worth to it. 50MB of packages is nothing in the current day and age.

TL;DR: No.

PR #3907, opened 2023-11-20
Fix floating clipbox scale

thejch:
Fix floating clipbox scale. Should fix the window getting clipped incorrectly during workspace transition? Not 100% sure but works for me

No

Ready


vaxerski:
you can use vecTransformedSize instead. It's the same.

thejch:
> you can use vecTransformedSize instead. It's the same.

added

vaxerski (approved):
I told you long ago, on the road

vaxerski:
wait, this needs to be rethinked.

vaxerski:
why are we adding the entire monitor here?

We should just translate it by the workspace offset

oops

thejch:
you can do it if you want. as long as the scale thing gets fixed lol

vaxerski:
I can't repro so I can just suggest you try this:
```cpp
CRegion rg = pWindow->getFullWindowBoundingBox().translate(-pMonitor->vecPosition).scale(pMonitor->scale);
rg.add(CBox{0, 0, pMonitor->vecSize.x, pMonitor->vecSize.y});
```

into

```cpp
CRegion rg = pWindow->getFullWindowBoundingBox().translate(-pMonitor->vecPosition + PWORKSPACE->m_vRenderOffset.vec()).scale(pMonitor->scale);
```

thejch:
yea seems to do it. I added that

vaxerski (approved):
top g thanks!

PR #3923, opened 2023-11-23
Make exec support compound bash commands

thejch:
Current behavior: exec/exec-once cannot run compound bash commands (if statements, loops, whatever the [[ ]] thing is called, etc) due to some hardcoded environment variables added in front.
These env vars are redundant because they are in `environ` already. Therefore, they are automaticlly used in `execl` anyway, so they should be safe to remove. 

no

ready

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -641,11 +641,6 @@ void CKeybindManager::spawn(std::string args) {
         args  = args.substr(args.find_first_of(']') + 1);
     }
 
-    if (g_pXWaylandManager->m_sWLRXWayland)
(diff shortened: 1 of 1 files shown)


vaxerski:
isn't that precisely the point of exec raw?

thejch:
But with execr you don't get rules?

thejch:
Another few things to consider:
1. The keyword `exec-once` uses spawn, not spawn raw, and there is no `exec-once-raw` as far as I know. Therefore, compound shell commands currently do not work with `exec-once`
2. In that regard, there is also no `execr` keyword. (there is only a `execr` dispatcher). So, `exec=compound command` doesn't work
3. The current code is not exactly following conventions, even if we were to add environment variables here, it should be using `setenv`.

vaxerski:
factual statements. Fair enough. Wiki MR needed though. (list of dispatchers, execr description would be outdated)

thejch:
added

Issue #3916, opened 2023-11-22
Drop shadow in special workspace be looking extra thicc


thejch:
### Hyprland Version

Hyprland, built from branch main at commit 812a3f6d78681abbfbc0b01f3b5956c38ae06c1c dirty (renderer: fix double render of tiled on workspace switch). Tag: v0.32.3-47-g812a3f6d

### Bug or Regression?

Bug

### Description

drop shadow in special workspace be looking extra thicc
maybe related to #3888 ?
also its not the dim. i disabled it

### How to reproduce

default config
run `hyprctl keyword decoration:shadow_range 50`
compare drop shadow in a special workspace and a regular workspace. it looks much thicker in special

vaxerski:
looks like double rendering

thejch:
seems to still happen

vaxerski:
??? was defo fixed for me. Repro steps??

thejch:
I can reproduce this every time i launch kitty in special. so i am not sure what to provide as repro steps.
but try the following so its more noticeable:
```sh
#!/bin/sh

hyprctl keyword misc:disable_hyprland_logo true
hyprctl keyword misc:background_color "rgb(ffffff)"
hyprctl keyword decoration:shadow_range 50
hyprctl keyword decoration:dim_special 0
hyprctl dispatch exec "[size 100 100;float;move 20% 20%] kitty"
sleep 1;
hyprctl dispatch workspace special
hyprctl dispatch exec "[size 100 100;float;move 50% 20%] kitty"

```

vaxerski:
check with above

thejch:
Seems partially fixed.
However, there's is a few new issues that I noticed.
1. have floating window in regular, and open special. now the floating window in regular is double rendered
2. following 1, now open a window in special, and close special, the regular window blinks a few times for some reason
3. and if the window in special is tiled, the window in regular gets rendered on top of it, while it should be under it
4. sometimes when you have a maximized window in regular, and you open up special, the special window gets double rendered during the animation

Dickby:
On my side, this seems to be fixed.

Issue #3889, opened 2023-11-18
Special workspace not visible during workspace switching


Dickby:
### Hyprland Version

Hyprland, built from branch makepkg at commit 8b57a1973ea8b242783069e9a3e3d31e3cd939a8 dirty (internal: Allow floating windows on special (3872)). Tag: v0.32.3-34-g8b57a197

### Bug or Regression?

Bug

### Description

Switching workspaces when a special workspace is visible, causes that special ws to hidden behind the windows on the non-special new workspace until the switching is finished.
Also, In my opinion switching a workspace with a visible special workspace should either keep the special focused (i would prefer that), or hide the special one.
The way it is now after switching the special is still visible but not focused, i think that is not usefull at all, and a state that can't be reached in other ways.

### How to reproduce

See description

Lillecarl:
confirmed 572fd554b8e86b3b3533f05631348c3374815ab5

It's especially entertaining when you have a window rule to fade in the windows 😸

Dickby:
@vaxerski after the last change, this issue is back too!

Dickby:
This seems to be fixed!
Thanks

PR #3928, opened 2023-11-24
Fixup double rendering cases with special

### LINKED ISSUE #3888: Transparency difference on special workspace
### Hyprland Version

Hyprland, built from branch makepkg at commit 8b57a1973ea8b242783069e9a3e3d31e3cd939a8 dirty (internal: Allow floating windows on special (3872)). Tag: v0.32.3-34-g8b57a197

### Bug or Regression?

Bug

### Description

Semi-Transparent windows are less transparent if they are in a special workspace.

### How to reproduce

Well, open a semitransparent window on a non-special, memorize how transparent it is, and then move it to a special workspace and compare

_No response_

vaxerski:
Aims to fix #3888 #3916 #3889

let me know if it works @Dickby @thejch

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -256,7 +256,7 @@ void CHyprRenderer::renderWorkspaceWindows(CMonitor* pMonitor, CWorkspace* pWork
         if (!shouldRenderWindow(w.get(), pMonitor, pWorkspace))
             continue;
 
-        if (w->m_iMonitorID == pWorkspace->m_iMonitorID && g_pCompositor->isWorkspaceSpecial(w->m_iWorkspaceID) && !pWorkspace->m_bIsSpecialWorkspace)
+        if (pWorkspace->m_bIsSpecialWorkspace != g_pCompositor->isWorkspaceSpecial(w->m_iWorkspaceID))
             continue;
 
         // render active window after all others of this pass
@@ -280,7 +280,7 @@ void CHyprRenderer::renderWorkspaceWindows(CMonitor* pMonitor, CWorkspace* pWork
         if (w->m_bIsFloating)
             continue; // floating are in the second pass
 
-        if (w->m_iMonitorID == pWorkspace->m_iMonitorID && g_pCompositor->isWorkspaceSpecial(w->m_iWorkspaceID) && !pWorkspace->m_bIsSpecialWorkspace)
+        if (pWorkspace->m_bIsSpecialWorkspace != g_pCompositor->isWorkspaceSpecial(w->m_iWorkspaceID))
             continue;
 
         if (!shouldRenderWindow(w.get(), pMonitor, pWorkspace))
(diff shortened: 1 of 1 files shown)


Dickby:
There seems to be only blur on regular-workspace-windows if the special workspace is visible.
With hidden special-ws the stuff that should be blured is just black

Dickby:
With either new_optimizations or xray enabled, blur works as expected, but both set to false leads to the problem is discribed above

vaxerski:
new_optimizations should be removed as it should NEVER be disabled. Code has expected it to be the case for a while now. Unrelated to this MR, anyways

Dickby:
So a tiled window that moves above another tiled window should never be visible in the blured alpha?

vaxerski:
tiled windows should not overlap.

Dickby:
But they do while window move animations

vaxerski:
that's an edge case that I don't want to fix for the sake of performance. It's barely noticeable, anyways.

vaxerski:
at any rate, that's unrelated to this MR. Lmk if it fixes your open issues.

Dickby:
I notice it very much and i would prefer to have that option, but it's your choice.
In that case the problem seems fixed, let's see if thejch finds something.

Dickby:
Wait there is still some double rendering while the special-ws is sliding in

vaxerski:
the thing that made it black for you is aedcade68dd0615fd919a7249633a554d0accd81

I can fix that later.

vaxerski:
> Wait there is still some double rendering while the special-ws is sliding in

on the special ws or regular ws? floating?

Dickby:
The special, it's not always the case. The active regular workspace has to be non_fullscreen. 
Newly created windows on the special-ws will be rendered at least twice while the special workspace is sliding in.
if the regular-ws is then fullscreened and the special workspace is activated again, the problem is gone even if the regular-ws window is unfullscreened again.
If you create a new window on that special however, that window will have that problem while siliding in, while the old window will be fine during the sliding animations.

thejch:
Yeah there's still occasional double rendering when the window slides in, especially when you have a maximized window open in the regular workspace. But it doesn't always happen. I'll try and find a consistent way to reproduce that.
Other than that it seems fixed.

thejch:
So it happens when I open special, create a window, close special, create a window in regular, fullscreen it, and then toggle special. Now the special window gets double rendered during the animation. Not sure if it's the same thing described by dickby, because in their case it says
> the active regular workspace has to be non_fullscreen.

which is the opposite of mine

Eclextic:
> So it happens when I open special, create a window, close special, create a window in regular, fullscreen it, and then toggle special. Now the special window gets double rendered during the animation. Not sure if it's the same thing described by dickby, because in their case it says
> 
> > the active regular workspace has to be non_fullscreen.
> 
> which is the opposite of mine

Probably is then exactly the same (Some kind of bool that should be reset/toggled, isn't getting toggled )

vaxerski:
fixed @thejch any other issues?

thejch:
I don't notice any on my end.

vaxerski:
@Dickby is your issue fixed?

vaxerski:
been running this myself for a while, seems to be working to me.

Issue #3029, opened 2023-08-19
X11 app not start in center


gfhdhytghd:
### Hyprland Version

28

### Bug or Regression?

Bug

### Description

As you could see, the x11 app which should be right in the center isn’t that center, moved a few pixels
![IMG_0332](https://github.com/hyprwm/Hyprland/assets/102898587/819d0dc1-268b-4f8d-96a0-62dd4d247665)

### How to reproduce

Build and install Hyprland from the main and install linuxqq or wechat-uos from AUR and start it

imxyy1soope1:
just try this:
```shell
#!/bin/bash

handle() {
	event="${line%%>>*}"
    edata="${line#"$event">>}"
	array=(${edata//,/ })
    if [ ${event} = "openwindow" -a ${array[2]} = "QQ" -a ${array[3]} = "QQ" ]; then
        hyprctl dispatch centerwindow
    fi
}

socat -U - UNIX-CONNECT:/tmp/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket2.sock | while read -r line; do handle "$line"; done
```

gfhdhytghd:
of course a windowsrull can solve it,  but this problem is obviously a bug and should not be solved by the user.

memchr:
What kind of error? I don't understand, are you launching this application floating?

Do you mind publishing your configurations?

gfhdhytghd:
there is no windowrule about QQ WeChat or such thing, this is a problem of the new hidpi xwayland solution, those x11 windows which should start in the center start right in the center if I use hidpi-patch instead of “force zero scaling”

gfhdhytghd:
# Setup monitors
# See https://wiki.hyprland.org/Configuring/Monitors/
#monitor=,preferred,auto,auto

# Dual monitor example on G15 Strix
# Both monitors here are at 1440 and 165Hz
# DP-1 is on the left and  eDP-1 is on the right
#monitor=eDP-1, 3120x2080@90,0x0, 2
monitor=eDP-1, 3120x2080@90, 550x2128, 2, bitdepth,10 
#monitor=eDP-1, 3120x2080@90, 550x2128, 2
#monitor=eDP-1, 1920x1080@90, 550x2128, 1, bitdepth,10 
monitor=DP-4, 1920x1080@60, 870x350, 1.078, transform,1
monitor=DP-2, 1920x1080@60, 430x1126, 1.078,
[...]

PR #3942, opened 2023-11-25
Fix float windows

h0tc0d3:
**Changelog:**
 - Add vector2 parameter general:float_min_crop for support minimal cropping size
    for windows that cannot transition to float. If 0 < value < 1.0 then use part of windows size, else pixel size. Example: `0.1 0.2`, `400 300`
 - Add parameter general:float_center_screen to control spawn new float windows.
   1 - New float windows spawn in center screen.
   0 - Сonsider the trimming at the top and bottom to calculate the center of the screen.
   For example, consider the size of the waybar.
 - Fix float windows
 - Fix togglefloat. Window did not return to the center of the screen, 
   for example, if there are two windows on the desktop, then the float window is 
   positioned in the center of half the screen.
 - Fix windows that start float and then go into a normal window, aka
   program launchers. Example: Blizzard battle.net client.
   The bug render full-screen programs partially behind the screen.
   Now the xwayland games launch normally.
 - Code cleanup and refactoring.
 - Replaced vec() with goalv() in logic other than animation,
   it seemed to me more correct because goalv() is used in many places.
   Not sure if this is correct.

Related bug reports: https://github.com/hyprwm/Hyprland/issues/3029

Tested programs: Blender, Jetbrains IDEs, Blizzard battle.net client, qt float program.

No

Yes.

Fix this:
screenshot_1700794200
screenshot_1700799073

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -94,6 +94,8 @@ void CConfigManager::setDefaultVars() {
     configValues["general:resize_on_border"].intValue        = 0;
     configValues["general:extend_border_grab_area"].intValue = 15;
     configValues["general:hover_icon_on_border"].intValue    = 1;
+    configValues["general:float_min_crop"].vecValue          = Vector2D(0.25, 0.25);
+    configValues["general:float_center_screen"].intValue     = 0;
     configValues["general:layout"].strValue                  = "dwindle";
     configValues["general:allow_tearing"].intValue           = 0;
 
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -455,9 +455,10 @@ void IHyprLayout::changeWindowFloatingMode(CWindow* pWindow) {
 
         // if the window is pseudo, update its size
         if (!pWindow->m_bDraggingTiled)
-            pWindow->m_vPseudoSize = pWindow->m_vRealSize.goalv();
+            pWindow->m_vPseudoSize = PSAVEDSIZE;
 
-        pWindow->m_vLastFloatingSize = PSAVEDSIZE;
+        pWindow->m_vLastFloatingPosition = PSAVEDPOS;
+        pWindow->m_vLastFloatingSize     = PSAVEDSIZE;
 
         // move to narnia because we don't wanna find our own node. onWindowCreatedTiling should apply the coords back.
         pWindow->m_vPosition = Vector2D(-999999, -999999);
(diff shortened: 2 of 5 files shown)


h0tc0d3:
@vaxerski I solved the problem with xwayland windows. I replaced vec() with goalv() in logic other than animation, it seemed to me more correct because goalv() is used in many places. Not sure if this is correct. There is also a bug that the xwayland window changes size and does not center the window, but as soon as you move the mouse, the window moves to the center and my code centers the window. I have not yet figured out where the window resize event occurs, this bug is not critical, but it will need to be fixed in the future.

vaxerski:
> I have not yet figured out where the window resize event occurs

configureX11 and unmanagedSetGeometry

h0tc0d3:
@vaxerski done. The problem turned out to be more widespread, i.e. there were several different bugs and the solution took longer. There is only one float window left, which is not positioned in the center, and the reason seems to be specific: `pWindow->m_uSurface.xwayland->window_type`. In all the programs I tested, now all float windows are displayed correctly.

src/managers/XWaylandManager.cpp
```diff
@@ -319,27 +317,15 @@ Vector2D CHyprXWaylandManager::getMaxSizeForWindow(CWindow* pWindow) {
...
-
-        double     distance =
-            vecToRectDistanceSquared(coord, {m->vecXWaylandPosition.x, m->vecXWaylandPosition.y}, {m->vecXWaylandPosition.x + SIZ.x - 1, m->vecXWaylandPosition.y + SIZ.y - 1});
-
-        if (distance < bestDistance) {
-            bestDistance = distance;
-            pMonitor     = m.get();
-        }
-    }
+    CMonitor* pMonitor = g_pCompositor->getMonitorFromVector(coord);
```
vaxerski, line 320: you can't do that

src/layout/IHyprLayout.cpp
```diff
@@ -215,16 +212,16 @@ void IHyprLayout::onBeginDragWindow() {
     m_vLastDragXY          = m_vBeginDragXY;
 
     // get the grab corner
-    if (m_vBeginDragXY.x < m_vBeginDragPositionXY.x + m_vBeginDragSizeXY.x / 2.0) {
-        if (m_vBeginDragXY.y < m_vBeginDragPositionXY.y + m_vBeginDragSizeXY.y / 2.0) {
+    if (m_vBeginDragXY.x < m_vBeginDragPositionXY.x + m_vBeginDragSizeXY.x / 2.f) {
```
vaxerski, line 215: what is with the obsession to change .0 to .f?

src/layout/IHyprLayout.cpp
```diff
@@ -477,18 +475,28 @@ void IHyprLayout::changeWindowFloatingMode(CWindow* pWindow) {
 
         g_pCompositor->changeWindowZOrder(pWindow, true);
 
-        if (DELTALESSTHAN(pWindow->m_vRealSize.vec().x, pWindow->m_vLastFloatingSize.x, 10) && DELTALESSTHAN(pWindow->m_vRealSize.vec().y, pWindow->m_vLastFloatingSize.y, 10)) {
-            pWindow->m_vRealPosition = pWindow->m_vRealPosition.goalv() + (pWindow->m_vRealSize.goalv() - pWindow->m_vLastFloatingSize) / 2.f + Vector2D{10, 10};
-            pWindow->m_vRealSize     = pWindow->m_vLastFloatingSize - Vector2D{20, 20};
+        const auto PSAVEDSIZE = pWindow->m_vRealSize.goalv();
+
+        Vector2D   sizeDelta     = {0, 0};
```
vaxerski, line 480: uhh... why?

h0tc0d3: [vaxerski](https://github.com/vaxerski) We try to use lastfloatposition whenever possible if it has been saved, and change it only if position.x == 0 or position.y == 0, so we add a delta size and position for windows that themselves do not know how to switch to float(allways maximazed ), so we crop the window correctly, I increased the window crop size. If we do not use lastfloatposition, then there are situations when the window after toggle float loses its coordinates and begins to be drawn in a different place.

src/managers/XWaylandManager.cpp
h0tc0d3: 2.0 - double, 2.f - float. The compiler will complain in strict mode. But the reason is different, when I wrote the code I used 2.0 in many places, then I corrected everything together.

src/layout/IHyprLayout.cpp
h0tc0d3: 2.0 - double, 2.f - float. The compiler will complain in strict mode. But the reason is different, when I wrote the code I used 2.0 in many places, then I corrected everything together.

src/managers/XWaylandManager.cpp
h0tc0d3: @vaxerski I looked at the code and it's identical.

src/layout/IHyprLayout.cpp
h0tc0d3: @vaxerski Also in some checks for == 0, I replaced it with <= 0 because In some places the coordinates and dimensions may be negative.

src/managers/XWaylandManager.cpp
vaxerski: well then you looked wrong.

src/layout/IHyprLayout.cpp
vaxerski: mhm, 50 too much though, maybe 20?

h0tc0d3: @vaxerski So now there are no small screens like 480x320, and smartphones also have larger screens. On a 480x320 screen, the window after cropping will have a size of 380x220, which visually makes the window 1/3 smaller. 100 seemed optimal to me to see that the window became float.

vaxerski: fair enough

h0tc0d3:
At the moment I decided to rework the float code and add more flexible cropping tile to float window.

vaxerski:
what value

h0tc0d3:
@vaxerski 

```C
IHyprLayout::changeWindowFloatingMode(CWindow* pWindow)
static auto* const FLOATCENTERSCREEN = &g_pConfigManager->getConfigValuePtr("general:float_center_screen")->intValue;
static auto* const FLOATMINCROP = &g_pConfigManager->getConfigValuePtr("general:float_min_crop")->vecValue;
```

vaxerski:
fwiw both options will pass review, extremely bad ideas

h0tc0d3:
@vaxerski This is the last version of the changes, there will be no more changes.

src/config/ConfigManager.cpp
```diff
@@ -94,6 +94,8 @@ void CConfigManager::setDefaultVars() {
     configValues["general:resize_on_border"].intValue        = 0;
     configValues["general:extend_border_grab_area"].intValue = 15;
     configValues["general:hover_icon_on_border"].intValue    = 1;
+    configValues["general:float_min_crop"].vecValue          = Vector2D(0.25, 0.25);
+    configValues["general:float_center_screen"].intValue     = 0;
```
vaxerski, line 98: these options I will not agree to, they are extremely bad ideas.

vaxerski (commented):
also no need to ping me all the time, I get notifications regardless

Issue #3174, opened 2023-09-04
Set workspace/monitor default master layout orientation


robertoash:
### Description

I may have missed something but I have scoured the wiki for a solution to this and have been unable to find one. 

I use 2 monitors, one horizontal and one vertical. I would like to set the default orientation of the master layout to "top" on the vertical monitor and to "left" on the horizontal one.

I have added keybinds as a workaround. The issue is that I open several apps on startup and they all open with the wrong layout on my vertical monitor. I have to manually go into every workspace on that monitor and change the layout with the keybinds.

These are the keybinds I've set up:
```
bind = SUPER CTRL SHIFT, up, layoutmsg, orientationleft
bind = SUPER CTRL SHIFT, right, layoutmsg, orientationleft
bind = SUPER CTRL SHIFT, left, layoutmsg, orientationtop
bind = SUPER CTRL SHIFT, down, layoutmsg, orientationtop
```
These let me "toggle" between the two orientations. I have not found a way to use the layoutmsg dispatcher together with `exec-once`.

This is an example of two apps I open on startup:
```
exec-once = [workspace 1 silent] vscodium
exec-once = [workspace 1 silent] slack
```
I have not been able either to find a way to tell Hyprland to change the orientation to bottom before or after launching slack on workspace 1, for example.

Another solution (even nicer in my opinion) would be to be able to set a config so that all workspaces that spawn on a specific monitor are set to default to a specific layout. I have bound workspaces to the vertical monitor so having this config for the monitor would achieve the desired results.

**TLDR:** layoutmsg can be used to change the orientation of the master layout but the wiki only mentions how it can be used together with keybinds and does not mention how or if it can be used together with `exec-once`, a workspace rule or a monitor config.

Any ideas? If not, then maybe it's a good feature to request.

vaxerski:
script?

listen to the activeworkspace event and issue a layoutmsg based on that

robertoash:
It would be nice if the functionality was built in in the same way that you can easily do a transform on a monitor to make it vertical. 

In any case, a script could be a workaround so, after a lot of trial and error, I came up with this:

```
#! /bin/bash
# This script is used to listen for newly created workspaces using hyprctl
# and changing the layout to orientationtop if the workspace is on a vertical monitor

#!/bin/bash

# Variable to store workspace information from the previous run
previous_run_workspaces=""

# Delay to give time for the workspaces to be created on boot
sleep 5

####################################################
############## Get a list of vertical monitors ################
[...]

PR #3964, opened 2023-11-26
Add workspace rule for master layout orientation

thejch:
Addresses #3174 
No monitor rules but we can at least have a workspace rule

Not sure if this is ideal because this is only for master but it's in the general workspace rules.

Ready.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -79,7 +79,7 @@ void CConfigManager::setDefaultVars() {
     configValues["general:no_border_on_floating"].intValue = 0;
     configValues["general:gaps_in"].intValue               = 5;
     configValues["general:gaps_out"].intValue              = 20;
-    configValues["general:gaps_workspaces"].intValue        = 0;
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
I'd do a similiar approach to dispatchers. `layoutopt` or something and then `orientation top` etc

thejch:
> I'd do a similiar approach to dispatchers. `layoutopt` or something and then `orientation top` etc

Changed it to `layoutopt:orientation:top`

vaxerski:
for future-proofing the code it'd be better if it was a map str -> str, we might add more layoutopts later.

vaxerski (dismissed):
lgtm, wiki mr needed though

thejch:
Actually, str -> str is not the best idea since you can't store other types (bool, int, whatever may come in the future). I want to use `std::any` for this, or even a struct, but I think `std::any` sounds easier to deal with atm. Thoughts?

thejch:
Made some changes. Wiki MR added as well

PR #3983, opened 2023-11-27
Add Meson optimizations

h0tc0d3:

Remove:
```
'optimization=3',
'buildtype=release',
'debug=false'
```
The usage is redundant. The -O3 compiler flag is not safe, in synthetic tests in certain sections of the code we can get acceleration, but -O3 greatly increases the amount of machine code, which creates more frequent misses in the processor cache and breaks branch prediction. `buildtype` already has optimization and debug flags.

Instead, I suggest adding the -ffast-math flag for better optimization. In computer graphics, using -ffast-math is safe, and we can sacrifice accuracy and some checks in mathematical calculations for the sake of performance. In computer graphics, everyone used less accurate but faster calculations, such as sqrt calculation. Added a flag to align function addresses on a 32-byte boundary, this is safe and gives better use of the processor cache. For example gamescope uses -ffast-math.

Remove the unnecessary X11 backend.

https://kristerw.github.io/2021/10/19/fast-math/
https://en.wikipedia.org/wiki/Fast_inverse_square_root

No

Yes

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,10 +2,7 @@ project('Hyprland', 'cpp', 'c',
   version : run_command('jq', '-r', '.version', join_paths(meson.source_root(), 'props.json'), check: true).stdout().strip(),
   default_options : [
     'warning_level=2',
-    'default_library=static',
-    'optimization=3',
-    'buildtype=release',
-    'debug=false'
+    'default_library=static'
     # 'cpp_std=c++23' # not yet supported by meson, as of version 0.63.0
     ])
 
@@ -18,6 +15,13 @@ elif cpp_compiler.has_argument('-std=c++2b')
   add_global_arguments('-std=c++2b', language: 'cpp')
 else
   error('Could not configure current C++ compiler (' + cpp_compiler.get_id() + ' ' + cpp_compiler.version() + ') with required C++ standard (C++23)')
+endif
+
+if get_option('fast_math').enabled()
+  add_global_arguments(
+  cpp_compiler.get_supported_arguments(
+    ['-ffast-math', '-falign-functions=32']
+  ), language: 'cpp')
 endif
 
 add_project_arguments(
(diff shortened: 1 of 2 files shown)


q234rty:
> Remove the unnecessary X11 backend.

The X11 backend of wlroots allows hyprland to run nested in a X11 environment as a client, so I won't simply call it unnecessary...

h0tc0d3:
> > Remove the unnecessary X11 backend.
> 
> The X11 backend of wlroots allows hyprland to run nested in a X11 environment as a client, so I won't simply call it unnecessary...
This option increases build time and binary size. To be honest, I have never seen anyone use an X11 backend for DE made for wayland.

vaxerski:
what about the wayland backend?

h0tc0d3:
> what about the wayland backend?

I don't see this in wlroots.

```
wlroots 0.18.0-dev
    drm-backend     : YES
    x11-backend     : NO
    libinput-backend: YES
    xwayland        : YES
    gles2-renderer  : YES
    vulkan-renderer : NO
    gbm-allocator   : YES
    session         : YES
    xcb-errors      : NO
    egl             : YES
    libliftoff      : YES
```

vaxerski:
hm maybe it's baked in

vaxerski:
what about -O2?

h0tc0d3:
> what about -O2?

Already in `--buildtype release` and debug flags in `--buildtype debug`
```
buildtype {plain, debug, debugoptimized, release, minsize, custom}
```

vaxerski:
can we still explicitly set it, please?

h0tc0d3:
> can we still explicitly set it, please?

What's the point of this? I've seen a lot of projects on meson and usually no one does this.
It turned out that in some version of meson for `--buildtype release` the author changed `optimization=2` to `optimization=3`. 
https://mesonbuild.com/Builtin-options.html

```
buildtype	debug	optimization
------------------------------------
plain	        false	plain
debug	        true	0
debugoptimized	true	2
release	        false	3
minsize	        true	s
```
On the one hand, compilers are now more advanced, but I think this is bad. 

https://github.com/ValveSoftware/gamescope/blob/master/meson.build

PR #4010, opened 2023-11-30
Fix overzealous mouse capture by resize_on_border for XWayland applications

### LINKED ISSUE #2456: resize_on_border makes dropdown menu unusable in xwayland
When `resize_on_border` is enabled, clicking/dragging a drop-down menu in an xwayland application that is floating outside the window border will cause it's parent window to resize instead of triggering the action.

## Steps to reproduce

Open a program that uses xwayland in a position that the drop-down menu is in.

Try clicking or dragging the menu item.

The parent window is resized.

https://github.com/hyprwm/Hyprland/assets/118117622/c517a813-ab90-4611-8b06-b5b46a2a1f29

## Sample program

```python
#!/usr/bin/python

import gi
import sys

gi.require_version("Gtk", "3.0")
from gi.repository import Gtk, Gio, Gdk

resizable = sys.argv[-1] != "1"


class Window(Gtk.ApplicationWindow):
    def __init__(self, app):
        super(Window, self).__init__(application=app)
        print(Gdk.Display.get_default())

        menu = Gio.Menu.new()
        menu_item = Gio.MenuItem.new("Menu")

        submenu = Gio.Menu.new()
        for i in range(15):
            submenu.append_item(Gio.MenuItem.new(f"Item {i}"))

        menu_item.set_submenu(submenu)
        menu.append_item(menu_item)

        app.set_menubar(menu)

        self.set_show_menubar(True)
        self.set_default_size(200, 200)
        self.set_resizable(not resizable)

    def on_close(self, *_):
        self.close()


def on_activate(app):
    win = Window(app)
    win.present()


app = Gtk.Application()
app.connect("activate", on_activate)
app.run(None)
```

launch with

```
env GDK_BACKEND=x11 python sample.py
```

wrenby:

Fixes #2456

If an XWayland application has a menu that extends outside the range of the originating window, it's currently impossible to interact with that menu while the `resize_on_border` config option is enabled. On native Wayland apps, this case seems to be safeguarded against by `CWindow::hasPopupAt`, but that function is unimplemented for XWayland, presumably because X11 wasn't designed with that kind of query in mind. This fix only allows the mouse to be captured when the cursor is explicitly inside the configured resize border, rather than incorrectly assuming that a border-resize is the only reason mouse input might fall to a window while the cursor is outside the client area.

Video of fix, using the example program provided in the original issue:

https://github.com/hyprwm/Hyprland/assets/22502477/e120ae9c-cb68-441c-ace2-60d51d7a0446

- Also fixes a similar bug for drag and drop between XWayland windows.
- Remaining bug: If the menu item you want to select is directly on top of the resize border, the resize will "punch through" the menu and capture the mouse. Shouldn't be a problem except for large values of `extend_border_grab_area` or tiny menu items, but a more proper fix would implement `hasPopupAt` for X11 clients.

Should be ready to merge provided the tiny bit of code duplication between `CInputManager::processMouseDownNormal` and `CCompositor::vectorToWindowIdeal` is acceptable and the code fits the style. Again it's not a perfect fix, but it's a strict improvement over current behavior.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -597,6 +597,9 @@ void CInputManager::processMouseDownNormal(wlr_pointer_button_event* e) {
     const auto         PASS            = g_pKeybindManager->onMouseEvent(e);
     static auto* const PFOLLOWMOUSE    = &g_pConfigManager->getConfigValuePtr("input:follow_mouse")->intValue;
     static auto* const PRESIZEONBORDER = &g_pConfigManager->getConfigValuePtr("general:resize_on_border")->intValue;
+    static auto* const PBORDERSIZE       = &g_pConfigManager->getConfigValuePtr("general:border_size")->intValue;
+    static auto* const PBORDERGRABEXTEND = &g_pConfigManager->getConfigValuePtr("general:extend_border_grab_area")->intValue;
+    const auto         BORDER_GRAB_AREA  = *PRESIZEONBORDER ? *PBORDERSIZE + *PBORDERGRABEXTEND : 0;
 
     if (!PASS && !*PPASSMOUSE)
         return;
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
lgtm

PR #4021, opened 2023-12-01
Handle fullscreen windows in vectorToWindowIdeal

### LINKED ISSUE #2847: Can't click menu bar of flatpak apps in full-screen mode due to gaps
### Hyprland Version

Hyprland, built from branch main at commit bf0d8ab4a3c44728cd1330ee32a8143ad1f5dcdd

### Bug or Regression?

Bug

### Description

Consider `inkscape` installed via flatpak. I have the following settings:
```
general {
    gaps_in = 6
    gaps_out = 18
    border_size = 3
}
```
When in full-screen mode (`fullscreen` mode 1 - bar still shown, and `no_gaps_when_only = 1`), clicking on menu bar does nothing, likely because of the `gaps_out`. I moved the mouse down a bit and the menu is clickable there (see the picture). Also, if I set `gaps_out = 0`, the menu is clickable normally.

Apps installed via pacman works normally, looks like only flatpak apps have this issue.

image

### How to reproduce

- Set `gaps_out` high enough, for example `gaps_out = 18`.
- Open in full-screen mode an app installed via flatpak with menu bar (e.g. inkscape, libreoffice, etc.).
- Try clicking on the menu bar at the top region, won't work.

_No response_

dranull:
Clicking in a fullscreen window over the decoration area (gaps, border, groupbar) of other windows will pass the mousedown event to the other window in some cases.

Example: With two (or more) tiled windows on a workspace, activating fullscreen for one of them, then clicking over the area where another window had a groupbar (or when using `no_gaps_when_only = yes` with `resize_on_border = yes`) wasn't possible (it was interactive with the groupbar).
Most notable annoyances were the inability to click on browser tabs, or to simply drag the scrollbars.

Problem: m_vWindows is sorted from bottom to top. Fullscreening a window will put it at the end of that list, so it will be always the last to be checked. However, other tiled windows can be under the cursor at the same time (beneath the fullscreen window), and if one matches, `vectorToWindowIdeal` will return it first.

Solution: check in reverse order, so the top window will be returned first.

I've left some other minor changes in the code, while testing, if you don't mind.
Also, most probably fixes #2847.

I guess.


vaxerski (dismissed):
yep, lgtm.

Tiled windows should never be focused if there is a fullscreen one :)

dranull:
Sure, but I was thinking about an edge case, like this:
HDMI-A-1_2023-12-02_16-07-12
But it turns out, even fake fullscreen will make the popup completely disappear. I should have tried...

vaxerski (approved):

PR #4055, opened 2023-12-05
Add meson option to enable plugins support

h0tc0d3:
The system of plugins and hooks noticeably slows down Hyprland.
Add a meson option to enable plugin support.
Thus, plugin support can only be enabled by those who need it.

No

Yes

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -380,7 +380,11 @@ void CConfigManager::init() {
 
 void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::string& VALUE) {
     if (!configValues.contains(COMMAND)) {
+#ifdef WITH_PLUGINS
         if (!COMMAND.starts_with("device:") /* devices parsed later */ && !COMMAND.starts_with("plugin:") /* plugins parsed later */) {
+#else
+        if (!COMMAND.starts_with("device:") /* devices parsed later */) {
+#endif
             if (COMMAND[0] == '$') {
                 // register a dynamic var
                 Debug::log(LOG, "Registered dynamic var \"{}\" -> {}", COMMAND, VALUE);
@@ -417,6 +421,7 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
         }
 
         CONFIGENTRY = &it->second.at(CONFIGVAR);
+#ifdef WITH_PLUGINS
     } else if (COMMAND.starts_with("plugin:")) {
         for (auto& [handle, pMap] : pluginConfigs) {
             auto it = std::find_if(pMap->begin(), pMap->end(), [&](const auto& other) { return other.first == COMMAND; });
@@ -431,6 +436,7 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
             m_vFailedPluginConfigValues.emplace_back(std::make_pair<>(COMMAND, VALUE));
             return; // silent ignore
         }
+#endif
     } else {
         CONFIGENTRY = &configValues.at(COMMAND);
     }
@@ -1347,6 +1353,7 @@ void CConfigManager::handleEnv(const std::string& command, const std::string& va
     }
 }
 
+#ifdef WITH_PLUGINS
 void CConfigManager::handlePlugin(const std::string& command, const std::string& path) {
     if (std::find(m_vDeclaredPlugins.begin(), m_vDeclaredPlugins.end(), path) != m_vDeclaredPlugins.end()) {
         parseError = "plugin '" + path + "' declared twice";
@@ -1355,6 +1362,7 @@ void CConfigManager::handlePlugin(const std::string& command, const std::string&
 
     m_vDeclaredPlugins.push_back(path);
 }
+#endif
 
 std::string CConfigManager::parseKeyword(const std::string& COMMAND, const std::string& VALUE, bool dynamic) {
     if (dynamic) {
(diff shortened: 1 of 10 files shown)


src/config/ConfigManager.cpp
```diff
@@ -1565,9 +1578,13 @@ void CConfigManager::loadConfigLoadVars() {
     m_dBlurLSNamespaces.clear();
     m_dWorkspaceRules.clear();
     setDefaultAnimationVars(); // reset anims
+#ifdef WITH_PLUGINS
     m_vDeclaredPlugins.clear();
+#endif
     m_dLayerRules.clear();
+#ifdef WITH_PLUGINS
     m_vFailedPluginConfigValues.clear();
```
fufexan, line 1586: @vaxerski can't this be moved below `m_vDeclaredPlugins.clear()`?

fufexan (commented):
The meson side looks good. I'll change the Nix code accordingly.

fufexan:
Looks like there are a still few places where plugins are referenced.

zakk4223:
Does this mean hyprland is moving to a release config that doesn't enable plugins by default?

fufexan:
No. Plugin support will be on by default, but it will be possible to disable it if the user so desires.

Having it disabled by default in this PR makes it easy to test in CI.

vaxerski:
wait a sec, I instinctively tagged fuf because meson, wtf is this?

vaxerski:
> The system of plugins and hooks noticeably slows down Hyprland.

source? how the fuck?

Eclextic:
Well every piece of extra code has some performance penalty...
Question is how much are we tanking?

If it's like 0.001s then this PR is practically useless, but I would be for it, if we are getting an improvement of 1/60th of a frame...

vaxerski:
> Well every piece of extra code has some performance penalty...

No. If it's not ran, it doesn't. Stop trying to convince yourself 3µs of launch time that those 2000 LOC add matter to you.

h0tc0d3:
@vaxerski minus call hooks, minus one library when linking = the compiler can optimize the code better, and the binary makes better use of the processor cache. The difference is noticeable in games, and the animations are smoother. The frame time has become more constant, and this is noticeable on a 240Hz monitor.

vaxerski:
call hooks aren't used in hyprland. Stop acting like code that is _never ran_ is impacting frame times...

Unless you provide credible benchmarks, I am saying this is placebo.

h0tc0d3:
> call hooks aren't used in hyprland.

Yep.

```
#define EMIT_HOOK_EVENT(name, param)                                                                                                                                               \
    {                                                                                                                                                                              \
        static auto* const PEVENTVEC = g_pHookSystem->getVecForEvent(name);                                                                                                        \
        SCallbackInfo      info;                                                                                                                                                   \
        g_pHookSystem->emit(PEVENTVEC, info, param);                                                                                                                               \
    }
#define EMIT_HOOK_EVENT_CANCELLABLE(name, param)                                                                                                                                   \
    {                                                                                                                                                                              \
        static auto* const PEVENTVEC = g_pHookSystem->getVecForEvent(name);                                                                                                        \
        SCallbackInfo      info;                                                                                                                                                   \
        g_pHookSystem->emit(PEVENTVEC, info, param);                                                                                                                               \
        if (info.cancelled)                                                                                                                                                        \
            return;                                                                                                                                                                \
    }
```

vaxerski:
those arent call hooks but just events. And those you can't even rid of because they are used internally. Please read the codebase first.

I know they are named sililiarly but function hooks != hook events

h0tc0d3:
@vaxerski I redefined them to empty context. This way they are disabled at compile time. At the moment, I have not noticed a single problem in the work.

```
#ifdef WITH_PLUGINS
#define EMIT_HOOK_EVENT(name, param)                                                                                                                                               \
    {                                                                                                                                                                              \
        static auto* const PEVENTVEC = g_pHookSystem->getVecForEvent(name);                                                                                                        \
        SCallbackInfo      info;                                                                                                                                                   \
        g_pHookSystem->emit(PEVENTVEC, info, param);                                                                                                                               \
    }
#define EMIT_HOOK_EVENT_CANCELLABLE(name, param)                                                                                                                                   \
    {                                                                                                                                                                              \
        static auto* const PEVENTVEC = g_pHookSystem->getVecForEvent(name);                                                                                                        \
        SCallbackInfo      info;                                                                                                                                                   \
        g_pHookSystem->emit(PEVENTVEC, info, param);                                                                                                                               \
        if (info.cancelled)                                                                                                                                                        \
            return;                                                                                                                                                                \
    }
#else
#define EMIT_HOOK_EVENT(name, param) {}
#define EMIT_HOOK_EVENT_CANCELLABLE(name, param) {}
#endif
```

zakk4223:
You might want to search for 'hookDynamic' in the code base.  It is actually used and it is actually important

vaxerski:
I will close this as it's going nowhere. You provided zero benchmarks, zero actual tests to support your magical "performance benefit" and are breaking completely/parts of:
- build
- notificationOverlay
- hyprerror
- keybindmanager
- inputmethodrelay
- screencopy
- xdgoutput
- renderer
- decorationpositioner

h0tc0d3:
@vaxerski Why does this list work for me?

h0tc0d3:
@vaxerski Oddly enough, everything works for me except NotificationOverlay, and does not show errors in the config. I specifically connected a second monitor and checked. Blur, shadows, rounded corners all work. Maybe blur_new_optimizations not work. keybindmanager works, screencopy works in my screen lock app. Probably need to do some research. As was written earlier, I didn’t touch `hookStatic` and `hookDynamic`.

vaxerski:
hookStatic and hookDynamic depend on EMIT_HOOK_EVENT to fire...

PR #4059, opened 2023-12-05
Stop propagating axis events after valid binds

dranull:
Currently using a mouse wheel shortcut, like SUPER + mouse_up/down for zooming/changing volume, will also activate group switching if the cursor is over a groupbar.

This change will make sure that only my binds are activated.

No.

Yes.


vaxerski (approved):
lgtm

PR #4102, opened 2023-12-09
screen_shader config fixes

dranull:
- Workaround: Setting `--batch "keyword decoration:screen_shader ~/test.frag"` wasn't possible because the slash is used for flags. The commit is more like a patch than a real solution, as it makes an assumption; if there is a space after the first word than it's probably a keyword. But now it's possible to use flags and slashes at the same time.
- Fix: `keyword decoration:screen_shader [[EMPTY]]` previously created a tick for a non-existent file.

No.

Yes.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -550,7 +550,7 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
         }
     }
 
-    if (COMMAND == "decoration:screen_shader") {
+    if (COMMAND == "decoration:screen_shader" && VALUE != STRVAL_EMPTY) {
         const auto PATH = absolutePath(VALUE, configCurrentPath);
 
         configPaths.push_back(PATH);
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1402,10 +1402,17 @@ std::string getReply(std::string request) {
     auto format = HyprCtl::FORMAT_NORMAL;
 
     // process flags for non-batch requests
-    if (!request.contains("[[BATCH]]") && request.contains("/")) {
+    if (!request.starts_with("[[BATCH]]") && request.contains("/")) {
         long unsigned int sepIndex = 0;
         for (const auto& c : request) {
             if (c == '/') { // stop at separator
+                break;
+            }
+
+            // after whitespace assume the first word as a keyword,
+            // so its value can have slashes (e.g., a path)
+            if (c == ' ') {
+                sepIndex = request.size();
                 break;
             }
 


src/debug/HyprCtl.cpp
```diff
@@ -1347,13 +1347,20 @@ std::string getReply(std::string request) {
     auto format = HyprCtl::FORMAT_NORMAL;
 
     // process flags for non-batch requests
-    if (!request.contains("[[BATCH]]") && request.contains("/")) {
-        long unsigned int sepIndex = 0;
+    if (!request.starts_with("[[BATCH]]") && request.contains("/")) {
+        uint8_t sepIndex = 0;
```
vaxerski, line 1351: why?

dranull: Unfortunately my week old cpp experience cannot explain (yet) why not. It seemed reasonable, since there won't be that many flags.

h0tc0d3: @dranull `long unsigned int` == `size_t` or `uint64_t`, better use `size_t`,
`uint8_t` has max value = 255 and we are limited to 255 chars string. Based on the logic of the code, you are getting an integer overflow. `sepIndex >= UINT8_MAX`. You have broken the logic of the code. The code parses the string and looks for '/', and removes everything before /. This way you have limited string to 255 characters.

dranull: @h0tc0d3 I see, though it doesn't overflow (because of the check), but indeed it will cut the string to 255, and `if (sepIndex < request.size())` will be true then. Every request I tested was way under 255, and I was focusing on the length of the flags to limit them further, so I didn't notice. My bad, sorry.

h0tc0d3: @dranull Don't create empty commits when you create a commit and then delete it.

Use interactive rebase instead. `git rebase -i COMMIT_HASH^`
Replace `pick` with `drop` or `edit`. 
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History 
After changes: `git push --force`

dranull: I know how that works, but I don't see the point in this case, it's a new commit not a deleted one. The commits are squashed on merge anyway, so you'll have only one commit in the repo, from which you can't tell what the individual commits did. But in the pull request at least you'll be able to see the review process. But if the boss wishes, I'll do that way.

src/config/ConfigManager.cpp
```diff
@@ -457,11 +457,38 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
             parseError = "Error setting value <" + VALUE + "> for field <" + COMMAND + ">.";
         }
     } else if (CONFIGENTRY->strValue != "") {
+        std::string newValue = VALUE.empty() ? STRVAL_EMPTY : VALUE;
```
vaxerski, line 460: this should not be done in here. It belongs better to `loadConfigLoadVars` (look at the end of the func)

dranull: Excuse my dumbness once again, but I don't see how. `loadConfigLoadVars` is not called at all when you dynamically update an option via `hyprctl keyword` (these issues exist only here, normal config reload is fine).

vaxerski: you can make it a func and call it at the end of loadConfigLoadVars and at the end if dynamically parsing the keyword of parseKeyword

dranull: I revisited this, and I would rather pass on these changes. I reverted them.

vaxerski (approved):
ok

PR #4104, opened 2023-12-09
Remake borders as window decorations

### LINKED ISSUE #4100: Support shadows being drawn within borders.
### Description

Shadows were changed from being drawn within borders to outside of them. This is understandably desirable as you generally don't want the shadow to interfere with the color you set borders to. The issue is that hyprland now doesn't support a 'clean' way to only have a border on the active window while also having shadows. Currently there are two options:

1. setting the alpha of `general:col.inactive_border` to 00. This is the solution that worked prior to the shadow update, but now produces a border sized lack of shadow around the window

20231209_15h39m50s_grim

2. Creating the windowrule `windowrulev2 = bordersize 0, focus: 0`, dynamically removing the border from inactive windows. This isn't viable as it causes windows to resize whenever focus changes which is generally quite jarring

https://github.com/hyprwm/Hyprland/assets/43202406/f76beba1-4855-4295-9584-18bd10718229

It would be great if there were an option to change whether or not shadows are drawn within borders our outside of them. Of course this would re-introduce the shadows interfering with  non-opaque border colors but for me that would be much more desirable than the current 2 options we have.

vaxerski:
Makes borders window decorations for parity and code uniformity

Fixes #4100 by adding an option to toggle the behavior, `general:border_part_of_window` `bool` default `1`


vaxerski:
no further comments arose, merging

PR #4111, opened 2023-12-10
Fix memory allocation for the background texture

Nmzik:
Enables the avoidance of memory allocation for the background texture (GL_TEXTURE), which is useful for users not utilizing default Hyprland wallpapers (e.g. swaybg). With the previous implementation, it allocated memory for the texture regardless of whether the user has opted out of it (`misc:disable_hyprland_logo`).


vaxerski (approved):
lgtm. Worth noting it will not free it if the opt is disabled at runtime

Nmzik:
@vaxerski Completely agree. I didn't want to complicate code, and I don't really think people will constantly change the `misc:disable_hyprland_logo` setting at runtime...

PR #4112, opened 2023-12-10
ci: use composite action to minimize code duplication

zephvr:
The idea behind this code refactoring is to factorize the image setup process used in CI. The goal is also to make it more readable and easily maintainable

GitHub composite action seemed like a good choice for this purpose, as it implement in between steps. I also choose local action compare to remote for better control.

I also changed the build directory for meson from `obj-x86_64-pc-linux-gnu` to `build`. I did not find any reason for the old name, while the new is explicitly given in the contributing guide.

I also updated the `actions/checkout` to v4

Is ready for merging

### DIFF
--- a/.github/actions/setup_base/action.yml
+++ b/.github/actions/setup_base/action.yml
@@ -1,0 +1,75 @@
+name: "Setup base"
+
+inputs:
+  INSTALL_XORG_PKGS:
+    description: 'Install xorg dependencies'
+    required: false
+    default: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Get required pacman pkgs
+      shell: bash
+      run: |
+        sed -i 's/SigLevel    = Required DatabaseOptional/SigLevel    = Optional TrustAll/' /etc/pacman.conf
+        pacman --noconfirm --noprogressbar -Syyu
+        pacman --noconfirm --noprogressbar -Sy \
+          base-devel \
+          cairo \
+          clang \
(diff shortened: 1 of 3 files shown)


.github/actions/setup_base/action.yml
```diff
@@ -0,0 +1,87 @@
...
+          meson \
+          ninja \
+          pango \
+          pixman \
+          pkgconf \
+          scdoc \
+          seatd \
+          systemd \
+          tomlplusplus \
+          vulkan-headers \
```
vaxerski, line 48: if we're refactoring, vulkan deps can be dropped

.github/actions/setup_base/action.yml
```diff
@@ -0,0 +1,87 @@
...
+          xorg-xkbcomp \
+          xorg-xwayland
+
+    - name: Set up user
+      shell: bash
+      run: |
+        useradd -m githubuser
+        echo -e "root ALL=(ALL:ALL) ALL\ngithubuser ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers
+
+    - name: Install libdisplay-info from the AUR
```
vaxerski, line 73: same as above, it's now in official pkgs

zephvr:
I just did what you requested. Should be good now

vaxerski (approved):
lgtm, I dont speak CI well but it should work

vaxerski:
thanks!

PR #4113, opened 2023-12-10
layout: fix interactions with fakefullscreen

zephvr:
Fix some edge case when using fakefullscreen, maximize and fullscreen at the same time 

This PR fixes two cases I encountered:
- In a maximized and fakefullscreen window, it was not possible to exit the fakefullscreen state
- In a fakefullscreen window, toggling fullscreen twice, the fakefullscreen would no longer be rendered when exiting fullscreen

Both can be easily tested with firefox, as the title bar goes away in fullscreen

Ready to merge

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -998,3 +998,8 @@ int CWindow::getRealBorderSize() {
 bool CWindow::canBeTorn() {
     return (m_sAdditionalConfigData.forceTearing.toUnderlying() || m_bTearingHint) && g_pHyprRenderer->m_bTearingEnvSatisfied;
 }
+
+bool CWindow::shouldSendFullscreenState() {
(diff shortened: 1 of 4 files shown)


src/Window.cpp
```diff
@@ -998,3 +998,8 @@ int CWindow::getRealBorderSize() {
 bool CWindow::canBeTorn() {
     return (m_sAdditionalConfigData.forceTearing.toUnderlying() || m_bTearingHint) && g_pHyprRenderer->m_bTearingEnvSatisfied;
 }
+
+bool CWindow::canBeFullscreen() {
```
vaxerski, line 1002: more like shouldSendFullscreenState

zephvr:
Just replace the method name

vaxerski (approved):
lgtm, thanks!

PR #4116, opened 2023-12-10
hyprpm: unbreak on 32-bit archs via static_cast

jbeich:
Similar to #3589. `length()` returns `std::string::size_type` which may be `unsigned int`.


vaxerski (approved):
lgtm

PR #4145, opened 2023-12-13
Link against tomlplusplus when using cmake

MikeWalrus:
This reduces the compilation time and is consistent with the meson version.


vaxerski (approved):
k

Issue #3990, opened 2023-11-28
Entering an invalid keyboard layout causes Hyprland to crash


BeauTheBeau:
### Hyprland Version

Hyprland, built from branch, at commit 1b48642fd15c433c53876f1b933dcd46265caf8f dirty ().

### Bug or Regression?

Bug

### Description

When changing `input.kb_layout` to a layout that does not exist, e.g. "en" rather than "gb", Hyprland crashes and does not start.

### How to reproduce

- Set `input.kb_layout` to an invalid layout, eg "en"
- If already running, Hyprland will crash
- When executing `Hyprland`, Hyprland will also crash

### Crash reports, logs, images, videos

[log.txt](https://github.com/hyprwm/Hyprland/files/13492936/bBYfI.txt)

romanstingler:
Is this resolved for you?

BeauTheBeau:
Yep!

PR #4157, opened 2023-12-14
Prevent crash with invalid keyboard layout

dranull:
#3990. `memset` sets members of `rules` to null instead of empty strings, which causes a crash at formatting.
Using `pKeyboard->currentRules` instead of `rules` has the same effect, without the crash.

No.

I hope so.


vaxerski (approved):
thanks

theotrst:
same problem. Could you please unpack a bit how to solve the issue? How should I modify .conf file?
thanks

PR #4187, opened 2023-12-18
general:cursor_inactive_position

dusanx:

When hidden, cursor still triggers hover events at the position where it was hidden, for instance web browser will show hyperlink tooltips during scroll when hyperlink falls under hidden cursor position.

This PR adds `general:cursor_inactive_position` [vec2] variable with the defaults [-1 -1]. Defining any other value than -1 -1, for instance

    cursor_inactive_position = 500 200

will move cursor to those x,y coordinates when hiding and restore previous position before becomes visible. Visually doesn't change anything but choosing out of the way x and y prevents tooltips and similar mouse interaction UI to show uninvited.

I have no idea how to add `general:cursor_inactive_position` to documentation (https://wiki.hyprland.org/Configuring/Variables/) so some kind soul must do that for me or tell me how to do it.

💯. Ready to merge.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -88,14 +88,15 @@ void CConfigManager::setDefaultVars() {
     ((CGradientValueData*)configValues["general:col.inactive_border"].data.get())->reset(0xff444444);
     ((CGradientValueData*)configValues["general:col.nogroup_border"].data.get())->reset(0xff444444);
     ((CGradientValueData*)configValues["general:col.nogroup_border_active"].data.get())->reset(0xffff00ff);
-    configValues["general:cursor_inactive_timeout"].intValue = 0;
-    configValues["general:no_cursor_warps"].intValue         = 0;
-    configValues["general:no_focus_fallback"].intValue       = 0;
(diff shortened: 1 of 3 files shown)


vaxerski:
I don't like this, why? You can script it, no?

dusanx:
This is fairly often used logic to hide cursor and prevent unwanted interactions, at least for Xorg equivalents (unclutter and similar). Hiding cursor but not moving it away triggers tooltips and such. Moving cursor but not restoring its position makes cursor 'too far' when needed. I don't know other option (used cursor hiding programs for many years back) to achieve cursor being here but not here when it is hidden.

Maybe I am missing some approach, can you explain what script can achieve? I can move the cursor away using script but then when it is needed and reappears it is away.

Usage example: web browser or text editor, I can select or click, wait for cursor to hide, type, select or click again. If the cursor is not very close that breaks workflow.

If wayland allows cursor moving outside of all surfaces (min_double, min_double) then we don't need variable and moving it away can be automatic. I don't know all the wayland limitations yet to be able to be sure so this variable was a safe bet.

dusanx:
I was suggested better approach on wlroots IRC this morning, so I'll test implementing that and send different MR if it works well.

PR #4199, opened 2023-12-20
filesystem: Set the sticky bit on `/tmp/hypr`

Harazi:
Any user can delete the lock file under `/tmp/hypr` due to permissive permissions. By setting the sticky bit on `/tmp/hypr`, only root or the owner can delete lock files.


vaxerski (approved):
lgtm, thanks!

PR #4207, opened 2023-12-20
bug: fix invalid pointer access

AhmedYasser5:
The compositor sometimes segfaults on exit, so this fix checks first in the handler if the global pointer "g_pHookSystem" is valid before calling "longjump".

If I'm not mistaken, this bug might mean that this behaviour isn't intended, so this might require further lookup in the future.

Yes, it's.


vaxerski (approved):
right, my bad

PR #4256, opened 2023-12-26
Trampoline hooks %rip patching improvements

vaxerski:
Now uses `as` and `objdump` to parse human-readable at&t asm instead of raw bytes.

Tested a bit, but @DreamMaoMao for further testing (hycov) and @jbeich for potential clang/bsd comments.

Seems to work alright on my end.

Should work with `%rip` being used in `mov`, `lea` and `call`

### DIFF
--- a/src/plugins/HookSystem.cpp
+++ b/src/plugins/HookSystem.cpp
@@ -1,5 +1,6 @@
 #include "HookSystem.hpp"
 #include "../debug/Log.hpp"
+#include "../helpers/VarList.hpp"
 
 #define register
 #include <udis86.h>
@@ -19,12 +20,12 @@ CFunctionHook::~CFunctionHook() {
         unhook();
 }
 
-size_t CFunctionHook::getInstructionLenAt(void* start) {
+CFunctionHook::SInstructionProbe CFunctionHook::getInstructionLenAt(void* start) {
     ud_t udis;
 
     ud_init(&udis);
     ud_set_mode(&udis, 64);
-    ud_set_syntax(&udis, UD_SYN_INTEL);
+    ud_set_syntax(&udis, UD_SYN_ATT);
 
     size_t curOffset = 1;
     size_t insSize   = 0;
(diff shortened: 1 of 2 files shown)


src/plugins/HookSystem.cpp
```diff
@@ -41,31 +42,83 @@ size_t CFunctionHook::getInstructionLenAt(void* start) {
...
+        }
+
+        lastAsmNewline = probe.assembly.find("\n", lastAsmNewline) + 1;
+        currentAddress += len;
+    }
+
+    std::ofstream ofs("/tmp/hypr/.hookcode.asm", std::ios::trunc);
+    ofs << assemblyBuilder;
+    ofs.close();
+    execAndGet("as /tmp/hypr/.hookcode.asm -o /tmp/hypr/.hookbinary.o && objcopy -O binary -j .text /tmp/hypr/.hookbinary.o /tmp/hypr/.hookbinary2.o");
```
jbeich, line 107: `as` may not be installed on Clang systems (e.g., FreeBSD). Try using the default compiler binary. With GCC it'd simply invoke `as` but with Clang it'd pass `-cc1as` to invoke integrated assembler.

```suggestion
    execAndGet("cc /tmp/hypr/.hookcode.asm -o /tmp/hypr/.hookbinary.o && objcopy -O binary -j .text /tmp/hypr/.hookbinary.o /tmp/hypr/.hookbinary2.o");
```

jbeich: Err, `cc ... -c ... *.as`.

vaxerski: will `-c` catch that it's an asm file?

jbeich: No, GCC needs explicit `-x assembler` as it doesn't recognize `*.asm` unlike `*.s` or `*.S`. Clang works fine as is but also supports `-x assembler`.

I've edited the suggestion above.

jbeich: Alternatively, rename `.hookcode.asm` to `.hookcode.s`.

vaxerski: thanks!

DreamMaoMao:
Hyprland, built from branch main at commit 191fa587f4daf9f3bf3e3f791de9e2dc5e95a459  (windowrules: add initialTitle and initialClass (4259)).
Date: Wed Dec 27 06:47:46 2023
Tag: v0.33.1-92-g191fa587

flags: (if any)

image

it seem that the problem is still exist

vaxerski:
you didn't checkout to this mr :P

DreamMaoMao:
my bad ,

Hyprland, built from branch trampoline-rip-improvements at commit d8ca95be7668375a4c576f8116bc6066281cc56f  (use cc instead of as).
Date: Wed Dec 27 02:35:45 2023
Tag: v0.33.1-88-gd8ca95be

flags: (if any)
debug


it work normally now.

vaxerski:
cool :D

PR #4259, opened 2023-12-26
implement support for initialTitle and initialClass, closes #2311

### LINKED ISSUE #2311: Adding the initialTitle as supported fields in the windowrulev2
Including the **initialTitle** parameter in the **windowrulev2** configuration allows for greater flexibility in managing applications like Telegram. This is particularly useful when the class name remains the same across different child windows while the title of the window changes based on the selected chat or initiated call. With this configuration, it becomes possible to have Telegram in tile mode while keeping calls within Telegram in float mode.

Here is an example of the configuration:

`windowrulev2 = float, class: ^(org.telegram.desktop)$, initialTitle: ^(?!Telegram$).*`

zjeffer:
This PR adds support for matching `initialTitle` and `initialClass` in `windowrulev2`. It closes #2311.

I only added `INITIALCLASSPOS`, `INITIALTITLEPOS`, `szInitialTitle` and `szInitialClass` in the same places where I found the `title` and `class` rule matchers. I don't know what else needs to be added as I'm not very familiar with the code.

This will also need a wiki update.

Tested and works on a nested debug session, ~~I'd like to try again without nesting to do some more testing before merging.~~ also works without nesting.

Tested with `windowrulev2 = opacity 0.90, initialTitle:^(Spotify( Premium)?)$`

### DIFF
--- a/src/Window.hpp
+++ b/src/Window.hpp
@@ -151,6 +151,8 @@ struct SWindowRule {
     bool        v2 = false;
     std::string szTitle;
     std::string szClass;
+    std::string szInitialTitle;
+    std::string szInitialClass;
     int         bX11         = -1; // -1 means "ANY"
     int         bFloating    = -1;
     int         bFullscreen  = -1;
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
lgtm wiki mr needed

PR #4262, opened 2023-12-27
groupbar: remove extra border size

MightyPlaza:
removes extra border size from before borders became decorations (increasing border increases bar size)

there is also a border reference in the shadows file, I haven't checked it, but it might have a similar issue


vaxerski (approved):
thanks!

PR #4263, opened 2023-12-27
variables update their value when set again

### LINKED ISSUE #4255: variable overshadowing or proper handling of variables when they are set multiple times
### Description

let variables that are set multiple times in config use the value that was last set

flick0:
fixes #4255

```
$var=1
// value of var is now '1'
$var=2
// value of var is now '2'
$var=3
// value of var is now '3'
```

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -394,7 +394,7 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
                 for (auto& [var, val] : configDynamicVars) {
                     if (var == COMMAND.substr(1)) {
                         Debug::log(LOG, "Registered new value for dynamic var \"{}\" -> {}", COMMAND, VALUE);
-                        val = VALUE;
+                        val   = VALUE;
                         found = true;
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/config/ConfigManager.cpp
```diff
@@ -390,9 +390,21 @@ void CConfigManager::configSetValueSafe(const std::string& COMMAND, const std::s
...
+                        found = true;
+                    }
+                }
 
+                if (!found) {
+                    Debug::log(LOG, "Registered dynamic var \"{}\" -> {}", COMMAND, VALUE);
+                    configDynamicVars.emplace_back(std::make_pair<>(COMMAND.substr(1), VALUE));
+                }
+                
                 std::sort(configDynamicVars.begin(), configDynamicVars.end(), [&](const auto& a, const auto& b) { return a.first.length() > b.first.length(); });
```
vaxerski, line 408: sort can be moved to !found, if we don't add a new var sorting is redundant.

vaxerski:
oops code style fails. run clang-format

flick0:
done

vaxerski (approved):
fwiw hyprlang already does this

PR #4275, opened 2023-12-28
hyprctl: add decorations

MightyPlaza:
adds `hyprctl decorations`
allows user to see priority for each decoration so they can (in the future) directly set priority for each one

https://github.com/hyprwm/hyprland-wiki/pull/427

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -424,7 +424,7 @@ int main(int argc, char** argv) {
         request(fullRequest, 1);
     else if (fullRequest.contains("/keyword"))
         request(fullRequest, 2);
-    else if (fullRequest.contains("/decoration"))
+    else if (fullRequest.contains("/decorations"))
         request(fullRequest, 1);
     else if (fullRequest.contains("/hyprpaper"))
         requestHyprpaper(fullRequest);
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -31,6 +31,7 @@ commands:
     binds
     clients
     cursorpos
+    decorations
     devices
     dispatch
     getoption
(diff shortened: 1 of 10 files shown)


Eclextic:
Is this for plugin use?
If so that could be quite useful...

hyprctl/main.cpp
```diff
@@ -423,6 +424,8 @@ int main(int argc, char** argv) {
         request(fullRequest, 1);
     else if (fullRequest.contains("/keyword"))
         request(fullRequest, 2);
+    else if (fullRequest.contains("/decoration"))
```
vaxerski, line 427: mismatch: help says decorations, here it's decoration? Later in hyprland it's decorations again?

src/render/decorations/IHyprWindowDecoration.cpp
```diff
@@ -26,4 +26,8 @@ eDecorationLayer IHyprWindowDecoration::getDecorationLayer() {
 
 uint64_t IHyprWindowDecoration::getDecorationFlags() {
     return 0;
-}
\ No newline at end of file
+}
+
+std::string IHyprWindowDecoration::getDisplayName() {
+    return std::string("Generic Decoration");
```
vaxerski, line 32: Unknown would be better. Also no need for `std::string()`

hyprctl/main.cpp
MightyPlaza: done

src/render/decorations/IHyprWindowDecoration.cpp
MightyPlaza: done

MightyPlaza:
> Is this for plugin use? If so that could be quite useful...

anything that uses that uses the `IHyprWindowDecoration` class

vaxerski (approved):
lgtm

PR #4280, opened 2023-12-28
Unify input handling on decorations

dranull:
As discussed in #4177. Just making a draft to see if I'm on the right path.

Added methods:
```cpp
// picks the right event handler based on eInputType (set in InputManager.hpp), each decoration can use their own logic
// true if the input is consumed
bool IHyprWindowDecoration::onInputOnDeco(const eInputType, const Vector2D&, std::any = 0);

// loops through the decorations and calls the first appropriate `onInputOnDeco` and stops
// true if one of the decos consumed the input
bool CWindow::checkInputOnDecos(const eInputType type, const Vector2D& mouseCoords, std::any data)
```

This PR should not have behavioral changes, it follows the current logic.

(I only changed the positions of `onEndWindowDragOnDeco` and `onBeginWindowDragOnDeco` in the file, git diff shows more dramatic changes)

### DIFF
--- a/src/Window.hpp
+++ b/src/Window.hpp
@@ -348,6 +348,7 @@ class CWindow {
     void                     addWindowDeco(std::unique_ptr<IHyprWindowDecoration> deco);
     void                     updateWindowDecos();
     void                     removeWindowDeco(IHyprWindowDecoration* deco);
+    bool                     checkInputOnDecos(const eInputType, const Vector2D&, std::any = 0);
     pid_t                    getPID();
     IHyprWindowDecoration*   getDecorationByType(eDecorationType);
     void                     removeDecorationByType(eDecorationType);
--- a/src/Window.hpp
+++ b/src/Window.hpp
@@ -348,7 +348,7 @@ class CWindow {
     void                     addWindowDeco(std::unique_ptr<IHyprWindowDecoration> deco);
     void                     updateWindowDecos();
     void                     removeWindowDeco(IHyprWindowDecoration* deco);
-    bool                     checkInputOnDecos(const eInputType, const Vector2D&, std::any = 0);
+    bool                     checkInputOnDecos(const eInputType, const Vector2D&, std::any = {});
     pid_t                    getPID();
     IHyprWindowDecoration*   getDecorationByType(eDecorationType);
     void                     removeDecorationByType(eDecorationType);
--- a/src/render/decorations/CHyprGroupBarDecoration.hpp
+++ b/src/render/decorations/CHyprGroupBarDecoration.hpp
@@ -35,14 +35,6 @@ class CHyprGroupBarDecoration : public IHyprWindowDecoration {
 
     virtual void                       damageEntire();
 
-    virtual bool                       onBeginWindowDragOnDeco(const Vector2D&);
-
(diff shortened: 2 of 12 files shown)


src/render/decorations/CHyprGroupBarDecoration.hpp
```diff
@@ -35,11 +35,15 @@ class CHyprGroupBarDecoration : public IHyprWindowDecoration {
...
-    virtual void                       onBeginWindowDragOnDeco(const Vector2D&);
+    virtual bool                       onBeginWindowDragOnDeco(const Vector2D&);
 
-    virtual bool                       onEndWindowDragOnDeco(CWindow* pDraggedWindow, const Vector2D&);
+    virtual bool                       onEndWindowDragOnDeco(const Vector2D&, CWindow*);
 
-    virtual void                       onMouseButtonOnDeco(const Vector2D&, wlr_pointer_button_event*);
+    virtual bool                       onMouseButtonOnDeco(const Vector2D&, wlr_pointer_button_event*);
+
+    virtual bool                       onScrollOnDeco(const Vector2D&, wlr_pointer_axis_event*);
```
vaxerski, line 44: these no longer need to be virtual; they do not override anything. They can also go private.

vaxerski (commented):
I like this. Way less clutter in the code, de-dup, and stuff.

src/render/decorations/CHyprGroupBarDecoration.hpp
dranull: Done.

Do any of those need to be marked as virtual, since as far as I understand virtual is inherited? Or you're just being explicit?

vaxerski: no point in these being virtual as nothing will inherit groupbar

src/Window.hpp
```diff
@@ -348,6 +348,7 @@ class CWindow {
     void                     addWindowDeco(std::unique_ptr<IHyprWindowDecoration> deco);
     void                     updateWindowDecos();
     void                     removeWindowDeco(IHyprWindowDecoration* deco);
+    bool                     checkInputOnDecos(const eInputType, const Vector2D&, std::any = 0);
```
vaxerski, line 351: leave any uninitialized, 0 might give it an `int` type with value `0` instead of it being empty

vaxerski (commented):
generally lgtm

src/render/decorations/CHyprGroupBarDecoration.hpp
dranull: No, I meant "those". "Those" as in any of those other public methods (`getDecorationType`, `updateWindow`...).

src/Window.hpp
dranull: Currently there is no `data` set with `INPUT_TYPE_DRAG_START`, so this is how I called it in `KeybindManager`: `pWindow->checkInputOnDecos(INPUT_TYPE_DRAG_START, mouseCoords);`

If I remove the default 0, then this won't work (won't compile).
Or should I always pass a third parameter? If yes, what type?

vaxerski: `= {}`

dranull: Done, rebased.

vaxerski (approved):
lgtm, thanks as usual

Issue #2831, opened 2023-07-28
Completely hide the groupbar


tgy:
### Description

Is there a way the groupbar can be completely hidden?

I currently have the following

```
misc {
    render_titles_in_groupbar = false
}
```

but the bar still remains (with no title)

MightyPlaza:
use workspace rules, and disable decorations.
not working for floating windows for some reason

spikespaz:
Added as item `#8` to #2415, lets keep aesthetic requests for group bars in that issue. Recommend to others to :+1: that issue.

PR #4299, opened 2023-12-30
groupbar: allow enabling groupbar and setting priority

MightyPlaza:
adds groupbar `enabled` and `priority`

fixes bug where moving to a workspace that is not decorated will not remove the reserved decoration area
also deletes / doesn't create gradient textures if not enabled in config 

closes: https://github.com/hyprwm/Hyprland/issues/2831

https://github.com/hyprwm/hyprland-wiki/pull/428


vaxerski (approved):
top

PR #4304, opened 2023-12-30
fix keys getting stuck + minor refactor & optimizations to keybind handling

### LINKED ISSUE #4013: Keys being "toggled" if while pressed SUPER is also pressed.
### Hyprland Version

Hyprland, built from branch main at commit e40e486f61f2643578b9977b86f408799dbc75fd dirty (renderer: better checks for special rendering in renderWorkspaceWindows). Tag: v0.32.3-50-ge40e486f  flags: (if any)

### Bug or Regression?

Bug

### Description

With a bind like `SUPER+Q` to open the terminal, if while holding `Q` I were to tap `SUPER`, then, even after releasing all keys, `Q` stays "pressed".

It only seems to happen with keys used in binds, I haven't tested if the same would happen using a modifier key other than `SUPER`.

key goes _**brrrrrrr**_

### How to reproduce

With a bind, like `SUPER+Q`  (MODS: `SUPER` + key: `Q`),
while holding `Q`, press `SUPER`, then release keys,
`Q` acts like its still being pressed.

_No response_

rszyma:

Fixes #4013 

It's fixed by keeping track of mods pressed at the time of key activation and using this info to correctly handle a case, where key was pressed before mod.

Analogous issue with `bindr` was also fixed.

Also:
- Remove unnecessary duplicate `handleKeybinds` calls, which were supposed to be "keycode check runs" but in reality didn't do anything useful.
- "Merge" 3 related params in `handleKeybinds` to a single `SPressedKeyWithMods`

Nothing in particular.

- ~I'm not sure what to do on line 323.~
- ~Maybe code around 323-327 could be optimized, since the lambdas are the same. But idk C++ good enough to think of any improvement.~

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -301,14 +301,14 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
             m_pActiveKeybind            = nullptr;
         }
 
-        const auto key = SPressedKeyWithMods{
+        const auto KEY = SPressedKeyWithMods{
             .keysym             = keysym,
             .keycode            = KEYCODE,
             .modmaskAtPressTime = MODS,
         };
-        m_dPressedKeys.push_back(key);
+        m_dPressedKeys.push_back(KEY);
 
-        found = handleKeybinds(MODS, "", key, true, e->time_msec);
+        found = handleKeybinds(MODS, "", KEY, true, e->time_msec);
 
         if (found)
             shadowKeybinds(keysym, KEYCODE);
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -310,7 +310,7 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
 
         m_dPressedKeys.push_back(KEY);
 
-        found = handleKeybinds(MODS, KEY, true, e->time_msec);
+        found = handleKeybinds(MODS, KEY, true);
 
         if (found)
             shadowKeybinds(keysym, KEYCODE);
@@ -326,7 +326,7 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
         for (auto it = m_dPressedKeys.begin(); it != m_dPressedKeys.end();) {
             if (it->keycode == KEYCODE) {
                 if (!foundInPressedKeys) {
-                    found              = handleKeybinds(MODS, *it, false, e->time_msec);
+                    found              = handleKeybinds(MODS, *it, false);
                     foundInPressedKeys = true;
                 }
                 it = m_dPressedKeys.erase(it);
@@ -337,7 +337,7 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
         if (!foundInPressedKeys) {
             Debug::log(ERR, "BUG THIS: key not found in m_dPressedKeys");
             // fallback with wrong `KEY.modmaskAtPressTime`, this can be buggy
-            found = handleKeybinds(MODS, KEY, false, e->time_msec);
+            found = handleKeybinds(MODS, KEY, false);
         }
 
         shadowKeybinds();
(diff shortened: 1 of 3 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -319,12 +320,12 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
             m_pActiveKeybind            = nullptr;
         }
 
-        m_dPressedKeycodes.erase(std::remove(m_dPressedKeycodes.begin(), m_dPressedKeycodes.end(), KEYCODE), m_dPressedKeycodes.end());
-        m_dPressedKeysyms.erase(std::remove(m_dPressedKeysyms.begin(), m_dPressedKeysyms.end(), keysym), m_dPressedKeysyms.end());
-
-        found = handleKeybinds(MODS, "", keysym, 0, false, e->time_msec) || found;
+        auto it = std::find_if(m_dPressedKeys.begin(), m_dPressedKeys.end(), [&](const SPressedKeyWithMods& other) { return other.keycode == KEYCODE; });
+        // If not found, crash on purpose, because it's wrong and should be fixed.
```
vaxerski, line 324: don't crash on purpose in something like this.
`Debug::log(ERR, "BUG THIS <some info>");`

src/managers/KeybindManager.cpp
```diff
@@ -319,12 +320,12 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
...
+        found = handleKeybinds(MODS, "", *it, false, e->time_msec);
 
-        found = handleKeybinds(MODS, "", 0, KEYCODE, false, e->time_msec) || found;
+        m_dPressedKeys.erase(std::remove_if(m_dPressedKeys.begin(), m_dPressedKeys.end(), [&](const SPressedKeyWithMods& other) { return other.keycode == KEYCODE; }),
```
vaxerski, line 327: `std::erase_if` please

src/managers/KeybindManager.cpp
```diff
@@ -302,12 +301,14 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
             m_pActiveKeybind            = nullptr;
         }
 
-        m_dPressedKeycodes.push_back(KEYCODE);
-        m_dPressedKeysyms.push_back(keysym);
+        const auto key = SPressedKeyWithMods{
```
vaxerski, line 304: const are CAPS

rszyma: yep

src/managers/KeybindManager.cpp
rszyma: d700d2a64878e433746cf820efb0f9b9e1cc586b

src/managers/KeybindManager.cpp
rszyma: done 52d2df9ab38ce8408103513c06d14e579b68284e

rszyma:
> Maybe code around 323-327 could be optimized, since the lambdas are the same. But idk C++ good enough to think of any improvement.

Addressed this in [d6d1f52](https://github.com/hyprwm/Hyprland/pull/4304/commits/d6d1f5243336fd7371657b1ce4c807b83e81a947)

rszyma:
Wait, there's a bug, when Press(Mod)+Press(Key)+Release(Mod)+Release(Key) a release for Key is sent

rszyma:
Oh wait, it was there before, so it's a different bug

src/managers/KeybindManager.cpp
```diff
@@ -424,7 +441,7 @@ int repeatKeyHandler(void* data) {
     return 0;
 }
 
-bool CKeybindManager::handleKeybinds(const uint32_t& modmask, const std::string& key, const xkb_keysym_t& keysym, const int& keycode, bool pressed, uint32_t time) {
+bool CKeybindManager::handleKeybinds(const uint32_t modmask, const std::string& key, const SPressedKeyWithMods keyNumeric, bool pressed, uint32_t time) {
```
vaxerski, line 444: since this is const yu can avoid copying with `const SPressedKeyWithMods&`

vaxerski (commented):
too tired to think, but apart from that I think its alright

src/managers/KeybindManager.cpp
rszyma: Sure thing chief

rszyma:
I think there's now a bug with `bindr` for mouse buttons not working (not tested though)

rszyma:
`bindr` with mouse buttons still not working

I figured out that it get shadowed, but it worked before, so I must've fked up somewhere, and I've yet to discover it.

In what situations is shadowing even useful? I can't figure it out.

vaxerski:
> In what situations is shadowing even useful? I can't figure it out.

```ini
bindr = SUPER, SUPER_L, exec, $launcher
bind = SUPER, E, exec, $fm
```

rszyma:
Another bug, probably related to the previous one:

This is not working (command is not executed):
```
bindr = SUPER, SUPER_L, exec, $launcher
```

But I see where's the problem: I assumed that `MODS` include the currently pressed key when pushing to the pressed keys here https://github.com/hyprwm/Hyprland/blob/ef2fe9e057be6fe75c9d674e322e761487d8b00c/src/managers/KeybindManager.cpp#L313

rszyma:
> Another bug, probably related to the previous one:
> 
> This is not working (command is not executed):
> 
> ```
> bindr = SUPER, SUPER_L, exec, $launcher
> ```

this is fixed now



> `bindr` with mouse buttons still not working

this is still not working

rszyma:
I've fixed every issue with this MR, ready for review again

src/managers/KeybindManager.hpp
```diff
@@ -12,7 +12,7 @@ class CPluginSystem;
 
 struct SKeybind {
     std::string key          = "";
-    int         keycode      = -1;
+    uint32_t    keycode      = 0;
```
vaxerski, line 15: why this change? keycode 0 is valid, iirc. Maybe some `std::optional`s here?

rszyma: >why this change?

Explained in commit message here [`6fcc7a9` (#4304)](https://github.com/hyprwm/Hyprland/pull/4304/commits/6fcc7a9179e20d7e57f02e50068ca1206b4b25c1)

Isn't keycode 0 used for lack-of-keycode in xkb too?

vaxerski: k then

vaxerski (approved):
top

PR #4306, opened 2023-12-30
Ignore some input events when focus is on a layer surface

### LINKED ISSUE #3737: Mouse scroll events are sent to wrong grouped window until the cursor is moved
### Hyprland Version

88b47dfa833db2d230a2d6e6bbbf92c69afbdc28

### Bug or Regression?

Bug

### Description

Similar to https://github.com/hyprwm/Hyprland/issues/3679, but for switching between grouped windows, if you switch between them without moving the cursor, scroll events are sent to the window that last had a cursor move while they were focused. Seems to only happen when native wayland windows are involved, switching from xwayland to xwayland works as intended, any other combinations and the bug happens.

### How to reproduce

Set `input:follow_mouse` to anything but 0, open 2 scrollable windows, at least one of them being native wayland, switch between them without moving the cursor

Older video, but exact same behavior from https://github.com/hyprwm/Hyprland/issues/3679#issuecomment-1784132717

dranull:
The video tries to demonstrate all the listed issues (using waybar):
- Clicking on workspace buttons sends a motion event to the focused window, and because of that the cursor changes, also the focus is lost from waybar (as you can see the (top-border) hover effect disappears). You have to move the cursor to bring back waybar focus and change back the cursor.
- Similar/same issue when scrolling, but since the window gets the focus, scrolling events are sent to the window, instead of waybar, so the window will scroll not the workspaces.
- (not waybar related) Moving the cursor over non reserved area, and switching workspaces via shortcuts changes the cursor even if there's nothing under it.

Maybe #3737, at least the second comment there mentions my second issue.

Also ignore scroll events when an overlay is over the groupbar (or any window decoration)

https://github.com/hyprwm/Hyprland/assets/150595692/e41fea0d-6391-46d6-b6f6-7d6d23038656

Yes.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -880,10 +880,12 @@ void CKeybindManager::changeworkspace(std::string args) {
     } else
         pWorkspaceToChangeTo->rememberPrevWorkspace(PCURRENTWORKSPACE);
 
-    if (!g_pCompositor->m_pLastFocus)
-        g_pInputManager->simulateMouseMovement();
-    else
-        g_pInputManager->sendMotionEventsToFocused();
+    if (!g_pInputManager->m_bLastFocusOnLS) {
+        if (g_pCompositor->m_pLastFocus)
+            g_pInputManager->sendMotionEventsToFocused();
+        else
+            g_pInputManager->simulateMouseMovement();
+    }
 }
 
 void CKeybindManager::fullscreenActive(std::string args) {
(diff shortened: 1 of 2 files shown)


bvr-yr:
ohmy, it finally happens :heart:

vaxerski:
since this is all about LS-es, why not check if focus is on a LS instead?

dranull:
Because I totally knew what LS is, did not skip ANY wayland basics, and I have a deep understanding of the codebase... :innocent:

vaxerski (approved):
lgtm

PR #4327, opened 2024-01-02
Drag single window instead of destroying group

dranull:
Prevents destroying a group with only one window when that's dragged from the groupbar. It will just normally drag/move the window instead. `togglegroup` still can be used if someone wants to deactivate the group.

Yes.


vaxerski (approved):
lgtm

PR #4361, opened 2024-01-05
Unbreak install with CMake on FreeBSD

jbeich:
Regressed by 78f9ba9fdd7e. See also [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ln.html).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ install:
 	chmod 755 ${PREFIX}/bin/Hyprland
 	chmod 755 ${PREFIX}/bin/hyprctl
 	chmod 755 ${PREFIX}/bin/hyprpm
-	ln -s -r -f ${PREFIX}/bin/Hyprland ${PREFIX}/bin/hyprland
+	cd ${PREFIX}/bin && ln -sf Hyprland hyprland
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,7 +42,7 @@ install:
 	chmod 755 ${PREFIX}/bin/Hyprland
 	chmod 755 ${PREFIX}/bin/hyprctl
 	chmod 755 ${PREFIX}/bin/hyprpm
-	ln -s -r -f ${PREFIX}/bin/Hyprland ${PREFIX}/bin/hyprland
+	cd ${PREFIX}/bin && ln -sf Hyprland hyprland
```
fufexan, line 45: Do we not need to `cd -` afterwards?

jbeich: Shouldn't be necassary as each command line is executed within a subshell. Otherwise, the style isn't consistent with other `cd` instances in Makefile.

```
$ cd /tmp
$ cat Makefile
all:
	cd /var/empty
	pwd

$ bmake
cd /var/empty
pwd
/tmp

$ gmake
cd /var/empty
pwd
/tmp
```

fufexan: Alright then, thanks for the explanation.

fufexan (approved):

Makefile
jbeich: One can't reliably `cd` out of an absolute path of arbitrary length. Another option is to wrap the command into parentheses to use current directory of `sh` instead of `make`.

PR #4386, opened 2024-01-08
Set radius for each window's corner

Micovec:

This PR opens a possibility to define a radius for each window's corner separately. This can come as a new way for ricing Hyprland in combination of modified `hyprwm/hyprland-plugins/hyprbars` (please don't judge color scheming, I set them to have a high contrast).

image

As you can see window has rounded corners only at the bottom and hyprbars only at the top. They become more unified and it actually seems like `hyprbars` is part of the window.

Rounding has additional definition:
```
decoration {
    rounding = 15 # old way
    rounding = 0,0,15,15 # new way: topLeft, topRight, bottomRight, bottomLeft
}
```

However it is not done yet and there are still some issues. For example: window border isn't part of the bar (in picture set to size 7) as it still renders around the original window and not around `hyprbars`.

image

On top of that as you can see (in highlighted red rectangle) the border is rounded even though radius is set to `0`. To keep it consistent there shouldn't be any rounding.

When merged it will break `hyprwm/hyprland-plugins/hyprbars` and `hyprwm/hyprland-plugins/borders-plus-plus` as they rely on `CWindow::rounding()` and not the changed `CWindow::getCornerRadii()`.

Also I'm not sure if substraction box for renderdata.damage has correct values. I've left a `TODO` comment there.

PR is <b>not</b> ready for merging. I just wanted to see if it's a valid feature to add into `Hyprland` and if there's any interest for it.

If it will be decided to merge this, I'll make another PR into `hyprwm/hyprland-plugins` to fix compatibility and rounding.


Eclextic:
Probably not supported by the API but can one add multiple ways of defining variables, vaxry?
In this example it would be perfect:
```
decoration {
    rounding = 15 # old way => would become the same as rounding = 15,15,15,15
    rounding = 0,0,15,15 # new way => when needed users can use this syntax
}
```

Micovec:
> Probably not supported by the API but can one add multiple ways of defining variables, vaxry? In this example it would be perfect:
> 
> ```
> decoration {
>     rounding = 15 # old way => would become the same as rounding = 15,15,15,15
>     rounding = 0,0,15,15 # new way => when needed users can use this syntax
> }
> ```

Yes. I kept the old way of defining rounding with one integer and added a new way with four separated by comma.

This
```
rounding = 15
```
then gets mapped as:
```
rounding = 15,15,15,15
```

vaxerski:
@Eclextic already possible.

cristian-ciobanu:
This would be a nice addition at least for people who use hyprbars as plugin.

charlesfire:
I was looking to see if it was possible to set corner-radius on a per-corner basis and I found this merge request. I'm in favor of this feature. Giving more customization options for users, when it doesn't add unnecessary complexity, is always a good thing.

Micovec:
> I was looking to see if it was possible to set corner-radius on a per-corner basis and I found this merge request. I'm in favor of this feature. Giving more customization options for users, when it doesn't add unnecessary complexity, is always a good thing.

I appreciate the support but instead of commenting it's better to give thumbs up to the PR (first comment) so that it's easier to see which PRs are wanted by people.

Currently there's big migration away from wlroots. That includes some changes in rendering. When that's done and polished I'll take a second look on this and will try to finish it.

vaxerski:
closing due to inactivity

PR #4390, opened 2024-01-08
groupbar: improve gradient handling

MightyPlaza:
uses all provided gradient colors instead of just using the 1st one
changes default behaviour from gradient using default color to solid fill using default color

didn't find any issues


vaxerski (approved):
lgtm

like and subscribe

PR #4401, opened 2024-01-10
Fix greedy mouse grab from keyboard_interactive layer

### LINKED ISSUE #2017: Strange mouse grab on layers with keyboard interactivity
I think this is probably a bug, found it testing the eww sys tray https://github.com/elkowar/eww/pull/743#issuecomment-1501716248

there are multiple ways to reproduce, enable `:focusable` in an eww window or just run any layer-shell example like below with keyboard interactivity

I have  `follow_mouse = 2`

the smithay example is with OnDemand but its same with Exclusive too

```
git clone https://github.com/Smithay/client-toolkit/
cd client_toolkit 
cargo run --release --example simple_layer
```

expecatation:
mouse events only go to the layer surface if the cursor is over the surface, if its over another surface events go to that other surface

actual:
mouse events seem to go to the layer surface no matter where i click on the screen

sway: mouse clicks go to the intended target surface, not to the layer if mouse is not over the layer surface
does not happen with kb interactivity disabled

so I'm wondering what might keyboard interactivity have to do with mouse events

earboxer:

Fixes #2017

This allows you to use your mouse on other (not covered) surfaces even when you have a keyboard_interactive layer active.

The bug was that a keyboard_interactive layer would always take precedence over everything else (even without KEYBOARD_INTERACTIVITY_EXCLUSIVE). The fix makes that layer not take precedence, and instead makes `refocus` respect KEYBOARD_INTERACTIVITY_EXCLUSIVE (i.e. the correct behavior).

So, you can
* use wvkbd to type into bemenu
   * click on an eww element even when rofi is open
* click or scroll another window when (some other keyboard_interactive layer) is open
    * this will not steal the keyboard focus from a KEYBOARD_INTERACTIVITY_EXCLUSIVE layer

One case I found where the KEYBOARD_INTERACTIVITY_EXCLUSIVE is not respected:

* keyboard focus will still be stolen (from the layer to a window) when you switch focus using movefocus (e.g. $mainMod, left)

this is an existing bug/feature (it may be intended behavior). Maybe it should be fixed now (so KEYBOARD_INTERACTIVITY_EXCLUSIVE layers keep the keyboard), or maybe fix it some other day?

This is ready. It fixes the unexpected and incorrect behavior that bothers me. (the existing movefocus behavior can be changed some other time, if that's what we want.)

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -359,7 +359,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
 
     bool allowKeyboardRefocus = true;
 
-    if (!refocus && g_pCompositor->m_pLastFocus) {
+    if (g_pCompositor->m_pLastFocus) {
         const auto PLS = g_pCompositor->getLayerSurfaceFromSurface(g_pCompositor->m_pLastFocus);
 
         if (PLS && PLS->layerSurface->current.keyboard_interactive == ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE)
@@ -421,7 +421,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
             m_bLastFocusOnLS = false;
             return; // don't enter any new surfaces
         } else {
-            if (((FOLLOWMOUSE != 3 && allowKeyboardRefocus) && (*PMOUSEREFOCUS || m_pLastMouseFocus != pFoundWindow)) || refocus) {
+            if (allowKeyboardRefocus && ((FOLLOWMOUSE != 3 && (*PMOUSEREFOCUS || m_pLastMouseFocus != pFoundWindow)) || refocus)) {
                 m_pLastMouseFocus = pFoundWindow;
                 g_pCompositor->focusWindow(pFoundWindow, foundSurface);
             }
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1102,14 +1102,6 @@ wlr_surface* CCompositor::vectorToLayerSurface(const Vector2D& pos, std::vector<
 
         auto SURFACEAT = wlr_layer_surface_v1_surface_at(ls->layerSurface, pos.x - ls->geometry.x, pos.y - ls->geometry.y, &sCoords->x, &sCoords->y);
 
-        if (ls->layerSurface->current.keyboard_interactive && ls->layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP) {
-            if (!SURFACEAT)
-                SURFACEAT = ls->layerSurface->surface;
-
-            *ppLayerSurfaceFound = ls.get();
-            return SURFACEAT;
-        }
-
         if (SURFACEAT) {
             if (!pixman_region32_not_empty(&SURFACEAT->input_region))
                 continue;


src/managers/input/InputManager.cpp
```diff
@@ -421,7 +421,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
             m_bLastFocusOnLS = false;
             return; // don't enter any new surfaces
         } else {
-            if (((FOLLOWMOUSE != 3 && allowKeyboardRefocus) && (*PMOUSEREFOCUS || m_pLastMouseFocus != pFoundWindow)) || refocus) {
+            if ((FOLLOWMOUSE != 3 && allowKeyboardRefocus) && ((*PMOUSEREFOCUS || m_pLastMouseFocus != pFoundWindow) || refocus)) {
```
earboxer, line 424: (not completely sure this is completely right (haven't tested with FOLLOWMOUSE = 3). Just wanted to make `refocus` respect `allowKeyboardRefocus`)

vaxerski: why is this even in this MR? doesn't seem related

vaxerski:
from a glimpse I don't think this will work well?

Some cases:
 - open a launcher, hit a keybind to `movefocus,l` to a window. LS will get completely unfocused...?
 - open a launcher and click on an OSK, will the launcher still have kb focus?

earboxer:
On Wed Jan 10, 2024 at 9:11 AM EST, Vaxry wrote:
>  - open a launcher, hit a keybind to `movefocus,l` to a window. LS will get completely unfocused...?
Yes, until you mouse over it. This is an existing bug (though currently its only unfocused until you wiggle the mouse).

>  - open a launcher and click on an OSK, will the launcher still have kb focus?
Yes, launcher will have focus. Existing code ensures that a non-keyboard layer can't take keyboard focus.

vaxerski:
fair enough.

src/managers/input/InputManager.cpp
vaxerski: I also don't understand the point here?

earboxer: I tweaked the logic so that this won't change how FOLLOWMOUSE = 3 behaves.

src/managers/input/InputManager.cpp
```diff
@@ -421,7 +421,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
             m_bLastFocusOnLS = false;
             return; // don't enter any new surfaces
         } else {
-            if (((FOLLOWMOUSE != 3 && allowKeyboardRefocus) && (*PMOUSEREFOCUS || m_pLastMouseFocus != pFoundWindow)) || refocus) {
+            if (allowKeyboardRefocus && ((FOLLOWMOUSE != 3 && (*PMOUSEREFOCUS || m_pLastMouseFocus != pFoundWindow)) || refocus)) {
```
earboxer, line 424: The idea is only allow 'refocus' if allowKeyboardRefocus is true.

(I could have changed `refocus` to `(allowKeyboardRefocus && refocus)`, but I just extracted allowKeyboardRefocus to the front since it's required in both cases)

src/managers/input/InputManager.cpp
earboxer: The point is that we shouldn't refocus (change keyboard focus) when `allowKeyboardRefocus` is false (even on a click/force refocus), because there's a layer that has grabbed the keyboard.

vaxerski: ok

vaxerski (approved):
lgtm

feschber:
#4465

feschber:
From my testing, the offending change is this:
```diff
diff --git a/src/managers/input/InputManager.cpp b/src/managers/input/InputManager.cpp
index 8bfdb65c..ce7e2b30 100644
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -359,7 +359,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
 
     bool allowKeyboardRefocus = true;
 
-    if (g_pCompositor->m_pLastFocus) {
+    if (!refocus && g_pCompositor->m_pLastFocus) {
         const auto PLS = g_pCompositor->getLayerSurfaceFromSurface(g_pCompositor->m_pLastFocus);
 
         if (PLS && PLS->layerSurface->current.keyboard_interactive == ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE)
```

earboxer:
There is a partial regression from the two commits discussed here

https://github.com/hyprwm/Hyprland/issues/4465#issuecomment-1913149150

(If I open bemenu/rofi/nwggrid while there's no window open, it won't get keyboard events from wvkbd)

PR #4407, opened 2024-01-10
Round window pos and size on togglefloating

Clyybber:
This fixes a window sometimes ending up blurry after a toggle from tiled to floating.

Also restore the behaviour introduced in bc4a51dbbb285391c9ed45b1b6ca7c2037b7ba71


vaxerski (approved):
lgtm

Issue #4076, opened 2023-12-06
Swithing to a workspace which has a fullscreen window instantly hides the previous workspace


dranull:
### Hyprland Version

5489f9f07a73c6b5b97702731a5092463a01fb5b

### Bug or Regression?

Bug

### Description

Video...
I already know that fullscreen hides all other windows on the same workspace (with an animation though), however when switching from another (non-fullscreen) workspace its content is already empty when the animation starts.

### How to reproduce

(default config works as well)
- Two workspaces
- Open a window on the first one
- Now on the second one, and make it fullscreen (normal or fake)
- Switch between the workspaces

### Crash reports, logs, images, videos

https://github.com/hyprwm/Hyprland/assets/150595692/ea2f0108-2169-4bd3-9332-1a0d625792f2

Roger-Roger-debug:
Just stumbled on this myself, I think #3928 introduced the behavior.

dranull:
Normal workspace to fullscreen animation now works nicely, though fullscreen to fullscreen doesn't (both direction).

Also, watch the close animation of the (fullscreen) windows at the end, they jump to the left side before animating. Normally they just slide down. But this might be a different issue, that I just noticed.

https://github.com/hyprwm/Hyprland/assets/150595692/dc2c0ec3-0d98-436e-8c23-f8a464ebce69

Dickby:
In my case this doesn't fix the issue at all!
I still get hidden windows during the animation if the new workspace is fullscreen

vaxerski:
oops

Dickby:
I have to correct myself, it's as @dranull wrote, it's working unless both workspaces have fullscreen windows, sorry!

Roger-Roger-debug:
I haven't tested the old behavior, but when toggling special workspaces on top of a fullscreen workspace you get the fade in animation but no fade out (window disappears instantly).

PR #4417, opened 2024-01-11
Prevent hiding current workspace when swiping

scorpion-26:
When workspace_swipe_use_r is enabled, swiping from WS 1 to a non-empty WS 2 would hide WS 1 (Similar effect to issue #4076). This is caused by a faulty check which doesn't consider, that workspaceIDLeft could be the current workspace.
This bug is only a problem for r, because m wraps around on WS 1 m-1, whereas r stays on WS 1.

The first check ```(workspaceIDRight != ...)``` is technically not needed for the bug fix. Though to be extra sure I added it, since there is no reason why you'd want to hide the current workspace when swiping. If you want, I can remove it though.
I also have not checked ```+/-1```, though I see absolutely no reason why this would break something.

Ready.


vaxerski (approved):
ok

PR #4425, opened 2024-01-12
Options for shadow/rounding when only 1 window

end-4:
- The options and their defaults:
```
dwindle:no_shadow_when_only = true
dwindle:no_rounding_when_only = true
master:no_shadow_when_only = true
master:no_rounding_when_only = true
```
- "when only" = there's only 1 window in workspace, similar to no_gaps_when_only

- nope

- i hope i don't mess up with only 12 changed lines 😬


vaxerski:
I dont like this. I want to ultimately deprecate no_gaps_when_only altogether in favor of window/workspace rules.

Why don't you just `windowrulev2=rounding 0, onworkspace:1` or something?

end-4:
Alright
I wanted rounding enabled even when I use `no_gaps_when_only`. Currently the option just disables it.
I didn't think of using this windowrule, though when I tried it just now it doesn't work with `no_gaps_when_only`.

Issue #747, opened 2022-09-29
Better workspace management with multiple displays/monitors


erlonbie:
Hi, first of all thanks for this project, I'm in love with it.

As the title suggests, I think there is room for improvement when managing multiple displays.

Supose that we have two monitors, monitor_1 (left) and monitor_2 (right) 

If I'm looking at the monitor_1 and I switch to a  workspace that is in the monitor_2, it woud be nice (in my opinion) if the worskpace that was in monitor_2 got swapped with the previous workspace that was in monitor_1.This way I don't have to constantly turn my head to a different monitor. I know this sound silly, but it makes a huge difference in usability (again, a personal opinion). In summary, the priority where the workspace should go is the current monitor of the focused window, and not the current monitor of the workspace that were are trying to access.

I'm aware of the Dispatchers `moveworkspacetomonitor` and `swapactiveworkspaces`, but couldn't figure it out how to solve this problem.

Should be something like this:

(I'm using monitor_1)

**if** workspace_desired is in monitor_2 && showing_in_monitor_2: `swapactiveworkspaces` 1 2
**else if** workspace_desired is in monitor_2: `moveworkspacetomonitor` workspace_desired 1,  `moveworkspacetomonitor` previous_workspace 2.
**else**: `workspace` desired_workspace

That said, is this achievable through hyprctl (maybe with some bash scripting) or just tweaking the config file? If so, any help would be appreciated

Thanks again!

vaxerski:
```
bind=SUPER,1,moveworkspacetomonitor,1 current
bind=SUPER,1,workspace,1
```
try this

erlonbie:
> ```
> bind=SUPER,1,moveworkspacetomonitor,1 current
> bind=SUPER,1,workspace,1
> ```
> 
> try this

wow, that just worked! I feel ashamed that was a simple binding.

so, I can put bindings and rules mapping the same thing and the first one is gets priority?

vaxerski:
you are allowed to assign multiple binds to one key. They will be executed in the order they appear in the config.

erlonbie:
sorry to reopen it, but it's still not accomplishing 100% that I mentioned in this issue. 

I noticed that it does not swapps if both workspaces are showing in each monitor. Is this possible to do?

If I'm focused in monitor_1 and app A is in monitor_1 and app B is in monitor_2. It brings app B to monitor_1 and leaves monitor_2 empety (no workspace).

vaxerski:
[...]

PR #4439, opened 2024-01-14
Implement a dispatcher for xmonad/qtile-style workspace switching

virchau13:

Implements the `focusWorkspaceOnCurrentMonitor` dispatcher and function, which implements XMonad/Qtile-style workspace switching.

https://github.com/hyprwm/Hyprland/assets/16955157/0e7922ad-95f2-4fd1-b229-431e8311e85d

When called, `focusworkspaceoncurrentmonitor` will:
1. Send the requested workspace to the current monitor,
2. If the requested workspace was previously active on a different monitor, replace it with the workspace that was previously active on the current monitor,
3. Focus the requested workspace on the current monitor.

#### Why is this necessary to implement in the core repository?

This isn't possible to implement properly as a plugin/script because it requires changes in core functions and is inherently racy. Note that all the samples you can find online trying to achieve this behaviour (everything in #747, everything in [this reddit post](https://www.reddit.com/r/hyprland/comments/11z8w7u/make_hyperland_workspaces_work_like_qtile_work/)) have race condition issues (either time-of-check-to-time-of-use or simultaneous executions resulting in an incorrect window state), aren't very performant, and most of them have issues with cursor warping.

(I have a contribution ready for the hyprland wiki documenting this dispatcher. Should I open a PR there too?)

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -19,64 +19,65 @@
 CKeybindManager::CKeybindManager() {
     // initialize all dispatchers
 
-    m_mDispatchers["exec"]                          = spawn;
-    m_mDispatchers["execr"]                         = spawnRaw;
-    m_mDispatchers["killactive"]                    = killActive;
-    m_mDispatchers["closewindow"]                   = kill;
-    m_mDispatchers["togglefloating"]                = toggleActiveFloating;
-    m_mDispatchers["workspace"]                     = changeworkspace;
-    m_mDispatchers["renameworkspace"]               = renameWorkspace;
-    m_mDispatchers["fullscreen"]                    = fullscreenActive;
-    m_mDispatchers["fakefullscreen"]                = fakeFullscreenActive;
-    m_mDispatchers["movetoworkspace"]               = moveActiveToWorkspace;
-    m_mDispatchers["movetoworkspacesilent"]         = moveActiveToWorkspaceSilent;
-    m_mDispatchers["pseudo"]                        = toggleActivePseudo;
-    m_mDispatchers["movefocus"]                     = moveFocusTo;
-    m_mDispatchers["movewindow"]                    = moveActiveTo;
-    m_mDispatchers["swapwindow"]                    = swapActive;
-    m_mDispatchers["centerwindow"]                  = centerWindow;
-    m_mDispatchers["togglegroup"]                   = toggleGroup;
-    m_mDispatchers["changegroupactive"]             = changeGroupActive;
-    m_mDispatchers["movegroupwindow"]               = moveGroupWindow;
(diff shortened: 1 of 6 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -1504,6 +1505,54 @@ void CKeybindManager::moveWorkspaceToMonitor(std::string args) {
...
+
+    if (!PCURRMONITOR) {
+        Debug::log(ERR, "focusWorkspaceOnCurrentMonitor monitor doesn't exist!");
+        return;
+    }
+
+    auto PWORKSPACE = g_pCompositor->getWorkspaceByID(WORKSPACEID);
+
+    if (!PWORKSPACE) {
+        // no such workspace, make one
```
vaxerski, line 1527: redundant

src/managers/KeybindManager.cpp
```diff
@@ -1504,6 +1505,54 @@ void CKeybindManager::moveWorkspaceToMonitor(std::string args) {
...
+        std::string workspaceName;
+        PWORKSPACE = g_pCompositor->createNewWorkspace(WORKSPACEID, PCURRMONITOR->ID, workspaceName);
```
vaxerski, line 1529: why are you passing an empty workspaceName? It's default init'd to empty, thr arg can be skipped

vaxerski (commented):
yes wiki mr pls

vaxerski (approved):
lgtm

dariogoetz:
Awesome! A thousand thanks to you! This removes the only remaining hurdle for me to productively use hyprland!

svyatoclav:
That's exactly what I want, thank you so much! But there is a small issue for me...

When I have an empty workspace `7` on monitor 1 (while workspace `4` with a browser is currently open on monitor 1) and I want to "pass" this empty workspace `7` to monitor 2 (with the mouse focused on monitor 2 in workspace `9`) via a keybind something like `bind = $mainMod CTRL, 7, focusworkspaceoncurrentmonitor, 7`, the actual behavior now is to focus and move the mouse to monitor 1 with the empty workspace `7`, and not to open that empty workspace `7` on monitor 2! And when I have the empty workspace `7` open on monitor 1 and workspace `9` open on monitor 2 (with the mouse focus on it), the actual behavior is to SWAP these two workspaces between monitors, and that's absolutely not what I want: I want to move the empty workspace `7` to monitor 2 and focus on it and leave workspace `9` on monitor 2 as well, but that's not what currently happens. 

Does anyone know how to fix this, or is there some setting I missed or broke in settings? Can I ask to fix this bug or implement a feature to my preferred behavior if it's not a bug tho? Thanks.

ClaireNeveu:
Thank you for doing this! Just getting into hyprland now and the number xmonad feature that never seems to get implemented is this.

Issue #4388, opened 2024-01-08
Floating windows do not retain their size when toggling fullscreen on/off


bvr-yr:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit d7d333d162da2d3fc852b2c7a3faa2709440cefa dirty (opengl: apply box rot to projections).
Date: Sun Jan 7 20:51:08 2024
Tag: v0.34.0-31-gd7d333d1

flags: (if any)


System Information:
System name: Linux
Node name: Arch-ASUS
Release: 6.6.10-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Fri, 05 Jan 2024 16:20:41 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation 4th Gen Core Processor Integrated Graphics Controller [8086:0416] (rev 06) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Bug or Regression?

Regression

### Description

this 813af393f1b61ecd0105e126681986e58e73ffee is a badboy

### How to reproduce

default `hyprland.conf`

- `togglefloating` / exec with [floating] rule
- dispatch fullscreen 0/1 (both affects) or app's native FS mode
- undo fullscreen

### Actual
window is floated and takes whole screen in size

### Expected
window keeps its original size before fullscreen

PR #4443, opened 2024-01-14
floating windows lose last size while fullscreen

ettancos:
As described in #4388 there is a regression where exiting fullscreen leaves the window size as maximized for floating windows, it happens because recalculateMonitors (called in updateDynamicRules) updates window.realSize.
Moving the update rules above the fullscreen properties will have recalculateMonitors skip the fullscreen ceremony.

Just a proposal to get it work again, we can also discuss better options.

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -776,12 +776,12 @@ void CHyprDwindleLayout::fullscreenRequestForWindow(CWindow* pWindow, eFullscree
         return;
     }
 
+    pWindow->updateDynamicRules();
+    pWindow->updateWindowDecos();
+
     // otherwise, accept it.
     pWindow->m_bIsFullscreen           = on;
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
we cant do that. It's in place to match rules that want a fullscreen window, and you update them before setting the fs state

bvr-yr:
so what can we do? kinda annoying thing, gotta use `pseudo` to workaround atm

bvr-yr:
@vaxerski can we move block with saving float props like this: https://github.com/bvr-yr/Hyprland/commit/3551cb058275ff9ddbd7704630bcfb13efb340cb ?

vaxerski:
that would most likely be the way to fix this. Though in your commit you kinda pasted it between a comment and the code it was referencing xd

bvr-yr:
that was intended cause i thought _accepting_ comment is about whole part after since we are not returning anymore :D. changed

should i open PR or wait on OP?

vaxerski:
you can make a MR and mention it supersedes this one

PR #4444, opened 2024-01-14
groupbar: separate gradients from title

MightyPlaza:
Separates gradient from titles logic
Adds `group:groupbar:height` to set bar height that was previously set using the font size

https://github.com/hyprwm/hyprland-wiki/pull/439
didn't find any issues


vaxerski (approved):
lgtm

PR #4486, opened 2024-01-20
fix: movewindoworgroup onto empy workspace on next monitor

muellerbernd:

I had the problem when using `movewindoworgroup`, that if I want to move the current window(that is not inside a group) onto an empty workspace on another monitor, the `movewindoworgroup` did not behave like `movewindow` and I was not able to move the window. With `movewindow` everything worked fine.

I tested the code on my NixOS machine and it worked like I expected. Please let me know if there is an easier fix  for that.


src/managers/KeybindManager.cpp
```diff
@@ -2203,7 +2203,12 @@ void CKeybindManager::moveWindowOrGroup(std::string args) {
         } else
             moveWindowOutOfGroup(PWINDOW, args);
     } else if ((*PIGNOREGROUPLOCK || !ISWINDOWGROUPLOCKED) && ISWINDOWGROUP) // no target window
+    {
```
vaxerski, line 2206: this brace should go before the comment on top.

muellerbernd: added the changes mentioned in your review

vaxerski (approved):
lgtm

PR #4489, opened 2024-01-20
Touchscreen workspace swipe

earboxer:

Adds support for workspace-swiping via touchscreen

It repurposes existing workspace swiping code, and so it supports most of the the ["Gestures" configuration parameters](https://wiki.hyprland.org/Configuring/Variables/#gestures):
* instead of `workspace_swipe`, it uses `workspace_swipe_touch` to enable/disable it.
* `workspace_swipe_fingers` not relevant
* `workspace_swipe_distance` is unused. Or rather, it is *ab*-used to map the touchscreen's width to [0,100%]
  * Probably should refactor that 'abuse' out before merging this.
* `workspace_swipe_forever` has no effect, because the touchscreen's width is mapped to 0-100x, so you can't swipe more than one at a time.
  * I imagine it could be useful to change multiple workspaces at a time, though I like how it feels to drag the workspace over and have it follow your thumb. Maybe we need some `workspace_swipe_touch_distance_multiplier` parameter to allow this? (but default to 1, which would be the preferred value)
      * This could be done in the future if someone wants it.
* `workspace_swipe_invert` is implemented (though it feels weird to use)
* `workspace_swipe_direction_lock` has no effect, because the direction is set based on which edge you grabbed

~~No, this is not ready for merging.~~  Done some work. I think now it could be fine to merge.
Work needed:
- [x] ~~Currently the margins are hard-coded (at 5%)~~. This needs to 
  - [x] Be based on gaps_out
  - [x] Account for a layer shell that might be in the way
  - [x] handle vertical workspace changing as well
- ~~probably need to test with a multimonitor setup (I imagine something needs to be done so we switch the right monitor)~~ I think this will work fine.
- [ ] Support no_gaps_when_only??? (maybe a nice enhancement)
    - ~~support fullscreened windows~~? (I don't use fullscreen on mobile, doing so one can get trapped, so actually a workspace swipe could help save from that)
    - [ ] maybe do something generic so that floating windows will obstruct this? (not sure how to detect that I touched a floating window).
- [X] hyprland-wiki PR to add documentation for this feature: https://github.com/hyprwm/hyprland-wiki/pull/448

### DIFF
--- a/src/managers/input/Touch.cpp
+++ b/src/managers/input/Touch.cpp
@@ -1,7 +1,15 @@
 #include "InputManager.hpp"
 #include "../../Compositor.hpp"
+#include "../../config/ConfigValue.hpp"
 
 void CInputManager::onTouchDown(wlr_touch_down_event* e) {
+    static auto PSWIPETOUCH  = CConfigValue<Hyprlang::INT>("gestures:workspace_swipe_touch");
+    static auto PGAPSOUTDATA = CConfigValue<Hyprlang::CUSTOMTYPE>("general:gaps_out");
+    auto* const PGAPSOUT     = (CCssGapData*)(PGAPSOUTDATA.ptr())->getData();
+    // TODO: WORKSPACERULE.gapsOut.value_or()
+    auto        gapsOut     = *PGAPSOUT;
+    static auto PBORDERSIZE = CConfigValue<Hyprlang::INT>("general:border_size");
+    static auto PSWIPEINVR  = CConfigValue<Hyprlang::INT>("gestures:workspace_swipe_invert");
     EMIT_HOOK_EVENT_CANCELLABLE("touchDown", e);
 
     auto       PMONITOR = g_pCompositor->getMonitorFromName(e->touch->output_name ? e->touch->output_name : "");
@@ -22,6 +30,30 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
         e.state = WL_POINTER_BUTTON_STATE_PRESSED;
         g_pInputManager->processMouseDownKill(&e);
         return;
+    }
+
+    // Don't propagate new touches when a workspace swipe is in progress.
+    if (m_sActiveSwipe.pWorkspaceBegin) {
+        return;
+        // TODO: Don't swipe if you touched a floating window.
+    } else if (*PSWIPETOUCH && (!m_pFoundLSToFocus || m_pFoundLSToFocus->layer <= ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM)) {
+        const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(PMONITOR->activeWorkspace);
+        const bool VERTANIMS  = PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle == "slidevert" ||
+            PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle.starts_with("slidefadevert");
+        // TODO: support no_gaps_when_only?
+        const double TARGETLEFT  = ((VERTANIMS ? gapsOut.top : gapsOut.left) + *PBORDERSIZE) / (VERTANIMS ? PMONITOR->vecSize.y : PMONITOR->vecSize.x);
+        const double TARGETRIGHT = 1 - (((VERTANIMS ? gapsOut.bottom : gapsOut.right) + *PBORDERSIZE) / (VERTANIMS ? PMONITOR->vecSize.y : PMONITOR->vecSize.x));
+        const double POSITION    = (VERTANIMS ? e->y : e->x);
+        if (POSITION < TARGETLEFT || POSITION > TARGETRIGHT) {
+            beginWorkspaceSwipe();
+            m_sActiveSwipe.touch_id = e->touch_id;
+            // Set the initial direction based on which edge you started from
+            if (POSITION > 0.5)
+                m_sActiveSwipe.initialDirection = *PSWIPEINVR ? -1 : 1;
+            else
+                m_sActiveSwipe.initialDirection = *PSWIPEINVR ? 1 : -1;
+            return;
+        }
     }
 
     m_bLastInputTouch = true;
@@ -55,6 +87,13 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
 
 void CInputManager::onTouchUp(wlr_touch_up_event* e) {
     EMIT_HOOK_EVENT_CANCELLABLE("touchUp", e);
+    if (m_sActiveSwipe.pWorkspaceBegin) {
+        // If there was a swipe from this finger, end it.
+        if (e->touch_id == m_sActiveSwipe.touch_id)
+            endWorkspaceSwipe();
+        return;
+    }
+
     if (m_sTouchData.touchFocusSurface) {
         wlr_seat_touch_notify_up(g_pCompositor->m_sSeat.seat, e->time_msec, e->touch_id);
     }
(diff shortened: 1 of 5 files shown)


src/managers/input/Touch.cpp
```diff
@@ -24,6 +28,25 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
...
+        const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(PMONITOR->activeWorkspace);
+        const bool VERTANIMS  = PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle == "slidevert" ||
+            PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle.starts_with("slidefadevert");
+        // TODO: support no_gaps_when_only?
+        const auto target_portion = (*PGAPSOUT + *PBORDERSIZE) / (VERTANIMS ? PMONITOR->vecSize.y : PMONITOR->vecSize.x);
+        const auto position       = (VERTANIMS ? e->y : e->x);
+        if (position < target_portion || position > (1 - target_portion)) {
+            beginWorkspaceSwipe();
+            // Set the initial direction based on which edge you started from
+            if (position > 0.5) {
```
vaxerski, line 41: style nit: no `{}` around short ifs

src/managers/input/Touch.cpp
```diff
@@ -24,6 +28,25 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
         return;
     }
 
+    if (*PSWIPETOUCH && (!m_pFoundLSToFocus || m_pFoundLSToFocus->layer == ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND)) {
```
vaxerski, line 31: <= ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM

src/managers/input/Touch.cpp
```diff
@@ -24,6 +28,25 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
...
+        const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(PMONITOR->activeWorkspace);
+        const bool VERTANIMS  = PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle == "slidevert" ||
+            PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle.starts_with("slidefadevert");
+        // TODO: support no_gaps_when_only?
+        const auto target_portion = (*PGAPSOUT + *PBORDERSIZE) / (VERTANIMS ? PMONITOR->vecSize.y : PMONITOR->vecSize.x);
```
vaxerski, line 36: no snake_case, we use camelCase

additionally, const values we SHOUT

so `TARGETPORTION` and `POSITION`

vaxerski:
review will be stalled until #4656 as it will cause tons of conflicts

earboxer:
(marked as a draft because I must have added some bugs when rebasing)

src/managers/input/Touch.cpp
```diff
@@ -24,6 +28,26 @@ void CInputManager::onTouchDown(wlr_touch_down_event* e) {
...
 
+    // TODO: Don't swipe if you touched a floating window.
+    if (**PSWIPETOUCH && (!m_pFoundLSToFocus || m_pFoundLSToFocus->layer <= ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM) && (!m_sActiveSwipe.pWorkspaceBegin)) {
+        const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(PMONITOR->activeWorkspace);
+        const bool VERTANIMS  = PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle == "slidevert" ||
+            PWORKSPACE->m_vRenderOffset.getConfig()->pValues->internalStyle.starts_with("slidefadevert");
+        // TODO: support no_gaps_when_only?
+        const auto TARGETPORTION = (**PGAPSOUT + **PBORDERSIZE) / (VERTANIMS ? PMONITOR->vecSize.y : PMONITOR->vecSize.x);
+        const auto POSITION      = (VERTANIMS ? e->y : e->x);
+        if (POSITION < TARGETPORTION || POSITION > (1 - TARGETPORTION)) {
```
earboxer, line 39: This logic is broken (since the rebase) -- seems to always return true

earboxer: Doing some printf debugging:
```
num: 187650383698822 / 404.000000TARGETPORTION: 464481147769.361389
POSITION: 0.650000
```

The problem is that `(**PGAPSOUT + **PBORDERSIZE)` returns *reallybignumber*

earboxer: I guess PGAPSOUT is a special data type instead of just an integer

vaxerski:
would be nice to configure the number of fingers.

earboxer:
> would be nice to configure the number of fingers.

as-is the gesture tracks a single-finger (the one which started the gesture by grabbing the edge). It's very useful on a slate/phablet mobile phone device (where I mostly use my thumb to do everything and using a second finger is an inconvenience).

I can see the use-case for having a multifinger touch swipe gesture (e.g. on a tablet fixed on a stand), but that doesn't share too much in common with the edge-based implementation that this PR represents; I don't think tacking on multifinger to this edge-based gesture would be worth the effort.

earboxer:
Here's a [very] quick example of this:

https://github.com/hyprwm/Hyprland/assets/12452738/d1202f6a-e798-4160-ab53-1ab6ecca4c6a

vaxerski:
sorry I forgot about this

I am thinking more about 2-in-1s or laptops with a touchscreen (like mine)

How would one otherwise activate and deactivate this? if they wanna use their screen and not swipe

earboxer:
Just don't touch the edge of the screen.
The idea is that you're grabbing the gaps (where there's no content) and moving them.

vaxerski:
aaah, okay

makes sense.

vaxerski (approved):
lgtm

leon-erd:
Wouldn't it be a good idea to have a separate `workspace_swipe_invert` for the touchscreen workspace swipe gesture? Bc on my 2-in-1 laptop I would love to use both touchscreen and touchpad for workspace swipes. For touchpad `workspace_swipe_invert=false` feels weird and for touchscreen `workspace_swipe_invert=true` feels weird...

earboxer:
On Mon Mar 18, 2024 at 10:13 AM EDT, Leon wrote:
> Wouldn't it be a good idea to have a separate `workspace_swipe_invert` for the touchscreen workspace swipe gesture? Bc on my 2-in-1 laptop I would love to use both touchscreen and touchpad for workspace swipes.

Yes, that makes sense. Feel free to make a PR and I'll review if you tag me in it.
I don't have a 2-in-1, which is probably why I didn't think of that.

leon-erd:
I thought this is probably really easy to implement and was just forgotten (as you said) that's why I commented. Unfortunately, I don't know cpp so I don't want to tinker around with that

earboxer:
@leon-erd I would argue that you don't need to know cpp to make that change. you just need to know how to
1. Add a new config value
2. Change which config value is read during the touch gesture

If you're interested, you can learn 1 by looking at https://github.com/hyprwm/Hyprland/pull/4489/files#diff-9993bc2bbc311ab9e2cb878dbc44bc4f9d5edc5216b9aa3172e4b10b9b0d61f5 (and the wiki PR for documentation)
and 2 by looking at https://github.com/hyprwm/Hyprland/pull/4489/files#diff-22374f0f24e1efb2102fb42abd58b3bae860444eb5b9cb60015a7608333fc53b (Touch.cpp containts the touch-specific code, while Swipe.cpp contains the code that's common to both)

Of course, being able to compile hyprland would be a good too...

Issue #3093, opened 2023-08-27
fullscreen windows on same workspace issues


kristoferus75:
### Description

Hi !

Issue with fullscreen windows (rule maximized or fullscreen 1 windows)

two or more fullscreen windows on same workspace:

**issues**

- if close one window on this workspace -> windows on ws become tiled -> **is not ok**
- if i open an window on other monitor -> windows on ws become tiled -> **is not ok** 
- if i move one window to other monitor -> windows on ws become tiled -> **is not ok** 
- if i open an window on same  which have float rule   -> windows on ws become tiled  -> **is not ok** 
- if i open an window on ws which have no window rule -> windows on ws become tiled -> **is ok**
- if i open an window on ws which have maximized rule -> windows on ws stay maximized -> **is ok**

Is it possible to fix this issues or enhance hyprland ? (let maximized windows stay maximized)

Thanks

vaxerski:
you can script this. Make a script check on window close whether it was maximized and if so, maximize the first window on that workspace.

kristoferus75:
ok but sorry this is only a workaround for close and the script must run all the time in background :-(

vaxerski:
same method can be applied for open events.

misuzu:
Likely related https://github.com/hyprwm/Hyprland/issues/2050

kristoferus75:
maximize the first window on that workspace.  

How i get this address with hyprctl ??

```
addr_focused="$(hyprctl -j activewindow | gojq -r '.address')"
active_win_fullscreen="$(hyprctl -j activewindow | gojq -r '.fullscreen')

if [[ "$active_win_fullscreen" == "true" ]]; then
  hyprctl --batch "dispatch focuswindow address:${addr_focused=}; dispatch fullscreen 1;"
fi
```

kristoferus75:
script does close windows but not fullscreen -> What is wrong ? 

```
#!/usr/bin/env bash

addr_focused="$(hyprctl -j activewindow | gojq -r '.address')"
active_win_fullscreen="$(hyprctl -j activewindow | gojq -r '.fullscreen')"

active_ws_clients="$(hyprctl clients -j | jq -r ".[] | select(.workspace.id == "$(hyprctl activewindow -j | jq -r '.workspace.id')\)"")"
first_address="$(echo $active_ws_clients | jq -r '.address' | head -1)"

if [[ "$active_win_fullscreen" == "true" ]]; then
[...]

PR #4490, opened 2024-01-21
Restore full screen while moving or closing a fullscreen window

jsecchiero:

currently windows go into tile mode after a window closes or moves to another workspace (https://github.com/hyprwm/Hyprland/issues/3093). In this PR the first window will be put in fullscreen mode keeping a consistent behavior while working on fullscreen windows

minor fix: I did add a few configurations to let Nix add symbols to the generated binary otherwise was not possible to debug

Tested on my dual monitor setup and with the single window in a nested environment and everything is working as expected

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -96,12 +96,15 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
       then "debug"
       else "release";
 
+    dontStrip = debug;
+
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
wouldnt it be better to fix the root cause, window stopping being fullscreen in the first place? This can switch which window is fullscreen, no?

vaxerski:
closing due to inactivity

PR #4514, opened 2024-01-23
focus window on mouse down on decoration

MightyPlaza:
focus window on mouse down on decoration
shouldn't cause issues


vaxerski (commented):
breaks follows mouse 3

MightyPlaza:
> breaks follows mouse 3

fixed

src/managers/input/InputManager.cpp
```diff
@@ -601,8 +601,12 @@ void CInputManager::processMouseDownNormal(wlr_pointer_button_event* e) {
     const auto mouseCoords = g_pInputManager->getMouseCoordsInternal();
     const auto w           = g_pCompositor->vectorToWindowIdeal(mouseCoords);
 
-    if (w && !m_bLastFocusOnLS && w->checkInputOnDecos(INPUT_TYPE_BUTTON, mouseCoords, e))
-        return;
+    if (w && !m_bLastFocusOnLS) {
```
vaxerski, line 604: w is ideal, but does ideal include decos? I don't think so... so we would get the window behind, no?

MightyPlaza: `CCompositor::vectorToWindowIdeal()` uses `CWindow::getWindowInputBox()` which does include, which does include all extents (not just reserved)
so we do get the proper window, which is the same that is checked for input on decos with the same logic

on another note, `CWindow::getWindowInputBox()` includes non reserved extents 
this makes it so clicking on shadows (for example) focuses the window
is this desired or should it include only the outer reserved extents?

vaxerski: ouh, I thought ideal uses vecposition and vecsize. Doh.

Ideal should be vecpos and vecsize. That's it.

MightyPlaza: should I make a clone of `CCompositor::vectorToWindowIdeal()` and use that here, so `CCompositor::vectorToWindowIdeal()` gets changed in the future
or should we try to unify the existing get window functions and adding extra parameters to get the same functionality?

vaxerski: we could unify them. Ideal should only use vecpos vecsize (it's for determining windows' "input boxes")

MightyPlaza: created 2 new functions with the purpose of achieving all possible queries
at the moment `CCompositor::vectorToWindowUnified` is just `CCompositor::vectorToWindowIdeal` with some extra logic to exclude/restrict to float and uses `CWindow::getWindowBoxUnified` instead of `CWindow::getWindowInputBox`

src/Window.hpp
```diff
@@ -30,6 +30,14 @@ enum eGroupRules {
     GROUP_OVERRIDE    = 1 << 6, // Override other rules
 };
 
+enum eGetWindowProperties {
+    WINDOW_ONLY      = 0, // 0, 1, 2 are mutually exclusive
+    RESERVED_EXTENTS = 1,
```
vaxerski, line 35: with bitfields please make all fields except 0 a `(1 << x)`

src/Window.hpp
```diff
@@ -342,6 +350,7 @@ class CWindow {
     // methods
     CBox                     getFullWindowBoundingBox();
     SWindowDecorationExtents getFullWindowExtents();
+    CBox                     getWindowBoxUnified(uint8_t props);
```
vaxerski, line 353: no reason for a uint8_t, uint64_t makes more sense.

vaxerski (commented):
if unified exists, shouldn't we remove `tiled` and `ideal`?

src/Window.hpp
MightyPlaza: since the beginning of the enum is mutually exclusive it doesn't make sense combining them
so we can use numbers that can't be represented as `1 << n`, for example 3

since they are less than `1 << 2` (1st element) they don't interfere with it

should I still use only `1 << n` for the enum's items?

MightyPlaza:
> if unified exists, shouldn't we remove `tiled` and `ideal`?

that is the end goal (as well as the `CWindow::get...Box()` functions)
I haven't tested it properly yet, so it isn't in this PR
but should just be a drop-in replacement

src/Window.hpp
vaxerski: it's a stylistic choice, for consistency

vaxerski:
we can do it here. No point in splitting.

MightyPlaza:
seems to work fine
used `FULL_EXTENTS | ALLOW_FLOATING` as args in most cases since that is closest to the behavior of the original function, and doesn't really matter for tiled windows since they use the node size/pos

in future more enum elements should probably be added (ALLOW_SPECIAL for example)

vaxerski (commented):
wherever `Ideal` is used, it should not count non-interactable decos, so `FULL_EXTENTS` is wrong.

The fact it used to was a bug.

MightyPlaza:
> wherever `Ideal` is used, it should not count non-interactable decos, so `FULL_EXTENTS` is wrong.
> 
> The fact it used to was a bug.

speaking of this

I was trying to implement border resize using the decoration input system
the issue is the extension of the resize area
I was thinking of allowing non reserved input area, which would be checked after nothing is found on the reserved area

is there any reason to want to get window using the actual full extents
or does it only make sense for extents (not necessarily reserved) that allow input?

vaxerski:
> is there any reason to want to get window using the actual full extents

rendering

Ideal -> input
Full -> rendering

> allowing non reserved input area

elaborate? isn't this just deco with the input flag?

MightyPlaza:
> elaborate? isn't this just deco with the input flag?

should just be, looks like most of the work is already done then

what should I used to replace the original function? (`WINDOW_ONLY`, `RESERVED_EXTENTS` or`INPUT_EXTENTS`)

vaxerski:
which original?

Ideal probably INPUT_EXTENTS and RESERVED_EXTENTS

Full probably ALL_EXTENTS

MightyPlaza:
> which original?
> 
> Ideal probably INPUT_EXTENTS and RESERVED_EXTENTS
> 
> Full probably ALL_EXTENTS
> 
> edit: you should also add a flag called `TILED_IDEAL` that will use pure ideal for tiled windows (m_vPosition and m_vSize)

done

added `USE_PROP_TILED` to use the window properties from the enum to get the window box

src/SharedDefs.hpp
```diff
@@ -52,4 +52,9 @@ struct SWindowDecorationExtents {
     bool operator==(const SWindowDecorationExtents& other) const {
         return topLeft == other.topLeft && bottomRight == other.bottomRight;
     }
-};
\ No newline at end of file
+
+    void addExtents(const SWindowDecorationExtents& other) {
+        this->topLeft     = this->topLeft.getComponentMax(other.topLeft);
```
vaxerski, line 57: we don't use `this` where unnecessary

vaxerski (commented):
lgtm except the two style nits

src/SharedDefs.hpp
MightyPlaza: done

vaxerski (approved):
lgtm

PR #4515, opened 2024-01-23
CoC: remove redundant verbage and simplify

binex-dsk:
This simplifies the Code of Conduct by removing language that is redundant, too verbose, or otherwise unnecessary for the reader to understand the point of the Code of Conduct.


vaxerski:
no

PR #4517, opened 2024-01-23
Allow disabling touchscreen input

earboxer:

Adds support for `input:touchdevice:enabled`/`device:$device:enabled` for touch devices.

Use case: I don't want to type/click on stuff when I turned the screen off.
```
# toggle touchscreen off/on with power button
bind = , XF86PowerOff, exec, hyprctl -i 0 dispatch dpms $(hyprctl -j -i 0 monitors | jq -r 'first|.dpmsStatus|if . then "off" else "on" end')
bind = , XF86PowerOff, exec, hyprctl -i 0 keyword input:touchdevice:enabled $(hyprctl -j getoption input:touchdevice:enabled | jq '.int==0')
```

I wonder why we don't have a 'generic' way of doing this, since the libinput device config code is generic.

Yes, should be good to go!


vaxerski (approved):
lgtm, wiki mr needed

PR #4526, opened 2024-01-24
fix random hang on exit

killown:
fix random hang on exit

still need testing, I did some hyprctl dispatch exist and still no random hang on exit


killown:
takes more time to catch the issue, exit(0) doesn't solve 100%, thus closed!

Issue #2661, opened 2023-07-06
Can't interact with certain steam windows


Roger-Roger-debug:
Since the new steam UI rolled out I can't interact with some windows, mainly the context menu opened by right-clicking someone my friendlist.

# Steps to reproduce
- Open steam
- Open friendlist
- Right click someone on friendlist

# Expected outcome
I can interact with the opened menu

# Actual outcome
Menu can't be interacted with

----
I ran a git bisect and the offending commit seems to be 
https://github.com/hyprwm/Hyprland/commit/fcf1bafb098c8f756550195269dcd95202124277

Had some trouble running the bisect as after the `git bisect skip` the menus wouldn't always show up or crash steam.

<details><summary>Bisect log</summary><p>

```
# bad: [51a930f802c71a0e67f05e7b176ded74e8e95f87] version: bump to 0.26.0
# good: [c02ac5e08a425d5ec44b48e24ca2695d9cafd38d] fix un-normalized missed color
git bisect start 'v0.26.0' 'v0.20.0beta'
# bad: [06244555915339967864292dd0b83cd9732516d8] Meson: add rdynamic ld flag
git bisect bad 06244555915339967864292dd0b83cd9732516d8
# bad: [38c25bb50d76409193db2438db60369f3868edf9] Added a crash reporter
git bisect bad 38c25bb50d76409193db2438db60369f3868edf9
# good: [12e293e3092b03dbb9d82e7a749f482ef2b7ea6e] update animated deco values after setprop
git bisect good 12e293e3092b03dbb9d82e7a749f482ef2b7ea6e
# bad: [be6325dd4b0a99561281ba557f1d37e945fa183f] ext_workspace_unstable: send done after output_enter when wl_output is bound late (#1481)
git bisect bad be6325dd4b0a99561281ba557f1d37e945fa183f
# skip: [5d35c0432b70e2d5e82296caf7f496dc4612f3e5] add a minimize ipc event
git bisect skip 5d35c0432b70e2d5e82296caf7f496dc4612f3e5
# good: [32381fe6c4e33232401d7a74f587ee7296a61fcf] send a 1 on unknown minimize event
git bisect good 32381fe6c4e33232401d7a74f587ee7296a61fcf
# good: [a46abd8b1adaf850e243380ea2a7d6b02677fc86] update wlroots dep
git bisect good a46abd8b1adaf850e243380ea2a7d6b02677fc86
[...]

PR #4539, opened 2024-01-26
remove m_vReportedSize set for xwayland

phonetic112:
Fixes https://github.com/hyprwm/Hyprland/issues/2661#issuecomment-1872569582

nope

ready

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -1137,8 +1137,7 @@ void Events::listener_unmanagedSetGeometry(void* owner, void* data) {
         g_pHyprRenderer->damageWindow(PWINDOW);
 
         PWINDOW->m_vReportedPosition    = PWINDOW->m_vRealPosition.goalv();
-        PWINDOW->m_vReportedSize        = PWINDOW->m_vRealSize.goalv();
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
but... unmanagedSetGeometry is a client request? Why would we wait for a pong?

phonetic112:
Steam pop-ups open themselves at size 0,0 for some reason before they request `unmanagedSetGeometry`, so setting `m_vReportedSize` like that would cause the surface to prematurely render at size 0,0 if the timing is off.

so setting `m_vPendingReportedSize` directly lets us actually wait for the pong before rendering the surface.

vaxerski:
okay but that's _before_ and we apply `.goalv()` which should have the value requested in the event struct?

phonetic112:
yes, with the `minsize 1 1` window rule it requests size (5, 5) first, and then the correct size afterwards (something like (100, 150))

vaxerski:
why would this fix it then?

phonetic112:
The surface would have no size until the pong is received

vaxerski (approved):
I guess if it fixes the issue, we can try that. Though I have bad feelings about this.

PR #4555, opened 2024-01-28
nix: fix overlay composition

GrizzlT:

The nix overlay defined in `nix/overlays.nix` defines a function called `mkJoinedOverlays` to combine the correct dependencies for hyprland. This function however only *combines* the result of multiple overlays but does not provide the dependency overlay outputs to `callPackage` as an input when defining the `hyprland` package. This means that the internal dependency `wlroots-hyprland` must be defined on the final nixpkgs fixpoint.

To avoid name clashes, I personally shadow my flake input's overlays to provide a cleaner package set when selecting packages (i.e. hide internal dependencies). Because the hyprland overlay depends on the final package set to have a root `wlroots-hyprland` package, this prohibits my workflow. This PR fixes that. 

This fix should be fully backwards compatible. I have tested that the overlay output package is a derivation. I have not tested the theoretical possibility of still manually overriding `wlroots-hyprland` using `hyprland.override { ... }`. This should however work since the hyprland package is a standard `mkDerivation` result.

ready, do I need formatting?

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -10,19 +10,16 @@
     (builtins.substring 4 2 longDate)
     (builtins.substring 6 2 longDate)
   ]);
-
-  mkJoinedOverlays = overlays: final: prev:
-    lib.foldl' (attrs: overlay: attrs // (overlay final prev)) {} overlays;
 in {
   # Contains what a user is most likely to care about:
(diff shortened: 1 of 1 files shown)


fufexan (approved):
LGTM. Thanks!

PR #4572, opened 2024-01-31
subprojects: remove wlroots.wrap

fufexan:

From @jbeich's testing, the wrap file doesn't work as it should, and having it
is not consistent with the other subprojects. See
https://github.com/hyprwm/hyprland-wiki/pull/200#issuecomment-1805708563 for
more info.

The wiki command `meson subprojects update --reset` won't work anymore (if it
even worked in the first place). It will be removed in the PR linked above.

Might be ready. @jbeich thoughts?


jbeich:
Looks fine. With d179501c4f4a there's no need for `soversion` hack while plugin API uses wlroots headers but doesn't need the library. I'm not sure if any plugins end up using wlroots symbols (only tested hy3) but even then Hyprland binary should provide wlroots symbols (on dlopen) due to static linking.

fufexan:
Thanks for the input, merging then.

PR #4577, opened 2024-01-31
Fix nullptr crash when disconnecting a monitor

abmantis:

I was able to reproduce the crash frequently by having a kitty terminal on an monitor running the following command and then unplugging that monitor:

`while true; do echo "" && sleep 0.02; done`

No.

I think it is ready, unless there is a better way to handle this race condition.


vaxerski (approved):
alr

PR #4585, opened 2024-02-01
hyprpm: install headers locally

vaxerski:
cc @zakk4223 @DreamMaoMao @fufexan 

hyprpm will now install headers locally to `$XDG_DATA_HOME/hyprpm/headersRoot`

Should completely eliminate duplicate headers / system headers issues.

`hyprpm` will now refuse to use system headers altogether:

https://github.com/hyprwm/Hyprland/blob/d0bfa567a89e8023898413a057e50dbb93b73bc5/hyprpm/src/core/PluginManager.cpp#L285-L290

They are unreliable all things considered and to be frank who cares.

Bonus: no more asking for password to install headers.

Needs some testing, seems to pass my short tests.

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -10,6 +10,7 @@
 #include <thread>
 #include <fstream>
 #include <algorithm>
+#include <format>
 
 #include <toml++/toml.hpp>
 
@@ -77,6 +78,8 @@ SHyprlandVersion CPluginManager::getHyprlandVersion() {
 }
 
 bool CPluginManager::addNewPluginRepo(const std::string& url) {
+
+    const auto HLVER = getHyprlandVersion();
 
     if (DataState::pluginRepoExists(url)) {
         std::cerr << "\n" << Colors::RED << "✖" << Colors::RESET << " Could not clone the plugin repository. Repository already installed.\n";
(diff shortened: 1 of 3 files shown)


vaxerski:
I like how i just realized I hardcoded the path with my user name lol

hyprpm/src/core/PluginManager.cpp
```diff
@@ -382,14 +409,19 @@ bool CPluginManager::updateHeaders(bool force) {
     progress.m_szCurrentMessage = "Installing sources";
     progress.print();
 
-    progress.printMessageAbove(
-        std::string{Colors::YELLOW} + "!" + Colors::RESET +
-        " in order to install the sources, you will need to input your password.\n  If nothing pops up, make sure you have polkit and an authentication daemon running.");
+    // progress.printMessageAbove(
+    //     std::string{Colors::YELLOW} + "!" + Colors::RESET +
+    //     " in order to install the sources, you will need to input your password.\n  If nothing pops up, make sure you have polkit and an authentication daemon running.");
```
fufexan, line 414: Is this still needed here?

fufexan (approved):
LGTM

hyprpm/src/core/PluginManager.cpp
vaxerski: no, not really

fufexan:
@jbeich would this change allow us to `--skip-subprojects` in https://github.com/hyprwm/hyprland-wiki/pull/200?

zakk4223:
Ok, finally got to test this after wasting a bunch of time discovering makepkg clobbers branch names :/

All seems good to me. Normal and forced update both worked without header conflicts.

When plugins don't load due to mismatch sometimes the notification gets hidden by the big red config syntax error banner, but that's not an issue with this PR; I'll (eventually) create another issue for that one.

vaxerski:
top. Merging then. Thanks bros

PR #4609, opened 2024-02-04
Command specifc usage messages on hyprctl

jones-josh:
It'd be nice to have usage information on commands in hyprctl, especially dispatch. Most times I try to use it from memory I make mistakes with its usage and end up keeping the wiki open. The commit in this PR is just a first start/draft, and I'm aiming to open the discussion on how you would feel it would be best included. Perhaps this is too much to keep in the executable and we could instead extend / add a further man page, or perhaps you think it's not worth the extra effort maintaining the new offline usage docs as well as the wiki. I'd be very interested to hear which you think would be more appropriate and would happily start work on it.

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -62,6 +62,76 @@ flags:
     --instance (-i) -> use a specific instance. Can be either signature or index in hyprctl instances (0, 1, etc)
 )#";
 
+const std::string DISPATCH_USAGE = R"#(usage: hyprctl [(opt)flags] dispatch [dispatcher] [(opt)args]
+
+dispatchers:
+    exec                            executes a shell command
+    execr                           executes a raw shell command (does not support rules)
+    pass                            passes the key (with mods) to a specified window. Can be used as a workaround
+                                    to global keybinds not working on Wayland.
+    killactive                      closes (not kills) the active window
+    closewindow                     closes a specified window
+    workspace                       changes the workspace
+    movetoworkspace                 moves the focused window to a workspace
+    movetoworkspacesilent           same as above, but doesnt switch to the workspace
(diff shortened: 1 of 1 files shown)


vaxerski:
this will get outdated really quickly. Why not redirect to the wiki?

jones-josh:
Offline documentation is preferable where possible as it does not require an internet connection. After giving it some further thought though I don't like what I've committed here and am now much more strongly on the side of expanding the man page(s). The first tool I reach for when using a shell program I'm not 100% familiar with is the program itself (`-h`), but very shortly after that I look it up on man long before I open my browser. Hyprctl already does have a man page, but it is far from usable as a standalone document and just provides some general hints. Having it populated with much more information with a disclaimer that it may not be 100% up to date, and to fallback to the wiki as truth in those cases, would be much more preferable to me and I'm sure many other consumers of the project.

In the case of dispatchers as an example, the dispatchers and parameter tables are the main bits of information I'd like to be able to more quickly reference. A script generating a man page out of the markdown in the wiki wouldn't be difficult, but I'm not confident enough in deployment to comfortably assert that it wouldn't be difficult to actually put in place.

Final say is yours, but is definitely something I personally would like to use.

stephentoth:
my two cents: having to open a web browser, navigate to the wiki, then find where the thing I need is located gets old fast.

In terms of it being outdated, it might (since it makes you have to update in two locations), but the top level commands are already in the help so you already have to do that.

also, the error messages for the dispatch command are less than helpful at the current moment. I'd be nice to just have the help as part of the error.

fufexan:
I will be working on updating the manpages soon.

jones-josh:
Closing this issue in that case. @fufexan feel free to tag me here when you start if you want to split the work.

PR #4643, opened 2024-02-08
renderer: Add dimaround layer rule

### LINKED ISSUE #1657: Dimaround rule for layers
Hi,

I've been using hyprland for a week know, and everything has been working fine. I absolutely love this WM.

On the [wiki page about rules](https://wiki.hyprland.org/Configuring/Window-Rules/#rules) it shows the `dimaround` windowrule. Also, at the end of that page, it talks about what layers are, and it seems that they only have the `noanim` rule. I have a custom rofi powermenu script for login out, reboot, etc. If I didn't understand uncorrectly, `rofi` would be a layer, and it seems to be since I can see it in `hyprctl layers` as `namespace: rofi`, and not in `hyprctl clients`.

It would be very much appreciated if we could also dim around a powermenu like that.

In that case, I think right now it wouldn't be possible to differentiate between more than one type, since `hyprctl layers` doesn't seem to report any difference between app launcher / powermenu script.

Anyway, thanks for all the work. I've finally moved from X11 + bspwm to wayland thanks to this project!

### LINKED ISSUE #1889: Add 'dimaround' layerrule
It would be great to be able to use the `dimaround` window rule for layers as well, to combine it with [rofi](https://github.com/davatorium/rofi) and similar floating-window-applications.

TheOnlyMrCat:

I literally just copied the code from the dimaround window rule and it seems to work.

Fixes #1657, Fixes #1889

Probably related to #3633. The same weird behaviour that shows up when applying `dimaround` to tiled windows would also happen here for shells on the same layer. There'd also be double-dimaround if there are any dimaround windows behind the layer.

It is ready for merging.

### DIFF
--- a/src/helpers/WLClasses.cpp
+++ b/src/helpers/WLClasses.cpp
@@ -10,6 +10,11 @@ SLayerSurface::SLayerSurface() {
     realPosition.registerVar();
     realSize.registerVar();
 
+    alpha.setUpdateCallback([this](void*) {
+        if (dimAround)
+            g_pHyprRenderer->damageMonitor(g_pCompositor->getMonitorFromID(monitorID));
+    });
+
     alpha.setValueAndWarp(0.f);
 }
 
@@ -26,6 +31,7 @@ void SLayerSurface::applyRules() {
     forceBlur        = false;
     ignoreAlpha      = false;
     ignoreAlphaValue = 0.f;
+    dimAround        = false;
     xray             = -1;
     animationStyle.reset();
 
@@ -47,6 +53,8 @@ void SLayerSurface::applyRules() {
                 if (!alphaValue.empty())
                     ignoreAlphaValue = std::stof(alphaValue);
             } catch (...) { Debug::log(ERR, "Invalid value passed to ignoreAlpha"); }
+        } else if (rule.rule == "dimaround") {
+            dimAround = true;
         } else if (rule.rule.starts_with("xray")) {
             CVarList vars{rule.rule, 0, ' '};
             try {
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -654,6 +654,13 @@ void CHyprRenderer::renderWindow(CWindow* pWindow, CMonitor* pMonitor, timespec*
 }
 
 void CHyprRenderer::renderLayer(SLayerSurface* pLayer, CMonitor* pMonitor, timespec* time, bool popups) {
+    static auto PDIMAROUND = CConfigValue<Hyprlang::FLOAT>("decoration:dim_around");
+
+    if (*PDIMAROUND && pLayer->dimAround && !m_bRenderingSnapshot && !popups) {
+        CBox monbox = {0, 0, g_pHyprOpenGL->m_RenderData.pMonitor->vecTransformedSize.x, g_pHyprOpenGL->m_RenderData.pMonitor->vecTransformedSize.y};
+        g_pHyprOpenGL->renderRect(&monbox, CColor(0, 0, 0, *PDIMAROUND * pLayer->alpha.value()));
+    }
+
     if (pLayer->fadingOut) {
         g_pHyprOpenGL->renderSnapshot(&pLayer);
         return;
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1910,8 +1910,8 @@ bool windowRuleValid(const std::string& RULE) {
 }
 
 bool layerRuleValid(const std::string& RULE) {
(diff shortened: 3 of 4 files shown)


src/render/Renderer.cpp
```diff
@@ -617,6 +619,11 @@ void CHyprRenderer::renderLayer(SLayerSurface* pLayer, CMonitor* pMonitor, times
 
     g_pHyprOpenGL->m_pCurrentLayer = pLayer;
 
+    if (*PDIMAROUND && pLayer->dimAround && !m_bRenderingSnapshot) {
```
vaxerski, line 622: de-dup: why not put this before the first check instead of duping this here and in renderSnapshot?

TheOnlyMrCat: Hmm. Doing that sorta works, but the dimaround flashes more-opaque at the start of the fade out and I can't tell why (it's in a separate branch against main [here](https://github.com/TheOnlyMrCat/Hyprland/commit/bfdee9f0d0cb6414fe4e9c9a606e9f375679eca9#diff-dbcd1ce1e0b916536f17c999473ad9e22c014646ddb2df15c45d726884a3481c) until I can figure that out).

I did it duplicated here because it was duplicated for the window dimaround. If I can get the dedup working here, should I try and deduplicate the code there as well?

TheOnlyMrCat: Ok I figured out why it was flashing and it's no longer doing it. I haven't done the dedup on the window code because it'd just result in more duplication again.

vaxerski:
stalling until #4656 is merged (will cause conflicts otherwise)

TheOnlyMrCat:
I've rebased onto main; should be ready to go now.

src/render/Renderer.cpp
```diff
@@ -654,6 +654,17 @@ void CHyprRenderer::renderWindow(CWindow* pWindow, CMonitor* pMonitor, timespec*
...
 
 void CHyprRenderer::renderLayer(SLayerSurface* pLayer, CMonitor* pMonitor, timespec* time, bool popups) {
+    static auto PDIMAROUND = CConfigValue<Hyprlang::FLOAT>("decoration:dim_around");
+
+    if (*PDIMAROUND && pLayer->dimAround && !m_bRenderingSnapshot && !popups) {
+        CBox monbox = {0, 0, g_pHyprOpenGL->m_RenderData.pMonitor->vecTransformedSize.x, g_pHyprOpenGL->m_RenderData.pMonitor->vecTransformedSize.y};
+        g_pHyprOpenGL->renderRect(&monbox, CColor(0, 0, 0, *PDIMAROUND * pLayer->alpha.value()));
+
+        if (pLayer->alpha.isBeingAnimated()) {
+            g_pHyprRenderer->damageMonitor(g_pCompositor->getMonitorFromID(pLayer->monitorID));
```
vaxerski, line 664: you cannot damage while rendering. I'd recommend damaging in an onUpdate callback for the layer alpha avar.

TheOnlyMrCat: Done (At least I'm pretty sure `startAnimation` is the correct place to register the callback)

src/helpers/WLClasses.cpp
```diff
@@ -173,6 +176,9 @@ void SLayerSurface::startAnimation(bool in, bool instant) {
         alpha = in ? 1.f : 0.f;
     }
 
+    if (this->dimAround)
```
vaxerski, line 179: we don't use `this->`

src/helpers/WLClasses.cpp
```diff
@@ -173,6 +176,9 @@ void SLayerSurface::startAnimation(bool in, bool instant) {
...
+        alpha.setUpdateCallback([&](void*) { g_pHyprRenderer->damageMonitor(g_pCompositor->getMonitorFromID(this->monitorID)); });
```
vaxerski, line 180: same as above, plus no need to `[&]`, you can just `[this]`

TheOnlyMrCat:
I've fixed the above. (Thanks for helping me with my admittedly patchy C++ knowledge)

src/helpers/WLClasses.cpp
```diff
@@ -173,6 +176,9 @@ void SLayerSurface::startAnimation(bool in, bool instant) {
         alpha = in ? 1.f : 0.f;
     }
 
+    if (dimAround)
+        alpha.setUpdateCallback([this](void*) { g_pHyprRenderer->damageMonitor(g_pCompositor->getMonitorFromID(monitorID)); });
```
vaxerski, line 180: the callback is not removed later. I'd put this in the layer constructor instead and in the callback itself check pLayer->dimAround

TheOnlyMrCat: Alright, done (unless you wanted me to get pLayer from the animatedvariable?)

vaxerski (approved):
lgtm

vaxerski:
wiki mr pls

PR #4646, opened 2024-02-08
Feature: `monitoraddedv2` IPC event for monitor description

spikespaz:
Wiki PR not started. Will make a PR to add them to the table when this PR is approved.

Adds five new `.socket2.sock` events, `monitoraddedv2`, `monitorremovedv2`, `focusedmonv2`, `moveworkspacev2`, and `activespecialv2`.

### [Why?](https://github.com/spikespaz/dotfiles/blob/7a01cfc0efd57545bddaaa1e9f0a8f17ddf20f2c/users/jacob%40odyssey/hyprland/monitors.nix#L122-L142)

Each of the original events provided the parameter `MONNAME`. Since then, monitor "names" (or DRM node names) are non-deterministic; if I plug a monitor into DisplayPort on a Thunderbolt dock, it may show up as `DP-1` or `DP-5`, or `DP-6`. One cannot know.

An easy way to mitigate this is to parse the output from `hyprctl -j monitors`, but this is a somewhat ugly solution that requires many more lines of code on my side. Alternatively, this PR introduces eight lines of code, saving me dozens when using the socket.

image

Each new event, where the former had only `MONNAME`, now has an added the monitor description (`szDescription`, `MONDESC` will be in the wiki), so where `MONNAME` was, now it is `MONNAME,MONDESC`.

I am currently still testing these, but the change is so trivial I think that it should probably be fine. I don't use C++, and have never contributed to Hyprland, so I might have broken something.

As long as Vaxry says it's fine, I'd trust.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -76,7 +76,7 @@ std::string monitorsRequest(eHyprCtlOutputFormat format, std::string request) {
     "vrr": {},
     "activelyTearing": {}
 }},)#",
-                m->ID, escapeJSONStrings(m->szName), escapeJSONStrings(m->szDescription), (m->output->make ? m->output->make : ""), (m->output->model ? m->output->model : ""),
+                m->ID, escapeJSONStrings(m->szName), escapeJSONStrings(m->szShortDescription), (m->output->make ? m->output->make : ""), (m->output->model ? m->output->model : ""),
                 (m->output->serial ? m->output->serial : ""), (int)m->vecPixelSize.x, (int)m->vecPixelSize.y, m->refreshRate, (int)m->vecPosition.x, (int)m->vecPosition.y,
                 m->activeWorkspace, (m->activeWorkspace == -1 ? "" : escapeJSONStrings(g_pCompositor->getWorkspaceByID(m->activeWorkspace)->m_szName)), m->specialWorkspaceID,
                 escapeJSONStrings(getWorkspaceNameFromSpecialID(m->specialWorkspaceID)), (int)m->vecReservedTopLeft.x, (int)m->vecReservedTopLeft.y,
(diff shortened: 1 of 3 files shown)


spikespaz:
I thought it was a Nix packaging issue, but I'm pretty sure that this doesn't do anything. Don't know why.

spikespaz:
I have introduced `szShortDescription` because wlroots having the DRM node name in the description is stupid and prevents me from doing what I want. I considered replacing other usages of `szDescription`, but only did so for `hyprctl` since there is also redundant information there. If the user reads the source code and discovers they can and decide they want to use the parenthesized node name suffix to match workspace rules, so be it, I did not change that. But I think it probably should be changed.

Also, it works now.

spikespaz:
`szShortDescription` needs to be trimmed.

spikespaz:
Only thing not tested is `activespecialv2`, but this is trivial so I'm not gonna.

vaxerski (approved):
ok. wiki mr needed

spikespaz:
Rebased to remove all changes except for introducing `monitoraddedv2`. The parameters of the event changed, they are now `ID, NAME, DESCRIPTION`.

src/debug/HyprCtl.cpp
```diff
@@ -76,7 +76,7 @@ std::string monitorsRequest(eHyprCtlOutputFormat format, std::string request) {
     "vrr": {},
     "activelyTearing": {}
 }},)#",
-                m->ID, escapeJSONStrings(m->szName), escapeJSONStrings(m->szDescription), (m->output->make ? m->output->make : ""), (m->output->model ? m->output->model : ""),
```
vaxerski, line 79: wait, don't change this! It's used by people to match monitors by `desc:` in rules :(

spikespaz: That's why I changed it. The parenthesized DRM node name at the end is redundant (in both pretty and JSON output). The wiki says to ignore parenthesized name at the end, so I removed it. People probably wouldn't guess that internally, you just check if the `szDescription` starts with `szShortDescription` when matching `desc:szShortDescription`. They can still use the full description from `wlroots`, but now we can remove comments about it in the wiki. I think this should be more clear.

vaxerski: oke

spikespaz:
> @spikespaz I actually had this implemented myself in a nasty, hackier way — so thank you for adding this!
> 
> Any reasoning for the `szShortDescription`?
> 
> I have a script that sets up monitors dependent on the monitors added. At home and work, I only have a specific vendor. With `monitoraddedv2` I get `monitoraddedv2>>1,DP-2,LG`. So there is no real gain here, when you use the same monitors/vendors, or do I miss something?

Can you please verify that the output `description` of `hyprctl -j monitors` for your LG monitor is the same before and after the change?

Wlroots has a monitor description, which appears to be make, model, and serial as a comma separated string. Hyprland replaces commas with spaces and deduplicates extra spaces.

The difference between description and short description is that the one from `wlroots` has the DRM node name parenthesised at the end, this is redundant and the wiki tells you to ignore it and use the first part. That first part is how separately in Hyprland source as `szShortDescription` but is made from make model serial. Might not match original wlroots description.

Also see original PR diff.

herrwusel:
> Can you please verify that the output description of hyprctl -j monitors for your LG monitor is the same before and after the change?

That is always the case. In my previous shitty implementation I got basically the entire `make`+`model`+`serial` back: "LG Electronics LG ULTRAFINE 308MASX9J302". With `monitoraddedv2` I only get the shortened `make` (see my initial comment). I don't know if thats intended.
According to
 ```
 szShortDescription =
        removeBeginEndSpacesTabs(std::format("{} {} {}", output->make ? output->make : "", output->model ? output->model : "", output->serial ? output->serial : ""));
```
it should have more information than just `LG`, no?

Here is the relevant `hyprctl` output:
```
,{
    "id": 1,
    "name": "DP-2",
    "description": "LG Electronics LG ULTRAFINE 308MASX9J302",
    "make": "LG Electronics",
    "model": "LG ULTRAFINE",
    "serial": "308MASX9J302",
    "width": 3840,
    "height": 2160,
    "refreshRate": 60.00000,
    "x": 1928,
    "y": 0,
    "activeWorkspace": {
        "id": 3,
        "name": "3"
    },
    "specialWorkspace": {
        "id": 0,
        "name": ""
    },
    "reserved": [0, 31, 0, 0],
    "scale": 1.67,
    "transform": 0,
    "focused": true,
    "dpmsStatus": true,
    "vrr": false,
    "activelyTearing": false
}]
```

PR #4261, opened 2023-12-27
Add some more splash texts

yamin-shihab:
#### Describe your PR, what does it fix/add?

It adds some additional splash texts; I just felt like there were some missing :\

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

It should work fine.

#### Is it ready for merging, or does it need work?

Yeah it's done (just five more strings in a vector).

PR #4680, opened 2024-02-11
added new splashes

DemonKingSwarn:
Percy Jackson splashes ftw


vaxerski:
how many times #4261

Eclextic:
> how many times #4261

Quick question... How would one request/vote for a new splash?
Through the Discord?

vaxerski:
you can suggest them on the discord but which and when get added is up to my discretion

Issue #2324, opened 2023-05-17
Smart Borders


Riolku:
As per #355, hyprland has smart gaps (yay!).

However, there isn't similar support for smart borders. What I'm looking for is the ability to hide borders along the screen edge when gaps_out = 0.

I can set gaps_in = 1 and disable borders, but then my borders aren't coloured.

Can we have smart borders?

vaxerski:
isnt this scriptable

Riolku:
Took a quick look, and I don't know. How would I script this?

vaxerski:
well, execute `hyprctl keyword general:border_size 0` whenever you change the gaps out to 0 I guess

Riolku:
That's not what I want. I want, with two horizontally split windows on screen, to have a border line drawn between the two, but no other border lines drawn

vaxerski:
oh, why not say that right from the start?

tcmal:
`no_gaps_when_only` in the master layout also disables borders (although the documentation only mentions gaps), which seems to do what you want, but I'm not sure if other layouts have a similar option.

Visne:
You can kind of get this effect by setting `gaps_out` to a negative value

Zeioth:
Not super prioritary, but it's a nice to have behavior.

Ghosthree3:
Would also appreciate support for this.
Previously under sway I would use `hide_edge_borders both` with no gaps. This would draw borders between windows, but not against the edge of my monitor.
Currently on Hyprland, I have to choose between either having borders in all cases, even when only a single window is present in a workspace (which reduces the size of the window and messes with aspect ratio a little), or having no borders at all, which makes windows look a little strange when side by side.

Basically, support for Sway's `hide_edge_borders` option would be very much appreciated.
> **hide_edge_borders** [--i3] none|vertical|horizontal|both|smart|smart_no_gaps
    Hides  window borders adjacent to the screen edges. Default is none. The --i3 option enables
    i3-compatible behavior to hide the title bar on  tabbed  and  stacked  containers  with  one
    child. The smart|smart_no_gaps options are equivalent to setting smart_borders smart|no_gaps
    and hide_edge_borders none.

haasn:
To anybody else frustrated this, you can *partially* alleviate this bug by doing the following:

```
windowrulev2 = noborder, fullscreen:1
```

[...]

PR #4704, opened 2024-02-14
windowrules: add 'solo' rule

haasn:

This adds a new window rule which matches a window if and only if it is the only non-floating window visible. Alternatively, it matches the window only if it is the 'root' of the layout tree.

See https://github.com/hyprwm/Hyprland/issues/2324

This is not necessarily the most elegant way of handling this condition. Perhaps it would be better to allow directly querying the number of floating / non-floating windows on the workspace, for example. That would allow encoding this rule with 'fullscreen:1` and `onworkspacenonfloating:1` (as two separate window rules).

But, I can't think of a good name for the "on workspace but non floating" rule. Maybe it would be better to instead generalize this substantially, so we could encode conditions in a more flexible way. But that is probably out-of-scope for this change.

Tested and working but see above.

### DIFF
--- a/src/Window.hpp
+++ b/src/Window.hpp
@@ -169,6 +169,7 @@ struct SWindowRule {
     int         bFullscreen  = -1;
     int         bPinned      = -1;
     int         bFocus       = -1;
+    int         bSolo        = -1;
     int         iOnWorkspace = -1;
     std::string szWorkspace  = ""; // empty means any
 };
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1058,6 +1058,7 @@ void CConfigManager::handleWindowRuleV2(const std::string& command, const std::s
     const auto PINNEDPOS       = VALUE.find("pinned:");
     const auto FOCUSPOS        = VALUE.find("focus:");
     const auto ONWORKSPACEPOS  = VALUE.find("onworkspace:");
+    const auto SOLOPOS         = VALUE.find("solo:");
 
     // find workspacepos that isn't onworkspacepos
     size_t WORKSPACEPOS = std::string::npos;
@@ -1072,7 +1073,7 @@ void CConfigManager::handleWindowRuleV2(const std::string& command, const std::s
 
     if (TITLEPOS == std::string::npos && CLASSPOS == std::string::npos && INITIALTITLEPOS == std::string::npos && INITIALCLASSPOS == std::string::npos &&
         X11POS == std::string::npos && FLOATPOS == std::string::npos && FULLSCREENPOS == std::string::npos && PINNEDPOS == std::string::npos && WORKSPACEPOS == std::string::npos &&
-        FOCUSPOS == std::string::npos && ONWORKSPACEPOS == std::string::npos) {
+        FOCUSPOS == std::string::npos && ONWORKSPACEPOS == std::string::npos && SOLOPOS == std::string::npos) {
         Debug::log(ERR, "Invalid rulev2 syntax: {}", VALUE);
         parseError = "Invalid rulev2 syntax: " + VALUE;
         return;
@@ -1105,6 +1106,8 @@ void CConfigManager::handleWindowRuleV2(const std::string& command, const std::s
             min = WORKSPACEPOS;
         if (FOCUSPOS > pos && FOCUSPOS < min)
             min = FOCUSPOS;
+        if (SOLOPOS > pos && SOLOPOS < min)
+            min = SOLOPOS;
 
         result = result.substr(0, min - pos);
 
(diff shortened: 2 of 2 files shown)


phonetic112:
can't you just do `windowrulev2 = [rule], onworkspace:1, floating:0`?

haasn:
> can't you just do `windowrulev2 = [rule], onworkspace:1, floating:0`?

Consider the situation of there being one non-floating and one floating window on the same workspace. Your condition would be `false` for both, but the condition I want would be `true` for the non-floating window, as it is the only non-floating window.

The motivating use-case here is that I want to remove borders around windows when there is no reason to draw them, which is the case whenever a window effectively covers the whole screen - either because it is fullscreened, or because it is the only non-floating window.

vaxerski:
I'd rather expand the onworkspace to support `tiled` and `floating`.

haasn:
> I'd rather expand the onworkspace to support `tiled` and `floating`.

Do we have a precedent for syntax? `onworkspace.tiled:1`, `onworkspace:tiled:1`?

vaxerski:
`1 tiled`, `2 floating`, `2 floating 1 tiled` sounds reasonable

sungyoonc:
@haasn Does the workspace selector with `onworkspace` window rule work for you?

I'm currently using this.
```ini
# When only 1 tiled window is visible
windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
windowrulev2 = rounding 0, onworkspace:w[tv1] f[-1], floating:0
workspace = w[tv1], gapsout:0

# When only 1 tiled window is visible (maximized)
windowrulev2 = noborder, fullscreen:1
windowrulev2 = rounding 0, fullscreen:1
workspace = f[1], gapsout:0
```

SymphonySimper:
> ```ini
> windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
> ```
```windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0```
This works thanks!

**Comment by SymphonySimper** [comment]
> > ```ini
> > windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
> > ```
> 
> This works thanks!



> @haasn Does the workspace selector with `onworkspace` window rule work for you?
> 
> I'm currently using this.
> 
> ```ini
> # When only 1 tiled window is visible
> windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
> windowrulev2 = rounding 0, onworkspace:w[tv1] f[-1], floating:0
> workspace = w[tv1], gapsout:0
> 
> # When only 1 tiled window is visible (maximized)
> windowrulev2 = noborder, fullscreen:1
> windowrulev2 = rounding 0, fullscreen:1
> workspace = f[1], gapsout:0
> ```



> @haasn Does the workspace selector with `onworkspace` window rule work for you?
> 
> I'm currently using this.
> 
> ```ini
> # When only 1 tiled window is visible
> windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
> windowrulev2 = rounding 0, onworkspace:w[tv1] f[-1], floating:0
> workspace = w[tv1], gapsout:0
> 
> # When only 1 tiled window is visible (maximized)
> windowrulev2 = noborder, fullscreen:1
> windowrulev2 = rounding 0, fullscreen:1
> workspace = f[1], gapsout:0
> ```



> @haasn Does the workspace selector with `onworkspace` window rule work for you?
> 
> I'm currently using this.
> 
> ```ini
> # When only 1 tiled window is visible
> windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
> windowrulev2 = rounding 0, onworkspace:w[tv1] f[-1], floating:0
> workspace = w[tv1], gapsout:0
> 
> # When only 1 tiled window is visible (maximized)
> windowrulev2 = noborder, fullscreen:1
> windowrulev2 = rounding 0, fullscreen:1
> workspace = f[1], gapsout:0
> ```

```ini
windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
```
This works thanks!

SymphonySimper:
> @haasn Does the workspace selector with `onworkspace` window rule work for you?
> 
> I'm currently using this.
> 
> ```ini
> # When only 1 tiled window is visible
> windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
> windowrulev2 = rounding 0, onworkspace:w[tv1] f[-1], floating:0
> workspace = w[tv1], gapsout:0
> 
> # When only 1 tiled window is visible (maximized)
> windowrulev2 = noborder, fullscreen:1
> windowrulev2 = rounding 0, fullscreen:1
> workspace = f[1], gapsout:0
> ```

```ini
windowrulev2 = noborder, onworkspace:w[tv1] f[-1], floating:0
```
This works thanks!

haasn:
Perfect, then this isn't needed

PR #4723, opened 2024-02-16
feat: Add css style gaps

DashieTM:
gaps can be configured via css syntax:
gaps_out = 5 -> 5 all
gaps_out = 5, 10 -> 5 top and bottom, 10 right and left
gaps_out = 5, 10, 15, 20 -> 5 top, 10 right, 15 bottom, 20 left

Note: for workspace rules I used a whitespace as delimiter e.g. 5 10 15 20 (without comma)

wiki mr: https://github.com/hyprwm/hyprland-wiki/pull/473

- existing configurations work as expected -> one number does the same
- changes the output of hyprctl workspacerules. -> instead of 1 number it will print 4

A quick check for formatting and naming should suffice ( i hope :P )


DashieTM:
Yeah, I thought about that one, and perfectly reasonable, gives me some time to make sure I didn't accidentally break something.

DashieTM:
Alright, updated to hyprlang.

There is one problem I have.
Currently a wrong gradient only creates a log, not a red banner error, as I have implemented the same custom data type, css gaps will also not cause a red banner.

DashieTM:
Yeah, that's why I was asking if this might be a missed interaction between the custom data types. It doesn't cause an error on main either, so it's not what I added.

vaxerski:
add it for the user's sake I guess

DashieTM:
Alright, how do you invoke HyprError within the set (GradientSet, CssGapSet) functions?
I tried the addParseError, but it doesn't do anything, I assume that is because this is separate from the regular parsing, e.g. after everything else is parsed. 
Perhaps a separate way to display the error?

vaxerski:
just return an error in the handler via the Hyprlang::ParseResult

DashieTM:
That's what I tried to explain. The returned error isn't handled properly. The gradient handler returns the error, just like the now implemented css gaps.

vaxerski:
fixed in hyprlang. Use hyprlang-git and it should pop up.

DashieTM:
Awesome :) the error works now for custom types.

Now the only thing that only causes logs is the workspace rules. 
-> bordersize caused a crash before on stoi, I put a try catch around it and logged the error.
-> gaps, both cause a log
Perhaps another PR later on can expand the workspace rules to also cause a banner error.

vaxerski:
I guess yea. Is this ready for review?

DashieTM:
I would say so.

src/config/ConfigManager.cpp
```diff
@@ -279,8 +306,8 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("general:border_size", {1L});
     m_pConfig->addConfigValue("general:no_border_on_floating", {0L});
     m_pConfig->addConfigValue("general:border_part_of_window", {1L});
-    m_pConfig->addConfigValue("general:gaps_in", {5L});
-    m_pConfig->addConfigValue("general:gaps_out", {20L});
+    m_pConfig->addConfigValue("general:gaps_in", Hyprlang::CConfigCustomValueType{&configHandleGapSet, configHandleGapDestroy, "5"});
```
vaxerski, line 309: why are you mixing `&func` and `func`? They are the same, pick one. Probably `func`

src/config/ConfigManager.cpp
```diff
@@ -2115,12 +2142,22 @@ std::optional<std::string> CConfigManager::handleWorkspaceRules(const std::strin
...
+        } else if ((delim = rule.find("gapsout:")) != std::string::npos) {
+            CVarList varlist = CVarList(rule.substr(delim + 8), 0, ' ');
+            wsRule.gapsOut   = CCssGapData();
+            try {
+                wsRule.gapsOut->parseGapData(varlist);
+            } catch (...) { Debug::log(ERR, "Error parsing workspace rule gaps: {}", rule.substr(delim + 8)); }
+        } else if ((delim = rule.find("bordersize:")) != std::string::npos)
+            try {
+                wsRule.borderSize = std::stoi(rule.substr(delim + 11));
+            } catch (...) { Debug::log(ERR, "Error parsing workspace rule bordersize: {}", rule.substr(delim + 11)); }
```
vaxerski, line 2160: throw an error here: `return "error string"`

src/layout/DwindleLayout.cpp
```diff
@@ -134,12 +134,14 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
 
     PWINDOW->updateSpecialRenderData();
 
-    static auto* const PGAPSIN         = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("general:gaps_in");
-    static auto* const PGAPSOUT        = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("general:gaps_out");
     static auto* const PNOGAPSWHENONLY = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("dwindle:no_gaps_when_only");
+    static auto* const PGAPSINLOCKED   = (Hyprlang::CUSTOMTYPE* const*)g_pConfigManager->getConfigValuePtr("general:gaps_in");
```
vaxerski, line 138: uhh why renamed to LOCKED? they aren't locked?

src/layout/MasterLayout.cpp
```diff
@@ -633,13 +634,15 @@ void CHyprMasterLayout::applyNodeDataToWindow(SMasterNodeData* pNode) {
 
     PWINDOW->updateSpecialRenderData();
 
-    static auto* const PGAPSIN         = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("general:gaps_in");
-    static auto* const PGAPSOUT        = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("general:gaps_out");
     static auto* const PNOGAPSWHENONLY = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("master:no_gaps_when_only");
     static auto* const PANIMATE        = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("misc:animate_manual_resizes");
+    static auto* const PGAPSINLOCKED   = (Hyprlang::CUSTOMTYPE* const*)g_pConfigManager->getConfigValuePtr("general:gaps_in");
+    static auto* const PGAPSOUTLOCKED  = (Hyprlang::CUSTOMTYPE* const*)g_pConfigManager->getConfigValuePtr("general:gaps_out");
```
vaxerski, line 640: same ere

src/layout/DwindleLayout.cpp
DashieTM: Unsure what to call those, PGAPSINDATA better?

vaxerski: sounds better

src/config/ConfigManager.cpp
DashieTM: Oh this resolved the banner issue as well, thanks.

vaxerski (approved):
looks god to me

PR #4724, opened 2024-02-16
feat: Allow passing a window argument to the moveoutofgroup dispatcher

morr0ne:
Very simple change that allows for moving a specific window out of a group. It follows the same convention as the `pin` and `togglefloating` dispatchers by accepting either no args or "active" to act on the focused windows and still remain backwards compatible

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -797,7 +797,7 @@ void CKeybindManager::clearKeybinds() {
 void CKeybindManager::toggleActiveFloating(std::string args) {
     CWindow* PWINDOW = nullptr;
 
-    if (args != "" && args != "active" && args.length() > 1)
+    if (args != "active" && args.length() > 1)
         PWINDOW = g_pCompositor->getWindowByRegex(args);
     else
         PWINDOW = g_pCompositor->m_pLastWindow;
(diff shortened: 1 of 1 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -2177,7 +2177,12 @@ void CKeybindManager::moveOutOfGroup(std::string args) {
     if (!*PIGNOREGROUPLOCK && g_pKeybindManager->m_bGroupsLocked)
         return;
 
-    const auto PWINDOW = g_pCompositor->m_pLastWindow;
+    CWindow* PWINDOW = nullptr;
+
+    if (args != "" && args != "active" && args.length() > 1)
```
vaxerski, line 2182: args.empty() not != "", but you already check for .length() so why is it there even

morr0ne: That's a good question. I wondered the same. The other two implementation did the same thing so I figured there must have been a good reason and just left it there

morr0ne: Fixed it to use the empty and also did the same with the other implementations

vaxerski:
honestly I'd rather check for `length() > 1`, some of these have had instances of single-char args before

morr0ne:
Should be good to go

vaxerski (approved):
lgtm, needs a wiki mr

morr0ne:
Done: https://github.com/hyprwm/hyprland-wiki/pull/475

PR #4727, opened 2024-02-16
monitor: don't damage twice

abmantis:
When scaled, the ring is already fully damaged, no need to add the region damage.

Also moved a variable that was being declared way to far for where it is actually used.

Ready.


vaxerski (approved):
ok

PR #4738, opened 2024-02-18
Fix outputmgr nullptr crash (kanshi/shikane like tools)

levnikmyskin:
This PR fixes two (related) issues:
1. fixes `nullptr` crash in `outputmgr` when using `kanshictl reload` repeteadly;
2. fixes `nullptr` crash in `outputmgr` when using `kanshictl switch` to re-enable a disabled monitor (laptop's internal)

The main thing here is that in `outputMgrApplyTest` we might get disabled monitors, and thus we should look into the `realMonitors` vector.  
I didn't check whether other code relying on `Compositor::getMonitorFromOutput` was well-managing disabled outputs, so I decided to create another `Compositor::getRealMonitorFromOutput` method. Maybe this can be changed, let me know

see above

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -881,6 +881,16 @@ CMonitor* CCompositor::getMonitorFromOutput(wlr_output* out) {
     return nullptr;
 }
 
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
lgtm

Issue #4667, opened 2024-02-10
hyprland 0.35 xkb_layout keyboard config no longer works


pschmitt:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 84ab8d11e8951a6551d1e1bf87796a8589da6d47  ().
Date: 2024-02-05
Tag:

flags: (if any)

```

</details>

### Bug or Regression?

Regression

### Description

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 84ab8d11e8951a6551d1e1bf87796a8589da6d47  ().
Date: 2024-02-05
Tag:

flags: (if any)

```

</details>

### How to reproduce

I have a custom xkb map set for my keyboard which works fine with Hyprland 0.34 (03ebbe18ed8517ee22591eac82cd54322f42cb7d)

With 0.35 it seems to be ignored, even though the log report that the layout has been set to what I was expecting (de_hhkb in my case), there are some warnings about xkb in the log though.

Relevant Hyprland config:

```
device:pfu-limited-hhkb-hybrid {
  kb_layout = de_hhkb
}

device:pfu-limited-hhkb-hybrid-consumer-control {
  kb_layout = de_hhkb
}

device:pfu-limited-hhkb-hybrid-keyboard {
  kb_layout = de_hhkb
}

device:hhkb-hybrid_1-keyboard {
  kb_layout = de_hhkb
}
```

`.config/xkb/symbols/de_hhkb`:

```
default partial alphanumeric_keys

xkb_symbols "basic" {
  include "de"
  name[Group1] = "Custom HHKB DE layout by pschmitt";
  // https://github.com/swaywm/sway/issues/4250

  // 1. Determine the scancode of the key
  // $ wev
  //
  // 2. Get the internal ID
  // $ </usr/share/X11/xkb/keycodes/evdev | grep -E "=\s+(132)"
  // <AE13> = 132;   // Yen
  //
  // 3. The key name to map to can be fetched via:
  // $ </usr/include/X11/keysymdef.h grep -E "#define XK_"

  // The mapping goes as follows:
  // key <KEY_ID> { [ map1, map2, map3, map4 ]};
  // Where:
  // map1: normal key press
  // map2: shift + key
  // map3: AltGr + key
  // map4: AltGr + shift + key

  // YEN KEY (Insert) -> "pipe" char
  key <AE13> { [ bar, bar, bar, bar ] };

[...]

PR #4739, opened 2024-02-18
fix: keys without keysyms triggering random binds

rszyma:

Fixes https://github.com/hyprwm/Hyprland/issues/4667

In #4667 it was deducted that only yen and ro keys are buggy, so I remapped my keyboard to be able to output yen key. In Hyprland I debug-logged yen key when passed to `handleKeybinds` it turns out its keysym can't be resolved correctly for some reason. So it's keysym is 0-value: 
```
key.keyName: ""
key.keycode: 248
key.keysym: 0
```

In `handleKeybinds` some keybinds kinda incorrectly take the `else` branch from the `if-elseif-else` conditional, e.g. `switch:off:Lid Switch` as in this branch `xkb_keysym_from_name` is called on them and that results in `KBKEY = 0`. That is later compared to keysym of current key event and voila - for yen/ro key it will be `0 == 0` comparison here and it will incorrectly trigger currenlty iterated keybind https://github.com/hyprwm/Hyprland/blob/b36a2009d2f400e0564af8cf6c5ec0d8e80523b5/src/managers/KeybindManager.cpp#L528-L529  

Honestly, I'm still not sure why 37b76cd made that regression, and I've already spent too much time trying to figure it out with no results, so I moved on and instead tried to fix it based on what we know about the bug.

Waiting for @pschmitt to confirm it's fixed, then should be good to go.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -510,7 +510,19 @@ bool CKeybindManager::handleKeybinds(const uint32_t modmask, const SPressedKeyWi
         } else {
             // oMg such performance hit!!11!
             // this little maneouver is gonna cost us 4µs
-            const auto KBKEY      = xkb_keysym_from_name(k.key.c_str(), XKB_KEYSYM_CASE_INSENSITIVE);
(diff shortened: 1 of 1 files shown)


pschmitt:
Tested and yup, this fixes my issue! Thank you very much.

vaxerski (approved):
lgtm thanks as usual

PR #4766, opened 2024-02-20
master layout: change the mfact dispatcher to use splitratio

thejch:
Change the mfact dispatcher to use split ratio, allowing increment/decrement.
I realized this dispatcher does the same thing as `splitratio`, so I'm just changing it to call the latter, this allows you to increment/decrement the mfact in addition to only being able to set the exact mfact.

I also thought about just removing it since splitratio exists, but the splitratio dispatcher is really confusing, as it does two different and not necessarily related things in dwindle and master. You also wont find it in the master layout wiki when you look for it since it's not a layout specific dispatcher, but it kinda should be imo. The mfact is a very layout-specific config that is specific to master, and you usually won't look in the universal dispatchers for such a thing, and it's not clear that splitratio changes the mfact in master. I think it would make more sense in the future to have two different dispatchers separately in dwindle and master that will do what this `splitratio` dispatcher does. Having a dispatcher for the mfact itself seems much more easier to understand. 

Needs wiki mr

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -1314,19 +1314,7 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
     } else if (command == "orientationcycle") {
         runOrientationCycle(header, &vars, 1);
     } else if (command == "mfact") {
-        if (vars.size() >= 2) {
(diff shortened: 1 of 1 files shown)


janpeterd:
I added a wiki mr

vaxerski (approved):

Issue #4760, opened 2024-02-19
Ability to contain user interaction with window size


Ligthiago:
### Description

Some windows just don't look right when they are tiled, very large or very small, so it would be nice to have a way of doing this:
- Set the maximum possible size for floating window. At the moment (as far as I can see in docs) there is only a static `maxsize` rule which only affects window creation, user is still able to resize the window more than this. I tried to find the upper limit, but Hyprland just crashed at 40000x40000.
- Set minimum possible size for floating window, same as before, but right now minimum size is at least exist - it is 20x20 window, just point on screen.  Can it be customized globally and per window?
![2024-02-19T23:03:26,857272455+04:00](https://github.com/hyprwm/Hyprland/assets/142721811/afbad4dd-175d-46bd-b6dd-5e5d1900c517)

- Same issue for tiling windows with use case examples https://github.com/hyprwm/Hyprland/issues/4414

- Add window rule that allows always floating window (set float and ignore togglefloating dispatcher, even if user accidentally press it). I guess it can be done with bash script, but window rule is more convenient.

### Some use case for max/min possible size:
![usecase](https://github.com/hyprwm/Hyprland/assets/142721811/88adc292-8fbb-4bda-a9fe-90d539081b33)

Ligthiago:
Btw, after some recent changes, minimal size seems to work fine in most GTK applications, at least they can't become barely existing unusable dot, smaller that it's content. Very good.
So examples is no longer reproducible, but issue is still valid for other apps and maxsize.

https://github.com/hyprwm/Hyprland/assets/142721811/f0137bb0-c7c2-45da-8d5d-819d0c074e47

vaxerski:
The app itself should set its maxsize properly.

PR #4775, opened 2024-02-20
Make min/maxsize rules dynamic

phonetic112:
Fixes https://github.com/hyprwm/Hyprland/issues/4760

Converts `minsize` and `maxsize` to dynamic rules. I was going to make it a window prop as well but I figured you could use `hyprctl keyword` instead.

Nah

Ready, needs wiki mr

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -570,26 +570,6 @@ void CWindow::applyDynamicRule(const SWindowRule& r) {
         m_sAdditionalConfigData.forceNoDim = true;
     } else if (r.szRule == "forcergbx") {
         m_sAdditionalConfigData.forceRGBX = true;
-    } else if (r.szRule.starts_with("maxsize")) {
-        try {
-            size_t      spacePos = r.szRule.find(' ');
-            std::string vecStr;
-            if (spacePos != std::string::npos) {
-                vecStr = r.szRule.substr(spacePos + 1);
-            }
-
-            m_sAdditionalConfigData.maxSize = configStringToVector2D(vecStr);
-        } catch (std::exception& e) { Debug::log(ERR, "maxsize rule \"{}\" failed with: {}", r.szRule, e.what()); }
-    } else if (r.szRule.starts_with("minsize")) {
-        try {
-            size_t      spacePos = r.szRule.find(' ');
-            std::string vecStr;
-            if (spacePos != std::string::npos) {
-                vecStr = r.szRule.substr(spacePos + 1);
-            }
-
-            m_sAdditionalConfigData.minSize = configStringToVector2D(vecStr);
-        } catch (std::exception& e) { Debug::log(ERR, "minsize rule \"{}\" failed with: {}", r.szRule, e.what()); }
     } else if (r.szRule == "opaque") {
         if (!m_sAdditionalConfigData.forceOpaqueOverridden)
             m_sAdditionalConfigData.forceOpaque = true;
@@ -704,6 +684,10 @@ void CWindow::applyDynamicRule(const SWindowRule& r) {
             m_eIdleInhibitMode = IDLEINHIBIT_FULLSCREEN;
         else
             Debug::log(ERR, "Rule idleinhibit: unknown mode {}", IDLERULE);
+    } else if (r.szRule.starts_with("maxsize")) {
+        m_sAdditionalConfigData.maxSize = configStringToVector2D(r.szRule);
+    } else if (r.szRule.starts_with("minsize")) {
+        m_sAdditionalConfigData.minSize = configStringToVector2D(r.szRule);
     }
 }
 
@@ -718,9 +702,9 @@ void CWindow::updateDynamicRules() {
     m_sAdditionalConfigData.forceNoDim       = false;
     if (!m_sAdditionalConfigData.forceOpaqueOverridden)
         m_sAdditionalConfigData.forceOpaque = false;
-    m_sAdditionalConfigData.forceNoAnims    = false;
-    m_sAdditionalConfigData.minSize         = Vector2D(-1, -1);
     m_sAdditionalConfigData.maxSize         = Vector2D(99999, 99999);
+    m_sAdditionalConfigData.minSize         = Vector2D(-1, -1);
+    m_sAdditionalConfigData.forceNoAnims    = false;
     m_sAdditionalConfigData.animationStyle  = std::string("");
     m_sAdditionalConfigData.rounding        = -1;
     m_sAdditionalConfigData.dimAround       = false;
@@ -733,6 +717,7 @@ void CWindow::updateDynamicRules() {
     m_eIdleInhibitMode                      = IDLEINHIBIT_NONE;
 
     const auto WINDOWRULES = g_pConfigManager->getMatchingRules(this);
+
     for (auto& r : WINDOWRULES) {
         applyDynamicRule(r);
     }
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -691,6 +691,11 @@ void CWindow::applyDynamicRule(const SWindowRule& r) {
             if (std::stoi(vars[0]))
                 m_sAdditionalConfigData.maxSize = configStringToVector2D(r.szRule + " " + vars[0]);
         } catch (std::invalid_argument& e) {}
+        const auto SIZE = Vector2D(std::min((double)m_sAdditionalConfigData.maxSize.toUnderlying().x, m_vRealSize.goal().x),
+                                   std::min((double)m_sAdditionalConfigData.maxSize.toUnderlying().y, m_vRealSize.goal().y));
+        m_vRealSize     = SIZE;
+        g_pXWaylandManager->setWindowSize(this, m_vRealSize.goal());
+        setHidden(false);
     } else if (r.szRule.starts_with("minsize")) {
         CVarList vars(r.szValue, 0, ',');
         m_sAdditionalConfigData.minSize = configStringToVector2D(r.szRule);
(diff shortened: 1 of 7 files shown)


src/Window.hpp
```diff
@@ -141,26 +141,28 @@ struct SWindowSpecialRenderData {
...
+    CWindowOverridableVar<bool>     forceOpaqueOverridden = false; // if true, a rule will not change the forceOpaque state. This is for the force opaque dispatcher.
+    CWindowOverridableVar<bool>     forceAllowsInput      = false;
+    CWindowOverridableVar<bool>     forceNoAnims          = false;
+    CWindowOverridableVar<bool>     forceNoBorder         = false;
+    CWindowOverridableVar<bool>     forceNoShadow         = false;
+    CWindowOverridableVar<bool>     forceNoDim            = false;
+    CWindowOverridableVar<bool>     noFocus               = false;
+    CWindowOverridableVar<bool>     windowDanceCompat     = false;
+    CWindowOverridableVar<bool>     noMaxSize             = false;
+    CWindowOverridableVar<Vector2D> maxSize               = Vector2D(99999, 99999);
```
vaxerski, line 157: if you're making it a prop, add it as one. If you're not making this a prop, don't make it a CWindowOverridableVar. Furthermore, since this is optional, make it such. `std::optional` exists.

phonetic112: If I make it a prop, how do I make it override the window rule? could I just use applyDynamicRule when setting the prop?

vaxerski: what?

phonetic112: If the window has a rule from the config the setprop command won't work.

phonetic112: I'm regarded, I wasn't using hyprctl from the build dir

src/Window.cpp
```diff
@@ -698,6 +702,8 @@ void CWindow::updateDynamicRules() {
     m_sAdditionalConfigData.forceNoDim       = false;
     if (!m_sAdditionalConfigData.forceOpaqueOverridden)
         m_sAdditionalConfigData.forceOpaque = false;
+    m_sAdditionalConfigData.maxSize         = Vector2D(99999, 99999);
+    m_sAdditionalConfigData.minSize         = Vector2D(20, 20);
```
vaxerski, line 706: dont default it to 20,20, do it to 1,1

src/Window.cpp
```diff
@@ -698,6 +702,8 @@ void CWindow::updateDynamicRules() {
     m_sAdditionalConfigData.forceNoDim       = false;
     if (!m_sAdditionalConfigData.forceOpaqueOverridden)
         m_sAdditionalConfigData.forceOpaque = false;
+    m_sAdditionalConfigData.maxSize         = Vector2D(99999, 99999);
```
vaxerski, line 705: `std::numeric_limits<double>::max()`

src/Window.hpp
```diff
@@ -142,26 +142,28 @@ struct SWindowSpecialRenderData {
...
```
vaxerski, line 158: same as above

src/helpers/MiscFunctions.cpp
```diff
@@ -722,6 +722,22 @@ int64_t configStringToInt(const std::string& VALUE) {
     return std::stoll(VALUE);
 }
 
+Vector2D configStringToVector2D(const std::string& VALUE) {
+    size_t      spacePos = VALUE.find(' ');
```
vaxerski, line 726: this function should not accept anything but `123 456` or `123, 456`

phonetic112: wouldn't allowing commas break the window rule syntax

vaxerski: thats why allow both

phonetic112: no like, say I have this rule:

`windowrulev2 = maxsize 500 500, class:foot`

this works, but this rule:

`windowrulev2 = maxsize 500, 500, class:foot`

now the second `500` is interpreted as the window var

vaxerski: thats why there they would be parsed with a space

phonetic112: ok check now

src/helpers/MiscFunctions.cpp
```diff
@@ -722,6 +722,14 @@ int64_t configStringToInt(const std::string& VALUE) {
     return std::stoll(VALUE);
 }
 
+Vector2D configStringToVector2D(const std::string& VALUE) {
+    const auto SIZESTR  = VALUE.substr(VALUE.find(' ') + 1);
+    const auto SIZEXSTR = SIZESTR.substr(0, SIZESTR.find(' '));
+    const auto SIZEYSTR = SIZESTR.substr(SIZESTR.find(' ') + 1);
+
+    return Vector2D(std::stoll(SIZEXSTR), std::stoll(SIZEYSTR));
```
vaxerski, line 730: this still won't parse `123,456` or `123, 456`

I want a function that parses all three:
 - `123 456`
 - `123,456`
 - `123 456`

src/layout/IHyprLayout.cpp
```diff
@@ -338,8 +338,10 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
     } else if (g_pInputManager->dragMode == MBIND_RESIZE || g_pInputManager->dragMode == MBIND_RESIZE_FORCE_RATIO || g_pInputManager->dragMode == MBIND_RESIZE_BLOCK_RATIO) {
         if (DRAGGINGWINDOW->m_bIsFloating) {
 
-            Vector2D MINSIZE = g_pXWaylandManager->getMinSizeForWindow(DRAGGINGWINDOW).clamp({20, 20});
-            Vector2D MAXSIZE = g_pXWaylandManager->getMaxSizeForWindow(DRAGGINGWINDOW);
+            Vector2D MINSIZE = g_pXWaylandManager->getMinSizeForWindow(DRAGGINGWINDOW)
+                                   .clamp({DRAGGINGWINDOW->m_sAdditionalConfigData.minSize.toUnderlying().x, DRAGGINGWINDOW->m_sAdditionalConfigData.minSize.toUnderlying().y});
```
vaxerski, line 342: `toUnderlying` will ret a vec, why are you decomposing it and then composing back??

src/helpers/MiscFunctions.cpp
phonetic112: ok should be good now

src/helpers/MiscFunctions.cpp
```diff
@@ -722,6 +722,14 @@ int64_t configStringToInt(const std::string& VALUE) {
     return std::stoll(VALUE);
 }
 
+Vector2D configStringToVector2D(const std::string& VALUE) {
+    const auto SIZESTR  = VALUE.substr(VALUE.find(' ') + 1);
```
vaxerski, line 726: brother. With all due respect, are you ok??

This function still takes only `x y`, you moved the logic OUTSIDE, not IN THIS FUNCTION.

Furthermore, why are you even passing a bullshit argument here as the first arg??

Ugh, I'll just fix this bs myself.

vaxerski:
[patch.txt](https://github.com/hyprwm/Hyprland/files/14509477/patch.txt)
cant commit to your branch for whatever reason from my laptop

src/helpers/MiscFunctions.cpp
phonetic112: >you moved the logic OUTSIDE, not IN THIS FUNCTION

It has to be outside, because how else would the window rule handle the extra arg?

>This function still takes only x y

It doesn't? Technically it could even take any additional character that isn't a comma or number right after it, like `500. 500` or even `500dsflskdfmlk 500sdnfkljfnslk`

Your patch didn't even work, because it interprets the `y` arg as the window arg itself:

`[LOG] Window rule minsize 1000 ->  500, class:foot matched`

when it should be:

`[LOG] Window rule minsize 1000 500 ->  class:foot matched`

vaxerski: you're piping `500hwufbgufrhbgeru` into `std::stoll`, that doesn't throw?

> Your patch didn't even work, because it interprets the y arg as the window arg itself:

simple fix

phonetic112: >you're piping 500hwufbgufrhbgeru into std::stoll, that doesn't throw?

no, only if there are non-numbers *before* the number like `alfdkjgfdlk500`

vaxerski: that's... cringe.

Still:
 - why does it take a useless word at the beginning? (the func wont parse `500 500` it will take `garbage 500 500` and extract the `500 500`)
 - why the try catch block calling it twice?

phonetic112: true, changed

src/Window.cpp
```diff
@@ -684,6 +684,28 @@ void CWindow::applyDynamicRule(const SWindowRule& r) {
             m_eIdleInhibitMode = IDLEINHIBIT_FULLSCREEN;
         else
             Debug::log(ERR, "Rule idleinhibit: unknown mode {}", IDLERULE);
+    } else if (r.szRule.starts_with("maxsize")) {
+        CVarList vars(r.szValue, 0, ',');
+        try {
+            std::stoi(vars[0]);
```
vaxerski, line 690: what could be there that we are testing for?

phonetic112: If the first var can convert, it's `y`. otherwise it's the window and thus probably formatted as `x y`

vaxerski: ...what?

`windowrulev2 = maxsize 20 30, class:something`

szValue will be `maxsize 20 30`, how will the first arg ever parse?

phonetic112: no, szValue is `class:something`, `maxsize 20 30` is szRule. If commas were used:

`windowrulev2 = maxsize 20, 30, class:something`

then `30` would be the first var of szValue and the try block would be executed

phonetic112: ah ok

phonetic112:
code style bugged?

vaxerski:
maybe

src/helpers/MiscFunctions.cpp
```diff
@@ -722,6 +722,32 @@ int64_t configStringToInt(const std::string& VALUE) {
...
 }
 
+Vector2D configStringToVector2D(const std::string& VALUE) {
+    std::istringstream iss(VALUE);
+    std::string        token;
+
+    if (!std::getline(iss, token, ' ') && !std::getline(iss, token, ','))
+        throw std::invalid_argument("Invalid string format");
+
+    if (!std::all_of(token.begin(), token.end(), ::isdigit))
```
vaxerski, line 732: we have `isNumber()`

phonetic112: ok

vaxerski (approved):
lgtm

Issue #3698, opened 2023-10-30
regression: cursor_inactive_timeout is broken


andresilva:
### Hyprland Version

1c9d6b94d178c234cc5174dfe6d2f952cf0438f3

### Bug or Regression?

Regression

### Description

Dispatching `movefocus` now activates the cursor. Did a bisect, regression was introduced in 7a5234a0cca2508b7a5ede1fc60dc3ae8b72a51c.

### How to reproduce

- enable `cursor_inactive_timeout`
- move window focus
- cursor reappears

andresilva:
What do you mean? There are two issue templates: bug and feature request.

vaxerski:
:)

andresilva:
friendly ping. I tried to debug this but couldn't figure it out.

dusanx:
I have started working on this, MR should be ready in a few. Cursor is being rewritten/forced in a weird way so fix requires slight logic change with cursor handling. I didn't go all the way trough code, so far seems very fixable.

andresilva:
Fixed in #4184. Thanks @dusanx!

andresilva:
Broken by: https://github.com/hyprwm/Hyprland/commit/b500e5699b9aa0f5a92805707bcc88ba151a7c82

When you move focus to a place where the cursor should change it gets re-activated.

vaxerski:
[patch.txt](https://github.com/hyprwm/Hyprland/files/14347690/patch.txt)

what if

vaxerski:
hm, right.

andresilva:
FWIW this isn't any different than it was before so it's probably worth it to merge that patch anyway (the changes in `setCursorFromName` aren't needed though).

PR #4780, opened 2024-02-21
renderer: ignore set cursor surface if cursor should be hidden

andresilva:

Fix https://github.com/hyprwm/Hyprland/issues/3698.

No.

Yes.


vaxerski (commented):
well two issues with this.

firstly, the check in the other method taking a shape is needed too (e.g. hidden because of a timeout, we cant let shape be set)

and second to avoid wrong restores it'd make sense to return after setting m_sLastCursorData so we can restore the proper cursor on unhide

andresilva:
Updated.

src/render/Renderer.cpp
```diff
@@ -2152,26 +2152,32 @@ bool CHyprRenderer::applyMonitorRule(CMonitor* pMonitor, SMonitorRule* pMonitorR
...
     m_bCursorHasSurface = true;
 
-    if ((name == m_sLastCursorData.name || m_bCursorHidden) && !force)
+    if (name == m_sLastCursorData.name && !force)
         return;
 
     m_sLastCursorData.name = name;
     m_sLastCursorData.surf.reset();
 
+    if (m_bCursorHidden)
```
vaxerski, line 2178: honestly for the sake of clarity `force` should also omit this check. I know it won't change anything rn, but in the future it might be annoying

vaxerski (commented):
lgtm apart from tat

vaxerski (approved):
top thanks

PR #4783, opened 2024-02-21
Fix building plugins

q234rty:

Fixes building plugins after ddf022d61c63fb36b4abba392682772690c06b5c

Long term it's probably better port the whole codebase to relative paths and remove https://github.com/hyprwm/Hyprland/blob/main/CMakeLists.txt#L87-L88 .
Alternatively we can just add those two dirs to the .pc file, but I remember that was explicitly rejected since plugins should use the full path.

Ready.


vaxerski (approved):
aiaiai I missed this

PR #4790, opened 2024-02-21
Implement notification dimissing

### LINKED ISSUE #4787: Add a way to dismiss notifications
### Description

Currently, there is `hyprctl notify` command, which can be used to create system notifications managed by hyprland. I'd love to be able to do something like `hyprctl notify-dismiss 2`, to dismiss the last 2 notifications, perhaps with a default to dismiss all notifications.

### My use-case

I'm using `hyprctl notify` to send a notification from `hypridle`, like so:

```
listener {
  timeout = 300
  # use hyprland notifications, in case dunst notifs are paused or whatever, also these have a nice progress bar
  on-timeout = hyprctl notify 0 10000 "rgb(ff0000)" "System going idle in 10 seconds..."
}
```

Which works perfectly with the 10s timeout, as that's when another listener kicks in locking the session. However, it's pretty annoying that I can't dismiss this notification with `on-resume`.

ItsDrike:

This adds logic for notification dismissing via hyprctl. Fixes #4787

I'm don't really love the name `dismissnotify` for the hyprctl command. I wanted to do `notifydismiss` instead originally. 

However, I wasn't able to do that with the way command registering works now, where if there's already a `notify` command, all commands prefixed with `notify` will be handled from there. There is an `exact` option when registering, but that wouldn't allow any arguments to be passed. 

Short of changing the way this logic works, one thing that comes to mind would be `if (vars[0] == "notifydismiss")` in the `dispatchNotify` func, but I didn't see this kind of logic anywhere else in the dispatch handlers, and I'm not sure this prefix based handling was actually intended, so I chose to go with the unique `dismissnotify` name.

Another thing I will mention is that if there isn't enough notifications to dismiss, the command will still respond with `ok`, if desired, I can add some other message for this case.

Yes

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -33,6 +33,7 @@ commands:
     cursorpos
     decorations
     devices
+    dismissnotify
     dispatch
     getoption
     globalshortcuts
@@ -385,6 +386,8 @@ int main(int argc, char** argv) {
         request(fullRequest, 3);
     else if (fullRequest.contains("/plugin"))
         request(fullRequest, 1);
+    else if (fullRequest.contains("/dismissnotify"))
+        request(fullRequest, 0);
     else if (fullRequest.contains("/notify"))
         request(fullRequest, 2);
     else if (fullRequest.contains("/output"))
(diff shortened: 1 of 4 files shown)


vaxerski (approved):
brother if I havent reviewed your mr in a week just ping me, I didn't notice this amongst the slew of notifications I get daily.

vaxerski:
wiki MR needed

ItsDrike:
> wiki MR needed

Opened: https://github.com/hyprwm/hyprland-wiki/pull/498

PR #4805, opened 2024-02-23
CrashReporter: Fix compilation with musl libc

earboxer:
It can be assumed this doesn't function correctly: my 'configuration does not support execinfo.h', so I have no backtrace to test against

Fixes compilation with musl libc, which does not include `dladdr1`.

https://github.com/hyprwm/Hyprland/commit/3d9ca6381df1cdbf1731d7e2b39ea3f7574fce1e#commitcomment-138447420

I used `#ifdef __GLIBC__` (as I don't think musl libc defines anything). If there are other libc's which provide dladdr1, it should be added to that ifdef (but how many libc's are there? [bionic doesn't include dladdr1](https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/dlfcn.h), but [uClibc does](https://git.uclibc.org/uClibc/tree/include/dlfcn.h)-- if anyone uses uClibc, they might want to change the ifdef to get better backtraces)

### DIFF
--- a/src/debug/CrashReporter.cpp
+++ b/src/debug/CrashReporter.cpp
@@ -108,18 +108,23 @@ void CrashReporter::createAndSaveCrash(int sig) {
 
     std::string addrs = "";
     for (size_t i = 0; i < CALLSTACK.size(); ++i) {
+#ifdef __GLIBC__
         // convert in memory address to VMA address
         Dl_info          info;
         struct link_map* linkMap;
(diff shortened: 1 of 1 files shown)


thejch:
> musl's link.h doesn't include it, (so adding it reduces the amount of stuff that needs to go in the ifdef)

Y'oure not using that stuff anyway

earboxer:
@thejch Fair enough. Changed as suggested.

vaxerski (approved):
lgtm

Issue #3558, opened 2023-10-12
Random hang on exit


coffebar:
### Hyprland Version

Hyprland, built from branch main at commit 3f09b14381e8b28dd2cc1d292763374f2d6c8484 dirty (makefile: fixup permissions on installed binaries). Tag: v0.30.0-22-g3f09b143

### Bug or Regression?

Bug

### Description

I noticed that Hyprland randomly hangs on exit. Instead of exit, it blocks my computer with a black screen and I can't access tty to do something with it. 

From jornalctl after reboot I see that other programs continued working in background (syncthing) while I see only black screen. 

### How to reproduce

Can't reproduce it manually, even on my machine. 

### Crash reports, logs, images, videos

I created this script to Force `exit`:

```bash
echo "Hyprland exit" | systemd-cat -t coffebar -p info
hyprctl dispatch exit &
sleep 10
echo "Hyprland failed to exit" | systemd-cat -t coffebar -p err
killall -9 Hyprland
```

I hoped that it will kill compositor when it hangs on exit (after 10 seconds). 

But it didn't work.  What I found in journal:

> Oct 12 19:26:18 crab coffebar[115698]: Hyprland exit
> Oct 12 19:26:18 crab polkitd[1182]: Unregistered Authentication Agent for unix-session:1 (system bus name :1.22, object path /org/kde/PolicyKit1/AuthenticationAgent, locale en_US.UTF-8) (disconnected from bus)
> Oct 12 19:26:28 crab coffebar[115714]: Hyprland failed to exit

After that, nothing happens in logs, except the syncthing does its job as usual. 

It's odd, I'm not sure what is going on here.

### Logs

Beginning part:
```[LOG] User-specified config location: '/home/abc/.config/hyprland/hyprland.conf'
Welcome to Hyprland!
[LOG] Instance Signature: 3f09b14381e8b28dd2cc1d292763374f2d6c8484_1697194161
[LOG] Hyprland PID: 30565
[LOG] ===== SYSTEM INFO: =====
[LOG] System name: Linux
[LOG] Node name: crab
[LOG] Release: 6.5.7-arch1-1
[LOG] Version: #1 SMP PREEMPT_DYNAMIC Tue, 10 Oct 2023 21:10:21 +0000


[LOG] GPU information:
[...]

PR #4811, opened 2024-02-23
Try to fix the exit hang

thejch:
Fixes https://github.com/hyprwm/Hyprland/issues/3558, hopefully
Remove some signals before starting the cleanup, so that the crash reporter wont get triggered when exiting.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1207,8 +1207,13 @@ void CInputManager::onKeyboardKey(wlr_keyboard_key_event* e, SKeyboard* pKeyboar
         updateKeyboardsLeds(pKeyboard->keyboard);
     }
 
-    if (m_bExitTriggered)
+    if (m_bExitTriggered) {
+        // remove signals so that the crash reporter won't get accidentally triggered
+        signal(SIGABRT, SIG_DFL);
+        signal(SIGSEGV, SIG_DFL);
+
         g_pCompositor->cleanup();
+    }
 }
 
 void CInputManager::onKeyboardMod(void* data, SKeyboard* pKeyboard) {


thejch:
> hm, it "fixes" it, but it will still coredump, which kinda sucks

Probably something in the cleanup procedure is still causing SIGSEGV or SIGABRT

vaxerski:
usually the case yes

src/managers/input/InputManager.cpp
```diff
@@ -1207,8 +1207,10 @@ void CInputManager::onKeyboardKey(wlr_keyboard_key_event* e, SKeyboard* pKeyboar
         updateKeyboardsLeds(pKeyboard->keyboard);
     }
 
-    if (m_bExitTriggered)
+    if (m_bExitTriggered) {
+        g_pCompositor->m_bShouldSaveCrash = false;
```
vaxerski, line 1211: wouldn't it be better to put this inside cleanup? this isn't the only place that triggers the cleanup.

src/Compositor.cpp
```diff
@@ -29,7 +29,8 @@ void handleUnrecoverableSignal(int sig) {
         return;
     }
 
-    CrashReporter::createAndSaveCrash(sig);
+    if (g_pCompositor->m_bShouldSaveCrash)
```
vaxerski, line 32: honestly this is not a great idea. It'd be better to remove the SEGV/ABRT signal handlers altogether in cleanup() probably

src/managers/input/InputManager.cpp
thejch: The point is only when manually exiting, skip the crash report

src/Compositor.cpp
thejch: Maybe this previous commit is closer to that? https://github.com/hyprwm/Hyprland/pull/4811/commits/530d71c7857e2998a80bd1f06fe4a95adeb17a88
I thought that wasnt good if we still want to receive the signal.

vaxerski: do it inside cleanup and that'd be my idea

src/managers/input/InputManager.cpp
vaxerski: cleanup doesn't get called on crash

src/Compositor.cpp
thejch: thats added. though the core dump still exists probably

PR #4823, opened 2024-02-24
Fix SEGV/ABRT core dump when exiting

thejch:
This tries to fix the core dump issue when exiting, descirbed here: https://github.com/hyprwm/Hyprland/pull/4811#issuecomment-1961278893

Limited testing done but working so far, there might be some uncovered cases still


vaxerski (approved):
makes sense I guess

PR #4837, opened 2024-02-25
compositor: allow source monitor to be provided to `getMonitorInDirec…

github-usr-name:
Add trivial overload for `CCompositor::getMonitorInDirection` which allows the caller to provide a source monitor, which would be useful for some plugin functionality that I'm working on [the plugin needs to directionally-determine the monitor _from an arbitrary window's perspective_]


vaxerski:
sure

vaxerski (approved):

PR #4844, opened 2024-02-26
surface: ensure global pointers valid before using in destructor

github-usr-name:
This fixes an observed SigSegV resulting from the cursor surface using `g_pInputManager` when invoked from the `CInputManager` destructor.  Event chain:

- Hyprland exit dispatcher invoked
- ....
- `g_pInputManager` is destroyed
- `CInputManager::m_sCursorSurfaceInfo` is destroyed
- `cursorSI::wlSurface` is destroyed
- `WLSurface::destroy()` is invoked (from the `WLSurface` destructor)
- `g_pInputManager->m_pLastMouseSurface` is accessed by the `WLSurface`
- `g_pInputManager` is no longer valid, and a SigSegV occurs

Change is trivial, just adds a guard to `g_pInputManager` before accessing.  Also added guards to `g_pCompositor` and `g_pHyprRenderer` in case of similar cycles occurring, though I have not observed this.

### DIFF
--- a/src/helpers/WLSurface.cpp
+++ b/src/helpers/WLSurface.cpp
@@ -83,11 +83,11 @@ void CWLSurface::destroy() {
     m_pWLRSurface->data = nullptr;
     m_pOwner            = nullptr;
 
-    if (g_pCompositor->m_pLastFocus == m_pWLRSurface)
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
lgtm

PR #4851, opened 2024-02-27
Add option to resolve keybinds by sym instead of code

### LINKED ISSUE #1881: Option to apply keybinds after `kb_*` options
My desktop's keyboard has custom firmware for the workman layout, and all my Hyprland keybinds are designed as such. For example:

```
# Move focus
bind = $main, N, movefocus, l
bind = $main, E, movefocus, d
bind = $main, O, movefocus, u
bind = $main, I, movefocus, r
```

On my laptop, however, I'm not so lucky, and I have to set `kb_variant = workman`. Unfortunately this means that, with the same configuration shown above, I need to press the physical NEOI keys rather than the ones mapped by setting the `kb_variant`, which are all over the place instead of on the home row like I want.

Cu3PO42:
#### Description

This commit adds the new configuration option 'resolve_binds_by_sym' which can be set globally or per-device. It is off by default, which preserves the current behavior.

This setting only affects the behavior of keybinds that are defined via key symbols, not those defined via keycode. Binds defined by symbols currently activate if the keycode pressed would generate the specified symbol on the first layout specified in the input section. If enabled, keys pressed on the relevant device will instead match keybinds by the symbols they produce with their current layout.

Closes #1881.

#### Status

This PR is ready to be merged from my end. If you'd prefer a different name for the option I'm happy to rename it.
I have prepared an appropriate update to the documentation in hyprwm/hyprland-wiki#493.

### DIFF
--- a/src/helpers/WLClasses.hpp
+++ b/src/helpers/WLClasses.hpp
@@ -127,15 +127,17 @@ struct SKeyboard {
     bool               active    = false;
     bool               enabled   = true;
 
-    xkb_layout_index_t activeLayout = 0;
+    xkb_layout_index_t activeLayout        = 0;
+    xkb_state*         xkbTranslationState = nullptr;
 
     std::string        name        = "";
     std::string        xkbFilePath = "";
 
     SStringRuleNames   currentRules;
-    int                repeatRate  = 0;
-    int                repeatDelay = 0;
-    int                numlockOn   = -1;
+    int                repeatRate        = 0;
+    int                repeatDelay       = 0;
+    int                numlockOn         = -1;
+    bool               resolveBindsBySym = false;
 
     // For the list lookup
     bool operator==(const SKeyboard& rhs) const {
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -304,7 +304,7 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
 
     const auto         KEYCODE = e->keycode + 8; // Because to xkbcommon it's +8 from libinput
 
-    const xkb_keysym_t keysym         = xkb_state_key_get_one_sym(m_pXKBTranslationState, KEYCODE);
+    const xkb_keysym_t keysym         = xkb_state_key_get_one_sym(pKeyboard->resolveBindsBySym ? pKeyboard->xkbTranslationState : m_pXKBTranslationState, KEYCODE);
     const xkb_keysym_t internalKeysym = xkb_state_key_get_one_sym(wlr_keyboard_from_input_device(pKeyboard->keyboard)->xkb_state, KEYCODE);
 
     if (handleInternalKeybinds(internalKeysym))
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -823,7 +823,8 @@ void CInputManager::applyConfigToKeyboard(SKeyboard* pKeyboard) {
     const auto REPEATRATE  = g_pConfigManager->getDeviceInt(devname, "repeat_rate", "input:repeat_rate");
     const auto REPEATDELAY = g_pConfigManager->getDeviceInt(devname, "repeat_delay", "input:repeat_delay");
 
-    const auto NUMLOCKON = g_pConfigManager->getDeviceInt(devname, "numlock_by_default", "input:numlock_by_default");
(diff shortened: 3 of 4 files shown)


src/helpers/WLClasses.hpp
```diff
@@ -133,9 +133,10 @@ struct SKeyboard {
     std::string        xkbFilePath = "";
 
     SStringRuleNames   currentRules;
-    int                repeatRate  = 0;
-    int                repeatDelay = 0;
-    int                numlockOn   = -1;
+    int                repeatRate        = 0;
+    int                repeatDelay       = 0;
+    int                numlockOn         = -1;
+    bool               resolveBindsBySim = false;
```
vaxerski, line 139: typo?

vaxerski (commented):
rest lgtm

src/helpers/WLClasses.hpp
Cu3PO42: Yup, that one got past me. Fixed it. Thanks for the review!

vaxerski (dismissed):
:car:

src/managers/KeybindManager.cpp
```diff
@@ -304,10 +304,10 @@ bool CKeybindManager::onKeyEvent(wlr_keyboard_key_event* e, SKeyboard* pKeyboard
 
     const auto         KEYCODE = e->keycode + 8; // Because to xkbcommon it's +8 from libinput
 
-    const xkb_keysym_t keysym         = xkb_state_key_get_one_sym(m_pXKBTranslationState, KEYCODE);
-    const xkb_keysym_t internalKeysym = xkb_state_key_get_one_sym(wlr_keyboard_from_input_device(pKeyboard->keyboard)->xkb_state, KEYCODE);
+    const xkb_keysym_t keysymByKeyboard = xkb_state_key_get_one_sym(wlr_keyboard_from_input_device(pKeyboard->keyboard)->xkb_state, KEYCODE);
```
vaxerski, line 307: you can't do that I believe. We had an issue with this back in the day. Some mods alter the syms (eg. SHIFT+1 will now not work, because SHIFT and 1 produces a ! and not 1)


You need to make a copy of the state, initialize it with the keymap and not update the mods.

Cu3PO42: Yeah, you're right. I did think about that at some point while reading xkb docs, but then promptly forgot again when I wrote the code. *puts on dunce hat*

Cu3PO42: I resolved this by creating the equivalent of `m_pXKBTranslationState` for every keyboard whenever a keyboard's layout is changed. This way I don't need to copy it for every key event.

src/managers/input/InputManager.cpp
```diff
@@ -906,6 +908,11 @@ void CInputManager::applyConfigToKeyboard(SKeyboard* pKeyboard) {
         KEYMAP = xkb_keymap_new_from_names(CONTEXT, &rules, XKB_KEYMAP_COMPILE_NO_FLAGS);
     }
 
+    if (pKeyboard->xkbTranslationState) {
+        xkb_state_unref(pKeyboard->xkbTranslationState);
```
vaxerski, line 912: 1: style, no {} around short ifs
2: leak: not destroyed on .destroy

Cu3PO42: Of course I would check to make sure the ref counting was correct for the keymap and then forget about a destructor 🤦

I checked and [xkb_state_unref](https://github.com/xkbcommon/libxkbcommon/blob/58d0bdc743f3b63cfc305ec06ad736072d9b7008/src/state.c#L609) does its own null check, so I just removed the if. FYI clang-format did not remove those braces for me. There's a `RemoveBracesLLVM` rule that might give you the behavior you want.

Thanks again for the review!

vaxerski: RemoveBracesLLVM IIRC removes too much. We only remove around short ifs.

vaxerski (approved):
lgtm thanks

vaxerski:
sehr schöner code digga danke für die geduld :+1:

Cu3PO42:
Danke auch! Bin ja froh, das upstream zu haben.

src/managers/input/InputManager.cpp
Cu3PO42: RemoveBracesLLVM does indeed also remove braces around single statement for and while, maybe others. Just to make sure I'm understandig you right for my next PR: you keep braces around the other control-flow statements?

PR #4863, opened 2024-02-28
keybinds: unconstrain mouse on focusmonitor and cyclenext

JManch:
When playing games, I noticed that switching to my second monitor using `focusmonitor +1` would not move the cursor as I'd expect with `follow_mouse=1`. The cursor would stay on my primary monitor unless I used `movefocus`.

This fixes that and also fixes the same problem for `cyclenext`.

No

Ready


vaxerski (approved):
yea lgtm thanks

PR #4886, opened 2024-02-29
fix libc++/clang build

pg83:
Just fix build


vaxerski:
the fail is unrelated to his code I made an oopsie

vaxerski (approved):

Issue #4650, opened 2024-02-08
Pointer constraints not properly inplemented, windows unable to lock or warp cursor properly


Ghosthree3:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 3d9ca6381df1cdbf1731d7e2b39ea3f7574fce1e dirty (crashreporter: fix logging of function data (4632)).
Date: Wed Feb 7 20:50:23 2024
Tag: v0.35.0-7-g3d9ca638

flags: (if any)


System Information:
System name: Linux
Node name: milkbar
Release: 6.7.4-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Mon, 05 Feb 2024 22:07:49 +0000


GPU information: 
0b:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 10 [Radeon RX 5600 OEM/5600 XT / 5700/5700 XT] [1002:731f] (rev c1) (prog-if 00 [VGA controller])
0c:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU104 [GeForce RTX 2070 SUPER] [10de:1e84] (rev a1) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

An active window/client should be able to lock or warp the cursor within its region, this behavior is buggy at best (warps inaccurately, or not at all) and missing at worst (never locks). This has been a long standing issue, though the behavior has changed several times over the last six months.
I've been too lazy to report it but a recent commit (cbadf3e3) has made the problem significantly worse because mouse up and down events can now cause a window to lose control of the cursor by _briefly_ selecting a different window (windows only have control of the pointer when focused).

[...]

Issue #4348, opened 2024-01-04
resize_on_border does not care about mouse constraints


zakk4223:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 46997a764304366d772456c20b1c719960927aa7 dirty (renderer: fix auto scale detection with fractional).
Date: Sun Dec 31 07:54:24 2023
Tag: v0.33.1-119-g46997a76
```

</details>

### Bug or Regression?

Bug

### Description

Multi-monitor setup: a app/game constraining the mouse on the main monitor (fullscreen), resize_on_border=true in the config.

Sometimes when you click the mouse the game will lose focus. I traced through the source and figured out it is coming from processMouseDownNomal calling resizeWithBorder().

There's a bunch of extra logging in the log file, sorry. Was trying to figure out wtf the focus grab was coming from. Just search for "RESIZE WITH BORDER", that's emited from within processMouseDownNormal

### How to reproduce

resize_on_border+Multi monitor -> run game on main that constrains mouse. Have a window open on the second monitor. "move" the mouse until it would "normally" be on the edge of the second display, and then click+drag. After a few tries you'll eventually trigger a border resize.

### Crash reports, logs, images, videos

[hyprlandd-constrain.log](https://github.com/hyprwm/Hyprland/files/13825973/hyprlandd-constrain.log)

vaxerski:
please check #4889

Issue #3669, opened 2023-10-26
Mouse grab doesn't freeze cursor in place, still interacts with borders


UltraBlackLinux:
### Hyprland Version

0.31.0.r36.6650e4ba-1

### Bug or Regression?

Bug

### Description

In Bitwig Studio there are knobs and faders, and if you turn a knob or move a fader, it grabs the cursor and lets you move the fader, but since a recent update it doesn't work anymore. When I release LMB I can clearly see that the mouse moved.
I can also reproduce a bit of weird behavior when going so far that the mouse goes off the window while having window border resizing enabled: I see the resize cursor and bitwig doesn't ungrab the mouse anymore. A mix of clicking and forcing and unforcing fullscreen gets it to release it again. Very weird behavior.

Bitwig has a free-of-charge 80 day demo, which you can install from the aur.

### How to reproduce

1. Turn on border resizing in hyprland
2. Turn a knob in the plugin view at the bottom and drag downwards so that the cursor goes off the edge of a window
3. You can see the window losing focus and the cursor at the knob turning into the resize cursor
4. Ungrabbing is now very difficult

### Crash reports, logs, images, videos

Github won't upload the log, so here's the bit which should include the part where I caused it:
```
[LOG] Window got assigned a surfaceTreeNode 0
[LOG] Registered signal for owner 56042c2fb220: 56042b4f5110 -> 56042c2fb490 (owner: XWayland Window Late)
[LOG] Registered signal for owner 56042c2fb220: 56042b4f5120 -> 56042c2fb700 (owner: XWayland Window Late)
[LOG] Registered signal for owner 56042c2fb220: 56042b4f5150 -> 56042c2fb3c0 (owner: XWayland Window Late)
[LOG] Registered signal for owner 56042c2fb220: 56042b4f50f0 -> 56042c2fb5c8 (owner: Xwayland Window Late)
[LOG] Registered signal for owner 56042c2fb220: 56042b4f5100 -> 56042c2fb630 (owner: Xwayland Window Late)
[LOG] Registered signal for owner 56042c2fb220: 56042b4f51f0 -> 56042c2fb428 (owner: XWayland Window Late)
[LOG] Registered signal for owner 56042b508820: 56042c2ee320 -> 56042b5088a8 (owner: SurfaceTreeNode)
[LOG] Registered signal for owner 56042b508820: 56042c2ee2f0 -> 56042b508910 (owner: SurfaceTreeNode)
[...]

Issue #4867, opened 2024-02-28
Cursor occasionally fails to lock causing games to lose focus


Sophed:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit f4f3aa2e505ca1b22688fb4a32f31c4bfb851dd0 dirty (layout: add size prediction for initial xdg commits).
Date: Wed Feb 28 11:45:43 2024
Tag: v0.36.0-2-gf4f3aa2e

flags: (if any)


System Information:
System name: Linux
Node name: Arc
Release: 6.7.6-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 23 Feb 2024 16:32:48 +0000


GPU information: 
08:00.0 VGA compatible controller [0300]: NVIDIA Corporation GP106 [GeForce GTX 1060 6GB] [10de:1c03] (rev a1) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:


```

</details>

### Bug or Regression?

Bug

### Description

Mainly noticeable when playing a game, I've found that if I use my keybind of SUPER + arrow key to switch focus to my game window, the cursor does not get locked. This means that looking around in game will cause my cursor to leave the window boundaries, unfocusing it and causing the game to pause or stop receiving input.

### How to reproduce

- Set a keybind for switching window focus (I'm using SUPER + arrow keys)
- Open a game (in my case I'm using Minecraft, I haven't had the chance to test if any other games have the same issue)
- Focus a different window (Discord, browser whatever)
- Use the keybind to refocus the game
- Attempt to move the camera using the mouse

I have tried to record a video showcasing the issue but it may not be super clear due to my bad OBS setup

### Crash reports, logs, images, videos

https://github.com/hyprwm/Hyprland/assets/65602967/8649abec-c0cb-492d-840c-982de41af727

[...]

PR #4889, opened 2024-02-29
Rewritten pointer constraints

vaxerski:
This PR rewrites constraints from the ground-up, cleans the code, yadda yadda yadda.

Possibly fixes, so CC for testing: (if your issue is in here, please report with a comment either "my issue #XYZ was fixed" or "my issue #XYZ was not fixed")

#4867 #4650 #4465 #4348 #3669 #2376 #1732 #3672 #3270

cc @fufexan for review


aanjkyks:
this helps with part of #4650, which affects me, that is - hyprland switching focus to another window on click of a "phantom pointer". tested on Arch with looking glass

mouse desync is still an issue. it seems as if hyprland remembers the pointer position at the time of "handing over" control of the pointer over to looking glass, and sets pointer back there when looking glass "releases" control of the pointer. expected behavior would be that both pointers are in sync, but that might be too application-specific. overall, this is an improvement

also, now, with input capture enabled in looking glass, keyboard input is also fully passed to the window, instead of hyprland reacting to keybinds from the config first, which seems to be a "more correct", but still, a change in behavior. for example, i used to be able to switch between workspaces without toggling input capture in looking glass

vaxerski:
> mouse desync is still an issue. it seems as if hyprland remembers the pointer position at the time of "handing over" control of the pointer over to looking glass, and sets pointer back there when looking glass "releases" control of the pointer. expected behavior would be that both pointers are in sync, but that might be too application-specific. overall, this is an improvement

that is true, but it should warp the cursor to the set region in the meantime... :thinking:

vaxerski:
how about now?

aanjkyks:
i do not notice any change, other than 6d82f07 leading to crashes, which seems to have been fixed by d2dc4b5. just in case, attaching a crash report
[hyprlandCrashReport21578.txt](https://github.com/hyprwm/Hyprland/files/14451450/hyprlandCrashReport21578.txt)

vaxerski:
no, those crashes are still on this MR, they've been fixed in head. Let me rebase

UltraBlackLinux:
unfortunately, this commit does not fix #3669, and instead reintroduces #4186. Furthermore, with my use case the cursor seems to jump to a seemingly random place after the ~4th immediate unfocus

Roger-Roger-debug:
I did a little bit of testing now, but will have to look closer tomorrow.

Regarding #3672: in FF the camera flick happened only once so far (might've been related to a steam notification I got at that moment?). It rarely happened on wallpaper change as well, but I didn't spend enough time playing to tell if that has changed.

This change also changes the behavior of #4497. It now doesn't happen on every wallpaper change but still happens sometimes. I think unlike before it now snaps to the same position each time.

vaxerski:
both of you check now

vaxerski:
@aanjkyks can you try the desync issue with this?
[patch.txt](https://github.com/hyprwm/Hyprland/files/14454149/patch.txt)

zakk4223:
Looks like it fixes #4348

rurigk:
@Roger-Roger-debug Do you have any other game that have this issue #3672 
A long time I had this issue with Apex legends but i assumed it was fixed a very long time ago

earboxer:
Doesn't fix my issue "where I can't type (with the virtual keyboard) on a layer surface when I don't have an open window beneath it.". Also, looks like exclusive keyboard focus can now be taken away by touching a different surface, but it looks like a bug with (bemenu) greedy pointer grabbing is re-resolved.

rurigk:
Im not sure if this is related to this but i get a very weird interactions when draging a window on games with gamescope
`gamescope -w 1920 -h 1080 -W 3840 -H 2160 -e -- %command%`

https://github.com/hyprwm/Hyprland/assets/6665402/6178288b-c6af-4e5f-b4ab-a12f9789a2dc

vaxerski:
@Roger-Roger-debug can you try the patch I posted here https://github.com/hyprwm/Hyprland/pull/4889#issuecomment-1972072597

UltraBlackLinux:
> both of you check now

#3669 is fixed, awesome!

Roger-Roger-debug:
> @Roger-Roger-debug can you try the patch I posted here [#4889 (comment)](https://github.com/hyprwm/Hyprland/pull/4889#issuecomment-1972072597)

That seems to have fixed all my Borderlands/Gamescope issues, spent a few minutes spamming new wallpapers but no flicking. I'm not sure about FF yet as I have no clear way to reproduce it there and haven't played enough.

vaxerski:
committed

Ghosthree3:
This mostly fixes #4650, at least in terms of the cursor remaining captured when being captured. The local cursor now _appears_ to track the application cursor accurately, and is aggressively warped to maintain that if the application is trying to 'hold' it, like in some video games (like seen in the 4th video of that issue).

There is one issue though, the true location of the cursor seems to be locked somewhere near where you entered the window (how far away from the border is somewhat random, I assume however far you managed to move the cursor before the application window, or Hyprland, I'm not sure, 'noticed' you were inside). When the application attempts to lock the cursor it can be seen to warp there instead of being frozen where it was when the lock occurred as it should (as shown in the sway demos in the mentioned issue). It also tries to move but is warped repeatedly back the same way it does when NOT locked but being held still by a game like previously mentioned, while it shouldn't be able to move at all. This wouldn't be the end of the world but for the next part.

Despite the local cursor appearing to be in sync with the application while within the application, when exiting the window the cursor is warped to that same position along the edge where you entered, causing you to have to exit twice. This is especially noticeable if you're attempting to exit the window from the opposite side that you entered from. For example, if I enter the window with the cursor from the left, then want to exit to the right, as I reach the right side edge the local cursor will warp all the way back to the left and have to be dragged all the way across the window for a second time (but this time out of sync with the application cursor) to actually exit. If exiting from the same side that I entered from the same thing occurs but the distance is significantly less.

If I had to guess, the issue is that when the window is not trying to warp the mouse anymore the cursor is reset to its entry position rather than to the last position it was moved/warped to by the window. It probably shouldn't have any 'memory' of that at all, it should just continue to be moved until it isn't.

See video below for a demo, reminder that the dot is the local/Hyprland cursor position.

https://github.com/hyprwm/Hyprland/assets/6110140/92b04e28-8c92-47ea-a27c-79be0c8dd309

This is still an overall improvement in that while leaving the window is a little strange now, at least you don't accidentally leave when you don't want to.

vaxerski:
try now @Ghosthree3

Ghosthree3:
```
Hyprland, built from branch rewrite-constraints at commit ed3842a81bb524f7ffc51361b1de4c9c4161899a  (update pos hint on region change).
```

Behavior seems identical to the above post.

vaxerski:
dam

alright but since it solves so many issues I'll merge it, seems like it's in a good state

Thanks all, further issues shall be dealt with in the futrue

pdf:
@vaxerski I just ran through a git bisect to find that d72ea5f2a7fdb0d0a7bf914412327195b05199b0 is what's causing the mouse cursor to be displaced by a random significant offset in fullscreen games, making it impossible to click on things reliably.

zakk4223:
> @vaxerski I just ran through a git bisect to find that [d72ea5f](https://github.com/hyprwm/Hyprland/commit/d72ea5f2a7fdb0d0a7bf914412327195b05199b0) is what's causing the mouse cursor to be displaced by a random significant offset in fullscreen games, making it impossible to click on things reliably.

Might need a bit more info; I've been playing fullscreen games with this commit and I'm not getting displaced mouse issues.

How many displays, which game(s) etc?

pdf:
@zakk4223 single monitor at 0x0.

Various games, Balatro is what I've been using as a repro because it's consistent and fast to start up, but also happens on League randomly, and those are the only two I've tried until now. I just tried a couple of others and they didn't produce the bug.

Sophed:
gotcha, did everything and it looks like i'm still having the same issues :/

tchofy:
You could have multiple hyprland installs on your system, try ``whereis hyprland`` and see if there's only one binary. If you've ever built hyprland manually then you probably want to nuke the manual install.

Sophed:
nope, i only have one binary on my system

pdf:
> Various games, Balatro is what I've been using as a repro because it's consistent and fast to start up, but also happens on League randomly, and those are the only two I've tried until now. I just tried a couple of others and they didn't produce the bug.

Also happens in-game in Dota 2 (initial client UI is fine, starting a game/arcade game results in offset cursor).

pdf:
> @vaxerski I just ran through a git bisect to find that [d72ea5f](https://github.com/hyprwm/Hyprland/commit/d72ea5f2a7fdb0d0a7bf914412327195b05199b0) is what's causing the mouse cursor to be displaced by a random significant offset in fullscreen games, making it impossible to click on things reliably.

Just FYI, I can no longer reproduce as of 66330281 so I would guess something between d72ea5f and now has resolved this.

Ghosthree3:
> > @vaxerski I just ran through a git bisect to find that [d72ea5f](https://github.com/hyprwm/Hyprland/commit/d72ea5f2a7fdb0d0a7bf914412327195b05199b0) is what's causing the mouse cursor to be displaced by a random significant offset in fullscreen games, making it impossible to click on things reliably.
> 
> Just FYI, I can no longer reproduce as of [6633028](https://github.com/hyprwm/Hyprland/commit/66330281ff19e84c986c5ce639b670929e5dadd6) so I would guess something between [d72ea5f](https://github.com/hyprwm/Hyprland/commit/d72ea5f2a7fdb0d0a7bf914412327195b05199b0) and now has resolved this.

Most likely 024d4dd

Ghosthree3:
This (https://github.com/hyprwm/Hyprland/pull/4889#issuecomment-1974666895) issue is now fixed and Hyprland's behavior is almost identical to Sway's. The only difference I can see is that during capture/lock mode the cursor technically isn't locked/frozen (which it should be), but is repeatedly warped back to the position it should be locked at, so to the user it is effectively the same.
I have only just now updated to master (bcba395) from 6c4e248 so am not really sure, but I suspect #5056 is what fixed it.

I will close #4650 as all the issues I raised there now seem to be resolved.

Fleimi:
https://github.com/hyprwm/Hyprland/issues/3672 is not fixed yet. In FFXIV in particular there used to be a "mouse spazzes out upon focusing the window" bug, but now there is also a bug where the camera is just randomly pointed to a random direction, especially on mouseclick (?). 

Is there something you'd need help with in order to diagnose this? For context, I'm spamming right click to take control of the camera and every so often when doing so the camera is pointed down and to one side, seemingly randomly.

https://github.com/hyprwm/Hyprland/assets/73422514/4008fb1d-2485-412f-8b06-3c20b9a82525

PR #4911, opened 2024-03-01
animation: Refactor AnimatedVariable

GartoxFR:

This commit decomposes the AnimatedVariable class into a base class with the common attribute to all variable types and a templated derived type containing strongly typed info on the type being animated.

Access to the typed version is perfomed using the visitor pattern. A utility is provided to build a visitor on the fly using lambdas.

Adding a new type to be animated should just be a matter of adding the typed in the list defined by the ANIMABLE_TYPES macro

The size of the commit is justified by the API change in the AnimatedVariable class. No more vec(), fl() or col() method but a unified value() method.

This PR reduces code duplication in the AnimatedVariable class and in the AnimationManager. Moreover, it improves it's extensibility to other types.

The tradeoff is a bit more complexity with all the templated arguments in AnimatedVariable.cpp but less code duplication and smaller memory footprint for these variables.

Note that the added complexity does't leak to much on the code that uses these variables.    

This can break compatibility as it changes the way to declare and use AnimatedVariables.

It's ready to merge. I'm running this branch to test it while I'm redaction this PR.

### DIFF
--- a/src/helpers/AnimatedVariable.hpp
+++ b/src/helpers/AnimatedVariable.hpp
@@ -3,6 +3,7 @@
 #include <functional>
 #include <any>
 #include <chrono>
+#include <type_traits>
 #include "Vector2D.hpp"
 #include "Color.hpp"
 #include "../macros.hpp"
@@ -14,6 +15,23 @@ enum ANIMATEDVARTYPE {
     AVARTYPE_VECTOR,
     AVARTYPE_COLOR
 };
+
+// Utility to bind a type with its corresponding ANIMATEDVARTYPE
+template <class T>
+struct typeToANIMATEDVARTYPE_t { static constexpr ANIMATEDVARTYPE value = AVARTYPE_INVALID; };
+
+template<>
+struct typeToANIMATEDVARTYPE_t<float> { static constexpr ANIMATEDVARTYPE value = AVARTYPE_FLOAT; };
+
+template<>
+struct typeToANIMATEDVARTYPE_t<Vector2D> { static constexpr ANIMATEDVARTYPE value = AVARTYPE_VECTOR; };
+
+template<>
+struct typeToANIMATEDVARTYPE_t<CColor> { static constexpr ANIMATEDVARTYPE value = AVARTYPE_COLOR; };
+
+template <class T>
+inline constexpr ANIMATEDVARTYPE typeToANIMATEDVARTYPE = typeToANIMATEDVARTYPE_t<T>::value;
+
 
 enum AVARDAMAGEPOLICY {
     AVARDAMAGE_NONE   = -1,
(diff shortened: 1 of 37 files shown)


vaxerski (commented):
A few comments:
 - why split the Base out? Is there any benefit to having it separate?
 - what the fuck is this: https://github.com/GartoxFR/Hyprland/blob/dccd0f5c1cf17e80673172b83896e9f914b35609/src/helpers/AnimatedVariable.hpp#L33-L83
 
Visitors are one of the most confusing design patterns I've ever seen in my life and I have yet to understand what they are, what they do, or what they're for.

To me, all the code you wrote in the linked snippet (L33-L83) looks like useless garbage

GartoxFR:
> A few comments:
> 
>     * why split the Base out? Is there any benefit to having it separate?
> 
>     * what the fuck is this: https://github.com/GartoxFR/Hyprland/blob/dccd0f5c1cf17e80673172b83896e9f914b35609/src/helpers/AnimatedVariable.hpp#L33-L83
> 
> 
> Visitors are one of the most confusing design patterns I've ever seen in my life and I have yet to understand what they are, what they do, or what they're for.
> 
> To me, all the code you wrote in the linked snippet (L33-L83) looks like useless garbage

- The base class exists so we can store `CBaseAnimatedVariable*` in the `AnimationManager`. And  it's also easier to implement the common methods in non-templated class. The benefit of separating then is to isolate common code under the base class but still consider `AnimatedVariable<float>` and `AnimatedVariable<Vector2D>` as different types at compile time and not at runtime.
- The snippet of code you linked defines which types can be animated (Vector2D, float and CColor). Then it defines a visitor class that have an empty visit method for each possible types. Then the CAnimatedVisitorOverload allows construction of the visitor with lambdas. All this code is used to safely down cast `CBaseAnimatedVariable*` into `CAnimatedVariable<T>&` in this snippet : https://github.com/GartoxFR/Hyprland/blob/dccd0f5c1cf17e80673172b83896e9f914b35609/src/managers/AnimationManager.cpp#L116-L143

In summary, the visitor pattern is a bit of a glorified switch statement that uses virtual functions to downcast pointers when needed.

vaxerski:
doesn't std have `std::visit`? 

Why isn't that used?

Generally the visit-related code is 100% unreadable to me, and thus I can't accept a pattern like this as I can't maintain it later. I'd rather have just the template and a switch later without the entire visitor stuff.

GartoxFR:
> A few comments:
> 
>     * why split the Base out? Is there any benefit to having it separate?
> 
>     * what the fuck is this: https://github.com/GartoxFR/Hyprland/blob/dccd0f5c1cf17e80673172b83896e9f914b35609/src/helpers/AnimatedVariable.hpp#L33-L83
> 
> 
> Visitors are one of the most confusing design patterns I've ever seen in my life and I have yet to understand what they are, what they do, or what they're for.
> 
> To me, all the code you wrote in the linked snippet (L33-L83) looks like useless garbage

- The base class exists so we can store `CBaseAnimatedVariable*` in the `AnimationManager`. And  it's also easier to implement the common methods in non-templated class. The benefit of separating then is to isolate common code under the base class but still consider `AnimatedVariable<float>` and `AnimatedVariable<Vector2D>` as different types at compile time and not at runtime.
- The snippet of code you linked defines which types can be animated (Vector2D, float and CColor). Then it defines a visitor class that have an empty visit method for each possible types. Then the CAnimatedVisitorOverload allows construction of the visitor with lambdas. All this code is used to safely down cast `CBaseAnimatedVariable*` into `CAnimatedVariable<T>&` in this snippet : https://github.com/GartoxFR/Hyprland/blob/dccd0f5c1cf17e80673172b83896e9f914b35609/src/managers/AnimationManager.cpp#L116-L143

In summary, the visitor pattern is a bit of a glorified switch statement that uses virtual functions to downcast pointers when needed.

> doesn't std have `std::visit`?
> 
> Why isn't that used?
> 
> Generally the visit-related code is 100% unreadable to me, and thus I can't accept a pattern like this as I can't maintain it later. I'd rather have just the template and a switch later without the entire visitor stuff.

`std::visit` is used for variant so it's a bit different as it doesn't use any Base class.

I proposed this pattern because I find it more readable than a switch as it's closer to pattern marching I can find in other languages but this is just a personal opinion. 

I can replace the visitor pattern by adding back an enum variable to tell at runtime which type to cast to. It's basically a matter of using this kind of variables or v-tables to tell which type we have.

vaxerski:
> I can replace the visitor pattern by adding back an enum variable to tell at runtime which type to cast to. It's basically a matter of using this kind of variables or v-tables to tell which type we have.

please do. Much more readable

GartoxFR:
> please do. Much more readable

Done. This needed some template to associate the types with the enum constants. I hope it's not to unreadable

src/helpers/AnimatedVariable.hpp
```diff
@@ -314,3 +185,86 @@ class CAnimatedVariable {
...
     friend class CHyprRenderer;
 };
+
+template <Animable VarType>
+class CAnimatedVariable : public CBaseAnimatedVariable {
+  public:
+    CAnimatedVariable() : CBaseAnimatedVariable(typeToANIMATEDVARTYPE<VarType>) {} // dummy var
+
+    void create(const VarType& value, SAnimationPropertyConfig* pAnimConfig, void* pWindow, AVARDAMAGEPOLICY policy) {
+
```
vaxerski, line 195: useless empty line

src/helpers/AnimatedVariable.hpp
```diff
@@ -28,175 +46,34 @@ struct SLayerSurface;
...
+    void         unregister();
+    void         registerVar();
 
-        if (endCallback)
-            onAnimationEnd();
-    }
+    virtual void warp(bool endCallback = true) = 0;
 
-    void setConfig(SAnimationPropertyConfig* pConfig) {
+    void         setConfig(SAnimationPropertyConfig* pConfig) {
```
vaxerski, line 76: could you add an empty comment above to avoid the alignment? `//`

vaxerski (commented):
minor style nits and this lgtm, much better! Thanks :)

vaxerski (dismissed):
lgtm thanks :)

vaxerski:
oops seems like clang-format is unhappy, can you slap the mr with it?

GartoxFR:
Of course, editing this much files I forgot pressing the format button on some of them. Should be fixed now

vaxerski (approved):

Issue #3638, opened 2023-10-23
Option to make all keys no-op in submaps unless explicitly bound


spikespaz:
### Description

For example:

```
submap=resize,noop
# binds...
submap=reset
```

Here, `noop` is added as an argument to `submap`. This means that key presses will do nothing unless they are bound in the `resize` submap.

This is useful if someone is borrowing a computer and they want to mash keys to figure out what works. They don't need to delete the random text they accidentally typed in.

spikespaz:
I also want `rebind` where binds from the default submap are used unless overridden.

So, `submap=resize,noop,remap` would mean "normal key presses are not sent to the focused window, bindings from the default submap can be used unless the `resize` submap replaces them".

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

Issue #3405, opened 2023-09-23
special "exit-sub-on-exec" bind for submaps and a wildcard key 


torwag:
### Description

Related to #1972 I tried to get the same sort of keybinding I am used in sway.
1. I enter a submap with a certain keybind
2. I press a certain sub-bind
3. If this is defined execute it and exit the submap
4. If it doesn't exist, do not pass it as keypress but ignore the command and exit the submap

Right now, pressing the wrong key by accident in a terminal gets you some glyphs on the terminal.
Exiting the submap after an successful execution requires to define each key twice once for the acutal action and once to exit the submap.

I would like to propose two possible solutions

1. Define a bind for submaps which executes the action and exist the current submap
That alone would make it unnecessary to define an exit and might help a lot of people not to get locked in a submap.

2. Create a special ALL_KEYS / wildcard key which basically maps to all and every character. 
Creating at the very final end of a submap a binding with such a wildcard-key would allow to exit the submap without action and would avoid to pass any "wrong" entry as a key-press to the active program.

Both proposals would be backward compatible and do not change the submap mechanics, thus it could be implemented without conflicting with current configs.

Issue #1972, opened 2023-04-07
Submaps are a pain


LordTlasT:
This is my opinion and I am missing something very simple, but scripting, sockets wouldn't help my problem.
If there is a way of achieving what I want please help me. I hope it's not in the wiki.

When I want to bind a submap I have to do this:

```ini
1. bind = $mainMod, M, submap, menuscripts
2. submap = menuscripts
3. bind = $mainMod, A, exec, dmapimg
4. bind = , A, exec, dmapimg
5. bind = $mainMod, A, submap, reset
6. bind = , A, submap, reset
7. submap = reset
```
I would want the behavior of submaps to be the following

1. submap key is pressed
2. a next key (combination) is pressed
-> only the recently pressed key (combination) should be checked, this would eliminate line 3
3a. If the key matches a correct bind, execute it  
3b. If not then exit the keymap

I would like to get 3b implemented, because it's really easy to get trapped in a submap group.
And it's very annoying even when you can reset it.

Thank you for making hyprland, this has been very fun and I am happy using a wonderful environment on my computer. :)
I don't how much of what I want can be achieved, but I wanted to ask.

vaxerski:
there is currently no way to make a "fallback" on a submap, so the only way I can see it done atm is just binding every key from a-z that isnt used to `submap, reset`

LordTlasT:
This is also the only way I see, so I seperated the binds in a seperate file.
Thanks for the response. I hope for your success!

LordTlasT:
That's a nice way to keep your submap groups readable. Will this slow down hyprland?

11xx:
> Will this slow down hyprland?

I'm not familiar with the internals but it can be safe to assume it won't since when Hyprland reloads the config file it should load every reference to all imports in memory, and since it executes keybinds in a linear way just like the rest of the main config file, so when there's a submap change it just skips to the specified scope [of the 'reset' submap in this case].

LordTlasT:
I still don't know, it seems elegant and inelegant at the same time.

PR #4930, opened 2024-03-02
Add the 'catchall' keyword that matches all keys

Cu3PO42:
## Description

This keyword can be used to define arbitrary keybinds. The only special behavior that it exhibits is that it matches every key, including modifier keys. Any flags still apply normally.

## Motivation

I use submaps to create modal keymaps similar to those implemented by many Vim configs, e.g. [LazyVim](http://www.lazyvim.org/keymaps). Then I might hit `<leader> l m` to launch my system monitor. While in a submap I would like to ignore any keys that are not mapped and open a help window that reminds me of my forgotten binds. Adding a catch-all bind seemed to be the most straight-forward way to achieve that.

This or a similar feature is also requsted by #3638, #3405 and #1972.

## Fixed Issues

This commit also fixes an issue that keys bound via the `code:KEYCODE` format were not unbound correctly. This problem stemmed from a duplication of the code that translates the key specification in a bind rule (i.e. its second argument) to keysym/keycode. Since I had to touch that code anyway and to avoid similar problems in the future, I extracted that code into a function.

### Syntax 

The keyword could theoretically collide with a keysym. However, I believe it to be unlikely that a keysym with such a name will ever exist, so this is a limited risk. To be extra sure or future proof for new keywords, it is conceivable to instead use syntax such as `special:catchall`. The current `code:` syntax for binding by keycode would be precedent for that.

I am also not 100% sold on the formatting by hyprctl or more specifically the casing of the name. I tried to remain consistent with the existing fields, specifically `non_consuming` being snake_case in the JSON output and everything being lowercase in the other output.

### Footgun

Using this keyword outside a submap will prevent essentially all keys from being forwarded to applications (unless defined as non-consuming). This might make it difficult to revert the setting. The setting could be made to only work in a submap to mitigate that.

Since switching to a TTY should still work, I don't view this as a huge problem. I would rather remain consistent with existing behavior and introduce fewer special cases.

## Documentation

I have prepared an accompanying PR at hyprwm/Hyprland-wiki#499.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1732,6 +1732,17 @@ std::optional<std::string> CConfigManager::handleAnimation(const std::string& co
     return {};
 }
 
+SParsedKey parseKey(const std::string& key) {
+    if (isNumber(key) && std::stoi(key) > 9)
+        return {.keycode = std::stoi(key)};
+    else if (key.starts_with("code:") && isNumber(key.substr(5)))
+        return {.keycode = std::stoi(key.substr(5))};
+    else if (key == "catchall")
+        return {.catchAll = true};
+    else
+        return {.key = key};
+}
+
 std::optional<std::string> CConfigManager::handleBind(const std::string& command, const std::string& value) {
     // example:
     // bind[fl]=SUPER,G,exec,dmenu_run <args>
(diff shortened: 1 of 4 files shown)


vaxerski:
the warning in the wiki is relatable - how about ignore catchall outside submaps?

Cu3PO42:
That's definitely an option. In fact, that is what I intended to propose under the 'Footgun' section above. Since it is known at parse-time whether the bind would be in a submap, maybe it could even be a validation error instead of just ignoring it?

vaxerski (approved):
idk looks alright

Cu3PO42:
Thank you for reviewing and merging that so quickly!

vaxerski:
:mountain:  :car:

PR #4931, opened 2024-03-02
events: update render data after workspace window rule

JManch:
This fixes workspace rules not applying to windows that match both a `float` window rule and a `workspace` window rule.

The problem was that `updateSpecialRenderData` got called before the `workspace` rule was applied (so the workspace ID was outdated). This wasn't a problem for non-floating windows because the dwindle/master layout code calls `updateSpecialRenderData` again when the window tiles.

No

Ready


vaxerski (approved):
thanks!

I am a baller

PR #4944, opened 2024-03-03
compositor: ignore grab extend behind special workspaces

JManch:
Fixes this problem where, when crossing monitor borders, special workspaces would automatically close in a jarring way if the gap between a floating window and the monitor edge was less than `extend_border_grab_area`:

https://github.com/hyprwm/Hyprland/assets/61563764/22d67519-3237-4a28-95cc-e3e04cd7be5b

It's only an issue when `special_fallthrough` is disabled and `extend_border_grab_area > 0`.

It's ready to merge, but I'm not sure if it's the best solution. Is there a cleaner way to check if a window is fully contained inside its monitor?


vaxerski (approved):
alright

PR #4952, opened 2024-03-03
Better handling of workspace_back_and_forth when using movefocustocurrentmonitor dispatcher

tomben13:
`workspace_back_and_forth` behaviour was a little wonky when using `movefocustocurrentmonitor`. If the previous workspace was active on another monitor, it would switch the focus to the other monitor, which is indesirable for `movefocustocurrentmonitor`.

I added checks in `CKeybindManager::focusWorkspaceOnCurrentMonitor` such that if `workspace_back_and_forth` option is set and the focus request is on the current active workspace, it replaces the `pWorkspace` variable so the rest of the code treats like the focus request is on the previous workspace.

I may require additional testing. It works with my setup, but I couldn't wrap my head around all the edge cases in other setups than mine.


tomben13:
Commit message is wrong, I have to change it

src/Compositor.cpp
```diff
@@ -2150,7 +2153,7 @@ void CCompositor::moveWorkspaceToMonitor(CWorkspace* pWorkspace, CMonitor* pMoni
 
     // fix old mon
     int nextWorkspaceOnMonitorID = -1;
-    if (!SWITCHINGISACTIVE || !POLDMON)
+    if (!SWITCHINGISACTIVE)
```
vaxerski, line 2156: why is this check removed

src/managers/KeybindManager.cpp
```diff
@@ -1560,30 +1560,43 @@ void CKeybindManager::focusWorkspaceOnCurrentMonitor(std::string args) {
...
+        pWorkspace = g_pCompositor->createNewWorkspace(workspaceID, PCURRMONITOR->ID);
         // we can skip the moving, since it's already on the current monitor
-        changeworkspace(PWORKSPACE->getConfigName());
+        changeworkspace(pWorkspace->getConfigName());
         return;
     }
 
-    if (PWORKSPACE->m_iMonitorID != PCURRMONITOR->ID) {
-        const auto POLDMONITOR = g_pCompositor->getMonitorFromID(PWORKSPACE->m_iMonitorID);
+    static auto* const PBACKANDFORTH = (Hyprlang::INT* const*)g_pConfigManager->getConfigValuePtr("binds:workspace_back_and_forth");
```
vaxerski, line 1572: invalid config acquisition

```
static auto PBACKANDFORTH = CConfigValue<Hyprlang::INT>("binds:workspace_back_and_forth");
```

src/Compositor.cpp
tomben13: `const bool SWITCHINGISACTIVE = POLDMON ? POLDMON->activeWorkspace == pWorkspace->m_iID : false;`

`POLDMON` can never be `false` when `SWITCHINGISACTIVE` is `true` due to the definition of `SWITCHINGISACTIVE`. So the check is redundant.

Should I open a separate PR for this change ?

src/managers/KeybindManager.cpp
tomben13: Fixed. I copied code from some previous version.

src/Compositor.cpp
vaxerski: right my bad

vaxerski (approved):
lgtm

Issue #2457, opened 2023-06-07
Escape commas in monitor config lines


adlrwbr:
### Discussed in https://github.com/hyprwm/Hyprland/discussions/2130

<div type='discussions-op-text'>

<sup>Originally posted by **Hubro** April 22, 2023</sup>
I want to reuse my hyprland config across multiple machines, so I can't configure ports like DP-1 directly, I need to use monitor descriptions.

Unfortunately, one of my monitors has a comma in the description:

```
Monitor DP-2 (ID 0):
	3840x2160@60.000000 at 0x0
	description: GIGA-BYTE TECHNOLOGY CO., LTD. M28U **** (DP-2)
	make: GIGA-BYTE TECHNOLOGY CO., LTD.
	model: M28U
	serial: ****
	active workspace: 9 (9)
	reserved: 0 0 0 28
	scale: 1.50
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: 1
```

If I try to use this description unchanged, I get a syntax error:

```
desc:GIGA-BYTE TECHNOLOGY CO., LTD. M28U 22120B003932
```

I have tried every method to escape the comma I can think of (double comma, backslash escape, deleting it altogether, quoting the entire desc with double and single quotes) but nothing works.

How do I escape this comma so I can target the correct monitor?</div>

OahzEgroeg:
Maybe can make regex working with monitor desc in config?

fzechert:
I have the same problem. If it helps you, here is my output of `hyprctl monitors` for the monitor in question (also a GIGABYTE):

```
Monitor DP-9 (ID 3):
	3440x1440@59.97300 at 0x1080
	description: GIGA-BYTE TECHNOLOGY CO., LTD. M34WQ 0x00000B52 (DP-9)
	make: GIGA-BYTE TECHNOLOGY CO., LTD.
	model: M34WQ
	serial: 0x00000B52
	active workspace: 4 (4)
	special workspace: 0 ()
	reserved: 0 0 0 0
	scale: 1.00
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: 0
```

In my case I don't need to share the configuration file across multiple devices. But my three monitors are connected over thunderbolt and a DisplayPort Switch. Depending on the order the monitors are detected during boot, the port name `DP-9` changes.

This makes addressing the monitors via the port name difficult.

[...]

PR #4970, opened 2024-03-05
monitor: remove commas from short description

MikeWalrus:
Since `hyprctl monitor` shows szShortDescription now, it needs to be sanitized. (See #2457)

Also, monitor selectors are now compared against szShortDescription and szDescription to avoid re-striping the string.


vaxerski (approved):
lgtm

PR #4981, opened 2024-03-05
Add custom fontsize support for notifications

### LINKED ISSUE #4866: Increase hyprland notification font size
### Description

I'd love it if it were possible to increase the font size of the notifications created via the `hyprctl notify` command.

ItsDrike:

This adds support for specifying a custom font size the notification will be rendered with.

Fixes #4866

image

This is technically a breaking change for anyone who is currently using `hyprctl notify`, as it requires an extra argument for the font size.

Note that I'm not entirely sure how the font size calculation worked in the original code here:

```cpp
        const auto            FONTSIZE        = std::clamp((int)(notif->fontSize * ((pMonitor->vecPixelSize.x * SCALE) / 1920.f)), 8, 40);
```

I pretty much just reused this code and used `notif->fontSize`, instead of the hard-coded value. I think this is the correct way to do it, but I'm not certain of it.

Ready.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1422,7 +1422,7 @@ std::string dispatchNotify(eHyprCtlOutputFormat format, std::string request) {
 
     CColor color = configStringToInt(vars[3]);
 
-    int    msgidx   = 4;
+    size_t msgidx   = 4;
     float  fontsize = 13.f;
     if (vars[msgidx].length() > 9 && vars[msgidx].compare(0, 10, "fontsize:")) {
         const auto FONTSIZE = vars[msgidx].substr(9);
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1440,13 +1440,9 @@ std::string dispatchNotify(eHyprCtlOutputFormat format, std::string request) {
     if (vars.size() <= msgidx)
         return "not enough args";
 
-    std::string message = "";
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
this is breaking, though. would be better not to be

ItsDrike:
Yeah, I agree, but I'm not sure how to achieve that when all of the arguments after 5th were just treated as parts of the msg. Any suggestions?

vaxerski:
why not make this just a config variable

ItsDrike:
I considered that, but I figured the issue author wanted to be able to set different sizes for different notifs. If a single size is enough though, I can modify this

vaxerski:
I am for specifying args at the beginning of the message then, like:
`fontsize:17 Hello Hyprland!`

src/debug/HyprCtl.cpp
```diff
@@ -1420,17 +1420,33 @@ std::string dispatchNotify(eHyprCtlOutputFormat format, std::string request) {
...
+        } catch (std::exception& e) { return "invalid fontsize karg"; }
+
+        ++msgidx;
     }
 
+    if (vars.size() <= msgidx)
+        return "not enough args";
+
+    std::string message = "";
+    for (size_t i = msgidx; i < vars.size(); ++i) {
```
vaxerski, line 1444: `.join` exists.

ItsDrike: k fixed (though I was just reusing your code there)

PR #4982, opened 2024-03-05
layout: warp the cursor when focusing windows

MikeWalrus:
Similar to the `focuswindow` dispatcher, when focusing a window with wlr-foreign-toplevel-management, the cursor should be warped. Otherwise, the focus is lost immediately after the cursor moves.


vaxerski (approved):
oki

PR #4989, opened 2024-03-06
Fix logic of "on" and "off"  for binding of switch

Hayak3:
according to [wiki](https://wiki.hyprland.org/Configuring/Binds/#switches)
```
# trigger when the switch is turning on
bindl=,switch:on:[switch name],exec,hyprctl keyword monitor "eDP-1, 2560x1600, 0x0, 1"
# trigger when the switch is turning off
bindl=,switch:off:[switch name],exec,hyprctl keyword monitor "eDP-1, disable"

```
when you turn off the lid,it trigger switch:off,but it behavior opposite now.
It is ready for merging

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1467,11 +1467,11 @@ void CInputManager::newSwitch(wlr_input_device* pDevice) {
 
             switch (E->switch_state) {
                 case WLR_SWITCH_STATE_OFF:
-                    Debug::log(LOG, "Switch {} turn on, triggering binds.", NAME);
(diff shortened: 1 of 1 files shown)


Hayak3:
> that's a problem with your hardware or kernel

according to the [libinput](https://wayland.freedesktop.org/libinput/doc/1.8.1/group__device.html#ga507e97278f2eb7d2271023ef2a3d31a4),lid is close when WLR_SWITCH_STATE_ON, so maybe hyprland [wiki](https://wiki.hyprland.org/Configuring/Binds/#switches) is wrong.in this config,eDP-1 would be open when lid is closed.

vaxerski:
feel free to make a mr to the wiki

PR #5009, opened 2024-03-07
Support libhyprcursor

vaxerski:
This adds support for libhyprcursor and using hyprcursors in Hyprland.

Automatically supported are qt and hypr*, other apps have to support either libhyprcursor, or preferably, `wp_cursor_shape`.

gtk....

## TODO
 - [x] Bring back xcursor support that has been yeeted, obviously for legacy cursor themes
 - [x] Support animated cursors
 - [x] Meson & nix @fufexan


CaptainJack2491:
It seems hyprcursor has not been added to the dependency yet as

`yay -S hyprland-git` 

gives error saying

`ERROR: Dependency "hyprcursor" not found, tried pkgconfig and cmake`

I would have added the log file as well but I just installed arch and currently don't have a GUI.

vaxerski:
image

vaxerski:
idk maybe the aur package is cringe, CI builds from source.

CaptainJack2491:
It'll probably take time to update idk, I'll just try some hours later or maybe build from source. Thanks for replying so fast. Love your work ❤️

Cruleo:
> It seems hyprcursor has not been added to the dependency yet as
> 
> `yay -S hyprland-git`
> 
> gives error saying
> 
> `ERROR: Dependency "hyprcursor" not found, tried pkgconfig and cmake`
> 
> I would have added the log file as well but I just installed arch and currently don't have a GUI.

You can get `hyprlang-git` and `hyprcursor-git`, then hyprland-git will build.

CaptainJack2491:
It seems to be working now, thanks a lot. Something like that happened previously as well and it turned out to be because of mirrors not being updated, probably the same reason here.

PR #5019, opened 2024-03-08
check if floating when resizing from maxsize

phonetic112:
Forgor

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -689,23 +689,23 @@ void CWindow::applyDynamicRule(const SWindowRule& r) {
             Debug::log(ERR, "Rule idleinhibit: unknown mode {}", IDLERULE);
     } else if (r.szRule.starts_with("maxsize")) {
         try {
-            if (m_bIsFloating) {
-                m_sAdditionalConfigData.maxSize = configStringToVector2D(r.szRule.substr(8));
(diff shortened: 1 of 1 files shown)


src/Window.cpp
```diff
@@ -689,19 +689,23 @@ void CWindow::applyDynamicRule(const SWindowRule& r) {
             Debug::log(ERR, "Rule idleinhibit: unknown mode {}", IDLERULE);
     } else if (r.szRule.starts_with("maxsize")) {
         try {
-            m_sAdditionalConfigData.maxSize = configStringToVector2D(r.szRule.substr(8));
-            m_vRealSize                     = Vector2D(std::min((double)m_sAdditionalConfigData.maxSize.toUnderlying().x, m_vRealSize.goal().x),
-                                                       std::min((double)m_sAdditionalConfigData.maxSize.toUnderlying().y, m_vRealSize.goal().y));
-            g_pXWaylandManager->setWindowSize(this, m_vRealSize.goal());
-            setHidden(false);
+            if (m_bIsFloating) {
```
vaxerski, line 692: please prefer guards

if (floating)
continue;

phonetic112: I assume you mean return

vaxerski (approved):

PR #5036, opened 2024-03-09
fix tap-to-click and tap-and-drag names

NAHTAIV3L:
this follows how every other config variable is formatted

Every other config variable is using underscores to separate words but these are the only ones that do not follow that convention.
All I did was change where it used dashes to underscores.

If this is merged the wiki will also have to be edited to fix the renaming


vaxerski:
no, this is intentional

Issue #5006, opened 2024-03-07
the mose scroll doesn't change focus after swapwith master


a-amr:
### Hyprland Version

Hyprland, built from branch  at commit v0.35.0  ().
Date: 2024-02-05
Tag: 

flags: (if any)

System Information:
System name: Linux
Node name: xxx
Release: 6.7.6
Version: #1-NixOS SMP PREEMPT_DYNAMIC Fri Feb 23 08:51:59 UTC 2024

GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA106M [GeForce RTX 3060 Mobile / Max-Q] [10de:2520] (rev a1) (prog-if 00 [VGA controller])
06:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Cezanne [Radeon Vega Series / Radeon Vega Mobile Series] [1002:1638] (rev c5) (prog-if 00 [VGA controller])

os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.05pre592007.9df3e30ce24f"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.05 (Uakari)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.05 (Uakari)"
VERSION_CODENAME=uakari
VERSION_ID="24.05"

plugins:

### Bug or Regression?

Bug

### Description

bind=ALT, K, layoutmsg, swapwithmaster master

after running this command in the master layout the mouse scrolling or pad scrolling doesn't work right until i move the curser even one pixel 
if i don't move the mouse it scroll in the old window which not active now 

### How to reproduce

use any bind that use this command layoutmsg, swapwithmaster master

dantefromhell:
Not sure if this is the same issue, but switching focus to a different monitor using `focusmonitor` sends scroll commands to the previously focused window until the mouse cursor has been moved.

Way to reproduce
- Have 2 monitors with windows that have scrollable content
- Setup `focusmonitor`
- Work on monitor 1, hit the bind for `focusmonitor` to switch to monitor 2
- Scroll mouse without moving cursor scrolls previously focused window
[...]

Issue #4127, opened 2023-12-11
Inconsistent focus when scrolling in master layout


LanderMoerkerke:
### Hyprland Version

v0.33.1

### Bug or Regression?

Bug

### Description

If in master layout and you switch focus using the `swapnext` and `swapprev` bindings to a different window, the scrolling affects the original focused window and not the current active window. This is the case when the mouse doesn't move and only the scroll is utilised.

The bug does not occur if the layout is set to the default, dwindle layout.

### How to reproduce

1. Open two windows that are scrollable
2. Move the focus to the other window
3. Use the scroll wheel or scroll functionality of the trackpad to scroll
4. Observe the scroll to affect the 1st window 

```
monitor=,preferred,auto,1.3

input {
    mouse_refocus = false
}

$mainMod = SUPER

general {
    border_size = 2
    col.active_border = 0xffff0000
    layout = master
}

bind = $mainMod, 1, workspace, 1
bind = $mainMod, 2, workspace, 2

bind = $mainMod, left, movefocus, l
bind = $mainMod, right, movefocus, r
bind = $mainMod, up, movefocus, u
bind = $mainMod, down, movefocus, d

bind = $mainMod, J, layoutmsg, cyclenext
bind = $mainMod, K, layoutmsg, cycleprev

bind = $mainMod, Return, exec, alacritty
```

### Crash reports, logs, images, videos

https://github.com/hyprwm/Hyprland/assets/20244592/fce277fc-388f-455e-9c0d-4c3bf754e998

LanderMoerkerke:
Might be related to: !3227

vaxerski:
can you test #5074

LanderMoerkerke:
> can you test #5074

I tried to build it but failed to do so. Manual building is currently a bit flaky on Arch linux with the added dependency of `hyprcursor` and not yet up to date dependencies. 

Will test once the repositories are up to date.

Issue #3679, opened 2023-10-28
Mouse scrolling is not detected after closing a window until the mouse is moved


vE5li:
### Hyprland Version

0.30.0

### Bug or Regression?

Bug

### Description

Very similar to this issue https://github.com/hyprwm/Hyprland/issues/1675

### Steps to reproduce:
- Open two windows
- Put your mouse over one of them and close it (cursor should be on the other window now)
- Scroll without moving the mouse

### Expected outcome:
The window should scroll

### Noted outcome:
It doesn't. However, the scrolling continues after I move my mouse inside the window.

Since I mostly scroll using my keyboard this is happening quite a lot for me. I would also be happy to create a PR myself depending on what the feedback on this issue is

vaxerski:
This was fixed in -git.

tchofy:
I've been able to replicate this on both OP's situation, and also with grouped windows, but only happens if `folow_mouse` is set to anything other than 0. Built from latest git as of writing (7f4b0aa).

vaxerski:
my bad

tchofy:
From some testing with wev and groups, seems like scroll events are being sent to the previous window, instead of the newly focused window. The behavior might be the same with closed windows.
https://github.com/hyprwm/Hyprland/assets/64620317/48e8559a-ab20-4bd0-abb1-a7195c2429ae

vaxerski:
among us indeed. I know whats up

vE5li:
I just realized the same thing is happening when opening a window, the previous window gets the scroll events

tchofy:
I can still replicate this bug with some grouped windows, but seems like it's only for native wayland, xwayland works as expected. I'll do some more testing and open a new issue if I can confirm this.

vE5li:
This is still an issue. I am unable to determine what the logic is exactly but something very weird is happening that is easy to reproduce:
- Open a single window on the workspace
- Scroll without moving the mouse

The scroll input will only be propagated correctly if my mouse cursor is the bottom right quarter of the window. It's extremely weird but I have tested extensively and at least on my device this is the case 100% of the time.

This behavior is consistent also when opening additional windows, every time my mouse cursor is on the bottom right it will get the scroll input correctly, otherwise it will not.

### Closing windows
[...]

PR #5074, opened 2024-03-12
Fix master layout window focus and scroll

thejch:
Fixes https://github.com/hyprwm/Hyprland/issues/4127
Fixes https://github.com/hyprwm/Hyprland/issues/5006
Removed some duplicate code

This shouldn't change any bahavior other than fixing the issue.
In the future it's probably better to make `switchToWindow` use common code with `CKeybindManager::switchToWindow`

Ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2297,6 +2297,11 @@ void CCompositor::setWindowFullscreen(CWindow* pWindow, bool on, eFullscreenMode
         return;
     }
 
+    if (!pWindow->m_bIsFullscreen && !on) {
+        Debug::log(LOG, "Cannot unfullscreen when Window is not in fullscreen");
+        return;
+    }
+
     const auto PMONITOR = getMonitorFromID(pWindow->m_iMonitorID);
 
     const auto PWORKSPACE = getWorkspaceByID(pWindow->m_iWorkspaceID);
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2297,8 +2297,8 @@ void CCompositor::setWindowFullscreen(CWindow* pWindow, bool on, eFullscreenMode
         return;
     }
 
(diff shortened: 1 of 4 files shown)


src/Compositor.cpp
```diff
@@ -2297,6 +2297,11 @@ void CCompositor::setWindowFullscreen(CWindow* pWindow, bool on, eFullscreenMode
         return;
     }
 
+    if (!pWindow->m_bIsFullscreen && !on) {
```
vaxerski, line 2300: why not just `if (fullscreen == on)`

(there is a very similar check for the opposite case a few lines down, but we can avoid a few calcs in that cdase)

vaxerski (approved):
lgtm but I'd also like someone to test this from the affected issues

src/Compositor.cpp
thejch: Well since there are actually two kinds of fullscreen, I thought there could be a scenario where someone might wanna call it with on = true to change it from fs full to maximized, or vice versa. However, in the current code it doesn't look like it would make a difference indeed. But I can change this if you want, doesn't hurt though imo

vaxerski: that's a job for a different MR. I'd do `==` in this, and eventually build upon later

thejch: fair enough, changed

vaxerski (approved):
<3

thejch:
Took a look at the issues, I realize there's still
https://github.com/hyprwm/Hyprland/issues/3737
https://github.com/hyprwm/Hyprland/issues/3679
https://github.com/hyprwm/Hyprland/issues/2628
tbh it's probably better to just put the simulate mouse in compositor::focusWindow to always call it anyway, which would just fix all these issues, and many more, and prevent potential ones to come
But theres quite a few calls to that method and you'd have to double check every place it's called to make sure nothing is broken

PR #5100, opened 2024-03-13
lock: fix red screen issues with multiple monitors

PaideiaDilemma:

Fixes hyprlock screenshot red screen issues with multiple monitors.

It might take a while for `handleSurfaceMap` to be called. It can happen that hyprlock's screenshot feature then screenshots the solid red color. We could wait in hyprlock for the screenshot to be taken as is currently done for other compositors.

But this way it also fixes some red flashes that happen for me with multi monitor setups regardless of waiting for the screenshot or not.

However, if the map event would never happen, then we would never fall back to the red color which could be a problem.

Not sure if
1. It is OK to not render anything on a call to `renderLockscreen` if the surface is not mapped.
2. As mentioned it could be a problem that we never render the solid color when the map event never happens.

But it fixes our red screen issues. Confirmed by a friend.

Ready if there is nothing fundamentally wrong with.

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -95,6 +95,8 @@ void CSessionLockManager::onNewSessionLock(wlr_session_lock_v1* pWlrLock) {
 
             m_sSessionLock.active = false;
 
+            m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.clear();
+
             g_pCompositor->m_sSeat.exclusiveClient = nullptr;
             g_pInputManager->refocus();
 
@@ -143,6 +145,20 @@ SSessionLockSurface* CSessionLockManager::getSessionLockSurfaceForMonitor(uint64
     }
 
     return nullptr;
+}
+
+// We don't want the red screen to flash.
+// This violates the protocol a bit, but tries to handle the missing sync between a lock surface beeing created and the red screen beeing drawn.
+float CSessionLockManager::getRedScreenAlphaForMonitor(uint64_t id) {
+    const auto& NOMAPPEDSURFACETIMER = m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.find(id);
+
+    if (NOMAPPEDSURFACETIMER == m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.end()) {
+        m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.emplace(id, CTimer());
+        m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers[id].reset();
+        return 0.f;
+    }
+
+    return std::clamp(NOMAPPEDSURFACETIMER->second.getSeconds(), 0.f, 1.f);
 }
 
 bool CSessionLockManager::isSurfaceSessionLock(wlr_surface* pSurface) {
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -15,8 +15,6 @@ static void handleSurfaceMap(void* owner, void* data) {
 
     if (PMONITOR)
         g_pHyprRenderer->damageMonitor(PMONITOR);
-
-    g_pSessionLockManager->activateLock(); // activate lock here to prevent the red screen from flashing before that
 }
 
 static void handleSurfaceCommit(void* owner, void* data) {
@@ -126,6 +124,8 @@ void CSessionLockManager::onNewSessionLock(wlr_session_lock_v1* pWlrLock) {
         pWlrLock, "wlr_session_lock_v1");
 
     wlr_session_lock_v1_send_locked(pWlrLock);
+
+    g_pSessionLockManager->activateLock();
 }
 
 bool CSessionLockManager::isSessionLocked() {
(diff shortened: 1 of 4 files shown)


vaxerski:
1. Yes (not by protocol but I do not care)
2. that's a problem, IMO add a timer for 1s or something and call it a day

PaideiaDilemma:
Like this?

I also moved `activateLock` to `onNewSessionLock`. I don't really know why, but that works without any red flashes.

src/managers/SessionLockManager.cpp
```diff
@@ -126,19 +125,28 @@ void CSessionLockManager::onNewSessionLock(wlr_session_lock_v1* pWlrLock) {
...
+    if (sls->timeAlive.getSeconds() > 1.f)
+        return true;
+
+    return false;
+}
+
 SSessionLockSurface* CSessionLockManager::getSessionLockSurfaceForMonitor(const int& id) {
     for (auto& sls : m_sSessionLock.vSessionLockSurfaces) {
         if (sls->iMonitorID == id) {
-            if (sls->mapped)
```
vaxerski, line 138: I would not make this change, functions calling this expect the surf to be mapped

Instead of this you can make `shouldLockSurfaceBeMapped` take a cmonitor*

PaideiaDilemma: I am just going to use the monitor id instead. Are you implicitly casting the monitor id from `uint64_t` to `int` on purpose btw? (via `const int&`)

PaideiaDilemma:
I need to rethink this. This will be ugly anyways buy I want to make it a bit less ugly if possible.
Will let you know when ready to review again.

vaxerski:
tag me for review then

PaideiaDilemma:
@vaxerski

now assures a red screen is drawn when
- the lock client fails to create a surface
- the map event never happens
- the lock client fails to create a new surface for a new monitor connection.

(red screen when the lock is abandoned is handled via `g_pCompositor->m_sSeat.exclusiveClient`)

One slight caveat is that the screencopy in hyprlock could be slightly red tinted.
But on my 3 Monitor setup I only got `0.006000` as the worst case alpha when spamming hyprlock a lot.

src/managers/SessionLockManager.cpp
```diff
@@ -145,6 +147,20 @@ SSessionLockSurface* CSessionLockManager::getSessionLockSurfaceForMonitor(const
...
+float CSessionLockManager::getRedScreenAlphaForMonitor(uint64_t id) {
+    const auto& NOMAPPEDSURFACETIMER = m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.find(id);
+
+    if (NOMAPPEDSURFACETIMER == m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.end()) {
+        m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers.emplace(id, CTimer());
+        m_sSessionLock.mMonitorsWithoutMappedSurfaceTimers[id].reset();
+        return 0.f;
+    }
+
+    return std::clamp(NOMAPPEDSURFACETIMER->second.getSeconds(), 0.f, 1.f);
```
vaxerski, line 161: nice idea, make it `.getSeconds() - 0.5f` to get rid of the alpha tint on screenshots.

src/render/Renderer.cpp
```diff
@@ -808,16 +808,23 @@ void CHyprRenderer::renderAllClientsForWorkspace(CMonitor* pMonitor, CWorkspace*
...
             // locked with no surface, fill with red
-            CBox boxe = {0, 0, INT16_MAX, INT16_MAX};
-            g_pHyprOpenGL->renderRect(&boxe, CColor(1.0, 0.2, 0.2, 1.0));
+            const auto ALPHA = g_pSessionLockManager->getRedScreenAlphaForMonitor(pMonitor->ID);
+
+            CBox       monbox = {translate.x, translate.y, pMonitor->vecTransformedSize.x * scale, pMonitor->vecTransformedSize.y * scale};
+            g_pHyprOpenGL->renderRect(&monbox, CColor(1.0, 0.2, 0.2, ALPHA));
+
+            if (ALPHA < 1.f) /* animate */
+                g_pCompositor->scheduleFrameForMonitor(pMonitor);
```
vaxerski, line 827: what about damage? :P

PaideiaDilemma: I added `damageMonitor(pMonitor);` :)

src/render/Renderer.cpp
```diff
@@ -808,16 +808,25 @@ void CHyprRenderer::renderAllClientsForWorkspace(CMonitor* pMonitor, CWorkspace*
...
             // locked with no surface, fill with red
-            CBox boxe = {0, 0, INT16_MAX, INT16_MAX};
-            g_pHyprOpenGL->renderRect(&boxe, CColor(1.0, 0.2, 0.2, 1.0));
+            const auto ALPHA = g_pSessionLockManager->getRedScreenAlphaForMonitor(pMonitor->ID);
+
+            CBox       monbox = {translate.x, translate.y, pMonitor->vecTransformedSize.x * scale, pMonitor->vecTransformedSize.y * scale};
+            g_pHyprOpenGL->renderRect(&monbox, CColor(1.0, 0.2, 0.2, ALPHA));
+
+            if (ALPHA < 1.f) /* animate */ {
+                damageMonitor(pMonitor);
```
vaxerski, line 827: damage implies schedule, other not needed kek

PaideiaDilemma: ok makes sense

vaxerski (approved):
cheri cheri lady

Issue #5044, opened 2024-03-09
Surface doesn't immediately update on resize


phonetic112:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch makepkg at commit 717d5b3cc2b9e2f94d3ec1ba794e21bc2bb787e3 dirty (hyprctl: hide unmapped windows without -a).
Date: Fri Mar 8 12:47:12 2024
Tag: v0.36.0-77-g717d5b3c

flags: (if any)


System Information:
System name: Linux
Node name: artix
Release: 6.7.8-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Sun, 03 Mar 2024 08:30:13 +0000


GPU information: 
07:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD104 [GeForce RTX 4070] [10de:2786] (rev a1) (prog-if 00 [VGA controller])


os-release: NAME="Artix Linux"
PRETTY_NAME="Artix Linux"
ID=artix
BUILD_ID=rolling
ANSI_COLOR="0;36"
HOME_URL="https://www.artixlinux.org/"
DOCUMENTATION_URL="https://wiki.artixlinux.org/"
SUPPORT_URL="https://forum.artixlinux.org/"
BUG_REPORT_URL="https://bugs.artixlinux.org/"
PRIVACY_POLICY_URL="https://terms.artixlinux.org/docs/privacy-policy/"
LOGO=artixlinux-logo


plugins:


```

</details>

### Bug or Regression?

Bug

### Description

When resizing a window the surface isn't immediately updated for the new window size, resulting in brief stretching artifacts. It's more noticeable with animations turned off.

### How to reproduce

1. Open a window
2. Toggle floating or resize the window somehow
3. Window is briefly stretched

### Crash reports, logs, images, videos

Slow motion video of this happening: Hyprland, 60hz, recorded at 1000 fps

https://github.com/hyprwm/Hyprland/assets/73647246/4d37a8ef-53eb-4798-a09e-0e51c0a47452

Sway, 60hz

https://github.com/hyprwm/Hyprland/assets/73647246/b4fb766d-a91f-4ee0-9458-5d7c52e955c4

vaxerski:
idk why this happens tbh, we send the configure the same way, maybe sway waits for the surface to update before commencing the fullscreen?

phonetic112:
probably

phonetic112:
[...]

PR #5104, opened 2024-03-14
Prevent stretching when resizing window

phonetic112:
kinda fixes https://github.com/hyprwm/Hyprland/issues/5044. It no longer makes the surface stretch but a frame of the original surface is still visible.

### DIFF
--- a/src/managers/XWaylandManager.cpp
+++ b/src/managers/XWaylandManager.cpp
@@ -176,7 +176,6 @@ void CHyprXWaylandManager::setWindowSize(CWindow* pWindow, Vector2D size, bool f
 
     pWindow->m_vReportedPosition    = windowPos;
     pWindow->m_vPendingReportedSize = size;
-    pWindow->m_vReportedSize        = size;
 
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
this is not the way to do this

phonetic112:
how about this

vaxerski:
also no. "reported size" is the last size the window has ACKd. in ::setWindowSize, the window has not ACKd it yet. The bug is in the renderer, not the protocol handling

phonetic112:
third time's a charm

vaxerski:
all three changes are wrong. I'd rather not touch this code if I were you. Give me a test client and I'll look into it.

phonetic112:
what do you mean by test client? just resize a window

vaxerski:
but which window?

And how?

Are windows even real?

What is a window?

phonetic112:
Any window will work. Just resize the window to the smallest it can go and then toggle floating on it. It's easier to notice if you turn animations off

PR #5117, opened 2024-03-14
fix: refocus after destroy focused surface

drendog:
This will try to fix the issue of losing focus when disconnecting a monitor while in a locked session. This happens when the focus is on a surface of the monitor that is disconnecting, so this means that it will be impossible to interact with the keyboard such as writing the password to unlock the locked session. These changes will try to put focus on one of the remaining session lock surfaces when the focus is lost

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -163,7 +163,7 @@ void CSessionLockManager::removeSessionLockSurface(SSessionLockSurface* pSLS) {
     for (auto& sls : m_sSessionLock.vSessionLockSurfaces) {
         if (!sls->mapped)
             continue;
-        
+
         g_pCompositor->focusSurface(sls->pWlrLockSurface->surface);
         break;
     }
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -157,7 +157,7 @@ bool CSessionLockManager::isSurfaceSessionLock(wlr_surface* pSurface) {
 void CSessionLockManager::removeSessionLockSurface(SSessionLockSurface* pSLS) {
     std::erase_if(m_sSessionLock.vSessionLockSurfaces, [&](const auto& other) { return pSLS == other.get(); });
 
-    if (g_pCompositor->m_pLastFocus != nullptr)
+    if (g_pCompositor->m_pLastFocus)
         return;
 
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
since the mouse will be warped, why not just `g_pInputManager->refocus();`?

drendog:
I already tried it this way before, but it didn't work

vaxerski:
Ah, I see why.

src/managers/SessionLockManager.cpp
```diff
@@ -156,6 +156,16 @@ bool CSessionLockManager::isSurfaceSessionLock(wlr_surface* pSurface) {
 
 void CSessionLockManager::removeSessionLockSurface(SSessionLockSurface* pSLS) {
     std::erase_if(m_sSessionLock.vSessionLockSurfaces, [&](const auto& other) { return pSLS == other.get(); });
+
+    if (g_pCompositor->m_pLastFocus != nullptr)
```
vaxerski, line 160: redundant != nullptr

src/managers/SessionLockManager.cpp
```diff
@@ -156,6 +156,16 @@ bool CSessionLockManager::isSurfaceSessionLock(wlr_surface* pSurface) {
...
+        return;
+
+    for (auto& sls : m_sSessionLock.vSessionLockSurfaces) {
+        if (sls->mapped) {
```
vaxerski, line 164: prefer guards
```cpp
if (!sls->mapped)
    continue;
```

src/managers/SessionLockManager.cpp
drendog: it will be setted to nullptr [here](https://github.com/hyprwm/Hyprland/blob/c62a50e9746009d15bc0ede0ae2b57917e789aeb/src/managers/SessionLockManager.cpp#L41) when the surface to be destroyed is focused. If i remove this condition, the focus will be changed when the surface to be destroyed is also not focused

vaxerski: I meant only the `!= nullptr` part

`if (g_pCompositor->m_pLastFocus)`

vaxerski (dismissed):
lgtm

vaxerski:
needs a clang-format

vaxerski (approved):

PR #5138, opened 2024-03-16
opengl: report shader compilation errors from screen_shader

oatmealine:

Currently, if a shader set by `screen_shader` errors for any reason, be it a type error, syntax error, etc., the user will receive no error, and no error will be logged. This is because a call to `glGetShaderInfoLog` is never actually made. This PR makes said call, and outputs the error to `addParseError`, making a similar edit to program compilation for completeness' sake.

I'm unsure of how exactly elevating an error like this from the OpenGL module should go (sorry for my limited C++ knowledge) - I've just left it at removing the usual `"Screen shader parser: Screen shader parse failed"` error and hoping whatever caused it will make its own call to `addParseError`, but this might not be the ideal solution. Additionally, this means any `dynamic` shaders will behave this way, but nothing else is marked as `dynamic` currently except the shader configured by `screen_shader`.

I'm not confident at all this would be the best way to do this; but if it is, it's good to go.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -61,24 +61,19 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() {
 
 static void CHyprOpenGLImpl::logError(const GLuint& shader, bool program = false) {
     GLint maxLength = 0;
-    if (program) {
+    if (program)
         glGetProgramiv(shader, GL_INFO_LOG_LENGTH, &maxLength);
-    } else {
+    else
         glGetShaderiv(shader, GL_INFO_LOG_LENGTH, &maxLength);
-    }
 
     std::vector<GLchar> errorLog(maxLength);
-    if (program) {
+    if (program)
         glGetProgramInfoLog(shader, maxLength, &maxLength, errorLog.data());
-    } else {
+    else
         glGetShaderInfoLog(shader, maxLength, &maxLength, errorLog.data());
-    }
     std::string errorStr(errorLog.begin(), errorLog.end());
 
-    g_pConfigManager->addParseError(
-        (program ? "Screen shader parser: Error linking program:" : "Screen shader parser: Error compiling shader: ")
-        + errorStr
-    );
+    g_pConfigManager->addParseError((program ? "Screen shader parser: Error linking program:" : "Screen shader parser: Error compiling shader: ") + errorStr);
 
     return 0;
 }
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -59,7 +59,7 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() {
     m_tGlobalTimer.reset();
 }
 
-static void logError(const GLuint& shader, bool program = false) {
+static void CHyprOpenGLImpl::logError(const GLuint& shader, bool program = false) {
     GLint maxLength = 0;
     if (program) {
         glGetProgramiv(shader, GL_INFO_LOG_LENGTH, &maxLength);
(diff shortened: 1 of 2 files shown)


src/render/OpenGL.cpp
```diff
@@ -108,9 +119,21 @@ GLuint CHyprOpenGLImpl::compileShader(const GLuint& type, std::string src, bool
...
+
     if (dynamic) {
-        if (ok == GL_FALSE)
+        if (ok == GL_FALSE) {
+            GLint maxLength = 0;
+            glGetShaderiv(shader, GL_INFO_LOG_LENGTH, &maxLength);
+
+            // The maxLength includes the NULL character
+            std::vector<GLchar> errorLog(maxLength);
+            glGetShaderInfoLog(shader, maxLength, &maxLength, &errorLog[0]);
```
vaxerski, line 130: oh god, `.data()` exists, man

src/render/OpenGL.cpp
```diff
@@ -89,8 +89,19 @@ GLuint CHyprOpenGLImpl::createProgram(const std::string& vert, const std::string
     GLint ok;
     glGetProgramiv(prog, GL_LINK_STATUS, &ok);
     if (dynamic) {
-        if (ok == GL_FALSE)
+        if (ok == GL_FALSE) {
+            GLint maxLength = 0;
```
vaxerski, line 93: make a `static` function above to log an error instead of duplicating code

src/render/OpenGL.cpp
oatmealine: ah i was just grabbing the example khronos provided for openGL error processing, i see

src/render/OpenGL.cpp
oatmealine: Not sure how possible that is since they call different functions (`glGetShaderInfoLog` vs `glGetProgramInfoLog`) and use different error phrasing (`"Error linking program"` vs `"Error compiling shader"`)

vaxerski: `bool link`?

oatmealine: Tried that, should be good now?

src/render/OpenGL.cpp
```diff
@@ -59,6 +59,30 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() {
     m_tGlobalTimer.reset();
 }
 
+static void CHyprOpenGLImpl::logError(const GLuint& shader, bool program = false) {
+    GLint maxLength = 0;
+    if (program) {
```
vaxerski, line 64: style nit: no {} around short ifs, here and below

vaxerski (commented):
also slap a clang-format and this should be gtg

src/render/OpenGL.cpp
oatmealine: Should be good now

src/render/OpenGL.cpp
oatmealine: This should also be fixed, forgot to comment

src/render/OpenGL.cpp
```diff
@@ -59,6 +59,23 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() {
     m_tGlobalTimer.reset();
 }
 
+void CHyprOpenGLImpl::logError(const GLuint& shader, bool program) {
```
vaxerski, line 62: `logShaderError`, `logError` is misleading

vaxerski (approved):
rest lgtm

src/render/OpenGL.cpp
oatmealine: Should also be good

vaxerski (approved):

Issue #5139, opened 2024-03-16
Window slide animation breaks with second screen


Kommynct:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit e566be7847b1d32c11291edb99b169f6b64fcebb dirty (LICENSE: Update year (5132)).
Date: Sat Mar 16 12:57:45 2024
Tag: v0.37.1-2-ge566be78

flags: (if any)


System Information:
System name: Linux
Node name: anarchy-desktop
Release: 6.7.9-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 08 Mar 2024 01:58:50 +0000


GPU information: 
06:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6650 XT / 6700S / 6800S] [1002:73ef] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:


```

</details>

### Bug or Regression?

Bug

### Description

If you set the animation for windows to "slide" and there's a window coming from where a monitor is, you can see the window appear on the second monitor as it slides into place on the first.

### How to reproduce

set 

animation = windows, 1, 1, whateverbez, slide

have two displays next to eachother

open two windows, notice if the window comes from the side the other monitor is on, you'll see the monitor appear.

vaxerski:
I assume it's @thejch 's fault in #5126

you wanna fix this or should I?

thejch:
> I assume it's @thejch 's fault in #5126
> 
> you wanna fix this or should I?

Thanks for the fix. I notice that the fix seems to have broken the multi-monitor rendering if you try to move a window while it's being created. I'll make a PR real quick.

PR #5141, opened 2024-03-17
Fix incorrect rendering when manually moving a window when its being created

thejch:
ref: https://github.com/hyprwm/Hyprland/issues/5139#issuecomment-2002346554

No

Ready

### DIFF
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -389,6 +389,8 @@ void CWindow::moveToWorkspace(int workspaceID) {
 
     const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(m_iWorkspaceID);
 
+    m_bAnimatingIn = false;
+
     updateSpecialRenderData();
 
     if (PWORKSPACE) {
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -389,7 +389,7 @@ void CWindow::moveToWorkspace(int workspaceID) {
 
     const auto PWORKSPACE = g_pCompositor->getWorkspaceByID(m_iWorkspaceID);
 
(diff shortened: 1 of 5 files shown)


vaxerski (commented):
shouldn't we also change animation style to windowsMove in this case?

vaxerski:
maybe make a func in CWindow as in `changeAnimationsToMove`

thejch:
> shouldn't we also change animation style to windowsMove in this case?

The windowsMove setting doesn't seem to be doing anything right now. I tried setting it to popin and it's the same. Are you saying the func should be added to fix that? This PR is only a fix for when moving between monitors though

vaxerski:
windowsMove is for all animations that are not open/close. AKA... move.

thejch:
> shouldn't we also change animation style to windowsMove in this case?

As far as I can tell, windowsMove is being set, no? Did some debugging and as soon as I create a new window the [setAnimToMove](https://github.com/hyprwm/Hyprland/blob/4ffcdc41ff8233a92f02c633475ac3472b0998df/src/events/Windows.cpp#L33) gets called from [here](https://github.com/hyprwm/Hyprland/blob/4ffcdc41ff8233a92f02c633475ac3472b0998df/src/events/Windows.cpp#L613)

vaxerski:
if you use popin it won't be, only on the end of the animation.

thejch:
> if you use popin it won't be, only on the end of the animation.

I see what you mean now. I'll add that stuff in a minute

thejch:
> shouldn't we also change animation style to windowsMove in this case?

Added

vaxerski (approved):

Issue #2708, opened 2023-07-14
When chromium uses text-input-v1, it can only  input  in the first window 


mxmf:
```
hyprctl version                                                                                                               
Hyprland, built from branch main at commit f49af187bc0755f9b5be9376bfd52c0cf12c1e11 dirty (Xdg config home support (2047)).
Tag: v0.26.0-96-gf49af187

flags: (if any)
```

# Steps to reproduce
1. lauch chromium with flags ```--enable-features=WaylandWindowDecorations --ozone-platform-hint=auto --enable-wayland-ime```
2. type something, it worked well 
3. create a new window, type something, It only shows the candidate word panel and cannot actually write into the input box. Similar to the behavior of versions below electron21 mentioned at #1778 .

https://github.com/hyprwm/Hyprland/assets/37976818/c91ddbfe-222b-4ab8-8421-82718420c1dc

vaxerski:
why not use chromium with its much better text input protocol by using `--gtk-version=4`?

JustSimplyKyle:
![image](https://github.com/hyprwm/Hyprland/assets/68589851/2229a095-0fd4-4918-a0ca-b6056ed1ba77)
it breaks.
while v1's position works flawlessly

mxmf:
> why not use chromium with its much better text input protocol by using `--gtk-version=4`?

use `--gtk-version=4` has wrong position of  Candidate Box position.

By the way, this bug does not only occur in chromium , it also occurs in other electron applications like vscode, which not support the option `--gtk-versinon=4' that can only use tiv1 to input.

mxj4:
When using gtk4, Chromium only creates an invisible dummy gtk window while rendering its real window using wayland APIs, hence the im panel need to calculate rendering position on its own. Usually the im module can get the input position relative to the window, but under wayland the impanel process cannot get the window position relative to the screen. This is not a problem in GNOME since the im panel and mutter share the same process. We may need each im panel implementation to add some custom code to communicate with hyprland to get the current window position relative to the screen.

leonasdev:
I also encountered this issue, and `--gtk-version=4`will cause crashe instantly.

vaxerski:
I've been typing in japanese forever on hyprland and never had any issues, lel, but cool yall can now too

sujoshua:
Then enjoy !

PR #5147, opened 2024-03-17
input: Handling multiple surfaces for the text-input-v1 protocol implementation and imporve InputMethodRelay logic

sujoshua:

This PR addresses the issue where the text-input-v1 protocol could not handle cases where a client creates multiple `TextInput`.

The case one client creates multiple `TextInput` is common. e.g. chromium-based client applications create a `TextInput` for each separate surface. They then send `activate` request with `surface` param which indirectly specifies the `TextInput`. Because every `TextInput` object is bind to specific `surface`. 

Previously, the code iterated through a linked list and returned upon finding a match by comparing the current focus surface's client with `TextInput`'s client. This approach always returned the first matched  `TextInput`, whose internal surface pointed to a different surface rather than the current focus surface. This was the root cause of issue #2708.

Moreover, the original code heavily relied on iterating through linked list is not so efficient comparing to a hash lookup way. Therefore, in the  `InputMethodRelay`, a map from `surface -> TextInput `has been established, centered around the `surface.` The `InputMethodRelay` now records the current focus `surface`, enabling the retrieval of `TextInput` based on the surface. Consequently, the logic for `pendingSurface` in the original code has been removed, as the `TextInput` can now be obtained at any time through the focused surface.

About issue #2708 mentioned 

> Similar to the behavior of versions below electron21 mentioned at [[WIP\] text-input-v1 support #1778](https://github.com/hyprwm/Hyprland/pull/1778) .

yeah it's the same problem. It seems that electron below electron 21 will create a transparent surface. So in former implementation, we always caught that transparent surface's `TextInput`.

Firstly, I rarely write C++, so there might be some naive mistakes.  Please forgive me ~. Though this patch has been running on my main computer for two days without any issues.  

Secondly, there's a comment in the `InputMethodRelay.cpp` code marked "BUG POSSIBLE". This is because the logic for handling  surfaces in `text-input-v3` is quite different from `v1`. It requires the  server to proactively inform the client. Apart from using the client for matching, I can't think of a better method. However, using the client  for matching is the original code's method, so the issue mentioned for text-input-v3 still persists. But as far as the user experience with Kitty is concerned, there doesn't seem to be any major problem.

yeah

https://github.com/hyprwm/Hyprland/assets/54235339/856e0aad-3299-4c53-8926-ccb83b3fd36a

### DIFF
--- a/src/managers/input/InputMethodRelay.cpp
+++ b/src/managers/input/InputMethodRelay.cpp
@@ -87,14 +87,9 @@ void CInputMethodRelay::onNewIME(wlr_input_method_v2* pIME) {
 
             Debug::log(LOG, "IME Destroy");
 
-            if (PTI) {
-                setPendingSurface(PTI, focusedSurface(PTI));
+            if (PTI)
+                onTextInputEnter(PTI->focusedSurface);
 
-                if (PTI->pWlrInput)
-                    wlr_text_input_v3_send_leave(PTI->pWlrInput);
-                else
-                    zwp_text_input_v1_send_leave(PTI->pV1Input->resourceImpl);
-            }
         },
         this, "IMERelay");
 
@@ -144,16 +139,9 @@ void CInputMethodRelay::onNewIME(wlr_input_method_v2* pIME) {
         },
         this, "IMERelay");
 
-    const auto PTI = getFocusableTextInput();
+    if (const auto PTI = getFocusedTextInput(); PTI)
+        onTextInputEnter(PTI->focusedSurface);
 
-    if (PTI) {
-        if (PTI->pWlrInput)
-            wlr_text_input_v3_send_enter(PTI->pWlrInput, PTI->pPendingSurface);
-        else
-            zwp_text_input_v1_send_enter(PTI->pV1Input->resourceImpl, PTI->pPendingSurface->resource);
-
-        setPendingSurface(PTI, nullptr);
-    }
 }
 
 wlr_surface* CInputMethodRelay::focusedSurface(STextInput* pTI) {
@@ -322,22 +310,8 @@ SIMEKbGrab* CInputMethodRelay::getIMEKeyboardGrab(SKeyboard* pKeyboard) {
 }
 
 STextInput* CInputMethodRelay::getFocusedTextInput() {
-
-    for (auto& ti : m_lTextInputs) {
-        if (focusedSurface(&ti)) {
-            return &ti;
-        }
-    }
-
-    return nullptr;
-}
-
-STextInput* CInputMethodRelay::getFocusableTextInput() {
-    for (auto& ti : m_lTextInputs) {
-        if (ti.pPendingSurface) {
-            return &ti;
-        }
-    }
+    if (m_pFocusedSurface)
+        return getTextInput(m_pFocusedSurface);
 
     return nullptr;
 }
@@ -347,6 +321,18 @@ void CInputMethodRelay::onNewTextInput(wlr_text_input_v3* pInput) {
 }
 
 void CInputMethodRelay::createNewTextInput(wlr_text_input_v3* pInput, STextInputV1* pTIV1) {
+    // if client already has a version, reject
+    if(pInput){
+        if(!setTextInputVersion(wl_resource_get_client(pInput->resource), 3))
+            //reject
+            return;
+
+    }else{
+        if(!setTextInputVersion(pTIV1->client, 1))
+            // reject
+            return;
+    }
+
     const auto PTEXTINPUT = &m_lTextInputs.emplace_back();
 
     PTEXTINPUT->pWlrInput = pInput;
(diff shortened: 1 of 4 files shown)


src/managers/input/InputMethodRelay.cpp
```diff
@@ -475,67 +473,119 @@ void CInputMethodRelay::commitIMEState(STextInput* pInput) {
 }
 
 void CInputMethodRelay::onKeyboardFocus(wlr_surface* pSurface) {
+    // input method not available
```
vaxerski, line 476: useless comment tbh

src/managers/input/InputMethodRelay.cpp
```diff
@@ -475,67 +473,119 @@ void CInputMethodRelay::commitIMEState(STextInput* pInput) {
...
-    auto client = [](STextInput* pTI) -> wl_client* { return pTI->pWlrInput ? wl_resource_get_client(pTI->pWlrInput->resource) : pTI->pV1Input->client; };
-
-    for (auto& ti : m_lTextInputs) {
-        if (ti.pPendingSurface) {
-
-            if (pSurface != ti.pPendingSurface)
-                setPendingSurface(&ti, nullptr);
-
-        } else if (focusedSurface(&ti)) {
+    // equal to last focused surface, nothing to do.
```
vaxerski, line 480: ditto

src/managers/input/InputMethodRelay.cpp
```diff
@@ -322,23 +312,9 @@ SIMEKbGrab* CInputMethodRelay::getIMEKeyboardGrab(SKeyboard* pKeyboard) {
 }
 
 STextInput* CInputMethodRelay::getFocusedTextInput() {
-
-    for (auto& ti : m_lTextInputs) {
-        if (focusedSurface(&ti)) {
-            return &ti;
-        }
+    if (m_pFocusedSurface){
```
vaxerski, line 315: style: no {} around short ifs

src/managers/input/InputMethodRelay.cpp
```diff
@@ -144,15 +139,10 @@ void CInputMethodRelay::onNewIME(wlr_input_method_v2* pIME) {
         },
         this, "IMERelay");
 
-    const auto PTI = getFocusableTextInput();
+    const auto PTI = getFocusedTextInput();
```
vaxerski, line 142: ditto

vaxerski (commented):
generally looks good, though needs a clang-format for sure.

src/managers/input/InputMethodRelay.cpp
sujoshua: done

src/managers/input/InputMethodRelay.cpp
```diff
@@ -144,16 +139,9 @@ void CInputMethodRelay::onNewIME(wlr_input_method_v2* pIME) {
...
         this, "IMERelay");
 
-    const auto PTI = getFocusableTextInput();
-
-    if (PTI) {
-        if (PTI->pWlrInput)
-            wlr_text_input_v3_send_enter(PTI->pWlrInput, PTI->pPendingSurface);
-        else
-            zwp_text_input_v1_send_enter(PTI->pV1Input->resourceImpl, PTI->pPendingSurface->resource);
+    if (const auto PTI = getFocusedTextInput())
```
vaxerski, line 142: needs a `; PTI` at the end
```cpp
if (const auto PTI = getFocusedTextInput(); PTI)
```

src/managers/input/InputMethodRelay.cpp
```diff
@@ -478,64 +474,114 @@ void CInputMethodRelay::onKeyboardFocus(wlr_surface* pSurface) {
...
+    if(pInput->focusedSurface){
+        m_mSurfaceToTextInput.erase(pInput->focusedSurface);
+        pInput->focusedSurface = nullptr;
+    }
+}
+
+
+STextInput* CInputMethodRelay::getTextInput(wlr_surface* pSurface){
+    auto result = m_mSurfaceToTextInput.find(pSurface);
+    if (result != m_mSurfaceToTextInput.end()){
```
vaxerski, line 562: no {}

src/managers/input/InputMethodRelay.cpp
```diff
@@ -88,12 +88,7 @@ void CInputMethodRelay::onNewIME(wlr_input_method_v2* pIME) {
             Debug::log(LOG, "IME Destroy");
 
             if (PTI) {
-                setPendingSurface(PTI, focusedSurface(PTI));
-
-                if (PTI->pWlrInput)
-                    wlr_text_input_v3_send_leave(PTI->pWlrInput);
-                else
-                    zwp_text_input_v1_send_leave(PTI->pV1Input->resourceImpl);
+                onTextInputEnter(PTI->focusedSurface);
```
vaxerski, line 91: no {}

src/managers/input/InputMethodRelay.cpp
```diff
@@ -478,64 +474,114 @@ void CInputMethodRelay::onKeyboardFocus(wlr_surface* pSurface) {
...
+    /*
+         * v3 only. v1 is handled by hyprListener_textInputEnable.
+         * POSSIBLE BUG here: if one client has multiple STextInput and multiple surfaces, for any pSurface we can only record the last found ti.
+         * since original code has the same problem, it may not be a big deal.
+    */
+    if(getTextInputVersion(wl_resource_get_client(pSurface->resource)) == 3){
+        if (!getTextInput(pSurface)){
+            auto client = [](STextInput* pTI) -> wl_client* { return pTI->pWlrInput ? wl_resource_get_client(pTI->pWlrInput->resource) : pTI->pV1Input->client; };
+            for (auto& ti : m_lTextInputs) {
+                if (client(&ti) == wl_resource_get_client(pSurface->resource)  && ti.pWlrInput){
```
vaxerski, line 499: no {}

src/managers/input/InputMethodRelay.cpp
```diff
@@ -478,64 +474,114 @@ void CInputMethodRelay::onKeyboardFocus(wlr_surface* pSurface) {
...
+        Debug::log(WARN, "Client attempt to register text-input-v{}, but it has already registered text-input-v{}, ignored", version, v);
+        return 0;
+    }
+    m_mClientTextInputVersion.insert({pClient, version});
+    return 1;
+}
+
+int  CInputMethodRelay::getTextInputVersion(wl_client* pClient) {
+    auto result = m_mClientTextInputVersion.find(pClient);
+    if (result != m_mClientTextInputVersion.end()){
```
vaxerski, line 579: no {}

vaxerski (commented):
also it's still not `clang-format`ted

sujoshua:
Sorry, It seems that I've picked up some bad habits from writing too much Go.

vaxerski:
generally lgtm, but I'll clang-format it for you cuz it still isn't, lol

vaxerski (approved):
thanks a lot!

I've ran clang-format, rebased and fixed a few style nits. Should be gtg once tests pass.

cold-bin:
good job👍🏻

PR #5165, opened 2024-03-19
Config error limit/hyprctl

zakk4223:

Adds debug:error_limit to limit the number of config errors displayed in the big red notification. 
Second commit adds 'hyprctl configerrors' so scripts can get at the errors if someone would prefer to have a bar widget that yells at them when the config is messed up.

The default value of error_limit is 5. I just picked an arbitrary number that still seemed reasonably visible but prevented half the screen from being blocked. If you really want the default behavior to be 'show all, even if it covers the screen' then we can either set the default to some large number, or set it to zero and make zero a special case of 'no limit'.

I'm storing an extra copy of the error string so I can expose it to hyprctl. The rendering function is destructive and relies on the string being cleared to know the state of displaying the hyprError. If the extra copy is offensive I can rework all the rendering stuff to preserve the one copy of the string. 

Should be ready, unless you have comments on the two things above.


Ligthiago:
> Second commit adds 'hyprctl configerrors' so scripts can get at the errors if someone would prefer to have a bar widget that yells at them when the config is messed up.

In this case, IMO, it would be useful to have the ability to completely remove the errors popup handled by Hyprland itself and show them as part of widgets (EWW, AGS, whatever), just parse `hyprctl configerrors -j` output with the `configreloaded` event and show them in a nice way, with styling, expand/collapse, whatever the user wants to do with the power of his widget system.

src/config/ConfigManager.cpp
```diff
@@ -738,7 +739,7 @@ void CConfigManager::postConfigReload(const Hyprlang::CParseResult& result) {
         g_pHyprOpenGL->m_bReloadScreenShader = true;
 
     // parseError will be displayed next frame
-    if (result.error && !std::any_cast<Hyprlang::INT>(m_pConfig->getConfigValue("debug:suppress_errors")))
+    if (result.error)
```
vaxerski, line 742: why remove this?

src/debug/HyprCtl.cpp
```diff
@@ -497,6 +498,40 @@ std::string layoutsRequest(eHyprCtlOutputFormat format, std::string request) {
...
+            const auto  NEWLPOS = currErrors.find('\n', currstart);
+            std::string current = currErrors.substr(currstart, NEWLPOS - currstart);
+            currstart           = NEWLPOS;
+            if (NEWLPOS != std::string::npos)
+                currstart++;
+            result += std::format(
+                R"#(
+	"{}",)#",
+
+                current);
```
vaxerski, line 517: missing escapeJSONStrings

src/debug/HyprCtl.cpp
```diff
@@ -497,6 +498,40 @@ std::string layoutsRequest(eHyprCtlOutputFormat format, std::string request) {
     return result;
 }
 
+std::string configErrorsRequest(eHyprCtlOutputFormat format, std::string request) {
+    std::string result     = "";
+    std::string currErrors = g_pHyprError->getErrors();
+    if (format == eHyprCtlOutputFormat::FORMAT_JSON) {
+        result += "[";
+        size_t currstart = 0;
+        while (currstart != std::string::npos) {
```
vaxerski, line 507: why not just use CVarList with `\n` as the delimiter?

src/hyprerror/HyprError.cpp
```diff
@@ -91,8 +101,9 @@ void CHyprError::createQueued() {
     cairo_set_font_size(CAIRO, FONTSIZE);
     cairo_set_source_rgba(CAIRO, textColor.r, textColor.g, textColor.b, textColor.a);
 
-    float yoffset = FONTSIZE;
-    while (m_szQueued != "") {
+    float yoffset     = FONTSIZE;
+    int   renderedcnt = 0;
+    while (m_szQueued != "" && renderedcnt < VISLINECOUNT) {
```
vaxerski, line 106: would be nice to add a `...` or something at the end if there is an overflow

src/config/ConfigManager.cpp
zakk4223: I meant to move it to HyprError to just suppress the drawing of the notification, but it appears I didn't do that? lol.
Does it make more sense to leave the suppress_error here as-is and move the 'extra' copy of the error string to ConfigManager and have hyprctl fetch it from there instead?

vaxerski: leave it here absolutely, let plugins still use it even if disabled for config

zakk4223: Ok, restored it there and moved the "permanent" copy of the current error messages to the ConfigManager

src/config/ConfigManager.cpp
```diff
@@ -738,6 +743,13 @@ void CConfigManager::postConfigReload(const Hyprlang::CParseResult& result) {
         g_pHyprOpenGL->m_bReloadScreenShader = true;
 
     // parseError will be displayed next frame
+
+    if (result.error) {
```
vaxerski, line 747: no {} around short ifs

vaxerski (approved):
rest lgtm

vaxerski (approved):

vaxerski:
wiki mr needed

PR #5177, opened 2024-03-20
Fix infinite loop in workspace selector caused by integer overflow 💀

thejch:
Fix infinite loop in workspace selector caused by integer overflow 💀

When `NEXTSPACE == npos`, `i++` at the end of the loop turns i into 0 again 💀


vaxerski (approved):
:skull: :skull: :skull: :skull:

vaxerski:
somehow never happened to me lol

PR #5199, opened 2024-03-20
Add options to set tablet's active area

khalid151:
Adding options to set the active\usable drawing area of graphics tablets. It's useful when you need to match the aspect ratio of the monitor to the drawing area.

`active_area_size` and `active_area_position` are in mm to match the reported drawing dimensions in `wlr_tablet`. 

Not yet. Better name for options? Also, `hyprctl devices` could report the tablet's size in mm.  `libinput list-devices` can report that for the time being.

### DIFF
--- a/src/managers/input/Tablets.cpp
+++ b/src/managers/input/Tablets.cpp
@@ -36,6 +36,12 @@ void CInputManager::newTabletTool(wlr_input_device* pDevice) {
             const auto EVENT = (wlr_tablet_tool_axis_event*)data;
             const auto PTAB  = (STablet*)owner;
 
+            // Get tablet active area in mm and calculate bounds
+            const auto ACTIVE_AREA_SIZE = g_pConfigManager->getDeviceVec(PTAB->name, "active_area_size", "input:tablet:active_area_size");
+            const auto ACTIVE_AREA_POS = g_pConfigManager->getDeviceVec(PTAB->name, "active_area_position", "input:tablet:active_area_position");
+            const auto MIN_BOUNDS = Vector2D { ACTIVE_AREA_POS.x / PTAB->wlrTablet->width_mm, ACTIVE_AREA_POS.y / PTAB->wlrTablet->height_mm };
+            const auto MAX_BOUNDS = Vector2D { ( ACTIVE_AREA_POS.x + ACTIVE_AREA_SIZE.x ) / PTAB->wlrTablet->width_mm, ( ACTIVE_AREA_POS.y + ACTIVE_AREA_SIZE.y ) / PTAB->wlrTablet->height_mm };
+
             switch (EVENT->tool->type) {
                 case WLR_TABLET_TOOL_TYPE_MOUSE:
                     wlr_cursor_move(g_pCompositor->m_sWLRCursor, PTAB->wlrDevice, EVENT->dx, EVENT->dy);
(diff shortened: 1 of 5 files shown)


vaxerski:
how does that differ from the other region options?

khalid151:
Other region options map the entire tablet's area to a portion of the screen. This maps a subset of the tablets area to that region.

vaxerski:
_what_?

khalid151:
`region_position` and `region_size` are used in [wlr_cursor_map_input_to_region()](https://wayland.emersion.fr/wlroots/wlr/types/wlr_cursor.h.html#func-wlr_cursor_map_input_to_region).
https://github.com/hyprwm/Hyprland/blob/4c796683c05a0eaccc14aae8875f06972f9f3c5e/src/managers/input/InputManager.cpp#L1440
This function maps the received input from tablet to a specified region on the screen.

For example, if you have a 1920x1080 monitor and you set `region_size = 1080 1080`, it will be mapped to this portion of the screen, a square. Pointer won't be able to move past it.

region

Meanwhile, `active_area_size` limits the stylus movement on the tablet itself. Pointer movement will only register while the stylus is over this area.

active_area

This helps maintain a similar aspect ratio to the monitor (or the `region_size`). Like for example, if you draw a perfect circle on the tablet, it will be a perfect circle on the screen. Current options don't allow that.

vaxerski:
okay, I get it. You need to add this to `proximity` then as well

vaxerski:
disregard, I thought I committed that patch.

src/managers/input/Tablets.cpp
```diff
@@ -36,6 +36,12 @@ void CInputManager::newTabletTool(wlr_input_device* pDevice) {
             const auto EVENT = (wlr_tablet_tool_axis_event*)data;
             const auto PTAB  = (STablet*)owner;
 
+            // Get tablet active area in mm and calculate bounds
+            const auto ACTIVE_AREA_SIZE = g_pConfigManager->getDeviceVec(PTAB->name, "active_area_size", "input:tablet:active_area_size");
```
vaxerski, line 40: Extreme performance kill. Move the variable setting to `setTabletConfigs` and store a copy inside STablet itself.

Using `getDevice*` outside `set*Config` is banned

src/managers/input/Tablets.cpp
```diff
@@ -52,7 +58,14 @@ void CInputManager::newTabletTool(wlr_input_device* pDevice) {
                     if (PTAB->relativeInput)
                         wlr_cursor_move(g_pCompositor->m_sWLRCursor, PTAB->wlrDevice, dx, dy);
                     else
+                    {
```
vaxerski, line 61: clang-format not ran

src/managers/input/Tablets.cpp
khalid151: Now that you mention it, I felt I was doing something off. Thanks for the heads-up, that makes much more sense.

khalid151:
What about reporting the tablet dimensions? I was thinking of doing something like this:
```diff
diff --git a/src/debug/HyprCtl.cpp b/src/debug/HyprCtl.cpp
index f13843a3..a9fb69ab 100644
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -633,7 +633,7 @@ std::string devicesRequest(eHyprCtlOutputFormat format, std::string request) {
         }
 
         for (auto& d : g_pInputManager->m_lTablets) {
-            result += std::format("\tTablet at {:x}:\n\t\t{}\n", (uintptr_t)&d, d.name);
+            result += std::format("\tTablet at {:x}:\n\t\t{}\n\t\t\tsize: {}x{}mm\n", (uintptr_t)&d, d.name, d.wlrTablet->width_mm, d.wlrTablet->height_mm);
         }
 
         for (auto& d : g_pInputManager->m_lTabletTools) {
```
```
$ hyprctl devices
...
Tablets:
	Tablet Pad at 5757c2673770 (belongs to 0 -> )
	Tablet at 5757c2a785e0:
		wacom-intuos-pro-m-pen
			size: 224x148mm
	Tablet Tool at 5757c3464e90 (belongs to 5757c3464e90)
```

khalid151:
@Kommynct You're thinking of something along the lines of mapping a pointer to a display `wlr_cursor_map_input_to_output()`. Unfortunately, I don't have a second monitor to test. And I have no idea how multiple pointers\cursors are handled in Hyprland.

vaxerski:
> What about reporting the tablet dimensions? I was thinking of doing something like this:

reasonable

vaxerski:
> Unfortunately, I don't have a second monitor to test. And I have no idea how multiple pointers\cursors are handled in Hyprland.

they aren't. You have only one.

khalid151:
> reasonable
great then, I'll push the changes

> You have only one.
which makes sense

khalid151:
https://github.com/hyprwm/hyprland-wiki/pull/549

vaxerski (approved):
lgtm

Issue #4626, opened 2024-02-05
focusworkspaceoncurrentmonitor doesn't fully focus after swap


ArtemSmaznov:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 181f651de2e67917c7fcc72f7a69e6f4707954e5 dirty (vector: avoid min0 clamps without a max being invalid).
Date: Mon Feb 5 17:37:34 2024
Tag: v0.35.0-4-g181f651d

flags: (if any)


System Information:
System name: Linux
Node name: izumi
Release: 6.7.3-zen1-2-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 02 Feb 2024 17:03:56 +0000


GPU information:
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 32 [Radeon RX 7700 XT / 7800 XT] [1002:747e] (rev c8) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

First of all, thanks for a resolution of the mouse dependency with this dispatcher.

Now I've noticed another unexpected behavior - after swapping workspaces by targeting one that is on another screen, the focus partially moves to the side screen:
* :x: col.active_border is seen on the side screen and col.inactive_border activates on the previously focused window on the current screen
* :x: keyboard input is on the windows on side screen 
*  ✔️ hitting a binding to open a new app opens it on the current screen

I've only tested this with both screens having windows on them, hyprland is a bit unpredictable where empty workspaces are involved :P

### How to reproduce

```gherkin
Given I have 2 monitors
And I have "Window A" in "Workspace 1" on "Monitor DP-1"
And I have "Window B" in "Workspace 2" on "Monitor DP-2"
And "Window A" is focused
And I don't use the mouse at all

[...]

PR #5202, opened 2024-03-21
Fix animation issue in focusworkspaceoncurrentmonitor

thejch:
Fix animation rendering issue caused by workspace offset
ref: https://github.com/hyprwm/Hyprland/issues/4626#issuecomment-2010743763

Did remove some code thats not needed if workspace offset is not used

Ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -349,6 +349,7 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("misc:background_color", Hyprlang::INT{0xff111111});
     m_pConfig->addConfigValue("misc:new_window_takes_over_fullscreen", Hyprlang::INT{0});
     m_pConfig->addConfigValue("misc:enable_hyprcursor", Hyprlang::INT{1});
+    m_pConfig->addConfigValue("misc:animate_across_monitors", Hyprlang::INT{1});
 
     m_pConfig->addConfigValue("group:insert_after_current", Hyprlang::INT{1});
     m_pConfig->addConfigValue("group:focus_removed_window", Hyprlang::INT{1});
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -349,7 +349,6 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("misc:background_color", Hyprlang::INT{0xff111111});
     m_pConfig->addConfigValue("misc:new_window_takes_over_fullscreen", Hyprlang::INT{0});
     m_pConfig->addConfigValue("misc:enable_hyprcursor", Hyprlang::INT{1});
-    m_pConfig->addConfigValue("misc:animate_across_monitors", Hyprlang::INT{1});
 
     m_pConfig->addConfigValue("group:insert_after_current", Hyprlang::INT{1});
     m_pConfig->addConfigValue("group:focus_removed_window", Hyprlang::INT{1});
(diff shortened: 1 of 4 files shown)


src/managers/AnimationManager.cpp
```diff
@@ -245,10 +236,7 @@ void CAnimationManager::tick() {
                                            BORDERSIZE + ROUNDINGSIZE); // bottom
 
                 // damage for new box
-                CBox       WLRBOXNEW        = {PWINDOW->m_vRealPosition.value(), PWINDOW->m_vRealSize.value()};
-                const auto PWINDOWWORKSPACE = g_pCompositor->getWorkspaceByID(PWINDOW->m_iWorkspaceID);
-                if (PWINDOWWORKSPACE)
-                    WLRBOXNEW.translate(PWINDOWWORKSPACE->m_vRenderOffset.value());
+                CBox WLRBOXNEW = {PWINDOW->m_vRealPosition.value(), PWINDOW->m_vRealSize.value()};
```
vaxerski, line 239: isnt there a util to get this? plus, shouldn't we add the workspace offset?

src/render/Renderer.cpp
```diff
@@ -228,6 +228,8 @@ bool CHyprRenderer::shouldRenderWindow(CWindow* pWindow, CMonitor* pMonitor, CWo
         return true;
 
     if (pWindow->m_vRealPosition.isBeingAnimated()) {
+        if (PWINDOWWORKSPACE && PWINDOWWORKSPACE->m_vRenderOffset.isBeingAnimated())
```
vaxerski, line 231: not that simple, if you are for example doing swapcurrentworkspaces, it should be visible IMO.

thejch: It is visible though. Tested `swapactiveworkspaces` and `movecurrentworkspacetomonitor`. The offset is not in effect during these operations.

src/managers/AnimationManager.cpp
thejch: Yeah I'll add the offset back. Problem is I noticed if you do that it will add a little damage to other monitors when you switch workspaces. But maybe it should be properly fixed later.

src/render/Renderer.cpp
vaxerski: my bad

src/managers/AnimationManager.cpp
vaxerski: I guess we can add an optional monitor param to `addDamage`?

vaxerski: up to you if you want to do this in this mr, ping me with your preferred solution

thejch: @vaxerski Probably better to merge this one first. I can draft up something to fix that issue later

vaxerski: oke

src/config/ConfigManager.cpp
```diff
@@ -349,6 +349,7 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("misc:background_color", Hyprlang::INT{0xff111111});
     m_pConfig->addConfigValue("misc:new_window_takes_over_fullscreen", Hyprlang::INT{0});
     m_pConfig->addConfigValue("misc:enable_hyprcursor", Hyprlang::INT{1});
+    m_pConfig->addConfigValue("misc:animate_across_monitors", Hyprlang::INT{1});
```
vaxerski, line 352: well now you need a wiki MR. My question is, why is this an option? I am not in favor of this being an option

thejch: Ref: https://discord.com/channels/961691461554950145/967398886530515067/1220460557753122928
if you have different scaling/DPI ratios on different monitors it might look weird

vaxerski: it already will if you have a window partially on one or the other. (floating)

thejch: That is indeed true. I couldn't think of a better way to address that except for adding that config option, but I will remove the config for now

thejch: changed

vaxerski (approved):
ergergergergerfgergfrtghbf

Issue #1348, opened 2023-01-09
Hyprland with on-click causes weird behavior


mrbman:
Clicking on any module with on-click in waybar or onclick in eww will make the action repeat if you click anything again. If I click an 'onclick' to launch application A, and then a different onclick to launch any other application, application A is always launched again.

This bug was initially reported [here](https://github.com/Alexays/Waybar/issues/1850) however after further testing this issue seems to be specific to hyprland.

Rydgel:
Does it only happens when launching certain kind of apps? Doesn't happen here with `pamixer -t` on-click

riley-martin:
> So if I right click then left click and repeat I can open all of the apps via waybar and eww.

Same here. I am using `hyprland-git` and `waybar-hyprland-git`.

Vanadiom:
System: Arch
Installed: `waybar-hyprland-git` `hyprland` via `paru`
I also have this issue on `waybar-hyprland-git`. 
It seems to be inconsistent somehow, because, this problem:
- does not appear when using the Workspace switcher first, and then something else
- does appear when using something else and then the Workspace mixer.

However I am unsure if the Workspace switcher works using an `on-click`

justinesmithies:
I am running latest Arch git version of Hyprland and the latest git version of Eww Bar and all my on-click scripts work fine exept for the one that uses foot terminal to run a script that basically opens a terminal window and runs paru -Syu and waits for obvs the user password then updates and closes. After clicking on that I get the error shown below and clicking on any other thing on the bar  then opens the same script again. I have to physically quit and go back in.
This is the error I see in Eew Bars logs after that.

`warn: input.c:2211: stray button release event (compositor bug?)`

All my other scripts for network , brightness and so on all run fine with no issues.

zoro11031:
I've noticed for me the issue only happens with clickable events that utilize swaync. Everything else behaves normally

dnkl:
> warn: input.c:2211: stray button release event (compositor bug?)

This is from foot. I'm guessing the bar only sees a button **press** event, and that the **release** event goes to the launched application.

The bar probably gets confused when it sees two press events without a release in between.

justinesmithies:
> > warn: input.c:2211: stray button release event (compositor bug?)
> 
> This is from foot. I'm guessing the bar only sees a button **press** event, and that the **release** event goes to the launched application.
> 
> The bar probably gets confused when it sees two press events without a release in between.

[...]

PR #5219, opened 2024-03-22
fix: release all mouse buttons before focus on new surface

drendog:
I'm using waybar with a custom module for swaync, and is configured with a on click event that it should toggle the notification center. When I click the swaync module, it will open the notification center but it causes weird behavior. The issue arises because, when the compositor create a new layer surface (in this case, the notification center), its surface gains focus, which results in waybar being unable to detect the mouse click release. This leads to undesired behaviors, such as the bar not responding as expected, like reopening the notification center when clicking somewhere else.
So a potential solution is to notify the release of all mouse button before focusing in another new surface.

#1348 https://github.com/Alexays/Waybar/issues/2473 https://github.com/ErikReider/SwayNotificationCenter/issues/292 https://github.com/Alexays/Waybar/issues/2166

Yes

### DIFF
--- a/src/events/Layers.cpp
+++ b/src/events/Layers.cpp
@@ -148,6 +148,7 @@ void Events::listener_mapLayerSurface(void* owner, void* data) {
         (!g_pCompositor->m_sSeat.mouse || !g_pInputManager->isConstrained());
 
     if (GRABSFOCUS) {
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
sure

Issue #4599, opened 2024-02-03
wlogout "logout" function hangs if activated by mouse click.  Hotkey ("e") works fine.


joekm:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
<Paste the output of the command here>
```

</details>

### Bug or Regression?

Bug

### Description

Just got the latest hyprland-git today and did not have this behavior prior to it.  

If I run wlogout and try to logout with a mouse click, it will hang until i hit a key.  If I use the hotkey for logout ("e"), it works just fine.  If I just type the command called by wlogout in a terminal ("hyprctl dispatch exit 0"), it also works just fine.    So, it seems to be an issue with the mouse click.   

Lock, reboot, and shutdown all work just fine (I don't use suspend or hibernate on this machine).

### How to reproduce

Call wlogout and attempt to logout using a mouse click.

vaxerski:
wouldn't that make it an issue with wlogout?

joekm:
wlogout code hasn't changed in years and this issue started with the latest hyperland-git update.  My suspicion is that it's something to do with grabbing the mouse input on the hyprland side.

MightyPlaza:
> wlogout code hasn't changed in years and this issue started with the latest hyperland-git update. My suspicion is that it's something to do with grabbing the mouse input on the hyprland side.

if it is a regression mark it as one

do a git bisect (aka find the commit where it broke)

zakk4223:
It's this commit: 5eeec8860e2dc8ff80dda26243778b5ecc923ca2

It's because the exit flag processing only happens inside of onKeyboardKey()

Even in the case where it 'works' it's actually firing on the e key-up event. If I navigate to the logout entry with the arrow keys and then hit enter, it usually misses the enter-up event, causing me to have to hit another key to trigger the exit.

Fxzzi:
I use wleave instead of wlogout. wleave can use existing wlogout configs, it's literally a drop in replacement. Try with that instead?

joekm:
> I use wleave instead of wlogout. wleave can use existing wlogout configs, it's literally a drop in replacement. Try with that instead?

I did actually, and the issue persist.  For now, I'm just using hotkey "e" to logout.

Fxzzi:
> I'm not a developer so I don't really know "bug" vs "regression".

Bug is "it's always been an issue"
Regression is "it used to work, now vaxry broke it"

zakk4223:
[...]

PR #5240, opened 2024-03-24
Move exit trigger to monitor frame

dmayle:
This is a fix for #4599 .  It works by moving the exit checking logic to the monitor frame event listener.

Because there can be multiple monitors, the first monitor to have a frame event resets the trigger.

It works, and is tested on my machine.

### DIFF
--- a/src/events/Monitors.cpp
+++ b/src/events/Monitors.cpp
@@ -117,6 +117,13 @@ void Events::listener_newOutput(wl_listener* listener, void* data) {
 }
 
 void Events::listener_monitorFrame(void* owner, void* data) {
+    if (g_pCompositor->m_bExitTriggered) {
+        // Only signal cleanup once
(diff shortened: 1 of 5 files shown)


src/events/Monitors.cpp
```diff
@@ -117,6 +117,12 @@ void Events::listener_newOutput(wl_listener* listener, void* data) {
 }
 
 void Events::listener_monitorFrame(void* owner, void* data) {
+    if (g_pCompositor->m_bExitTriggered) {
+        // Only signal cleanup once
+        g_pCompositor->m_bExitTriggered = false;
+        g_pCompositor->cleanup();
```
vaxerski, line 123: we probably should return; after this.

dmayle: Yes, that makes perfect sense.  Done.

vaxerski (approved):
lgtm

PR #5251, opened 2024-03-24
Send workspace events during swapActiveWorkspaces and focusWorkspaceOnCurrentMonitor (when swapping)

thejch:
When swapping active workspaces, it should send an event because it's focusing on the new workspace that gets swapped in.
Since swapping is also called in focusWorkspaceOnCurrentMonitor, this should also fix the issue that the workspace event is not sent sometimes when focusWorkspaceOnCurrentMonitor is called.

Not really. Just events anyway doesnt change any behavior

Probably ready


vaxerski (approved):
lgtm

PR #5262, opened 2024-03-25
Fix resizeparams

### LINKED ISSUE #5261: Dispatchers resizeparams broken regression
Somewhere from v0.37.1 (c5e28eb) to 356414639fc06ace7045236ec41d673342981057 , dispatchers resizeparams got broken.

Make a floating window and execute:
`hyprctl dispatch resizeactive exact 80% 40%`
It should make widht=80% and height=40%
Instead it makes width=40% and height=40%

Aqa-Ib:
Fixes #5261

A better solution would be to be able to do what a94b902 author wanted to achieve without breaking resizeparams. But in the meantime, reverting the breakage would be reasonable.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -10,6 +10,7 @@
 #include <systemd/sd-daemon.h> // for sd_notify
 #endif
 #include <ranges>
+#include "helpers/VarList.hpp"
 
 int handleCritSignal(int signo, void* data) {
     Debug::log(LOG, "Hyprland received signal {}", signo);
@@ -2531,8 +2532,9 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
     bool        yIsPercent = false;
     bool        isExact    = false;
 
-    std::string x = args.substr(0, args.find_first_of(' '));
-    std::string y = args.substr(args.find_first_of(' ') + 1);
+    CVarList varList(args, 2, ' ', true);
+    std::string x = varList.size() > 0 ? varList[0] : "";
+    std::string y = varList.size() > 1 ? varList[1] : "";
 
     if (x == "exact") {
         x       = y.substr(0, y.find_first_of(' '));
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2532,7 +2532,7 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
     bool        yIsPercent = false;
     bool        isExact    = false;
 
-    CVarList    varList(args, 3, ' ', true);
+    CVarList    varList(args, 3, 's', true);
     std::string x = varList.size() > 0 ? varList[0] : "";
     std::string y = varList.size() > 1 ? varList[1] : "";
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2533,12 +2533,12 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
     bool        isExact    = false;
 
     CVarList    varList(args, 3, 's', true);
-    std::string x = varList.size() > 0 ? varList[0] : "";
-    std::string y = varList.size() > 1 ? varList[1] : "";
+    std::string x = varList[0];
+    std::string y = varList[1];
 
     if (x == "exact") {
-        x       = varList.size() > 1 ? varList[1] : "";
-        y       = varList.size() > 2 ? varList[2] : "";
+        x       = varList[1];
+        y       = varList[2];
         isExact = true;
     }
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2523,7 +2523,7 @@ SLayerSurface* CCompositor::getLayerSurfaceFromSurface(wlr_surface* pSurface) {
 
 // returns a delta
 Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, const Vector2D& relativeTo) {
-    if (!args.contains(' '))
+    if (!args.contains(' ') && !args.contains('\t'))
         return relativeTo;
 
     const auto  PMONITOR = m_pLastMonitor;
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2532,7 +2532,7 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
     bool        yIsPercent = false;
     bool        isExact    = false;
 
-    CVarList varList(args, 2, ' ', true);
+    CVarList    varList(args, 2, ' ', true);
     std::string x = varList.size() > 0 ? varList[0] : "";
     std::string y = varList.size() > 1 ? varList[1] : "";
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2532,7 +2532,7 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
     bool        yIsPercent = false;
     bool        isExact    = false;
 
-    CVarList    varList(args, 3, 's', true);
+    CVarList    varList(args, 0, 's', true);
     std::string x = varList[0];
     std::string y = varList[1];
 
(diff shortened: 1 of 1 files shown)


vaxerski:
Makes sense. Maybe just fix it right now? It's easy with a `CVarList`, it has a prop to ignore empty.

Aqa-Ib:
This is my first time writing C++ code, take it into account xD

src/Compositor.cpp
```diff
@@ -2531,8 +2532,9 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
     bool        yIsPercent = false;
     bool        isExact    = false;
 
-    std::string x = args.substr(0, args.find_first_of(' '));
-    std::string y = args.substr(args.find_first_of(' ') + 1);
+    CVarList varList(args, 2, ' ', true);
+    std::string x = varList.size() > 0 ? varList[0] : "";
+    std::string y = varList.size() > 1 ? varList[1] : "";
```
vaxerski, line 2537: the point of that fix kinda got fucked, tho. Make it max 3 and in `exact` just use `[1]` and `[2]`

Aqa-Ib: My bad. Thanks for the review! I hope that now it is good.

bvr-yr: and probably `'s'` as sep? or it will fuckup tabs in conf

Aqa-Ib: It is working fine with ' ' as delimiter even if I put tabs between args, but I don't if in some case 's' would be preferred here.

src/Compositor.cpp
```diff
@@ -2531,12 +2532,13 @@ Vector2D CCompositor::parseWindowVectorArgsRelative(const std::string& args, con
...
-    std::string x = args.substr(0, args.find_first_of(' '));
-    std::string y = args.substr(args.find_last_of(' ') + 1);
+    CVarList    varList(args, 3, 's', true);
+    std::string x = varList.size() > 0 ? varList[0] : "";
+    std::string y = varList.size() > 1 ? varList[1] : "";
 
     if (x == "exact") {
-        x       = y.substr(0, y.find_first_of(' '));
-        y       = y.substr(y.find_first_of(' ') + 1);
+        x       = varList.size() > 1 ? varList[1] : "";
```
vaxerski, line 2540: no need to check size, CVarList will return `""` if it's outside of its size

Aqa-Ib: Excuse me, if I understand you correctly, is it safe to remove all four size checks in the PR ?

Also, I have tested this and tabs characters are not working even when using 's' as delim.
Thanks for your patience.

bvr-yr: are you testing sith `\t`? you need actual tab, which may appear in config
ctrv+v and press tab if u wish from cmdline. or insert
```sh
hyprctl dispatch resizeactive exact 80%		40%
```

Aqa-Ib: Yes, I have inserted an actual tab character in different editors. Also, if you want to test it, notice that in the line of code that you wrote there is a space before the tab.

Aqa-Ib: Thank you, now that you tell, I have realized where is the error. It only works with a tab character when using the `exact` argument.

Aqa-Ib: hyprctl dispatch resizeactive exact 80%\t40%    = fail
hyprctl dispatch resizeactive exact\t80%\t40%  = fail
hyprctl dispatch resizeactive exact\t80% 40%    = success
hyprctl dispatch resizeactive exact 80%\t40%    = success

vaxerski: maybe `std::isspace` doesn't catch a tab? tho it should I believe.

Anyways, `is it safe to remove all four size checks in the PR ?`

si senior.

bvr-yr: yeah I tested it, it seem like you need to set `0` instead of `3` maxsize
I forgot I had 0 here locally

```ini
bind = SUPER CTRL, Z, resizeactive,     	    exact    	      60%		40%  56
```
works, which is mix of spaces and tabs


about `\t`, it's not expanded to real tab when you run like that.
if you want to test it from hyprctl
```sh
hyprctl dispatch resizeactive $(echo -e "exact\t60%\t40%")
```

Aqa-Ib: It catches the tab, but the arguments also need to contain a space for some reason.

Aqa-Ib: Yea, I wrote \t for it to be clear that they were actual tabs.
Can you try with one of these, replacing <tab> with an actual tab ?
hyprctl dispatch resizeactive 80%<tab>40%
hyprctl dispatch resizeactive exact<tab>80%<tab>40%

It works when you mix tabs and spaces, but not when you only write tabs. Look at this, I hope it is clear now:

Works:
+ Arg1<tab>Arg2<space>Arg3
+ Arg1<space>Arg2<tab>Arg3

Fails:
-  Arg1<tab>Arg2<tab>Arg3
 
Thanks.

bvr-yr: > yeah I tested it, it seem like you need to set 0 instead of 3 maxsize

or `4` if you wish, that makes no difference
dunno why it returns wrong when 3
```cpp
    Debug::log(WARN, "exact:{}  x:@{}@  y:@{}@",  isExact, x, y);
    Debug::log(WARN, "size:{}", varList.size());
    Debug::log(WARN, "1:{}", varList[0]);
    Debug::log(WARN, "2:{}", varList[1]);
    Debug::log(WARN, "3:{}", varList[2]);
```

```
[WARN] exact:true  x:@80%@  y:@80%	 	40%@
[WARN] size:3
[WARN] 1:exact
[WARN] 2:80%
[WARN] 3:80%	 	40%
```
with `0` or `4`
```
[WARN] exact:true  x:@80%@  y:@40%@
[WARN] size:3
[WARN] 1:exact
[WARN] 2:80%
[WARN] 3:40%
```

Aqa-Ib: I have compiled it with `0` and `4` with the same results as with `3`

bvr-yr: ohmygod sure xD
look at this [line](https://github.com/hyprwm/Hyprland/pull/5262/files#diff-3c089c6c6b23a85ab5b4cffb002e8880ad069c8a8b3462f7aafa9e5345c0333dR2526)

just do
```cpp
if (!args.contains(' ') && !args.contains('\t'))
    return relativeTo;
```

and with `0` maxsize all stuff works, spaces, tabs, mixed, hyprctl or conf

Aqa-Ib: Actually, if I use Control+V+Tab to get a tab character in the terminal as @bvr-yr suggested, it works perfectly fine in every situation.

However, it has problems with tab characters from GUI editors like VSCode, kate, or gedit.

If someone else can test this it would be good. Thanks.

bvr-yr: I'm not sure what's wrong with 3 maxsize, but it returns strange things like this
```
[WARN] exact:true  x:@60%@  y:@0%  40%@
[WARN] size:3
[WARN] 1:exact
[WARN] 2:60%
[WARN] 3:0%  40%
[ERR] parseWindowVectorArgsRelative: args not numbers
```
for example with bind
```ini
bind = SUPER CTRL, Z, resizeactive, exact    60%  40%
```
anyways, initial fix was config issues, so let's see what boss will tell about
```cpp
if (!args.contains(' ') && !args.contains('\t'))
    return relativeTo;
```
and 0 maxsize

vaxerski: tag me when this is ready. And yes, generally 0 maxsize is used

Aqa-Ib:
@vaxerski ready

bvr-yr:
maaan.. doesn't work with maxsize `3`

Aqa-Ib:
Thank you for your help @bvr-yr
It works fine on my end like this, I don't know what is wrong on your side though.

Aqa-Ib:
Ok I got you, It does not work when mixing zeroes and tabs with `3`.
Let me fix it, sorry @bvr-yr

bvr-yr:
that's ok, it just makes no difference, anyway with `0` right params will be returned
> And yes, generally 0 maxsize is used

Aqa-Ib:
Let me check if it needs sizechecks when using `0`

Aqa-Ib:
Seems good @vaxerski 
Kudos @bvr-yr

vaxerski (approved):
lgtm

PR #5277, opened 2024-03-26
Update hyprctl to output json with `--batch`

khalid151:
Using `-j` now along with `--batch` would output JSON as expected.

Not sure if using regex is overkill for this but it looks neat this way and saves the headache of dealing with mixed whitespace after semicolons like this `"command; command;command;  command"` (if somebody decided to write commands this way).

I guess?


vaxerski (approved):
lgtm

PR #5291, opened 2024-03-27
fix fullscreen + floating on monitor focus change

MightyPlaza:
doesn't reject floating windows if they're also fullscreened in `CCompositor::getWindowInDirection()` (also opposite case)

shouldn't cause issues


vaxerski (approved):
makes sense

PR #5307, opened 2024-03-28
fix bottom layer slide

MightyPlaza:
fixes typo so bottom layer slide animation works as intended

shouldn't cause issues


vaxerski (approved):
ooOps

PR #5308, opened 2024-03-28
layer: don't change workspace when restoring focus

MightyPlaza:
checks if the previously focused window is in a visible workspace before giving it focus, when closing a layer with input

shouldn't cause issues


vaxerski (approved):
makes sense, thanks.

PR #5340, opened 2024-03-30
Match all workspace rules instead of the first one only

thejch:
Make workspaces match all workspace rules instead of the first one only.
Currently if you have two workspace rules that match the same workspace like this
```
workspace = 1, monitor:WL-1, default:true
workspace = w[t1], border:no, gapsout:0
```
only the first one gets applied

This will fix that and apply all the rules.

not really

ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1259,8 +1259,13 @@ void CCompositor::sanityCheckWorkspaces() {
     auto it = m_vWorkspaces.begin();
     while (it != m_vWorkspaces.end()) {
 
-        const auto WORKSPACERULE = g_pConfigManager->getWorkspaceRuleFor(it->get());
-        if (WORKSPACERULE.isPersistent) {
+        const auto WORKSPACERULES = g_pConfigManager->getWorkspaceRulesFor(it->get());
+        bool       isPersistent   = false;
+        for (auto& wsRule : WORKSPACERULES) {
+            if (wsRule.isPersistent)
+                isPersistent = true;
+        }
+        if (isPersistent) {
             ++it;
             continue;
         }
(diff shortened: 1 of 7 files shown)


vaxerski (approved):
lgtm. No backdoors present it seems

zakk4223:
Apologizes for the late comment here but...

This feels really clunky to actually use. Even in this MR almost every instance of 'I want a WS rule value' is getWorkspaceRulesFor....->loop so the last one gets used. Can we maybe have something like this (https://github.com/zakk4223/hyprNStack/blob/46a0b2a6755e68effc5c5c5e1d9d87e9de96f468/nstackLayout.cpp#L49) as an 'official' API? You could even use the old function name!

Having an API that does the merging means you're also less likely to have plugin authors roll their own and screw it up.

thejch:
> Apologizes for the late comment here but...
> 
> This feels really clunky to actually use. Even in this MR almost every instance of 'I want a WS rule value' is getWorkspaceRulesFor....->loop so the last one gets used. Can we maybe have something like this (https://github.com/zakk4223/hyprNStack/blob/46a0b2a6755e68effc5c5c5e1d9d87e9de96f468/nstackLayout.cpp#L49) as an 'official' API? You could even use the old function name!
> 
> Having an API that does the merging means you're also less likely to have plugin authors roll their own and screw it up.

good point tbh
didnt think too much of it but i agree

thejch:
> Apologizes for the late comment here but...
> 
> This feels really clunky to actually use. Even in this MR almost every instance of 'I want a WS rule value' is getWorkspaceRulesFor....->loop so the last one gets used. Can we maybe have something like this (https://github.com/zakk4223/hyprNStack/blob/46a0b2a6755e68effc5c5c5e1d9d87e9de96f468/nstackLayout.cpp#L49) as an 'official' API? You could even use the old function name!
> 
> Having an API that does the merging means you're also less likely to have plugin authors roll their own and screw it up.

i drafted a pr real quick https://github.com/hyprwm/Hyprland/pull/5433
waiting for @vaxerski 's comment

Issue #4300, opened 2023-12-30
hyprctl should be more helpful discovering usage


hholst80:
### Description

```
[I] root@neo ~
# man hyprctl 
No manual entry for hyprctl
[I] root@neo ~ [16]
# 
```

```
[I] root@neo ~ [16]
# hyprctl --help
usage: hyprctl [(opt)flags] [command] [(opt)args]

commands:
    activewindow
    activeworkspace
    binds
    clients
    cursorpos
    devices
    dispatch
    getoption
    globalshortcuts
    hyprpaper
    instances
    keyword
    kill
    layers
    layouts
    monitors
    notify
    plugin
    reload
    setcursor
    seterror
    setprop
    splash
    switchxkblayout
    version
    workspacerules
    workspaces

flags:
    -j -> output in JSON
    --batch -> execute a batch of commands, separated by ';'
    --instance (-i) -> use a specific instance. Can be either signature or index in hyprctl instances (0, 1, etc)

[I] root@neo ~ [1]
# 
```

So far not terrible. But individual commands should help the user to explore and figure out the usage.

![image](https://github.com/hyprwm/Hyprland/assets/6200749/81e933ae-11c0-4469-b506-2f99906cffb2)

vaxerski:
I mean that's what [the wiki](https://wiki.hyprland.org/Configuring/Using-hyprctl/#notify) is for no?

hholst80:
I think the CLI tooling should be helpful enough not to require me to have a working internet connection to do anything useful.

vaxerski:
fair enough I guess

JPDucky:
I agree heavily with this, trying to figure out more advanced usage of hyprctl is a nightmare as the current given documentation does not go into much detail.

vaxerski:
MRs welcome.

PR #5385, opened 2024-04-02
Improve hyprctl help pages

Micovec:

References #4300 

I hope I didn't miss anything. There's a lot of stuff hyprctl is capable of and some of the stuff isn't documented so I had to check for it in the code. Few descriptions are copy-pasted from the wiki.

As all help pages these are limited to 80 columns.

Some commands have several arguments which with their description didn't fit into the help page somewhat nicely so I replaced them with three dots. You can use `hyprctl [command] --help` to print additional help for the command. I'm not sure if it's the best solution but it was the most optimal I came up with.

Ready for merging

### DIFF
--- a/hyprctl/Strings.hpp
+++ b/hyprctl/Strings.hpp
@@ -1,6 +1,6 @@
 #pragma once
 
-const std::string_view USAGE = R"#(usage: hyprctl [(opt)flags] [command] [(opt)args|--help]
+const std::string_view USAGE = R"#(usage: hyprctl [flags] <command> [args...|--help]
 
 commands:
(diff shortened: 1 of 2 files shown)


fufexan (commented):
The `[]` everywhere should be replaced with `<>` where required, and `[]` where optional, just like most other help texts. This will also make the `(opt)` redundant, thus freeing some visual space.

fufexan:
Sure, that sounds good.

Micovec:
Done. Changed brackets.

khaneliman:
Love the update, thanks for doing this.

PR #5394, opened 2024-04-02
Fix master center resizing

thejch:
Fix a small resiziing bug on master layout, which causes the resizing to go the wrong direction


vaxerski (approved):
fair

Issue #5402, opened 2024-04-03
Hyprland crashes when quitting some applications


Ciflire:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 347b83903474ab9e95fa9d4ba123f6f78ab38d40  ().
Date: 2024-04-03
Tag:

flags: (if any)


System Information:
System name: Linux
Node name: nixos
Release: 6.8.2
Version: #1-NixOS SMP PREEMPT_DYNAMIC Tue Mar 26 22:23:34 UTC 2024


GPU information:
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA106M [GeForce RTX 3060 Mobile / Max-Q] [10de:2520] (rev a1) (prog-if 00 [VGA controller])
06:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Rembrandt [Radeon 680M] [1002:1681] (rev c8) (prog-if 00 [VGA controller])


os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.05.20240403.a1c35e4"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.05 (Uakari)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.05 (Uakari)"
VERSION_CODENAME=uakari
VERSION_ID="24.05"


plugins:
```

</details>

### Bug or Regression?

Regression

### Description

Hyprland crashes when quitting some applications, the ones i noted are:
- pavucontrol
- kitty
- alacritty
- konsole
- qterminal

### How to reproduce

Launch one of those applications
Quit it
Hyprland crashes

### Crash reports, logs, images, videos

[hyprlandCrashReport12059.txt](https://github.com/hyprwm/Hyprland/files/14852818/hyprlandCrashReport12059.txt)
[hyprlandCrashReport32720.txt](https://github.com/hyprwm/Hyprland/files/14852819/hyprlandCrashReport32720.txt)
[hyprlandCrashReport36745.txt](https://github.com/hyprwm/Hyprland/files/14852821/hyprlandCrashReport36745.txt)
[...]

PR #5409, opened 2024-04-03
Fix crash in isWorkspaceVisible

thejch:
Fixes https://github.com/hyprwm/Hyprland/issues/5402


vaxerski:
it'd be better to actually track down the root cause.

furthermore, use `valid(w)`

thejch:
changed

vaxerski:
I fixed it anyways in main with a more robust patch

thejch:
cool

PR #5414, opened 2024-04-03
Use proper path to update repo when processing user provided revision

zakk4223:
When doing a plugin update hyprpm wasn't using the correct directory when a user provided revision was used.

Yes


vaxerski (approved):
oops

PR #5428, opened 2024-04-04
IME: fix crash and auto enter

sungyoonc:
Fixes a crash with IME.
Fix auto entering a focused text input when new IME.

When the last focused surface before IME exit gets closed, it's not removed from `m_vTextInputs`. After this when you launch IME and try to move to a different text input, hyprland crashes.

Steps to reproduce
1. Open any application with text input v1 or v3
2. Exit/Kill IME
3. Close the application from step 1
4. Open two application with text input v1 or v3
5. Focus on one application and launch IME
6. Try change focus to another application
7. Crash

Not really

I think so


vaxerski (approved):
sure

PR #5433, opened 2024-04-04
Fix workspace rule loops

thejch:
Addressess https://github.com/hyprwm/Hyprland/pull/5340#issuecomment-2038081538
Merge the rules before returning.

breaking API change again smfh, but this wasnt the best design to begin with

ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1257,13 +1257,8 @@ void CCompositor::sanityCheckWorkspaces() {
     auto it = m_vWorkspaces.begin();
     while (it != m_vWorkspaces.end()) {
 
-        const auto WORKSPACERULES = g_pConfigManager->getWorkspaceRulesFor(*it);
-        bool       isPersistent   = false;
-        for (auto& wsRule : WORKSPACERULES) {
-            if (wsRule.isPersistent)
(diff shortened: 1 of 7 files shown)


vaxerski:
is pineapple an animal

thejch:
> I think the processing of onCreatedEmptyExecuted needs to use a vectorized version. Although I guess it depends on which behavior is wanted.
> 
> Before this PR it used the vectorized version and ran every rule that matched. Now it runs the last one. I kinda feel like running them all is more intuitive, but it didn't work that way prior to 0.38 anyways soooo
> 
> It might be ok to leave getWorkspaceRulesFor since removing it is going to break anyone that converted their code for 0.38. But that's not up to me ;)

first point makes sense to me
second point i'll wait for vaxry's input

vaxerski:
> It might be ok to leave getWorkspaceRulesFor since removing it is going to break anyone that converted their code for 0.38. But that's not up to me ;)

do I look like a person that cares that much about this one lmao

I am still waiting on input regarding my previous question

zakk4223:
pineapples are only animals if they scream when you cut them open

vaxerski (approved):
lgtm

PR #5466, opened 2024-04-06
Event loop rework

vaxerski:
This makes the event loop multi-threaded so we can plant our own timers with more precision than wayland allows.

Give a :+1: or comment if everything is alright

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2,6 +2,7 @@
 #include "helpers/Splashes.hpp"
 #include "config/ConfigValue.hpp"
 #include "managers/CursorManager.hpp"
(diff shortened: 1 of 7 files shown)


vaxerski:
cool, thanks! Merging then

ErrorNoInternet:
Found a regression; pressing a key for just under `repeat_delay` milliseconds makes it register twice. I have my repeat_delay set to 150 so this is extremely noticeable.

Reproducible with `ydotool key -d580 38:1 38:0` which will hold down the l key, wait 580 milliseconds then release it (assuming your repeat_delay is set to 600, the default). Might have to do it a few times to trigger it.

PR #5499, opened 2024-04-08
Add count group flag in windowCount workspace selector prop

### LINKED ISSUE #5488: workspace rule selector like window counts, but for 'containers'
### Description

Some way to have the functionality of say, w[t2] but only count the number of 'containers' that are visible in the workspace. I guess just counting groups as a single window would work?

sungyoonc:
Adds the groupCount workspace selector prop.
The groupCount is the same as the windowCount but it counts groups instead of windows in a workspace.
Resolves #5488

Workspace rules can now do
```ini
workspace = g[t1], bordersize:10
```

A comment here shows an example using windowCount with range `0-4`. But this is invalid according to the code. Did you meant to make `0-4` a valid range or just typo in the comment?
```cpp
// w - windowCount: w[0-4] or w[1], optional flag t or f for tiled or floating, e.g. w[t0-1]
```
```cpp
if (to < from || to < 1 || from < 1) {
    Debug::log(LOG, "Invalid selector {}", selector);
    return false;
}
```

Ready, but wiki mr needed.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1280,6 +1280,16 @@ int CCompositor::getWindowsOnWorkspace(const int& id, std::optional<bool> onlyTi
     return no;
 }
 
+int CCompositor::getGroupsOnWorkspace(const int& id, std::optional<bool> onlyTiled) {
+    int no = 0;
+    for (auto& w : m_vWindows) {
+        if (w->workspaceID() == id && w->m_bIsMapped && !(onlyTiled.has_value() && !w->m_bIsFloating != onlyTiled.value()) && w->m_sGroupData.head)
+            no++;
+    }
+
+    return no;
+}
+
 CWindow* CCompositor::getUrgentWindow() {
     for (auto& w : m_vWindows) {
         if (w->m_bIsMapped && w->m_bIsUrgent)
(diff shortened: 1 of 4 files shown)


vaxerski:
why not integrate it with w? `w[gt1-2]`

sungyoonc:
> why not integrate it with w? `w[gt1-2]`

How about this?

vaxerski:
okay, but I don't get why you're doing all that refactoring outside of what you are adding. I'd rather it wasn't there.

sungyoonc:
Removed all the refactoring bits. Should I make a separate mr or just scrap it?

vaxerski:
> Removed all the refactoring bits. Should I make a separate mr or just scrap it?

No, they are fine, just unrelated to this MR. Feel free to make it separately

vaxerski:
can you make a wiki mr?

sungyoonc:
Here hyprwm/hyprland-wiki#585

zakk4223:
Hmm, coming back to this now that it is in...

Are there any sets of selectors I can use that match tiled(windows+groups) = <num>?

(I guess you can kinda do it with multiple workspace rules?)

sungyoonc:
I think mixed counting flag(non grouped window + groups) need to be added to support that.

Issue #5496, opened 2024-04-08
Replace pkg-config with pkgconf in hyprpm (and anywhere else if relevant)


blacklightpy:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch v0.38.1-b at commit 360ede79d124ffdeebbe8401f1ac4bc0dbec2c91  (props: bump ver to 0.38.1).
Date: Sat Apr 6 14:24:13 2024
Tag: v0.38.1

flags: (if any)


System Information:
System name: Linux
Node name: void
Release: 6.6.21_1
Version: #1 SMP PREEMPT_DYNAMIC Fri Mar  8 01:01:43 UTC 2024


GPU information:


os-release: NAME="Void"
ID="void"
PRETTY_NAME="Void Linux"
HOME_URL="https://voidlinux.org/"
DOCUMENTATION_URL="https://docs.voidlinux.org/"
LOGO="void-logo"
ANSI_COLOR="0;38;2;71;128;97"

DISTRIB_ID="void"


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

As I mentioned in #5480, my Hyprland install was getting the missing header error, and updating did no help.

I tried debugging it for a while, and found out that the problem was getting `HEADERS_MISSING` after doing `headersValid()`

I figured out that the problems was in the line:
```cpp
std::string cmd     = std::format("PKG_CONFIG_PATH=\"{}/share/pkgconfig\" pkg-config --cflags --keep-system-cflags hyprland", DataState::getHeadersPath());
```

The problem was that `--keep-system-cflags` was not recognized by `pkg-config`, but it was by `pkgconf`. [`pkgconf`](https://github.com/pkgconf/pkgconf) is an active fork (last updated 2 weeks ago) of Freedesktop's [`pkg-config`](https://gitlab.freedesktop.org/pkg-config/pkg-config) (last updated 3 years ago). Arch Linux, FreeBSD, NetBSD, Fedora, Mageia and OpenSUSE had [moved to](https://lists.archlinux.org/pipermail/arch-dev-public/2018-May/029252.html) `pkgconf` by 2018, as a replacement for `pkg-config`.

`PKG_CONFIG_PATH="/home/user/.local/share/hyprpm/headersRoot/share/pkgconfig" pkgconf --cflags --keep-system-cflags hyprland` returned the three `-I/path`s.

[...]

PR #5502, opened 2024-04-08
Change pkg-config -> pkgconf for hyprpm and cmake

vaxerski:
Ref #5496

@fufexan for nix, you can merge when you're done


fufexan:
Pretty sure hyprpm doesn't work on Nix anyway but sure.

PR #5504, opened 2024-04-08
nix: install shell completions

postsolar:

Uses nixpkgs function [`installShellCompletion`](https://nixos.org/manual/nixpkgs/stable/#installshellfiles) to install the newly-added Bash, Fish and Zsh completions for `hyprctl` and `hyprpm` commands

Nope

Ready

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -3,6 +3,7 @@
   stdenv,
   pkg-config,
   pkgconf,
+  installShellFiles,
   makeWrapper,
   meson,
   ninja,
@@ -86,6 +87,7 @@ in
     '';
 
     nativeBuildInputs = [
+      installShellFiles
       jq
       makeWrapper
       meson
@@ -149,6 +151,16 @@ in
           pkgconf
         ]}
       ''}
+
+      installShellCompletion --cmd hyprctl \
+        --bash $src/hyprctl/hyprctl.bash   \
+        --fish $src/hyprctl/hyprctl.fish   \
+        --zsh  $src/hyprctl/hyprctl.zsh
+
+      installShellCompletion --cmd hyprpm \
+        --bash $src/hyprpm/hyprpm.bash    \
+        --fish $src/hyprpm/hyprpm.fish    \
+        --zsh  $src/hyprpm/hyprpm.zsh
     '';
 
     passthru.providedSessions = ["hyprland"];


fufexan:
As far as I know they're already being handled by Meson. Is there a reason to use this instead?

fufexan:
In my case, hyprctl ends up in `/run/current-system/sw/share/zsh/site-functions/_hyprctl`. As for `hyprpm`'s completions, I think I forgot to add them to meson.

postsolar:
Ok, I think the ZSH ones *do* end up in `/etc/profiles/per-user/me/share/zsh/site-functions/_hyprctl` for me. But the fish ones do not. I also notice that in the resulting derivation the path used is incorrect (`/nix/store/<hash>-hyprland-0.38.0+date=2024-04-09_d0d1ba5/share/fish/completions/hyprctl.fish`), where that last directory should be `vendor_completions.d`. But in the Makefile it seems to be correct, so I'm now confused.

https://github.com/hyprwm/Hyprland/blob/d0d1ba5918206f16128e56988dd4db5d6d143338/Makefile#L47

postsolar:
Ok, I think the ZSH ones *do* end up in `/etc/profiles/per-user/me/share/zsh/site-functions/_hyprctl` for me. But the fish ones do not. I also notice that in the resulting derivation the path used is incorrect (`/nix/store/<hash>-hyprland-0.38.0+date=2024-04-09_d0d1ba5/share/fish/completions/hyprctl.fish`), where that last directory should be `vendor_completions.d`. But in the Makefile it seems to be correct, so I'm now confused.

https://github.com/hyprwm/Hyprland/blob/d0d1ba5918206f16128e56988dd4db5d6d143338/Makefile#L47

ErrorNoInternet:
Modifying hyprctl/meson.build seems to have fixed it for me.
image

postsolar:
@ErrorNoInternet nice find! Thank you. Would you mind sending a quick patch then?

ErrorNoInternet:
Sure, opened at #5510

leon-erd:
For me the completions are also not working on nixos for both zsh and bash.


> In my case, hyprctl ends up in `/run/current-system/sw/share/zsh/site-functions/_hyprctl`. As for `hyprpm`'s completions, I think I forgot to add them to meson.

I can also find them in `/run/current-system/sw/share/zsh/site-functions/_hyprctl`. Do they work for you out of the box @fufexan or do I have to enable them somehow?

fufexan:
@leon-erd I think what makes them work is adding this line to the nixos config.
```nix
  # enable zsh autocompletion for system packages (systemd, etc)
  environment.pathsToLink = ["/share/zsh"];
```

leon-erd:
Hm still not working for me unfortunately...

postsolar:
> I can also find them in /run/current-system/sw/share/zsh/site-functions/_hyprctl.

Should work OOTB, at least it does for me. Do you have this path displayed if you do `print -l $fpath`?

leon-erd:
Yup its there
```bash
➜  ~ print -l $fpath | grep /run/current-system/sw/share/zsh/
/run/current-system/sw/share/zsh/site-functions
/run/current-system/sw/share/zsh/5.9/functions
/run/current-system/sw/share/zsh/vendor-completions
```

postsolar:
Does it work if you explicitly instruct zsh to use it?
```console
$ autoload _hyprctl
$ compdef _hyprctl hyprctl
```

If yes then I think you maybe have an outdated `.zcompdump` file, if you can find it in your `$HOME` or `$ZDOTDIR` then deleting it and restarting ZSH might help.

leon-erd:
You are a genius! Deleting the `.zcompdump` actually worked :partying_face:

PR #5518, opened 2024-04-09
Expand on window matching

SoSeDiK:
Adds `activewindow`, `initialtitle` and `initialclass` as valid targets for window matching. Also allows explicitly specifying `class:`.

\-

Ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2410,9 +2410,8 @@ void CCompositor::scheduleFrameForMonitor(CMonitor* pMonitor) {
 }
 
 CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
-    if (regexp == "activewindow") {
+    if (regexp.starts_with("active"))
         return m_pLastWindow;
-    }
 
     eFocusWindowMode mode = MODE_CLASS_REGEX;
 
@@ -2464,7 +2463,7 @@ CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
                 break;
             }
             case MODE_INITIAL_CLASS_REGEX: {
-                const auto initialWindowClass = w.get()->m_szInitialClass;
+                const auto initialWindowClass = w->m_szInitialClass;
                 if (!std::regex_search(initialWindowClass, regexCheck))
                     continue;
                 break;
@@ -2476,7 +2475,7 @@ CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
                 break;
             }
             case MODE_INITIAL_TITLE_REGEX: {
-                const auto initialWindowTitle = w.get()->m_szInitialTitle;
+                const auto initialWindowTitle = w->m_szInitialTitle;
                 if (!std::regex_search(initialWindowTitle, regexCheck))
                     continue;
                 break;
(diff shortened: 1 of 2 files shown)


src/Compositor.cpp
```diff
@@ -2389,13 +2389,25 @@ void CCompositor::scheduleFrameForMonitor(CMonitor* pMonitor) {
 }
 
 CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
+    if (regexp == "activewindow") {
```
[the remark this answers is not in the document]
SoSeDiK: Removed {}, oops!

I'd like to reiterate on the other proposals though, since I'm not confident enough it'd be a good change. While options are great to have, I would consider these ones an unneeded clutter.
1) Aliases can be neat in some cases (e.g. color vs colour / --list vs -l), but having *that much* different things doing the same thing is not IMHO. Opinionated, but makes sharing things more cumbersome. Someone could share their rules, and you'd have to remember to change them to your liking or end up with a mixed config, while there could simply be one entry that everyone uses.
2) Pretty much the same as 1), but also allows for crazy stuff like `activewindowRand0mStuuf`, which I'm not sure when would be actually preferred over the explicit rule.

As for the current naming, I went with `activewindow` initially to be consistent with the hyprctl's activewindow dispatcher (I think it's quite intuitive).
Will add proposals if preferred, though

src/Compositor.cpp
```diff
@@ -2389,13 +2389,24 @@ void CCompositor::scheduleFrameForMonitor(CMonitor* pMonitor) {
 }
 
 CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
+    if (regexp == "activewindow")
```
vaxerski, line 2392: `.starts_with("active")` makes more sense

src/Compositor.cpp
```diff
@@ -2436,6 +2453,12 @@ CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
                     continue;
                 break;
             }
+            case MODE_INITIAL_TITLE_REGEX: {
+                const auto initialWindowTitle = w.get()->m_szInitialTitle;
```
vaxerski, line 2457: `w->`

src/Compositor.cpp
```diff
@@ -2426,7 +2437,13 @@ CWindow* CCompositor::getWindowByRegex(const std::string& regexp) {
         switch (mode) {
             case MODE_CLASS_REGEX: {
                 const auto windowClass = g_pXWaylandManager->getAppIDClass(w.get());
-                if (!std::regex_search(g_pXWaylandManager->getAppIDClass(w.get()), regexCheck))
+                if (!std::regex_search(windowClass, regexCheck))
+                    continue;
+                break;
+            }
+            case MODE_INITIAL_CLASS_REGEX: {
+                const auto initialWindowClass = w.get()->m_szInitialClass;
```
vaxerski, line 2445: `w->`

vaxerski:
sorry for the delay, this got lost in my inbox somehow

vaxerski (approved):
lgtm, wiki mr needed

SoSeDiK:
> lgtm, wiki mr needed

https://github.com/hyprwm/hyprland-wiki/pull/587

PR #5530, opened 2024-04-10
Improve Hyprpm Update Performance

https123456789:

Improves the performance of the `hyprpm update` command by reducing the amount of data fetched over the network. The current version of hyprpm clones the entire Hyprland repo to do an update. This includes tens of thousands objects that are not used or referenced and are deleted after building the headers. In addition, the tracy submodule is also cloned, however it is never included in the build as hyprpm always builds headers in release mode. This PR changes these behaviors by:
- Using a shallow clone based on only history since the currently running commit, dramatically reducing the number of objects that need to be fetched (from a couple thousand to ~300 for any commit within the past couple months).
- Removing unused submodules (tracy) to avoid expensive cloning.

This performance improvement is only really noticeable on systems that have a slower internet connection (like mine). For me, running `hyprpm update` took around 4.5 minutes before the changes and around 2.4 minutes after.

Ready for merging.

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -70,6 +70,9 @@ SHyprlandVersion CPluginManager::getHyprlandVersion() {
     std::string hlbranch = HLVERCALL.substr(HLVERCALL.find("from branch") + 12);
     hlbranch             = hlbranch.substr(0, hlbranch.find(" at commit "));
 
+    std::string hldate = HLVERCALL.substr(HLVERCALL.find("Date: ") + 6);
+    hldate = hldate.substr(0, hldate.find("\n"));
+
     std::string hlcommits;
 
     if (HLVERCALL.contains("commits:")) {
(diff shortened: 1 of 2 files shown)


hyprpm/src/core/PluginManager.cpp
```diff
@@ -398,7 +401,7 @@ bool CPluginManager::updateHeaders(bool force) {
 
     progress.printMessageAbove(std::string{Colors::YELLOW} + "!" + Colors::RESET + " Cloning https://github.com/hyprwm/hyprland, this might take a moment.");
 
-    std::string ret = execAndGet("cd /tmp/hyprpm && git clone --recursive https://github.com/hyprwm/hyprland hyprland");
+    std::string ret = execAndGet("cd /tmp/hyprpm && git clone --recursive https://github.com/hyprwm/hyprland hyprland --shallow-since='" + HLVER.date + "'");
```
vaxerski, line 404: this won't work, unfortunately. A bunch of cases will violate this:
 - packagers might compile a release days or weeks later
 - source-based compiles still may use releases (gentoo etc)
 
 In these cases, the date will be too far ahead to clone their release

https123456789: The date returned by `hyprctl version` is the date that the commit was made, not the date of compilation. Therefore the clone operation would work regardless of when a release is build.

https://github.com/hyprwm/Hyprland/blob/303b9956b2ae15508b09dffae602550ca17e6539/src/debug/HyprCtl.cpp#L814-L815

Please correct me if I am wrong.

vaxerski: right, it is. My bad.

vaxerski (approved):
thanks :)

PR #5534, opened 2024-04-10
Fallback to xcursor if failed to render hyprcursor

SoSeDiK:
Fallback to xcursor as last resort. Should bandaid invisible cursors on broken themes.

\-

Ready

### DIFF
--- a/src/managers/CursorManager.cpp
+++ b/src/managers/CursorManager.cpp
@@ -122,6 +122,7 @@ void CCursorManager::setCursorFromName(const std::string& name) {
 
         if (m_sCurrentCursorShapeData.images.size() < 1) {
             Debug::log(ERR, "BUG THIS: No fallback found for a cursor in setCursorFromName");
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
is this a good idea though? An Xcursor theme might not be in use at all - which would result in a random ugly cursor, instead of the old one staying...?

SoSeDiK:
This is an edge case when no cursor could be found at all, so the expectation is that either the fancy cursor didn't load in the first place and mouse will be invisible without this, or something went so horribly wrong that fancy cursor is not a priority.

The proper fix is to change/repair a broken theme, but this should help those who had invisible cursor in e.g. kitty and had to hover Waybar for cursor to reappear.

SoSeDiK:
https://github.com/hyprwm/hyprcursor/pull/31 should fix invisible cursors too during initialization. I still think *this* is a good caution in case it ever breaks during runtime.

vaxerski:
I guess it's okay since if you dont have a left_ptr your theme is cringe

vaxerski (approved):

PR #5548, opened 2024-04-11
Added `toggleswallow` dispatcher

tomben13:
New feature: `toggleswallow` dispatcher which allows to toggle the swallowed window with a keybind. Only if the window was originally swallowed will it work. Meaning that you can only "unswallow" and then "swallow" again a window, and so on.

Useful when you want to have your terminal swallowed, but sometimes you also want to check the logs of your running application.

Ready

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -385,6 +385,9 @@ void Events::listener_mapWindow(void* owner, void* data) {
     // Verify window swallowing. Get the swallower before calling onWindowCreated(PWINDOW) because getSwallower() wouldn't get it after if PWINDOW gets auto grouped.
     const auto SWALLOWER  = PWINDOW->getSwallower();
     PWINDOW->m_pSwallowed = SWALLOWER;
+    if (PWINDOW->m_pSwallowed) {
+        PWINDOW->m_pSwallowed->m_bCurrentlySwallowed = true;
+    }
 
     if (PWINDOW->m_bIsFloating) {
         g_pLayoutManager->getCurrentLayout()->onWindowCreated(PWINDOW);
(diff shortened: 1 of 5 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -1820,6 +1821,28 @@ void CKeybindManager::focusWindow(std::string regexp) {
     g_pCompositor->warpCursorTo(PWINDOW->middle());
 }
 
+void CKeybindManager::toggleSwallow(std::string args) {
+    CWindow* pWindow = g_pCompositor->m_pLastWindow;
+    if (pWindow->m_pSwallowed && g_pCompositor->windowExists(pWindow->m_pSwallowed)) {
```
vaxerski, line 1826: windowExists should be avoided in favor of methods that don't leave dangling pointers, like e.g. removing them in ::onUnmap

tomben13: I'm not sure what to replace it with. I used the same condition found in `src/events/windows.cpp:690`.
Should I remove the check altogether ?
Is it safe to do so and assume the pointer isn't dangling ? `if (pWindow->m_pSwallowed) {`

vaxerski: I think it should be valid if you add a method to remove the `swallowed` ptr in `::onUnmap` (basically when a window is unmapped, loop through all windows and set swallowed to null wherever it finds a ref)

src/events/Windows.cpp
```diff
@@ -687,10 +688,16 @@ void Events::listener_unmapWindow(void* owner, void* data) {
     g_pHyprOpenGL->makeWindowSnapshot(PWINDOW);
 
     // swallowing
-    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed)) {
+    if (PWINDOW->m_pSwallowed) {
```
vaxerski, line 691: what about previouslySwallowed? Also reset this :P

src/managers/KeybindManager.cpp
vaxerski: sure

src/events/Windows.cpp
tomben13: Replaced with a `bool`. Is there a better method to handle the hide and unhide in C++ ?

vaxerski: why not check `isHidden()`? If swallowed != null and swallowed->ishidden(), it's swallowed, if it's not hidden, it's not

tomben13: I thought about this, but it could be hidden for other reasons, like if another window is fullscreen, no ?

MightyPlaza: grouped windows are hidden

daniel-hernandez-sony:
Would love to see this coming online :)

tomben13:
Hi @vaxerski ,
I have reworked the PR.
It's simplified because `PWINDOW->m_pSwallowed` cannot be dangling now.
Using `bool m_bCurrentlySwallowed` instead of `isHidden()` to work with grouped windows.

src/events/Windows.cpp
```diff
@@ -385,6 +385,9 @@ void Events::listener_mapWindow(void* owner, void* data) {
     // Verify window swallowing. Get the swallower before calling onWindowCreated(PWINDOW) because getSwallower() wouldn't get it after if PWINDOW gets auto grouped.
     const auto SWALLOWER  = PWINDOW->getSwallower();
     PWINDOW->m_pSwallowed = SWALLOWER;
+    if (PWINDOW->m_pSwallowed) {
```
vaxerski, line 388: no {}

vaxerski (commented):
rest ok

vaxerski (approved):
wiki mr needed

PR #5554, opened 2024-04-11
Fix crash when when swallowed window is closed

Bartkk0:
Fixes a crash when a swallowed window is closed. From what I've discovered it happens randomly depending on which window is closed first, at least it appears to be the cause.

I don't know if it's the right way to fix this crash, but it works.

Ready

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -687,7 +687,7 @@ void Events::listener_unmapWindow(void* owner, void* data) {
     g_pHyprOpenGL->makeWindowSnapshot(PWINDOW);
 
     // swallowing
-    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed)) {
+    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed) && !PWINDOW->m_pSwallowed->m_bReadyToDelete) {
         PWINDOW->m_pSwallowed->setHidden(false);
         g_pLayoutManager->getCurrentLayout()->onWindowCreated(PWINDOW->m_pSwallowed);
         PWINDOW->m_pSwallowed = nullptr;


src/events/Windows.cpp
```diff
@@ -687,7 +687,7 @@ void Events::listener_unmapWindow(void* owner, void* data) {
     g_pHyprOpenGL->makeWindowSnapshot(PWINDOW);
 
     // swallowing
-    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed)) {
+    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed) && !PWINDOW->m_bReadyToDelete) {
```
vaxerski, line 690: I dont think the last condition will ever be false

Bartkk0: Yeah, I messed up

src/events/Windows.cpp
```diff
@@ -687,7 +687,7 @@ void Events::listener_unmapWindow(void* owner, void* data) {
     g_pHyprOpenGL->makeWindowSnapshot(PWINDOW);
 
     // swallowing
-    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed)) {
+    if (PWINDOW->m_pSwallowed && g_pCompositor->windowExists(PWINDOW->m_pSwallowed) && !PWINDOW->m_pSwallowed->m_bReadyToDelete) {
```
vaxerski, line 690: check should be for m_bIsMapped

Bartkk0: Tried it, doesn't work

vaxerski:
should be fixed by #5548 I think

Bartkk0:
Yeah, compiled the PR and Hyprland no longer crashes, closing.

PR #5566, opened 2024-04-12
Add variable to focus on a noinitialfocus window if the workspace is empty

c4llv07e:
I have configured `windowrule=noinitialfocus,(.*)`, so when I open a new window, hyprland doesn't focus on it. This is good when you are using some app and don't want a popup to steal focus.
This works as intended, but I think there is no reason not to focus on a new window if there is no focused one.
This PR adds a new variable `misc:focus_noinitial_if_no_window`, which allows to change this behaviour. It will focus on a new window even if it is marked as `noinitialfocus` when the workspace is empty.

I can't think a better name for this variable. Apart from that, I can't find any bugs/problems.

If you good with this name, it is ready to merge, I think.


vaxerski:
`windowrulev2 = noinitialfocus,onworkspace:[t1-999]`?

c4llv07e:
> `windowrulev2 = noinitialfocus,onworkspace:[t1-999]`?

I have never thought about it in this way...

`windowrulev2=noinitialfocus,onworkspace:w[t2-999]` works, PR isn't required anymore.

PR #5582, opened 2024-04-14
remove libsystemd dependency

cnt0:
Hello. This PR introduces home-baked `sd_booted` and `sd_notify` functions, so it's now possible to completely remove the `libsystemd` dependency as per Lennart Poettering's [advice](https://github.com/systemd/systemd/issues/32028#issuecomment-2031366922)

> I'd recommend apps to just embed their own minimal version of the sd_notify() protocol, in their native language. It's trivial, and nowadays even in crazy languages like Java you have AF_UNIX support that is sufficient to implement it in a few lines of code.

I copied `sd_booted` almost as-is from libsystemd. As for `sd_notify`, I've been looking at [python implementation](https://github.com/bb4242/sdnotify/blob/master/sdnotify/__init__.py) and its [documented behavior](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html).

This code is entirely untested, other than "it compiles", so please test it before merging (if you'd like to merge this PR, that is)


src/helpers/SdDaemon.hpp
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+int sd_booted(void);
```
vaxerski, line 3: if we are rewriting this, some style things to consider.

Firstly, this could be wrapped in a namespace e.g. `namespace Systemd`

Secondly, no_snake, we use camelCase

PR #5587, opened 2024-04-14
Fix xwl cursor locking for scaled monitors

LivingCodeX:

When an xwayland app locks the cursor, xwayland:force_zero_scaling=true is set and the monitor is scaled in the config, the xwayland app will use the unscaled display size to calculate the cursor position within the window. For scales greater than 1, this position can be off or out of bounds, so it will not warp the cursor at all.

This PR fixes this behavior.

This fix has only been tested with Minecraft, but as long as other xwayland apps work the same, it should be fine.

Also, I'm not sure if m_pOwner->m_fLastScale in line 70 is the ideal way to get the scale of the monitor the app window is in.

Ready for merging

### DIFF
--- a/src/desktop/Constraint.cpp
+++ b/src/desktop/Constraint.cpp
@@ -67,14 +67,14 @@ void CConstraint::onCommit() {
 
         m_bHintSet = true;
 
-        auto SCALE = 1.f;
-        const auto window = m_pOwner->getWindow();
-        if (window) {
-            const auto ISXWL = window->m_bIsX11;
-            SCALE = ISXWL && *PXWLFORCESCALEZERO ? window->m_fX11SurfaceScaledBy : 1.f;
+        float scale = 1.f;
+        const auto PWINDOW = m_pOwner->getWindow();
+        if (PWINDOW) {
+            const auto ISXWL = PWINDOW->m_bIsX11;
+            scale = ISXWL && *PXWLFORCESCALEZERO ? PWINDOW->m_fX11SurfaceScaledBy : 1.f;
         }
 
-        m_vPositionHint = {m_pConstraint->current.cursor_hint.x / SCALE, m_pConstraint->current.cursor_hint.y / SCALE};
+        m_vPositionHint = {m_pConstraint->current.cursor_hint.x / scale, m_pConstraint->current.cursor_hint.y / scale};
         g_pInputManager->simulateMouseMovement();
     }
 
--- a/src/desktop/Constraint.cpp
+++ b/src/desktop/Constraint.cpp
@@ -67,11 +67,11 @@ void CConstraint::onCommit() {
 
         m_bHintSet = true;
 
-        float scale = 1.f;
+        float      scale   = 1.f;
         const auto PWINDOW = m_pOwner->getWindow();
         if (PWINDOW) {
             const auto ISXWL = PWINDOW->m_bIsX11;
-            scale = ISXWL && *PXWLFORCESCALEZERO ? PWINDOW->m_fX11SurfaceScaledBy : 1.f;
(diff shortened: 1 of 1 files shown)


src/desktop/Constraint.cpp
```diff
@@ -62,8 +63,12 @@ void CConstraint::onCommit() {
     const auto COMMITTED = m_pConstraint->current.committed;
 
     if (COMMITTED & WLR_POINTER_CONSTRAINT_V1_STATE_CURSOR_HINT) {
-        m_bHintSet      = true;
-        m_vPositionHint = {m_pConstraint->current.cursor_hint.x, m_pConstraint->current.cursor_hint.y};
+        static auto PXWLFORCESCALEZERO = CConfigValue<Hyprlang::INT>("xwayland:force_zero_scaling");
+
+        m_bHintSet       = true;
+        const auto ISXWL = m_pOwner->getWindow()->m_bIsX11;
```
vaxerski, line 69: getWindow() may be null

src/desktop/Constraint.cpp
```diff
@@ -62,8 +63,12 @@ void CConstraint::onCommit() {
...
+        const auto SCALE = ISXWL && *PXWLFORCESCALEZERO ? m_pOwner->m_fLastScale : 1;
```
vaxerski, line 70: you should use the window property for xwayland scale, forgot the name, iirc x11scaledby

src/desktop/Constraint.cpp
LivingCodeX: Fixed. Btw this is my first PR. Is it considered bad to commit to the main branch for a PR? Also, am I generally supposed to resolve the conversation or should I leave it to you?

src/desktop/Constraint.cpp
LivingCodeX: Got it.

src/desktop/Constraint.cpp
LivingCodeX: Is window in lower case fine?

vaxerski: > Is it considered bad to commit to the main branch for a PR?

No

> Also, am I generally supposed to resolve the conversation or should I leave it to you?

For me it's whatever

src/desktop/Constraint.cpp
```diff
@@ -62,8 +63,18 @@ void CConstraint::onCommit() {
...
 
     if (COMMITTED & WLR_POINTER_CONSTRAINT_V1_STATE_CURSOR_HINT) {
-        m_bHintSet      = true;
-        m_vPositionHint = {m_pConstraint->current.cursor_hint.x, m_pConstraint->current.cursor_hint.y};
+        static auto PXWLFORCESCALEZERO = CConfigValue<Hyprlang::INT>("xwayland:force_zero_scaling");
+
+        m_bHintSet = true;
+
+        auto SCALE = 1.f;
+        const auto window = m_pOwner->getWindow();
```
vaxerski, line 71: style nit: const values in caps + type
`PWINDOW`

vaxerski (approved):
rest lgtm

src/desktop/Constraint.cpp
LivingCodeX: K, does that also apply to ISXWL and other primitives except pointers?

vaxerski: type isn't a hard mandatory but caps is

LivingCodeX: I also changed scale to lower case since it's not a const value

vaxerski (approved):
thanks!

vaxerski:
needs a clang-format it seems

vaxerski (approved):

PR #5599, opened 2024-04-15
notify: fix notifications visibility on manually rotated monitor

FUFSoB:

If you rotate monitor with eg `hyprctl keyword monitor HDMI-A-1,highrr,0x0,1,transform,3`, you then cannot see any notifications produced by `hyprctl notify` on the rotated monitor.

Fix is pretty simple: use `vecTransformedSize` instead of `vecPixelSize`

I also added extra check if monitor has changed its rotation/size. Without this check number of notifs on screen is limited by monitor's initial height (for 1080 with default settings that number is 37).

Cannot test with multiple monitors, but this shouldn't be a problem.

Should be ready.


vaxerski (approved):
lgtm, thanks!

Issue #3806, opened 2023-11-08
Adding noise on Nvidia only darkens the window background and creates dotted diagonal lines


MathisP75:
### Hyprland Version

Hyprland, built from branch main at commit 751d2851cc270c3322ffe2eb83c156e4298a0c0e dirty (props: bump ver to 0.32.0).

### Bug or Regression?

Bug

### Description

`noise = 0.0`
![image](https://github.com/hyprwm/Hyprland/assets/98901170/e8e8771c-6e33-4892-af64-abcf1e265027)

`noise = 1.0`
![image](https://github.com/hyprwm/Hyprland/assets/98901170/fc53dc42-2b39-4cb6-b5c0-ca8c86373a45)

### How to reproduce

On an Nvidia GPU, set noise to a value higher than 0 and see the darkness increase as you increase the value. No noise is being generated. The dotted lines also become more visible.

MathisP75:
My card is a 1660ti. Applies on `nvidia-dkms`for both 535 and 545.

MathisP75:
Another user reproduced the issue on a 3090
![image](https://github.com/hyprwm/Hyprland/assets/98901170/8d2ec768-d668-42a8-928a-af11c016075c)

This is his result for a `pacman -Qs nvidia`:
```
local/cuda 12.3.0-1
    NVIDIA's GPU programming toolkit
local/egl-wayland 2:1.1.13-1
    EGLStream-based Wayland external platform
local/ffnvcodec-headers 12.1.14.0-1
    FFmpeg version of headers required to interface with Nvidias codec APIs
local/lib32-nvidia-utils 535.113.01-1
    NVIDIA drivers utilities (32-bit)
local/libva-nvidia-driver-git 0.0.11.r1.gea6d862-1
    VA-API implementation that uses NVDEC as a backend (git version)
local/libvdpau 1.5-2
    Nvidia VDPAU library
local/libxnvctrl 545.29.02-1
    NVIDIA NV-CONTROL X extension
local/nvidia-open-dkms 535.113.01-8
    NVIDIA open kernel modules
local/nvidia-utils 535.113.01-2
    NVIDIA drivers utilities
local/opencl-nvidia 535.104.05-1
    OpenCL implemention for NVIDIA
```

MathisP75:
This is mine:
```
local/egl-wayland 2:1.1.13-1
    EGLStream-based Wayland external platform
local/ffnvcodec-headers 12.1.14.0-1
[...]

PR #5607, opened 2024-04-16
shaders: Use sin-less hash for noise

Epikastema:
Changes the hash function in the noise shader to avoid sin(). Fixes diagonal lines from noise observed in #3806.

Needs testing on older nvidia hardware/drivers, and AMD/Intel hardware to make sure it works on all platforms. I have only tested on 4000 series GPUs with the 550 driver.

### DIFF
--- a/src/render/shaders/Textures.hpp
+++ b/src/render/shaders/Textures.hpp
@@ -400,7 +400,9 @@ uniform float     noise;
 uniform float     brightness;
 
 float hash(vec2 p) {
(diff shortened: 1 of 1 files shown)


MahouShoujoMivutilde:
Gtx 970 on 550.67 works fine, lines are gone.

<details>
  <summary>Before and after</summary>





</details>

vaxerski (approved):
lgtm

PR #5648, opened 2024-04-19
Hyprland session target example

Loara:

A simple Hyprland user session target that automatically calls `graphical-session-pre.target` and `graphical-session.target` when launched by `systemctl --user`. A lot of Wayland applications usually ship their own user units that depends on `graphical-session.target` in order to be executed only when a graphical environment is running. Hyprland doesn't automatically ships its own session target so users have to explicitly execute installed systemd units with `exec-once = systemctl --user start ...` without the ability of enabling/disabling them selectively without modifying each time the configuration file.

For these reasons a lot of users prefer to manually create an `hyprland-session.target` unit file and start it with `exec-once = systemctl --user start hyprland-session.target` (you can't directly start `graphical-session.target` via `systemctl`). But this approach has an important drawback: when you exit from Hyprland without logging out from your user session then neither the graphical session target nor all the unit services depending on it would stop. There are two possible ways to solve this:
- automatically integrate `hyprland-session.target` in Hyprland code;
- allow users to add custom hook that should be executed before Hyprland exits, currently one could do this by defining custom key bindings but this wouldn't work when `hyprctl exit` is called elsewhere.

For the moment I only provide a simple `hyprland-session.target` file which is not automatically installed by meson in order to avoid overwriting custom target units. In this way users and package managers can freely decide whenever use it or not.

Automatically works


fufexan:
I've been thinking about adding an `exec-exit` hook too, but that won't work when Hyprland crashes, so it doesn't fully solve the issue.
Another thing that may help is polling `pgrep Hyprland` every few seconds and stop the target when it's no longer found. It's dirty but it may work.

Loara:
For `hyprland-session.target` we could solve it by simply execute `Hyprland; systemctl --user stop hyprland-session.target` in place of `Hyprland` only, so the target would be stopped whenever Hyprland exit.

A more elegant way is to write an `hyprland.service` unit service that both executes `Hyprland` and starts/stops any user unit bounded to `hyprland-session.target` when Hyprland exists or crashes. But at the moment I don't think Hyprland could be executed as a Systemd service.

fufexan:
Also, related sway discussion https://github.com/swaywm/sway/issues/5160.

I believe the first point you suggested might be the way to go. Additionally, distro packagers can patch https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.desktop to add `systemctl --user stop hyprland-session.target` for systemd/other init systems.

Loara:
I agree, the first solution is the most portable one. However, I think is a good thing to provide this `hyprland-session.target` file along with a patched `hyprland.desktop` and the following (commented) line in the default configuration file of Hyprland so that users and package managers can decide to use or not the Hyprland session target:

    # exec-once = systemctl --user start hyprland-session.target

Loara:
Put in another repository.

luigir-it:
> I agree, the first solution is the most portable one. However, I think is a good thing to provide this `hyprland-session.target` file along with a patched `hyprland.desktop` and the following (commented) line in the default configuration file of Hyprland so that users and package managers can decide to use or not the Hyprland session target:
>
> ```
> # exec-once = systemctl --user start hyprland-session.target
> ```

According to [this](https://github.com/swaywm/sway/wiki/Systemd-integration#managing-user-applications-with-systemd) you should run `systemctl --user start hyprland-session.target` after importing the environment; this way you avoid issues with user services that rely on that environment.

So a better option would be:
`# exec-once = dbus-update-activation-environment --systemd --all; systemctl --user start hyprland-session.target`

Loara:
> > I agree, the first solution is the most portable one. However, I think is a good thing to provide this `hyprland-session.target` file along with a patched `hyprland.desktop` and the following (commented) line in the default configuration file of Hyprland so that users and package managers can decide to use or not the Hyprland session target:
> > ```
> > # exec-once = systemctl --user start hyprland-session.target
> > ```
> 
> According to [this](https://github.com/swaywm/sway/wiki/Systemd-integration#managing-user-applications-with-systemd) you should run `systemctl --user start hyprland-session.target` after importing the environment; this way you avoid issues with user services that rely on that environment.
> 
> So a better option would be: `# exec-once = dbus-update-activation-environment --systemd --all; systemctl --user start hyprland-session.target`

Added [here](https://github.com/Loara/hyprland-target).

luigir-it:
What I meant, at least according to that wiki page, is that the two commands should be executed one after the other. This way, the target and all units that depend on it will inherit the full environment. If you use two different exec-once, there is a chance that the second command will finish before the first one, effectively inheriting just part of the environment or none at all.

So the preferred way would be like you did on [your first commit](https://github.com/Loara/hyprland-target/commit/8e212c496aef2706d56f559076077efe03de8a6d)

Loara:
> What I meant, at least according to that wiki page, is that the two commands should be executed one after the other. This way, the target and all units that depend on it will inherit the full environment. If you use two different exec-once, there is a chance that the second command will finish before the first one, effectively inheriting just part of the environment or none at all.
> 
> So the preferred way would be like you did on [your first commit](https://github.com/Loara/hyprland-target/commit/8e212c496aef2706d56f559076077efe03de8a6d)

Fixed in the README.

Issue #5573, opened 2024-04-13
Add More Monitor 'Auto' Positions


modestbadger1773:
### Description

Currently the `auto` position of a secondary monitor is 'Right' of the main monitor.

The solution to this is to place the monitor at the position you would like it to be in your config file. But, if you would always want the second monitor to be 'above' your main monitor, this can be annoying (as you will need to activate a new line for every time a monitor of new resolution is attached).

The suggested enhancement would be to add the following options available for monitor configuration:

`auto-up`: Automatically places the monitor at a position above the primary monitor.
`auto-left`: Automatically places the monitor at a position left of the primary monitor.
`auto-down`: Automatically places the monitor at a position below the primary monitor.
`auto-right` Automatically places the monitor at a position right of the primary monitor. (This is the current behaviour of `auto`).

I think `auto` should itself remain an option, to avoid breaking configs, even though it would be redundant with the existence of `auto-right`

----
An argument could be made that the option should exist for the 8 main directions `up`, `up-left`, `up-right`, `left`, `right`, `down`, `down-left`, & `down-right`. But just having the `auto-up` option would solve my use case, so /shrug.

The-Briel-Deal:
Thanks for the feature request! Working on this now, should hopefully have a PR up by EOD.

The-Briel-Deal:
PR Created https://github.com/hyprwm/Hyprland/pull/5670. I only implemented the 4 new directions as the other 4 directions suggested have a few problems in concept. I described the issue deeper in my PR description. Thanks again for the feature request!

PR #5670, opened 2024-04-21
Add More Monitor 'Auto' Positions.

The-Briel-Deal:
Resolves https://github.com/hyprwm/Hyprland/issues/5573.

The objective of this PR is to add more automatic monitor placement options. Currently, you can specify a monitors position as `auto` and it will automatically place the monitor to the right. It was requested however that we enable automatically placing monitor in other directions. So this PR adds the ability to place monitors up, down, left, and right with the keywords `auto-up`, `auto-down`, `auto-left`, and `auto-right`.

It was also mentioned that having corner directions would be desirable as well. But adding this leads to significant ambiguity as to what would actually happen. For example with `auto-up-right` where would the monitors connect? On the right top or right side? What happens if this monitor overlaps with another monitor due to an odd resolution or aspect ratio?

So, in this PR I kept things simple which I believe will cover most general use cases without introducing many new edge cases.

That said, in the future I would like to propose further improvements to the monitor placement system. For example, I think it would be nice to make a visual tui curses tool to align your monitors with a set scale and resolution. I also think we should implement a way to correct or warn on poor monitor placement if a user accidentally overlaps monitors or makes it so that they aren't connected.

I have done basic testing but I would like to play with some edge cases tomorrow as I've only been able to test my two monitors (both 16:9 with one rotated). I think its fine to review now though.


vaxerski:
> For example, I think it would be nice to make a visual tui curses tool to align your monitors with a set scale and resolution

https://github.com/artizirk/wdisplays

> I also think we should implement a way to correct or warn on poor monitor placement if a user accidentally overlaps monitors or makes it so that they aren't connected.

sounds fair

src/helpers/Monitor.hpp
```diff
@@ -11,7 +11,16 @@
 #include "Region.hpp"
 #include <optional>
 
+// Enum for the different types of auto directions, e.g. auto-left, auto-up.
+enum class AutoDirs {
+    auto_up,
```
vaxerski, line 16: enums in CAPS and with their short name in the beginning, e.g.:
`DIR_AUTO_UP`

src/helpers/Monitor.hpp
```diff
@@ -11,7 +11,16 @@
 #include "Region.hpp"
 #include <optional>
 
+// Enum for the different types of auto directions, e.g. auto-left, auto-up.
+enum class AutoDirs {
```
vaxerski, line 15: enums prefix `e`: `eAutoDirs`

src/config/ConfigManager.cpp
```diff
@@ -1641,6 +1641,16 @@ std::optional<std::string> CConfigManager::handleMonitor(const std::string& comm
 
     if (ARGS[2].starts_with("auto")) {
         newrule.offset = Vector2D(-INT32_MAX, -INT32_MAX);
+        if (ARGS[2] == "auto-right" || ARGS[2] == "auto") {
```
vaxerski, line 1644: style nit: no {} around short ifs

vaxerski (commented):
rest lgtm

The-Briel-Deal:
> > For example, I think it would be nice to make a visual tui curses tool to align your monitors with a set scale and resolution
> 
> https://github.com/artizirk/wdisplays

Woahhh, thanks for bringing that to my attention. Thanks for the quick feedback though. Fixing now.

src/helpers/Monitor.hpp
The-Briel-Deal: I see all of the other classes and structs are prefixed with a capital letter. e.g. SMonitorRule. Is that prefered over lower case? I'll stick with what you said here for now.

The-Briel-Deal:
@vaxerski All your feedback has been addressed. Just testing a bunch of different layouts to see if I can break this but feel free to review it now.

The-Briel-Deal:
Edge case found. If you declare your first monitor as auto-up and then your second monitor as auto left they will only connect at the bottom left and top right corner. The solution to this could be to always make users have their first declared monitor be `auto` or `auto-right` that way there is always a monitor in the center of others.

Alternatively we could just leave this behavior and note it in the docs + give users a warning if one of their monitors isn't connected.
Sad monitors

vaxerski:
We can make the first declared always at pos 0,0, sounds reasonable

The-Briel-Deal:
> We can make the first declared always at pos 0,0, sounds reasonable

Implemented. Not really sure why why the Nix stuff is failing. It looks like it is in upstream too.

vaxerski:
ignore the nix stuff

vaxerski (dismissed):
lgtm, just needs a wiki mr and a `clang-format`

fufexan:
You can rebase to fix the Nix CI. Well you need to rebase anyway due to conflicts.

The-Briel-Deal:
@vaxerski  clang-format has been wrong, and docs have been written! https://github.com/hyprwm/hyprland-wiki/pull/610

vaxerski (approved):
thanks :)

The-Briel-Deal:
> thanks :)

Np! You've been a great maintainer and help with me getting started contributing to open source! I appreciate that and all the work you do!

PR #5674, opened 2024-04-21
git: ignore wlroots dirty changes

thejch:
when patches are applied in wlroots they are shown as unstaged modified content.
we can ignore wlroots dirty changes caused by the applied patches.
after this it will only show up as modified if you set the subproject to a different commit.


vaxerski (approved):
lgtm, thanks :)

PR #5681, opened 2024-04-21
protocols: add missing header for g_pCompositor

gulafaran:

adds missing header to RelativePointer.cpp, build fails for me on gentoo otherwise on g_pCompositor.


vaxerski (approved):
oops, thanks

PR #5683, opened 2024-04-21
Make ci build use hyprland cachix

fred21O4:
I noticed that the nix ci build task is not using the hyprland cachix due to it not trusting the flake substituters
This commit makes it use https://hyprland.cachix.org 
This can also be done by passing --accept-flake-config instead of --substituters however I prefer to be explicit in what caches to trust.
It can be done the other way if you prefer

This makes ci builds use the hyprland cachix which will hopefully speed up builds slightly

This probably doesn't improve build speeds that much but it wont hurt

should be good to merge


fufexan:
Please squash, rebase on main and then it's good to go.

fred21O4:
I messed up the rebase I'll just submit another pr

PR #5738, opened 2024-04-24
Always recheck idle inhibitors on creation and deletion

zakk4223:
Formatting

Idle  inhibitors without associated windows were only being processed if some "window event" (focus, fullscreen map/unamp etc) happened. This caused them to not function in some cases.

Ready(?)


vaxerski (approved):
lgtm, thanks!

PR #5741, opened 2024-04-25
Default unconfigured monitors to open to the right

eriedaberrie:
This PR returns the functionality of making unconfigured monitors default to opening to the right (previously all the if statements fell through and they ended up opening up, the enum zero value). It also improves the logging for invalid auto directions. Finally, it makes the logging performed in arrangeMonitors actually accurate under the new directional auto system, instead of just stating the next available rightwards position even when the new monitor isn't being arranged to the right.

I also noticed that since the max monitor positions are being stored as ints, there was inconsistent use of decimals in the arrangeMonitor logging. Seeing as how monitor positions can only be integers values anyways, I opted to just remove all decimals in the logging.

I'm not too sure if the logging changes are really in the scope of the PR but the inconsistencies were really bugging me and I didn't want to make a whole new PR to edit logging.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1659,7 +1659,12 @@ std::optional<std::string> CConfigManager::handleMonitor(const std::string& comm
             newrule.autoDir = eAutoDirs::DIR_AUTO_UP;
         else if (ARGS[2] == "auto-down")
             newrule.autoDir = eAutoDirs::DIR_AUTO_DOWN;
-
+        else {
+            Debug::log(WARN,
+                       "Invalid auto direction. Valid options are 'auto',"
+                       "'auto-up', 'auto-down', 'auto-left', and 'auto-right'.");
+            error += "invalid auto direction ";
+        }
     } else {
         if (!ARGS[2].contains("x")) {
             error += "invalid offset ";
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
lgtm

vaxerski:
ah clang-format needed

eriedaberrie:
Whoops I accidentally pushed to the wrong branch just now.
(clang-format has been performed)

src/config/ConfigManager.cpp
```diff
@@ -1647,15 +1651,20 @@ std::optional<std::string> CConfigManager::handleMonitor(const std::string& comm
     if (ARGS[2].starts_with("auto")) {
         newrule.offset = Vector2D(-INT32_MAX, -INT32_MAX);
         // If this is the first monitor rule needs to be on the right.
-        if (ARGS[2] == "auto-right" || ARGS[2] == "auto" || m_dMonitorRules.empty())
+        if (ARGS[2] == "auto-right" || ARGS[2] == "auto" || m_dMonitorRules.empty()) {
```
vaxerski, line 1654: ???

eriedaberrie: wait what's the issue

vaxerski: no {} around short ifs, why did you add them

eriedaberrie: I added a multiline else clause so I thought I should add braces around the previous if statements as well, is that not a thing?

vaxerski: nope

vaxerski (approved):
lgtm

Issue #5399, opened 2024-04-03
IME popup not showing in fullscreen when scale is set to 1


sungyoonc:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit fbdaf74a82b38e11a0a6914b3c9f7cb934ac1624 dirty (master: fix swapped workspaces (5397)).
Date: Wed Apr 3 09:22:59 2024
Tag: v0.38.0-10-gfbdaf74a

flags: (if any)
```

</details>

### Bug or Regression?

Bug

### Description

When the monitor's scale is set to 1, when in fullscreen the ime popup disappear but comes back when not in fullscreen

I tested this with alacritty, foot, kitty, thunar, obsidian, discord(flatpak, xwayland) but it did not happen in firefox(via text-input-v3) and okular(which use QT_IM_MODULE=fcitx env to access ime directly).

When I set the scale to anything except 1(e.g. 1.2, 2 etc), it doesn't have this issue.

This is my monitor config
```sh
monitor = eDP-1,2880x1800@120,auto,auto,vrr,1 (scale 2)
monitor = DP-3,1920x1080@60,auto,auto (scale 1)
monitor=,preferred,auto,auto
```

When I change the scale of eDP-1 to 1 the issue happens on it.
When I change the scale of DP-3 to 2, the ime popup doesn't disappear.

### How to reproduce

1. Set the scale to 1.
2. Open an application (alacritty, foot, kitty, thunar, obsidian or discord(flatpak, xwayland) and presumably many others)
  If you want I can test other applications.
3. Type in a textbox -> IME Popup is visible
4. Go to fullscreen -> IME Popup disappears

### Crash reports, logs, images, videos

https://github.com/hyprwm/Hyprland/assets/65142096/2b3181b4-9aaf-45a4-b460-37c3c66dc675

https://github.com/hyprwm/Hyprland/assets/65142096/1cfc0827-ae54-4cb0-aaf5-e09f4aaa1952

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

Issue #4781, opened 2024-02-21
Firefox freezes when fullscreened on a second monitor


dropbear32:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch HEAD at commit 84ab8d11e8951a6551d1e1bf87796a8589da6d47  (props: bump ver to 0.35.0).
Date: Mon Feb 5 01:59:02 2024
Tag: v0.35.0

flags: (if any)


System Information:
System name: Linux
Node name: arch
Release: 6.7.5-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Sat, 17 Feb 2024 14:02:33 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Alder Lake-P Integrated Graphics Controller [8086:4626] (rev 0c) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```
</details>

### Bug or Regression?

Bug

### Description

When fullscreening Firefox (e.g. F11), the window will no longer update. Moving the cursor does cause the window to update, but only for as long as the cursor is being moved.

Chromium is not affected by this problem, nor is Firefox when on my laptop's built-in monitor. This only happens when Firefox is on an external monitor. (In my case, it's over HDMI.)

<details> <summary>hyprctl monitors</summary>

```
Monitor eDP-1 (ID 0):
	2256x1504@59.99900 at 0x0
	description: BOE 0x095F (eDP-1)
	make: BOE
	model: 0x095F
	serial: 
	active workspace: 1 (1)
	special workspace: 0 ()
	reserved: 0 0 0 0
	scale: 1.57
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: 0
	activelyTearing: false

Monitor DP-1 (ID 1):
	2560x1440@59.95100 at 1439x0
[...]

Issue #4474, opened 2024-01-18
Scaled Display not Covering Full Screen Width


mewoocat:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit dcd7a92b01a360bbe7afa7299604c4040abb4c62  ().
Date: 2024-01-15
Tag: 

flags: (if any)


System Information:
System name: Linux
Node name: scythe
Release: 6.1.72
Version: #1-NixOS SMP PREEMPT_DYNAMIC Wed Jan 10 16:10:37 UTC 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation HD Graphics 620 [8086:5916] (rev 02) (prog-if 00 [VGA controller])


os-release: BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.05.20240113.9b19f5e"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.05 (Uakari)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.05 (Uakari)"
VERSION_CODENAME=uakari
VERSION_ID="24.05"


plugins:

```

</details>

### Bug or Regression?

Bug

### Description

Setting the monitor scale to something besides 1x results in a small vertical line of rendered space on the right side of the screen.  I have discussed this issue in the Discord and it was suggested that this is a rounding error.  

### How to reproduce

This issue has been observed at 1.33 and 1.5 scale on a 2560x1440 display.

### Crash reports, logs, images, videos

For example, this prevents the ability to click on a button that is at the far right when the mouse is at the edge of the screen.

Photo: (The mouse can be observed to be rendered in the blank space)
![20240118_082351](https://github.com/hyprwm/Hyprland/assets/97066700/184b948d-f256-4655-b5e7-4811c2d19006)

Screenshot: (The extra space seems to be transparent so it's difficult to see)
![image](https://github.com/hyprwm/Hyprland/assets/97066700/12d75612-df35-4ac6-b0cd-f31b723031f3)

vaxerski:
can you attach `hyprctl monitors`?

mewoocat:
```
Monitor eDP-1 (ID 0):
	2560x1440@59.95100 at 0x0
[...]

PR #5748, opened 2024-04-25
Fix the rounding of the size of fractionally scaled monitors

eriedaberrie:
This fixes the issue of some monitors with fractional scaling having their vecSize 1 pixel less than they should be. In my case, my 2560x1600 monitor with 1.33x scaling would end up with dimensions 1919.999...x1119.999..., which would then get floored to 1919x1119. This caused a 1-pixel outline around the edges of the monitor where nothing happened (very visible if I set background_color to 0x00ff0000 or something and fullscreened anything or had a dark wallpaper. The problem would be exacerbated if external monitors are positioned with auto, because due to the incorrect vecSize the monitors would end up actually overlapping by a single pixel. This would cause all sorts of problems like vfr breaking on certain fullscreened apps (firefox in particular) and also popup menus not working on fullscreened apps for some reason.

Fixes:
#3511
#3778
#4474
#4781
and *possibly* #5399 (not sure if related)

I'm not sure if this is the right way to go about this and it might still have problems if there is no clean divisor for the scaling factor but it's better than nothing I think.

Seems ready to me.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -2260,7 +2260,7 @@ bool CHyprRenderer::applyMonitorRule(CMonitor* pMonitor, SMonitorRule* pMonitorR
 
     int x, y;
     wlr_output_transformed_resolution(pMonitor->output, &x, &y);
-    pMonitor->vecSize            = (Vector2D(x, y) / pMonitor->scale).floor();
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
oh man, right. Thanks

YourSandwich:
This causes some notifications like Steam Notifications to be slightly cut off

PR #5771, opened 2024-04-27
fix(CrashReporter): make it mostly async-signal-safe

virchau13:

I was randomly experiencing Hyprland crashes (due to other issues) that resulted in the crash-handler indefinitely hanging and never saving logs. After doing some debugging I figured out why logs weren't saving: because the SIGSEGV/SIGABRT signal handler wasn't async-signal-safe (see [`signal-safety(7)`](https://man7.org/linux/man-pages/man7/signal-safety.7.html).) 

In particular, `CrashReporter::createAndSaveCrash()` was not async-signal-safe, resulting in random deadlocks/double-crashes during Hyprland crashes. This changes the function to be (mostly) async-signal-safe.

This also adds an `alarm()` statement to the unrecoverable signal handler to kill the process if the signal-handler somehow gets deadlocked. This is necessary because Hyprland steals keyboard access, meaning that if a user hasn't enabled SysRq or TTYs, a Hyprland crash can soft-lock their computer. (To escape, one had to use SysRq+R, switch to another TTY with Ctrl+Alt+F2, and then SIGKILL the Hyprland process.)

A lot of otherwise-regular functions had to be reimplemented purely in system calls, because POSIX only gives guarantees for *certain* functions to be async-signal-safe (the full list is in [`signal-safety(7)`](https://man7.org/linux/man-pages/man7/signal-safety.7.html)). Specifically, in signal handlers, you _cannot_
* allocate memory
* call any other library function other than the ones listed in `signal-safety(7)`, even the ones that sound trivial
* trust the validity of any global variables, mutexes, or semaphores

The code is written to do absolutely everything on the stack, minimizing stack usage as much as possible. It is also written to save as much information as possible in the crash report file in case a global invariant is broken by accident.

It wasn't possible to make the backtrace code async-signal-safe (`dladdr1()` and `backtrace_symbols()` aren't async-signal-safe) so instead we do a best-effort attempt at obtaining the backtrace.  
(This is strictly better than the previous code: the previous code would always fail if any global invariants were broken, this code only fails if the `malloc()` or `std::allocator` invariants are broken.)

Ready for merge.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -35,6 +35,14 @@ void handleUnrecoverableSignal(int sig) {
         return;
     }
 
+    // Kill the program if the crash-reporter is caught in a deadlock.
+    signal(SIGALRM, [](int _) {
+        char const* msg = "\nCrashReporter exceeded timeout, forcefully exiting\n";
+        write(2, msg, strlen(msg));
+        abort();
+    });
+    alarm(15);
+
     CrashReporter::createAndSaveCrash(sig);
 
     abort();
--- a/src/debug/CrashReporter.cpp
+++ b/src/debug/CrashReporter.cpp
@@ -1,59 +1,127 @@
 #include "CrashReporter.hpp"
-#include <random>
+#include <fcntl.h>
 #include <sys/utsname.h>
-#include <fstream>
-#include <signal.h>
 #include <link.h>
+#include <time.h>
+#include <errno.h>
+#include <sys/stat.h>
 
 #include "../plugins/PluginSystem.hpp"
+#include "../signal-safe.hpp"
 
 #if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__)
 #include <sys/sysctl.h>
 #endif
 
-std::string getRandomMessage() {
+static char const* const MESSAGES[] = {"Sorry, didn't mean to...",
+                                       "This was an accident, I swear!",
+                                       "Calm down, it was a misinput! MISINPUT!",
+                                       "Oops",
+                                       "Vaxry is going to be upset.",
+                                       "Who tried dividing by zero?!",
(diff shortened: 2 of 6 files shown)


src/Compositor.cpp
```diff
@@ -35,6 +35,14 @@ void handleUnrecoverableSignal(int sig) {
         return;
     }
 
+    // Kill the program if the crash-reporter is caught in a deadlock.
+    signal(SIGALRM, [](int _) {
+        char const* msg = "\nCrashReporter exceeded timeout, forcefully exiting\n";
+        write(2, msg, strlen(msg));
+        abort();
+    });
+    alarm(15);
```
vaxerski, line 44: 15 is a bit too long IMO, something like 10 should do

src/debug/CrashReporter.cpp
```diff
@@ -1,81 +1,163 @@
...
+                                       "Maybe you should try dusting your PC in the meantime?",
+                                       "I tried so hard, and got so far...",
+                                       "I don't feel so good...",
+                                       "*thud*",
+                                       "Well this is awkward.",
+                                       "\"stable\"",
+                                       "I hope you didn't have any unsaved progress.",
+                                       "All these computers..."};
+
+char const*              getRandomMessage() {
```
vaxerski, line 31: put a `//` above to get rid of the wonky indent

vaxerski (commented):
rest lgtm

src/Compositor.cpp
virchau13: In my testing it takes about 7 seconds to get the backtrace, and I'm on a fairly decent computer. Worse computers would probably exceed the 10 second threshold.

vaxerski: hm, I wonder where that is spent, maybe in nm. Alright then

vaxerski (approved):
pretty cool, thanks :)

Issue #5745, opened 2024-04-25
[Feature Request] warp cursor on openwindow/closewindow


mklan:
### Description

When we spawn a new window the cursor is currently not warped, even tho the new window is focused from the beginning. Currently I am listening to events and then focusing once again using the dispatcher. It would be nice to have this integraded.

The-Briel-Deal:
Fixed in https://github.com/hyprwm/Hyprland/pull/5777. Try it out if you can!

youegraillot:
what's the state on this ?

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #5777, opened 2024-04-27
Warp cursor on openwindow/closewindow.

The-Briel-Deal:
Fixes/addresses https://github.com/hyprwm/Hyprland/issues/5745.

Currently when creating a new window, the cursor doesn't warp to the new window like it does on a window focus change. As such I used the same method (warpCursorTo) to warp the cursor. This will also be disabled with no_cursor_warps.

This is a very straightforward change. It's unlikely this will introduce unexpected behavior as new windows are already focused, this just moves mouse too.

Ready for merging.

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -668,7 +668,6 @@ void Events::listener_mapWindow(void* owner, void* data) {
     if (PMONITOR && PWINDOW->m_iX11Type == 2)
         PWINDOW->m_fX11SurfaceScaledBy = PMONITOR->scale;
 
-    // Move cursor to middle of new window if enabled.
     g_pCompositor->warpCursorTo(PWINDOW->middle());
 }
 
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -667,6 +667,9 @@ void Events::listener_mapWindow(void* owner, void* data) {
 
     if (PMONITOR && PWINDOW->m_iX11Type == 2)
         PWINDOW->m_fX11SurfaceScaledBy = PMONITOR->scale;
+
+    // Move cursor to middle of new window if enabled.
+    g_pCompositor->warpCursorTo(PWINDOW->middle());
 }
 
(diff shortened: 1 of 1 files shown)


src/events/Windows.cpp
```diff
@@ -667,6 +667,9 @@ void Events::listener_mapWindow(void* owner, void* data) {
 
     if (PMONITOR && PWINDOW->m_iX11Type == 2)
         PWINDOW->m_fX11SurfaceScaledBy = PMONITOR->scale;
+
+    // Move cursor to middle of new window if enabled.
```
vaxerski, line 671: useless comment

```
// this enables the thing
thing->enable();
```

src/events/Windows.cpp
```diff
@@ -667,6 +667,9 @@ void Events::listener_mapWindow(void* owner, void* data) {
...
+    g_pCompositor->warpCursorTo(PWINDOW->middle());
```
vaxerski, line 672: you're not checking if:
- there is a constraint
- the window gained focus
- the window is focusable

vaxerski (commented):
I like cursor warps but would absolutely hate this. IMO this should be configurable separately and off by default (probably in `misc:` or something)

src/events/Windows.cpp
The-Briel-Deal: Good points, especially the checking if its gained focus. I definitely put this out a little per-maturely.

For the first point though, in https://github.com/hyprwm/Hyprland/blob/2e763764bfdaee2039961c6c220b5d6af31712d4/src/Compositor.cpp#L2425-L2443. I see that this is only going to warp if the user hasn't put `general:no_cursor_warps` in their config. I kind of figured it fell under the same bucket as the change focus warps.

I'm going to think on this some more and ideally try to integrate this solution with what changing focus uses.

Thanks for the feedback! Going to put this in Draft for a bit.

vaxerski:
closing due to inactivity

PR #5778, opened 2024-04-27
Add some colors to stdout log 🔴🟡🟢🔵🟣

thejch:
Add some colors to stdout log 🔴🟡🟢🔵🟣
🔴 error
🟡 warning
🟢 info
🔵 trace
🟣 critical

not really

should be ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -819,7 +819,7 @@ void CConfigManager::postConfigReload(const Hyprlang::CParseResult& result) {
     if (Debug::disableStdout && isFirstLaunch)
         Debug::log(LOG, "Disabling stdout logs! Check the log for further logs.");
 
-    Debug::coloredLogs = std::any_cast<Hyprlang::INT>(m_pConfig->getConfigValue("debug:colored_stdout_logs"));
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
this should be disableable

thejch:
> this should be disableable

added

src/config/ConfigManager.cpp
```diff
@@ -818,6 +819,8 @@ void CConfigManager::postConfigReload(const Hyprlang::CParseResult& result) {
     if (Debug::disableStdout && isFirstLaunch)
         Debug::log(LOG, "Disabling stdout logs! Check the log for further logs.");
 
+    Debug::coloredLogs = std::any_cast<Hyprlang::INT>(m_pConfig->getConfigValue("debug:colored_stdout_logs"));
```
vaxerski, line 822: this will make it unchangeable dynamically. Why not do the same as with `debug:disable_logs`?

thejch: i think its the opposite. this is on config reload so it is changeable dynamically. if i put it next to disable_logs it only gets run once and that will be unchangeable

vaxerski: dynamically means with hyprctl.

`disable_logs` is a static pointer, not a value copy.

thejch: changed

vaxerski (approved):
lgtm, thanks. Wiki mr needed.

PR #5781, opened 2024-04-28
feat: add compile_commands.json output to `make debug` build

virchau13:

This makes all the include resolution and diagnostic information work properly (with the `clangd` language server on NixOS, at least.) Probably makes this line unnecessary: https://github.com/hyprwm/Hyprland/blob/2e763764bfdaee2039961c6c220b5d6af31712d4/src/includes.hpp#L6

* As far as I can tell, the `xwayland` dependency provides literally nothing.
```
/nix/store/i3yspjhpqz8n25phd91p5cyrh2jl1xvh-xwayland-23.2.6
├── bin
│  └── Xwayland
├── lib
│  ├── pkgconfig
│  │  └── xwayland.pc
│  └── xorg
│     └── protocol.txt
└── share
   ├── applications
   │  └── org.freedesktop.Xwayland.desktop
   └── man
      └── man1
         ├── Xserver.1.gz
         └── Xwayland.1.gz
```
I'm not sure why it's even in CMakeLists.txt in the first place.

* This shouldn't conflict with #5667.

* If you add `-DCMAKE_LINKER_TYPE=MOLD` and install the Mold linker, the debug builds become *significantly* faster, but I don't know if I should add that to this PR.

Ready for merge.


fufexan:
> Probably makes this line unnecessary:

If it's the case, can you also remove it in this PR?

virchau13:
I don't have VSCode, so someone else should test if this actually fixes it.

BvngeeCord:
Can confirm, this fixes all include errors for me in Neovim with Clangd. Note: by default, clangd shouldn't need `compile_commands.json` to be symlinked to the project root, as [looks upwards](https://clangd.llvm.org/design/compile-commands) naturally starting from where the source file is. Though it might still be good to be safe.

Alternatively, `set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)` can be set inside of `CMakeLists.txt` instead of the env var if that is preffered (which is what I've been doing).

khushal-solves:
> Alternatively, set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)

Yes or simply pass -DCMAKE_EXPORT_COMPILE_COMMANDS=1 to cmake while compiling.

It fixes all errors.

vaxerski:
closing due to inactivity

PR #5796, opened 2024-04-29
log on watchdog activation

BvngeeCord:

When watchdog is triggered, log something so that users that are debugging / working on Hyprland don't get confused when a simple breakpoint crashes their program. Context: [this discord message](https://discord.com/channels/961691461554950145/1070436481912549497/1234261662081744916).

Thinking about the new user experience, `debug:enable_stdout_logs` is still disabled by default, meaning this won't take effect unless users have already looked through all the debugging variables and happened to set this. Maybe it makes sense to set the default value of disable_logs/enable_stdout_logs based on whether it's a debug build or not?

Should be ready, though feel free to reword the message

### DIFF
--- a/src/helpers/Watchdog.cpp
+++ b/src/helpers/Watchdog.cpp
@@ -22,8 +22,10 @@ CWatchdog::CWatchdog() {
             if (!m_bWillWatch)
                 m_cvWatchdogCondition.wait(lk, [this] { return m_bNotified; });
             else {
-                if (m_cvWatchdogCondition.wait_for(lk, std::chrono::milliseconds((int)(*PTIMEOUT * 1000.0)), [this] { return m_bNotified; }) == false)
+                if (m_cvWatchdogCondition.wait_for(lk, std::chrono::milliseconds((int)(*PTIMEOUT * 1000.0)), [this] { return m_bNotified; }) == false) {
+                    Debug::log(LogLevel::CRIT, "Watchdog timeout met: killing main thread!");
                     pthread_kill(m_iMainThreadPID, SIGUSR1);
+                }
             }
 
             if (m_bExitThread)


vaxerski:
but it is logged: https://github.com/hyprwm/Hyprland/blob/33e0bb14786dc22a0c82eaaf097b469d3fdeceea/src/helpers/WLListener.cpp#L13-L17

BvngeeCord:
> but it is logged: https://github.com/hyprwm/Hyprland/blob/33e0bb14786dc22a0c82eaaf097b469d3fdeceea/src/helpers/WLListener.cpp#L13-L17

Hmm, that doesn't seem to trigger for me to while debugging Hyprland with breakpoints. I didn't even notice that was there.

vaxerski:
generally we can't log from watchdog because Debug::log is not thread-safe

BvngeeCord:
> we can't log from watchdog because Debug::log is not thread-safe

Ah, I completely missed that. Well I'm not too sure then how to go about generic watchdog logging from outside of the watchdog thread (such that it triggers when GDB pauses at a breakpoint). Feel free to close this, though it's def still worth thinking about

vaxerski:
I mean generally on my end the log gets printed... You can try adding a `std::chrono::this_thread::sleep_for(std::chrono::seconds(10))` in a random handler and seeing if the log is printed

BvngeeCord:
> I mean generally on my end the log gets printed... You can try adding a `std::chrono::this_thread::sleep_for(std::chrono::seconds(10))` in a random handler and seeing if the log is printed

Turns out I just had to set `disable_logs`  to false. This surprises me though, given that its description on the wiki says it "disables logging to a file" but it also apparently disables watchdog logging to stdout (even though there's a dedicated `enable_stdout_logs` which I had set to true already). I can't quite figure out why this is the case (and don't rly have the time rn) but I think it's pretty weird that the watchdog log only occurs when disabled_logs is false *and* enable_stdout_logs is true. Either that or the wiki description is just outdated

Issue #5803, opened 2024-04-29
Default monitor for cursor


ThatOneCalculator:
### Description

I wasn't able to find this in the documentation, so I'm not sure if this exists already, but the option to set the monitor that the cursor will appear in on login would be great. I have both DP-1 and HDMI-A-2, and although DP-1 is my primary monitor, my cursor always appears on HDMI-A-2 on login.

vaxerski:
yea it currently is on the first one that gets recognized by drm I believe

PR #5847, opened 2024-05-02
added option to choose the default monitor that the cursor will appea…

ikalco:
Default monitor for cursor #5803 

Adds an option (string) under general:default_cursor_monitor.
Then, whenever a new monitor is added, it checks whether it's name matches the above string.
If it does, then it will set the cursor to the middle of that monitor.

If the option isn't set then it functions the same as before

Also, technically if a new monitor is added after startup and it matches the name, then the cursor will be set to there.
To fix this you would need to make the code stop checking after startup or after all the initial monitors have been added.
I don't know how you would do that though

Probably, I tested it on my PC and changing the option does change which monitor the mouse appears in on startup

### DIFF
--- a/src/events/Monitors.cpp
+++ b/src/events/Monitors.cpp
@@ -60,12 +60,44 @@ void Events::listener_change(wl_listener* listener, void* data) {
     wlr_output_manager_v1_set_configuration(g_pCompositor->m_sWLROutputMgr, CONFIG);
 }
 
+static void checkDefaultCursorWarp(std::shared_ptr<CMonitor>* PNEWMONITORWRAP, std::string monitorName) {
+    const auto  PNEWMONITOR = PNEWMONITORWRAP->get();
+
+    static auto PCURSORMONITOR    = CConfigValue<std::string>("general:default_cursor_monitor");
+    static auto firstMonitorAdded = std::chrono::system_clock::now();
+    static bool cursorDefaultDone = false;
+    static bool firstLaunch       = true;
+
+    const auto  POS = PNEWMONITOR->middle();
+
+    // by default, cursor should be set to first monitor detected
+    // this is needed as a default if the monitor given in config above doesn't exist
+    if (firstLaunch) {
+        firstLaunch = false;
+        g_pCompositor->warpCursorTo(POS, true);
+        g_pInputManager->refocus();
+    }
+
+    if (cursorDefaultDone || *PCURSORMONITOR == STRVAL_EMPTY)
+        return;
+
+    // after 10s, don't set cursor to default monitor
+    auto timePassedSec = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - firstMonitorAdded);
+    if (timePassedSec.count() > 10) {
+        cursorDefaultDone = true;
+        return;
+    }
+
+    if (*PCURSORMONITOR == monitorName) {
+        cursorDefaultDone = true;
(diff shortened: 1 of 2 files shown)


src/events/Monitors.cpp
```diff
@@ -114,6 +114,14 @@ void Events::listener_newOutput(wl_listener* listener, void* data) {
             }
         }
     }
+
+    static auto PCURSORMONITOR = *CConfigValue<std::string>("general:default_cursor_monitor");
```
vaxerski, line 118: ay ay ay ay you can't do that get that `*` outta here no no no

vaxerski (commented):
To make this one-time:
 - add a `static bool once = true` inside the if and set it to false once you match one output
 - add a `static std::chrono::system_clock::time_point firstTimeOrSomethign = std::chrono::system_clock::now();` with the same idea and give an arbitrary max time threshold (10s should be reasonable) after which the thing disables itself

thejch:
Would it be better to make this part of the monitor rules? it would probably also fix this issue:
> Also, technically if a new monitor is added after startup and it matches the name, then the cursor will be set to there.
To fix this you would need to make the code stop checking after startup or after all the initial monitors have been added.
I don't know how you would do that though

vaxerski:
don't think so, because it can only match one monitor at a time, so there is no benefit to it being a rule.

src/events/Monitors.cpp
```diff
@@ -114,6 +114,24 @@ void Events::listener_newOutput(wl_listener* listener, void* data) {
...
+    static auto PCURSORMONITOR    = CConfigValue<std::string>("general:default_cursor_monitor");
+    static auto firstMonitorAdded = std::chrono::system_clock::now();
+    static bool cursorDefaultDone = false;
+
+    if (cursorDefaultDone || *PCURSORMONITOR == STRVAL_EMPTY)
+        return;
+
+    auto timePassedSec = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - firstMonitorAdded);
+    if (timePassedSec.count() > 10)
+        cursorDefaultDone = true;
```
vaxerski, line 127: should also return, otherwise setting this at a random point will warp the cursor once.

src/events/Monitors.cpp
```diff
@@ -114,6 +114,24 @@ void Events::listener_newOutput(wl_listener* listener, void* data) {
...
+
+    if (*PCURSORMONITOR == OUTPUT->name) {
+        cursorDefaultDone = true;
+        const auto POS    = PNEWMONITOR->middle();
+        if (g_pCompositor->m_sSeat.mouse != nullptr)
```
vaxerski, line 132: entire thing can be g_pCompositor->warpCursor no?

ikalco:
I also saw that whenever a new window is created it doesn't follow where the mouse was.
Ex. mouse, by default, on left monitor, window created on right

Is that related?
Maybe the focus needs to be set as well?

vaxerski (commented):
as it's getting quite large, I'd move it to a separate function e.g. `static void checkDefaultCursorWarp` or something

vaxerski:
> I also saw that whenever a new window is created it doesn't follow where the mouse was.

you should also g_pCompositor->setactivemonitor and g_pInputManager->refocus() for that

vaxerski (approved):
lgtm, wiki mr needed

ikalco:
Does this work?
hyprwm/hyprland-wiki#627

PR #5852, opened 2024-05-03
fix a few asan reported leaks on exit of hyprland

gulafaran:
a few of the leaks are mostly unimportant since the OS will reclaim the memory anways, but it reduces the asan spam if one builds with it atleast. 

the changes could use some testing and reviewing so i didnt miss anything obvious, perhaps some weird combination of HW and situation will cause some use after free?

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -421,8 +421,22 @@ void CCompositor::cleanup() {
     g_pWatchdog.reset();
     g_pXWaylandManager.reset();
 
-    wl_display_terminate(m_sWLDisplay);
+    if (m_sWLRCursor)
+        wlr_cursor_destroy(m_sWLRCursor);
 
+    if (m_sSeat.seat)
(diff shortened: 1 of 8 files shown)


vaxerski (approved):
dunno why this is a problem but sure I guess

gulafaran:
> dunno why this is a problem but sure I guess

i knew i missed something obvious, im on a laptop and went into the closet digging for some dusty old logitech usb keyboard. hottplugging its gonna double free because its not set to nullptr here

```
index 58b5caa3..fb9d2b8a 100644
@@ -1195,6 +1195,7 @@ void CInputManager::destroyKeyboard(SKeyboard* pKeyboard) {
     pKeyboard->hyprListener_keyboardKey.removeCallback();
 
     xkb_state_unref(pKeyboard->xkbTranslationState);
+    pKeyboard->xkbTranslationState = nullptr;
 
     m_lKeyboards.remove(*pKeyboard);
```

PR #5855, opened 2024-05-03
inputmgr: dont double free on hotplug

gulafaran:
since we are also unrefing the state on hotplugging the keyboard set the state to nullptr so the destructor if case actually catches its been already freed.

#5852 got merged so fast it didnt feel right, im on a laptop so had to dig through the closet for an old logitech usb keyboard and lo and behold there was a missing nullptr here, without it double free occurs.


vaxerski (approved):
lgtm

vaxerski:
> got merged so fast it didnt feel right

hyprland

PR #5866, opened 2024-05-04
togglegroup ipc event added when create/destroy-ing window groups

### LINKED ISSUE #5687: togglegroup IPC event
### Description

i need it for my script and overall would be nice to have it

VPavliashvili:
I found myself in a need of a IPC event of creating or destroying window groups. I read the [ipc wiki article](https://wiki.hyprland.org/IPC/) and saw this feature was missing

This PR will add new `togglegroup` IPC event which will return 0/1 after creating or destroying [a window group](https://wiki.hyprland.org/0.18.0beta/Configuring/Dwindle-Layout/#grouped-tabbed-windows) 
Nothing specifically mentioned in question but this PR resolves #5687 issue
Should be ready for merge without conflicts

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -836,6 +836,8 @@ void CWindow::createGroup() {
         g_pCompositor->updateWorkspaceSpecialRenderData(workspaceID());
         g_pLayoutManager->getCurrentLayout()->recalculateMonitor(m_iMonitorID);
         g_pCompositor->updateAllWindowsAnimatedDecorationValues();
+
+        g_pEventManager->postEvent(SHyprIPCEvent{"togglegroup", std::format("1,{:x}", (uintptr_t)this)});
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/desktop/Window.cpp
```diff
@@ -836,6 +836,8 @@ void CWindow::createGroup() {
         g_pCompositor->updateWorkspaceSpecialRenderData(workspaceID());
         g_pLayoutManager->getCurrentLayout()->recalculateMonitor(m_iMonitorID);
         g_pCompositor->updateAllWindowsAnimatedDecorationValues();
+
+        g_pEventManager->postEvent(SHyprIPCEvent{"togglegroup", "1"});
```
vaxerski, line 840: I would also add a window handle as another argument (`handle,state`) so that one can know which window was toggled

VPavliashvili: In the case of creating group when 1 should be returned I understand your point. But what about when user is destroying the window group containing more than one windows. 
Should `togglegroup` event return 0 and empty string (something like this `"",0`) or 0 with array of window addresses inside the handle argument?
e.g `"0x64cea2525760,0x64cea2522380",0`

vaxerski: no need for quotes. Make `state` first then and the rest is a list of windows.

VPavliashvili: before we proceed this pr I realized that I also need ipc event for `moveintogroup` and `moveoutofgroup`  commands. I think they can be done in the following way
`moveintogroup` will give single window address which was merged in a group, e.g `0x64cea2525760`
`moveoutofgroup` will also give single window address which was separated from a group

What do you think about this?  and should I create a new PR for this or merge inside this one?

vaxerski: sounds ok

VPavliashvili: Implementation and squashing commits done, I think its ready to merge if you approve

VPavliashvili:
[PR request](https://github.com/hyprwm/hyprland-wiki/pull/637) sent to hyrpland-wiki repo

PR #5877, opened 2024-05-04
Regenerate completions, update man pages.

The-Briel-Deal:
I noticed, some options (specifically -i, --instance) were missing from my fish auto-completions. I found that the reason is because my completions were generated from the hyprctl manpages which was missing --instance. In this process I also tried that the completions generated by complgen were out of date so I regenerated these as well.

I've only tested fish completions. I can also drop the regenerated completions if the PR is too large.

It should be ready for merging! Going to try testing the zsh and bash completions before though.


vaxerski:
closing due to inactivity

PR #5883, opened 2024-05-05
New shared_ptr implementation

vaxerski:
The reasoning is quite simple - this is a more useful shared_ptr implementation. Whether the impl itself is better - God knows.

I am not sure if this is a great idea either, but:

weak_ptrs got two new superpowers that the stl ones dont have:

- you can dereference a weak_ptr directly. This will obviously segfault on a nullptr deref if it's expired.
   - this is useful to avoid the .lock() hell where we are 100% sure the pointer _should_ be valid. (and if it isn't, it should throw.)
- weak_ptrs are still valid while the SP is being destroyed.
   - reasoning: while an object (e.g. CWindow) is being destroyed, its `weak_ptr self` should be accessible (the sp is still alive, and so is CWindow), but it's not because by stl it's already expired (to prevent resurrection)
   - this impl solves it differently. w_p is expired, but can still be dereferenced and used. Creating `s_p`s is not possible anymore, though.
   - this is useful in destructors and callbacks.
   
review pls @outfoxxed @fufexan and everyone else.

TODO:
 - [x] Battle testing.
 - [x] Fix invalid delete

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,8 @@ if(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES DEBUG)
         message(STATUS "Enabling ASan")
 
         target_link_libraries(Hyprland asan)
+        pkg_check_modules(ffidep REQUIRED IMPORTED_TARGET libffi)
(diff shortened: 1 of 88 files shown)


vaxerski:
should be working. Also ran a quick test for possible leaks, doesn't seem to leak.

fufexan (approved):

vaxerski:
thanks, cheers

Issue #5899, opened 2024-05-05
lockscreen is not rendered above floating wine games


gulafaran:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit cddeec47a1fc0e70d8598fd10c29cd0e9489999f dirty (keybinds: make the keybind manager check for session lock (5894)).
Date: Sun May 5 20:28:14 2024
Tag: v0.40.0-11-gcddeec47, commits: 4617

flags: (if any)


System Information:
System name: Linux
Node name: tom-legion
Release: 6.8.9-gentoo-tom
Version: #1 SMP PREEMPT_DYNAMIC Fri May  3 08:23:20 CEST 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Raptor Lake-S UHD Graphics [8086:a788] (rev 04) (prog-if 00 [VGA controller])
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD104M [GeForce RTX 4080 Max-Q / Mobile] [10de:27e0] (rev a1) (prog-if 00 [VGA controller])


os-release: NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
VERSION_ID="2.15"
```

</details>

### Bug or Regression?

Bug

### Description

if i leave the computer with the floating wine game its rendered above the lockscreen, it still gets locked since i have to blindly type the password with the game rendering on to get it "unlocked".

i got a window rule if it matters to set all games as floating. 
windowrule = float, *.exe

picture of me just triggering the keybind to lock the screen https://i.imgur.com/cRcRIA9.jpeg

as can be seen the Adwaita cursor is now showing as its "locked" but the game is still above the lockscreen itself.

### How to reproduce

run a game in this case diablo2 mod median xl, in wine fullscreened. either wait for the idle timeout or lock with keybind.

PR #5906, opened 2024-05-06
always render lock screen above other stuff

ikalco:
trying to fix
lockscreen is not rendered above floating wine games #5899 

This isn't only wine btw, this affects all fullscreen xwayland clients
the real problem is that we're still rendering the fullscreen xwayland clients
and that doesn't follow the [session lock protocol](https://wayland.app/protocols/ext-session-lock-v1)
> This means that the compositor must stop rendering and providing input to normal clients.

Currently, I believe that Hyprland allows fullscreen xwayland clients use a dmabuf and direct scanout to render more efficiently
The question is... how do we make those clients stop rendering (through the dmabuf) when the session is locked?

rn I just always render the session locker, which works, but its a hack

if you are ok with the current solution

I'm going to **try** to actually fix it

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -1322,8 +1322,11 @@ void CHyprRenderer::renderMonitor(CMonitor* pMonitor) {
                         damageBlinkCleanup = 0;
                 }
             }
-        } else
+        } else {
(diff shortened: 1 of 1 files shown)


vaxerski:
wrong approach, I'll do it myself

Issue #5923, opened 2024-05-07
language onevent with error


coffebar:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit fe7b748eb668136dd0558b7c8279bfcd7ab4d759  (props: bump version to 0.39.1).
Date: Tue Apr 16 16:01:03 2024
Tag: v0.39.1, commits: 4460

flags: (if any)


System Information:
System name: Linux
Node name: crab
Release: 6.8.7-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Wed, 17 Apr 2024 15:20:28 +0000


GPU information:
00:02.0 VGA compatible controller [0300]: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] [8086:9a49] (rev 01) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

Hyprland is sending errors to the IPC sometimes:
> hyprland language onevent with error

in result, Waybar shows empty layout

Here is log from the Waybar:

> [2024-05-06 18:51:55.642] [debug] hyprland language onevent with English (US)
> [2024-05-06 18:51:55.648] [debug] hyprland language onevent with English (US)
> [2024-05-06 18:51:55.649] [debug] hyprland IPC received activelayout>>wlr_virtual_keyboard_v1,English (US)
> [2024-05-06 18:51:55.655] [debug] hyprland language onevent with English (US)
> [2024-05-06 18:51:55.661] [debug] hyprland language onevent with English (US)
> [2024-05-06 18:51:55.662] [debug] hyprland IPC received activelayout>>wlr_virtual_keyboard_v1,error
> [2024-05-06 18:51:55.667] [debug] hyprland language didn't find matching layout
> [2024-05-06 18:51:55.667] [debug] hyprland language onevent with error
[...]

PR #5925, opened 2024-05-07
check value correctly with xkb_state_layout_index_is_active()

ikalco:
fixes
language onevent with error #5923

in cpp `-1` is truthy and Hyprland uses this
```cpp
std::string IKeyboard::getActiveLayout() {
    const auto WLRKB      = wlr();
    const auto KEYMAP     = WLRKB->keymap;
    const auto STATE      = WLRKB->xkb_state;
    const auto LAYOUTSNUM = xkb_keymap_num_layouts(KEYMAP);

    for (uint32_t i = 0; i < LAYOUTSNUM; ++i) {
        if (xkb_state_layout_index_is_active(STATE, i, XKB_STATE_LAYOUT_EFFECTIVE)) {
            const auto LAYOUTNAME = xkb_keymap_layout_get_name(KEYMAP, i);

            if (LAYOUTNAME)
                return std::string(LAYOUTNAME);
            return "error";
        }
    }

    return "none";
}
```

Not sure how it would get an invalid index though

Yes

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1084,7 +1084,7 @@ std::string switchXKBLayoutRequest(eHyprCtlOutputFormat format, std::string requ
     const auto         LAYOUTS      = xkb_keymap_num_layouts(PWLRKEYBOARD->keymap);
     xkb_layout_index_t activeLayout = 0;
     while (activeLayout < LAYOUTS) {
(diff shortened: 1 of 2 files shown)


vaxerski:
ah yeah I should rtfm indeed

vaxerski (approved):
thanks!

PR #5936, opened 2024-05-07
empty on monitor and next empty flags

wagyourtail:

add a selector to select the first empty workspace on the monitor. and a flag to only select empty workspaces with a higher id.

flags work by suffix'ing empty
ie.
`emptynm` for next empty on the same monitor
`emptym` for first empty on same monitor

so basically, I want this so when I hit my fullscreen bind the current window moves to the next empty workspace on the same monitor, without my program just spamming `r+1` until it finds one.

yes.


Agent00Ming (approved):
Works as described. Very convenient addition especially for larger multi-monitor setups. :+1:

PR #5943, opened 2024-05-08
meson: require hyprwayland-scanner >= 0.3.5

rtgiskard:
hyprwayland-scanner version has been bumped to 0.3.5, cmake build system is present, however the meson build system is not synced

I'm using arch, with AUR, it's default using the meson build system, which I prefer, however the previous installed `hyprwayland-scanner` from AUR is v0.3.4, and the latest version requirement is not synced in `meson.build`, it fails on build process without explicit prompt.

I think it's better to keep only one single build system, and meson is definitely a more modern and tidy build system for developers, maintain duplicate settings for two build system is unnecessary.

### DIFF
--- a/protocols/meson.build
+++ b/protocols/meson.build
@@ -17,7 +17,7 @@ wayland_scanner = find_program(
 	wayland_scanner_dep.get_variable('wayland_scanner'),
 	native: true,
 )
-hyprwayland_scanner_dep = dependency('hyprwayland-scanner', native: true)
+hyprwayland_scanner_dep = dependency('hyprwayland-scanner', version: '>=0.3.5', native: true)
 hyprwayland_scanner = find_program(
 	hyprwayland_scanner_dep.get_variable('hyprwayland_scanner'),
 	native: true,


fufexan (approved):

rtgiskard:
For myself, when it comes to choose the build system, the first and straightforward comparison is between `CMakeLists.txt` and `meson.build`, if I have to read one of the two files, I'd definite prefer `meson.build`, which is generally much less and clean, and I believe less is better.

From my experience, on writing `meson.build`, things are straightforward, you just need to declare things in a native manner. It's like coding to some degree, but as the author said, meson is not a Turing complete, you can't really code with that, which helps to keep things simple and easy to understand.

It's really doing things good, uniform doc, and elegant solutions for various customization.

Just a share of personal experience, no means to underestimate `cmake`, it has been there for long time, there is mature and complete ecosystem, being widely used. `Meson`, on the other hand, is young and distinctive.

vaxerski:
I've had both systems for over 2 years now under not only hyprland but other tools like hyprpaper or xdph and I have chosen cmake a while back for multiple reasons.

fwiw I hate writing meson

rtgiskard:
😂 Anyway, the build system is personal perspective, feel free to choose what you like.
Just to select a single one for a specific part to avoid unnecessary duplication of settings, which would generally lead to mess.

vaxerski:
I agree, that's why we are moving off of meson

PR #5952, opened 2024-05-08
nix: build improvements

andresilva:

Fixes populating the version header and removes unused stuff.

No.

Yes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -94,9 +94,4 @@
     nixosModules.default = import ./nix/module.nix inputs;
     homeManagerModules.default = import ./nix/hm-module.nix self;
   };
-
-  nixConfig = {
-    extra-substituters = ["https://hyprland.cachix.org"];
-    extra-trusted-public-keys = ["hyprland.cachix.org-1:a7pgxzMz7+chwVL3/pzj6jIBMioiJM7ypFP8PwtkuGc="];
-  };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -74,9 +74,6 @@
         # hyprland-extras
         
         xdg-desktop-portal-hyprland
-        # dependencies
-        
-        hyprland-protocols
         ;
     });
 
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -5,7 +5,6 @@
   pkgconf,
   makeWrapper,
   cmake,
-  meson,
   ninja,
   binutils,
   cairo,
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -12,7 +12,6 @@
   fribidi,
   git,
   hyprcursor,
-  hyprland-protocols,
   hyprlang,
   hyprwayland-scanner,
   jq,
@@ -107,7 +106,6 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
         fribidi
         git
         hyprcursor.dev
-        hyprland-protocols
         hyprlang
         libGL
         libdrm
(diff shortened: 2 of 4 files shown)


nix/default.nix
```diff
@@ -72,22 +70,12 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
...
-        --replace-warn "@HASH@" '${commit}' \
-        --replace-warn "@BRANCH@" "" \
-        --replace-warn "@MESSAGE@" "" \
-        --replace-warn "@DATE@" "${date}" \
-        --replace-warn "@TAG@" "" \
-        --replace-warn "@DIRTY@" '${
-        if commit == ""
-        then "dirty"
-        else ""
-      }'
```
andresilva, line 88: This is now done by CMake.

scripts/generateVersion.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/sh
 cp -fr ./src/version.h.in ./src/version.h
 
-HASH=$(git rev-parse HEAD)
-BRANCH=$(git branch --show-current)
-MESSAGE=$(git show ${GIT_COMMIT_HASH} | head -n 5 | tail -n 1 | sed -e 's/#//g' -e 's/\"//g')
```
andresilva, line 6: GIT_COMMIT_HASH was always unpopulated, it worked because it assumes HEAD when given no arguments.

vaxerski: I'd **highly** recommend to get nix to set  _at least_ GIT_COMMIT_HASH, otherwise hyprpm will not work.

andresilva: the nix build is setting the git hash, but the variable name is actually HASH whereas the script was using GIT_COMMIT_HASH here (which is undefined in the nix or regular build). I could have fixed this by replacing `${GIT_COMMIT_HASH}` with `$HASH`, but there's no point in doing that since we just want to use HEAD anyway (which is the default for git show if no argument is passed).

flake.nix
```diff
@@ -74,9 +74,6 @@
         # hyprland-extras
         
         xdg-desktop-portal-hyprland
-        # dependencies
-        
-        hyprland-protocols
```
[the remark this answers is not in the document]
fufexan: True. Can be removed.

fufexan:
@vaxerski are you alright with the `hyprland.pc.in` changes? CMake now installs the wlroots headers under `@PREFIX@/include/hyprland/wlr`. Only meson still installs them to `wlroots-hyprland`, but I'll change it.

Tested building the plugins, all green.

vaxerski:
dont care, as long as shit works

Agent00Ming:
On arch:
```
./documents/git/hyprland/build/hyprpm/hyprpm update

! Cloning https://github.com/hyprwm/hyprland, this might take a moment.                                                                                                                                   
✔ cloned                                                                                                                                                                                                  
✔ checked out to running ver                                                                                                                                                                              
! configuring Hyprland                                                                                                                                                                                    
✔ configured Hyprland                                                                                                                                                                                     
✖ failed to install headers with error code 4                                                                                                                                                             
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  5 / 5  Failed                                                                                                                                        

✖ Headers version mismatch. Please run hyprpm update to fix those.
```
not sure if this is good or bad

fufexan:
I hate that error code. I need more info. Thanks for testing though.

fufexan:
I've decided the pkg-config stuff + meson are out of scope for this PR, and I'll open a new one for those. Merging as soon as CI passes.

fufexan (approved):

PR #5966, opened 2024-05-09
Allow Multiple Simultaneous keys in a Keybind and add Left and Right mod distinction.

The-Briel-Deal:
I've seen many requests like: 
https://github.com/hyprwm/Hyprland/issues/5517, 
https://github.com/hyprwm/Hyprland/discussions/1703,
https://www.reddit.com/r/hyprland/comments/12weag0/binding_capshjkl_to_arrow_keys_setting_all_window/,
https://www.reddit.com/r/hyprland/comments/12g1axe/unable_to_bind_keys_properly/,
https://www.reddit.com/r/hyprland/comments/17ybq1n/keybind_help/, and many more.

This has also been a highly desired feature myself, and I don't find the solution of submaps to be a particularly intuitive solution. In all of these threads and issues people are asking for the ability to arbitrarily bind multiple normal keys or specify the specific mod key they want to use (e.g. Control_L, Alt_L, Meta_R...). This change allows both of these things without breaking existing behaviors or adding large amounts of overhead to existing configs.

To use this use the sym flag "s", with this flag you must use the symbols shown in xev like "Control_L" for mods or "k" for the k key. You also must seperate mods and keys with "&" like this:
```
# You can use a single mod with multiple keys.
binds = Control_L, A&Z, exec, kitty
# You can also specify multiple specific mods.
binds = Control_L&Shift_L, K, exec, kitty
# You can also do both!
binds = Control_R&Super_R&Alt_L, J&K&L, exec, kitty
```
(Note that I had to use & as the separator because some keysyms have underscores like the mod keys above)

None of the above from my testing! Please do try it out and see if you can break it though.

It should be ready for merging!


vaxerski:
feel free to request my review when you feel this is ready

The-Briel-Deal:
> feel free to request my review when you feel this is ready

Hey @vaxerski, I'm happy with where this is. I've gone through quite a few revisions of this and I'm happy with where this is at and can't find any edge cases. (There may be a weird flag combination since there are so many possible options with 9 flags)

Please take a look and let me know your thoughts. Here are two example keybinds below if you'de like to test. The keys are case insensitive and mods and keys are separated by &. I also updated the PR description with more info since you last saw.


```
binds = Control_R, a&z&s, exec, kitty
binds = Control_L&Control_R&Super_L, Y&u, exec, kitty
```

src/managers/KeybindManager.cpp
```diff
@@ -536,9 +538,50 @@ int repeatKeyHandler(void* data) {
...
+        return MK_NO_MATCH;
+
+    return mkKeysymSetMatches(keybind.sMkKeys, m_sMkKeys);
+}
+
 bool CKeybindManager::handleKeybinds(const uint32_t modmask, const SPressedKeyWithMods& key, bool pressed) {
     bool found = false;
 
+    if (pressed) {
+        if (keycodeToModifier(key.keycode)) {
```
vaxerski, line 573: no {} around short ifs

vaxerski (commented):
How will this work with shadowing, though?

Multi-key binds to me make no sense - you can define submaps already and achieve the same thing.

e.g.:
```
bindr = SUPER, A, exec, kitty
binds = SUPER, R&A, exec, dolphin
```
user presses super, a, and r. Will shadowing work and block super+a from executing?

The-Briel-Deal:
> How will this work with shadowing, though?
> 
> Multi-key binds to me make no sense - you can define submaps already and achieve the same thing.
> 
> e.g.:
> 
> ```
> bindr = SUPER, A, exec, kitty
> binds = SUPER, R&A, exec, dolphin
> ```
> 
> user presses super, a, and r. Will shadowing work and block super+a from executing?
> 
> For mods, why not just store an internal modmask that distinguishes L/R mods?



> How will this work with shadowing, though?
> 
> Multi-key binds to me make no sense - you can define submaps already and achieve the same thing.
> 
> e.g.:
> 
> ```
> bindr = SUPER, A, exec, kitty
> binds = SUPER, R&A, exec, dolphin
> ```
> 
> user presses super, a, and r. Will shadowing work and block super+a from executing?
> 
> For mods, why not just store an internal modmask that distinguishes L/R mods?

Good catch on shadowing, I'll look into this deeper this evening.

My primary justification for this feature is to have an un-opinionated option for users to bind anything they want. I don't really want to say "hey, these are the mods you are allowed to use". I really just want it to look at keysyms and treat them all the same no matter which you use, the only difference between mods and keys here is that if mods aren't an exact match then keypresses that are in the bind won't be be consumed (so if you have a `Super_L, A&B` bind and you press Super then A, then the key will be consumed, but A without the mod will not be consumed).

Let me give you a real world example I ran into. I personally never use my caps_lock, page_down, and page_up keys. So I wanted to be able to use these keys as mods for different hotkeys. But previously if your mod wasn't one of the 8 allowed mods, you just couldn't use it. I understand that a submap may work here, but they can get messy really quickly. And you can only use one layer, so `caps_lock,e&r,exec,ranger` was just not allowed.

So to sum it all up, I personally feel that we should leave these sort of options open to the user while staying out of the way for people who don't care about it.

vaxerski:
fair enough. As long as shadowing works, this looks ok to me

The-Briel-Deal:
> fair enough. As long as shadowing works, this looks ok to me

It should be good now, shadowing is good and I resolved conflicts and merged in latest main. Wait to merge till tomorrow though, going to make a PR to the docs and run it as my main install for a bit and make sure there aren't any issues.

src/managers/KeybindManager.cpp
The-Briel-Deal: Fixed

The-Briel-Deal:
Docs made https://github.com/hyprwm/hyprland-wiki/pull/652! Everything should be good and tested!

vaxerski (approved):
lgtm

noahfraiture:
Hello, it seems that super_l is no different that super_l in configuration. I set mainMod to super_l but both super key do the same. I checked with wev and they are two distinct key

The-Briel-Deal:
> Hello, it seems that super_l is no different that super_l in configuration. I set mainMod to super_l but both super key do the same. I checked with wev and they are two distinct key

Are you using the s flag? Can you show me the line in your config.

noahfraiture:
The command is as simple as 
```
$mainMod = SUPER_L # windows key
```
and then later things like this 
```
bind = $mainMod, T, exec, kitty  # open terminal
```
But I saw that this wasn't merge in the 0.40.0 which is on pacman. Does it mean that I will have to simply replace all my `bind` by `binds` ?

The-Briel-Deal:
If you want to use Keysyms for mods instead of words like SUPER you'll have to use the s (sym) flag. If you want to change that you can set it as default true instead of false in the source code, it's a 1 line change in ConfigManager.cpp.


Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

graham33:
Thanks very much for this @The-Briel-Deal !

> Does it mean that I will have to simply replace all my bind by binds ?

Yes, that worked for me. Just posting to make it super clear for others. For example I use:

```
$mainMod = Alt_R
binds = $mainMod, Q, exec, kitty
...
```

The-Briel-Deal:
> Thanks very much for this @The-Briel-Deal !
> 
> > Does it mean that I will have to simply replace all my bind by binds ?
> 
> Yes, that worked for me. Just posting to make it super clear for others. For example I use:
> 
> ```
> $mainMod = Alt_R
> binds = $mainMod, Q, exec, kitty
> ...
> ```

Ty @graham33, I'm glad it wasn't just me that wanted this! I'm happy it works alright for you!

istir:
Thank you @The-Briel-Deal, this feature is very useful to me!

However, I noticed a bug

NathanC:
Since the left/right mod distinction is under `s`, there doesn't seem to be a way to leverage it for `bindm` mouse binds still.

PR #5976, opened 2024-05-10
add focusonactivate window rule

MightyPlaza:
adds `focusonactivate` window rule to allow finer control than the global variable
maybe the variable could be removed

https://github.com/hyprwm/hyprland-wiki/pull/643


vaxerski (approved):
top g

PR #5902, opened 2024-05-05
Cursor/Tablet: move to a hyprland impl

vaxerski:
This moves wlr_cursor to a completely new impl mostly under CPointerManager

Still needs tablet support to be shipped.

TODO:
 - [x] Tablet support
 - [ ] Testing
 - [x] Wiki (https://github.com/hyprwm/hyprland-wiki/pull/641)
 - [x] Fix hardware cursors being a bit inefficient with rendering
 - [x] Fix software cursors being a broken mess
 - [x] ~~Allow `WLR_NO_HARDWARE_CURSORS`~~ (`cursor:no_hardware_cursors`. Dynamic as well.)
 - [x] Fix surface cursors being wonky
 
 New stuff:
 - `cursor:no_hardware_cursors` replaces the wlr env
 - `cursor:hotspot_padding` default `1`, the amount of padding around the hotspot around monitor edges. Set to 10 to see what I mean.

PR #6027, opened 2024-05-11
fix: change min cursor padding to 0

drendog:
Since ed411f53bdcf8d455f9e6ca769693aa84dc8372f a padding has been added to the cursor, which means that some application that expect a hit in the window border bounds, will never detects the hit there. 
For example when using firefox at full screen mode, that expect a hit on top of the windows to show the hidden tabs, it will never detects the hit.

ed411f53bdcf8d455f9e6ca769693aa84dc8372f #5902

Ready


Agent00Ming:
but my scroll bar :sob: https://github.com/hyprwm/Hyprland/pull/5902#issuecomment-2098567103

Agent00Ming (approved):
Firefox fullscreen functionality is reduced by the 1 pixel minimum. This does indeed fix it.

vaxerski:
wouldnt it be better to add a setting to have the padding only on the bottom right?

morr0ne:
> wouldnt it be better to add a setting to have the padding only on the bottom right?

Maybe I am not understanding this right, but what is the purpose of the padding? It mainly seems to break stuff.

Agent00Ming:
> > wouldnt it be better to add a setting to have the padding only on the bottom right?
> 
> Maybe I am not understanding this right, but what is the purpose of the padding? It mainly seems to break stuff.

https://github.com/hyprwm/Hyprland/pull/5902#issuecomment-2098567103 and https://github.com/hyprwm/Hyprland/pull/5902#issuecomment-2098575700 should explain the reason behind it.

Aqa-Ib:
Well, in my case the firefox bug I was having is fixed with the 1 pixel. That is because I am using "fakefullscreen" instead of normal "fullscreen", and with fakefullscreen you need that pixel to be able to show the tab bar. 

TLDR: the 1 pixel fixes "fakefullscreen" but breaks "fullscreen" firefox tab bar.

morr0ne:
Okay after reading through the pr I understand the purpose of the padding. However it feels like it should default to 0 and then users affected can add the 1px padding themselves. If this gets released as is most desktop users will end up with a broken config by default without even knowing why.

+1 for this pr

vaxerski:
I still think this is the wrong solution and instead https://github.com/hyprwm/Hyprland/pull/6027?notification_referrer_id=NT_kwDOApT3W7QxMDYyMTk2NjkwMTo0MzMxNzA4Mw&notifications_before=Y3Vyc29yOjI2#issuecomment-2106052387 should be done

morr0ne:
> I still think this is the wrong solution and instead [#6027 (comment)](https://github.com/hyprwm/Hyprland/pull/6027#issuecomment-2106052387) should be done

That could be a solution. Regardless this should be considered a blocker for 0.41.0 until resolved imo

src/managers/PointerManager.cpp
```diff
@@ -592,7 +592,7 @@ CBox CPointerManager::getCursorBoxGlobal() {
 Vector2D CPointerManager::closestValid(const Vector2D& pos) {
     static auto PADDING = CConfigValue<Hyprlang::INT>("cursor:hotspot_padding");
 
-    auto        CURSOR_PADDING = std::clamp((int)*PADDING, 1, 100); // 1px
+    auto        CURSOR_PADDING = std::clamp((int)*PADDING, 0, 100);
```
charchil-zel, line 595: the range from 0 to 100 is verry good

src/config/ConfigManager.cpp
```diff
@@ -525,7 +525,7 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("cursor:no_hardware_cursors", Hyprlang::INT{0});
     m_pConfig->addConfigValue("cursor:no_break_fs_vrr", Hyprlang::INT{0});
     m_pConfig->addConfigValue("cursor:min_refresh_rate", Hyprlang::INT{24});
-    m_pConfig->addConfigValue("cursor:hotspot_padding", Hyprlang::INT{1});
+    m_pConfig->addConfigValue("cursor:hotspot_padding", Hyprlang::INT{0});
```
charchil-zel, line 528: good default because it will prevent firefox to not show toolbars when in fullscreen, which statistically is more probable and if users don't research on this matter they won't even know what is wrong. the minimum hotspot should be 0 so that firefox will work properly or else hyprland is unusable.

drendog:
any news about it? Moving the problem is not a solution, since this compositor is supposed to be desktop-first, putting padding on the bottom will break the applications that expect a hit on bottom edge

charchil-zel (changes requested):
The default minimum of hotspot padding in the recent update has made hyprland so unusable with firefox that i have temporarily swithched to i3. Without firefox what is a computer? i so frustrating as i usually have 100 tabs open so i cannot cycle through them and i think so is the case with many people.

drendog:
@vaxerski

drendog:
putting the padding at the bottom will break those applications that expect a bottom edge cursor hit, in this way it's only a moving of the problem and not a real solution of it

vaxerski:
was the cursor able to go to (x,y) on a (w,h) sized monitor before? I don't think so

drendog:
ok i got it, i didn't check that the cursor can go out of screen limit, i'll fix it soon

drendog:
if i didn't misunderstand, the problem is the cursor hotspot can reach `(w, h)` instead of the maximum `(w - 1, h - 1)`. I figured out that `closestPoint` of Box, [here](https://github.com/hyprwm/hyprutils/blob/ff343e027951837f27f6afcd718dd9725a56d2b8/src/math/Box.cpp#L207) the maximum is `x + w` or `y + h`. I currently adjusted the coords in `NEAREST_LAYOUT`, if that was an error on hyprutils, i can open a pr for it and remove the adjustment here

vaxerski:
this won't work if you have two monitors side-by-side touching each other on the right or bottom.

drendog:
> this won't work if you have two monitors side-by-side touching each other on the right or bottom.

what do you mean? I tried every combo using 2 monitors and i don't see any issue

vaxerski:
mon1: 1920x1080 at 0x0
mon2: 1920x1080 at 1920x0

```
|--------------------------|--------------------------|
|                          |                          |
|                          |                          |
|                          |                          |
|                          |                          |
|--------------------------|--------------------------|
```

The point (1919.5, 69) is now treated as out of bounds, despite not being such.

drendog:
> The point (1919.5, 69) is now treated as out of bounds, despite not being such.

should we care since the functions like [move_cursor](https://github.com/hyprwm/wlroots-hyprland/blob/422207dbcf0949e28042403edab539159282885e/include/wlr/interfaces/wlr_output.h#L51) takes the cursor position which are integers, and `NEAREST_LAYOUT` takes care of giving the correct position while having the position as double?

vaxerski:
we should because wlroots is being removed and aquamarine takes doubles. Furthermore internally it's stored as a double, so when a user would be moving their mouse veeery slowly (e.g. 0.2px a frame) they would be blocked from leaving their screen:

1918 ok
1918.2 ok
1918.4 ok
1918.6 ok
1918.8 ok
1919 ok
1919.2 clamped to 1919
1919.2 clamped to 1919
....

drendog:
this should fix the sub-pixel coords clamp on edges, but can I ask you why use floating points for the cursor position? Shouldn't using fixed points be more appropriate maybe using `wl_fixed_t`?

vaxerski:
why not?

drendog:
just asking, nvm, then i found out that using double for cursor position it's not as bad an idea as I thought. Btw, can you give a feedback about last commit?

src/managers/PointerManager.cpp
```diff
@@ -635,7 +635,14 @@ Vector2D CPointerManager::closestValid(const Vector2D& pos) {
         float    distanceSq = __FLT_MAX__;
 
         for (auto& b : currentMonitorLayout.monitorBoxes) {
-            auto p      = b.closestPoint(vec);
+            auto p = b.closestPoint(vec);
+
+            auto max_x = std::nextafter(b.x + b.w, b.x + b.w - 1);
```
vaxerski, line 640: wth is nextafter?

drendog: "Returns the next representable value of `from` in the direction of `to`. "
So in this case returns the prev representable value from `x + width`.
If `x + width` equals `3840` like the 4k display, the prev representable by `double` type is `3839.9999999999995453...`, because floating point precision and consistency is a joke, and is for this that i'm worried about using floating point values for cursor position

vaxerski: round exists

src/managers/PointerManager.cpp
```diff
@@ -635,7 +635,13 @@ Vector2D CPointerManager::closestValid(const Vector2D& pos) {
         float    distanceSq = __FLT_MAX__;
 
         for (auto& b : currentMonitorLayout.monitorBoxes) {
-            auto p      = b.closestPoint(vec);
+            auto p = b.closestPoint(vec);
+
+            auto maxPoint = Vector2D{std::nextafter(b.x + b.w, -INFINITY), std::nextafter(b.y + b.h, -INFINITY)};
```
vaxerski, line 640: doesnt that in this case mean that maxPoint on a 1080p will be 1919,1079 which runs into problem numba 1 of 1919.5?

vaxerski:
I feel like the reasonable solution is:

Make padding a 4-value special value (like gaps_in for example) which describes per-side padding.

for example: 0,1,1,0 for 1px on bottom and right.

Then, assume there is a box around the cursor, in this case at cursor x,y with size 1,1

clamp that box to the region of all screens

return coord.

src/managers/PointerManager.cpp
drendog: no, on a 1080p `maxPoint` will be (1919.9999999999997726..., 1079.9999999999997726...)

drendog:
Isn't that padding solution, a solution to a problem that shouldn't even exist?
I mean, from what i got, it seems that the origin of the problem is due to wrong calcs of some method ([in this case `closestPoint`](https://github.com/hyprwm/hyprutils/blob/ff343e027951837f27f6afcd718dd9725a56d2b8/src/math/Box.cpp#L202C33-L202C45)) of the `CBox`, which returns an extra pixel when we deal with bottom and right edges, because it clamps to `x + w`, insteand of the prev representable value of `x + w`. 
Just look at the [current wlroots implementation of `wlr_box_closest_point`](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/util/box.c?ref_type=heads#L7), that it clamps up to `x + w - arbitrary number`, which in this case arbritrary number is `1/65536.0`.
If you agree about it i can open a PR to fix it on hyprutils.

Fixing by padding hack, may not be the ideal solution, and it makes the code unnecessarily complex when such a problem can be solved by modifying a few lines.

if we were to fix by adopting padding in each direction, should i do it with 4 boxes?

vaxerski:
returning x - epsilon,y - epsilon is a fine solution by me, as long as it doesnt break something else

vaxerski (approved):
with the hyprutils mr merged, this is fine. Thanks!

PR #6039, opened 2024-05-12
use add_custom_command for generating protocols

cnt0:
This fixes an issue with build error in case of e.g.

```bash
$ rm protocols/*.h
$ cmake --build build
```

The workaround was to 
```bash
$ touch CMakeLists.txt
``` 
With this PR, protocol files are properly regenerated with no extra efforts.

Also, resolve hyprwayland-scanner dependency via cmake instead of pkg-config.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,8 +110,10 @@ pkg_check_modules(deps REQUIRED IMPORTED_TARGET
     wayland-server wayland-client wayland-cursor wayland-protocols
     cairo pango pangocairo pixman-1
     libdrm libinput hwdata libseat libdisplay-info libliftoff libudev gbm
-    hyprwayland-scanner>=0.3.7 hyprlang>=0.3.2 hyprcursor>=0.1.7
+    hyprlang>=0.3.2 hyprcursor>=0.1.7
 )
+
+find_package(hyprwayland-scanner 0.3.7 REQUIRED)
 
 file(GLOB_RECURSE SRCFILES "src/*.cpp")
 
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -109,9 +109,11 @@ pkg_check_modules(deps REQUIRED IMPORTED_TARGET
     wayland-server wayland-client wayland-cursor wayland-protocols
     cairo pango pangocairo pixman-1
     libdrm libinput hwdata libseat libdisplay-info libliftoff libudev gbm
-    hyprwayland-scanner>=0.3.6 hyprlang>=0.3.2 hyprcursor>=0.1.7
+    hyprlang>=0.3.2 hyprcursor>=0.1.7
 )
 
+find_package(hyprwayland-scanner 0.3.6 REQUIRED)
```
fufexan, line 115: @vaxerski do we need 0.3.7?

vaxerski: now we do anyways (see conflict)

fufexan: Right. @cnt0 can you update it?

cnt0: fixed & rebased

github seems to show false diffs in this PR, I have no idea why. `CMakeLists.txt` is the only file here that is actually changed compared to the main branch.

fufexan: Caused by merging main into your branch. A rebase should fix it. If it doesn't, `git rebase --interactive HEAD~10` and `delete` any commits that aren't your own.

PR #6066, opened 2024-05-14
Add support for linux-drm-syncobj-v1

lucasreis1:
Adds support for linux-drm-syncobj-v1, the new explicit sync protocol. 

This **stricitly depends** on the [wlroots MR](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4262). Also, I'm heavily copying Simon's implementation on sway (https://github.com/swaywm/sway/pull/8156).

Draft for now while I follow both wlroots and sway's implementation. Comments from mantainers welcome to ensure this code makes any sense :)

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -206,6 +206,12 @@ void CCompositor::initServer() {
         m_sWLRLinuxDMABuf = wlr_linux_dmabuf_v1_create_with_renderer(m_sWLDisplay, 4, m_sWLRRenderer);
     }
 
+    if (wlr_renderer_get_drm_fd(m_sWLRRenderer) >= 0 &&
+        m_sWLRRenderer->features.timeline) {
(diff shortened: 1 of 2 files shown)


vaxerski:
if only it was that easy. It absolutely won't work unfortunately as just like you said on discord "the heavy load seems to be done entirely on wlroots", but we don't use those modules of wlroots _at all_, so all the heavylifting needs to be implemented here separately :)

lucasreis1:
> but we don't use those modules of wlroots at all

Do you mean the ones added by the new MR for explicit sync? This PR is done expecting these to be merged to hyprland's fork

vaxerski:
the changes are done to scene, renderer, and render_pass, all of which are not used in hl

lucasreis1:
Yeah, I see it now. Thanks for the input Vaxry

PR #6083, opened 2024-05-15
Add: Allow requestHyprpaper() and request() to return exit codes instead of void.

rofe33:

It allows `request()` and `requestHyprpaper()` to return exit codes.

This allows other script to check for the `hyprctl` return code, to diagnose the command.

For example, in python:

```python
    # ...
    hyprpaper = subprocess.run(
        [
            'hyprctl',
            'hyprpaper',
        ],
        capture_output=True,
    )

    # Check returncode and not the stdout string
    if (hyprpaper.returncode == 3 or
            hyprpaper.stdout.decode().strip().endswith('(3)')):
        print('Is hyprpaper running?')
        sys.exit(34)
     # ...
```

This code is from a script that I've created, to change randomly wallpapers for all the monitors.

- [pyprpaper](https://github.com/rofe33/pyprpaper) 

What should be the exit code for:

```cpp
    // request() function (line 98)
    if (ARGS < minArgs) {
        std::cout << "Not enough arguments, expected at least " << minArgs;
        return -1;
    }
```

All the other exit codes are the ones available in the string, for example:

```cpp
    if (SERVERSOCKET < 0) {
        std::cout << "Couldn't open a socket (1)";
        return 1;
    }
```

Yes.

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -90,24 +90,24 @@ std::vector<SInstanceData> instances() {
     return result;
 }
 
-void request(std::string arg, int minArgs = 0) {
+int request(std::string arg, int minArgs = 0) {
     const auto SERVERSOCKET = socket(AF_UNIX, SOCK_STREAM, 0);
 
     const auto ARGS = std::count(arg.begin(), arg.end(), ' ');
 
     if (ARGS < minArgs) {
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
ok, thanks

PR #6104, opened 2024-05-16
use add_custom_command for generating protocols

cnt0:
(this is v2 of [this PR](https://github.com/hyprwm/Hyprland/pull/6039), I made a new one to keep git history clean)

This fixes an issue with build error in case of e.g.

```bash
$ rm protocols/*.h
$ cmake --build build
```

The workaround was to 
```bash
$ touch CMakeLists.txt
``` 
With this PR, protocol files are properly regenerated with no extra efforts.

Also, resolve hyprwayland-scanner dependency via cmake instead of pkg-config.


fufexan (approved):
LGTM, thanks!

Issue #3241, opened 2023-09-10
[FR] Option to change position of red "Config error" bar


zjeffer:
### Description

This is an extremely small, low priority issue but when an error occurs, the red config error bar is displayed: 

![image](https://github.com/hyprwm/Hyprland/assets/4633209/1cc593d4-ae83-4292-8ece-0011d322e27d)

This bar always appears at the top of the screen, but this obscures my waybar. 
I'd like an option to always put this bar at the bottom of the screen. 

I'm aware it is possible to disable the bar, but I like to see if there are any problems. I often get errors due to plugins needing to be updated, so it's nice that the bar is there, but I don't want it to be in the same location as my waybar while I'm fixing the issue.

zjeffer:
> how can I disable the bar?

In your Hyprland config, you can enable `suppress_errors`:

https://wiki.hyprland.org/Configuring/Variables/#debug

debug {
  suppress_errors = true
}

Dreded:
> or u can close current one with `hyprctl seterror disable`

Thank you, the error is always Overtop of the text or website url I need to copy in order to fix the issue..

KeepThisTicket:
The list of errors doesn't fit in the preview window for me - where can I see the full list of errors?

PR #6111, opened 2024-05-16
Add option to change position of HyprError bar (#3241)

zjeffer:

Implements my own feature request: #3241.

Many people have their system bar up top. This means that when they have a config error, the error goes above the bar. 
Depending on your set `error_limit`, the errors might also cover up things like the top of your browser (which you might want to use to search for help), or the top of your terminal (which is where your prompt, and you'll use that to fix the error).

This PR makes the position configurable: 

* top bar (current): `debug:error_position = 0` (default)
* bottom bar: `debug:error_position = 1` (or I guess it would work with any non-zero value).

Tested in a nested debug instance. No problems found.

I would actually argue the default should be bottom, as I think more people have their bar up top than at the bottom. I assume the top is default for a reason?

Before merging this, I need to create a wiki PR to explain the new option. I'll create that after I get approval for this PR.


vaxerski (approved):
lgtm, thanks! wiki mr needed

PR #6120, opened 2024-05-17
cmake: make gprof optional for debug builds

cnt0:
Hyprland can't be built with cmake in debug mode on musl systems, because musl doesn't support the `-pg` flag.

This PR fixes that by making this flag optional (enabled by default).

Also, fix PIE flags: it should be `-fno-pie` for [compilation](https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html), `-no-pie` for [linking](https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html).


vaxerski (approved):
lgtm

PR #6136, opened 2024-05-18
Replace monitor rule when disabling head. Closes #5978

### LINKED ISSUE #5978: Can't temporarily disable monitor using external tools
### Hyprland Version

```sh
Hyprland, built from branch  at commit cba1ade848feac44b2eda677503900639581c3f4  ().
Date: 2024-05-04
Tag: , commits: @COMMITS@

flags: (if any)


System Information:
System name: Linux
Node name: upaya
Release: 6.6.30
Version: #1-NixOS SMP PREEMPT_DYNAMIC Thu May  2 14:32:50 UTC 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation HD Graphics 630 [8086:591b] (rev 04) (prog-if 00 [VGA controller])


os-release: BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="23.11.20240504.651b470"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 23.11 (Tapir)"
SUPPORT_END="2024-06-30"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="23.11 (Tapir)"
VERSION_CODENAME=tapir
VERSION_ID="23.11"


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

With 0.39.1, you could disable or enable a monitor using other tools, e.g. `kanshi` or `wdisplays`. With 0.40, these tools no longer work. Once a monitor is disabled, hyprland immediately re-enables it. This is a problem for those us that regular get up from their desks and plug and unplug to external monitors multiple times a day. On Discord, it was suggested to just manually update hyprland config, but this isn't a great solution as it involves reloading hyprland, and a rebuild if you are on nixos.

### How to reproduce

Have a laptop plugged into an external monitor. Use `wdisplays` to disable the laptop monitor. Note that it turns back on immediately.

_No response_

levnikmyskin:
Fixes Issue #5978. The main reason was that the monitor rule was not updated, and thus the head would get re-enabled when calling `performMonitorReload`

I'm not sure this is the best way of doing it. Would it be a better idea to make `Monitor.activeMonitorRule` a pointer to the actual rule?

See above.

### DIFF
--- a/src/protocols/OutputManagement.cpp
+++ b/src/protocols/OutputManagement.cpp
@@ -305,8 +305,8 @@ COutputConfiguration::COutputConfiguration(SP<CZwlrOutputConfigurationV1> resour
         LOGM(LOG, "disableHead on {}", PMONITOR->szName);
 
         PMONITOR->activeMonitorRule.disabled = true;
-        g_pHyprRenderer->applyMonitorRule(PMONITOR, &PMONITOR->activeMonitorRule, false);
         g_pConfigManager->replaceMonitorRule(PMONITOR->activeMonitorRule);
+        g_pHyprRenderer->applyMonitorRule(PMONITOR, &PMONITOR->activeMonitorRule, false);
     });
 
     resource->setTest([this](CZwlrOutputConfigurationV1* r) {
--- a/src/protocols/OutputManagement.cpp
+++ b/src/protocols/OutputManagement.cpp
@@ -357,6 +357,7 @@ bool COutputConfiguration::applyTestConfiguration(bool test) {
 
         SMonitorRule newRule = PMONITOR->activeMonitorRule;
         newRule.name         = PMONITOR->szName;
+        newRule.disabled     = false;
 
         if (head->committedProperties & COutputConfigurationHead::eCommittedProperties::OUTPUT_HEAD_COMMITTED_MODE) {
             newRule.resolution  = {head->state.mode->getMode()->width, head->state.mode->getMode()->height};
@@ -381,7 +382,8 @@ bool COutputConfiguration::applyTestConfiguration(bool test) {
         // reset properties for next set.
         head->committedProperties = 0;
 
-        g_pConfigManager->appendMonitorRule(newRule);
+        if (!g_pConfigManager->replaceMonitorRule(newRule))
+            g_pConfigManager->appendMonitorRule(newRule);
         g_pConfigManager->m_bWantsMonitorReload = true;
     }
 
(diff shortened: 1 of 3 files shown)


src/protocols/OutputManagement.cpp
```diff
@@ -306,6 +306,7 @@ COutputConfiguration::COutputConfiguration(SP<CZwlrOutputConfigurationV1> resour
 
         PMONITOR->activeMonitorRule.disabled = true;
         g_pHyprRenderer->applyMonitorRule(PMONITOR, &PMONITOR->activeMonitorRule, false);
+        g_pConfigManager->replaceMonitorRule(PMONITOR->activeMonitorRule);
```
vaxerski, line 309: you're making it a bool but ignoring the retval. Plus, shouldnt this be first?

levnikmyskin: Yeah about the retval, it looks stupid. The idea was that `replaceMonitorRule` could be used somewhere else in the code, where it might be useful to have the retval. 
I can remove it if you don't like that. Regarding the position, yes probably you're right. In general I would have preferred turning the `activeMonitorRule` to a pointer, but I don't know how much of a good idea that would be (I mean, shouldn't this rule replacing happen also somewhere else?). For now, I'll just swap the position

levnikmyskin:
As mentioned in the issue thread, it looks like with this I can disable monitor via kanshi, but cannot re-enable it. Probably because of a similar problem. I'll try to fix that too

vaxerski:
tag me when ready and tested

levnikmyskin:
@vaxerski this is ready to be reviewed

src/protocols/OutputManagement.cpp
```diff
@@ -305,6 +308,7 @@ COutputConfiguration::COutputConfiguration(SP<CZwlrOutputConfigurationV1> resour
         LOGM(LOG, "disableHead on {}", PMONITOR->szName);
 
         PMONITOR->activeMonitorRule.disabled = true;
+        g_pConfigManager->replaceMonitorRule(PMONITOR->activeMonitorRule);
```
vaxerski, line 311: should be if(!...) append() like below no?

vaxerski (commented):
rest ok, I assume tested and fixes the issue?

levnikmyskin:
> rest ok, I assume tested and fixes the issue?

Yes, a few people reported back and it looks like we can close the issue

vaxerski (approved):
thanks!

Issue #5631, opened 2024-04-18
Allow pass dispatcher to map to different keys before sending to window


joshheyse:
### Description

I'd like to take a look at implementing https://github.com/hyprwm/Hyprland/discussions/3510. After looking in into the existing `pass` dispatcher in (CKeyboardManager::pass), I am proposing the following config and implentation.

```
bind = <MODKEYS><KEY>,passThru,<NEW_MODKEYS><NEW_KEY>[,WINDOW_RULES]
```

The passThru dispatcher would bind to `MODEKEYS/KEY` and only be triggered iff the current window matched the `WINDOW_RULE` or the `WINDOW_RULE` was empty.  The `passThru` would then pass the `NEW_MODKEYS/NEW_KEY` down to the current window. `WINDOW_RULES` would follow the same logic as [Window Rules V2](https://wiki.hyprland.org/Configuring/Window-Rules/#window-rules-v2). 

I am also proposing that the `pass` dispatcher (config, commandName, functions, and documentation) be deprecated in-favor of `passTo` to help disambiguate between the explicit global hot key support `passTo` and `passThru`.  A warning would be added to the logs if `pass` was used in the user configuration and 1 - 2 releases prior to the removal a message would be displayed using the same functionality as the "you are using a default config" message.

Any suggestions or changes are welcome, just tossing out one potential solution to get the conversation started.

## Here is an example of a personal use case for this feature.

I switch between macOS and Hyperland on my Asahi Linux laptop frequently and battling with different keyboard shortcuts combinations.  (This is my second time typing this issue up in GitHub because I accidently hit Super+C when I meant to copy and instead it closed my Firefox window.)

```
bind = SUPER,C,passThru,CTRL,C
bind = SUPER,V,passThru,CTRL,V
bind = SUPER,X,passThru,CTRL,X
bind = SUPER,C,passThru,CTRL_SHIFT,C,class:(kitty),title:(kitty)
bind = SUPER,V,passThru,CTRL_SHIFT,V,class:(kitty),title:(kitty)
bind = SUPER,X,passThru,CTRL_SHIFT,X,class:(kitty),title:(kitty)
```

joshheyse:
If this looks good, let me know and I will polish it up and submit a PR. Thanks!

vaxerski:
it should be named something else (e.g. sendshortcut) but other than that makes sense

joshheyse:
### Description

[...]

PR #6174, opened 2024-05-21
Added new dispatcher (sendshortcut)

yungcxn:
Implemented the Request from https://github.com/hyprwm/Hyprland/issues/5631

`bind = <MODKEYS><KEY>,sendShortcut,<NEW_MODKEYS><NEW_KEY>[,WINDOW_RULES]`

Window rules are optional, if not given, the keys are sent to the current focused window.
This is a really nice feature so one wouldn't need simple key remaps aswell.

I want to mention the problem with mapping a single keyname (e.g. "g") to a keycode (not keysym).
Needed to add a cache where I add keycodes to keys at the first time they are pressed (only those keys that are sent to a certain window, not all keys). Therefore I added a new field in the KeyManager Header. Code is commented adequately

Ready I guess, did the testing, but I don't know how it behaves in xwayland world....


src/managers/KeybindManager.hpp
```diff
@@ -95,6 +95,11 @@ class CKeybindManager {
 
     std::list<SKeybind>                                               m_lKeybinds;
 
+    //since we cant find keycode through keyname in xkb:
+    //on sendshortcut call, we once search for keyname (e.g. "g") the correct keycode (e.g. 42)
+    //and cache it in this map to make sendshortcut calls faster
+    std::unordered_map<std::string, xkb_keycode_t>                    m_mKeyToCodeCache;
```
vaxerski, line 101: different keyboards can have different layouts, this can break.

vaxerski (commented):
instead of finding a code like a caveman, you can create a temporary keymap with your stuff at 1, 2, 3, etc, set that keymap, send it to clients, sent the codes, reset the keymap to what it was, and smile

yungcxn:
What do you mean by temporary keymap?

yungcxn:
wtype does exactly the same thing with iterating through all codes (https://github.com/atx/wtype/blob/d71be3a7b3f93b534a2823fd68cabd7ac2a02359/main.c#L171),  "caveman-like"  aswell :)

I could add the current keyboard to the map key, that  would allow differentiation between keyboards, but completely building a whole keymap from file would be 1. unnecessary 2. time consuming. Or do you know of any other alternative?

vaxerski:
you can add the current keyboard I guess

src/managers/KeybindManager.hpp
yungcxn: Done, added the pointer to the current keyboard to the string to differentiate. Didnt use `std::pair<wlr_keyboard*, std:.string>` since hashing doesnt work with this smh

yungcxn:
Note: I didn't use your XWTOXW variable (checked if windows are xwayland) from pass. Can you check, if everything works in XWayland aswell? I did it, everything worked besides sending to "xev"

vaxerski (commented):
can you clang-format it first, it's painful to read

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+    //args=<NEW_MODKEYS><NEW_KEY>[,WINDOW_RULES]
+    const auto ARGS = CVarList(args, 3);
+    if (ARGS.size() != 3) {
+        Debug::log(ERR, "sendshortcut: invalid args");
+        return;
+    }
+
+    const auto MOD = g_pKeybindManager->stringToModMask(ARGS[0]);
+    const auto KEY = ARGS[1];
+    uint32_t   keycode;
```
vaxerski, line 2028: no uninitialized primitives

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+    bool       isMouse = 0;
+
+    //oriented from parseKey in ConfigManager
+    if (isNumber(KEY) && std::stoi(KEY) > 9) {
```
vaxerski, line 2032: no {} around short ifs

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+
+        // here, we need to find the keycode from the key name
+        // this is not possible through xkb's lib, so we need to iterate through all keycodes
+        // once found, we save it to the cache
+
+        const auto KEYSYM = xkb_keysym_from_name(KEY.c_str(), XKB_KEYSYM_CASE_INSENSITIVE);
+        keycode           = 0;
+
+        const auto KB            = g_pSeatManager->keyboard;
+        const auto KEYPAIRSTRING = std::to_string(KB) + KEY;
```
vaxerski, line 2053: wot de hell does to_string even do here? address? this is not verbose enough, it'd be better to use std::format.

Furthermore, KB can be an `::expired()` ptr.

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+
+        // check if KEY is in m_mKeyToCodeCache
+        if (g_pKeybindManager->m_mKeyToCodeCache.find(KEYPAIRSTRING) == g_pKeybindManager->m_mKeyToCodeCache.end()) {
+            // find keycode and add to cache
+            // get current window
+            xkb_keymap*   km = KB->wlr()->keymap;
```
vaxerski, line 2059: same as above, KB can be expired

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+                    g_pKeybindManager->m_mKeyToCodeCache[KEYPAIRSTRING] = keycode;
+                }
+            }
+
+            if (keycode == 0) { //wasnt found
+                Debug::log(ERR, "sendshortcut: key not found");
+                return;
+            }
+
+        } else {
```
vaxerski, line 2083: no {}

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+        // once found, we save it to the cache
+
+        const auto KEYSYM = xkb_keysym_from_name(KEY.c_str(), XKB_KEYSYM_CASE_INSENSITIVE);
+        keycode           = 0;
+
+        const auto KB            = g_pSeatManager->keyboard;
+        const auto KEYPAIRSTRING = std::to_string(KB) + KEY;
+
+        // check if KEY is in m_mKeyToCodeCache
+        if (g_pKeybindManager->m_mKeyToCodeCache.find(KEYPAIRSTRING) == g_pKeybindManager->m_mKeyToCodeCache.end()) {
```
vaxerski, line 2056: map has `.contains` iirc

src/managers/KeybindManager.hpp
```diff
@@ -95,6 +95,12 @@ class CKeybindManager {
 
     std::list<SKeybind>                                               m_lKeybinds;
 
+    //since we cant find keycode through keyname in xkb:
+    //on sendshortcut call, we once search for keyname (e.g. "g") the correct keycode (e.g. 42)
+    //and cache it in this map to make sendshortcut calls faster
+    //we also store the keyboard pointer (in the string) to differentiate between different keyboard (layouts)
+    std::unordered_map<std::string, xkb_keycode_t> m_mKeyToCodeCache;
```
vaxerski, line 102: this cache needs to be cleared on a keyboard's keymap change. Easiest solution is to just clear it in .keymap of any kb

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+            Debug::log(ERR, "sendshortcut: window not found");
+            return;
+        }
+
+        if (!g_pSeatManager->keyboard) {
+            Debug::log(ERR, "No kb in sendshortcut?");
+            return;
+        }
+
+        if (isMouse == 0)
```
vaxerski, line 2120: !isMouse

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+                Debug::log(ERR, "sendshortcut: key not found");
+                return;
+            }
+
+        } else {
+            keycode = g_pKeybindManager->m_mKeyToCodeCache[KEYPAIRSTRING];
+        }
+    }
+
+    if (keycode == 0) {
```
vaxerski, line 2088: !keycode

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,162 @@ void CKeybindManager::pass(std::string regexp) {
...
+        return;
+    }
+
+    const std::string regexp = ARGS[2];
+
+    PHLWINDOW         PWINDOW          = nullptr;
+    const auto        LASTWINDOWEXISTS = g_pCompositor->m_pLastWindow.lock();
+
+    if (!LASTWINDOWEXISTS) {
+        Debug::log(ERR, "sendshortcut: no window to return to");
```
vaxerski, line 2099: this is not fatal, is it? Just unfocus afterwards in that case

vaxerski (commented):
a bit of a chonker but most are style nits

src/managers/KeybindManager.cpp
yungcxn: Is unfocus just putting the focus to the last focused window? If so, I would do that

src/managers/KeybindManager.hpp
vaxerski: better place would be the .keyboard listener in setupKeyboard in CInputManager

src/managers/KeybindManager.cpp
vaxerski: unfocus is putting focus on nothing

src/managers/KeybindManager.cpp
vaxerski: whatever you please, worth noting the above likely wont compile

src/managers/KeybindManager.cpp
yungcxn: What should I do then exactly in this case? Unfocus before focusing the old window?

vaxerski: if there is no old window to return to, after sending the shortcut, unfocus?

yungcxn: this is before changing the focus

vaxerski: it will not perform a sendshortcut if there is no focused window

yungcxn: ahhh i get it :)

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done (did it with :p and void pointer cast)

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.hpp
yungcxn: Done, please check the location, I put it in the if block

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,163 @@ void CKeybindManager::pass(std::string regexp) {
...
+        const auto KB = g_pSeatManager->keyboard;
+
+        if (!KB) {
+            Debug::log(ERR, "sendshortcut: no kb");
+            return;
+        }
+
+        // create a unique string for the keypair
+        // casting wlr_keyboard* to void* for formatting
+        const auto KEYPAIRSTRING = std::format("{:p}{}", static_cast<void*>(KB->wlr()), KEY);
```
vaxerski, line 2061: generally we do `{}` with `(uintptr_t)var.get()`

src/managers/KeybindManager.cpp
```diff
@@ -2013,6 +2014,163 @@ void CKeybindManager::pass(std::string regexp) {
...
+                g_pSeatManager->state.keyboardFocusResource.reset();
+            } else {
+                g_pSeatManager->state.pointerFocus = nullptr;
+                g_pSeatManager->state.pointerFocusResource.reset();
+            }
+        }
+
+        const auto SL = PWINDOW->m_vRealPosition.goal() - g_pInputManager->getMouseCoordsInternal();
+
+        if (!LASTWINDOW) {
```
vaxerski, line 2165: this if is wrong, remove it. If you found a window and focused to it, you need to get back to the last.

src/managers/KeybindManager.cpp
yungcxn: Done

src/managers/KeybindManager.cpp
yungcxn: Done

yungcxn:
Can you please confirm that sending to XWayland works aswell? Now I can't send shortcuts to something like xev

vaxerski:
doesn't look solvable to me. I think XWayland processes events separately and thus our timing tricks dont work. The question is whether we care. I would say "no".

This diff removes some redundant comments, cleans up code, and makes wl -> xwl work, leaving xwl -> xwl in a kinda broken state (will send to current)

[patch.txt](https://github.com/hyprwm/Hyprland/files/15415842/patch.txt)

test it, commit it, and if everything's ok we can ship this

yungcxn:
Is it my fault or where is the diff? File is empty

vaxerski:
aw fuck

vaxerski:
are you happy with this? does it work for your usecase or are there things left?

yungcxn:
It works
I am happy
and this feature is extremely nice

vaxerski (approved):
lgtm, just needs a wiki mr

PR #6182, opened 2024-05-21
Add option to disable window direction monitor fallback

shezdy:
Adds an option to disable monitor fallback when focusing or moving a window in a direction. A value of `0` will disable moving to an adjacent monitor when focusing/moving a window over the edge of a monitor. A value of `1` (default) will keep behavior as it is now.

With `window_direction_monitor_fallback = 0`:
`movefocus l` will do nothing if the focused window is on the left edge of the monitor (or focus the next window on the workspace depending on `no_focus_fallback`)

`movewindow l` will do nothing if the focused window is already on the left edge of the monitor

This was a feature request for my plugin (https://github.com/shezdy/hyprsplit/issues/20), but I think it makes sense to implement it here.

I don't think the name `window_direction_monitor_fallback` is ideal but I can't think of anything better so please suggest a better name if you have one.

ready, needs wiki mr

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1386,7 +1386,7 @@ PHLWINDOW CCompositor::getWindowInDirection(PHLWINDOW pWindow, char dir) {
         return nullptr;
 
     // 0 -> history, 1 -> shared length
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
coolio, wiki mr needed

PR #6206, opened 2024-05-23
Fix tablet mapping when mapped region is specified

sifmelcara:

When `region_position` or `region_size` is set in the config (non-empty `boundBox`), cursor is mapped to wrong coordinate because `CBox::translate` mutates `TAB->boundBox`, making all subsequent coordinate calculations wrong.

Ready

### DIFF
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -674,8 +674,11 @@ void CPointerManager::warpAbsolute(Vector2D abs, SP<IHID> dev) {
                 }
             }
 
-            if (!TAB->boundBox.empty())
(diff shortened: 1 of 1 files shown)


sifmelcara:
Updated the commit to also account for the case where `region_position` is set but `region_size` is unset. (I doubt anyone would do this though)

Tested
* Only set `region_position`
* Only set `region_size`
* Set both

with libwacom and multi monitor set-up; my `input:tablet:output` is set to a monitor that is not at `(0, 0)`.

Agent00Ming (approved):
Can confirm that this does indeed fix region size and position mapping defined inside of hyprland config.

vaxerski (approved):
thanks :)

Issue #6079, opened 2024-05-15
add a `tag` property to window, allow update tag with dispatch, and match with tag


rtgiskard:
### Description

Currently windowrule only match with limitted fields, which can't be modified by the user with convenience, if every window has a list of tags, and there is a dispatcher to manipulate (add/delete tag) it, and windowrule is able to match with tags, then user will be able to do many things, for example:

1. adjust opacity of a specific window with keybindings on the fly
2. mark windows in any workspace, and operate with the marks

PR #6211, opened 2024-05-23
feat: add tag dispacther and window rule

rtgiskard:

this resolves: https://github.com/hyprwm/Hyprland/issues/6079

1. ~the tag rule will update tags which affect the rule matched by tag, may be a rule rescan after tag applied~
2. ~dynamic tag shouldn't be permernant~
3. ~tag matched rule not get executed right after tag is applied by dispatcher~
4. hyprctl bash/fish/zsh completion is not synced yet, just not sure how to update

Likely to be ready except for the above note

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1090,7 +1090,8 @@ std::vector<SWindowRule> CConfigManager::getMatchingRules(PHLWINDOW pWindow, boo
             try {
                 if (rule.szValue.starts_with("tag:") && !tags.isTagged(rule.szValue.substr(4)))
                     continue;
-                else if (rule.szValue.starts_with("title:")) {
+
+                if (rule.szValue.starts_with("title:")) {
                     std::regex RULECHECK(rule.szValue.substr(6));
 
                     if (!std::regex_search(title, RULECHECK))
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -995,7 +995,7 @@ SMonitorRule CConfigManager::getMonitorRuleFor(const CMonitor& PMONITOR) {
     Debug::log(WARN, "No rule found for {}, trying to use the first.", PMONITOR.szName);
 
     for (auto& r : m_dMonitorRules) {
-        if (r.name == "") {
+        if (r.name.empty()) {
             return r;
         }
     }
@@ -1088,7 +1088,7 @@ std::vector<SWindowRule> CConfigManager::getMatchingRules(PHLWINDOW pWindow, boo
                 if (rule.szValue.starts_with("tag:")) {
                     const auto tag = rule.szValue.substr(4);
 
-                    if (!pWindow->m_tags.contains(tag))
+                    if (!pWindow->isTagged(tag))
                         continue;
                 } else if (rule.szValue.starts_with("title:")) {
                     std::regex RULECHECK(rule.szValue.substr(6));
@@ -1107,33 +1107,33 @@ std::vector<SWindowRule> CConfigManager::getMatchingRules(PHLWINDOW pWindow, boo
             }
         } else {
             try {
-                if (rule.szTag != "") {
-                    if (!pWindow->m_tags.contains(rule.szTag))
+                if (!rule.szTag.empty()) {
+                    if (!pWindow->isTagged(rule.szTag))
                         continue;
                 }
 
-                if (rule.szClass != "") {
+                if (!rule.szClass.empty()) {
                     std::regex RULECHECK(rule.szClass);
 
                     if (!std::regex_search(appidclass, RULECHECK))
                         continue;
                 }
 
-                if (rule.szTitle != "") {
+                if (!rule.szTitle.empty()) {
                     std::regex RULECHECK(rule.szTitle);
 
                     if (!std::regex_search(title, RULECHECK))
                         continue;
                 }
 
-                if (rule.szInitialTitle != "") {
+                if (!rule.szInitialTitle.empty()) {
                     std::regex RULECHECK(rule.szInitialTitle);
 
                     if (!std::regex_search(pWindow->m_szInitialTitle, RULECHECK))
                         continue;
                 }
 
-                if (rule.szInitialClass != "") {
+                if (!rule.szInitialClass.empty()) {
                     std::regex RULECHECK(rule.szInitialClass);
 
                     if (!std::regex_search(pWindow->m_szInitialClass, RULECHECK))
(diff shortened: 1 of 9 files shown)


vaxerski (commented):
one thing stands out to me: you apply a tag in dynamic but dont clear them before. This means a window won't lose a tag when it stops matching a dynamic prop (e.g. windowrulev2 = tag:cock, floating:1 -> toggling float will permanently add the tag cock)

I think a reasonable approach is to just make a struct SWindowTag which has a `bool dynamic`, and clear dynamic tags before reloading dynamic rules

vaxerski:
tags added via a dispatcher should be permanent though so those would not be tagged as dynamic I guess

rtgiskard:
Well, I'll check later, I need to have a late-night snack. 😂

rtgiskard:
Dynamic tag has been fixed with a "*" suffix to tell it apart

rtgiskard:
Dynamic tag is implemented with a "*" suffix compared to static tag via dispatcher.

Tag rule will queue another rescan for window rules, while tag from tag will not be queued for another rescan to avoid infinite recursion like this:

```
windowrulev2 = tag code, tag:cpp
windowrulev2 = opacity 0.7, tag:code
```
though, it would be ok with this if allow tag from tag recursion:

```
windowrulev2 = tag +code, tag:cpp
windowrulev2 = opacity 0.7, tag:code
```

src/desktop/Window.cpp
```diff
@@ -856,21 +856,53 @@ bool CWindow::isTagged(const std::string& tag, bool strict) {
...
+                auto w = static_cast<CWindow*>(data);
+
+                w->m_vMatchedRules = g_pConfigManager->getMatchingRules(w->m_pSelf.lock());
+                for (auto& r : w->m_vMatchedRules)
+                    if ((r.v2 && !r.szTag.empty() && w->isTagged(r.szTag)) || (!r.v2 && r.szValue.starts_with("tag:") && w->isTagged(r.szValue.substr(4))))
+                        w->applyDynamicRule(r);
+
+                // NOTE: if tag from tag, then the nested tag's rescan will be ignored
+                wl_event_source_remove(w->m_tagScanSource);
+                w->m_tagScanSource = nullptr;
```
rtgiskard, line 903: reset `m_tagScanSource` at the end of callback to avoid tag toggle from tag match, which will cause infinite loop in such case.

src/desktop/Window.cpp
```diff
@@ -856,21 +856,53 @@ bool CWindow::isTagged(const std::string& tag, bool strict) {
...
+
+    // re run tag rules after update tag
+    queueScanForTagMatch();
+}
+
+void CWindow::queueScanForTagMatch() {
+    // TODO: assume single thread for window operation
+    if (!m_tagScanSource)
+        m_tagScanSource = wl_event_loop_add_idle(
+            g_pCompositor->m_sWLEventLoop,
```
rtgiskard, line 892: Here window operation is assumed to be run in single thread, if that's not TRUE, there'll need to be an extra lock or something better as the fix

src/managers/KeybindManager.cpp
```diff
@@ -1928,6 +1930,21 @@ void CKeybindManager::focusWindow(std::string regexp) {
...
+
+    if (vars.size() == 1)
+        PWINDOW = g_pCompositor->m_pLastWindow.lock();
+    else if (vars.size() == 2)
+        PWINDOW = g_pCompositor->getWindowByRegex(vars[1]);
+    else
+        return;
+
+    if (PWINDOW && PWINDOW->m_tags.applyTag(vars[0]))
+        PWINDOW->updateDynamicRules();
```
rtgiskard, line 1945: Upon `updateDynamicRules()`, which one should I call to ensure necessary render update?

src/desktop/Window.cpp
rtgiskard: For a case like this, it's reasonable:
```
windowrulev2 = tag +code, tag:cc
windowrulev2 = tag +code, tag:go
windowrulev2 = tag +code, tag:py

windowrulev2 = bordercolor rgba(800000cc), tag:cc
windowrulev2 = bordercolor rgba(008000cc), tag:go
windowrulev2 = bordercolor rgba(000080cc), tag:py

windowrulev2 = opacity 0.7, tag:code
```

While, with latest commit, the matched rules returned from config is already a full match with tag rule considered, there is no need to queue rescan in `applyTag()`

rtgiskard: It might be reasonable for a case like this:

```
windowrulev2 = tag +code, tag:cpp
windowrulev2 = tag +code, tag:go
windowrulev2 = tag +code, tag:py

windowrulev2 = bordercolor rgba(800000cc), tag:cpp
windowrulev2 = bordercolor rgba(008000cc), tag:go
windowrulev2 = bordercolor rgba(000080cc), tag:py

windowrulev2 = opacity 0.8 override, tag:code
windowrulev2 = opacity 0.4 override, tag:cpp
```

rtgiskard:
Now tags works perfect, seems to be ready

rtgiskard:
Now tags works perfect, seems to be ready

rtgiskard:
Doc for tag is updated here: https://github.com/hyprwm/hyprland-wiki/pull/660

src/desktop/Window.cpp
vaxerski: why not just assign 2 tags?

src/managers/KeybindManager.cpp
vaxerski: yes

src/desktop/Window.cpp
rtgiskard: Dynamic tags is more to be automatic, static tag is generally assigned manually. With dynamic tag, static tags can be grouped, or let me say, tags can be tagged too. It's like abstract class and sub subclass to some degree, and currently this does not bring more footprint than a local tag emulation in matching tags. the general use case is documented in https://github.com/hyprwm/hyprland-wiki/pull/660

vaxerski: mhm ok

src/config/ConfigManager.cpp
```diff
@@ -1080,12 +1081,16 @@ std::vector<SWindowRule> CConfigManager::getMatchingRules(PHLWINDOW pWindow, boo
...
+
     for (auto& rule : m_dWindowRules) {
         // check if we have a matching rule
         if (!rule.v2) {
             try {
-                if (rule.szValue.starts_with("title:")) {
-                    // we have a title rule.
+                if (rule.szValue.starts_with("tag:") && !tags.isTagged(rule.szValue.substr(4)))
+                    continue;
+                else if (rule.szValue.starts_with("title:")) {
```
vaxerski, line 1093: else is redundant

src/helpers/TagKeeper.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <string>
+#include <set>
+
+class TagKeeper {
```
vaxerski, line 6: CTagKeeper

src/helpers/TagKeeper.hpp
```diff
@@ -0,0 +1,18 @@
...
+  private:
```
vaxerski, line 7: private comes last

vaxerski (commented):
sum style nits, rest ok

vaxerski:
class CTagKeeper but file TagKeeper (just like everything else) and then gtg

rtgiskard:
> class CTagKeeper but file TagKeeper (just like everything else) and then gtg

Well, mistake. By the way, would you like to try the bleeding-edge cpp modules?

vaxerski:
nope, time will come for them in the future thugh prolly

vaxerski (approved):
lgtm

PR #6220, opened 2024-05-23
Add back nvidia hardware cursor

thejch:
Fixes the issue that nvidia hardware cursors no longer work after the new cursor implementation
Code is simply taken from https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4596/diffs#diff-content-e3ea164da86650995728d70bd118f6aa8c386797

Don't know if it would break anything but this just restores the behavior from before the new cursor impl

Should be ready


vaxerski:
the hw cursor is likely not cleared.

PR #6246, opened 2024-05-26
fix: prevent potential crash in handleBezier and in refreshRate due to invalid input

tachyglossues:

before, when the conf file contained anything other than a float in the 4 bezier curve generation points for annimations, it crashed Hyprland, which now returns an error message.

This is my first PR, so please check that I haven't done anything wrong. 

It's ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1821,26 +1821,20 @@ std::optional<std::string> CConfigManager::handleBezier(const std::string& comma
     const auto  ARGS = CVarList(args);
 
     std::string bezierName = ARGS[0];
+        if (ARGS[1] == "" || ARGS[2] == "" || ARGS[3] == "" || ARGS[4] == "")
+            return "too few arguments";
+        if (ARGS[5]!= "")
+            return "too many arguments";
 
-    if (ARGS[1] == "")
-        return "too few arguments";
-    float p1x = std::stof(ARGS[1]);
-
-    if (ARGS[2] == "")
-        return "too few arguments";
-    float p1y = std::stof(ARGS[2]);
-
-    if (ARGS[3] == "")
-        return "too few arguments";
-    float p2x = std::stof(ARGS[3]);
-
-    if (ARGS[4] == "")
-        return "too few arguments";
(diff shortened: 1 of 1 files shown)


src/config/ConfigManager.cpp
```diff
@@ -1821,26 +1821,20 @@ std::optional<std::string> CConfigManager::handleBezier(const std::string& comma
     const auto  ARGS = CVarList(args);
 
     std::string bezierName = ARGS[0];
+        if (ARGS[1] == "" || ARGS[2] == "" || ARGS[3] == "" || ARGS[4] == "")
```
vaxerski, line 1824: you can check .size

tachyglossues:
```suggestion
        if (ARGS.size() < 5)
```

Micovec:
Related #6249 I think that a util function should be created instead (as suggested by Vaxry in that issue).

tachyglossues:
> Related #6249 I think that a util function should be created instead (as suggested by Vaxry in that issue).

In this PR I've just fixed #6249 and I don't know if it's better to create a function. Because a simple try/catch is enough to solve this type of problem. And from the few tests I've done I haven't found the same problem elsewhere.

Micovec:
I don't want to be rude, but there are **many more** places with the same problem. Just look two lines up from your first change.

Also I'm not a reviewer but you don't want to do:

```cpp
...
} catch (...) {
    // An oopsie happened
}
```

It catches everything including things like [std::bad_alloc](https://en.cppreference.com/w/cpp/memory/new/bad_alloc).

bakicelebi:
@tachyglossues are you going to continue to work on this? This issue is annoying the hell out of me with vscode autosave. If you are not gonna continue working on this I will raise my own PR.

tachyglossues:
> @tachyglossues are you going to continue to work on this? This issue is annoying the hell out of me with vscode autosave. If you are not gonna continue working on this I will raise my own PR.

No, because the arguments are handled differently in the code, and I don't have the necessary C++ skills to refactor the code without making a disaster of it.

vignesh1507 (changes requested):
You can add few type hints to your updated code snippet.

PR #6086, opened 2024-05-15
xwayland/xwayland_shell: move to hyprland impl

vaxerski:
Moving wlr_xwayland to a hyprland impl.

TODO:
 - [x] c+p
     - [x] support outgoing
     - [ ] support INCR
     - [x] support incoming
 - [ ] dnd
 - [x] Support no_xwayland builds
 - [x] set xwayland cursor
 - [x] Meson
 - [x] Nix
 - [x] Wiki
 - [x] Testing

Bugs:
 - [x] Some apps can't copy/paste because of mime mismatches

PR #6262, opened 2024-05-27
meson: explicitly depend on xcb-icccm

jbeich:
Regressed by #6086. Prevents disabling xwayland support in wlroots independently from Hyprland.

Also, affects unusual downstream packaging. For example, FreeBSD uses shared wlroots (with sway-git) to aid debugging and avoid leaking wlroots implementation details into hyprland package.


fufexan (approved):
LGTM.

PR #6263, opened 2024-05-27
fix: make new_window_takes_over_fullscreen use the new window's workspace

cdubthecoolcat:
Currently the setting `new_window_takes_over_fullscreen` is applied to the currently focused workspace instead of the workspace the new window is appearing on.

Minimally tested but should be fine

I haven't run into issues regarding this so i think it's okay. Let me know if something isn't though.


vaxerski (approved):
thanks!

Issue #6154, opened 2024-05-19
Windows spawn somewhere in the corner


DADA30000:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit f8857e6072bd85b95393499688872aaf7f088b5b  ().
Date: 2024-05-18
Tag: , commits: 

flags: (if any)


System Information:
System name: Linux
Node name: nixos
Release: 6.9.1
Version: #1-NixOS SMP PREEMPT_DYNAMIC Fri May 17 10:18:09 UTC 2024


GPU information: 
	Flags: bus master, VGA palette snoop, 66MHz, medium devsel, latency 64, NUMA node 0, IOMMU group 11
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU117 [GeForce GTX 1650] [10de:1f82] (rev a1) (prog-if 00 [VGA controller])


os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.05.20240517.4a6b83b"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.05 (Uakari)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.05 (Uakari)"
VERSION_CODENAME=uakari
VERSION_ID="24.05"


plugins:
  hyprexpo by Vaxry ver 1.0

```

</details>

### Bug or Regression?

Bug

### Description

https://github.com/hyprwm/Hyprland/assets/73874445/eda0c1b4-1b97-451d-b9da-d6bf8ba0b465

### How to reproduce

Use latest hyprland from git

The-Briel-Deal:
Please attach config. Cannot repro.

DADA30000:
[hyprland.conf.zip](https://github.com/hyprwm/Hyprland/files/15374722/hyprland.conf.zip)

The-Briel-Deal:
Can you also give steps to reproduce this issue? And can you explain what the issue is? It looks like the auth pop up is popping up in the top right. Is that the the problem?

DADA30000:
And steps to reproduce
1. Install hyprland using flake on NixOS
2. Open floating window

[...]

PR #6285, opened 2024-05-31
fix a few asan reported issues and a coredump on exit

gulafaran:
the wl_event_resource was running upon destruction of the compositor causing a null pointer segfault in onX11Event so ensure the event is removed upon destruction, also free the memory allocated by xcb_errors_context_new and finally call xcb_disconnect on the connection to free the fd and its memory.

also free the eventsources in hyprctl and eventloopmanager.

### DIFF
--- a/src/debug/HyprCtl.hpp
+++ b/src/debug/HyprCtl.hpp
@@ -8,6 +8,7 @@
 class CHyprCtl {
   public:
     CHyprCtl();
+    ~CHyprCtl();
 
     std::string         makeDynamicCall(const std::string& input);
     SP<SHyprCtlCommand> registerCommand(SHyprCtlCommand cmd);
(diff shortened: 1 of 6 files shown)


src/debug/HyprCtl.hpp
```diff
@@ -25,6 +26,7 @@ class CHyprCtl {
     void                             startHyprCtlSocket();
 
     std::vector<SP<SHyprCtlCommand>> m_vCommands;
+    wl_event_source*                 m_eventSource{nullptr};
```
vaxerski, line 29: = nullptr; please

vaxerski (commented):
rest lgtm

src/debug/HyprCtl.hpp
gulafaran: changed. <3

vaxerski (approved):
thanks!

obivan:
This one is merged, but I don't see any related commits in main. Pardon my ignorance, but... how so?

MightyPlaza:
> This one is merged, but I don't see any related commits in main. Pardon my ignorance, but... how so?

possibly https://github.com/hyprwm/Hyprland/issues/6154#issuecomment-2144034315

vaxerski:
thanks for spotting, I accidentally nuked a bit of history yesterday and missed restoring this one. Will do later today.

PR #3186, opened 2023-09-05
Formatter rework

vaxerski:
Uses `std::format` everywhere instead of the poo poo `vasnprintf`

Needs to be merged ASAP - will make a conflict with pretty much anything

PR #6331, opened 2024-06-04
log: log with local timezone

rtgiskard:

Now that local timezone is there, why not?

`libc++` seems to support the format of `std::chrono::hh_mm_ss`, so the conditional compilation statement was removed. Not for sure, please confirm

ready

### DIFF
--- a/src/debug/Log.hpp
+++ b/src/debug/Log.hpp
@@ -48,12 +48,9 @@ namespace Debug {
 
         // print date and time to the ofs
         if (disableTime && !**disableTime) {
-#ifndef _LIBCPP_VERSION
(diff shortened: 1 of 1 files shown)


vaxerski:
@jbeich iirc was the one to put it in due to bsd issues.

rtgiskard:
> @jbeich iirc was the one to put it in due to bsd issues.

It's introduced from here: https://github.com/hyprwm/Hyprland/pull/3186#issuecomment-1707434227

While with clang++ 17 (current archlinux default), `std::chrono::zoned_time` is still not supported ([c++20 status](https://en.cppreference.com/w/cpp/compiler_support/20)), add a TODO there

vaxerski (approved):
thanks!

PR #6338, opened 2024-06-05
feat: cursor warps maintain relative position within a window

memchr:
Enables hyprland to remember the position of the cursor relative to a window when the window is unfocused (unless caused by **continuous** cursor movement). later, when the window is refocused, the cursor will be warped back to its previously remembered position within that window, rather than to the centre of the window.

in addition, when moving a focused window, the relative cursor position would also be maintained instead of jumping to the centre. this includes
- swap windows
- moving windows
- adding or removing windows from groups
- moving windows to different workspaces 

as a fallback, the cursor is still moved to the centre of the window when 
- the cursor is outside the window when it is unfocused
- the cursor would be distorted outside the window if the relative position is remembered.

this behaviour is controlled by `cursor:persistent_warps`.

## use case

improves usability for workflows where it is beneficial to maintain the cursor position within a window context, such as when frequently switching between multiple windows, e.g. between a browser and an ide.

this should be more noticeable to laptop users with trackpads, as these devices are relatively close to the keyboard and tend to be imprecise and require a lot of finger movement.

### terms

- *warp*: a **discontinuous** cursor movement.
- *discontinuous unfocus*: losing focus of a window that is **not** caused by moving the mouse continuously. currently it is usually caused by the movefocus dispatcher.
- *relative cursor position*: the coordination of the cursor relative to the upper left corner of a window.

### cursor position tracking

`vector2d cwindow::m_vrelativecursorcoordsonlastwarp` stores the cursor position relative to the window. it was updated whenever the window was about to lose focus discontinuously.

### cursor warpping

`void cwindow::warpcursor()` handles the logic for warping the cursor.
- if presistent_warps is disabled or the saved position is invalid, it behaves identically to `g_pcompositor->warpcursorto(window->middle())`, i.e. warp the cursor to the middle of the window;
- otherwise it warps the cursor to the previously stored relative position.

### integration with keybindings

- `updaterelativecursorcoords` will update the relative cursor position before the last focused window is about to discontinuously lose focus. 
- several dispatchers are updated to call `updaterelativecursorcoords` before change of window position and to use the `cwindow::warpcursor()` method instead of warping the cursor to the centre of the window directly.

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1356,7 +1356,7 @@ void CWindow::activate(bool force) {
         g_pCompositor->changeWindowZOrder(m_pSelf.lock(), true);
 
     g_pCompositor->focusWindow(m_pSelf.lock());
-    g_pCompositor->warpCursorTo(middle());
+    warpCursor();
 }
 
 void CWindow::onUpdateState() {
@@ -1529,4 +1529,17 @@ void CWindow::onX11Configure(CBox box) {
         g_pInputManager->refocus();
 
     g_pHyprRenderer->damageWindow(m_pSelf.lock());
+}
+
+void CWindow::warpCursor() {
+    const auto coords = m_vRelativeCursorCoordsOnLastWarp;
+
+    // effectivly reset m_vRelativeCursorCoordsOnLastWarp
+    m_vRelativeCursorCoordsOnLastWarp.x = -1;
+
+    // we don't wanna warp cursor outside the window
+    if (true && coords.x >= 0 && coords.y >= 0 && coords < m_vSize) {
+        g_pCompositor->warpCursorTo(m_vPosition + coords);
+    } else
+        g_pCompositor->warpCursorTo(middle());
 }
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1532,13 +1532,11 @@ void CWindow::onX11Configure(CBox box) {
 }
 
 void CWindow::warpCursor() {
-    const auto coords = m_vRelativeCursorCoordsOnLastWarp;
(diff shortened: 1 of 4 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -257,6 +257,16 @@ bool CKeybindManager::ensureMouseBindState() {
     return false;
 }
 
+void updateRelativeCursorCoords() {
+    static auto PNOWARPS = CConfigValue<Hyprlang::INT>("cursor:no_warps");
+
+    if (*PNOWARPS)
+        return;
+
+    if (const auto PLASTWINDOW = g_pCompositor->m_pLastWindow.lock(); PLASTWINDOW)
```
vaxerski, line 266: no need to lock here, you can just `if(g_pCompositor->m_pLastWindow)`

src/desktop/Window.cpp
```diff
@@ -1530,3 +1530,14 @@ void CWindow::onX11Configure(CBox box) {
 
     g_pHyprRenderer->damageWindow(m_pSelf.lock());
 }
+
+void CWindow::warpCursor() {
+    static auto PERSISTENTWARPS         = CConfigValue<Hyprlang::INT>("cursor:persistent_warps");
+    const auto  coords                  = m_vRelativeCursorCoordsOnLastWarp;
+    m_vRelativeCursorCoordsOnLastWarp.x = -1; // reset m_vRelativeCursorCoordsOnLastWarp
+
+    if (*PERSISTENTWARPS && coords.x > 0 && coords.y > 0 && coords < m_vSize) { // don't warp cursor outside the window
```
vaxerski, line 1539: no {}

PR #6342, opened 2024-06-06
Add gestures:workspace_swipe_min_fingers option

JohnMH:
When gestures:workspace_swipe_min_fingers is enabled, gestures:workspace_swipe_fingers is considered to be the minimum number of fingers required to swipe.

This behavior is more similar to the sway and macOS.

For example, this allows you to set workspace_swipe_fingers to 3, but swipe with 4 or more fingers instead of 3.

To avoid changing behavior for existing users, this setting is disabled by default.


vaxerski (dismissed):
lgtm, wiki mr needed

JohnMH:
See https://github.com/hyprwm/hyprland-wiki/pull/676

vaxerski:
rebase on main and run clang-format pls

vaxerski (approved):

vaxerski:
thanks!

PR #6344, opened 2024-06-06
also test cross build

Mic92:

We noticed that a recent update broken cross compilation in nixpkgs and therefore added tests for this upstream - I saw that this project uses the magic-nix-cache so hopefully all cross-build dependencies will be cached.

CMake Error at /nix/store/q1nssraba326p2kp6627hldd2bhg254c-cmake-3.29.2/share/cmake-3.29/Modules/FindPkgConfig.cmake:634 (message):
  The following required packages were not found:

   - hwdata

Adding `hwdata` and `hyprwayland-scanner` seems to make the pkg-config happy, but than it looks like it's pulling `hyprwayland-scanner` for the wrong architecture. I saw that your meson build seems to be somewhat cross-compiling aware, so I am not quite sure what is going on here.

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         package:
           - hyprland
+          - hyprland-cross
           - xdg-desktop-portal-hyprland
 
     runs-on: ubuntu-latest
--- a/flake.lock
+++ b/flake.lock
@@ -154,11 +154,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1726755586,
-        "narHash": "sha256-PmUr/2GQGvFTIJ6/Tvsins7Q43KTMvMFhvG6oaYK+Wk=",
+        "lastModified": 1727122398,
+        "narHash": "sha256-o8VBeCWHBxGd4kVMceIayf5GApqTavJbTa44Xcg5Rrk=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "c04d5652cfa9742b1d519688f65d1bbccea9eb7e",
+        "rev": "30439d93eb8b19861ccbe3e581abf97bdc91b093",
         "type": "github"
       },
       "original": {
--- a/flake.nix
+++ b/flake.nix
@@ -67,6 +67,15 @@
           hyprland-extras
         ];
       });
+    pkgsCrossFor = eachSystem (system: crossSystem:
+      import nixpkgs {
+        localSystem = system;
+        crossSystem = crossSystem;
+        overlays = with self.overlays; [
+          hyprland-packages
+          hyprland-extras
+        ];
+      });
   in {
     overlays = import ./nix/overlays.nix {inherit self lib inputs;};
 
@@ -92,6 +101,7 @@
         
         xdg-desktop-portal-hyprland
         ;
+      hyprland-cross = (pkgsCrossFor.${system} "aarch64-linux").hyprland;
     });
 
     devShells = eachSystem (system: {
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -102,7 +102,6 @@ in
         ninja
         pkg-config
         python3 # for udis86
-        wayland-scanner
       ];
 
       outputs = [
@@ -130,6 +129,7 @@ in
           tomlplusplus
           wayland
           wayland-protocols
+          wayland-scanner
           xorg.libXcursor
         ]
         (optionals customStdenv.hostPlatform.isMusl [libexecinfo])


Mic92:
cc @fufexan

fufexan:
Great addition! We should also add the cross build to the [CI matrix](https://github.com/hyprwm/Hyprland/blob/main/.github/workflows/nix-build.yml#L13).

Mic92:
> Great addition! We should also add the cross build to the [CI matrix](https://github.com/hyprwm/Hyprland/blob/main/.github/workflows/nix-build.yml#L13).

Maybe, but first I would appreciate help with the issue. Thanks!

fufexan:
Great addition! We should also add the cross build to the [CI matrix](https://github.com/hyprwm/Hyprland/blob/main/.github/workflows/nix-build.yml#L13).

> Adding `hwdata` and `hyprwayland-scanner` seems to make the pkg-config happy, but than it looks like it's pulling `hyprwayland-scanner` for the wrong architecture.

Sorry, adding those packages where? I've added hwdata and basically mimicked the `(native)BuildInputs` from the Nixpkgs derivation, yet CMake still can't find `hwdata`.

fufexan:
Managed to fix the `hwdata` error, and I've updated `hyprwayland-scanner` to work in aarch64 as well.

However, `wayland-scanner` now fails to generate proper headers.

Mic92:
This is the same issue I also ended up having:

`/nix/store/xq1l3m0fsrb6ql8qbii4fg9cbkwz5gvi-wayland-aarch64-unknown-linux-gnu-1.22.0-bin/bin/wayland-scanner` shouldn't be used because its build for aarch64, we want the native version here. meson seems to have some `native: true` flag, which indicates that you could request this feature, but I don't know how it works out here.

fufexan:
Looks like building with Meson worked, but I can't get CMake to select the right wayland-scanner.

Mic92:
Do we need both to build hyprland? I wouldn't mind if cross-compiling would work with one of the two as a good first step.

fufexan:
Hyprland can build fine in Meson it seems, so assuming we can override *only the pkgsCross derivation*, we can introduce Meson before CMake (so its configure hook takes priority) in `nativeBuildInputs.

Mic92:
> Hyprland can build fine in Meson it seems, so assuming we can override _only the pkgsCross derivation_, we can introduce Meson before CMake (so its configure hook takes priority) in `nativeBuildInputs`.

Can we not also just disable cmake everywhere like this?

```
dontUseCmakeConfigure = true;
```

fufexan:
We can, I was just not aware of that option.

fufexan:
I've removed hwdata as it is no longer needed in Hyprland itself, but in aquamarine. So I fixed aquamarine's cross build.
However, the new problem is that CMake doesn't seem to pick up the correct `wayland-scanner`. It should pick the host architecture, but instead it picks the target system architecture, which leads to errors like this:
```
hyprland-aarch64-unknown-linux-gnu> FAILED: /build/source/protocols/text-input-unstable-v1-protocol.c
hyprland-aarch64-unknown-linux-gnu> cd /build/source && /nix/store/q0396wa1xyhcwbzv0m7vv5bssnq3diab-wayland-aarch64-unknown-linux-gnu-1.22.0-bin/bin/wayland-scanner private-code //nix/store/swrc0afg0ppby4xamsx37m8diswdf4si-wayland-protocols-aarch64-unknown-linux-gnu-1.35/share/wayland-protocols/unstable/text-input/text-input-unstable-v1.xml protocols/text-input-unstable-v1-protocol.c
hyprland-aarch64-unknown-linux-gnu> /nix/store/q0396wa1xyhcwbzv0m7vv5bssnq3diab-wayland-aarch64-unknown-linux-gnu-1.22.0-bin/bin/wayland-scanner: line 1: syntax error: unexpected word (expecting ")")
```

@Mic92 do you have a clue why this happens or how to fix it?

Mic92:
I believe cmake needs to pass to pkg-config to differentiate between host and build dependencies: https://www.freedesktop.org/wiki/Software/pkg-config/CrossCompileProposal/
It looks like meson has figured this out.

fufexan:
Sure, we can use meson for the cross build. However I wanted to make sure CMake can also cross-compile as it's the prevalent build system used throughout hyprwm.

vaxerski:
I still would want to drop meson

fufexan:
@Mic92 I've managed to fix this (along with the previously-failing aquamarine). We're currently building Hyprland with Meson, but when testing, the cross compilation also worked with CMake. This is relevant as we may (once again) [go back to using CMake](https://github.com/hyprwm/Hyprland/pull/7784).

When building I've also been noticing that `hyprcursor` couldn't cross-compile due to `xcur2png` failing due to malloc issues. I've found a way to fix that, but I'm unsure whether that's a suitable fix for nixpkgs inclusion.

fufexan:
Got the patch merged in Nixpkgs https://github.com/NixOS/nixpkgs/pull/343074. Will wait a few days for it to propagate to unstable, and then merge this PR.

fufexan (approved):
LGTM. Thank you @Mic92

Mic92:
@fufexan thank you as well!

PR #6356, opened 2024-06-07
HookSystemManager: Make needsDeadCleanup volatile

JohnMH:
The value of needsDeadCleanup would be clobbered after longjmp, having an undefined value.

By making this variable volatile, it will always be read/written in memory instead of being stored in a register.


vaxerski (approved):
thanks!

PR #6358, opened 2024-06-07
Add description do key binds

### LINKED ISSUE #6357: Bind description
### Description

I'm bad at remembering key binds and was always a fan of nvim plugins like which key. For something like this, a way of describing your key binds is necessary.

My proposal would look something like this:
```conf
bind = SUPER, T, "open my terminal", exec, kitty
```
but obviously, it should be backward compatibl so this should also be no problem:
```conf
bind = SUPER, T, exec, kitty
```

Working with any bind the description would always come after the KEY (so at index 2) and the indicator would be double quotes. I think some kind of indicator is needed to distinguish it from the dispatcher options.

PR is a WIP

Moerliy:
I'm bad at remembering key binds and was always a fan of nvim plugins like which key. For something like this, a way of describing your key binds is necessary.

My proposal would look something like this:
```conf
bind = SUPER, T, "open my terminal", exec, kitty
```
but obviously, it should be backward compatible so this should also be no problem:
```conf
bind = SUPER, T, exec, kitty
```

Feedback is wanted but it's merge ready from my end.

closes #6357

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1991,11 +1991,17 @@ std::optional<std::string> CConfigManager::handleBind(const std::string& command
     if (mouse && (repeat || release || locked))
         return "flag m is exclusive";
 
-    const auto ARGS = CVarList(value, 4);
+    auto ARGS = CVarList(value, 4);
 
+    const bool HAS_DESCRIPTION = ARGS[2].starts_with("\"") && ARGS[2].ends_with("\"");
+
+    if (HAS_DESCRIPTION) {
+        ARGS = CVarList(value, 5);
+    }
+    const int DESCR_OFFSET = HAS_DESCRIPTION ? 1 : 0;
     if ((ARGS.size() < 3 && !mouse) || (ARGS.size() < 3 && mouse))
         return "bind: too few args";
-    else if ((ARGS.size() > 4 && !mouse) || (ARGS.size() > 3 && mouse))
+    else if ((ARGS.size() > 4 + DESCR_OFFSET && !mouse) || (ARGS.size() > 3 + DESCR_OFFSET && mouse))
         return "bind: too many args";
 
     std::set<xkb_keysym_t> KEYSYMS;
@@ -2014,12 +2020,14 @@ std::optional<std::string> CConfigManager::handleBind(const std::string& command
 
     const auto KEY = multiKey ? "" : ARGS[1];
 
-    auto       HANDLER = ARGS[2];
+    const auto DESCRIPTION = HAS_DESCRIPTION ? ARGS[2] : "";
 
-    const auto COMMAND = mouse ? HANDLER : ARGS[3];
+    auto       HANDLER = ARGS[2 + DESCR_OFFSET];
 
-    if (mouse)
-        HANDLER = "mouse";
+    const auto COMMAND = mouse ? HANDLER : ARGS[3 + DESCR_OFFSET];
+
+    // const auto HAS_DESCRIPTION = mouse ? ARGS.size() == 4 : ARGS.size() == 5;
+    //
 
     // to lower
     std::transform(HANDLER.begin(), HANDLER.end(), HANDLER.begin(), ::tolower);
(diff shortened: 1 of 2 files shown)


vaxerski:
why not add a flag for description like `bindd`?

Also, this will break with descriptions with `,`

Moerliy:
> why not add a flag for description like `bindd`?

Good idea, to get ridge of  " inside the key bind itself

> Also, this will break with descriptions with `,`

Yeah. had that in the back of my mind too. First reaction was to just use the docs to let the users know they shouldn't use the delimiter in the description.
I couldn't come up with a solution. It's the delimiter.
As far as I'm aware you can only get away with using the delimiter in the args part of the bind.
@vaxerski

vaxerski:
yes, because we limit the amount of arguments to 4, so anything after the 3rd `,` is treated as one. With this, you can't do it. I'm fine with mentioning that `,` is not allowed in descriptions, but a flag would be nice.

Also, no need to ping, I have notifications on, you know.

Moerliy:
I've added hyprctl functionality too now but I have a question on how you want the description to be handled.

1. It can either be always in the printout and if there is no description it will have an empty value. Similar to args.
2. Or we use the description flag to find out if there should be a description Key-Value pair or not.

Pro of the lather: Would be more in line with how it should be in my opinion, especially with the JSON out.
Con:  more complex code where it isn't needed and people don't care if it's an empty string or no Key-Value pair.

The current implementation is the first.

vaxerski:
for json you can omit the k-v if it's not set.

Moerliy:
> for json you can omit the k-v if it's not set, but it doesn't matter.

then it's fine with me :+1: 

I did the docs real quick too so I would say this PR and the docs PR are ready to review.
I can't assign a review tho or maybe I'm just too blind.

src/config/ConfigManager.cpp
```diff
@@ -1953,15 +1954,16 @@ std::optional<std::string> CConfigManager::handleBind(const std::string& command
...
-    const auto BINDARGS     = command.substr(4);
+    bool       locked          = false;
+    bool       release         = false;
+    bool       repeat          = false;
+    bool       mouse           = false;
+    bool       nonConsuming    = false;
+    bool       transparent     = false;
+    bool       ignoreMods      = false;
+    bool       multiKey        = false;
+    bool       has_description = false;
```
vaxerski, line 1965: no snake :skull:

hasDescription

src/debug/HyprCtl.cpp
```diff
@@ -789,9 +791,12 @@ std::string bindsRequest(eHyprCtlOutputFormat format, std::string request) {
                 ret += "e";
             if (kb.nonConsuming)
                 ret += "n";
+            if (kb.hasDescriptions) {
```
vaxerski, line 794: no {}

src/managers/KeybindManager.hpp
```diff
@@ -14,23 +14,25 @@ class CPluginSystem;
...
+    std::string            submap          = "";
+    std::string            description     = "";
+    bool                   release         = false;
+    bool                   repeat          = false;
+    bool                   mouse           = false;
+    bool                   nonConsuming    = false;
+    bool                   transparent     = false;
+    bool                   ignoreMods      = false;
+    bool                   multiKey        = false;
+    bool                   hasDescriptions = false;
```
vaxerski, line 35: hasDescription (no plural)

src/config/ConfigManager.cpp
Moerliy: python muscle mem, it seems.
fixed in commit 05a1e22ed93afefb84b19bec05ed66d3de2e568b

src/debug/HyprCtl.cpp
Moerliy: fixed in commit 05a1e22ed93afefb84b19bec05ed66d3de2e568b

src/managers/KeybindManager.hpp
Moerliy: fixed in commit 05a1e22ed93afefb84b19bec05ed66d3de2e568b

Moerliy:
Alright, let's try this again.

src/config/ConfigManager.cpp
```diff
@@ -2014,12 +2020,14 @@ std::optional<std::string> CConfigManager::handleBind(const std::string& command
...
+    const auto DESCRIPTION = hasDescription ? ARGS[2] : "";
 
-    const auto COMMAND = mouse ? HANDLER : ARGS[3];
+    auto       HANDLER = ARGS[2 + DESCR_OFFSET];
 
-    if (mouse)
-        HANDLER = "mouse";
+    const auto COMMAND = mouse ? HANDLER : ARGS[3 + DESCR_OFFSET];
+
+    // const auto HAS_DESCRIPTION = mouse ? ARGS.size() == 4 : ARGS.size() == 5;
```
vaxerski, line 2029: this comment is useless

vaxerski (commented):
rest lgtm

src/config/ConfigManager.cpp
Yusuf-Duran: is removed :+1:

vaxerski (approved):
thanks!

gksudolol:
I get an invalid dispatcher error when firing up hyprland from this commit with:

`bindm = $mainMod, mouse:273, resizewindow`

specifically with `resizewindow`, any other dispatcher works fine (this doesn't occur with the commit before)

Moerliy:
I'm not sure here, but there really is no dispatcher called `resizewindow` but one `resizewindowpixel`.
Does the second one do what you are after?

MightyPlaza:
> I'm not sure here, but there really is no dispatcher called `resizewindow` but one `resizewindowpixel`. Does the second one do what you are after?

mouse dispatcher

nonetrix:
Yep same

Moerliy:
Oh yeah, my bad. Found it. I did a push to my fork and I can also provide a patch if you want.

PR #6380, opened 2024-06-09
hyprctl: add --quiet flag

diniamo:

Adds a `--quiet` flag to hyprctl that supresses its output.

Might want to use `setstate(failbit)` instead of `rdbuf(nullptr)`, I'm not sure.

It's ready.


hyprctl/main.cpp
```diff
@@ -341,6 +344,10 @@ int main(int argc, char** argv) {
         fullRequest += ARGS[i] + " ";
     }
 
+    if (quiet) {
+        std::cout.rdbuf(nullptr);
```
vaxerski, line 348: I'd rather wrap std::cout in something like `log` that does `if (quiet) return;`

diniamo:
How is this?

Note:
- I left normal cout for usage prints, should those also be included?
- Removed the endl print at the end, since I highly doubt it's necessary, and is also annoying

hyprctl/main.cpp
```diff
@@ -30,6 +30,7 @@
 #define PAD
 
 std::string instanceSignature;
+bool        quiet;
```
vaxerski, line 33: please initialize primitives

hyprctl/main.cpp
```diff
@@ -39,6 +40,11 @@ struct SInstanceData {
     bool        valid = true;
 };
 
+void log(std::string str) {
+    if (!quiet)
```
vaxerski, line 44: if (quiet)
   return;

hyprctl/main.cpp
```diff
@@ -419,6 +427,5 @@ int main(int argc, char** argv) {
         exitStatus = request(fullRequest);
     }
 
-    std::cout << std::endl;
```
vaxerski, line 422: this is useful as it flushes stdout, though you can also just flush

hyprctl/main.cpp
diniamo: Why?

hyprctl/main.cpp
diniamo: Why is flushing useful?

hyprctl/main.cpp
vaxerski: always prefer guards, they are cleaner.

hyprctl/main.cpp
vaxerski: to avoid half-written logs

diniamo:
That should be all.

vaxerski (approved):
lgtm thanks!

PR #6385, opened 2024-06-09
core: move to hyprutils for utils

vaxerski:
This MR moves a bunch of utilities over to hyprutils.

To be merged after 0.41 lands, and @fufexan pls for nix


vaxerski:
@fufexan looks like you forgot to add the dep to meson's hyprpm/hyprctl

PR #6399, opened 2024-06-10
github issues: improve the chances of the user specifying bug or regression

Aqa-Ib:
I believe this would do the trick, because currently when you expand the system info box; the bug/regression box is left in a weird spot between the huge system info box and the description box, making it harder to pay attention to it.

Cheers


vaxerski (approved):
sure, thanks!

PR #6414, opened 2024-06-11
sessionLock: fix focus not being tied to mouse

outfoxxed:
Fixes session lock focus not following the mouse between monitors. Previously focus would stay forever on a single surface.

no

ready to merge


wxlyyy:
Can this be solved?
https://github.com/jovanlanik/gtklock/issues/86

outfoxxed:
gtklock likely only accepts input on one surface, in which case yes, though you can also solve it by forcing gtk to accept inputs from the seat even though it doesn't think it's focused. In my own quickshell lock it now follows the mouse correctly.

vaxerski (approved):
thanks!

PR #6424, opened 2024-06-11
layer: don't close special ws when restoring focus

MightyPlaza:
same as #5308 but works for special workspaces
seems to work fine


vaxerski (approved):
tanks!

vaxerski:
konflikt

MightyPlaza:
can't really build main after new dependency was added yet to test, but should be fine since I just rebased and changed the function in use.

vaxerski (approved):
tanks!

Issue #6378, opened 2024-06-09
Hyprland crashes when restarting fcitx5 (Random)


hdm9527:
### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 9994b73ad0af5c9ba5fc4058234cea77d0a1ddb1 dirty (buffer: track asynchronous buffers and don't release them until unref).
Date: Sat Jun 8 23:27:56 2024
Tag: v0.40.0-176-g9994b73a, commits: 4782

flags: (if any)


System Information:
System name: Linux
Node name: Arch
Release: 6.9.3-2-cachyos-echo-lto
Version: #1 SMP PREEMPT_DYNAMIC Thu, 06 Jun 2024 06:05:56 +0000


GPU information: 
64:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Phoenix1 [1002:15bf] (rev c7) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

Hyprland crashes when restarting fcitx. It's crashed once so far and I haven't found a way to reproduce it yet.

### How to reproduce

above

### Crash reports, logs, images, videos

[hyprlandCrashReport1055.txt](https://github.com/user-attachments/files/15750549/hyprlandCrashReport1055.txt)
[coredumpctl.log](https://github.com/user-attachments/files/15750550/coredumpctl.log)

hdm9527:
[hyprlandCrashReport67619.txt](https://github.com/user-attachments/files/15750648/hyprlandCrashReport67619.txt)
[hyprland_coredumpctl.log](https://github.com/user-attachments/files/15750650/hyprland_coredumpctl.log)

cunbidun:
I also encountered this problem a number of times before

sungyoonc:
[...]

PR #6460, opened 2024-06-12
check if VirtualKeyboard is valid before accessing.

yangyingchao:
This fixes a crash when restarting fcitx (#6378).

An extra newline is added by the editor automatically, as mentioned in https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline.


vaxerski (approved):
thanks!

PR #6470, opened 2024-06-13
last bits of asan output

gulafaran:
with these commits and the [hyprutils PR](https://github.com/hyprwm/hyprutils/pull/1) im down to 

` SUMMARY: AddressSanitizer: 880 byte(s) leaked in 8 allocation(s).` of a nested run of opening a few windows/closing and exiting hyprland. making asan VERY silent :) not that these wouldnt have been harvested by the OS anyways but makes asan much more tolerable when trying to debug things. could use a test or two so the eventloopmanager doesnt have to be running at exit? and freeing the idle frame means some artifacting on some weird combination of display destroying.

the remaining leaks is gonna be solved when the 3 old protocols gets rewritten to the new way, and when the wlroots/wayland implentation proceeds.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -352,6 +352,7 @@ void CCompositor::cleanup() {
     g_pPointerManager.reset();
     g_pSeatManager.reset();
     g_pHyprCtl.reset();
+    g_pEventLoopManager.reset();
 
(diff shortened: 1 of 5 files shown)


vaxerski (approved):
thanks!

PR #6479, opened 2024-06-13
layout: refine master layout new window handling

memchr:
## Open window relative to active window

`new_on_active`:
  - `none` (default):
  - `before`: above of the focused window
  - `after`: below the focused window

If the focused window is the solo master window, or the new window replaces master, this option has no effect and new_on_top are respected.

## Refine new window status control

**BREAKING CHANGE**: new_is_master removed in favour of new variable

`new_status`:
  - `slave` (default): new window open as slave
  - `master`: new window open as master
  - `inherit`: new window inherit status from active window, i.e. when the focused window is master, new window will become new master, otherwise new window are added to slaves


vaxerski (approved):
sure, wiki mr needed

Ghosthree3:
Can this be updated, https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.conf#L128
Default example config is now broken.

PR #6520, opened 2024-06-15
github: reword bug or regression

Aqa-Ib:
I have been watching and a lot of people still don't mark a clear regression. Rewording would fix this, I believe.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -5,7 +5,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        **Before opening a new issue, take a moment to search through the current open and closed issues to check if it already exists.**
+        <h2>Before opening a new issue, please take a moment to search through the current open and closed issues to check if it already exists.</h2>
 
         ---
 
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -13,7 +13,7 @@ body:
     id: type
     attributes:
       label: Regression?
-      description: Is this bug present in a previous version of Hyprland?
+      description: Is this bug absent in a previous version of Hyprland?
       options:
         - I don't know
         - Yes
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -5,7 +5,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        <h2>Before opening a new issue, please take a moment to search through the current open and closed issues to check if it already exists.</h2>
+        ## Before opening a new issue, please take a moment to search through the current open and closed issues to check if it already exists.
 
         ---
 
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -13,7 +13,7 @@ body:
     id: type
     attributes:
       label: Regression?
-      description: Is this bug absent in a previous version of Hyprland?
+      description: Regression means that this bug is absent in a previous version of Hyprland. If it is, in order to fix it we need you to bisect it with the help of the wiki: https://wiki.hyprland.org/Crashes-and-Bugs/#bisecting-an-issue 
       options:
         - I don't know
         - "Yes"
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -5,7 +5,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Before opening a new issue, take a moment to search through the current open and closed issues, to check if it already exists.
+        Before opening a new issue, take a moment to search through the current open and closed issues to check if it already exists.
 
         ---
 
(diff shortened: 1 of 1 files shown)


MightyPlaza:
might also be a good idea to ask them to bisect and provide the wiki link https://wiki.hyprland.org/Crashes-and-Bugs/#bisecting-an-issue

Aqa-Ib:
Let me know if it can be worded better, (english is not my mother tongue).

fufexan (dismissed):
LGTM

fufexan:
> Also, do you think that we should give "I don't know" as an option?

Probably not, as we would still tell them to bisect. It's either yes/no.

Also, we should tell users to also search closed issues, not only open ones.

MahouShoujoMivutilde:
Maybe something like

* Bug - This is an unexpected behavior (something is broken).
* Regression - Like bug, but you also know that it used to work as expected and now doesn't (provide bisect).

?

Because "absent in a previous version" is IMO not as clear - you could read it and be like "of course it was absent, this feature didn't exist".

Aqa-Ib:
I understand your concern about a new feature. However, I don't see your definition of regression clear enough.

fufexan:
@Aqa-Ib what about "Regression - something no longer works in newer versions"?

Aqa-Ib:
Done. One thing that is not clear: if we force the users to check for a regression, one can wonder: how many version do I have to roll back to check for it ?

MightyPlaza:
> Done. One thing that is not clear: if we force the users to check for a regression, one can wonder: "how many versions do I have to roll back to check for it?" If we are omitting the "I don't know" option, maybe we should clarify this.

i don't think this matters much
the point is that they write that it's a regression if they know it was working before and now isn't
not go looking for a working commit (this is also a good idea when debugging issues, but not necessary imo)

Aqa-Ib:
If that is the case, then it makes sense for the "I don't know" option to be included.

fufexan:
> Done. One thing that is not clear: if we force the users to check for a regression, one can wonder: "how many versions do I have to roll back to check for it?" If we are omitting the "I don't know" option, in my opinion we should clarify this. Maybe 1 or 2 major releases ?

If the user just updated, they should know which version they were on before, and start with that. If they don't, then 1-2 major versions sounds alright.

Aqa-Ib:
@vaxerski please, as you are the one that deal with every bug report, when you have a moment do you mind sharing your thoughts on this?

Aqa-Ib:
@fufexan @MightyPlaza @Agent00Ming would you make a change to improve bug reports ?

Agent00Ming:
Maybe the command paste thing is a bit obscure but I don't know if we can wrap the user input in a plain text box and treat it with markdown afterwards. Otherwise, it looks as good as it's going to get considering the difficulty that some experience reading a sentence longer than 5 words.

fufexan:
> @fufexan @MightyPlaza @Agent00Ming would you make a change to improve bug reports ?

Unsure what we could improve further here.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,18 +5,20 @@ body:
...
 
   - type: dropdown
     id: type
     attributes:
-      label: Bug or Regression?
-      description: Is this a bug or a regression?
+      label: Regression?
+      description: |
+        Regression means that something used to work but no longer does.
+        PLEASE, before continuing check if this bug is a regression or not, and if it is, we need you to bisect with the help of the wiki: https://wiki.hyprland.org/Crashes-and-Bugs/#bisecting-an-issue
```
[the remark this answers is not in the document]
Aqa-Ib: Done

Agent00Ming (approved):

Aqa-Ib:
I found a way to force the user to select yes or no.
At the moment I am exploring simplifying the command paste textbox.

MightyPlaza (approved):

Aqa-Ib:
I wasn't able to simplify the command paste textbox.

Aqa-Ib:
This is ready.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,18 +5,21 @@ body:
...
   - type: dropdown
     id: type
     attributes:
-      label: Bug or Regression?
-      description: Is this a bug or a regression?
+      label: Regression?
+      description: |
+        Regression means that something used to work but no longer does.
+        **BEFORE CONTINUING**, please check if this bug is a regression or not, and if it is, we need you to bisect with the help of the wiki: https://wiki.hyprland.org/Crashes-and-Bugs/#bisecting-an-issue
+      multiple: true
```
fufexan, line 19: I don't think users should be able to select both Yes and No.

Aqa-Ib: That option achieves that there is no option selected by default, so it forces the user to select one option to be able to submit the bug report. That outweighs the possibility of one mad lad deciding to select both IMO.

fufexan: Then why not set `default: 0` instead? This will force the user to choose one of the provided options, according to the [GitHub docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#validations-for-dropdown).

fufexan: What's the problem with selecting "yes" as a default? It can still be changed.

Aqa-Ib: The problem is that people have attention problems and just ignore it. If we don't have a default, people can't send the bug report unless they explicitly select yes or no, so you can expect that the value of regression is hand picked, and not just "yes" because the user ignored it.

fufexan: Alright then, makes sense.

fufexan (approved):
LGTM, thanks!

Aqa-Ib:
Thank you as always. Wish you all a wonderful day!

PR #6521, opened 2024-06-15
xdg-shell: Continue transform of popup until size fits

DashieTM:
- applies all 3 steps of the protocol until it fits or doesn't find a suitable position
This is before this change:

https://github.com/hyprwm/Hyprland/assets/72016555/d4a4efb2-da65-48fd-b4d6-1efff0859be4

There is an issue with opening the popup at the exact point where it should flip upwards -> it pops up, and then flips (this also exists in main)

Should be ready other than what's above

### DIFF
--- a/src/protocols/XDGShell.cpp
+++ b/src/protocols/XDGShell.cpp
@@ -534,29 +534,25 @@ CBox CXDGPositionerRules::getPosition(const CBox& constraint, const Vector2D& pa
     if (success)
         return predictedBox.translate(-parentCoord - constraint.pos());
 
+    CBox test = predictedBox;
+
     if (state.constraintAdjustment & (XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_X | XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_Y)) {
         // attempt to flip
         const bool flipX = state.constraintAdjustment & XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_X;
         const bool flipY = state.constraintAdjustment & XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_Y;
 
-        CBox       test = predictedBox;
-        success         = true;
-        if (flipX && test.copy().translate(Vector2D{-predictedBox.w - state.anchorRect.w, 0}).expand(-1).inside(constraint))
-            test.translate(Vector2D{-predictedBox.w - state.anchorRect.w, 0});
-        else if (flipY && test.copy().translate(Vector2D{0, -predictedBox.h - state.anchorRect.h}).expand(-1).inside(constraint))
-            test.translate(Vector2D{0, -predictedBox.h - state.anchorRect.h});
-        else if (flipX && flipY && test.copy().translate(Vector2D{-predictedBox.w - state.anchorRect.w, -predictedBox.h - state.anchorRect.h}).expand(-1).inside(constraint))
+        if (flipX && flipY)
             test.translate(Vector2D{-predictedBox.w - state.anchorRect.w, -predictedBox.h - state.anchorRect.h});
-        else
-            success = false;
+        else if (flipX)
+            test.translate(Vector2D{-predictedBox.w - state.anchorRect.w, 0});
+        else if (flipY)
+            test.translate(Vector2D{0, -predictedBox.h - state.anchorRect.h});
+
+        success = test.copy().expand(-1).inside(constraint);
 
         if (success)
             return test.translate(-parentCoord - constraint.pos());
     }
-
-    // if flips fail, we will slide and remember.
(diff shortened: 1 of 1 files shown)


vaxerski:
image
excuse me?

DashieTM:
But why would they put slide as 1 and 2... ffs
aight, I'm gonna find a way to make the current order work and change the title and commit after.

DashieTM:
There we go, proper order. It just didn't apply the flip because it tested if the flip change by itself would fix the position. Hence it continued and eventually failed to position the popup.

src/protocols/XDGShell.cpp
```diff
@@ -534,30 +534,26 @@ CBox CXDGPositionerRules::getPosition(const CBox& constraint, const Vector2D& pa
...
+
     if (state.constraintAdjustment & (XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_X | XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_Y)) {
         // attempt to flip
         const bool flipX = state.constraintAdjustment & XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_X;
         const bool flipY = state.constraintAdjustment & XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_Y;
 
-        CBox       test = predictedBox;
-        success         = true;
-        if (flipX && test.copy().translate(Vector2D{-predictedBox.w - state.anchorRect.w, 0}).expand(-1).inside(constraint))
+        if (flipX && flipY)
```
vaxerski, line 544: no, no, no. What if the flip gets us _away_ from the desired positioning?

DashieTM: The only Idea I would have is to create a flipped copy and a regular copy, and then check if either of those provide a constraint that is ok.

But the translate check can't be within the if, otherwise the transform doesn't get applied and the slide won't do anything anymore. E.g. if a popup needs flip and transform it can't work with the translate within the if.

vaxerski: how about test how many sides are out of bounds and flip if it decreases the count?

DashieTM: Woops, sorry, you did the same with the other cases, so just check if horizontal out or vertical out.

src/protocols/XDGShell.cpp
```diff
@@ -534,30 +534,29 @@ CBox CXDGPositionerRules::getPosition(const CBox& constraint, const Vector2D& pa
...
         const bool flipX = state.constraintAdjustment & XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_X;
         const bool flipY = state.constraintAdjustment & XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_FLIP_Y;
 
-        CBox       test = predictedBox;
-        success         = true;
-        if (flipX && test.copy().translate(Vector2D{-predictedBox.w - state.anchorRect.w, 0}).expand(-1).inside(constraint))
+        const bool horizontalEdgeOut = test.x < constraint.x || test.x + test.w > constraint.x + constraint.w;
+        const bool verticalEdgeOut   = test.y < constraint.y || test.y + test.h > constraint.y + constraint.h;
+
+        if (flipX && horizontalEdgeOut && flipY && verticalEdgeOut)
```
vaxerski, line 547: this still doesn't test if flipping fixes this, confront this professional drawing:
image

DashieTM: But the flip would just invert the x direction. I just tested this to be sure.
In the picture it would mean the popup gets repositioned to the other side, which is not the case.

src/protocols/XDGShell.cpp
```diff
@@ -534,30 +534,36 @@ CBox CXDGPositionerRules::getPosition(const CBox& constraint, const Vector2D& pa
...
+            edgeCount += test.x < constraint.x ? 1 : 0;
+            edgeCount += test.x + test.w > constraint.x + constraint.w ? 1 : 0;
+            edgeCount += test.y < constraint.y ? 1 : 0;
+            edgeCount += test.y + test.h > constraint.y + constraint.h ? 1 : 0;
+            return edgeCount;
+        };
+        int edgeCount = countEdges(test);
+
+        // -1 as we want 2 edges to be resolved with the flip, otherwise move on to only x or only y
+        if (flipX && flipY && edgeCount - 1 > countEdges(test.copy().translate(Vector2D{-predictedBox.w - state.anchorRect.w, -predictedBox.h - state.anchorRect.h})))
```
vaxerski, line 554: this case makes no sense. You can drop this one and instead use the two ifs below without else

DashieTM: Ah yeah, thanks.

vaxerski (approved):
gangsta, thanks

PR #6523, opened 2024-06-15
Fix hardware cursor not showing on mirror displays

### LINKED ISSUE #6455: mouse cursor visibility on mirrored screen
### Description

I *love* the mirror feature in hyprland, where I can eg. use

```monitor=DP-1,preferred,auto,auto,mirror,eDP-1```

in order to mirror `eDP-1` to `DP-1`. 

What I find particularly great about this is that the resolutions on both displays can be different! I use this a lot to present slides to students.

It seems though, that the mouse cursor is hidden on the screen that is mirrored **to** (`DP-1` in the example) from the main screen (`eDP-1`). However, if I want to highlight something on the screen by hovering over or circling around something on my screen (`eDP-1`) the mouse cursor is not visible on the other (`DP-1`) and the students cannot see it.

It seems this behavior (ie. hidden mouse cursor on the mirrored-to-screen) is on purpose, but in the case described above, a visible cursor on both screens would be nice.

Would it be possible to add a way to make this happen? It would not have to be the default, but some way to toggle this on/off would be great!

Have a great day everyone!

### PR DESCRIPTION
**Pull request by lucasreis1**
Ensure cursor is rendered on mirror displays when using hardware cursor. (Fixes #6455 )

AFAIK this is a hack but works well on my system (using nvidia :)).

I guess it is, but again it's a hack so there could be tons of better solutions for this

### DIFF
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -360,30 +360,39 @@ void CPointerManager::resetCursorImage(bool apply) {
     }
 }
 
+void CPointerManager::updateCursorBackendForMonitor(CSharedPointer<CMonitor>& m, Hyprlang::INT PNOHW) {
+    auto state = stateFor(m);
+
+    if (!m->m_bEnabled || !m->dpmsStatus) {
+        Debug::log(TRACE, "Not updating hw cursors: disabled / dpms off display");
+        return;
+    }
+
+    if (state->softwareLocks > 0 || PNOHW || !attemptHardwareCursor(state)) {
+        Debug::log(TRACE, "Output {} rejected hardware cursors, falling back to sw", m->szName);
+        state->box            = getCursorBoxLogicalForMonitor(state->monitor.lock());
+        state->hardwareFailed = true;
+
+        if (state->hwApplied)
+            setHWCursorBuffer(state, nullptr);
+
+        state->hwApplied = false;
+        return;
+    }
+
+    state->hardwareFailed = false;
+}
+
 void CPointerManager::updateCursorBackend() {
     static auto PNOHW = CConfigValue<Hyprlang::INT>("cursor:no_hardware_cursors");
 
     for (auto& m : g_pCompositor->m_vMonitors) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/managers/PointerManager.cpp [hunk 1]
```diff
@@ -360,30 +360,50 @@ void CPointerManager::resetCursorImage(bool apply) {
...
-        }
+    state->hardwareFailed = false;
+}
+
+void CPointerManager::updateCursorBackend() {
+    static auto PNOHW = CConfigValue<Hyprlang::INT>("cursor:no_hardware_cursors");
+
+    for (auto& m : g_pCompositor->m_vMonitors) {
+        updateCursorBackendForMonitor(m, *PNOHW);
+        for (auto& mirror : m->mirrors)
```
vaxerski, line 391: {} around for()

vaxerski:
also maybe we should just lock software? this will break if the output mirroring is a different res

lucasreis1:
> this will break if the output mirroring is a different res
How? I tried to account for that here: https://github.com/hyprwm/Hyprland/pull/6523/commits/3364e760e7caf14a6e87c38dc16496b131c7ee1c#diff-649995ddb095587b539b1ea9cbe9bb01f749557ff5f34229f0688a276b957493R396

Tested on my system mirroring a 1440p screen to a 1080p monitor and it worked fine.

src/managers/PointerManager.cpp
lucasreis1: Oops, fixed!

vaxerski:
what about different _aspect ratios_? What about stretching? Size?

lucasreis1:
If you think that'd be solved by some transforms when handling pointer positions on `onCursorMoved` I can take a look, but if it's way more complicated than that I can make another PR to force software cursors when mirrors exist. But I'd be a shame

vaxerski:
I believe software cursors on mirrors are the way to go.

lucasreis1:
Sure, I'll convert this to a draft if you don't mind because I'd still like to tackle it later

PR #6537, opened 2024-06-16
update example hyprland.conf

memchr:
#6479 replace master:new_is_master with master:new_status

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -125,7 +125,7 @@ dwindle {
 
 # See https://wiki.hyprland.org/Configuring/Master-Layout/ for more
 master {
-    new_is_master = true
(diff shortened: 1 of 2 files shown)


fufexan (approved):

fufexan:
https://github.com/hyprwm/Hyprland/blob/main/src/config/defaultConfig.hpp#L141 should also be updated.

fufexan (approved):

noahfraiture:
What is the point except break config ?

TSmigielski:
I don't mind such changes, but a heads up in the changelog about a breaking config update would be nice.

onyasumi:
> What is the point except break config ?

not much, it seems

Issue #622, opened 2022-09-04
different layout for each monitor


beckend:
useful for rotated monitors, master is not useful for standing monitors

vaxerski:
the current system is very much not made with this in mind, but it probably might be an option in the future

proycon:
Just switched to hyprland (from dwm). It seems hyprland already does pretty much everything I need out of the box (good work!) except for this. Rather than tie it to a monitor, perhaps it could even be tied to a workspace? (allowing different layouts on different workspaces akin to dwm), since we can already bind workspaces to monitors with wsbind.

vaxerski:
read my comment.

jheroy:
Any progress on this?

vaxerski:
nope

Trinxic:
Guys, a form of this is in the wiki: https://wiki.hyprland.org/Configuring/Master-Layout/#config

`workspace = 2, layoutopt:orientation:top`

this will change the orientation of just workspace-2. From there, you could also bind the vertical monitor to that workspace:

`workspace = 2, monitor:[m], layoutopt:orientation:top`

tmccombs:
That is using different layout configuration for different workspaces, not different layouts.

gnusenpai:
@devraza You cannot change the layout (master/dwindle) based on workspace *or* monitor (what this issue is about). The example given above just changes the orientation of the master layout.

Truenya:
seems interesting to try

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #6544, opened 2024-06-16
Get current layout from workspace instead of g_pLayoutManager

UjinT34:
* added CWorkspace::getCurrentLayout
* replaced g_pLayoutManager->getCurrentLayout with CWorkspace::getCurrentLayout
* renamed CLayoutManager::getCurrentLayout to CLayoutManager::getCurrentGlobalLayout
* removed calls to getCurrentLayout from layouts

Initial work on per workspace layouts. For now `CWorkspace::getCurrentLayout` just returns `g_pLayoutManager->getCurrentGlobalLayout()`. So any existing logic shouldn't be affected by these changes.
Renamed `CLayoutManager::getCurrentLayout` to break the api and get conflicts with other PRs which may add new calls to `g_pLayoutManager->getCurrentLayout`.
Also removed calls to `getCurrentLayout` from layouts since it should point to `this`.

This code relies on `CWindow::m_pWorkspace` and `CMonitor::activeWorkspace` being set before the calls to `getCurrentLayout`.
In some cases `getCurrentLayout` might be called on an incorrect workspace. This should not be an issue for now since all workspaces will return the same layout.
When a window is moved to another workspace there are no checks that both workspaces have the same layout and only one of them gets updated. Again not an issue for now.

Ready for merging

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -875,7 +875,8 @@ void CCompositor::focusWindow(PHLWINDOW pWindow, SP<CWLSurfaceResource> pSurface
     if (pWindow && pWindow->m_bIsX11 && pWindow->m_iX11Type == 2 && !pWindow->m_pXWaylandSurface->wantsFocus())
         return;
 
-    g_pLayoutManager->getCurrentLayout()->bringWindowToTop(pWindow);
+    if (pWindow)
+        pWindow->m_pWorkspace->getCurrentLayout()->bringWindowToTop(pWindow);
 
     if (!pWindow || !validMapped(pWindow)) {
 
@@ -898,7 +899,8 @@ void CCompositor::focusWindow(PHLWINDOW pWindow, SP<CWLSurfaceResource> pSurface
 
         EMIT_HOOK_EVENT("activeWindow", (PHLWINDOW) nullptr);
 
-        g_pLayoutManager->getCurrentLayout()->onWindowFocusChange(nullptr);
+        if (m_pLastMonitor)
+            m_pLastMonitor->activeWorkspace->getCurrentLayout()->onWindowFocusChange(nullptr);
 
         m_pLastFocus.reset();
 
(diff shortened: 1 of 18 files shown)


vaxerski:
Makes no sense to split that though, as per-workspace layouts by the pure design of layouts cannot work properly. Am I missing something?

UjinT34:
Well, have to start with something. Too many changes to do it in one go. And it won't be easy to review it in one go...
I can continue to work on this as a single PR. I am just not sure if what I am doing is right. Is it the right way forward?

vaxerski:
my numba 1 question would be: why? It adds complexity for no reason, tbh. Currently Hyprland is designed so that each "layout" is like a standalone window manager - so it's easy to swap them out and also write new ones using plugins.

UjinT34:
This will implement https://github.com/hyprwm/Hyprland/issues/622
A plugin for this will have to duplicate the existing layouts and any new ones. I don't understand why a layout should care about what happens offscreen on other workspaces. Any layout work on active workspace doesn't depend on the state of other workspaces and doesn't affect the state of other workspaces. Am I missing some features which require simultaneous access to nodes across several workspaces? From a glance there are bunch of helper functions that limit the work to the given workspace id.
If you don't have any plans to change current design and support per-monitor or per-workspace layouts within hyprland itself than I won't pursue this any further.

zakk4223:
I've had a plugin that does this for a while now. It _does_ work and it's nice for people with multiple monitors that may have different sizes and/or orientations. Or just people who want different layouts in workspaces based on workflows/content etc. 

It works by adding a layout that just proxies layout calls to the "real" layout for a given workspace, and adding/removing windows from the 'real' layouts so a given window is never tracked by more than one layout.

I think if the only layouts possible in  hyprland were the two default ones, having native per-workspace layouts might seem a bit overkill. With plugins able to add as many layouts as users want, it might be nice to allow that extra flexibility.

vaxerski:
swapwindow between two workspaces will break this, for example.

zakk4223:
If my hacky proxy layout can deal with swapwindow across workspace/layout a native implementation of it can too.

vignesh1507 (commented):
it it necessary to update the code, i think the original code was perfect.

vaxerski:
closing due to inactivity

PR #6571, opened 2024-06-17
Customizable directional window resize logic

### LINKED ISSUE #6152: Bind or setting to precisely resize a window with a pointing device
### Description

As pointed out in [this currently unanswered discussion](https://github.com/hyprwm/Hyprland/discussions/6129), one can easily resize a window precisely with keyboard shortcuts, but, when grabbing on a window's sides, resizing with a pointing device like a mouse always acts on another axis that I do not want to be acted upon, as seen in this video.

https://private-user-images.githubusercontent.com/160250128/331737779-5101e38a-f604-4e71-ab31-d70597b09f59.mp4

I have looked at the wiki to see if I could change this behaviour, but to no avail.

---

This resizing behaviour is *generally* not the one I want nor expect, for it means that any imprecise movement with a pointing device like a mouse in an undesired axis will change the size of the tiles currently on the screen in an undesirable manner.

If possible, having a binding `[key + mouse]` would allow one to resize a window from its sides *(here in white)* in one axis, without affecting the others. This, ideally, could also or otherwise be done with a setting under `hyprland.conf`.

For example, on the image below, I could move the right side of `L1` horizontally, and even if I moved my pointing device vertically during that horizontal resize, only the horizontal axis would be affected.
Layout

---

Any thoughts on this?

Agent00Ming:
Closes #6152, adds a drop-in replacement (and configurable) window resizing logic. Also will fix a bug with pseudotiled windows + resize on border being unreliable (sometimes resizes the node box, sometimes resizes the inner window).

New function: `IHyprLayout::getWindowRegion()` determines where a point is on an arbitrary rectangle defined as a CBox with an optional CBox for edge size calculation (defaults to 0) and an optional `raw` boolean for the raw untranslated integer. Returns an integer between 0 and 23 (inclusive) when translated in-function.

Drop-in compatibility, no behavioural changes unless configured to be. Pseudotiled `general:resize_on_border` interaction now always resizes the inner window instead of the window node's box.

`general:edge_depth` float value that
 * acts as a % of the window's shortest side divided into two edges when between 0.0 and 1.0 (exclusive)
 * acts as a pixel count clamped between 1 and 1/2 of the window's shortest side when above 1.0 
 * disables the calculation of edges when 0 or under.
 
`general:resize_mouse_bind_pattern` integer value that sets the pattern of the window resize direction. 
 * 0 is the default, the 4 corners that we have now.
 * 1 makes the edges of the window single axis resize and the middle of the window is divided into 4 corners.
 * 2 is similar to the default with the exception that the edges are single axis resize
 * 3 corners do the dual axis resize and a 'cross' shape that connects the edges are single axis resize
 * 4 window is divided into 4 triangles for single axis resizes only
 * Any invalid value is considered default
 
It's presently in a working state for dwindle and for floating windows.

TODO:
- [x] floating windows don't update their reported size and position when resized (bug in main branch)
- [x] `general:extend_border_grab_area` no longer does anything (bug in main branch) (broken by addd3e7f1aeb670dd91d26005aaeccce3efb1ae7)
- [x] pseudotiled windows [dwindle] have a inconsistent `general:resize_on_border` interaction (bug in main branch)
- [x] pseudotiled windows [dwindle] don't respect single axis resize
- [ ] master layout does not respect single axis resize
- [ ] primary + alternate pattern ?
- [ ] settle on config variable names
- [ ] cleanup
- [ ] docs

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -701,7 +701,7 @@ PHLWINDOW CCompositor::vectorToWindowUnified(const Vector2D& pos, uint8_t proper
     if (properties & ALLOW_FLOATING) {
         for (auto& w : m_vWindows | std::views::reverse) {
             const auto BB  = w->getWindowBoxUnified(properties);
-            CBox       box = BB.copy().expand(w->m_iX11Type == 2 ? BORDER_GRAB_AREA : 0);
+            CBox       box = BB.copy().expand(w->m_iX11Type != 2 ? BORDER_GRAB_AREA : 0);
             if (w->m_bIsFloating && w->m_bIsMapped && !w->isHidden() && !w->m_bX11ShouldntFocus && w->m_bPinned && !w->m_sAdditionalConfigData.noFocus && w != pIgnoreWindow) {
                 if (box.containsPoint(g_pPointerManager->position()))
                     return w;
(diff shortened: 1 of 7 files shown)


vaxerski:
closing due to inactivity

PR #6572, opened 2024-06-17
Lower sleep times for rollinglog -f

phonetic112:
Make it faster

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -132,7 +132,7 @@ int rollingRead(const int socket) {
             buffer.fill('\0');
         }
 
-        usleep(100000);
+        usleep(1000);
     }
     close(socket);
     return 0;
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1753,7 +1753,6 @@ void runWritingDebugLogThread(const int conn) {
     std::thread([conn]() {
         while (Debug::RollingLogFollow::Get().IsRunning()) {
             if (Debug::RollingLogFollow::Get().isEmpty(conn)) {
-                std::this_thread::sleep_for(1000ms);
                 continue;
             }
 
@@ -1762,7 +1761,7 @@ void runWritingDebugLogThread(const int conn) {
                 // We cannot write, when connection is closed. So thread will successfully exit by itself
                 break;
 
-            std::this_thread::sleep_for(100ms);
+            std::this_thread::sleep_for(1ms);
         }
         close(conn);
         Debug::RollingLogFollow::Get().StopFor(conn);


vaxerski:
why spam it?

phonetic112:
Before it was slower than just using `tail -f`, this matches that speed

vaxerski:
1000 times / s is still too much IMO. Maybe 50ms?

phonetic112:
8ms is the fastest I could go without increasing CPU too much.

vaxerski:
well then 50ms is a good compromise, no? 20 times / s

Truenya:
@phonetic112 probably the best way is to make it configurable with slow default settings?

phonetic112:
how bout this

vaxerski (commented):
default 50ms

phonetic112:
yes.

src/debug/HyprCtl.cpp
```diff
@@ -1753,7 +1755,6 @@ void runWritingDebugLogThread(const int conn) {
     std::thread([conn]() {
         while (Debug::RollingLogFollow::Get().IsRunning()) {
             if (Debug::RollingLogFollow::Get().isEmpty(conn)) {
-                std::this_thread::sleep_for(1000ms);
```
Truenya, line 1756: this change creates unnecessary 100% CPU load when no logs and thread is activated by os scheduler

if u decided to make configurable sleep, better to do it configurable in both hyprctl and core. and leave the default settings as they are, please

hyprctl/main.cpp
```diff
@@ -351,6 +352,8 @@ int main(int argc, char** argv) {
             } else if ((ARGS[i] == "-f" || ARGS[i] == "--follow") && !fullArgs.contains("f")) {
                 fullArgs += "f";
                 needRoll = true;
+                if (ARGS.size() >= 3)
+                    rollDelay = std::stoi(ARGS[2]);
```
Truenya, line 356: it is not such comfortable to simple users, to count milliseconds

maybe max rps (requests per second) option will be more easy to understand, like fps in games

src/debug/HyprCtl.cpp
phonetic112: a 1ms sleep should be good enough here, right? I don't see a point in making both of them configurable

hyprctl/main.cpp
phonetic112: sure, done

src/debug/HyprCtl.cpp
Truenya: not exactly. quite a lot of users, including me, using linux to reanimate old machines with low capabilities

i think here we can sleep for similar to code bellow time.

16.(6) ms is nice speep time (60 checks in second)

but it's better to make it customizable imho

Truenya: in this loop we block 2 mutexes, which slows down the entire system (logs can be anywhere)

Therefore, ideally, it is necessary to adhere to the idea that the default cycle repetition periods should be sufficiently large.

phonetic112: how about now

src/debug/HyprCtl.cpp
```diff
@@ -1677,6 +1677,8 @@ std::string CHyprCtl::getReply(std::string request) {
 
     std::string result = "";
 
+    request = request.substr(0, request.find(" "));
```
Truenya, line 1680: Is that intended?

src/debug/HyprCtl.cpp
```diff
@@ -1821,7 +1824,17 @@ int hyprCtlFDTick(int fd, uint32_t mask, void* data) {
...
+        std::string        val;
+        if (iss >> val) {
+            if (iss >> val) {
+                runWritingDebugLogThread(ACCEPTEDCONNECTION, std::stoi(val));
+            } else {
+                runWritingDebugLogThread(ACCEPTEDCONNECTION);
+            }
+        } else {
+            Debug::log(ERR, "Invalid rollinglog request");
+        }
```
Truenya, line 1837: this construction does not look safe

is it working?

Truenya (changes requested):

src/debug/HyprCtl.cpp
phonetic112: It's there so it doesn't throw invalid request if we have an additional arg

phonetic112: oops, that actually breaks all other commands. fixed

src/debug/HyprCtl.cpp
phonetic112: yes

littleblack111:
What is the advantages of this? It just makes the compositor laggier? No? Why make this faster. Unnecessary imo

PR #6585, opened 2024-06-18
Small refactor of 'else if' statements in hyprctl

Yaraslaut:
Small refactor of else if statements inside hyperctl 

Also, some of the outputs of hyperctl does not have new lines in the end,  namely all calls to `log` function, if you are interested i can refactor it for a bit and introduce usage of `std::println` 

ready

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -450,34 +450,31 @@ int main(int argc, char** argv) {
         batchRequest(fullRequest, json);
     else if (fullRequest.contains("/hyprpaper"))
         exitStatus = requestHyprpaper(fullRequest);
-    else if (fullRequest.contains("/switchxkblayout"))
-        exitStatus = request(fullRequest, 2);
-    else if (fullRequest.contains("/seterror"))
-        exitStatus = request(fullRequest, 1);
-    else if (fullRequest.contains("/setprop"))
-        exitStatus = request(fullRequest, 3);
-    else if (fullRequest.contains("/plugin"))
-        exitStatus = request(fullRequest, 1);
-    else if (fullRequest.contains("/dismissnotify"))
-        exitStatus = request(fullRequest, 0);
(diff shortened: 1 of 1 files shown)


hyprctl/main.cpp
```diff
@@ -450,34 +450,31 @@ int main(int argc, char** argv) {
...
+                {"/dismissnotify", 0},
+                {"/notify", 2},
+                {"/output", 2},
+                {"/setcursor", 1},
+                {"/dispatch", 1},
+                {"/keyword", 2},
+                {"/decorations", 1}
+            };
+            for (const auto& [cmd, minArgs] : commands) {
+                if(fullRequest.contains(cmd))
```
vaxerski, line 473: hm, this might bug out if the rest of the request has the thing. Maybe substring only the part after the first / and before the first space?

Yaraslaut: I think that i do not change any behavior of parsing in this pr.
for safer way of handling cli option i would go directly with something like https://github.com/contour-terminal/contour/blob/121d7ade938ac99a18ff0b4f650254778baddafd/src/contour/ContourApp.cpp#L297 
and then just 
```
_syntax = parameterDefinition();
optional<CLI::flag_store> flagsOpt = CLI::parse(_syntax.value(), argc, argv);
```

vaxerski: you didn't but when already changing it why not improve?

Yaraslaut: my feeling is that this is not a good place for input validation, especially inside for loop during check for the flags

vaxerski: it's not input validation - it's making sure you choose the right value in case of e.g.

```
/notify blah blah /decorations
```

Yaraslaut: If `/notify blah blah /decorations` is not ill-formed then logic is not right, since only  `/notify` command will be handled

PR #6587, opened 2024-06-18
pointer: use software rendering when monitor is mirrored

### LINKED ISSUE #6455: mouse cursor visibility on mirrored screen

lucasreis1:
Force software locks to monitors when they are being mirrored to another screen, so we can render the pointer properly in mirror monitors. Fixes #6455 

I'm incrementing the lock counter for each new mirror. Maybe we could just increment/decrement it once if mirror list isn't/is empty, but I found it more concise this way.

Ready.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -6,6 +6,7 @@
 #include "../devices/ITouch.hpp"
 #include "../protocols/LayerShell.hpp"
 #include "../protocols/PresentationTime.hpp"
+#include "managers/PointerManager.hpp"
 #include <hyprutils/string/String.hpp>
 using namespace Hyprutils::String;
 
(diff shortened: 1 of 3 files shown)


src/helpers/Monitor.cpp
```diff
@@ -6,6 +6,7 @@
 #include "../devices/ITouch.hpp"
 #include "../protocols/LayerShell.hpp"
 #include "../protocols/PresentationTime.hpp"
+#include "managers/PointerManager.hpp"
```
vaxerski, line 9: please use relative paths

lucasreis1: sorry, fixed

vaxerski (approved):
cool, thanks!

Issue #6545, opened 2024-06-16
Ability to select previous workspace per monitor


leon-erd:
### Description

Currently there only exists the `previous` parameter to select a workspace.
It would be really nice to have a `previous_per_monitor` parameter that selects the previously focused workspace on the currently focused monitor. This means it will select a different workspace depending on what monitor you are on.
For me this would be really convenient since I usually focus the different monitors just by moving the mouse (having no benefit from the current `previous` parameter) and the `$mainMod + Tab` keybind is really easy to reach which is why I use it very often. Also when moving the mouse a lot it is hard to remember what the previous workspace was whereas it is really easy to remember the previous workspace on a monitor resulting in more intentional usage of keybinds. This new parameter would greatly improve my workflow!
On a single monitor setup the behaviour is identical to the current `previous` parameter.

Truenya:
Can i try this?

vaxerski:
feel free

leon-erd:
@Truenya works like a charm! Thanks for implementing this so fast :)
Only thing I noticed:
1. go to monitor 1
2. open two workspaces with apps on it
3. use dispatcher `movecurrentworkspacetomonitor` to move one of the workspaces to monitor 2
4. go back to monitor 1
5. dispatcher `workspace, previous_per_monitor` will change to monitor 2 instead of staying on 1

For me this is no big deal :) I dont know if you want to fix that so I will just close this issue.

Truenya:
@leon-erd can u please create an issue with this?

PR #6598, opened 2024-06-19
feat: [compositor, misc] ability to select previous workspace per monitor

Truenya:
added `previous_per_monitor` parameter that selects the previously focused workspace on the currently focused monitor
https://github.com/hyprwm/Hyprland/issues/6545

no

ready

### DIFF
--- a/src/macros.hpp
+++ b/src/macros.hpp
@@ -24,7 +24,8 @@
 
 #define STRVAL_EMPTY "[[EMPTY]]"
 
-#define WORKSPACE_INVALID -1L
+#define WORKSPACE_INVALID     -1L
+#define WORKSPACE_NOT_CHANGED -101
 
 #define LISTENER(name)                                                                                                                                                             \
     void               listener_##name(wl_listener*, void*);                                                                                                                       \
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1648,8 +1648,7 @@ PHLWORKSPACE CCompositor::getWorkspaceByString(const std::string& str) {
     }
 
     try {
-        std::string name = "";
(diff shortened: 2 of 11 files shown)


src/macros.hpp
```diff
@@ -25,6 +25,7 @@
 #define STRVAL_EMPTY "[[EMPTY]]"
 
 #define WORKSPACE_INVALID -1L
+constexpr int WORKSPACE_NOT_CHANGED = -42;
```
vaxerski, line 28: uhh no thats not a good idea for an index. Also, `#define` like others.

I'd suggest something like -101

Reasoning:
Regular workspaces start at 1 and grow up
Named start at -1337 and grow down
Special start at -99 and grow up

Truenya: Thanks! Fixed it.
About defines, I was guided by this rule when I did this
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros2

vaxerski: I realize but we use them there and for the sake of consistency we should keep doing so.

vaxerski (dismissed):
cool, needs a wiki mr

vaxerski:
and a clang-format

Truenya:
about clang format, I'm using version 18.1.6
used such script for formatting

```bash
#!/bin/bash
for file in $(git diff --name-only); do
	echo "$file formatted"
	clang-format --style=file -i $file
done
```

vaxerski:
`clang-format -i src/**/*{cpp,hpp}` works for me.
image

Just remove those changes

vaxerski (approved):
lgtm, thanks!

PR #6620, opened 2024-06-21
add virtual destructor to surfacerole to avoid undefined behaviour

gulafaran:
all classes that will be derived from should have a virtual destructor otherwise deleting an instance via pointer to a base class is undefined behaviour, CLayerShellResource/CXDGSurfaceResource hits this with std::default_delete in the new sharedptr implentation.


gulafaran:
also added a commit to fix missing includes for no precompiled headers builds and removed a redefinition of HYPRATOMS in the second commit

vaxerski (approved):
thanks!

Issue #2800, opened 2023-07-24
Import environment variables into systemd in NixOS modules


Mikilio:
## Issue

Frankly speaking, using the nixos modules without modification is a XDG-"nightmare".

This is because almost all XDG/GIO calls happen through portals. The problem being that the portals are being started as a user session service by systemd. This means that important environment variables like PATH and XDG_DATA_DIRS are not accessible. Then again there are many applications that ignore default mime applications completely and look at [Well-known Message Bus Instances](https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-types) instead.

To bring order into this situation I personally did 2 things:
- Some packages install well-known bus type dbus services (usually located in `.../share/dbus-1/`) . To select a default I will symlink the corrects one the most dominant name (usually something like org.freedesktop.BasicFoo1).
- I add the following line  in my config after declaring variables and before starting any services:
```
exec-once = ${pkgs.dbus}/bin/dbus-update-activation-environment --systemd --all
```

## Proposal

I would like to:
1. Add this information to documentation. Even the first point is necessary because Hyperland does not ship with any default tools, so every n00b will throw together their own distro and cry about XDG issues.
2. Add the line in the config by default (at least when using nix module). Home-manager already provides a way to declare environment variables for a session, so it should be fine to add the line at the top of the config. 

## Related Issues
[NixOS/nixpkgs](https://github.com/NixOS/nixpkgs/issues/160923)

PatrickShaw:
Are you running home-manager?

To my knowledge, something similar is already handled for those running HM: https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix#L138

Mikilio:
I think I remember seeing that. However it still still is not enough.
Important variables for xdg are missing like XDG_DATA_DIRS and PATH. If
there isn't any security issue I would just shorten the line by using the
flag `--all`.

fufexan:
Are you sure we want to import all environment variables when starting the session? I don't have info on this so I just kept the var list small so as not to potentially conflict with other things.

Mikilio:
I wasn't too concerned about any side effects because greetd is started by a greetd user anyways. I agree, in general it might be better to import only the ones you need. However right now some seems to be missing. I actually based my config off of yours and I still encountered the 25sec issue when regreet starts.

spikespaz:
[...]

PR #6640, opened 2024-06-23
systemd,dbus: import PATH XDG_DATA_DIRS variables into systemd and dbus session

JohnRTitor:

When opening links from a browser that would otherwise open in another app present in the system, it fails to find the app. This is because PATH is not imported into systemd/dbus.

We are currently using a [workaround in NixOS](https://github.com/NixOS/nixpkgs/blob/d603719ec6e294f034936c0d0dc06f689d91b6c3/nixos/modules/programs/wayland/hyprland.nix#L66) to explicitly define the systemd environment PATH, but ideally it should be set by the Window manager/DE at login.

An open issue highlighted this for the past one year #2800 and several threads (ie. [on discourse](https://discourse.nixos.org/t/nautilus-not-working-properly-due-to-missing-path-in-dbus/38824)) on the internet have popped up asking the user to manually import PATH. Lets not hold this any longer and let this issue die in peace.   

In addition to `PATH`, `XDG_DATA_DIRS` should be imported as well, due to it containing useful things like desktop files.

This issue affects flatpak apps as well.

This is backwards compatible.

Ready. I am using the [patch on NixOS](https://github.com/JohnRTitor/nix-conf/commit/04c0d2b6feb2aaecfa4db0ca2e9658b47d394e88) and it works.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -311,10 +311,10 @@ void CCompositor::cleanEnvironment() {
     if (m_sWLRSession) {
         const auto CMD =
 #ifdef USES_SYSTEMD
-            "systemctl --user unset-environment DISPLAY WAYLAND_DISPLAY HYPRLAND_INSTANCE_SIGNATURE XDG_CURRENT_DESKTOP QT_QPA_PLATFORMTHEME && hash "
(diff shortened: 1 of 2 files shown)


fufexan (approved):
LGTM

PR #6658, opened 2024-06-25
sessionLock: focus lock on creation based on mouse position

outfoxxed:
Causes created lock surfaces to be focused only if they would be focused by mouse movement. This makes it so on lock, the focused surface is the one the mouse is on instead of the last surface created by the locker.


vaxerski (commented):
what if no surface is under the mouse? Should we fallback?

outfoxxed:
We should probably kick focus from the currently focused surface immediately on lock creation, then only let lock surfaces be focusable.

vaxerski (approved):
lgtm, thanks!

Issue #5900, opened 2024-05-05
No mouse lock while in a match in Dota 2


Okkurat:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit cddeec47a1fc0e70d8598fd10c29cd0e9489999f  (keybinds: make the keybind manager check for session lock (5894)).
Date: Sun May 5 21:28:14 2024
Tag: v0.40.0-11-gcddeec47, commits: 4617

flags: (if any)


System Information:
System name: Linux
Node name: archlinux
Release: 6.8.9-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Thu, 02 May 2024 17:49:46 +0000


GPU information: 
28:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M/6850M XT] [1002:73df] (rev c0) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

```

</details>

### Bug or Regression?

Bug

### Description

In the AUR package of hyprland (as opposed to the current package in Arch (0.39.1-5) introduces an issue in Dota 2.

While in a match, the mouse is supposed to be locked in to the game window and moving the cursor to the edge should not move the cursor outside the game. This works as expected in the version 0.39.1-5, but in the git package, the mouse moves out of bounds and the Dota 2 mouse icon also stays.

My setup is 3 monitors and the Dota 2 is in the middle screen.

### How to reproduce

Launch Dota 2. Start a match with bots, pick a hero, spawn and try to move the mouse out out of the game.

Aqa-Ib:
Can you try to bind the fullscreen dispatcher to some key and press it twice?

Okkurat:
[...]

PR #6679, opened 2024-06-27
pointer-constraints: keep constraint alive until actual destruction

Trimutex:
Current constraint deactivation kills the constraint even if it tries to reactivate due to lifetime check always being true. Removing this fixes refocusing of windows not actually restarting constraint, requiring a recreation of the constraints by the program.
- Ex: Minecraft will not recreate constraint until menu is opened in game, so swapping to another window and back rapidly will kill the constraint without recreation, leaving pointer to roam free.

Possible fix for #5900 but can not recreate issue myself to test.

No problems with usage as proper destruction gets called anyways later when actually necessary. Perhaps a better option is to change the default lifetime of the pointer constraint, but I'm not too knowledgeable the implications of that.

Ready for merging

### DIFF
--- a/src/protocols/PointerConstraints.cpp
+++ b/src/protocols/PointerConstraints.cpp
@@ -112,6 +112,7 @@ void CPointerConstraint::deactivate() {
 
     active = false;
 
+    /* Requires additional lifetime implementations
     if (lifetime == ZWP_POINTER_CONSTRAINTS_V1_LIFETIME_ONESHOT) {
         dead = true;
         // remove from inputmgr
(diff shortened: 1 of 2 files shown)


vaxerski:
tag me when ready

Trimutex:
@vaxerski should already be ready? this still fixes the Minecraft example already. Sorry if I didn't make that clear

vaxerski:
why not just set the properties in the appropriate constructors like I've obviously forgotten to? xD

vaxerski:
should be fixed.

Trimutex:
because I missed it too lol. this does fix it properly, thanks!

PR #6696, opened 2024-06-28
Fix `listener_sessionActive` to handle an event correctly when the session get inactivated

harenayo:

I found that Hyprland handles `m_sWLRSession->events.active` signal by `listener_sessionActive`.

https://github.com/hyprwm/Hyprland/blob/b7f42a1e88a5b6c9d2dbdba31e0f35f6a02461e7/src/Compositor.cpp#L285

https://github.com/hyprwm/Hyprland/blob/b7f42a1e88a5b6c9d2dbdba31e0f35f6a02461e7/src/events/Misc.cpp#L32-L43

`listener_sessionActive` always does an activation process.

But I think `m_sWLRSession->events.active` is emitted whether the session get activated or inactivated.

https://github.com/hyprwm/wlroots-hyprland/blob/422207dbcf0949e28042403edab539159282885e/backend/session/session.c#L24-L35

That's why I made the following changes:

- Check `m_sWLRSession->active` in `listener_sessionActive`.
- Move the code for switching VTs into `listener_sessionActive`.

I'm not sure that the second change (the code for VTs) works correctly.

It is ready.

### DIFF
--- a/src/events/Misc.cpp
+++ b/src/events/Misc.cpp
@@ -30,14 +30,27 @@ void Events::listener_RendererDestroy(wl_listener* listener, void* data) {
 }
 
 void Events::listener_sessionActive(wl_listener* listener, void* data) {
-    Debug::log(LOG, "Session got activated!");
+    if (g_pCompositor->m_sWLRSession->active) {
+        Debug::log(LOG, "Session got activated!");
 
-    g_pCompositor->m_bSessionActive = true;
+        g_pCompositor->m_bSessionActive = true;
 
-    for (auto& m : g_pCompositor->m_vMonitors) {
-        g_pCompositor->scheduleFrameForMonitor(m.get());
-        g_pHyprRenderer->applyMonitorRule(m.get(), &m->activeMonitorRule, true);
+        for (auto& m : g_pCompositor->m_vMonitors) {
+            g_pCompositor->scheduleFrameForMonitor(m.get());
+            g_pHyprRenderer->applyMonitorRule(m.get(), &m->activeMonitorRule, true);
+        }
+
+        g_pConfigManager->m_bWantsMonitorReload = true;
+    } else {
+        Debug::log(LOG, "Session got inactivated!");
+
+        g_pCompositor->m_bSessionActive = false;
+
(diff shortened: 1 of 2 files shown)


src/events/Misc.cpp
```diff
@@ -30,14 +30,27 @@ void Events::listener_RendererDestroy(wl_listener* listener, void* data) {
...
+
+        g_pCompositor->m_bSessionActive = false;
 
-    g_pConfigManager->m_bWantsMonitorReload = true;
+        for (auto& m : g_pCompositor->m_vMonitors) {
+            m->noFrameSchedule = true;
+            m->framesToSkip    = 1;
+        }
+
+        Debug::log(LOG, "Destroyed all render data, frames to skip for each: 2");
```
vaxerski, line 54: log is wrong, can be removed

harenayo: I just removed it.

src/managers/KeybindManager.cpp
```diff
@@ -1048,14 +1037,20 @@ SWorkspaceIDName getWorkspaceToChangeFromArgs(std::string args, PHLWORKSPACE PCU
         return {WORKSPACE_NOT_CHANGED, ""};
     }
 
-    const auto ID = PCURRENTWORKSPACE->m_iID;
-    if (const auto PWORKSPACETOCHANGETO = g_pCompositor->getWorkspaceByID(PPREVWS.id); PWORKSPACETOCHANGETO) {
-        if (PER_MON && PCURRENTWORKSPACE->m_iMonitorID != PWORKSPACETOCHANGETO->m_iMonitorID)
-            return {WORKSPACE_NOT_CHANGED, ""};
+    const auto ID                   = PCURRENTWORKSPACE->m_iID;
```
vaxerski, line 1040: unintentional?

harenayo: This change is from b7f42a1e88a5b6c9d2dbdba31e0f35f6a02461e7 and is unintentional if it should not be here.

vaxerski: correct, you'll need to remove that. The commit was reverted.

vaxerski (approved):
thanks!

PR #6745, opened 2024-07-02
Add functions to hyprctl header

Yusuf-Duran:
Adds certain Object to json format functions to the header so they can be called by other files (in this case I need this for my plugin. The other option would be for me to write the function myself which would be kinda dumb). I also refactored one function to have the monitor to json code it's own function. 

I'm not sure if this is the best way to make them available for other files but this works. 

ready


src/debug/HyprCtl.hpp
```diff
@@ -29,4 +29,8 @@ class CHyprCtl {
     wl_event_source*                 m_eventSource = nullptr;
 };
 
+std::string                      getWindowData(PHLWINDOW w, eHyprCtlOutputFormat format);
```
vaxerski, line 32: hm, I don't like leaving globals like this, without any namespace or anything.

Does anyone need those outside of the one cpp file anyways? Why should they be exposed?

Yusuf-Duran: In my case, I'm developing a plugin for Hyprland that acts as a proxy for events. This plugin allows you to run scripts in Bash, Python, etc., when event hooks are called. This makes some dotfile functionality possible without having to write a plugin each time you got an idea. To enhance functionality, I want to pass the JSON representation of windows, monitors, etc., as a parameter to these scripts.

Since there's already maintained code for generating these JSON representations, it would be redundant and inefficient for me to write it from scratch, given it means copying the existing code from hyprctl.cpp and maintaining it myself. I'm looking for better options to make this accessible from outside the plugin. Any suggestions for a more efficient solution would be greatly appreciated.

Yusuf-Duran: One possible option would be moving the functions to the class and making them static so they can get called through the `g_pHyprCtl` pointer.

zakk4223: "Why should they be exposed?"

Because you write a plugin that hooks something in that file, like say you want to slightly reimplement it to add some extra data or whatever. But the function uses a bunch of 'private' functions in HyprCtl.cpp so you either have to cut+paste them or use pointers to them or something similarly ugly.

If they were just in a namespace at least you could get to them somehow.

vaxerski: fair enough.

Just move it somewhere that isnt global then, please.

src/debug/HyprCtl.hpp
```diff
@@ -29,4 +29,11 @@ class CHyprCtl {
     wl_event_source*                 m_eventSource = nullptr;
 };
 
+namespace HyprCtl {
+    std::string getWindowData(PHLWINDOW w, eHyprCtlOutputFormat format);
```
vaxerski, line 33: why not put them inside CHyprCtl? It's a singleton. If you prefer, you can even make these static.

Yusuf-Duran:
it was a mistake getting the latest changes from main :standing_man: . Now the codestyle check won't do :+1:

src/debug/HyprCtl.hpp
Yusuf-Duran: Aight done chief :saluting_face: . Should I just add every method into the class and make it static in case someone else needs it in the future, or you want it not to be more cluttered than needed?

vaxerski (approved):
thanks!

PR #6776, opened 2024-07-05
Improve handling of window properties

MightyPlaza:
In this PR, `CWindowOverridableVar` allows setting different priorities for each value to fix issues where the layout, the workspace/window rules and `setprop` want different things.
Most properties have become window rules as well, and vice versa.
`SWindowAdditionalConfigData` and `SWindowSpecialRenderData` have been unified into SWindowData
`mbWindowProperties` and `miWindowProperties` have been created to avoid having to specify window rules / `setprop` in multiple places (ideally there would be a global struct for all types but I'm not sure how to do that without using `void*`)

Details:
`allowsinput`, `dimaround`, `decorate`, `focusonactivate`, `keepaspectratio`, `nearestneighbor`, `noanim`, `noblur`
`noborder`, `nodim`, `nofocus`, `nomaxsize`, `norounding`, `noshadow`, `opaque`, `forcergbx`, `immediate`, `xray`, `windowdance` can now be set with both windowrules and `setprop`
windowrule: `allowsinput` (set as true for backwards compatibility), `allowsinput 0` or `allowsinput 1`
`setprop`: `allowsinput 0` or `allowsinput 1`, `allowsinput toggle` and `allowsinput unset` (to unset value previously set with `set prop` or exec rules)
setprop also allows `bordersize unset` and `rounding toggle`

some workspace rule values have been flipped internally to match the negative nature of the window properties
`forceinput` has been removed because it wasn't used anywhere in the codebase (???)
exec rules now have the same priority as `setprop` so they won't disappear when the window rules get reloaded neither get overridden by a window rule that would match the same window

and I think that's mostly it
seems to work fine for some rules and cases I've tested but still haven't tested all case
also needs a wiki MR, but should be ready for review

### DIFF
--- a/src/config/ConfigManager.hpp
+++ b/src/config/ConfigManager.hpp
@@ -169,12 +169,10 @@ class CConfigManager {
     std::string                                                                             configCurrentPath;
 
     std::unordered_map<std::string, std::function<CWindowOverridableVar<bool>*(PHLWINDOW)>> mbWindowProperties = {
-        {"dimaround", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.dimAround; }},
+        {"allowsinput", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.allowsInput; }},
+        {"dimaround", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.dimAround; }},
+        {"decorate", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.decorate; }},
         {"focusonactivate", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.focusOnActivate; }},
-        {"forceinput", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.allowsInput; }},
-        {"forceopaque", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.opaque; }},
-        {"forcergbx", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.RGBX; }},
-        {"immediate", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.tearing; }},
         {"keepaspectratio", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.keepAspectRatio; }},
         {"nearestneighbor", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.nearestNeighbor; }},
         {"noanim", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.noAnim; }},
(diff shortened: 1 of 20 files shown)


src/desktop/Window.hpp
```diff
@@ -59,126 +59,129 @@ enum eSuppressEvents {
 
 class IWindowTransformer;
 
+struct sAlphaValue {
```
vaxerski, line 62: SAlphaValue

src/desktop/Window.hpp
```diff
@@ -59,126 +59,129 @@ enum eSuppressEvents {
...
+    T value() {
+        if (!values.empty())
+            return std::prev(values.end())->second;
+        else
+            throw std::bad_optional_access();
     }
 
-    bool operator>(T const& other) {
-        return value > other;
+    T value_or(T const& other) {
```
vaxerski, line 121: no snake

src/config/ConfigManager.hpp
```diff
@@ -148,25 +148,50 @@ class CConfigManager {
...
+        {"nodim", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.noDim; }},
+        {"nofocus", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.noFocus; }},
+        {"nomaxsize", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.noMaxSize; }},
+        {"norounding", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.noRounding; }},
+        {"noshadow", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.noShadow; }},
+        {"opaque", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.opaque; }},
+        {"forcergbx", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.RGBX; }},
+        {"immediate", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.tearing; }},
+        {"xray", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.xray; }},
+        {"windowdance", [](PHLWINDOW pWindow) { return &pWindow->m_sWindowData.windowDanceCompat; }},
```
vaxerski, line 190: should be removed, it's broken and unused

MightyPlaza: `windowdance` I'm guessing?

vaxerski: ya

vaxerski:
value or default is still snake

MightyPlaza:
> value or default is still snake

only now realized what you meant by snake

vaxerski (approved):
thanks!

vaxerski:
wiki mr pls

MightyPlaza:
PR done in https://github.com/hyprwm/hyprland-wiki/pull/709

PR #6822, opened 2024-07-09
Allow vrr with vfr=false

UjinT34:
Allows VRR with `vfr = false` as a workaround for nvidia ghosting/tearing on aquamarine branch. With 
```
misc {
    vfr = false
    vrr = 2
}

experimental {
    explicit_sync = true
}

```
those issues will be present for some apps in fullscreen mode only.

Ready


vaxerski:
it's not ready because half of the checks fail

UjinT34:
> it's not ready because half of the checks fail

It's for aquamarine branch. main is ok without this feature.

vaxerski:
isnt this a bug worth fixing, though? I mean, vfr off will come at a sizeable cpu/gpu hit

UjinT34:
Any progress on this bug?
I don't understand why VRR requires `vfr = true`. Looks like these settings should be independent. This feature "auto enables" vfr when vrr is active.

UjinT34:
I understand. That's why this feature till it properly fixed. I don't think anyone will miss `vfr = false` blocking VRR %)
Too much fuss about one line that doesn't break anything...

PR #6843, opened 2024-07-11
sessionLock: fix the check for locking a locked session

PaideiaDilemma:
It fixes the following misbehavior:

When a client holds a locked session lock object and another client tries to lock the session as well, the second client will receive `finished`. When the second client never calls `destroy` or `unlock_and_destroy` it effectively blocks the user from locking the session, because `m_vLocks` still contains the lock object that was not destroyed.

I think this is the correct behavior. Since `RESOURCE->inert` is set whenever `sendFinished` is called anyways, it should also be safe for the left over lock object to send `unlock_and_destroy` at some point. 

Please check if i am correct, but should be fine.

Ready.


vaxerski (approved):
thanks!

PR #6850, opened 2024-07-12
session-lock: send `locked` after the lock screen is properly rendered

MikeWalrus:
The protocol says:
> The locked event "must not be sent until a new "locked" frame (either from a
> session lock surface or the compositor blanking the output) has been presented
> on all outputs and no security sensitive normal/unlocked content is possibly
> visible".

This helps users ensure the screen is properly locked before suspending the machine. (e.g. with swaylock --ready-fd)

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -3,6 +3,7 @@
 #include "../config/ConfigValue.hpp"
 #include "../protocols/FractionalScale.hpp"
 #include "../protocols/SessionLock.hpp"
+#include <algorithm>
 
 SSessionLockSurface::SSessionLockSurface(SP<CSessionLockSurface> surface_) : surface(surface_) {
     pWlrSurface = surface->surface();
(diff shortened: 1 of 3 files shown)


src/managers/SessionLockManager.cpp
```diff
@@ -118,6 +117,19 @@ float CSessionLockManager::getRedScreenAlphaForMonitor(uint64_t id) {
     return std::clamp(NOMAPPEDSURFACETIMER->second.getSeconds() - /* delay for screencopy */ 0.5f, 0.f, 1.f);
 }
 
+void CSessionLockManager::onLockscreenRenderedOnMonitor(uint64_t id) {
+    if (!m_pSessionLock || m_pSessionLock->m_hasSentLocked) {
```
vaxerski, line 121: no {}

vaxerski (commented):
what if a faulty lock doesnt lock all monitors?

MikeWalrus:
It'll show the fade-in red screen until it becomes opaque. Then it'll send the locked event.

> what if a faulty lock doesnt lock all monitors?

vaxerski:
right, I see. Alright, just fix the style and this is gtg

src/managers/SessionLockManager.cpp
MikeWalrus: done

vaxerski (approved):
thanks!

Issue #6852, opened 2024-07-12
foot terminal - broken transparency/blur


izmyname:
### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit e728e56cbc6af0a9be1276b2cf8e019a894016e7  (meson: install wayland.hpp header).
Date: Thu Jul 11 22:58:23 2024
Tag: v0.41.2-31-ge728e56c, commits: 4917

flags: (if any)


System Information:
System name: Linux
Node name: archbtw
Release: 6.9.9-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 12 Jul 2024 00:06:19 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation CometLake-H GT2 [UHD Graphics] [8086:9bc4] (rev 05) (prog-if 00 [VGA controller])
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU116M [GeForce GTX 1660 Ti Mobile] [10de:2191] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  555.58.02  Tue Jun 25 01:39:15 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/arch-mage/.config/hypr/hyprland.conf: Read Succeeded
#set default monitor
#use 'hyprctl monitors' 
monitor=,preferred,auto,1

#theme
source=$HOME/.config/hypr/mocha.conf

general {
    # See https://wiki.hyprland.org/Configuring/Variables/ for more

    gaps_in = 5
    gaps_out = 20
    border_size = 1
    col.active_border = rgba($blueAlphaee) rgba($sapphireAlphaee) 45deg
    col.inactive_border = rgba($baseAlphaaa)
    layout = dwindle

[...]

PR #6855, opened 2024-07-12
fix: xray regressions

MightyPlaza:
fixes regression where `shouldUseNewBlurOptimizations` would return `false` where it shouldn't .
fixes regression where `passRequiresIntrospection` would ignore a window since `-1 == true` .

no idea how xray (windowrule/prop) works but seems to be making foot opaque when using it's config file

fixes: https://github.com/hyprwm/Hyprland/issues/6852


vaxerski (approved):
thanks!

PR #6857, opened 2024-07-12
Send done after title and app id events for ForeignTopLevel protocol

caffineehacker:
According to the spec (https://wayland.app/protocols/ext-foreign-toplevel-list-v1#ext_foreign_toplevel_handle_v1:event:title), clients should wait for the done signal before applying updates

Adds the done event after the title event to comply with they Wayland spec.

I don't expect this to break any users unless they are expecting done only for new windows.

Ready for merging


vaxerski (approved):
thanks!

PR #6862, opened 2024-07-13
feat: dispatchers, add hidespecialworkspace

Bartkk0:
Adds a dispatcher `hidespecialworkspace` that hides the current special workspace

I don't think so

I think it's ready for merging

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -87,6 +87,7 @@ CKeybindManager::CKeybindManager() {
     m_mDispatchers["focusworkspaceoncurrentmonitor"] = focusWorkspaceOnCurrentMonitor;
     m_mDispatchers["moveworkspacetomonitor"]         = moveWorkspaceToMonitor;
     m_mDispatchers["togglespecialworkspace"]         = toggleSpecialWorkspace;
+    m_mDispatchers["hidespecialworkspace"]           = hideSpecialWorkspace;
     m_mDispatchers["forcerendererreload"]            = forceRendererReload;
     m_mDispatchers["resizeactive"]                   = resizeActive;
     m_mDispatchers["moveactive"]                     = moveActive;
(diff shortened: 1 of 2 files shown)


Eclextic:
Hide as in, don't show the special workspace anymore or hide as in, make it inaccessible?

Bartkk0:
> Hide as in, don't show the special workspace anymore or hide as in, make it inaccessible?

Hide as in don't show it anymore

MightyPlaza:
i think it would make more sense to add an arg to `togglespecialworkspace` (like `0`/`1`/`toggle`) or create a new dispatcher that could open, close and toggle instead of creating one just to close it

vaxerski:
mightyplaza is right on both fronts

Eclextic:
> i think it would make more sense to add an arg to `togglespecialworkspace` (like `0`/`1`/`toggle`) or create a new dispatcher that could open, close and toggle instead of creating one just to close it also vaxry will want a wiki MR

Yeah I agree, that's what confused me a bit... The `togglespecialworkspace` functionality closely resembles what you are trying to do, so it should probably become an argument.

Also I can't think of a case where anyone would need to *just* close the special workspace instead of toggling it...

Bartkk0:
for me the use case is to hide the currently visible special workspace, `togglespecialworkspace` could have an additional argument, but it would also need to handle referring to the current workspace

Kommynct:
Always hiding the special workspace would be useful for making it so that when you switch to a workspace the specia workspace always hides 

I currently use this bound to every single workspace hotkey for this and it'd certainly simplify that 

hyprctl monitors -j | jaq -r '.[] | select(.focused == true) | .specialWorkspace.name' | sd 'special:' '' | xargs -I [] hyprctl dispatch togglespecialworkspace []

dxrcy:
#8594 provides alternative solutions may be of use.

PR #6868, opened 2024-07-13
partial revert previous xray fix

MightyPlaza:
partially reverts https://github.com/hyprwm/Hyprland/pull/6855 since `-1 == true` is actually false


vaxerski (approved):
life is life

PR #6608, opened 2024-06-20
Core: Move to aquamarine

vaxerski:
This is for now very basic work to get Hyprland to move from wlroots to aquamarine https://github.com/hyprwm/aquamarine for the backend.

Very, very wip.

After this MR, we'll no longer be based on wlroots, at all.

## Aquamarine Support lacking
 - [x] DRM sessions
 - [x] Headless
 - [x] Session
 - [x] Libinput
   - [x] tablet
   - [x] touch
   - [x] gesture
   - [x] switch
 
## Features disabled and not re-implemented
 - [x] Input devices
 - [x] Gamma ctrl
 - [x] hyprctl output
 - [x] VT switching (session)
 - [x] Direct Scanout
 - [x] Screencopy
 - [x] Cursors
 - [x] hyprctl switchxkblayout
 - [x] Keyboard LED control
 - [x] vt (drm sessions also are broken)
 - [x] edid
 
## DRM support lacking (aq / hl)
 - [x] Atomic modesetting
 - [x] multigpu
 - [x] gamma
 - [x] planes (cursor)
 - [x] vt restore
 - [x] adaptive sync
 - [x] lease

## Bugs (aq/hl)
 - [x] Wayland session struggles to update
 - [x] DMABUFs are broken
 - [x] dt issues as usual
 - [x] modifiers on drm are wonky
 - [x] drm graphical glitches
 - [x] coming back from vt is broken
 - [x] direct scanout looks fucked
 - [x] Atomic modesetting stutters... I think?
 - [x] DRM sessions use 100% of a core
 - [x] CPU usage is kinda high on drm
 - [x] Nvidia bugs and glitches from the vrr fix
 - [x] Input devices die after replug
 - [x] https://github.com/hyprwm/Hyprland/pull/6608#issuecomment-2218792487 
 - [x] multigpu graphical glitches issues
 - [x] hotplug issues
 
## Misc
 - [x] We still link wlroots for a few utilities. Stop doing that.
 
## Notes
 - requires (always) hyprwayland-scanner git. I'll bump the ver once we're close to being done with this MR.

PR #6875, opened 2024-07-14
[aquamarine] try to fix direct scanout

ikalco:
will try to fix direct scanout and linux-dmabuf protocol for #6608

 - [x] use monitor primary plane formats for ds (hyprwm/aquamarine#13)
 - [x] support monitor connect/disconnect for format_table (hyprwm/aquamarine#14)
 - [x] fix weird bug that disables ds randomly
 - [x] fix weird artifacts that happen randomly when ds [^1] **(thanks vaxry!!!)**
 - [x] <s> only send formats than can fallback to renderer even in scanout tranches</s> \
    add renderer fallback tranche when attempting ds [^2]
 - [x] maybe use sets in format table to conserve memory, would be hard to keep track of indicies tho
 - [ ] <s>maybe create seperate format_tables to mitigate bad clients [^3]</s>

[^1]:  after reading some of weston's code this might be related to plane damaging (drm.c@470)? this is a guess so far tho
[^2]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/linux-dmabuf/feedback.rst
    >Compositors should not send feedback parameters if they don't have a fallback path. For instance, compositors shouldn't send a format/modifier supported for direct scan-out but not supported by the rendering API for texturing.

    after some more reading the ds tranche is allowed to have formats that the egl doesnt support as long as there is a fallback tranche that the renderer does support
[^3]: the format_table contains all the formats from egl and planes but some wl clients only use format_table to find formats instead of tranches
    only "bad" client i could find was mpv so I just bug reported there

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -10,6 +10,7 @@
 #include "../protocols/DRMLease.hpp"
 #include "../protocols/core/Output.hpp"
 #include "../managers/PointerManager.hpp"
+#include "../protocols/core/Compositor.hpp"
 #include "sync/SyncTimeline.hpp"
 #include <aquamarine/output/Output.hpp>
 #include <hyprutils/string/String.hpp>
@@ -784,6 +785,53 @@ void CMonitor::scheduleDone() {
         return;
 
     doneSource = wl_event_loop_add_idle(g_pCompositor->m_sWLEventLoop, ::onDoneSource, this);
+}
+
+bool CMonitor::attemptDirectScanout() {
+    if (!mirrors.empty() || isMirror() || g_pHyprRenderer->m_bDirectScanoutBlocked)
+        return false; // do not DS if this monitor is being mirrored. Will break the functionality.
+
+    if (g_pPointerManager->softwareLockedFor(self.lock()))
+        return false;
+
+    const auto PCANDIDATE = solitaryClient.lock();
+
+    if (!PCANDIDATE)
+        return false;
+
+    const auto PSURFACE = g_pXWaylandManager->getWindowSurface(PCANDIDATE);
+
+    if (!PSURFACE || !PSURFACE->current.buffer || PSURFACE->current.buffer->size != vecPixelSize || PSURFACE->current.transform != transform)
+        return false;
+
+    // we can't scanout shm buffers.
+    if (!PSURFACE->current.buffer->dmabuf().success)
+        return false;
+
+    // FIXME: make sure the buffer actually follows the available scanout dmabuf formats
+    // and comes from the appropriate device. This may implode on multi-gpu!!
+
+    output->state->setBuffer(PSURFACE->current.buffer);
+
+    if (!state.test())
+        return false;
+
+    timespec now;
+    clock_gettime(CLOCK_MONOTONIC, &now);
+    Debug::log(TRACE, "presentFeedback for DS");
+    PSURFACE->presentFeedback(&now, this, true);
+
+    if (state.commit()) {
+        if (lastScanout.expired()) {
+            lastScanout = PCANDIDATE;
+            Debug::log(LOG, "Entered a direct scanout to {:x}: \"{}\"", (uintptr_t)PCANDIDATE.get(), PCANDIDATE->m_szTitle);
+        }
+    } else {
+        lastScanout.reset();
+        return false;
+    }
+
+    return true;
 }
 
 CMonitorState::CMonitorState(CMonitor* owner) {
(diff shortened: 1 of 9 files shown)


ikalco:
Found an interesting repro case for the random artifacts during ds, xonotic
if run with `SDL_VIDEODRIVER=x11 xonotic-sdl` there is one or two arifacts when starting the game up
but run with `SDL_VIDEODRIVER=wayland xonotic-sdl` there is major artifacting that looks like its related to damaging

ikalco:
another repro case for artifacting
https://github.com/ascent12/compositor-killer.git

ikalco:
gamescope has a bug report that looks a lot like the our artifacting
https://github.com/ValveSoftware/gamescope/issues/1368

vaxerski:
so is that a kernel bug?

ikalco:
> so is that a kernel bug?

not entirely sure, it supposedly works on nvidia and igpu without artifacting but amd gets artifacting
but then again amd doesnt get artifacting on sway so maybe they have a mitigation related to below
>edit:
can repro that lowering frame rate in game (xonotic) to below vsync removed all artifacts

not sure how to test how many fps the client is actually rendering to test if this is true though

Agent00Ming:
recorded on this branch, intel iGPU

https://github.com/user-attachments/assets/7891a87f-b6cc-4d90-b84b-34358ac25054

ikalco:
After a lot of testing the artifacts dont seem to be a kernel/driver/mesa bug
its definitely related to vsync, the more accurate the fps is to the refresh rate the less the artifacts
I've tested a few games and enabling vsync will make most artifacts go away

PS:
I dont really understand enough about how Hyprland maintains fps and rr to be able to fix the artifacting
is fps and rr maintained by DRM? is page flip event called by driver exactly right to line up with refresh rate?
either way I think something in attemptDirectScanout isn't the same as the rest of renderMonitor in relation to vsync/fps/rr which causes the bug

vaxerski:
@phonetic112 in 6608 said the glitches go away on legacy drm api, perhaps something there is off?

> how is fps and rr maintained by Hyprland?

render ---> submit with flip event

submit submits a frame to be presented on next vblank (monitor changes image = vblank)
then, when that vblank happens and our image is presented, we get an event (see handlePF in aq) in which (if applicable) we send a .frame event to signify its a good time to render the next frame. If there is no damage, and a frame isnt needed, we dont. This stops the cycle and will resume it the next time something schedules a frame (e.g. damage)

ikalco:
no disabling hw cursors doesnt fix artifacting

btw to get artifacting to go away i used vrrtest, vsync on, fps cap 1 fps above vsync, and busy wait on (busy wait says it more accurately tracks vsync)
you'll also see that the closer you get to vsync the artifacting will slow down

ikalco:
> @phonetic112 in 6608 said the glitches go away on legacy drm api, perhaps something there is off?
> 
> > how is fps and rr maintained by Hyprland?
> 
> render ---> submit with flip event
> 
> submit submits a frame to be presented on next vblank (monitor changes image = vblank) then, when that vblank happens and our image is presented, we get an event (see handlePF in aq) in which (if applicable) we send a .frame event to signify its a good time to render the next frame. If there is no damage, and a frame isnt needed, we dont. This stops the cycle and will resume it the next time something schedules a frame (e.g. damage)

wait, so how do we make sure we dont schedule more than, lets say, 60 fps?

vaxerski:
backend will err out if you commit a new buffer before handlePF

ikalco:
last commit was able to make artifacting only appear when vrrtest frame cap is above monitor refresh rate except when starting it up when it happens for a few frames

also arch ci seems to be broken, something with updating packages

ikalco:
the wlroots ds pr mentions holding client buffers longer
> We need to hold client buffers for longer than usual. When compositing, we need to hold buffers until the buffer swap (and the kernel will take care of the rest with implicit synchronization). When scanning out, we need to hold buffers as long as they're visible on screen.

didn't have time to check this out but if we arent then that could be causing artifacts

vaxerski:
heh I did once make a patch to do that but it didnt stop the glitches, I can do it again

vaxerski:
committed the new infra impl here, feel free to drop it if it doesnt work

ikalco:
> committed the new infra impl here, feel free to drop it if it doesnt work

it worked!!! artifacting is now completely gone on AMD, should be gtg other than testing on nvidia and igpu

Agent00Ming:
works on nvidia but something is going wrong with the cursor hanging the whole desktop

ikalco:
yeah i just  saw that, i think i might know that it is

Agent00Ming:
Yeah, works on nvidia. Occasionally spams entered/left direct scanout in logs when initially starting a fullscreen window or when system is under heavy load when starting the direct scanout but that's a non-issue and stabilizes in a few seconds Will test intel in a moment.

vaxerski:
ready to merge then? or no

ikalco:
yes as far as i can tell this is ready

vaxerski:
thanks!

Issue #5918, opened 2024-05-07
VRR is broken in Hyprland


Gwenodai:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit cba1ade848feac44b2eda677503900639581c3f4  (props: bump version to 0.40.0).
Date: Sat May 4 15:42:32 2024
Tag: , commits: 1

flags: (if any)


System Information:
System name: Linux
Node name: gwen-x570
Release: 6.8.9-300.fc40.x86_64
Version: #1 SMP PREEMPT_DYNAMIC Thu May  2 18:59:06 UTC 2024


GPU information: 
05:00.0 VGA compatible controller [0300]: NVIDIA Corporation GP106 [GeForce GTX 1060 6GB] [10de:1c03] (rev a1) (prog-if 00 [VGA controller])
0c:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M/6850M XT] [1002:73df] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="Fedora Linux"
VERSION="40 (Sway)"
ID=fedora
VERSION_ID=40
VERSION_CODENAME=""
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Linux 40 (Sway)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f40/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://gitlab.com/fedora/sigs/sway/SIG/-/issues"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=40
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=40
SUPPORT_END=2025-05-13
VARIANT="Sway"
VARIANT_ID=sway


plugins:
```

</details>

### Bug or Regression?

Bug

### Description

[...]

PR #6877, opened 2024-07-14
Fix VRR for constrained cursors

UjinT34:

- allows `cursor:no_break_fs_vrr` to work with locked cursors
- skips window z order changes if window is in correct position
- uses `m_pLastMonitor` instead of `getMonitorFromCursor` with locked cursor in some places

When cursor is constrained (and maybe also locked) `g_pSeatManager->sendPointerFrame()` results in an extra `listener_commitWindow` call which schedules a frame for monitor. These extra frames break VRR. https://github.com/hyprwm/Hyprland/issues/5918#issuecomment-2189672071
Since there is no direct connection (at least I couldn't find one) between `sendPointerFrame` and `listener_commitWindow` and the latter also happens when an app sends a new frame it is hard to filter unnecessary frame schedules.

This PR delays `sendPointerFrame` to be called when an app sends a new frame and then skips the damage from next `listener_commitWindow`. This fixes VRR ~but loses most of the cursor movement between the frames~.
I don't think that it is the correct way to fix this issue. If it is possible to not get those extra window commit events than it will be a better solution.

If something is wrong with this PR as a whole consider cherry-picking 05de4f3
`if clicked on a floating window make it top` comment is misleading. `changeWindowZOrder` is called for every clicked window. Is it intended behaviour or a bug?
`CInputManager::onMouseMoved` might set the cursor position outside the constraint. Some code uses this incorrect position before the constraints are handled. This PR fixes several of such situations. There might be more.

Works more reliable when the app is the only one on its workspace.

Ready if there is no better way to do it.

- [x] check this hack is disabled when `vrr = 0`
- [x] test GW2
- [x] deal with movement speed
- [x] test games on special workspaces
- [x] fast mouse movement can briefly jump outside of constraint


Myned:
With this PR, `vrrtest` on a regular workspace with the below relevant config results in:
- No cursor movement = functional VRR
- Cursor movement = lowered framerate while moving despite software cursors and `no_break_fs_vrr`
- Mouse clicking = refresh rate spikes consistent with the main issue

<details>
<summary>hyprland.conf</summary>

```
... omitted: exec-once

plugin=/nix/store/4ciqvzb102b03k8gx3qsirvnbm0wn9bh-hyprbars-0.1/lib/libhyprbars.so
binds {
  allow_workspace_cycles=true
  disable_keybind_grabbing=true
  scroll_event_delay=0
}

cursor {
  hotspot_padding=10
  min_refresh_rate=0
  no_break_fs_vrr=true
  no_hardware_cursors=true
  no_warps=true
  zoom_rigid=true
}

decoration {
  blur {
    enabled=false
  }
  col.shadow=rgba(00000040)
  col.shadow_inactive=rgba(0000001a)
  dim_special=0.500000
  rounding=16
  shadow_range=50
  shadow_render_power=4
}

device {
  name=kensington-orbit-wireless-tb-mouse
  accel_profile=adaptive
  left_handed=true
  middle_button_emulation=true
  natural_scroll=true
  sensitivity=-0.500000
}

device {
  name=logitech-m570
  accel_profile=adaptive
  sensitivity=-1
}

device {
  name=wireless-controller-touchpad
  enabled=false
}

general {
  border_size=2
  col.active_border=rgb(93a1a1)
  col.inactive_border=rgba(93a1a140)
  col.nogroup_border=rgba(dc322f40)
  col.nogroup_border_active=rgb(dc322f)
  gaps_in=10
  gaps_out=20
  layout=master
  resize_on_border=true
}

gestures {
  workspace_swipe=true
  workspace_swipe_cancel_ratio=0.200000
  workspace_swipe_distance=1000
  workspace_swipe_forever=true
  workspace_swipe_min_speed_to_force=10
}

group {
  groupbar {
    col.active=rgb(6c71c4)
    col.inactive=rgba(6c71c440)
    col.locked_active=rgb(d33682)
    col.locked_inactive=rgba(d3368240)
    font_size=12
    height=10
    render_titles=false
    text_color=rgb(93a1a1)
  }
  col.border_active=rgb(6c71c4)
  col.border_inactive=rgba(6c71c440)
  col.border_locked_active=rgb(d33682)
  col.border_locked_inactive=rgba(d3368240)
  insert_after_current=false
}

input {
  touchpad {
    clickfinger_behavior=true
    natural_scroll=true
    scroll_factor=0.300000
  }
  accel_profile=flat
  float_switch_override_focus=0
  follow_mouse=1
  mouse_refocus=false
  repeat_delay=400
  repeat_rate=40
  sensitivity=0.500000
}

master {
  always_center_master=true
  mfact=0.500000
  new_on_active=before
  orientation=center
}

misc {
  background_color=rgb(073642)
  disable_hyprland_logo=true
  disable_splash_rendering=true
  focus_on_activate=true
  font_family=Iosevka NF SemiBold
  force_default_wallpaper=0
  initial_workspace_tracking=0
  key_press_enables_dpms=true
  middle_click_paste=false
}

plugin {
  hyprbars {
    bar_button_padding=12
    bar_color=rgb(002b36)
    bar_height=36
    bar_padding=12
    bar_precedence_over_border=true
    bar_title_enabled=false
    hyprbars-button=rgb(dc322f), 16,, /nix/store/wcwpgpb6ragsn2vz96r19xp2fr4l2x6d-hyprland-0.41.2/bin/hyprctl dispatch killactive
    hyprbars-button=rgb(d33682), 16,, /nix/store/wcwpgpb6ragsn2vz96r19xp2fr4l2x6d-hyprland-0.41.2/bin/hyprctl dispatch fullscreen 1
    hyprbars-button=rgb(6c71c4), 16,, minimize
  }
}

xwayland {
  force_zero_scaling=true
}
animation=global, 1, 5, default
animation=specialWorkspace, 1, 5, default, fade

... omitted: bind, envd, exec-once, windowrulev2, workspace
```

</details>

Myned:
To address the rest of the tests:
- `vrr = 0` correctly disables VRR with no cursor side effects

With `vrr = 2`, `= 1` not tested:
- GW2 is <1 fps while moving cursor with `min_refresh_rate = 0`, `= 60` results in dropping to 60 during movement, followed by a large stutter when stopping movement that seems visually identical to mouse click
- GW2 has large stutters when clicking mouse, unsure if framerate or refresh rate as I do not have a monitor OSD
- Cursor update speed does seem lower than the current framerate while VRR is active
- VRR with `vrrtest` does not work in a special workspace

UjinT34:
GW2 internal fps limit doesn't seem to do a good job, fps fluctuates around the limit. Mangohud limit fixes that.
Can't reproduce mouse click issues in GW2.

`cursor:no_break_fs_vrr` works only on regular workspaces. Is there any need to have it on special?

Note that there is an issue with switching `vrr` from 2 to 1 https://github.com/hyprwm/Hyprland/issues/6893
This fix and `cursor:no_break_fs_vrr` doesn't rely on `vrr` setting. They check the real adaptive sync status.

UjinT34:
This PR shouldn't change the `vrrtest` behaviour.
Still can't detect any spikes when clicking. Does this happen with mouse or with touchpad?

Myned:
I wasn't sure initially, but the latest commit reverted functionality to the same as Hyprland main @ bd526822deb9ed47c0b51b534817aa8541fff07b, as far as I can tell. The spikes happen on both main and this PR for my system. It does not occur with a bluetooth controller nor the keyboard. None of the usual suspect keywords (under general, input, misc, or cursor) seem to have an effect on whether it spikes on click.

Tested with hardware:
- Razer Viper Ultimate in wireless mode, all buttons including side
- Kensington Orbit trackball, both buttons
- AMD RX 6750 XT with `pp_power_profile_mode = 1`
- Samsung S34CG50 monitor @ 48?-100hz, target framerate set to 60

Gwenodai:
I can confirm what [Myned](https://github.com/Myned) is saying. Clicking M1, M2, M3 (mmb), M4, or M5 cause refresh rate spikes. as seen in the video below

https://github.com/user-attachments/assets/477612da-40b8-4afe-812f-ff824f012b7a

On the other hand I've noticed another bug happening with mouse movements that I'd not encountered before. With `no_break_fs_vrr = true` mouse movements in any direction except right are fine. However moving right which is the direction of my second monitor causes spikes. I've also tested changing the position of my monitors and it moves the trouble direction as you'd expect. Clip below

https://github.com/user-attachments/assets/4364754c-005d-4873-b8e0-8f74918de797

This can be mitigated by leaving a large gap between monitors (0x0 and 2000x0 vs 0x0 and 1920x0), however this means that you have to move the mouse fast to move between monitors normally and also means that you still get spikes when moving the mouse fast enough to that side. So this leads me to believe that the mouse is leaving the game and moving to the other monitor briefly which is what's causing the spikes in this case.

Also I don't know if this is supposed to be the case for `no_break_fs_vrr = false` on this pr or on the latest git, but with that disabled my refresh rate sits at 144 when moving the mouse at all in fps games despite it being invisible anyway.

Gwenodai:
> Also I don't know if this is supposed to be the case for no_break_fs_vrr = false on this pr or on the latest git, but with that disabled my refresh rate sits at 144 when moving the mouse at all in fps games despite it being invisible anyway.

Ok after a quick test I can confirm this is just how Hyprland functions now on the latest git no matter the state of `no_break_fs_vrr`. On the latest git no matter what settings used Hyprland will sit at max refresh rate when moving the mouse in fps games. On this PR, `no_break_fs_vrr=true` fixes this with the added issues of the second clip I explained in my previous comment.

> This PR shouldn't change the vrrtest behaviour.
> Still can't detect any spikes when clicking. Does this happen with mouse or with touchpad?

Just to clarify, the mouse button spikes [Myned](https://github.com/Myned) and I mentioned have been a thing in all Hyprland versions I've ever used. I don't think this PR has anything to do with that. Recent Hyprland versions prior to this PR just don't work with vrr at all when looking around in games so that might have made it harder to notice the mouse button thing between 0.41.0 and this PR where vrr had completely broken.

UjinT34:
Does the spike happen on mouse down, mouse up or both? Can't test it myself, my osd doesn't refresh itself and I don't see such short spikes.
This PR requires `cursor:no_break_fs_vrr = true`.
LFC might work incorrectly for some setups. This is a hw or driver issue. My monitor advertises 48hz min refresh but starts doing weird stuff when fps goes below 52.

Gwenodai:
> Does the spike happen on mouse down, mouse up or both?

I can test this tomorrow. But I'm not sure considering it doesn't happen on every single input as you can see from the video I posted. Which is just really odd.

Myned:
The spike occurs on mouse down only, for barely a millisecond as indicated above. Movement while mouse down does not incur additional spikes. If I had to guess it could've been a bind, but it still happens with all binds and window rules excluded from the config. I also tested without building with hyprbars to the same effect.

I cannot test the multi-monitor issues as I only have one display.

Additionally, my entire system freezes occasionally when VRR is active. It just happened recently while testing in vrrtest. There are no logs and I cannot SSH into the machine nor switch to a different TTY.

<details>
<summary>hyprctl systeminfo</summary>

```
Hyprland, built from branch  at commit f412e11aa530650a37080edd834dcec7fd7b38b5  ().
Date: 2024-07-15
Tag: , commits: f412e11aa530650a37080edd834dcec7fd7b38b5

flags: (if any)


System Information:
System name: Linux
Node name: mynix
Release: 6.9.8-zen1
Version: #1-NixOS ZEN SMP PREEMPT_DYNAMIC Tue Jan  1 00:00:00 UTC 1980


GPU information: 
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M/6850M XT] [1002:73df] (rev c0) (prog-if 00 [VGA controller])


os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.11.20240714.693bc46"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.11 (Vicuna)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.11 (Vicuna)"
VERSION_CODENAME=vicuna
VERSION_ID="24.11"


plugins:
  hyprbars by Vaxry ver 1.0
```

</summary>

Myned:
Some additional testing in vrrtest:
- Dispatcher does not spike via `watch -n 1 'hyprctl dispatch sendshortcut , mouse:272, class:love'`
- Hardware scroll wheel does not spike in either direction (but middle mouse still does)

Atemu:
I can reproduce the mouse click issue but only in some games and combinations of windowed/fullscreen.

***

What's worse though is that VRR doesn't appear to work properly without inputs to begin with:

When I run vrrtest at 139 fps, actual vfreq is 144 and stutters ~twice a second.

When I run vrrtest at 130 fps, actual vfreq jumps between 144~130 and sometimes dips far below into the 80s-90s, quite stuttery overall.

When I run vrrtest at 75 fps, actual vfreq appears to be quite stable but actually jumps around very often (causing stutter) and results in jumps to 90~100 in vfreq from time to time.

(Vsync on and AMDGPU in the VR power profile (4).)

***

BTW, if you have an LG monitor, the debug menu has a more frequent vfreq display without as much averaging that also stays on screen permanently via the service menu. Power off, left, left, left, right, power on and then settings, debug info.

Other monitors may have similar service menus.

Myned:
@Atemu 

> but only in some games and combinations of windowed/fullscreen.

If you're using `vrr = 1`, there are still many events that can interrupt the stability of VRR in windowed mode outside of input, as [commented](https://github.com/hyprwm/Hyprland/issues/5918#issuecomment-2190768835). This can cause issues without proper direct scanout, if I understand correctly, and is worth revisiting after the wlroots migration. If `vrr = 2`, you wouldn't have VRR active without fullscreen, which would cause stutter. I'd suggest testing only with `vrr = 2` for the purposes of this PR.

Gwenodai:
It's definitely on mouse down for me too. Oddly enough, I found a ["fix" for this back in 0.40.0](https://github.com/hyprwm/Hyprland/issues/5918#issuecomment-2182973284) but that doesn't work on this PR. I'm not sure when that changed as I've been sticking with 0.40.0 since it's the last good vrr build I've found on my testing.

> What's worse though is that VRR doesn't appear to work properly without inputs to begin with:
> 
> When I run vrrtest at 139 fps, actual vfreq is 144 and stutters ~twice a second.
> 
> When I run vrrtest at 130 fps, actual vfreq jumps between 144~130 and sometimes dips far below into the 80s-90s, quite stuttery overall.
> 
> When I run vrrtest at 75 fps, actual vfreq appears to be quite stable but actually jumps around very often (causing stutter) and results in jumps to 90~100 in vfreq from time to time.

I can't reproduce this at all. My vfreq on all values in vrr test and other applications is within +-1hz from the reported fps. The only change is below 51fps when LFC kicks in, though oddly enough LFC is supposed to be 48 but idk.

Myned:
Also, in most games, VSync adds additional overhead without benefit for adaptive refresh, and vrrtest runs by default with it disabled. If there's no difference, it's usually better to disable it.

Myned:
@Gwenodai 

Your linked comments sound like fullscreen wasn't being enabled. Did you confirm that VRR is active while the spikes do not occur? A decent way to visually see whether the window is fullscreened properly is with reserved space such as overlays (e.g. waybar). If the overlay is not hidden, the window isn't in true fullscreen.

Gwenodai:
@Myned

Yes I can confirm fullscreen was enabled and that vrr was functioning in both instances. I limit the fps to 60 when testing to make it very easy to tell if spikes occur both visually and by seeing the vfreq go from 2 digits to 3 digits. I have waybar at the top of my screen, it is not visible during fullscreen. On top of that, the testing to confirm that was done with the help of hyprctl's `activewindow` output which confirms that in all cases the application was in proper fullscreen.

I have no idea why `"floating": true` has any effect on a fullscreen window, but it is consistent and reproducible on 0.40.0 for me. In fact I'll include a video of it to demonstrate it.

https://github.com/user-attachments/assets/552f2d0f-d3ab-458c-a8f9-2a0dd04848c2

This video was taken with `vfr = 1` and `vrr = 2` on Hyprland 0.40.0 with vrrtest set to 87fps. The first and third tests have the properties of  `"floating": true`, `"fullscreen": true`, `"fullscreenMode": 0`. The second one has the properties of `"floating": false`, `"fullscreen": true`, `"fullscreenMode": 0`.

As you can see this bug is completely eliminated under these circumstances on 0.40.0. This however does not prevent it on this PR, so something has changed since then. This however really shouldn't be necessary and makes no sense as to why it works. But the weird thing is that it does and because of that I've written a script that checks for floating on fullscreen game windows and fixes it for me, essentially eliminating the bug in a hacky behind the scenes way for me. However since discovering this back on June 22nd I've not been effected by the spikes again until trying this new PR.

UjinT34:
Do mouse clicks cause spikes with `input:follow_mouse = 3`?

Myned:
> Do mouse clicks cause spikes with input:follow_mouse = 3?

They do not! Looks like it's related to focus as `cyclenext` while vrrtest is the only window also triggers the spike.

I can also confirm that `follow_mouse = 3` works on v0.40.0 as I was testing the floating fix mentioned, without being able to reproduce.

Gwenodai:
> Do mouse clicks cause spikes with `input:follow_mouse = 3`?

No they do not. So it appears mouse buttons must be doing something like shifting the focus or refocusing for a single frame. Then why does this behaviour not happen on fullscreen applications in 0.40.0 when `"floating": false`?

For reference I use `follow_mouse = 1`

UjinT34:
Looks like these spikes are caused by `refocus` or bringing a floating window to top.

Gwenodai:
@UjinT34

Then wouldn't `mouse_refocus = false` solve the issue with `follow_mouse = 1`? Because it seems to have no effect for me.

Myned:
My bad, it does indeed spike on `alterzorder`. I had it bound to the same key as `cyclenext`.

Myned:
@Gwenodai 

As for `floating: 0` not causing spikes, I believe that's due to some configuration differences as I wasn't able to reproduce myself. Best guess would be this rule from your posted config, which would mean tiled windows might paint differently than floating ones:

```
windowrulev2 = bordercolor rgba(00fbff85),fullscreen:1
```

Gwenodai:
@Myned

It may well be something in my config, but it's definitely not that as that only changes the border color of maximized windows not fullscreened ones and removing that has no effect.

The odd thing is nothing in my config changes between Hyprland versions outside of `WLR_NO_HARDWARE_CURSORS`, `default_cursor_monitor = DP-1`, `cursor_inactive_timeout = 0` being removed from my config and these being added to cursor: `no_hardware_cursors = true`, `inactive_timeout = 0`, `default_monitor = DP-1`, `no_break_fs_vrr = true`, `min_refresh_rate = 48`

Myned:
It's likely because tiled windows don't use z-order, thus can't be brought to the top on click, though that doesn't explain why `floating: 0` doesn't work in recent versions or why I can't reproduce.

Gwenodai:
> though that doesn't explain why `floating: 0` doesn't work in recent versions or why I can't reproduce.

That's really odd to me. Both my partner and I have completely different hardware from each other (Mobo, cpu, gpu, ram, monitors, mouse, keyboard, etc) as well as similar but different software setups to each other. We're both running Fedora 40, different mesa driver versions, completely different kernels (6.9.6-200_tkg_zen2_preempt.fc40.x86_64 vs 6.8.10-300.fc40.x86_64) and a bunch of different applications though her Hyprland config is almost the same as mine. But despite that all of this is reproducible on her PC too. So surely it can't be a hardware thing then.

UjinT34:
The code doesn't really checks the floating state, it checks `g_pSeatManager->state.pointerFocus` variable. I didn't bother to look what really sets its value. `changeWindowZOrder` calls `damageMonitor` which schedules an additional frame even when the order doesn't really change. I've made it return early if the window is already in the correct position. Please, verify that spikes on mouse click are gone and several floating windows have no trouble changing their order when clicked.

Myned:
Seems fixed with the latest commit, no spiking on click for vrrtest nor GW2, and multiple floating windows were changing z-order just fine via `follow_mouse = 1` on click and an `alterzorder, top` bind. I'll roll with this PR for a while to stability test.

Gwenodai:
That latest commit is amazing! The spiking is completely gone from mouse button clicks now. Now the only issue is the one I mentioned [here](https://github.com/hyprwm/Hyprland/pull/6877#issuecomment-2230423256) to do with multiple monitors and cursor constraints.

UjinT34:
> Now the only issue is the one I mentioned [here](https://github.com/hyprwm/Hyprland/pull/6877#issuecomment-2230423256) to do with multiple monitors and cursor constraints.

Does latest commit fix this issue? Is there any issues with switching focus between monitors?

Gwenodai:
> Does latest commit fix this issue? Is there any issues with switching focus between monitors?

There's no issues switching focuses between monitors that I noticed. It's just the issue I highlighted prior, which is basically:

Moving the mouse in the direction of my second monitor causes spikes. This can be mitigated by leaving a large gap between monitors (0x0 and 2000x0 vs 0x0 and 1920x0), however this presents it's own issues and you'd not want to use a computer like this. This also means that you still get spikes when moving the mouse fast enough to that side. So this leads me to believe that the mouse is leaving the game and moving to the other monitor briefly which is what's causing the spikes in this case.

Gwenodai:
@UjinT34

Sadly the latest commit doesn't seem to have any effect on the bug whatsoever. There are also no issues switching focus between monitors that I can find.

UjinT34:
@Gwenodai
How about now? If this doesn't help then it should be a separate issue.

Gwenodai:
@UjinT34

That fixed it completely! I haven't found any other issues either. I'm gonna stick with this PR for a bit and see how things go.

UjinT34:
@vaxerski please take a look at this PR

Atemu:
VRR with the cursor being moved appears to have regressed for me.

If the cursor is visible, I'm at max refresh rate again.

When the cursor is not visible, it sort of works but somehow only when I'm not moving it very quickly. Flinging the camera around nets me max refresh rate regardless of the frame rare.

Clicking appears to work correctly though now.

UjinT34:
> If the cursor is visible, I'm at max refresh rate again.
> 
> When the cursor is not visible, it sort of works but somehow only when I'm not moving it very quickly. Flinging the camera around nets me max refresh rate regardless of the frame rare.

Looks like you have hw cursors enabled.
Multi-monitor? What games have you tested?

Gwenodai:
> If the cursor is visible, I'm at max refresh rate again.
> 
> When the cursor is not visible, it sort of works but somehow only when I'm not moving it very quickly. Flinging the camera around nets me max refresh rate regardless of the frame rare.

To add onto what [UjinT34](https://github.com/UjinT34) said, also make sure `no_break_fs_vrr` is enabled.

Atemu:
`no_hardware_cursors = true`. All other prerequisites are set correctly too AFAICT:

```
misc {
    force_default_wallpaper = 1 # Set to 0 or 1 to disable the anime mascot wallpapers
    disable_hyprland_logo = true # If true disables the random hyprland logo / anime girl background. :(

    # I'd like my screen to turn on when I press a button please
    mouse_move_enables_dpms = true
    key_press_enables_dpms = true

    # Variable frame rate?
    # TODO need to test whether this affects input delay
    vfr = true
    vrr = 2

    no_direct_scanout = true
}

cursor {
    no_hardware_cursors = true
    hide_on_key_press = false
    no_break_fs_vrr = true
    min_refresh_rate = 10
}
```

I've also toggled HW cursors to no avail.

As mentioned, it does partially work too when the mouse cursor is not visible.

Gwenodai:
I've discovered a strange bug with mouse movements in games. If I move my mouse very slowly it doesn't seem to transfer all of the movement into the game, some frames seem to ignore mouse movement alltogether when I move the mouse slowly. Disabling `no_break_fs_vrr` (thereby essentially disabling this entire PR) "fixes" this issue. So there's an odd effect on mouse movement being impacted/ignored during slow movements.

UjinT34:
@Gwenodai
Should be fixed now

Myned:
@Atemu 

Do you have `WLR_NO_HARDWARE_CURSORS` in your environment? Definitely sounds like hardware cursors. Also make sure to do a full session restart or system reboot, if you haven't already.

@UjinT34 

Was able to reproduce the ignoring of slow mouse movement when the cursor is locked @ https://github.com/hyprwm/Hyprland/pull/6877/commits/658e79e0d340765b62e60f82c4139ee06fc46b7e, tis fixed @ https://github.com/hyprwm/Hyprland/pull/6877/commits/b325df22dfc8ef54bbb9619bb98ff5fd56030a7c (latest)

Gwenodai:
@UjinT34

Sadly something has completely broken now for me with this new commit.

Moving the camera in games now brings the refresh rate up to the max again (But not moving a visible mouse). But along with this I've now got new bugs that weren't present before.

The random spikes are back and now there's even more weirdness with them. Locking it to 60fps results in random spikes to 144 every few seconds, whereas locking it to something larger like 89fps results in completely random spikes all over the place constantly. This behavior appears in both vrrtest and games for me. I've attached a video below that demonstrates the 60fps and 89fps comparison. (I only set it to 89fps because I just held the up arrow for a bit, the issue appears on other framerates as well.)

https://github.com/user-attachments/assets/c1fc2c06-e8ed-4d3f-9cec-5a9df883bcb3

Gwenodai:
Ok more odities.

The spikes in vrrtest only seem to happen when there's a game open on another workspace (The workspace is hidden/not visible). This doesn't happen when vrrtest is the only fullscreen application open, and also oddly enough doesn't happen when I fullscreen other applications in the background. (The games I tested with are Monster Hunter World and Warhammer 40,000: Boltgun). Without a game open in the background it works perfectly.

Moving the camera with the mouse in Monster Hunter World causes the refresh rate to sit at max during said mouse movements, however moving the camera in Warhammer 40,000: Boltgun does not cause this issue for some reason.

Gwenodai:
Ok so the refresh rate that vrrtest is jumping up to seems to be the framerate Monster Hunter World was running at in the other workspace, so that explains why it didn't happen with other fullscreen applications. Something weird is going on and it's mixing together the two refresh rates. `no_direct_scanout` also seems to have no effect on this.

This might actually be happening on https://github.com/hyprwm/Hyprland/commit/658e79e0d340765b62e60f82c4139ee06fc46b7e as well, I just didn't test Monster Hunter World, nor did I test multiple vrr applications at different refresh rates on other workspaces simultaneously before.

UjinT34:
@Gwenodai
b325df2 is very unlikely to cause any trouble
658e79e might be a regression with multi monitor setup

I don't have any trouble with vrrtest and a game on separate workspaces. Having them on the same workspace breaks vrr. But why anyone would want that?) This issue is outside the scope of this PR. If you have several fullscreen apps on a single workspace and switch between them the focus might get stuck with the wrong app and fullscreen status might get lost. Most games (especially online) keep rendering in the bg. I don't know how it'll affect other apps.

Myned:
I can reproduce as well with GW2 and vrrtest on separate workspaces:
- GW2 fullscreen on workspace 1, mangohud limited to 100 (max refresh), vsync off, internal limiter set to Unlimited
- vrrtest fullscreen on workspace 2, target fps at 60
- Observe stutter/flickering as the refresh rate fluctuates on vrrtest
- Limit GW2 to 60 fps
- Observe no stutter/flickering on vrrtest

I do agree, however, that this is probably a different issue to track as it is not related to input.

Gwenodai:
Yes I'd assume the new issues I mentioned and [Myned](https://github.com/Myned) reproduced might be unrelated to input and a different issue entirely. And just to be clear I was talking about apps on other workspaces entirely not on the same one.

However for some reason I'm still able to reproduce mouse movements causing spikes in MHW. I'm not sure why I can in that game but not others, but that might be something else to look into. Maybe there's another way games handle locking input which at least MHW uses that's causing this issue to reappear there.

Clip of it below taken on the latest commit.

https://github.com/user-attachments/assets/bc3cd486-a91d-4ee4-b0d0-48b187c8489e

Gwenodai:
Also to add onto my last comment, I've just confirmed that the refresh rate does not spike up during mouse movements in MHW back on 0.40.0 prior to `cursor:no_break_fs_vrr`. So I assume that does have something to do with input and this PR.

UjinT34:
Reproduced the issue with GW2. Even the launcher is enough to break VRR. Seems to be limited to specific apps. Doesn't happen with vrrtest + quake 2 + noita. As soon as I start GW2 VRR breaks for all other apps.

Myned:
The fact that it's only some games and that GW2 is included makes me think this is related to my [comment](https://github.com/hyprwm/Hyprland/issues/5918#issuecomment-2227934967) about frame accumulation in the background. I'm interested to see how aquamarine handles that once it's in main.

Atemu (approved):
After a restart it appears to be working now; odd that it didn't work before. It feels super smooth and works really well. Thank you so much!

It still feels like sometimes the refresh rate doesn't match for a split second when frame rate drops slightly but I couldn't reproduce that in vrrtest and it's generally getting to the point where it's hard to noticeably see these imperfections that I think I feel in-game. You'd need specialised hardware equipment to do any further QA from my perspective.

I can't comment on code quality but the user experience gets an approval from me.

This does not need to be addressed in this PR but one minor issue I noticed is that LFC does not apply to the cursor. This is trivial to reproduce in vrrtest by setting the target framerate extremely low (i.e. 5 fps). The monitor is being fed at 70-80Hz but the cursor is extremely choppy (likely the 5fps). Fixing this would obviate the need for `cursor:min_refresh_rate` because LFC would ensure that the cursor is always at the smoothest rate possible.

Another thing that needs to happen is for the required settings to be documented as you do need to tweak some magical knobs for VRR to work correctly and those aren't at all obvious.

Here's what I've gathered so far:

| Setting                      | Requirements/Effect                                                                 |
|------------------------------|-------------------------------------------------------------------------------------|
| `misc:vrr`                   | `1` or `2` or else VRR does not work, cannot be disabled at runtime                 |
| `misc:vfr`                   | `true` or else `misc:vrr` has no effect                                             |
| `cursor:no_break_fs_vrr`     | `true` or else VRR will lock to max refresh when moving the mouse                   |
| `cursor:no_hardware_cursors` | `true` or else `cursor:no_break_fs_vrr` has no effect                               |
| `cursor:min_refresh_rate`    | (Optional) An integer fps value below which the cursor will force its own refreshes |

Does `misc:no_direct_scanout` have any effect on VRR?

`cursor:no_break_fs_vrr` should probably imply `cursor:no_hardware_cursors` (or the other way around); it doesn't make sense to have one but not the other (outside of debugging).  
Similar story with `misc:vfr`: If `misc:vrr` cannot function without it, it should be enabled if `misc:vrr` is enabled. Though I believe `misc:vrr = 0` + `misc:vfr = true` does have some use?

Atemu:
Oh, I did not see the force-push in time, I will test that it appears to use an entirely different backend now instead of wlroots? O.o

Myned:
@Atemu 

> It still feels like sometimes the refresh rate doesn't match for a split second when frame rate drops slightly but I couldn't reproduce that in vrrtest

I've noticed this as well, even in vrrtest, but I chalked it up to the current situation with wlroots. Will retest on aquamarine.

> Another thing that needs to happen is for the required settings to be documented

Thanks for the writeup, I agree it should be documented, perhaps in a wiki PR once commits have settled in this area. Only thing to note is that `vrr = 2` can be toggled while a session is running (it's how I tested), but it only takes effect when switching to another workspace and back. I can't comment on the rest myself.

Atemu (approved):
Okay that's quite incredible, everything still works. Super smooth. Certainly a big step up to Xorg which is what I've been using up until today. I didn't notice any slight mismatches either but I'll keep an eye out.

I did not expect a migration of the rendering backend to go this smoothly. It actually fixed a few outstanding issues aswell. I'm impressed.

UjinT34:
LFC repeats the same frame to keep monitor refresh rate within supported range. The content still refreshes at low rate when LFC is active. `cursor:min_refresh_rate` will force a minimum content refresh rate when mouse moves. This forced rate is not guaranteed to be smooth. 

Settings are documented https://wiki.hyprland.org/Configuring/Variables/#cursor All settings can be changed at runtime (except one specific change https://github.com/hyprwm/Hyprland/issues/6893)

Direct scanout requires HW cursors. With `cursor:no_hardware_cursors = true` ds setting has no effect. With `cursor:no_hardware_cursors = false` `cursor:no_break_fs_vrr = true` will only work for locked cursors with this PR. I'll look more into hw cursors if this PR gets merged.

Atemu:
> LFC repeats the same frame

I don't think it's that simple. I never had my cursor drop below ~60 fps with LFC on Xorg or on any Windows implementation of VRR.

Or is that behaviour due to usage of a HW cursor?

Gwenodai:
> > LFC repeats the same frame
> 
> I don't think it's that simple. I never had my cursor drop below ~60 fps with LFC on Xorg or on any Windows implementation of VRR.
> 
> Or is that behaviour due to usage of a HW cursor?

It really is that simple. LFC is done entirely on your monitor, not on the computer. It can't move your mouse in between frames and render it separately from the given frame. 

LFC actually works by multiplying the frames the monitor gets when it's below the lower limit of your display. It does this by doubling, tripping, etc the current frame based on your current frame rate in order to get it above your minimum. You may see 30fps become 90hz (3x) or 46fps become 92hz (2x) depending on how the monitor decides to multiply your framerate at its current number.

Atemu:
Well, evidently, it is not that simple as the cursor updates far more frequently than the synced content in other systems. You can easily try this out yourself using vrrtest on Xorg. At 5 fps, my mouse is at a perfectly smooth 120Hz and the bars stutter along at a constant rate with no tearing.

I also don't believe the monitor plays any role in this as they're generally quite dumb. Even VRR itself is little more than the graphics card varying the front porch of the output signal sent to the monitor AFAIK. The fact that [the same code path used for VRR on modern displays can be made to work with analogue CRTs](https://forums.blurbusters.com/viewtopic.php?t=3234) should be telling.  
What you're describing sounds more like PSR (panel self refresh) which is something entirely different (though I guess it could be used for LFC in theory?).

I know how LFC works conceptually, I'm just unsure of the concrete implementation details.

Myned:
Retested @ https://github.com/hyprwm/Hyprland/pull/6877/commits/7dc595ec4100be4fbff099508cddb0c576efdb6e, seems to be working the same as before aquamarine in both vrrtest and GW2. The occasional stutter is still present, but that's likely due to external factors outside of Hyprland's control, such as hardware.

@Atemu 

> Or is that behaviour due to usage of a HW cursor?

From my understanding, the difference in behavior you're seeing is indeed because of hardware cursors, due to them being on a separate plane than the main framebuffer, thus can be updated independently at a multiple of the main framerate (LFC).

X11 and Windows may have the ability to update cursors as a multiple of refresh without triggering a page flip (which discards the previous frame), but it seems that Wayland does not currently have an API path to do so, though I was unable to confirm this. [zamundaaa](https://invent.kde.org/plasma/kwin/-/issues/85#note_463762) from KDE explains it better than I can - by "AMS API" I believe they mean AMD, so FreeSync only.

`cursor:min_refresh_rate` is needed for software cursors because it's a single framebuffer for both cursor and application, which can be under minimum refresh, and is essentially a workaround for cursor movement not refreshing the display, but is not dynamic like with LFC and hardware cursors.

According to a [thread](https://invent.kde.org/plasma/kwin/-/issues/85#note_463853) in that issue, a type of pseudo-LFC could theoretically be implemented by a compositor, where the frame could be repeated at the highest supported multiple with an updated cursor position instead of relying on the monitor's handling below minimum refresh, which would also work for software cursors, but this was not expanded upon further from what I could find.

Disclaimer: I am paraphrasing quite a lot here, but I tried to compile as much info as possible since LFC is such a misunderstood topic in the general VRR space (and possibly by myself).

@UjinT34 

If you haven't seen it already, [KWin](https://invent.kde.org/plasma/kwin/-/issues/194) has some additional context that may be worth looking at, and some potential options for working around hardware cursors [here](https://invent.kde.org/plasma/kwin/-/issues/85).

Atemu:
> [zamundaaa](https://invent.kde.org/plasma/kwin/-/issues/85#note_463762) from KDE explains it better than I can - by "AMS API" I believe they mean AMD, so FreeSync only.

AMS likely means atomic mode setting in this context.

The thread linked is very helpful because they had to solve the exact same issue at hand here and the currently chose workaround was also presented as one of the options.

Atemu:
The newest version does not appear as smooth as the previous iteration anymore. When moving the camera around in Planetside 2 at a locked 7.2ms (Mangohud shows variation of 7.1~7.2), refresh rate often jumps between 137Hz and 143Hz though it appears to primarily focus around 138.something and 139.something. It feels off too, this is not an accurate representation of the frame rate.

It's not easy to tell in vrrtest but if you make it fluctuate a bit (I chose 120 min 139 max), you can tell it judders a bit with a frequency of roughly ~4Hz but it's not constant. Super hard to tell reproducibly but irritating in practice.

Atemu:
I tried the previous version again (7dc595ec4100be4fbff099508cddb0c576efdb6e) and it does indeed feel better, so I wasn't imagining it.

However, it also shows the issue I noticed on the newer version with the refresh rate jumping a bit. Though it appears not quite as much? It's that hard to tell.

The small judder is not visible in vrrtest however (or at least not to a noticeable degree).

Myned:
vrrtest and GW2 @ 3d3955f0720c8d18ba7be1ff92e919610c974019

I don't notice anything different with software cursors myself and rapid fluctuations would result in pretty noticeable flicker on my monitor.

With hardware cursors it still fluctuates on cursor movement, but very minorly and only in GW2. When the cursor is locked there is no noticeable difference to software cursors, so that's great.

Atemu:
A merge conflict arose btw.

Gwenodai:
Is this still being worked on? I miss having reliable vrr in hyprland without having to swap between versions depending on what input device I want to use.

UjinT34:
This is ready to be merged if there no other way to get rid of unwanted `_CWlSurfaceCommit` calls. Still might not work for every game https://github.com/hyprwm/Hyprland/discussions/7386

Cloudperry:
Tested this PR to see how games work with it. Quake champions worked well, but Apex legends was very laggy with this PR compared to main branch. I can't really check if vrr is working, because my monitor doesn't have an fps overlay, but I didn't notice any tearing with allow_tearing on. I can maybe add some logs or other details about the Apex legends issue later. Basically the game is running well (Mangohud frametimes), but appears extremely stuttery and possibly shows out of order frames.

I also checked if this happens to fix https://github.com/hyprwm/Hyprland/issues/7414, but it didn't.

Atemu:
@Cloudperry Have you ensured you're not running into https://gitlab.freedesktop.org/drm/amd/-/issues/1500?

Also look up whether there's a debug/service menu for your display which shows this info. I believe all of them do for QA purposes but it's not always documented (usually unofficially).

Cloudperry:
> @Cloudperry Have you ensured you're not running into [gitlab.freedesktop.org/drm/amd/-/issues/1500](https://gitlab.freedesktop.org/drm/amd/-/issues/1500)?
> 
> Also look up whether there's a debug/service menu for your display which shows this info. I believe all of them do for QA purposes but it's not always documented (usually unofficially).

I have ran into that quirk before and fixed it, but actually that was the cause. My Corectrl autostart was somehow broken. Now VRR is working really nicely (no tearing when allow_tearing is on) with games capped to 142 fps using Mangohud. On main branch with no_break_fs_vrr I had occasional tearing with the same games, but that testing was with a 143 fps cap.

Keksmo:
when can we expect this to be merged? I would really love to see vrr working on the current build

vaxerski:
idk it's still a PoC

Atemu:
FWIW, this makes VRR work much better than it did before; it's quite essential honestly.

vaxerski:
@UjinT34

Atemu:
I finally found a way to spot the imperfections I've been feeling: Set vrrtest to the lowest speed.

The stutters become immediately obvious.

This needs a rebase btw. I've done it myself at https://github.com/Atemu/Hyprland/tree/vrr-new-rebased-rebased, resolving conflicts to the best of my ability but I have no way of verifying whether the changes still make sense in integration with the rest of the codebase. It appears to work just like before. (It's atop v43.0 but applies cleanly to 44.0.)

UjinT34:
I don't think there is a better way to fix it. Looks like something similar is required for HW cursors in general.

Ready for merging

Atemu:
Sorry, I'm not following.

HW cursors are disabled in my case and it's not visible on screen. I'll have to re-test whether this is the status quo in which case we can move the discussion to another issue.

@Gwenodai do you also notice the stutters in low-speed vrrtest?

src/managers/input/InputManager.cpp
```diff
@@ -87,6 +87,11 @@ void CInputManager::onMouseMoved(IPointer::SMotionEvent e) {
 
     const auto  DELTA = *PNOACCEL == 1 ? e.unaccel : e.delta;
 
+    if (g_pSeatManager->isPointerFrameSkipped)
```
vaxerski, line 90: am I blind or are the stored values never used

src/events/Windows.cpp
```diff
@@ -758,16 +759,28 @@ void Events::listener_commitWindow(void* owner, void* data) {
...
+
+    PMONITOR->debugLastPresentation(g_pSeatManager->isPointerFrameCommit ? "listener_commitWindow skip" : "listener_commitWindow");
+    if (g_pSeatManager->isPointerFrameCommit) {
+        g_pSeatManager->isPointerFrameSkipped = false;
+        g_pSeatManager->isPointerFrameCommit  = false;
+    } else
+        g_pHyprRenderer->damageSurface(PWINDOW->m_pWLSurface->resource(), PWINDOW->m_vRealPosition.goal().x, PWINDOW->m_vRealPosition.goal().y,
+                                       PWINDOW->m_bIsX11 ? 1.0 / PWINDOW->m_fX11SurfaceScaledBy : 1.0);
+
+    if (g_pSeatManager->isPointerFrameSkipped) {
```
vaxerski, line 772: "frame" refers to a monitor frame? In that case renderer would be a better place for it? "frame" in seat means wl_pointer.frame

src/helpers/Timer.hpp
```diff
@@ -6,7 +6,7 @@ class CTimer {
   public:
     void                                         reset();
     float                                        getSeconds();
-    long                                         getMillis();
+    float                                        getMillis();
```
vaxerski, line 9: it'd be better to just make a new `getMicros` or `getMillisFloat`

src/managers/PointerManager.cpp
```diff
@@ -676,8 +678,11 @@ void CPointerManager::warpTo(const Vector2D& logical) {
     damageIfSoftware();
 
     pointerPos = closestValid(logical);
-    recheckEnteredOutputs();
-    onCursorMoved();
+
+    if (!g_pInputManager->isLocked()) {
```
vaxerski, line 682: what if we are changing a workspace? Wouldn't that lock our cursor to the current window? Why this change?

src/managers/input/InputManager.cpp
UjinT34: listener_commitWindow sends them

src/events/Windows.cpp
UjinT34: `isPointerFrameSkipped` refers to `g_pSeatManager->sendPointerFrame();`

src/helpers/Timer.hpp
UjinT34: It is converted to float in every place it is used anyway. And all of those places benefit from better precision.

src/managers/PointerManager.cpp
UjinT34: Fast movements can escape the lock briefly and will trigger active monitor switch causing unnecessary frames and frame rate spikes https://github.com/hyprwm/Hyprland/pull/6877#issuecomment-2230423256

UjinT34: Changing a workspace doesn't seem to have any issues with this

vaxerski: ok

src/events/Windows.cpp
vaxerski: ok good then

src/managers/input/InputManager.cpp
vaxerski: right.

vaxerski (approved):
sure, let's roll

Gwenodai:
@Atemu I didn't notice the stutters from my short testing of it. I was just giving a thumbs up for the rebasing and such. Sorry for the confusion 😅

Issue #6874, opened 2024-07-13
Broken workspace rules


diniamo:
### Regression?

Yes

### System Info and Version

The last working commit is b03f41efec14273cf25c42d4cef326acc36cb319, meaning a443902abca6549839910720b80bd01faf10f387 broke it.

### Description

I have these 2 workspace rules in my config:
```
workspace=w[v1] s[false], gapsin:0, gapsout:0, border:false, shadow:false, rounding:false, decorate:false
workspace=f[1], gapsin:0, gapsout:0, border:false, shadow:false, rounding:false, decorate:false
```

And I noticed 2 issues:
1. After opening 2 windows in a workspace, the first one retains the properties of the first workspace rule (it has no border and no rounding)
2. When I have 2 windows, maximize one of them, only the gaps disappear, the border and rounding stay

### How to reproduce

1. Have the 2 workspace rules
2. Follow the issues above

diniamo:
@MightyPlaza - in case you care

MightyPlaza:
PR above should fix the regressions
however with your rules, floating the only window in a workspace will result in it not getting updated later on when another window gets open which is not a regression from that commit

diniamo:
I know, I already have another issue open for that from a while ago: #5467 

Thanks for the PR.

PR #6878, opened 2024-07-14
update workspace rules on layout change

MightyPlaza:
updates workspace rules on layout changes for workspace rules that depend on layout state like window count or fullscreen state

should fix: https://github.com/hyprwm/Hyprland/issues/6874


vaxerski (approved):
thanks!

PR #6903, opened 2024-07-16
Make shader time always count from zero

PartyWumpus:

At the moment, a shader's `time` uniform is initially set to whatever the current value of `m_tGlobalTimer` is. This means that any shaders that are enabled after hyprland starts have no way of knowing how long it has been since they were enabled (this only applies to the user defined screen shader as far as I can tell).

This PR fixes this by storing the initial time and setting the `time` uniform to `time - initialTime`, so a shader's first `time` value is zero (and then counts up normally from there).

This only sets the initial time value for the screen shader (everything else defaulting to zero), as I assume nothing else actually needs it. This does mean the initial time could just be stored outside of the `CShader`, but it's easier this way.

It means that whenever the config is edited and the user has a screen shader active it will reset the screen shader's time back to zero instead of continuing roughly from where they were. If this is a problem then it could instead just expose an `initial_time` uniform to the shader but the only thing you could possibly do with it is just `float timer = time-initial_time;` so I think it's best not to, for ease of use. (the first commit does it this way)

Should be ready to merge, it works fine on my machine.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -861,6 +861,9 @@ void CHyprOpenGLImpl::applyScreenShader(const std::string& path) {
     m_sFinalScreenShader.proj      = glGetUniformLocation(m_sFinalScreenShader.program, "proj");
     m_sFinalScreenShader.tex       = glGetUniformLocation(m_sFinalScreenShader.program, "tex");
     m_sFinalScreenShader.time      = glGetUniformLocation(m_sFinalScreenShader.program, "time");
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
oke doke

vaxerski:
can you rebase on main and run clang-format to make the ci happy?

PartyWumpus:
Sure, done. I thought I already ran clang-format, but clearly I forgot.

vaxerski (approved):
coolio, thanks!

PR #6930, opened 2024-07-19
compositor: implement wayland socket handover

Ferdi265:

This PR implements the compositor side of the Wayland socket handover protocol as described in the [KDE Wiki](https://invent.kde.org/plasma/kwin/-/wikis/Restarting). The CLI options are chosen so that they are compatible with Kwin:

Hyprland now accepts the CLI options `--socket NAME` and `--wayland-fd FD`, which pass the name of the Wayland display socket and the Wayland display socket fd number respectively. If given, Hyprland will use these instead of finding a free Wayland display socket name and creating it.

This allows for Hyprland to be used with a helper tool that automatically restarts the compositor on crashes, and allowing clients to reconnect after the compositor restart and survive the crash. The reconnecting clients look like new clients to Hyprland and perform all the same steps as a fresh Wayland client would. No additional handling is required on the side of Hyprland.

The helper tool finds and creates the Wayland socket and launches Hyprland with the new CLI options, restarting it when it exits with a non-zero exit code. Two such helper tools currently exist as far as I know:

- [`kwin_wayland_wrapper`](https://invent.kde.org/plasma/kwin/-/blob/master/src/helpers/wayland_wrapper/kwin_wrapper.cpp?ref_type=heads) (only for KDE)
- [`wl-restart`](https://github.com/ferdi265/wl-restart) (developed by me)

This PR works well from my personal testing.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -27,6 +27,7 @@
 #include <hyprutils/string/String.hpp>
 using namespace Hyprutils::String;
 
+#include <fcntl.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/resource.h>
@@ -533,19 +534,28 @@ void CCompositor::prepareFallbackOutput() {
     wlr_headless_add_output(headless, 1920, 1080);
 }
 
-void CCompositor::startCompositor() {
+void CCompositor::startCompositor(std::string socketName, int socketFd) {
     initAllSignals();
 
-    // get socket, avoid using 0
-    for (int candidate = 1; candidate <= 32; candidate++) {
-        const auto CANDIDATESTR = ("wayland-" + std::to_string(candidate));
-        const auto RETVAL       = wl_display_add_socket(m_sWLDisplay, CANDIDATESTR.c_str());
+    if (!socketName.empty() && socketFd != -1) {
+        fcntl(socketFd, F_SETFD, FD_CLOEXEC);
+        const auto RETVAL = wl_display_add_socket_fd(m_sWLDisplay, socketFd);
         if (RETVAL >= 0) {
-            m_szWLDisplaySocket = CANDIDATESTR;
-            Debug::log(LOG, "wl_display_add_socket for {} succeeded with {}", CANDIDATESTR, RETVAL);
-            break;
-        } else {
-            Debug::log(WARN, "wl_display_add_socket for {} returned {}: skipping candidate {}", CANDIDATESTR, RETVAL, candidate);
+            m_szWLDisplaySocket = socketName;
+            Debug::log(LOG, "wl_display_add_socket_fd for {} succeeded with {}", socketName, RETVAL);
+        } else
+            Debug::log(WARN, "wl_display_add_socket_fd for {} returned {}: skipping", socketName, RETVAL);
+    } else {
+        // get socket, avoid using 0
+        for (int candidate = 1; candidate <= 32; candidate++) {
+            const auto CANDIDATESTR = ("wayland-" + std::to_string(candidate));
+            const auto RETVAL       = wl_display_add_socket(m_sWLDisplay, CANDIDATESTR.c_str());
+            if (RETVAL >= 0) {
+                m_szWLDisplaySocket = CANDIDATESTR;
+                Debug::log(LOG, "wl_display_add_socket for {} succeeded with {}", CANDIDATESTR, RETVAL);
+                break;
+            } else
+                Debug::log(WARN, "wl_display_add_socket for {} returned {}: skipping candidate {}", CANDIDATESTR, RETVAL, candidate);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main.cpp
```diff
@@ -113,7 +135,7 @@ int main(int argc, char** argv) {
     Debug::log(LOG, "Hyprland init finished.");
 
     // If all's good to go, start.
-    g_pCompositor->startCompositor();
+    g_pCompositor->startCompositor(socketName, socketFd);
```
vaxerski, line 138: since we require both to work, would be nice to print an error if only one is set

Ferdi265: Thanks, fixed!

vaxerski (dismissed):
thanks!

src/main.cpp
```diff
@@ -46,6 +50,24 @@ int main(int argc, char** argv) {
...
+            socketName = *std::next(it);
+            it++;
+        } else if (it->compare("--wayland-fd") == 0) {
+            if (std::next(it) == args.end()) {
+                help();
+
+                return 1;
+            }
+
+            socketFd = std::atoi(std::next(it)->c_str());
```
vaxerski, line 69: ah wait a sec, this can crash without a useful message. Wrap it in a try cath

Ferdi265: Good catch, I didn't think about error cases much when I wrote the original PoC and didn't get to cleaning this up yet. Though, I think a try catch is not sufficient here: `std::atoi()` returns 0 on invalid numbers and causes undefined behaviour if the number is out of range. I should have used `std::stoi()` which does throw an exception.

Ferdi265: Fixed. I also added a check that `socketFd` refers to a valid file descriptor, and isn't just a random number.

src/main.cpp
```diff
@@ -46,6 +51,37 @@ int main(int argc, char** argv) {
...
+                help();
+
+                return 1;
+            }
+
+            try {
+                socketFd = std::stoi(std::next(it)->c_str());
+
+                // check if socketFd is a valid file descriptor
+                if (fcntl(socketFd, F_GETFD) == -1) {
```
vaxerski, line 74: no {} around short ifs

src/Compositor.cpp
```diff
@@ -533,19 +534,30 @@ void CCompositor::prepareFallbackOutput() {
...
+    } else {
+        // get socket, avoid using 0
+        for (int candidate = 1; candidate <= 32; candidate++) {
+            const auto CANDIDATESTR = ("wayland-" + std::to_string(candidate));
+            const auto RETVAL       = wl_display_add_socket(m_sWLDisplay, CANDIDATESTR.c_str());
+            if (RETVAL >= 0) {
+                m_szWLDisplaySocket = CANDIDATESTR;
+                Debug::log(LOG, "wl_display_add_socket for {} succeeded with {}", CANDIDATESTR, RETVAL);
+                break;
+            } else {
```
vaxerski, line 558: no {}

src/Compositor.cpp
```diff
@@ -533,19 +534,30 @@ void CCompositor::prepareFallbackOutput() {
...
+        const auto RETVAL = wl_display_add_socket_fd(m_sWLDisplay, socketFd);
         if (RETVAL >= 0) {
-            m_szWLDisplaySocket = CANDIDATESTR;
-            Debug::log(LOG, "wl_display_add_socket for {} succeeded with {}", CANDIDATESTR, RETVAL);
-            break;
+            m_szWLDisplaySocket = socketName;
+            Debug::log(LOG, "wl_display_add_socket_fd for {} succeeded with {}", socketName, RETVAL);
         } else {
-            Debug::log(WARN, "wl_display_add_socket for {} returned {}: skipping candidate {}", CANDIDATESTR, RETVAL, candidate);
+            Debug::log(WARN, "wl_display_add_socket_fd for {} returned {}: skipping", socketName, RETVAL);
```
vaxerski, line 547: no {}

vaxerski (commented):
rest lgtm

src/main.cpp
Ferdi265: should be fixed

src/Compositor.cpp
Ferdi265: should be fixed

src/Compositor.cpp
Ferdi265: should be fixed

Ferdi265:
rebased onto latest main

vaxerski (approved):
thanks!

vaxerski:
oopsie, someone didn't run clang-format :P

Ferdi265:
> oopsie, someone didn't run clang-format :P

Ahh sorry, should be correct now.

Ferdi265:
also, cleaned up the history.

vaxerski (approved):

Issue #6375, opened 2024-06-08
XWayland game turns black after switching to another XWayland window.


BluewyDiamond:
### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 9994b73ad0af5c9ba5fc4058234cea77d0a1ddb1  (buffer: track asynchronous buffers and don't release them until unref).
Date: Sat Jun 8 17:27:56 2024
Tag: v0.40.0-176-g9994b73a, commits: 4782

flags: (if any)


System Information:
System name: Linux
Node name: CachyOS
Release: 6.9.3-4-cachyos-lto
Version: #1 SMP PREEMPT_DYNAMIC Sat, 08 Jun 2024 12:42:12 +0000


GPU information: 
2f:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 10 [Radeon RX 5600 OEM/5600 XT / 5700/5700 XT] [1002:731f] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="CachyOS Linux"
PRETTY_NAME="CachyOS"
ID=cachyos
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://cachyos.org/"
DOCUMENTATION_URL="https://wiki.cachyos.org/"
SUPPORT_URL="https://forum.cachyos.org/"
BUG_REPORT_URL="https://github.com/cachyos"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=cachyos


plugins:

======Config-Start======


Config File: /home/bluewy/.config/hypr/hyprland.conf: Read Succeeded
# -----------------------------------------------------
# Monitors
# -----------------------------------------------------
source = $HOME/.config/hypr/sections/monitors.conf

# -----------------------------------------------------
# Inputs
# -----------------------------------------------------
source = $HOME/.config/hypr/sections/inputs.conf

# -----------------------------------------------------
# Keybinds
# -----------------------------------------------------
source = $HOME/.config/hypr/sections/keybinds.conf

# -----------------------------------------------------
# Environment Variables
# -----------------------------------------------------
source = $HOME/.config/hypr/sections/env.conf

# -----------------------------------------------------
# Window Rules
# -----------------------------------------------------
source = $HOME/.config/hypr/sections/window_rules.conf

# -----------------------------------------------------
[...]

PR #6966, opened 2024-07-21
Fix xwayland black window bug after losing focus

UjinT34:
Sets and removes the withdrawn state for fullscreen xwayland apps. Fixes https://github.com/hyprwm/Hyprland/issues/6375

Might mess with withdrawn state from `handleUnmapNotify`.

Ready

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -378,6 +378,20 @@ void CXWM::handleFocusIn(xcb_focus_in_event_t* e) {
         focusWindow(focusedSurface.lock());
 }
 
+void CXWM::handleFocusOut(xcb_focus_out_event_t* e) {
+    Debug::log(TRACE, "[xwm] focusOut mode={}, detail={}, event={}", e->mode, e->detail, e->event);
+
+    const auto XSURF = windowForXID(e->event);
+
+    if (!XSURF)
+        return;
+
+    Debug::log(TRACE, "[xwm] focusOut for {} {} {} surface {}", XSURF->mapped ? "mapped" : "unmapped", XSURF->fullscreen ? "fullscreen" : "windowed",
+               XSURF == focusedSurface ? "focused" : "unfocused", XSURF->state.title);
+
+    // do something?
+}
+
 void CXWM::sendWMMessage(SP<CXWaylandSurface> surf, xcb_client_message_data_t* data, uint32_t mask) {
     xcb_client_message_event_t event = {
         .response_type = XCB_CLIENT_MESSAGE,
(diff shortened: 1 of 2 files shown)


BluewyDiamond:
in my case switching from xwayland fullscreen game back to steam and then back to the game will result the game going black, switching to wayland window and then back to the xwayland fullscreen game works fine

BluewyDiamond:
i've retried it just to be sure and now the xwayland game does not turn black even if i switch to steam (xwayland window) or in other words it works for me now, not sure what happened when i first tried it where it did turn black

src/xwayland/XWM.cpp
```diff
@@ -891,6 +891,9 @@ void CXWM::activateSurface(SP<CXWaylandSurface> surf, bool activate) {
 }
 
 void CXWM::sendState(SP<CXWaylandSurface> surf) {
+    if (surf->fullscreen)
+        surf->setWithdrawn(surf != focusedSurface);
```
vaxerski, line 895: this is not correct. Generally, we don't want to say a surface is withdrawn as long as its mapped on our side, ever.

Withdrawn is like "minimized to tray", from what I understand.

IMO we should just always set it to `false`, as long as the surface is mapped, if that doesn't break anything.

UjinT34: This is what happens with proton games in i3. And that is what makes the issue go away. Maybe the withdrawn state should come from some unhandled event but I don't know which one it is.

jaschutte: Whether or not it should always be set to false, the current patch does fix the black screen issue in all the games I've experienced it in.

vaxerski: The question is whether changing `surf != focusedSurface` to `!surf->mapped` doesn't fix the issue as well, while being more of what we want.

UjinT34: It turned out that resending normal state is enough. Still might not be 100% correct but works. Also added some debug logging.

vaxerski (approved):
thanks! This is way better

turboshark:
Sorry if this is a dumb question but how do I get this to work? I'm trying to play GTA IV and it still crashes when I switch workspaces. If I try to play my game in windowed mode and use Hyprland fullscreen it works but it has an thin line chopped off the screen at the bottom. I'm using Arch with everything updated in this not in the repo yet?
image

BluewyDiamond:
are u using the aur package (hyprland-git)?

turboshark:
> are u using the aur package (hyprland-git)?

No I'm using upstream, how long do you think it'll take for this to be added to the main Arch repo?

vaxerski:
whenever we release 0.42.0 and then a couple days, most likely.

PR #7017, opened 2024-07-24
add bindp and noshortcutsinhibit

MightyPlaza:
adds `bindp` and `noshorcutsinhibit` window rule to allow control over https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1 per bind and per client
using `bindp` because all the good letters have already been used

seems to work fine on qemu
wiki PR: https://github.com/hyprwm/hyprland-wiki/pull/724


src/protocols/ShortcutsInhibit.cpp
```diff
@@ -73,6 +73,9 @@ bool CKeyboardShortcutsInhibitProtocol::isInhibited() {
     if (!g_pCompositor->m_pLastFocus)
         return false;
 
+    if (g_pCompositor->getWindowFromSurface(g_pCompositor->m_pLastFocus.lock())->m_sWindowData.noShortcutsInhibit.valueOrDefault())
```
vaxerski, line 76: ehhhh is this a good idea?

MightyPlaza: yeah you're probably right
what's the solution? storing a `PHLWINDOWREF` on each `CWLSurface` when created or will that cause issues?

vaxerski: not my point, the question is should the state of isInhibited() be altered by the focus. IMO no

MightyPlaza: it already is doe
the next bit of the function is
```cpp
for (auto& in : m_vInhibitors) {
     if (in->surface() != g_pCompositor->m_pLastFocus)
         continue;
```
which already depends on focus

vaxerski: ah, right. It's fine then

vaxerski (approved):
cool

MightyPlaza:
just to leave here, I initially thought the concern was the unnecessary looping over all windows for each bind on each key press

vaxerski:
it's not there, read the source for the func
I fixed it an hour ago

PR #7024, opened 2024-07-24
remove toggleopaque

MightyPlaza:
removes `toggleopaque` dispatcher in favour of `hyprctl setprop active opaque toggle`

needs wiki MR

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -98,7 +98,6 @@ CKeybindManager::CKeybindManager() {
     m_mDispatchers["pass"]                           = pass;
     m_mDispatchers["sendshortcut"]                   = sendshortcut;
     m_mDispatchers["layoutmsg"]                      = layoutmsg;
-    m_mDispatchers["toggleopaque"]                   = toggleOpaque;
     m_mDispatchers["dpms"]                           = dpms;
     m_mDispatchers["movewindowpixel"]                = moveWindow;
     m_mDispatchers["resizewindowpixel"]              = resizeWindow;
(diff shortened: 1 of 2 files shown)


MightyPlaza:
https://github.com/hyprwm/hyprland-wiki/pull/726

noahfraiture:
Is it possible to not push breaking change without a notice ? The command was good, what is the point of pushing such breaking change ?

MightyPlaza:
> Is it possible to not push breaking change without a notice ? The command was good, what is the point of pushing such breaking change ?

there was no need to point in having a dispatcher just to toggle a single window property, since an previous commit allowed to toggling other properties like noborder or noshadow

like I said above just replace it with `hyprctl setprop active opaque toggle`

hrideshmg:
Could this be documented in the wiki? I don't think the average user should have to waddle through github issues to figure out how to fix their config.

deng232:
> > Is it possible to not push breaking change without a notice ? The command was good, what is the point of pushing such breaking change ?
> 
> there was no need to point in having a dispatcher just to toggle a single window property, since an previous commit allowed toggling other properties like noborder or noshadow
> 
> like I said above just replace it with `hyprctl setprop active opaque toggle`
> 
> on another note, maybe setprop should be added to the dispatchers to not require using hyprctl

but it was a frequently used function isnt it? what is the cost of having toggleopaque dispatcher anyway.

Eclextic:
Hyprland is a distro known to have lots of breaking changes.

> but it was a frequently used function isnt it? what is the cost of having toggleopaque dispatcher anyway.

Consistency. That's better than anything else. Someone searching the docs for toggle will now easily find everything they are searching for.

shogeki:
I found this change really annoying because previously I was toggling certain windows opaque on launch but retaining their opacity settings to change them transparent with hotkey if I want.  Trying to do that with hyprctl setprop active doesn't work since there is no guarantee if/when the window I just launched might be active.

MightyPlaza:
> I found this change really annoying because previously I was toggling certain windows opaque on launch but retaining their opacity settings to change them transparent with hotkey if I want. Trying to do that with hyprctl setprop active doesn't work since there is no guarantee if/when the window I just launched might be active.

what do you mean?
`toggleopaque` always used the active window as well
besides you can just use the `opaque` window rule

AdrienLemaire:
toggleopaque can be removed from the completion list.

PR #7026, opened 2024-07-24
fix crash

MightyPlaza:
fixes: https://github.com/hyprwm/Hyprland/issues/7025
tested on older hyprland build and works fine, can't currently build -git


khaneliman:
Tested patch locally, fixes my instant hyprland crash with anyrun launcher.

gksudolol:
Can confirm this fixes crashing on git

khaneliman (approved):

PR #7028, opened 2024-07-24
add date and flags to crash report

MightyPlaza:
adds date and flags to crash report
on another note, I can't seem to get the `debug` flag to appear here or in `hyprctl version`


vaxerski (approved):
thanks!

debug is fed from cmake, prolly has to be fixed (maybe added? maybe got removed) there

PR #7040, opened 2024-07-25
Not so dumb cursor copy

UjinT34:
- Adds support for dumb buffers with `cursor:allow_dumb_copy = true`.
- Adds `cursor:allow_slow_copy`

With `cursor:allow_slow_copy = true` creates an additional renderable swapchain if rendering to cursor swapchain fails. Renders to this swapchain buffer and copies the result to main cursor swapchain.

I don't really know what I am doing here %) Some bits might be totally wrong. But this works for me and doesn't seem to cause any issues.
This code skips many sanity checks because it assumes that fallback buffer will get the same size and format as the main one.

Requires https://github.com/hyprwm/aquamarine/pull/31


Honkazel:
@UjinT34 ig this PR can be closed?

PR #7123, opened 2024-07-31
some minor fd/socket cleanups and make logging thread safe

gulafaran:
the logger change is something i kinda made apparent on my own by placing various Debug::log() in places to figure things out and suddenly i hit bad things, so make it thread safe and only open file once on start and close it on exit.

the rest is a few fd/socket leaks and closing already closed or closing -1 fd's making valgrind a bit spammy.

also change the way bezier loops over getYForPoint,

looping a float and incrementing it can cause subtle precision errors like looping 8 times or 10 times depending on how it turns out. god willing it does it right. so loop 9 iterations and get the float value from that instead

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -518,6 +518,8 @@ void CCompositor::cleanup() {
 
     // this frees all wayland resources, including sockets
     wl_display_destroy(m_sWLDisplay);
+
+    Debug::close();
 }
 
 void CCompositor::initManagers(eManagersInitStage stage) {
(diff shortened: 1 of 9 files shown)


vaxerski (approved):
danke

vaxerski:
no pch failed

gulafaran:
> no pch failed

ah yeah just noticed, fixing. uno minuto.

gulafaran:
> no pch failed

fixed.

vaxerski (approved):
sehr schoen

PR #7128, opened 2024-07-31
allows toggling fullscreenstate

MightyPlaza:
toggles `fullscreenstate value` between `value` and `FSMODE_NONE` if it is already set to `value`


vaxerski (approved):
oke doke

PR #7131, opened 2024-08-01
fix buffer crash with cursors

gulafaran:
current buffer->buffer can turn out to be null actually check for its existence or use the lastbuffer when calling updateCursorShm()

was hitting crashes at `auto shmAttrs = current.buffer->buffer->shm();`  when running nested session, was this how its supposed to be?


vaxerski (approved):

vaxerski:
> was hitting crashes at auto shmAttrs = current.buffer->buffer->shm(); when running nested session, was this how its supposed to be?

dunno

Issue #7138, opened 2024-08-01
Window decoration doesn't get properly updated when toggling fullscreen


sungyoonc:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 60571cd5ccc76f91209ef2faac93ecea542de221  (border: fixup infinite recursion).
Date: Thu Aug 1 19:36:15 2024
Tag: v0.41.2-152-g60571cd5, commits: 5038

flags: (if any)


System Information:
System name: Linux
Node name: acacia
Release: 6.10.2-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Sat, 27 Jul 2024 16:49:55 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Raptor Lake-P [Iris Xe Graphics] [8086:a7a0] (rev 04) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/fox/.config/hypr/hyprland.conf: Read Succeeded
# See https://wiki.hyprland.org/Configuring/Keywords/ for more

# See https://wiki.hyprland.org/Configuring/Monitors/
monitor = eDP-1,2880x1800@120,auto-left,1.8,vrr,1
monitor = DP-3,1920x1080@60,auto-right,auto
monitor=,preferred,auto,auto

# Idle configuration, only run when hypridle is not running
exec-once = pidof hypridle || hypridle &

source = ~/.config/hypr/config.d/env.conf

source = ~/.config/hypr/config.d/ime.conf

source = ~/.config/hypr/config.d/input.conf

source = ~/.config/hypr/config.d/layout.conf

source = ~/.config/hypr/config.d/visual.conf

source = ~/.config/hypr/config.d/rules.conf

source = ~/.config/hypr/config.d/keybinds.conf

[...]

PR #7169, opened 2024-08-03
fix infinite recursion on no_gaps when only

MightyPlaza:
fixes infinite recursion when using `no_gaps_when_only`
partially reverts https://github.com/hyprwm/Hyprland/commit/60571cd5ccc76f91209ef2faac93ecea542de221

closes: #7138 and closes: #7167


vaxerski (approved):
thats one way to do it

PR #7206, opened 2024-08-07
fix leds on kb creation

ikalco:
when starting or vt switching, numlock led isnt set until a key is pressed
this fixes that

yes


vaxerski (approved):
thanks

PR #7212, opened 2024-08-07
hyprctl: link to much less libraries

sliedes:
I got annoyed by latency caused by implementing some operations in shell scripts; particularly if the shell script needed to read some status using hyprctl and then decide to do more hyprctl executions.

This makes hyprctl start significantly faster.

`$ time for ((i=0; i<1000; i++)); do hyprctl/hyprctl -j activewindow >/dev/null; done`

Before: 12.269 s (about 12.3 ms/execution)
After: 2.142 s (about 2.1 ms/execution)

The one thing I don't like is how this adds another mention (and version constraint) of `hyprutils>=0.2.1` in the top level `CMakeLists.txt`. I don't understand CMake well enough to tell if the `deps` could be made to somehow inherit `hyprctl_deps`.

I believe it is ready to merge as is, unless you know a way to only mention hyprutils once :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,12 @@ endif()
 find_package(OpenGL REQUIRED COMPONENTS ${GLES_VERSION})
 
 pkg_check_modules(
+  hyprctl_deps
+  REQUIRED
+  IMPORTED_TARGET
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
thanks!

PR #7219, opened 2024-08-07
core: Move to C++26 and use native_handle to CLOEXEC the debug fd

vaxerski:
Moves hyprland to C++26 to use `native_handle` to CLOEXEC the debug fd and avoid it being passed to children (every process spawned...)

@fufexan CI
@jbeich will this work on BSD / clang / libc++? I can see it has support for it since clang18

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ else()
 endif()
 
 include_directories(. "src/" "subprojects/udis86/" "protocols/")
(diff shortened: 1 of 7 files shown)


jbeich:
I confirm, builds fine on FreeBSD -CURRENT with Clang/libc++ 18. For older versions (until EOL) downstream can temporarily revert or add workarounds.

vaxerski:
alright, thanks!

Issue #6298, opened 2024-06-01
IPC: improve activelayout event


cnt0:
### Description

Hello. Hyprland IPC has the following event emitted on the layout change
```
activelayout>>KEYBOARDNAME,LAYOUTNAME
```
`KEYBOARDNAME` and `LAYOUTNAME` are separated by comma. The problem is that `KEYBOARDNAME` can contain commas as well. At least it does in my case, it starts with `mistel-co.,ltd.-`. 

Such behavior complicates parsing, for example, [shellevents](https://github.com/hyprwm/contrib/tree/main/shellevents) can't correctly handle it right now. 

I believe that instead of polluting IPC clients with workarounds we can fix the IPC itself. For example we can introduce the `activelayoutv2>>LAYOUTNAME` event, because who cares about keyboard names anyway. But I'd listen to suggestion first.

MahouShoujoMivutilde:
```bash

~ 
❯ event_data='some,strange,name,layout' 

~ 
❯ echo "${event_data%,*}"                      
some,strange,name

~ 
❯ echo "${event_data#*,}" 
strange,name,layout


~ 
❯ echo "${event_data##*,}"
layout

```

But if we're talking about this - there are apparently layouts with commas too e.g.:
* Czech (QWERTY, Macintosh)
* Arabic (QWERTY, Eastern Arabic numerals)
* Czech (US, Colemak, UCW support)

https://gist.github.com/MahouShoujoMivutilde/8a12c0d40e7b262f2c78ef6e01193542

zjeffer:
This is indeed a very annoying issue for IPC clients. I see the event is posted in InputManager.cpp:

https://github.com/hyprwm/Hyprland/blob/22138ac259b2f4253be29311f6b60fbd675074b4/src/managers/input/InputManager.cpp#L873-L874

and two other places in this file. 

We could maybe replace the `','` in this line with something like a `;`, but I feel this is not a very robust as some keyboard names or layout could contain a `;` as well. 

Maybe a better solution could be to escape the existing commas in the keyboard name and layout with a `\`? Does that make sense to do?

Vladimir-csp:
Since issue title is quite generic, i'll add another improvement request: also output layout index. Otherwise it is harder to translate name to index for use in `hyprctl switchxkblayout`.

ENDERZOMBI102:
encountered this `,` issue, but with window titles ;w;

[...]

PR #7275, opened 2024-08-10
fix: change layout event's separator

s1syph0s:

The layout event separator is changed from "," to ";" to improve parsing ability, since some keyboard can also have "," in its name. https://github.com/Alexays/Waybar/issues/2137 https://github.com/Alexays/Waybar/pull/3224

Breaks compability with waybar's hyprland/language module.

This is ready for merging


MahouShoujoMivutilde:
I think adding another event, e.g. `activelayoutv2` with `;` as a separator would better, to not break compatibility with everything else that uses it.

Original `activelayout` could be then declared deprecated in the next version, if needed, and in the meantime a note in the wiki can be added to explain why v2 is the one that should be used now.

s1syph0s:
Thanks for the input! I'll fix this soon

MahouShoujoMivutilde:
Rereading a related issue, someone here https://github.com/hyprwm/Hyprland/issues/6298#issuecomment-2212585258 said that some keyboards could contain `;` in their names. I didn't see anything like that, but sure.

Maybe let's just use `;;;`, or `|`, or something else extremely unlikely to appear in the name, but easy to split by?

Btw, see `activewindow` and `activewindowv2`, no need to EMIT_... twice, as it seems to be internal Hyprland's events, which aren't serialized in strings where splitting would matter, as opposed to the ones posted at socket2.

https://github.com/hyprwm/Hyprland/blob/4fdc0d55e4b44bb5300679025d2378fb6de0cae4/src/Compositor.cpp#L1007-L1010

or 

https://github.com/hyprwm/Hyprland/blob/4fdc0d55e4b44bb5300679025d2378fb6de0cae4/src/Compositor.cpp#L2031-L2033

vaxerski:
you're avoiding the problem instead of fixing it. Just replace `,`s with a space or something?

MahouShoujoMivutilde:
That would mean that IPC users would need to do that too, if they want to match a particular keyboard / layout, because now `.name` and `.active_keymap` won't match between `hyprctl devices` (or config) and events at socket2.

Having more unique separator character feels like a simpler change.

vaxerski:
`\,`?

MahouShoujoMivutilde:
As for escaping commas in keyboard / layouts - that would mean you still can't just split for it and get what you want (the same thing that `hyprctl devices -j` has).

```sh
❯ echo "activelayout>>some\,weird\,keyboard,Some\,Layout" | awk -F '>>' '{print $2}' | awk -F ',' '{print $2}'
weird\
```
You'll need to (pre/post)process it like e.g.

```sh
❯ echo "activelayout>>some\,weird\,keyboard,Some\,Layout" | awk -F '>>' '{print $2}' | sed 's#\\,#~#g' | awk -F ',' '{print $1}' | sed 's#~#,#g'
some,weird,keyboard

~ 
❯ echo "activelayout>>some\,weird\,keyboard,Some\,Layout" | awk -F '>>' '{print $2}' | sed 's#\\,#~#g' | awk -F ',' '{print $2}' | sed 's#~#,#g'
Some,Layout
```

As separator - it still has problem, at least for awk, 

```sh
~ 
❯ echo "activelayout>>some,weird,keyboard\,Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\,' '{print $1}' 
awk: warning: escape sequence `\,' treated as plain `,'
some

~ 
❯ echo "activelayout>>some,weird,keyboard\,Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\\,' '{print $1}' 
awk: warning: regexp escape sequence `\,' is not a known regexp operator
some

```

but that's fixable in awk and isn't a problem in a saner `choose`

```sh
❯ echo "activelayout>>some,weird,keyboard\,Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\\\\,' '{print $1}' 
some,weird,keyboard

~ 
❯ echo "activelayout>>some,weird,keyboard\,Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\\\\,' '{print $2}'
Some,Layout
```

This could work, but is worse than just unique not escaped character.

Admittedly, `||` is also not as neat as I thought it would be:

```sh
 echo "activelayout>>some,weird,keyboard||Some,Layout" | awk -F '>>' '{print $2}' | awk -F '||' '{print $1}'
some,weird,keyboard||Some,Layout

~ 
❯ echo "activelayout>>some,weird,keyboard||Some,Layout" | awk -F '>>' '{print $2}' | awk -F '||' '{print $2}'


~ 
❯ echo "activelayout>>some,weird,keyboard||Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\|\|' '{print $2}' 
awk: warning: escape sequence `\|' treated as plain `|'


~
❯ echo "activelayout>>some,weird,keyboard||Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\\|\\|' '{print $2}' 
Some,Layout

~ 
❯ echo "activelayout>>some,weird,keyboard||Some,Layout" | awk -F '>>' '{print $2}' | awk -F '\\|\\|' '{print $1}'
some,weird,keyboard
```

But that's, again, purely awk's problem. For waybar it wouldn't matter.

But `;;;` as a separator would work everywhere, and is more intuitive

```
❯ echo "activelayout>>some,weird,keyboard;;;Some,Layout" | awk -F '>>' '{print $2}' | awk -F ';;;' '{print $1}' 
some,weird,keyboard

~ 
❯ echo "activelayout>>some,weird,keyboard;;;Some,Layout" | awk -F '>>' '{print $2}' | awk -F ';;;' '{print $2}'
Some,Layout
```

You get what you split for, no need to learn any escaping.

s1syph0s:
Should I just replace the separator from `||` to `;;`?

cnt0:
I'd propose something like
```
activalayoutv2>>KEYBOARDNAME,LAYOUTNAME,LAYOUTNAMEPOS
```
where integer `LAYOUTNAMEPOS` points to the beginning of `LAYOUTNAME` in the event string.

This allows us to 1) correctly parse the event regardless of layout and keyboard name contents 2) maintains the backwards compatibility with `activelayout`, so we actually don't need to introduce the new event at all.

s1syph0s:
I think that we still need to introduce a new event `activelayoutv2`, since adding `LAYOUTNAMEPOS` at the end will break waybar's current parser..

vaxerski:
> This could work, but is worse than just unique not escaped character.

dumb statement, what is there to make you so sure there is any character that will ever be unique?

MahouShoujoMivutilde:
Practical experience of how humans make up names for things meant to be read by other humans? 

`;` is rather common part of written language, so to see it in some `Gaymer Keyboard (uwu edition; international)` is more likely than `;;;` or `||` or a `\t` (tab). 

Maybe someone who has a keyboard with a fully programmable controller in it can mess this up.

But if you expect a completely arbitrary text in the name and layout, sure, escaping the separator character is the way. 

If we're being this pedantic, escaping the separator during the serialization and unescaping on user side should **always** be done, for all events with arbitrary output, and maybe all for the sake of consistency / simpler maintainability should the expected text change later.



I just think that the less unescaping the user needs to do, the better.

vaxerski:
escaping is a common practice and I still believe `\,` is the way

s1syph0s:
screenshot_12-08-2024-194517

I tried to use `\,` as the separator, but it is an unknown escape sequence. If I build this and gets the event output, I will receive this, which has no difference:

```
activelayout>>wl_keyboard,English (US)
activelayoutv2>>wl_keyboard,English (US)
```

If I remove the `v2` for the second output and put both in a diffcheck tool, the tool won't find any difference.

screenshot_12-08-2024-200146

Using rust to parse the event output, trying to read `\,` will result in an error, since the escape sequence is not known. Do you mean to use a literal `\,`, which leads to `activelayoutv2>>wl_keyboard\,English (US)`, or escape the comma?

vaxerski:
you need to put `\\,` in the source

s1syph0s:
Hi, is there any update to this issue? I'll be happy to adjust the PR

vaxerski:
well I am still convinced escaping the commas with `\` is the way to go.

s1syph0s:
I created a [commit](https://github.com/hyprwm/Hyprland/pull/7275/commits/3b55c64c2978cafcb20e26b73f9795c3c8b68c4d) that escapes the comma. Could you review the code?

vaxerski:
it's the opposite of what we want. We want to escape the commas in the layout name

s1syph0s:
Hi,

sorry for the late response.. I was taking a break from software dev in my free time..

I tried escaping the comma in both the layout and the language, but the event still can't be parsed properly..

```
raw: activelayoutv2>>hp\,-inc-hyperx-alloy-origins,German (no dead keys)
split: hp\ || -inc-hyperx-alloy-origins,German (no dead keys)

raw: activelayoutv2>>hp\,-inc-hyperx-alloy-origins,English (US)
split: hp\ || -inc-hyperx-alloy-origins,English (US)
```

Another idea that I would propose is to remove the comma completely from the keyboard name and the layout, but this is also not backwards compatible. To fix this in waybar for example, we also need to remove the commas from the input keyboard name. What do you think about this?

vaxerski:
closing due to inactivity

PR #7278, opened 2024-08-10
fix data race and a unsigned int rollover

gulafaran:
just some minor ubsan finds, keycodeToModifier was called with a 0 keycode on mouse keys, and the switch case rolled it over on 0 - 8, no harm but unnecessery.

the watchdog is a seperate thread and hit data races in both reading the config values and reading the bools with the main thread, add a atomic bool for the main thread to wait until the watchdog has read its configvalues so it doesnt hit races in reading the config values with main thread. and change the members to atomic aswell.

also with enabling various ubsan/threadsanitizer/asan flags it gave a few false positives from the non unicode characters in hyprdebug, just change them to a unicode name.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -552,6 +552,10 @@ void CCompositor::initManagers(eManagersInitStage stage) {
 
             g_pConfigManager->init();
             g_pWatchdog = std::make_unique<CWatchdog>(); // requires config
+            // wait for watchdog to initialize to not hit data races in reading config values.
+            while (!g_pWatchdog->m_bWatchdogInitialized) {
+                std::this_thread::yield();
+            }
 
             Debug::log(LOG, "Creating the PointerManager!");
             g_pPointerManager = std::make_unique<CPointerManager>();
--- a/src/debug/HyprDebugOverlay.cpp
+++ b/src/debug/HyprDebugOverlay.cpp
@@ -7,8 +7,8 @@ CHyprDebugOverlay::CHyprDebugOverlay() {
     m_pTexture = makeShared<CTexture>();
 }
 
-void CHyprMonitorDebugOverlay::renderData(CMonitor* pMonitor, float µs) {
-    m_dLastRenderTimes.push_back(µs / 1000.f);
+void CHyprMonitorDebugOverlay::renderData(CMonitor* pMonitor, float durationUs) {
+    m_dLastRenderTimes.push_back(durationUs / 1000.f);
 
     if (m_dLastRenderTimes.size() > (long unsigned int)pMonitor->refreshRate)
         m_dLastRenderTimes.pop_front();
@@ -17,8 +17,8 @@ void CHyprMonitorDebugOverlay::renderData(CMonitor* pMonitor, float µs) {
         m_pMonitor = pMonitor;
 }
 
-void CHyprMonitorDebugOverlay::renderDataNoOverlay(CMonitor* pMonitor, float µs) {
-    m_dLastRenderTimesNoOverlay.push_back(µs / 1000.f);
+void CHyprMonitorDebugOverlay::renderDataNoOverlay(CMonitor* pMonitor, float durationUs) {
+    m_dLastRenderTimesNoOverlay.push_back(durationUs / 1000.f);
 
     if (m_dLastRenderTimesNoOverlay.size() > (long unsigned int)pMonitor->refreshRate)
         m_dLastRenderTimesNoOverlay.pop_front();
@@ -188,12 +188,12 @@ int CHyprMonitorDebugOverlay::draw(int offset) {
     return posY - offset;
 }
 
-void CHyprDebugOverlay::renderData(CMonitor* pMonitor, float µs) {
-    m_mMonitorOverlays[pMonitor].renderData(pMonitor, µs);
+void CHyprDebugOverlay::renderData(CMonitor* pMonitor, float durationUs) {
+    m_mMonitorOverlays[pMonitor].renderData(pMonitor, durationUs);
 }
 
-void CHyprDebugOverlay::renderDataNoOverlay(CMonitor* pMonitor, float µs) {
-    m_mMonitorOverlays[pMonitor].renderDataNoOverlay(pMonitor, µs);
+void CHyprDebugOverlay::renderDataNoOverlay(CMonitor* pMonitor, float durationUs) {
+    m_mMonitorOverlays[pMonitor].renderDataNoOverlay(pMonitor, durationUs);
 }
 
 void CHyprDebugOverlay::frameData(CMonitor* pMonitor) {
--- a/src/debug/HyprDebugOverlay.hpp
+++ b/src/debug/HyprDebugOverlay.hpp
@@ -13,8 +13,8 @@ class CHyprMonitorDebugOverlay {
   public:
     int  draw(int offset);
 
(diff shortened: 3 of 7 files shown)


src/debug/HyprDebugOverlay.cpp
```diff
@@ -7,8 +7,8 @@ CHyprDebugOverlay::CHyprDebugOverlay() {
     m_pTexture = makeShared<CTexture>();
 }
 
-void CHyprMonitorDebugOverlay::renderData(CMonitor* pMonitor, float µs) {
-    m_dLastRenderTimes.push_back(µs / 1000.f);
+void CHyprMonitorDebugOverlay::renderData(CMonitor* pMonitor, float duration) {
```
vaxerski, line 10: this is a bannable offense in software development. "duration" without a unit specifier, that is. `durationUs` is ok if you really want to avoid µ

src/config/ConfigValue.hpp
```diff
@@ -11,7 +11,8 @@ template <typename T>
 class CConfigValue {
   public:
     CConfigValue(const std::string& val) {
-        const auto PVHYPRLANG = g_pConfigManager->getHyprlangConfigValuePtr(val);
+        std::lock_guard<std::mutex> guard(m_configMutex);
```
vaxerski, line 14: No. Make the main thread wait for watchdog to get the static pointers, thats it. You can use a condition_variable for that

vaxerski (approved):

src/config/ConfigValue.hpp
gulafaran: would this be more appropiate?

```
commit 48fde62a6fe1cf4083b0c775f60ee21f2554763e
Author: Tom Englund <tomenglund26@gmail.com>
Date:   Sat Aug 10 16:40:09 2024 +0200

    watchdog: avoid data races in watchdog
    
    asan thread sanitizer reported data races in the watchdog from reading
    and setting the bool variables make them std::atomic bools. also add a
    atomic bool for the main thread to wait for to avoid data race when
    reading the config values.

diff --git a/src/Compositor.cpp b/src/Compositor.cpp
index e3a347fd9..b2dc2cf01 100644
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -552,6 +552,9 @@ void CCompositor::initManagers(eManagersInitStage stage) {
 
             g_pConfigManager->init();
             g_pWatchdog = std::make_unique<CWatchdog>(); // requires config
+            // wait for watchdog to initialize to not hit data races in reading config values.
+            while (!g_pWatchdog->m_bWatchdogInitialized)
+                std::this_thread::yield();
 
             Debug::log(LOG, "Creating the PointerManager!");
             g_pPointerManager = std::make_unique<CPointerManager>();
diff --git a/src/helpers/Watchdog.cpp b/src/helpers/Watchdog.cpp
index b9f654daf..c7ff648ba 100644
--- a/src/helpers/Watchdog.cpp
+++ b/src/helpers/Watchdog.cpp
@@ -18,15 +18,14 @@ CWatchdog::CWatchdog() {
     m_pWatchdog = std::make_unique<std::thread>([this] {
         static auto PTIMEOUT = CConfigValue<Hyprlang::INT>("debug:watchdog_timeout");
 
-        while (1337) {
-            std::unique_lock lk(m_mWatchdogMutex);
+        m_bWatchdogInitialized = true;
+        while (!m_bExitThread) {
+            std::unique_lock<std::mutex> lk(m_mWatchdogMutex);
 
             if (!m_bWillWatch)
-                m_cvWatchdogCondition.wait(lk, [this] { return m_bNotified; });
-            else {
-                if (m_cvWatchdogCondition.wait_for(lk, std::chrono::milliseconds((int)(*PTIMEOUT * 1000.0)), [this] { return m_bNotified; }) == false)
-                    pthread_kill(m_iMainThreadPID, SIGUSR1);
-            }
+                m_cvWatchdogCondition.wait(lk, [this] { return m_bNotified || m_bExitThread; });
+            else if (m_cvWatchdogCondition.wait_for(lk, std::chrono::milliseconds((int)(*PTIMEOUT * 1000.0)), [this] { return m_bNotified || m_bExitThread; }) == false)
+                pthread_kill(m_iMainThreadPID, SIGUSR1);
 
             if (m_bExitThread)
                 break;
diff --git a/src/helpers/Watchdog.hpp b/src/helpers/Watchdog.hpp
index 7bb499d69..a21e692e9 100644
--- a/src/helpers/Watchdog.hpp
+++ b/src/helpers/Watchdog.hpp
@@ -14,18 +14,19 @@ class CWatchdog {
     void startWatching();
     void endWatching();
 
+    std::atomic<bool>                              m_bWatchdogInitialized{false};
   private:
     std::chrono::high_resolution_clock::time_point m_tTriggered;
 
     pthread_t                                      m_iMainThreadPID = 0;
 
-    bool                                           m_bWatching  = false;
-    bool                                           m_bWillWatch = false;
+    std::atomic<bool>                              m_bWatching  = false;
+    std::atomic<bool>                              m_bWillWatch = false;
 
     std::unique_ptr<std::thread>                   m_pWatchdog;
     std::mutex                                     m_mWatchdogMutex;
-    bool                                           m_bNotified   = false;
-    bool                                           m_bExitThread = false;
+    std::atomic<bool>                              m_bNotified   = false;
+    std::atomic<bool>                              m_bExitThread = false;
     std::condition_variable                        m_cvWatchdogCondition;
 };
 ```
 
 adding a atomic bool m_bWatchdogInitialized , making main thread wait and shed_yield until the watchdog has read its values? seeing as a conditional_value seems to require mutex locking if im not misunderstanding its usage, this means i can drop that mutex completly in configvalue

File: src/debug/HyprDebugOverlay.cpp (hunk 1 above)
Comment (gulafaran) [reply src/debug/HyprDebugOverlay.cpp]: heh sorry :), shall remember that. changed it to durationUs.

**Review by vaxerski — APPROVED** [review]

File: src/config/ConfigValue.hpp (hunk 2 above)
Comment (vaxerski) [reply src/config/ConfigValue.hpp]: oh wait didnt see this, I'll take a look later

Comment (gulafaran) [reply src/config/ConfigValue.hpp]: i added the code to the PR already if its easier to use the github compare/commits/files changed to review

File: src/Compositor.cpp [hunk 3]
```diff
@@ -552,6 +552,9 @@ void CCompositor::initManagers(eManagersInitStage stage) {
 
             g_pConfigManager->init();
             g_pWatchdog = std::make_unique<CWatchdog>(); // requires config
+            // wait for watchdog to initialize to not hit data races in reading config values.
+            while (!g_pWatchdog->m_bWatchdogInitialized)
```
vaxerski, line 556: where are {}

src/Compositor.cpp
```diff
@@ -552,6 +552,9 @@ void CCompositor::initManagers(eManagersInitStage stage) {
...
+                std::this_thread::yield();
```
vaxerski, line 557: what? sleep for a ms?

gulafaran: Hm took it from here https://cplusplus.com/reference/thread/this_thread/yield/

gulafaran: That example section

vaxerski: fair enough

src/Compositor.cpp
gulafaran: added

vaxerski (approved):
amongus?

Issue #7270, opened 2024-08-10
since move to aquamarine mouse doesnt warp to correct position on first launch


gulafaran:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 4fdc0d55e4b44bb5300679025d2378fb6de0cae4  (eventloop: don't dispatch in enterLoop).
Date: Sat Aug 10 00:04:26 2024
Tag: v0.42.0-5-g4fdc0d55, commits: 5074

flags: (if any)


System Information:
System name: Linux
Node name: tom-legion
Release: 6.10.3-gentoo-tom
Version: #1 SMP PREEMPT_DYNAMIC Wed Aug  7 10:11:12 CEST 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Raptor Lake-S UHD Graphics [8086:a788] (rev 04) (prog-if 00 [VGA controller])
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD104M [GeForce RTX 4080 Max-Q / Mobile] [10de:27e0] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  560.31.02  Release Build  (portage@localhost)  tor  8 aug 2024 14:49:46 CEST


os-release: NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
VERSION_ID="2.15"
```

</details>

### Description

upon first launch cursor warps to top left, 0,0. i tried debugging it seems https://github.com/hyprwm/Hyprland/blob/4fdc0d55e4b44bb5300679025d2378fb6de0cae4/src/Compositor.cpp#L2927 is returning an empty vector.

i think the monitor vecSize is never being set. which also causes hyprpicker to always launch in topleft corner aswell. position 0,0

### How to reproduce

first launch of hyprland or launch hyprpicker

vaxerski:
MRs welcome

gulafaran:
bisected and debugged it further, hyprpicker is a seperate issue.

vecSize is actually being set.

[...]

PR #7286, opened 2024-08-11
fix initial cursorwarp

gulafaran:
change the hook to monitorAdded instead of newMonitor so its finalized in the compositor and added to vMonitors, move the checkDefaultCursorWarp to PointerManager and check for it upon mode change. and also ensure it doesnt go out of bounds by replacing it in the middle again on resolution changes.

this im a bit unsure of if its the right approach to the problem so il put up a PR for discussion.

fixes: #7270


vaxerski (approved):
lgmt

PR #7288, opened 2024-08-11
Fix fullscreenState toggling behaviour

txkyel:

When the user toggles the either the `internal` or `client` fullscreen states (using `-1` to not alter the other state) the relevant state is toggled. 

e.g. when the initial state is `1 2` and `1 -1` is dispatched, the state is now set to `0 2` instead of syncing the states to `2 2`.

None

Ready for merge

fyi @MightyPlaza

Thread on previous commit relevant to changes: https://github.com/hyprwm/Hyprland/commit/2b520571e897be2a0e88c8692da607b062000038#r145242308

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2305,6 +2305,10 @@ void CCompositor::setWindowFullscreenState(const PHLWINDOW PWINDOW, sFullscreenS
 
     const bool            CHANGEINTERNAL = !(PWINDOW->m_bPinned || CURRENT_EFFECTIVE_MODE == EFFECTIVE_MODE || (PWORKSPACE->m_bHasFullscreenWindow && !PWINDOW->isFullscreen()));
 
+    // TODO: update the state on syncFullscreen changes
+    if (!CHANGEINTERNAL && PWINDOW->m_sWindowData.syncFullscreen.valueOrDefault())
+        return;
+
     PWINDOW->m_sFullscreenState.client = state.client;
     g_pXWaylandManager->setWindowFullscreen(PWINDOW, state.client & FSMODE_FULLSCREEN);
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2305,10 +2305,6 @@ void CCompositor::setWindowFullscreenState(const PHLWINDOW PWINDOW, sFullscreenS
 
     const bool            CHANGEINTERNAL = !(PWINDOW->m_bPinned || CURRENT_EFFECTIVE_MODE == EFFECTIVE_MODE || (PWORKSPACE->m_bHasFullscreenWindow && !PWINDOW->isFullscreen()));
 
-    // TODO: update the state on syncFullscreen changes
-    if (!CHANGEINTERNAL && PWINDOW->m_sWindowData.syncFullscreen.valueOrDefault())
-        return;
-
     PWINDOW->m_sFullscreenState.client = state.client;
     g_pXWaylandManager->setWindowFullscreen(PWINDOW, state.client & FSMODE_FULLSCREEN);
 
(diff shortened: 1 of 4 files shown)


txkyel:
I'd like to add documentation for this behaviour but I'm not too sure whether it'd be more appropriate to place on the dispatchers page or the rules page. 

Any suggestions?

MightyPlaza:
I don't think setting `syncFullscreen` default to false is a good idea, in most cases you'll want it set to true (for example when using a fullscreen button on a client itself)

src/Compositor.cpp
```diff
@@ -2305,10 +2305,6 @@ void CCompositor::setWindowFullscreenState(const PHLWINDOW PWINDOW, sFullscreenS
 
     const bool            CHANGEINTERNAL = !(PWINDOW->m_bPinned || CURRENT_EFFECTIVE_MODE == EFFECTIVE_MODE || (PWORKSPACE->m_bHasFullscreenWindow && !PWINDOW->isFullscreen()));
 
-    // TODO: update the state on syncFullscreen changes
-    if (!CHANGEINTERNAL && PWINDOW->m_sWindowData.syncFullscreen.valueOrDefault())
```
MightyPlaza, line 2309: this is needed to check if to parse the current state when `syncFullscreen` is set to true

txkyel:
It looks like I've completely misunderstood the functionality and importants of `syncFullscreen`. I expected `syncFullscreen` to only affect the behaviour of the `fullscreenState` dispatcher. 

I still don't quite agree with the toggle behaviour of the `fullscreenState`. Perhaps it would instead make sense to add an additional parameter determine the toggle behaviour to toggle the individual state or to sync the states. What do you think?

MightyPlaza:
> It looks like I've completely misunderstood the functionality and importants of `syncFullscreen`. I expected `syncFullscreen` to only affect the behaviour of the `fullscreenState` dispatcher. I'll revert these changes.
> 
> I still don't quite agree with the toggle behaviour of the `fullscreenState`. Perhaps it would instead make sense to add an additional parameter determine the toggle behaviour to toggle the individual state or to sync the states. What do you think?

I think the best solution for now would be to toggle it to `FSMODE_NONE` individually and only re-sync if both modes get toggled to `FSMODE_NONE`.

txkyel:
> re-sync if both modes get toggled to FSMODE_NONE

Do you mean we should set `syncFullscreen` to false in the cases where only one state is toggled? What would be the effect of leaving `syncFullscreen` as true when the internal and client states are not in sync?

MightyPlaza:
> > re-sync if both modes get toggled to FSMODE_NONE
> 
> Do you mean we should set `syncFullscreen` to false in the cases where only one state is toggled? What would be the effect of leaving `syncFullscreen` as true when the internal and client states are not in sync?

I mean when toggling back to `FSMODE_NONE` set `syncFullscreen` to true, to fully toggle the behaviour of the dispatcher

when `syncFullscreen` is set to true the internal and client modes should be kept the same

src/managers/KeybindManager.cpp
```diff
@@ -1181,14 +1181,14 @@ void CKeybindManager::fullscreenStateActive(std::string args) {
     }
 
     if (internalMode != -1 && clientMode == -1 && PWINDOW->m_sFullscreenState.internal == STATE.internal) {
-        g_pCompositor->setWindowFullscreenState(PWINDOW, sFullscreenState{.internal = PWINDOW->m_sFullscreenState.client, .client = PWINDOW->m_sFullscreenState.client});
-        PWINDOW->m_sWindowData.syncFullscreen = CWindowOverridableVar(true, PRIORITY_SET_PROP);
+        g_pCompositor->setWindowFullscreenState(PWINDOW, sFullscreenState{.internal = FSMODE_NONE, .client = PWINDOW->m_sFullscreenState.client});
+        PWINDOW->m_sWindowData.syncFullscreen = CWindowOverridableVar(false, PRIORITY_SET_PROP);
```
MightyPlaza, line 1185: `syncFullscreen` should be set to false above

vaxerski (approved):

Issue #7240, opened 2024-08-08
Crash while attempting to create CDRMLeaseDevice for render-only node


jpalus:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 9a09eac79b85c846e3a865a9078a3f8ff65a9259  (props: bump version to 0.42.0).
Date: Wed Aug 7 19:17:10 2024
Tag: v0.42.0, commits: 5069

flags: (if any)


System Information:
System name: Linux
Node name: rock
Release: 6.1.43-14-rk2312
Version: #3b95df6d8 SMP Tue Jul  9 09:03:31 UTC 2024


GPU information: 


os-release: NAME="PLD Linux"
VERSION="3.0 (Th)"
ID="pld"
VERSION_ID="3.0"
PRETTY_NAME="PLD Linux 3.0 (Th)"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:pld-linux:pld:3.0"
HOME_URL="https://www.pld-linux.org/"
BUG_REPORT_URL="http://bugs.pld-linux.org/"


plugins:

======Config-Start======


Config File: /home/users/jan/.config/hypr/hyprlandd.conf: Read Succeeded

# #######################################################################################
# AUTOGENERATED HYPR CONFIG.
# PLEASE USE THE CONFIG PROVIDED IN THE GIT REPO /examples/hypr.conf AND EDIT IT,
# OR EDIT THIS ONE ACCORDING TO THE WIKI INSTRUCTIONS.
# #######################################################################################

# This is an example Hyprland config file.
# Refer to the wiki for more information.
# https://wiki.hyprland.org/Configuring/Configuring-Hyprland/

# Please note not all available settings / options are set here.
# For a full list, see the wiki

# You can split this configuration into multiple files
# Create your files separately and then link them to this file like this:
# source = ~/.config/hypr/myColors.conf


################
### MONITORS ###
################

# See https://wiki.hyprland.org/Configuring/Monitors/
monitor=,preferred,auto,auto


###################
### MY PROGRAMS ###
###################

# See https://wiki.hyprland.org/Configuring/Keywords/

# Set programs that you use
$terminal = footclient
$fileManager = dolphin


#################
### AUTOSTART ###
#################

# Autostart necessary processes (like notifications daemons, status bars, etc.)
# Or execute your favorite apps at launch like this:

# exec-once = $terminal
exec-once = nm-applet &
[...]

Issue #6949, opened 2024-07-21
X server no longer running for xwayland apps after aquamarine merge


calebaden:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit cf373d315e9fb060576ed407bd5ee2dfb8a6d2e2  (touch: add touch swipe invert config (6940)).
Date: Sun Jul 21 21:59:09 2024
Tag: v0.41.2-50-gcf373d31, commits: 4936

flags: (if any)


System Information:
System name: Linux
Node name: arch
Release: 6.9.10-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Thu, 18 Jul 2024 18:06:13 +0000


GPU information: 
2d:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6600/6600 XT/6600M] [1002:73ff] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/caleb/.config/hypr/hyprland.conf: Read Succeeded
# See https://wiki.hyprlaad.org/Configuring/Monitors/
monitor = DP-1,highrr,0x0,1
workspace = 1, monitor:DP-1, default:true
workspace = 2, monitor:DP-1
monitor = DP-2,preferred,2560x0,1.25
workspace = 3, monitor:DP-2, default:true
workspace = 4, monitor:DP-2

# Rotate pen display
monitor = DP-2,transform,3

#env = AQ_NO_ATOMIC,1
#env = WLR_DRM_NO_ATOMIC,1
env = HYPRCURSOR_THEME,volantes
env = HYPRCURSOR_SIZE,24

# See https://wiki.hyprland.org/Configuring/Keywords/ for more

# Execute your favorite apps at launch
[...]

PR #7290, opened 2024-08-11
protocols avoid crashing in drmlease

gulafaran:
instead of resetting the uniqueptr in itself in the constructor and making it destruct before constructing is done add it to doLater.

by the looks of it the if check was also missing a ! making it reset even if it succeded.

seems either through unique_ptr globals or classforwarding? we cant use the defined LOGM in the constructor that causes crashes aswell, see https://godbolt.org/z/aYEbEfrYa

fixes: #7240 

maybe: #6949

### DIFF
--- a/src/protocols/DRMLease.cpp
+++ b/src/protocols/DRMLease.cpp
@@ -1,5 +1,6 @@
 #include "DRMLease.hpp"
 #include "../Compositor.hpp"
+#include "managers/eventLoop/EventLoopManager.hpp"
 #include <aquamarine/backend/DRM.hpp>
 #include <fcntl.h>
 
@@ -225,7 +226,7 @@ CDRMLeaseDevice::CDRMLeaseDevice(SP<Aquamarine::CDRMBackend> drmBackend) : backe
     auto fd = drm->getNonMasterFD();
 
     if (fd < 0) {
-        LOGM(ERR, "Failed to dup fd for drm node {}", drm->gpuName);
+        Debug::log(ERR, "[DRMLease] Failed to dup fd for drm node {}", drm->gpuName);
         return;
     }
 
@@ -247,10 +248,8 @@ CDRMLeaseProtocol::CDRMLeaseProtocol(const wl_interface* iface, const int& ver, 
             break;
     }
 
-    if (!primaryDevice || primaryDevice->success) {
-        PROTO::lease.reset();
-        return;
-    }
+    if (!primaryDevice || !primaryDevice->success)
+        g_pEventLoopManager->doLater([]() { PROTO::lease.reset(); });
 }
 
 void CDRMLeaseProtocol::bindManager(wl_client* client, void* data, uint32_t ver, uint32_t id) {


src/protocols/WaylandProtocol.hpp
```diff
@@ -35,6 +35,6 @@ class IWaylandProtocol {
     IWaylandProtocolDestroyWrapper m_liDisplayDestroy;
 
   private:
-    std::string m_szName;
+    std::string m_szName  = "";
```
vaxerski, line 38: isnt the default ctor for string exactly this

gulafaran: yeah std::string is such a special type it seems. reverted

src/protocols/DRMLease.cpp
```diff
@@ -225,7 +225,7 @@ CDRMLeaseDevice::CDRMLeaseDevice(SP<Aquamarine::CDRMBackend> drmBackend) : backe
     auto fd = drm->getNonMasterFD();
 
     if (fd < 0) {
-        LOGM(ERR, "Failed to dup fd for drm node {}", drm->gpuName);
+        Debug::log(ERR, "DRMLease Failed to dup fd for drm node {}", drm->gpuName);
```
jpalus, line 228: Note that it does not preserve format in which `DRMLease` is in square brackets. I guess there's no need to use `Debug::log` directly but just `protoLog` without `LOGM` indirection.

gulafaran: Should just refactor the entire logging in these protocols heh

gulafaran: added the brackets back but yeah a logging refactoring is needed in the protocols so this doesnt accidently occur again, but its out of the scope of this PR. is like 40 different files with various places it needs a checkover.

src/protocols/DRMLease.cpp
```diff
@@ -246,14 +246,12 @@ CDRMLeaseProtocol::CDRMLeaseProtocol(const wl_interface* iface, const int& ver,
...
     }
-
-    if (!primaryDevice || primaryDevice->success) {
-        PROTO::lease.reset();
-        return;
-    }
 }
 
 void CDRMLeaseProtocol::bindManager(wl_client* client, void* data, uint32_t ver, uint32_t id) {
+    if (!success())
```
vaxerski, line 252: this is wrong. We cannot refuse binding because that will crash clients.

We _have to_ remove the global. Which, IMO, should be done by destroying this protocol object

gulafaran: yeah but isnt that what we do in ProtocolManager.cpp line 159. if the fd fails to dup we both dont bind and we destroy the protocol object. if it succeeds it binds and doesnt destroy

vaxerski: destroy it in doLater then

gulafaran: wont that cause a possible race with us binding the protocol on a failed fd? since doLater is technically inside the wayland eventloop? and some client trying to use this while it cant, just "thinking" out load heh

gulafaran: roger, il change it back to dolater etc. give me a few minutos

gulafaran: now, we are down to very few line changes heh

vaxerski (approved):
thanks

asiantuntija:
Can confirm https://github.com/hyprwm/Hyprland/issues/6949 is still happening

PR #7304, opened 2024-08-12
Fix disabling tap-to-click

sliedes:

If input:touchpad:tap-to-click is false, the code nevertheless does not disable it in libinput. This change fixes this for me (i.e. after this change, tap-to-click is disabled when set to false in the config file).

It is ready.


src/managers/input/InputManager.cpp
```diff
@@ -1114,6 +1114,8 @@ void CInputManager::setPointerConfigs() {
             if (libinput_device_config_tap_get_finger_count(LIBINPUTDEV)) // this is for tapping (like on a laptop)
                 if (g_pConfigManager->getDeviceInt(devname, "tap-to-click", "input:touchpad:tap-to-click") == 1)
                     libinput_device_config_tap_set_enabled(LIBINPUTDEV, LIBINPUT_CONFIG_TAP_ENABLED);
```
vaxerski, line 1116: cant you just `g_pConfigManager->getDeviceInt(devname, "tap-to-click", "input:touchpad:tap-to-click") ? LIBINPUT_CONFIG_TAP_ENABLED : LIBINPUT_CONFIG_TAP_DISABLED`?

sliedes: Sure. I copied the style from the tap-and-drag and drag_lock options just above this code. Would you like me to change those at the same time, or only this?

vaxerski: feel free

sliedes: I updated this and a few places in the surrounding code. Only managed to give it quite limited testing so far because of the NixOS/C++26 isssue.

vaxerski (approved):
thanks

PR #7324, opened 2024-08-13
protocols refactor protocol logging to a macro

gulafaran:
this avoids the usage of the unique_ptr PROTO::protocol before it has been constructed incase one wants to log something inside the constructor itself, move the logging to macros and print file:linenumber on ERR,CRIT,WARN and classname on the rest of the levels.

this is how it looks in the log with two test printing lines added to CCursorShapeProtocol constructor.
```
PROTOLOG(ERR, "constructing CCursorShapeProtocol");
PROTOLOG(LOG, "constructing CCursorShapeProtocol");
```

2024-08-13-224450_hyprshot


src/macros.hpp
```diff
@@ -112,3 +112,32 @@
...
+            size_t end   = colons - begin;                                                                                                                                         \
+            return prettyFunction.substr(begin, end);                                                                                                                              \
+        } else {                                                                                                                                                                   \
+            return "Global";                                                                                                                                                       \
+        }                                                                                                                                                                          \
+    }()
+
+#define PROTOLOG(level, ...)                                                                                                                                                       \
```
vaxerski, line 129: LOGM

src/macros.hpp
```diff
@@ -112,3 +112,32 @@
     namespace Aquamarine {                                                                                                                                                         \
         class name;                                                                                                                                                                \
     }
+
+#define EXTRACT_CLASS_NAME()                                                                                                                                                       \
```
vaxerski, line 116: all this should be in IWaylandProtocol.hpp prolly

src/macros.hpp
```diff
@@ -112,3 +112,32 @@
...
+    []() -> std::string {                                                                                                                                                          \
```
vaxerski, line 117: this can be constexpr

src/macros.hpp
gulafaran: moved it there

src/macros.hpp
gulafaran: changed

src/macros.hpp
gulafaran: but can std::string really be constexpr? hitting compiler errors trying todo so

vaxerski: std::string_view

vaxerski: mhm

vaxerski (approved):
:+1:

Issue #7083, opened 2024-07-28
Option to disable activation environment management


Vladimir-csp:
### Description

Hi!
Following  #4217, could you please also add an env option `HYPRLAND_NO_SD_VARS` to disable manipulating activation environments, for the same case of delegating it to an external session manager.

Related: if dbus implementation used is `dbus-broker`, there is no need to separately run `dbus-update-activation-environment`, it reuses activation environment of systemd.

izmyname:
> Related: if dbus implementation used is `dbus-broker`, there is no need to separately run `dbus-update-activation-environment`, it reuses activation environment of systemd.

Wait what? If I use Arch with dbus-broker, I don't need it? ```
exec-once = dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
  exec-once = systemctl --user import-environment WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
```

P.S I'm sorry for off-topic, but I'm really curious about this.

MightyPlaza:
> > Related: if dbus implementation used is `dbus-broker`, there is no need to separately run `dbus-update-activation-environment`, it reuses activation environment of systemd.
> 
> Wait what? If I use Arch with dbus-broker, I don't need it?
> 
> ```
> exec-once = dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
>   exec-once = systemctl --user import-environment WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
> ```
> 
> P.S I'm sorry for off-topic, but I'm really curious about this.

```cpp
void CCompositor::startCompositor() {
    signal(SIGPIPE, SIG_IGN);

    if (m_pAqBackend->hasSession() /* Session-less Hyprland usually means a nest, don't update the env in that case */) {
        const auto CMD =
#ifdef USES_SYSTEMD
            "systemctl --user import-environment DISPLAY WAYLAND_DISPLAY HYPRLAND_INSTANCE_SIGNATURE XDG_CURRENT_DESKTOP QT_QPA_PLATFORMTHEME PATH XDG_DATA_DIRS && hash "
            "dbus-update-activation-environment 2>/dev/null && "
#endif
[...]

PR #7345, opened 2024-08-15
Add DesktopNames to hyprland.desktop

Vladimir-csp:

Adds `DesktopNames` with canonical `Hyprland` name to `hyprland.desktop`. DMs or SMs would use this to set `$XDG_CURRENT_DESKTOP`.

No.

Yes.

### DIFF
--- a/example/hyprland.desktop
+++ b/example/hyprland.desktop
@@ -3,3 +3,4 @@ Name=Hyprland
 Comment=An intelligent dynamic tiling Wayland compositor
 Exec=Hyprland
 Type=Application
+DesktopNames=Hyprland;


Vladimir-csp:
Related to this and #7083, do I understand correctly that currently Hyrpland [unsets ](https://github.com/hyprwm/Hyprland/blob/main/src/Compositor.cpp#L425) `XDG_CURRENT_DESKTOP` and does not consider existing value? That might not be a good idea.

vaxerski (approved):
sure, though Hyprland will set the env itself

vaxerski:
> Related to this and https://github.com/hyprwm/Hyprland/issues/7083, do I understand correctly that currently Hyrpland [unsets ](https://github.com/hyprwm/Hyprland/blob/main/src/Compositor.cpp#L425)XDG_CURRENT_DESKTOP and does not consider existing value? That might not be a good idea.

No, that's on exit. On launch it overwrites it

Vladimir-csp:
> sure, though Hyprland will set the env itself

Compositor can be a part of some desktop environment, which can be launched by a DM or SM, which can set `XDG_CURRENT_DESKTOP` according to `DesktopNames` from entry. This is why existing value should be left as is if no explicit `env = XDG_CURRENT_DESKTOP,...` statement exists in config.

vaxerski:
sooo... we should only set it iff it's unset/empty?

Vladimir-csp:
Yes.

vaxerski:
feel free to make a MR

Vladimir-csp:
#7347

Vladimir-csp:
#8110 resolved it

PR #7347, opened 2024-08-15
Preserve existing XDG_CURRENT_DESKTOP

Vladimir-csp:

Preserves existing `XDG_CURRENT_DESKTOP`, only set if empty.  For compatibility with DMs/SMs.

My 1st cpp commit. Yay!

I have no idea

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -36,7 +36,10 @@ int main(int argc, char** argv) {
     setenv("XDG_BACKEND", "wayland", 1);
     setenv("_JAVA_AWT_WM_NONREPARENTING", "1", 1);
     setenv("MOZ_ENABLE_WAYLAND", "1", 1);
-    setenv("XDG_CURRENT_DESKTOP", "Hyprland", 1);
+
+    // Set XDG_CURRENT_DESKTOP if it was not set for us
+    if (!getenv("XDG_CURRENT_DESKTOP"))
+        setenv("XDG_CURRENT_DESKTOP", "Hyprland", 1);
 
     // parse some args
     std::string              configPath;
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -36,7 +36,10 @@ int main(int argc, char** argv) {
     setenv("XDG_BACKEND", "wayland", 1);
     setenv("_JAVA_AWT_WM_NONREPARENTING", "1", 1);
     setenv("MOZ_ENABLE_WAYLAND", "1", 1);
-    setenv("XDG_CURRENT_DESKTOP", "Hyprland", 1);
+
+    // Set XDG_CURRENT_DESKTOP if it was not set for us
+    if (!getenv("XDG_CURRENT_DESKTOP"))
+        setenv("XDG_CURRENT_DESKTOP", "Hyprland", 1);
```
vaxerski, line 42: you should also add a flag to not `unsetenv` it on exit if it wasn't overwritten

Vladimir-csp: Does this look right?
```
    // Set XDG_CURRENT_DESKTOP if it was not set for us
    bool xdgcdExisted = False;
    if (!getenv("XDG_CURRENT_DESKTOP")) {
        setenv("XDG_CURRENT_DESKTOP", "Hyprland", 1);
    } else {
        xdgcdExisted = True;
    }
```
I'm afraid, connecting this to `CCompositor::cleanEnvironment()` might be beyond my current knowledge of C++ which is practically nonexistent.

leiserfg: Maybe instead of storing `xdgcdExisted` it will be simpler to unset  XDG_CURRENT_DESKTOP only if it is `Hyprland`?

Vladimir-csp: If it came from desktop entry, this would be ambiguous.

vaxerski: I'll do it myself then

Vladimir-csp: Thanks!

vaxerski (approved):

PR #7358, opened 2024-08-15
Add HYPRLAND_NO_SD_VARS env condition

### LINKED ISSUE #7083: Option to disable activation environment management

Vladimir-csp:

Closes #7083. This was pretty simple.

Not sure if hardcoding `systemctl` & `dbus-update-activation-environment` stuff is really a good idea. Config would be more appropriate IMHO.

Especially the way `dbus-update-activation-environment` and the original dbus' environment works: it can not unset stuff.

Calling `dbus-update-activation-environment` with a bunch of already unset vars does literally nothing. From manpage:

> VAR
... If VAR is not present in the environment, this argument is silently ignored.

The only cleanup option with original dbus is to export empty strings, and execute without `--systemd`.

And with `dbus-broker` there is no need for `dbus-update-activation-environment` at all, which begs for special treatment.

Patched my debian's hyprland 0.41.2, seems to work.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -660,8 +660,10 @@ void CCompositor::startCompositor() {
     signal(SIGPIPE, SIG_IGN);
 
     if (
-        m_pAqBackend->hasSession() /* Session-less Hyprland usually means a nest, don't update the env in that case */
-        && !envEnabled("HYPRLAND_NO_SD_VARS") /* Activation environment management is not disabled */
+        /* Session-less Hyprland usually means a nest, don't update the env in that case */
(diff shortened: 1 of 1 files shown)


vaxerski (approved):

Vladimir-csp:
Thanks!

Issue #7191, opened 2024-08-05
kitty terminal stops rendering if moved from non-visible workspace to visible workspace


MicroTransactionsMatterToo:
### Regression?

No

### System Info and Version

<details><summary>System/Version Info</summary>

```sh
Hyprland, built from branch main at commit 2b520571e897be2a0e88c8692da607b062000038  (keybinds: improve fullscreenstate toggling (7174)).
Date: Mon Aug 5 07:40:34 2024
Tag: v0.41.2-165-g2b520571, commits: 5051

flags: (if any)


System Information:
System name: Linux
Node name: abyssian-trigrade
Release: 6.10.2-zen1-2-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Sat, 03 Aug 2024 18:22:59 +0000


GPU information: 
0a:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD106 [GeForce RTX 4060 Ti] [10de:2803] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  555.58.02  Tue Jun 25 01:39:15 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/mtx/.config/hypr/hyprland.conf: Read Succeeded
# Startup Apps
exec-once = sleep 4 && waybar
exec-once = /usr/lib/polkit-kde-authentication-agent-1
exec-once = python3 /home/mtx/.config/hypr/blender_float.py
exec-once = sleep 10 && aw-qt
# Variable Declarations
$terminal = kitty
$fileManager = dolphin
$menu = wofi --show drun

# Env Vars
env = XCURSOR_SIZE,24

# == QT Configuration
env = GDK_BACKEND,wayland,x11,*
env = QT_QPA_PLATFORM,wayland;xcb
[...]

PR #7396, opened 2024-08-18
compositor: update suspendstate on window move

gulafaran:
hyprctl dispatch -- movetoworkspacesilent x,"^kitty$" where X is the current workspace makes kitty stops updating until current workspace is changed while it is on the screen. update the suspend state after it has been moved.

fixes: #7191


vaxerski (approved):
thanks!

PR #7409, opened 2024-08-19
feat: more systemd examples

JL2210:
These allow launching hyprland with a systemd service. They provide `graphical-session.target` which allows enabling services such as the ones for Waybar and Mako.

The existing examples are more examples than actually being usable. For example the load ordering that waybar and mako expect for `graphical-session.target` is not upheld with them.

They are rather flaky if modified. In particular:
* If `Before=graphical-session.target` is removed then services part of `graphical-session.target` will be launched before hyprland has time to update the user environment variables.
* If `--wait` is removed, the service will exit and immediately put you back into your display manager (e.g. regreet, sddm).

Ready for merging as far as I'm concerned. I would even consider installing these by default on systemd systems.

### DIFF
--- a/example/hyprland-session.service
+++ b/example/hyprland-session.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Hyprland - Tiling compositor with the looks
+Documentation=man:Hyprland(1)
+BindsTo=graphical-session.target
+Before=graphical-session.target
+Wants=graphical-session-pre.target
+After=graphical-session-pre.target
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/Hyprland
+ExecStop=/usr/bin/hyprctl dispatch exit
+Restart=on-failure
+Slice=session.slice
--- a/example/hyprland-systemd.desktop
+++ b/example/hyprland-systemd.desktop
@@ -1,0 +1,5 @@
+[Desktop Entry]
+Name=Hyprland
+Comment=An intelligent dynamic tiling Wayland compositor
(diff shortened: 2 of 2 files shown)


vaxerski (approved):
thanks!

izmyname:
> For example the load ordering that waybar and mako expect for `graphical-session.target` is not upheld with them.

Sorry for necroing, but what does that mean and how can I fix this? I'm using hyprland-session.service and start it from tty and, sometimes, waybar.service and mpd-notification.service don't start if I dispatch exit, then start hyprland-session.service again.

izmyname:
UPD:  to force start waybar - add
 `[Units] 
 Upholds=waybar.service ` to hyprland-session.service. Likewise, to always start mpd-notification - add` [Units] Upholds=mpd-notification.service` to` mpd.service`. There will be 10 seconds delay or so, but the services will restart.
 
 Maybe this info will help someone.

izmyname:
IMHO, it would be cool if both  .service and .desktop files, from this PR,  became a part of Hyprland itself, just like hyprpolkit and hypridle have their services available ootb.

vaxerski:
MRs welcome

vaxerski:
"if at first you don't succeed, try, try again"

izmyname:
nah

JL2210:
`/usr/share/wayland-sessions` and `/usr/lib/systemd/system` for the desktop file and service, respectively

izmyname:
Adding them to the build turned out to be way more complicated, I'm afraid.

PR #7439, opened 2024-08-20
fix: static asan patch

JL2210:
Rebases the static asan patch on top of current `CMakeLists.txt`

no

Yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,8 @@ asan:
 	#git reset --hard
 
 	@echo -en "If you want to apply a patch, input its path (leave empty for none):\n"
-	@read patchvar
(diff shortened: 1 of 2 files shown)


fufexan:
Can you explain what the Makefile change does? I'm not usually writing Makefile so I have no idea.

JL2210:
Each line in a Makefile is run in an individual shell. So the patch variable from that line doesn't propogate to any of the lines afterwards. To solve this I added a backslash at the end of the line so both commands are run in the same shell.

Make uses `$` as a variable substitution. To pass a literal `$` to a command you need to write `$$`.

As for the if check, `[` is a command that requires its last argument to be `]`. It's not special syntax, so it needs spaces around its brackets.

fufexan:
Okay then, LGTM!

fufexan (approved):

Issue #7181, opened 2024-08-04
Hyprland crashes if a mirror is enabled and disabled on a monitor


Dr0med4r:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit ae50f8614d92132d918663ea7551bd68eb13953a  ().
Date: 2024-08-03
Tag: , commits: ae50f8614d92132d918663ea7551bd68eb13953a

flags: (if any)


System Information:
System name: Linux
Node name: profugus
Release: 6.10.2
Version: #1-NixOS SMP PREEMPT_DYNAMIC Sat Jul 27 09:40:36 UTC 2024


GPU information:
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU116 [GeForce GTX 1650 SUPER] [10de:2187] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  560.28.03  Release Build  (nixbld@)  Mon Jul 29 22:33:20 UTC 2024


os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.11.20240802.81610ab"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.11 (Vicuna)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.11 (Vicuna)"
VERSION_CODENAME=vicuna
VERSION_ID="24.11"


plugins:
  hypr-darkwindow by micha4w ver 1.0.0

======Config-Start======


Config File: /home/jakob/.config/hypr/hyprland.conf: Read Succeeded
exec-once = /nix/store/j8l2v22xypah1srykhwjlkgwwgdxnnpw-dbus-1.14.10/bin/dbus-update-activation-environment --systemd --all && systemctl --user stop hyprland-session.target && systemctl --user start hyprland-session.target
plugin=/nix/store/55rmgwrq4f6nq12qk8g4ik9fcysh3mfn-Hypr-DarkWindow-0.1/lib/libHypr-DarkWindow.so
animations {
  bezier=wind, 0.05, 0.9, 0.1, 1.05
[...]

PR #7457, opened 2024-08-21
layersurface remove expired layers on arranging

gulafaran:
remove expired layers weakptrs on arranging layers, we can hit multiple null ptr derefs in renderer on mirroring and unmirroring displays otherwise.

fixes: #7181


vaxerski:
probably should be removed on layer... unmap?

also man what are those {}

gulafaran:
> probably should be removed on layer... unmap?
> 
> also man what are those {}

those brackets come from a 18 year old habit, its never gonna stop, sorry xD but yeah one could say this was more of a proof of concept of the issue, it is expired layers are never removed

vaxerski:
yeah thats bad

gulafaran:
ok how about now? lets say i mirror eDP to HDMI in SetMirror it removes eDP from mVmonitors so getmonitorfromid will fail because its gone, so instead im iterating over m_vRealMonitors incase the layersurface being destroyed actually exist on a monitor and remove it from its m_aLayerSurfaceLayers if it exist.

gulafaran:
i tried following the code i hope this doesnt break any sort of animation or other destroying logic since this seems to only occur on mirroring/unmirroring where as the background wallpaper layersurface is removed inbetween the monitor modechanges and recreated but the pointer to the old surface was sticking around in the monitors m_aLayerSurfaceLayers but expired since it was never removed and causing the renderer to hit null ptr derefs

src/desktop/LayerSurface.cpp
```diff
@@ -107,6 +107,15 @@ void CLayerSurface::onDestroy() {
         g_pHyprRenderer->damageBox(&geomFixed);
     }
 
+    for (auto& mon : g_pCompositor->m_vRealMonitors) {
+        for (size_t i = 0; i < mon->m_aLayerSurfaceLayers.size(); i++) {
```
vaxerski, line 111: `auto& lsl : mon->m_aLayerSurfaceLayers`?

src/desktop/LayerSurface.cpp
```diff
@@ -107,6 +107,15 @@ void CLayerSurface::onDestroy() {
...
+            auto it = std::find_if(mon->m_aLayerSurfaceLayers[i].begin(), mon->m_aLayerSurfaceLayers[i].end(), [this](auto& ls) { return ls.get() == this; });
```
vaxerski, line 112: `std::erase_if`

and then you can `return ls.expired() || ls.get() == this`

src/desktop/LayerSurface.cpp
gulafaran: changed

src/desktop/LayerSurface.cpp
gulafaran: changed

vaxerski (approved):
thanks tom!

Issue #7069, opened 2024-07-27
Modifier keys remain pressed after reload


g3r:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 682865632fe986d179ca07e45ae89e65b8058b33  (xwayland: fix high cpu idle usage).
Date: Sat Jul 27 08:58:30 2024
Tag: v0.41.2-101-g68286563, commits: 4987

flags: (if any)


System Information:
System name: Linux
Node name: KRONOS
Release: 6.10.1-1-MANJARO
Version: #1 SMP PREEMPT_DYNAMIC Wed Jul 24 17:43:42 UTC 2024


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GM206 [GeForce GTX 960] [10de:1401] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  555.58.02  Tue Jun 25 01:39:15 UTC 2024


os-release: NAME="Manjaro Linux"
PRETTY_NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://forum.manjaro.org/"
BUG_REPORT_URL="https://docs.manjaro.org/reporting-bugs/"
PRIVACY_POLICY_URL="https://manjaro.org/privacy-policy/"
LOGO=manjarolinux


plugins:
```

</details>

### Description

Modifier keys remain pressed after hyprctl reload. I tested SUPER ALT CONTROL and SHIFT

### How to reproduce

```sleep 2 && hyprctl reload```
While on sleep keep pressed a modifier and release it after reload is complete
After reload, the modifier key remains active until it is pressed again

MahouShoujoMivutilde:
Running on a399f98c68d017152883fbf81d67624ac3254073

Now they still get stuck, but you have to press some other modifier after reload for it to kick in, and you **can** get them unstuck by pressing the stuck modifier another time.

MahouShoujoMivutilde:
Wrote a patch

```diff
diff --git a/src/devices/IKeyboard.cpp b/src/devices/IKeyboard.cpp
index e05cbd04..29c05a6d 100644
[...]

PR #7486, opened 2024-08-23
Fix modifier keys getting stuck if depressed during config reload

MahouShoujoMivutilde:
##### The problem

If `input:numlock_by_default = true`, depressed mods will get stuck on config reload; this takes effect after some **other** mod is pressed. Can be unstuck by pressing the stuck mod another time.

##### With patch

Mods won't be stuck

`modifiersState` is still updated to the correct values, but it's done inside `updateModifiersState()` after new `xkbState` takes effect. Thus `SModifiersEvent` is also emitted, which is what I suspect actually fixes it.

This restores 0.41.2 behavior

https://github.com/hyprwm/Hyprland/blob/918d8340afd652b011b937d29d5eea0be08467f5/src/managers/input/InputManager.cpp#L993-L1002

with the exception that selected keyboard layout is preserved.

Fixes https://github.com/hyprwm/Hyprland/issues/7069

Someone [says](https://github.com/hyprwm/Hyprland/issues/7069#issuecomment-2303328558) that you could

> open hyprland config in micro editor, edit config, hold CTRL, press S then Q to save then quit

But even on 0.41.2, regardless of `numlock_by_default` value, I can't reproduce it **if** I wait for config autoreload before pressing Q.

But if you press it fast enough - then it works.

Yes

### DIFF
--- a/src/devices/IKeyboard.cpp
+++ b/src/devices/IKeyboard.cpp
@@ -119,7 +119,8 @@ void IKeyboard::setKeymap(const SStringRuleNames& rules) {
         if (IDX != XKB_MOD_INVALID)
             modifiersState.locked |= (uint32_t)1 << IDX;
 
-        updateModifiers(modifiersState.depressed, modifiersState.latched, modifiersState.locked, modifiersState.group);
+        // 0 to avoid mods getting stuck if depressed during reload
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks!

Issue #7422, opened 2024-08-20
Fails to build on 32-bit architectures (e.g., armv7, i386, powerpc)


jbeich:
### Regression?

Yes

### System Info and Version

v0.42.0-46-gc86db7bb

### Description

Regressed by #7216 (75fe7c890d2b). CC @gulafaran

<details><summary>Clang error</summary>

```c++
$ export CC=clang CXX=clang++
$ export CFLAGS=-m32 CXXFLAGS="-m32 -stdlib=libc++"
$ cmake -B _build
$ cmake --build _build
[...]
src/helpers/MiscFunctions.cpp:356:33: error: no matching function for call to 'max'
  356 |                 predictedWSID = std::max(predictedWSID, 0L);
      |                                 ^~~~~~~~
src/helpers/MiscFunctions.cpp:379:35: error: no matching function for call to 'max'
  379 |                     currentItem = std::max(currentItem, 0UL);
      |                                   ^~~~~~~~
src/helpers/MiscFunctions.cpp:416:42: error: no matching function for call to 'clamp'
  416 |                             namedWSIdx = std::clamp(namedWSIdx, 0UL, namedWSes.size() - 1);
      |                                          ^~~~~~~~~~
/usr/include/c++/v1/__algorithm/max.h:40:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('WORKSPACEID' (aka 'long long') vs. 'long')
   40 | max(_LIBCPP_LIFETIMEBOUND const _Tp& __a, _LIBCPP_LIFETIMEBOUND const _Tp& __b)
      | ^
/usr/include/c++/v1/__algorithm/max.h:51:1: note: candidate template ignored: could not match 'initializer_list<_Tp>' against 'WORKSPACEID' (aka 'long long')
   51 | max(initializer_list<_Tp> __t, _Compare __comp)
      | ^
/usr/include/c++/v1/__algorithm/max.h:60:1: note: candidate function template not viable: requires single argument '__t', but 2 arguments were provided
   60 | max(initializer_list<_Tp> __t)
      | ^   ~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/__algorithm/max.h:31:1: note: candidate function template not viable: requires 3 arguments, but 2 were provided
   31 | max(_LIBCPP_LIFETIMEBOUND const _Tp& __a, _LIBCPP_LIFETIMEBOUND const _Tp& __b, _Compare __comp)
[...]

PR #7510, opened 2024-08-25
build: fix 32bit builds

gulafaran:
ensure the correct type is passed to std::clamp and std::max int64_t is different on 64bit compared to 32bit, also in presentationtime tv_sec is __time_t and on 32bit its a 32bit type so right shift count >= width of type. so only bit shift on 64bit. and avoid potential nullptr deref in the for loops, check for .end() before *it <= endID.

fixes: #7422 

atleast it builds now, but since im not planning on installing a 32bit OS and properly test if we are overflowing/rolling over on other places i cant say if it fixes any runtime issues.

### DIFF
--- a/src/protocols/PresentationTime.cpp
+++ b/src/protocols/PresentationTime.cpp
@@ -58,8 +58,12 @@ void CPresentationFeedback::sendQueued(SP<CQueuedPresentationData> data, timespe
     if (reportedFlags & Aquamarine::IOutput::AQ_OUTPUT_PRESENT_HW_COMPLETION)
         flags |= WP_PRESENTATION_FEEDBACK_KIND_HW_COMPLETION;
 
+    __time_t tv_sec = 0;
+    if (sizeof(__time_t) > 4)
+        tv_sec = when->tv_sec >> 32;
+
     if (data->wasPresented)
-        resource->sendPresented((uint32_t)(when->tv_sec >> 32), (uint32_t)(when->tv_sec & 0xFFFFFFFF), (uint32_t)(when->tv_nsec), untilRefreshNs, (uint32_t)(seq >> 32),
+        resource->sendPresented((uint32_t)tv_sec, (uint32_t)(when->tv_sec & 0xFFFFFFFF), (uint32_t)(when->tv_nsec), untilRefreshNs, (uint32_t)(seq >> 32),
                                 (uint32_t)(seq & 0xFFFFFFFF), (wpPresentationFeedbackKind)flags);
     else
         resource->sendDiscarded();
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -353,13 +353,13 @@ SWorkspaceIDName getWorkspaceIDNameFromString(const std::string& in) {
                 char        walkDir       = in[1];
 
                 // sanitize. 0 means invalid oob in -
-                predictedWSID = std::max(predictedWSID, 0L);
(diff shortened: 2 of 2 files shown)


src/protocols/PresentationTime.cpp
```diff
@@ -58,9 +58,15 @@ void CPresentationFeedback::sendQueued(SP<CQueuedPresentationData> data, timespe
...
+    auto tv_sec = (uint32_t)when->tv_sec;
+
+#ifdef __LP64__ //64-bit architecture check
+    tv_sec = (uint32_t)(when->tv_sec >> 32);
+#endif
+
     if (data->wasPresented)
-        resource->sendPresented((uint32_t)(when->tv_sec >> 32), (uint32_t)(when->tv_sec & 0xFFFFFFFF), (uint32_t)(when->tv_nsec), untilRefreshNs, (uint32_t)(seq >> 32),
-                                (uint32_t)(seq & 0xFFFFFFFF), (wpPresentationFeedbackKind)flags);
+        resource->sendPresented(tv_sec, (uint32_t)(when->tv_sec & 0xFFFFFFFF), (uint32_t)(when->tv_nsec), untilRefreshNs, (uint32_t)(seq >> 32), (uint32_t)(seq & 0xFFFFFFFF),
```
vaxerski, line 68: this looks wrong indeed

gulafaran: pretty much this that had me worried https://godbolt.org/z/rKfovscxz but no idea

gulafaran: i mean isnt this trying to obtain the upper 32bits of the 64bit tv_sec type, only on 32bit it isnt a 64bit type. meaning we cant do it there and just pass it the tv_sec directly?

vaxerski: I'd just leave it as-is

gulafaran: its gonna break on 32bit tho , i mean i could technically even do something like 

```
if (sizeof(__time_t) > 4)
    tv_sec = (uint32_t)(when->tv_sec >> 32);
```

i think

gulafaran: or should it even be 0 if there is no upper bits hm

gulafaran: since the next argument is taking the lower bits and as on 32bit just passes it along as is

gulafaran: something like this maybe. idk, otherwise its gonna bork. as the gcc warning hinted when compiling its UB and the value turned out to something like -51237 or similiar on the 32bit build i did.

```
    __time_t tv_sec = 0;

    if (sizeof(__time_t) > 4)
        tv_sec = (uint32_t)(when->tv_sec >> 32);

    if (data->wasPresented)
        resource->sendPresented(tv_sec, (uint32_t)(when->tv_sec & 0xFFFFFFFF), (uint32_t)(when->tv_nsec), untilRefreshNs, (uint32_t)(seq >> 32), (uint32_t)(seq & 0xFFFFFFFF),
                                (wpPresentationFeedbackKind)flags);
```

vaxerski: uh huh, I guess ok

gulafaran: ok changed, but could really use a test on some 32bit system if this works as intended heh

vaxerski (approved):

PR #7535, opened 2024-08-26
CI: disable stale workflow on forks

NotAShelf:
The stale workflow will run unconditionally, but will fail on forks due to
`STALEBOT_PAT` not being set.

Trigger the workflow _only_ if we are on the main repo, where we can guarantee
the PAT. Also formats the YML syntax to be slightly more readable.

Suppresses annoying notifications if you have a longstanding fork of Hyprland on
your personal account.

_P.S. Please don't nit the formatting or else I will cry._

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -7,22 +7,22 @@ name: Mark stale issues and pull requests
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


NotAShelf:
Cheers

Issue #7424, opened 2024-08-20
Setting a keyword with an empty name prints garbage data to the console


JL2210:
### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit c86db7bbb0cf14d4955ee3a4d13c0ed9f8a0e0ae  (monitor: avoid dangling references to old monitors being undestroyed).
Date: Mon Aug 19 12:44:22 2024
Tag: v0.42.0-46-gc86db7bb, commits: 5115

flags: (if any)


System Information:
System name: Linux
Node name: WileECoyote
Release: 6.10.5-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Thu, 15 Aug 2024 00:25:30 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA106M [GeForce RTX 3060 Mobile / Max-Q] [10de:2560] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  560.31.02  Tue Jul 30 21:02:43 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
  hyprexpo by Vaxry ver 1.0

======Config-Start======


Config File: /home/james/.config/hypr/hyprland.conf: Read Succeeded
# #######################################################################################
# AUTOGENERATED HYPR CONFIG.
# PLEASE USE THE CONFIG PROVIDED IN THE GIT REPO /examples/hypr.conf AND EDIT IT,
# OR EDIT THIS ONE ACCORDING TO THE WIKI INSTRUCTIONS.
# #######################################################################################

# This is an example Hyprland config file.
# Refer to the wiki for more information.
# https://wiki.hyprland.org/Configuring/Configuring-Hyprland/

# Please note not all available settings / options are set here.
# For a full list, see the wiki

[...]

PR #7544, opened 2024-08-27
hyprctl: avoid parsing string::npos on invalid cmd

gulafaran:
invalid lines passed to hyprctl keyword made the string parsing try to parse std::string::npos, avoid that and return an error text instead.

caused it to return garbled output etc when giving it invalid lines.

Fixes: #7424


vaxerski (approved):
lgtm, I'll fix the {} myself cuz I know you are gone for a few days

vaxerski:
> are there keywords that can be called without a value provided etc? as im checking now it returns a fail message if no value is parsed.

doesn't make sense, no

vaxerski (approved):

PR #7563, opened 2024-08-29
Remove setenv("MOZ_ENABLE_WAYLAND", 1, 1)

AzimathGstan:

Firefox don't need this now from version 121, according to arch wiki. 

Keep it will prevent from user overriding MOZ_ENABLE_WAYLAND.
Like: "MOZ_ENABLE_WAYLAND=0 firefox" won't really change to 0
It still being forced to 1.

Which in some case, might lead to confusion for user who wish to run Firefox
under XWayland.

Not really sure if only Firefox depends on this env.
And I am not really sure if setenv() is intended to stop anything from changing it.

Yes, or just move this to default hyprland.conf.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -35,7 +35,6 @@ int main(int argc, char** argv) {
     setenv("HYPRLAND_CMD", cmd.c_str(), 1);
     setenv("XDG_BACKEND", "wayland", 1);
     setenv("_JAVA_AWT_WM_NONREPARENTING", "1", 1);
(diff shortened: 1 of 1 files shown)


vaxerski:
> Like: "MOZ_ENABLE_WAYLAND=0 firefox" won't really change to 0

??? it will.

AzimathGstan:
Oh, sorry. I found out that I misinterpret firefox's about:support.
That is working.
Sorry for the interruption

vaxerski:
yeah then I don't see the point of this, it's easily overridden.

Issue #7059, opened 2024-07-26
After aquamarine kanshi does not work well


leiserfg:
### Regression?

Yes

### System Info and Version

Nixos with hyprland compiled from https://github.com/hyprwm/Hyprland/commit/57371b93a058d17affb0ca92f13b8d40582b3ea0

### Description

I use hyprland with kanshi (basically to turn off the laptop screen whenever I connect it to an external monitor). After aquamarine that does not work well, if I run `hyprctl monitors` it only shows the active (external) monitor as it should, but the laptop screen is still on (just not getting new frames). Disconnecting the external monitor (which should restart the laptop screen) crashes hyprland.

Besides that, the cursor looks tiny in apps not based in gtk  (looks fine in firefox and waybar, but wrong in the wallpaper and kitty).

Those issues don't happen in `0.41.2+date=2024-07-20_f642fb9`.

### How to reproduce

In case it matters, I'm using a framework13 with AMD igpu.

leiserfg:
I also noticed that the screen turns white when hyprland starts (maybe it happens when kanshi rule gets applied).

vaxerski:
where crash report

mauricekraus:
@leiserfg, regarding the white flashing this is a dupe: https://github.com/hyprwm/Hyprland/issues/6950

leiserfg:
I'm testing now in [04b40ea](https://github.com/hyprwm/Hyprland/commit/04b40ea2ec85dd1d74ff18edc046a233b65024ac)
Unplugging the external monitor no longer crashes, but the laptop screen is still on when it should be off (dpms maybe?). Given that it's not crashing I don't have a crash report to send, is there something I can?

leiserfg:
I ran `hyprctl monitors all` and I'm getting:
```
Monitor eDP-1 (ID 0):
	2256x1504@59.99900 at 0x0
	description: BOE 0x0BCA
	make: BOE
	model: 0x0BCA
	serial: 
	active workspace: 0 ()
	special workspace: 0 ()
	reserved: 0 27 0 0
	scale: 1.17
	transform: 0
	focused: no
	dpmsStatus: 1
	vrr: 0
	activelyTearing: false
	disabled: true
	currentFormat: XRGB8888
[...]

PR #7571, opened 2024-08-29
Improve output configuration via wlr-output-configuration

vaxerski:
Improves the handling of wlr-output-configuration (e.g. wdisplays, kanshi)

Notable changes:
 - Output configuration will be valid for as long as the client has the global bound (so likely for as long as it's running)
 - Wlr output configuration will no longer be a rule, it will be treated as a hard override of whatever the user sets in the cfg (for as long as point 1)
 - Older clients have precedence (but who runs 2 monitor managers?)
 
Ref #7059

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -7,6 +7,7 @@
 #include "helpers/varlist/VarList.hpp"
 #include "../protocols/LayerShell.hpp"
 #include "../xwayland/XWayland.hpp"
+#include "../protocols/OutputManagement.hpp"
 
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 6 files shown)


leiserfg:
I just built it and now kanshi thinks the internal monitor was disabled, but it is still enabled (fully working not as before the patch when it was disabled but the last frame was still shown). Will try wdisplay.

leiserfg:
Wdisplay does not work either, I disable the `Enabled` checkbox and as soon as I click Apply it gets enabled again.

vaxerski:
oh wait sec

vaxerski:
now

leiserfg:
Do you want a log with kanshi, with wdisplay, or both?

vaxerski:
whichever doesnt work, and explain what doesnt work

leiserfg:
After `e` it worked with kanshi, let me play with it a bit more.

vaxerski:
the power of e

leiserfg:
It disabled the internal monitor correctly, but didn't enable it when I unplugged the external one 
[hyprland.log](https://github.com/user-attachments/files/16800799/hyprland.log)

My Kanshi rules are:
```
profile undocked {
  output "eDP-1" enable scale 1.175000
}

profile docked-left {
  output "DP-3"
  output "eDP-1" disable
}

profile docked-right {
  output "DP-2"
  output "eDP-1" disable
}


```
the logs of kanshi are:
```
Aug 29 16:32:54 rahmen systemd[1417]: Started Dynamic output configuration.
Aug 29 16:32:55 rahmen kanshi[952746]: applying profile 'docked-left'
Aug 29 16:32:55 rahmen kanshi[952746]: applying profile output 'DP-3' on connected head 'DP-3'
Aug 29 16:32:55 rahmen kanshi[952746]: applying profile output 'eDP-1' on connected head 'eDP-1'
Aug 29 16:32:55 rahmen kanshi[952746]: configuration for profile 'docked-left' applied
Aug 29 16:33:17 rahmen kanshi[952746]: no profile matched
```

it seems to me like the internal monitor never get re-exposed to kanshi.

vaxerski:
eee

vaxerski:
can you post WAYLAND_DEBUG=1 of kanshi when it fails to re-enable edp1

vaxerski:
I don't see anything wrong? What didn't work?

leiserfg:
The laptop's display stays off.

vaxerski:
ye from that log I can't see what's wrong? shikane just... doesnt ask us to enable it.

Can you grab a log from something like kde or sway where it works?

leiserfg:
I will test the PR further tonight.

leiserfg:
I'm testing again. I started hyprland with this as part of my config:
```
 exec-once = env WAYLAND_DEBUG=1 shikane 2> /tmp/shikane.log
```
and I'm doing `tail -f  /tmp/shikane.log`, and I don't see logs appearing when I plug and unplug the external monitor. So it is kinda worse than it is in master. Right now my external monitor is showing a fix frame but does not show as enabled in `hyprctl monitors`.

hypr

shikane logs:
[hyprland.log](https://github.com/user-attachments/files/16805171/hyprland.log)

vaxerski:
hm, thats odd. I'll test on my laptop tomorrow... Does socket2 log anything?

`socat -v unix-connect:$XDG_RUNTIME_DIR/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket2.sock -`

leiserfg:
@vaxerski checking in the git history I found this PR that was what fixed kanshi when a similar problem was happening in the days of wlroots

https://github.com/hyprwm/Hyprland/pull/3718/files#diff-dbcd1ce1e0b916536f17c999473ad9e22c014646ddb2df15c45d726884a3481cR1293-R1300

vaxerski:
@leiserfg I've rebased this on top of main. I remember some other fixes on main, is this better than the current head?

leiserfg:
I was chatting with the author of shikane and he told me the issue is:
> When a display gets unplugged the compositor should send Finished events for the output itself and all its modes to shikane.
In your log shikane only receives events that say that one display is disabled (but not unplugged) and one is enabled.
The Finished events are missing.
And once you plug it back in shikane gets all the initial information again.
For all of its runtime shikane knew that 2 displays existed.
Which should not have been the case because you unplugged your monitor.
There is nothing shikane can do about this.
It sounds interesting as to why the Finished events are not sent to shikane.
However, this is a bug on hyprlands side of things and I'm not willing to put in maintenance effort to work around another software's bug.
Also, please disclose beforehand if you are aware of any bugs which might be related.

leiserfg:
I think this PR adds more complexity to the whole monitor management and it will be better to keep it closer to what it is in master and try to introduce the proper events, what do you think?

vaxerski:
No, this system is generally more robust IMO.

Let me try to fix that, thank the shikane dev for me.

leiserfg:
If you wanna try yourself you can use shikane with
 `env SHIKANE_LOG=trace shikane  2>&1 | grep  Event:: `

vaxerski:
yeah I managed, something is seriously wrong around monitor SPs

vaxerski:
fixed, try now.

leiserfg:
Now it works (I even see the Finished events).

Issue #7507, opened 2024-08-25
Keyboard input doesn't work in Black Desert Online Steam Launcher for 2FA after xwayland rewrite


xadips:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 9a09eac79b85c846e3a865a9078a3f8ff65a9259  (props: bump version to 0.42.0).
Date: Wed Aug 7 19:17:10 2024
Tag: v0.42.0, commits: 5069

flags: (if any)


System Information:
System name: Linux
Node name: actimelis
Release: 6.9.9-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 12 Jul 2024 00:06:19 +0000


GPU information: 
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 31 [Radeon RX 7900 XT/7900 XTX/7900M] [1002:744c] (rev c8) (prog-if 00 [VGA controller])
0f:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Raphael [1002:164e] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
  hyprsplit by shezdy ver 1.0

======Config-Start======


Config File: /home/spidax/.config/hypr/hyprland.conf: Read Succeeded
source = ~/.config/hypr/conf.d/options.conf
source = ~/.config/hypr/conf.d/monitor.conf
source = ~/.config/hypr/conf.d/env.conf
source = ~/.config/hypr/conf.d/exec.conf
source = ~/.config/hypr/conf.d/rules.conf
source = ~/.config/hypr/conf.d/binds.conf

Config File: /home/spidax/.config/hypr/conf.d/options.conf: Read Succeeded
input {
    follow_mouse = 1
[...]

PR #7575, opened 2024-08-30
xwayland: update overrideRedirect on map and configure

Trimutex:
Fixes surfaces that are originally starting with `overrideRedirect = true` not being updated on configure and map notifies.

I kept the update logic in the function in case the other pieces from wlroots needs to be added back in, but I think these already get handled? Might want to double check that.

Works on my end for account creation typing, but want to confirm #7507 works as I don't actually play the game anymore.


vaxerski (approved):
preliminary lgtm I'll wait for op of that issue to report back

vaxerski:
op busy but patch looks good, merging, thanks!

PR #7582, opened 2024-08-30
Add a window rule to render while unfocused

### LINKED ISSUE #2701: Window rule to allow client to render when unfocused / on another workspace
Useful with some games like dark souls which will disconnect you if the fps flutters too much. When running inside gamescope it can limit FPS when unfocused as well. (though not sure if this feature works in wayland?)

vaxerski:
Adds a `renderunfocused` window rule.

Needs testing

fixes #2701

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1037,6 +1037,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_BOOL,
         .data        = SConfigOptionDescription::SBoolData{true},
     },
+    SConfigOptionDescription{
+        .value       = "misc:render_unfocused_fps",
+        .description = "the maximum limit for renderunfocused windows' fps in the background",
+        .type        = CONFIG_OPTION_INT,
+        .data        = SConfigOptionDescription::SRangeData{15, 1, 120},
+    },
 
     /*
      * binds:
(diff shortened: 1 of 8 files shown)


ntnn:
Tried it on glxgears with `windowrulev2 = renderunfocused, title:^(glxgears)$`, the rule was recognized but the FPS still dropped after moving it to an inactive workspace.

vaxerski:
it should drop to ~15

vaxerski:
image
seems to be working for me

the point isnt to render at max fps all the time is it? The point is to avoid apps stalling

ntnn:
Oh ok - then it works, it was about 15fps.

Do you think we could get a configuration option for that?

I see that it's a timeout in milliseconds here: https://github.com/hyprwm/Hyprland/pull/7582/files#diff-dbcd1ce1e0b916536f17c999473ad9e22c014646ddb2df15c45d726884a3481cR131

I think it'd be enough to make that value configurable.

vaxerski:
prolly yea

ntnn:
> the point isnt to render at max fps all the time is it? The point is to avoid apps stalling

Yeah not max fps all the time but some games (like Elden Ring as mentioned in the issue) are fps sensitive and being able to configure it would be great.

vaxerski:
`misc:render_unfocused_fps`

ntnn:
Works like a charm!

Myned:
FWIW, I had Elden Ring in the background for about 15 minutes @ 6634117f868bf087b93d92d9349b5200b78cc609 with the hardcoded 15fps while switching back and forth between workspaces, and it didn't trigger the framerate detection. The ability to configure that value is very welcome though. Thanks!

enzi:
great work! fixed!

PR #7614, opened 2024-09-01
textinput: send deactivate to ime on destory ti

sungyoonc:
#### Describe your PR, what does it fix/add?
Should Fix #5288
Fix #7593 (possible dupe of #5288)

IME wasn't being deactivated when text input was destroyed, due to `INPUT` pointing to a nullptr.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Nope

#### Is it ready for merging, or does it need work?
Ready

PR #5455, opened 2024-04-06
IME: fix ime on closing window

sungyoonc:
#### Describe your PR, what does it fix/add?
Fix IME not working sometimes when closing a window.

When closing a window, sometimes IME gets deactivated by textInputDestroy callback after IME have been activated for another surface.
This causes the IME to not work on the another surface and have to leave and reenter to get the IME working.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Not really

#### Is it ready for merging, or does it need work?
Ready for merging

PR #7617, opened 2024-09-01
textinput: don't deactivate ime if another ti is focused

sungyoonc:
Fix ime getting deactivated when closing a window on certain applications.

Some applications like alacritty, doesn't explicitly `destroy` text input object when closing the application.
This results in ime getting deactivated when it's already focused on another window, due to the destory event getting emitted after the keyboard focus event.

This is a follow up MR of #7614. Because the original fix (#5455) was broken, I've split it into a separate MR.

Nope.

Ready for merging.


vaxerski (approved):
thanks!

PR #7633, opened 2024-09-03
config: add option to enable/disable xwayland

### LINKED ISSUE #7607: Disable XWayland on runtime
### Description

While there is a compile time option to disable XWayland, many distros ship versions with Xwayland enabled by default for convenience. A config option or a flag would be helpful as a troubleshooting step and a security enhancement.

Trimutex:
Allows enabling and disabling XWayland during runtime. #3371 appears to be stale and not being worked on.

1. Disables "DISPLAY" env variable to prevent X11 launches. Due to this, a pop up will occur notifying of this. Is this to be left alone or is there a better way of informing the user that XWayland is not enabled?
2. While it closes the windows, applications like Steam are still running in the background with a tray icon. Is it possible to close these or just ignore them?

Current state of commit works, but see above for things I would like to address first. Closes #7607

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -890,6 +890,9 @@ CXWM::~CXWM() {
 
     if (eventSource)
         wl_event_source_remove(eventSource);
+
+    for (auto const& sr : surfaces)
+        sr->events.destroy.emit();
 }
 
 void CXWM::setActiveWindow(xcb_window_t window) {
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -891,8 +891,9 @@ CXWM::~CXWM() {
     if (eventSource)
         wl_event_source_remove(eventSource);
 
-    for (auto const& sr : surfaces)
+    for (auto const& sr : surfaces) {
         sr->events.destroy.emit();
+    }
 }
 
 void CXWM::setActiveWindow(xcb_window_t window) {
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -615,7 +615,7 @@ void CCompositor::initManagers(eManagersInitStage stage) {
             g_pCursorManager = std::make_unique<CCursorManager>();
 
             Debug::log(LOG, "Starting XWayland");
-            g_pXWayland = std::make_unique<CXWayland>();
+            g_pXWayland = std::make_unique<CXWayland>(g_pCompositor->m_bEnableXwayland);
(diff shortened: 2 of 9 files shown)


vaxerski (commented):
will this work at runtime? e.g. if someone enables it while running hl

vaxerski:
if yes it'd be nice to also do xwaylanf restore while we at it (restart if crash)

Trimutex:
> will this work at runtime? e.g. if someone enables it while running hl

Yeah, the option is meant to work while Hyprland is running and I want to keep it that way

> if yes it'd be nice to also do xwaylanf restore while we at it (restart if crash)

calling g_pXWayland->reset(), or resetting is members (pServer, pWM) crashes it when disabling due to lingering apps, probably due to the same issue as 2 above

Trimutex:
Unless there is some way to simply test if it crashed that I'm unaware of? That would just be oversight from my part if there's a good way for that

vaxerski:
I think if the wm is yeeten off then it crashed or died

Trimutex:
Update note: This PR currently only prevents creation of new XWayland windows by unsetting the env variable. I'll continue to dig deeper on actually shutting down XWayland and applications with it off.

Trimutex:
> Few bugs but currently handles turning off/on multiple times just fine

# Current Bugs
- Steam stays alive after closing its windows and can only exit if XWayland is re-enabled (not sure if fixable)
- Possibly taking another window with it (seems to be if previous window was X11?)

Trimutex:
@vaxerski Should be all good now for merge.

Leaving steam issue alone as I don't think its an XWayland problem, just would require keeping track of PIDs and killing them. That might be leaking into territory better made for another PR.

Additional mention as you asked about it that XWayland should fully restore on re-enable. I can't test that for sure though as I don't have a reliable way of crashing it in such a way that it can't restart.

src/xwayland/XWM.cpp
```diff
@@ -890,6 +890,9 @@ CXWM::~CXWM() {
 
     if (eventSource)
         wl_event_source_remove(eventSource);
+
+    for (auto const& sr : surfaces)
```
vaxerski, line 894: {}

Trimutex: added, didn't realize it was necessary for loops mb

Trimutex:
Managed to crash XWayland and can confirm this restores successfully!

vaxerski (approved):
sankyu!

PR #7649, opened 2024-09-04
input: don't emit idle activity when calling simulateMouseMovement

PaideiaDilemma:

Fixes https://github.com/hyprwm/hyprlock/issues/430

Now `simulateMouseMovement` does not lead to idle activity any more.

Idk if there are cases where `simulateMouseMovement` should post idle activity but I don't think so.

Ready.


vaxerski (approved):
makes sense

vaxerski:
tho in the future it'd be better to move the idle resetting stuff directly to stuff like IMouse and IKeyboard

PR #7650, opened 2024-09-04
config: Add a variable to prevent groups from merging after being dragged

Sonotsugipaa:

The PR adds a config variable that (dis)allows window groups from being "swallowed" into other groups when being dragged over them.  
This setting does **not** prevent _ungrouped_ windows from being dragged into groups.

The default value of the variable (`merge_groups_on_drag`) is `true`, which reflects the behavior of Hyprland prior to this change.

It is ready for merging.

#### Code Style

I tried to follow the code style as seen, but the [PR Guidelines page on the wiki](https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/) only mentions the `/.clang-format` file and I'm not sure I got VSCode to use it; still, I did my best to follow it.

#### Implementation

I implemented the check in `CWindow::canBeGroupedInto` (`/src/desktop/Window.cpp`), but I am not familiar with the codebase so I'm not entirely sure that's the correct place for it.

#### Rationale

My usual workflow has two window groups, with one being the "master" (despite using the dwindle layout); I also use a mouse binding to to drag windows.  
When I want to move a window from one group to the other, my muscle memory has me `moveoutofgroup`ing the window, then swiftly dragging it to its target.

However, occasionally I accidentally press the wrong keys, _forget_ to press the right keys, or even simply twitch and click twice.  
This means that the window group, often containing up to ~8 windows, "hovers" above the other group for a split second, and I end up with one big group containing everything;  
consequently, I need to _manually_ sort the windows back to the previous layout if I can even remember it.

You could prevent this from happening by locking groups or preventing ungrouped windows from joining them, but doing so is still either prone to whoopsies or much slower than having this PR's config variable set to `false`.  
Still, as far as I could gather from social media, I seem to be the only person to have ever had this issue; **someone suggested doing this was already possible**, but when I asked for clarification, silence followed.

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -743,6 +743,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .data        = SConfigOptionDescription::SBoolData{true},
     },
     SConfigOptionDescription{
+        .value       = "group:merge_groups_on_drag",
+        .description = "whether window groups can be dragged into other groups",
+        .type        = CONFIG_OPTION_BOOL,
+        .data        = SConfigOptionDescription::SBoolData{true},
+    },
+    SConfigOptionDescription{
         .value       = "general:col.border_active",
         .description = "border color for inactive windows",
         .type        = CONFIG_OPTION_GRADIENT,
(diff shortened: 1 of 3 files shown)


vaxerski (dismissed):
wiki mr needed

Sonotsugipaa:
I see that the code style is wrong, as I suspected - I'll work on that and the wiki right away.

Sonotsugipaa:
... done.

I also have the code style related change downstream, should I convert this PR to a draft and change it that way?

Sonotsugipaa:
Nevermind, I figured it out - forgive my clumsiness, I don't have much experience with PRs.

vaxerski (approved):

PR #7672, opened 2024-09-06
Added default button mapping for laptop multimedia keys for volume an…

DarthChillash:
This adds sensible defaults for multimedia keys for volume and brightness. Ideally for laptop usage.
I tested this on my Thinkpad T480 and I am using this exact keys config.

It is ready to merge, in my opinion, unless you wish to comment these first so they are disabled by default.

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -228,6 +228,14 @@ bind = $mainMod, mouse_up, workspace, e-1
 bindm = $mainMod, mouse:272, movewindow
 bindm = $mainMod, mouse:273, resizewindow
 
+# Laptop multimedia keys for volume and LCD brightness
+bindel = ,XF86AudioRaiseVolume, exec, wpctl set-volume @DEFAULT_AUDIO_SINK@ 5%+
+bindel = ,XF86AudioLowerVolume, exec, wpctl set-volume @DEFAULT_AUDIO_SINK@ 5%-
+bindel = ,XF86AudioMute, exec, wpctl set-mute @DEFAULT_AUDIO_SINK@ toggle
+bindel = ,XF86AudioMicMute, exec, wpctl set-mute @DEFAULT_AUDIO_SOURCE@ toggle
+bindel = ,XF86MonBrightnessUp, exec, brightnessctl s 10%+
+bindel = ,XF86MonBrightnessDown, exec, brightnessctl s 10%-
+
 
 ##############################
 ### WINDOWS AND WORKSPACES ###
--- a/src/config/defaultConfig.hpp
+++ b/src/config/defaultConfig.hpp
@@ -241,6 +241,13 @@ bind = $mainMod, mouse_up, workspace, e-1
 bindm = $mainMod, mouse:272, movewindow
 bindm = $mainMod, mouse:273, resizewindow
 
+# Laptop multimedia keys for volume and LCD brightness
+bindel = ,XF86AudioRaiseVolume, exec, wpctl set-volume @DEFAULT_AUDIO_SINK@ 5%+
+bindel = ,XF86AudioLowerVolume, exec, wpctl set-volume @DEFAULT_AUDIO_SINK@ 5%-
+bindel = ,XF86AudioMute, exec, wpctl set-mute @DEFAULT_AUDIO_SINK@ toggle
+bindel = ,XF86AudioMicMute, exec, wpctl set-mute @DEFAULT_AUDIO_SOURCE@ toggle
+bindel = ,XF86MonBrightnessUp, exec, brightnessctl s 10%+
+bindel = ,XF86MonBrightnessDown, exec, brightnessctl s 10%-
 
 ##############################
 ### WINDOWS AND WORKSPACES ###


DarthChillash:
You might be right that "not many" people will need this. More heads think clearer than one head. When I first installed Hyprland I got a bit stuck on this part. If you have a laptop then most likely you would be using volume and brightness keys, so I thought it would be nice to have this in the default config that people can just use without researching. Kind of like a jump starter for not-so-advanced users.

But if the goal of the default config is to be as minimal as possible then this change is likely not a good fit.

romanstingler:
> `wpctl`? Also I personally don't know if this really is fit for the default config, because we aren't _only_ having a default supplier (e.g. Firefox, Kitty), but also a default _keybinding_ AND step options (here 10%)...
> 
> I'm not against it, just pointing smth out, as I don't use **any** of these defaults you're proposing myself (then again I'm not the maintainer, soooo)

`wpctl` which is pipeWire wich is the leading replacement for pulseaudio and jack, 
If you have any modern Linux you will probably have pipewire. 
There is no BSD port as far as I know. 

Linux Distributions with PipeWire by Default:

    Fedora:
        Starting from Fedora 34, PipeWire has replaced PulseAudio and JACK as the default audio server.
    Ubuntu:
        Ubuntu 22.10 and later versions have PipeWire as the default audio server for handling audio and video. However, PulseAudio is still available.
    Pop!_OS:
        Pop!_OS, which is based on Ubuntu, has moved to PipeWire as the default in recent versions.
    Arch Linux:
        Arch Linux users can install PipeWire manually, but many Arch-based distributions (such as EndeavourOS) ship PipeWire by default in newer installations.
    Manjaro:
        Manjaro offers PipeWire as the default for audio in some desktop environments, particularly GNOME, and it's easy to switch to PipeWire in other editions.
    Garuda Linux:
        A rolling-release distro based on Arch, Garuda uses PipeWire out of the box for its audio management.
    KDE Neon:
        Newer versions of KDE Neon include PipeWire as part of their audio and video handling setup.

Linux Distributions with Optional PipeWire Support:

    Debian:
        PipeWire is available in the Debian repositories (since Bullseye). Users can switch from PulseAudio to PipeWire manually.
    openSUSE:
        PipeWire is available in both openSUSE Leap and Tumbleweed, but users need to configure it as a replacement for PulseAudio.
    Gentoo:
        Gentoo users can enable PipeWire through custom builds, though it requires manual configuration.
    Slackware:
        PipeWire is available in the Slackware repositories, but manual setup is required.
    Linux Mint:
        Starting from Linux Mint 20.3, PipeWire is supported and can be installed and configured as an alternative to PulseAudio, although it’s not the default.
    Zorin OS:
        Zorin OS supports PipeWire as an option, especially useful for those running GNOME or using video conferencing applications.

Rolling Release and Niche Distributions:

    EndeavourOS:
        An Arch-based rolling-release distro that uses PipeWire by default in many of its configurations.
    Clear Linux:
        Developed by Intel, Clear Linux uses PipeWire for its multimedia management, especially in GNOME setups.
    NixOS:
        NixOS supports PipeWire, and users can easily configure their system to use it instead of PulseAudio.

DarthChillash:
Oh, I did not think that PipeWire was the questionable thing in my pull. If you are doing Wayland you are most likely doing PipeWire too.

vaxerski (approved):

vaxerski:
changes ok but need to be mirrored to exampleconfig in assets

DarthChillash:
Good now, boss?

vaxerski (approved):
yea, thanks :)

RichAyotte:
Would you accept a PR for these too?

```
bindl = , XF86AudioNext, exec, playerctl next
bindl = , XF86AudioPause, exec, playerctl play-pause
bindl = , XF86AudioPlay, exec, playerctl play-pause
bindl = , XF86AudioPrev, exec, playerctl previous
```

Playerctl is a command-line utility and library for controlling media players that implement the MPRIS D-Bus Interface Specification. Compatible players include audacious, cmus, mopidy, mpd, mpv, quod libet, rhythmbox, spotify, and vlc.

vaxerski:
:+1:

JcBernack:
@RichAyotte I would recomment adding `--limit 1` to the `wpctl` arguments when raising volume, otherwise the volume will be raised past 100%. I really don't understand why that is the default behavior of `wpctl`, but it is what is is.

RichAyotte:
@JcBernack this isn't my PR, did you mean to tag @DarthChillash?

DarthChillash:
@JcBernack it's merged already, but I actually use 100+ % on my laptop because T480 speakers are poor so 110-120 % helps a lot. If you guys think Hyprland defaults need to halt at 100% then I will make a new PR.

JcBernack:
@DarthChillash Interesting, but doesn't that cause clipping? In any case, as this is about a default config I would prefer the behavior to be in line with what most users would expect, which I would argue is a maximum of 100% volume.
Not too important though, as it's also really easy to change if a user doesn't like the default :man_shrugging:

Issue #7671, opened 2024-09-06
a command to be executed during shutdown


littleblack111:
### Description

like a command that executes before anything start to shutdown.

use cases of this can be for sth like: https://github.com/hyprwm/Hyprland/issues/2475#issuecomment-2333359383
or stuff like writing script to see if system is updating and cancel the shutdown if it is

Aqa-Ib:
You can script this with a systemd unit targetting shutdown.target

Not a hyprland issue imo.

littleblack111:
ik. i mean this isnt really a issue... its a feature request

but if i do before shutdown.target
it would be like display manger closed, pipwire and everything.

thus stuff like https://github.com/hyprwm/Hyprland/issues/2475#issuecomment-2333359383 won't be achivable

Aqa-Ib:
Still a systemd thing. You need a hyprland.service started by systemd and a script.service that triggers `After=hyprland.service`

littleblack111:
uncertain wdym by a hyprland.service. but if its a service that runs along hyprland or starts hyprland. then i'm pretty sure After=hyprland will make it start after hyprland start instead of when hyprland is shutting down. but even if it runs after hyprland shuts. still wont be able to do stuff like https://github.com/hyprwm/Hyprland/issues/2475#issuecomment-2333359383 cuz
> it would be like display manger closed, pipwire and everything.

Aqa-Ib:
True my bad, it is not After. You can use `ExecStopPost` probably in the systemd unit. You have to play with systemd to achieve this.

littleblack111:
I see.. May I know wdym by hyprland.service though? like replace display manager and just start hyprland?

Aqa-Ib:
Yes

littleblack111:
Actually, that sounds like a good idea. since my display manager config is just autologin... ill have a try

littleblack111:
nope. doesnt work. i'm pretty sure theres a service that sets xdg_runtime_dir. suggested by https://www.reddit.com/r/archlinux/comments/167v652/hyprland_refusing_to_start_says_xdg_runtime_dir/ its a pam_systemd probelm. which idk what the service is. thus this wont work...

Aqa-Ib:
what problem do you face? Why are you looking for that service? Create a new systemd service to start Hyprland from the tty or automatic at boot time, whatever you want, and then use systemd flags in the service file to achieve what you want.

littleblack111:
[...]

Issue #2475, opened 2023-06-09
Persistent Desktop that keeps the apps open that you shut down using


speightashley:
I read through the documentation and guidelines for bash scripts but couldn't find anything that would be suitable to do what I want to do.

I was previously using a tiling window environment called Material Shell that was build on top of Gnome. We could have a persistent desktop with this so that when we restart, it saves the state of the programs that are open upon shutdown.

Is there a way we can do this with Hyprland?

I've done a workaround with starting my regular applications on different desktops but it's not really the same. I might be working on a specific project and when I restart, it will open VScode without that project in its current state.

vaxerski:
I don't think hyprland can magically understand what every program you use is doing and how to restart it in a way that it comes back to that state. That would be the app's job. 

For me, when I open vs code after a restart it does come back to my last opened state, but it's vs code doing it and not hyprland.

speightashley:
So how about just retaining that apps that were opened upon shutdown?

Let's forget about the state of each application but instead just retaining
the apps that were open in the first place.

Would that be more feasible?

vaxerski:
I think that has already been suggested somewhere. In the meantime you can make a script to do that for you. Hyprland exposes PIDs in `hyprctl clients` which you can use to read `/proc/$PID/exe`, save that, and launch next time you start up hyprland :)

speightashley:
Beautiful. Thanks dude.
Will have a play around with that tomorrow.

vaxerski:
you might also want to check out `/proc/$PID/cmdline` if there are any launch args.

speightashley:
Got it. I'm just reading the man page for proc as we speak to see how I can
do it. I'm sure this would be a common request for a lot of people ya know.

I'll be back once I have something working :-D

Amazing work btw :)

mairs8:
Hey, is it possible to post your solution to this? was trying to do the same thing

littleblack111:
if this not get any attention. then how about allowing a exec when shutting down(like storing info to disk) so at least we can write script ourselves to do so

littleblack111:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/18

there is a wayland protocol that is awaiting merge that does exactly this

littleblack111:
im currently using hyprsession, which is a very buggy solution

PR #7683, opened 2024-09-07
config: add exec-shutdown for running commands on shutdown

Trimutex:
Adds a config option for executing programs on shutdown.

Nothing that could crash Hyprland while running immediately found.

Ready, but want tested by #7671 for any edge cases I might be able to fix

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -460,8 +460,6 @@ void CCompositor::cleanup() {
     m_bIsShuttingDown   = true;
     Debug::shuttingDown = true;
 
-    g_pConfigManager->dispatchExecShutdown();
-
 #ifdef USES_SYSTEMD
     if (Systemd::SdBooted() > 0 && !envEnabled("HYPRLAND_NO_SD_NOTIFY"))
         Systemd::SdNotify(0, "STOPPING=1");
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -460,6 +460,8 @@ void CCompositor::cleanup() {
     m_bIsShuttingDown   = true;
     Debug::shuttingDown = true;
 
+    g_pConfigManager->dispatchExecShutdown();
+
 #ifdef USES_SYSTEMD
     if (Systemd::SdBooted() > 0 && !envEnabled("HYPRLAND_NO_SD_NOTIFY"))
         Systemd::SdNotify(0, "STOPPING=1");
--- a/src/Compositor.hpp
+++ b/src/Compositor.hpp
@@ -91,6 +91,7 @@ class CCompositor {
     bool                                       m_bNextIsUnsafe   = false;
     CMonitor*                                  m_pUnsafeOutput   = nullptr; // fallback output for the unsafe state
     bool                                       m_bIsShuttingDown = false;
+    bool                                       m_bFinalRequests  = false;
     bool                                       m_bDesktopEnvSet  = false;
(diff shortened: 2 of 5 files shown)


littleblack111:
@Trimutex I tried it. doesnt really work(I tried hyprctl clients > ~/asdf, which reads "Couldn't read (5)". I think `wl_display_terminate` prob already terminated all the windows? I tried to move `g_pConfigManager->dispatchExecShutdown();` to `CCompositor::stopCompositor` right before `wl_display_terminate` but it doesnt seem to execute then. I tried to just putting logs(`Debug::log(INFO, "stuff"`) but the whole `stopCompositor` part seem to not execute any of the code inside. uncertain how does this work(cuz im noob to coding wayland) but it seems to be the first function that get called when its terminating(by signal).

But anyways. I don't know how would I continue this... But, I hope we can add a exec-pre-shutdown and a exec-post-shutdown. the pre one for first immediate action before any window gets destroyed or anything(e.g. to save info like hyprctl clients, but im not sure if this not working is becuz the IPC shuts down?). But thank you for giving it a go! Hope we can find a solution soon

src/Compositor.cpp
```diff
@@ -460,6 +460,8 @@ void CCompositor::cleanup() {
     m_bIsShuttingDown   = true;
     Debug::shuttingDown = true;
 
+    g_pConfigManager->dispatchExecShutdown();
```
vaxerski, line 463: should be executed just before signal() a few lines up

Trimutex: Execs were started too late anyways, resolved.

Trimutex:
@littleblack111 this should provide what you are looking for now I think

vaxerski (approved):

littleblack111:
@Trimutex is there any way to make it execute it when is shutdown not via hyprctl dispatch exit? I tried sigterm, it doenst execute it.

Trimutex:
I'm a little confused about this. Are you asking for it to be handled post-hyprland-shutdown or as its resetting the variables?

littleblack111:
so, if I shut off hyprland via the the exit dispatcher. it will execute. but any other way(reboot, soft-reboot, restart display manager) will not trigger it.

Trimutex:
> so, if I shut off hyprland via the the exit dispatcher. it will execute. but any other way(reboot, soft-reboot, restart display manager) will not trigger it.

The closest you can come to that is a wrapper for Hyprland in a script I think. This seems like a bit of an XY problem to me tho

Trimutex:
If you want stuff to run on reboot or similar a bash script attached to a restart button inside of Hyprland that is run which calls reboot at end of script is my best guess of what you actually want

littleblack111:
as described at #7671 I want to do #2475 and possibly pause the shutdown or something if some important task is being executed

Trimutex:
> as described at #7671 I want to do #2475 and possibly pause the shutdown or something if some important task is being executed

I don't think Hyprland itself is capable of stopping reboot for updates. You can defer the reboot/shutdown call though by using a script that is called to replace shutdown by whatever method you prefer.

littleblack111:
is the shutdown process if linux un-pausable?

Trimutex:
> is the shutdown process if linux un-pausable?

theoretically yes but realistically only if its scheduled for later. attempting to pause it for updates seems unnecessary anyways as other solutions make more sense (e.g. wrapper script for shutdown that checks if a file contains a "1", skipping shutdown if it does)

Issue #7508, opened 2024-08-25
Build fails with clang/llvm due to C++26


JohnRTitor:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

NixOS 24.11, but irrelevant

</details>

### Description

Build fails with Clang. Happens on Clang 18, 19.

### How to reproduce

Compile Hyprland on NixOS with clang.

On nix, you can just override `stdenv`, set `pkgs.clangStdenv` or `pkgs.llvmPackages.stdenv`.

### Crash reports, logs, images, videos

```log
@nix { "action": "setPhase", "phase": "configurePhase" }
fixing cmake files...
cmake flags: -GNinja -DCMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY=OFF -DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF -DCMAKE_EXPORT_NO_PACKAGE>
-- The C compiler identification is Clang 19.1.0
-- The CXX compiler identification is Clang 19.1.0
-- Detecting C compiler ABI info
...skipping...
[53/247] Building CXX object CMakeFiles/Hyprland.dir/cmake_pch.hxx.pch
[54/247] Building CXX object CMakeFiles/Hyprland.dir/src/debug/Log.cpp.o
FAILED: CMakeFiles/Hyprland.dir/src/debug/Log.cpp.o 
/nix/store/rjsfx6sxjpkgd4f9hl9apm0n8dk7jd9w-clang-wrapper-19.1.0-rc2/bin/clang++ -DDATAROOTDIR=\"/nix/store/srlw63sjjl2xirnlv9qxn>
/build/source/src/debug/Log.cpp:13:26: error: no member named 'native_handle' in 'std::basic_ofstream<char>'
   13 |     auto handle = logOfs.native_handle();
      |                   ~~~~~~ ^
1 error generated.
[55/247] Building CXX object hyprpm/CMakeFiles/hyprpm.dir/src/core/DataState.cpp.o
[56/247] Building CXX object hyprctl/CMakeFiles/hyprctl.dir/main.cpp.o
[57/247] Building CXX object hyprpm/CMakeFiles/hyprpm.dir/src/core/PluginManager.cpp.o
[58/247] Building CXX object CMakeFiles/Hyprland.dir/src/debug/HyprNotificationOverlay.cpp.o
[...]

PR #7774, opened 2024-09-13
nix: use meson

diniamo:
CC @fufexan

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -4,6 +4,7 @@
   pkg-config,
   pkgconf,
   makeWrapper,
+  meson,
   cmake,
   ninja,
   aquamarine,
@@ -89,6 +90,7 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
       hyprwayland-scanner
       jq
       makeWrapper
+      meson
       cmake
       ninja
       pkg-config


fufexan (approved):

JohnRTitor:
I wonder why though, didn't we explicitly move from Meson to CMake? And I'll make a guess here that only one works here at any given time, presumably CMake.

JohnRTitor:
Oh I guess CMake building of Nix package is being phased out (#7778). If it solves issues like this https://github.com/hyprwm/Hyprland/issues/7508#issuecomment-2343619191, I have no complaints.

fufexan:
There is a possibility, but the main reason we've moved back to meson is due to the faster builds (almost twice as fast).

diniamo:
Yes, as weird is at is, meson builds twice as fast.

vaxerski:
sounds wrong and doesn't seem to be the case on my end. CMake builds a bit faster.

Have you tried:
 - not using ninja for cmake
 - building w/o pch

diniamo:
@fufexan ?

fufexan:
2 hours and a bad graph later, I'm back with some rough results
Hyprland build systems build times

Looks like we'll go back to using CMake.

PR #7778, opened 2024-09-13
nix: remove cmake remnants

diniamo:
CC @fufexan


nix/default.nix
```diff
@@ -65,8 +65,6 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
     patches = [
       # forces GCC to use -std=c++26
       ./stdcxx.patch
```
fufexan, line 67: Also not needed.

nix/default.nix
```diff
@@ -144,20 +142,9 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
       (lib.optionals withSystemd [systemd])
     ];
 
-    cmakeBuildType =
```
fufexan, line 147: Should be replaced with `mesonBuildType` iirc. Check Meson's setup hook for more info.

nix/default.nix
```diff
@@ -144,20 +142,9 @@ assert lib.assertMsg (!hidpiXWayland) "The option `hidpiXWayland` has been remov
...
-      if debug
-      then "Debug"
-      else "RelWithDebInfo";
-
     # we want as much debug info as possible
     dontStrip = debug;
 
-    cmakeFlags = [
```
fufexan, line 155: Same here.

fufexan (approved):

Issue #7412, opened 2024-08-19
scuffed cursor with transformed secondary display 


rbutera:
### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 9a09eac79b85c846e3a865a9078a3f8ff65a9259  (props: bump version to 0.42.0).
Date: Wed Aug 7 19:17:10 2024
Tag: v0.42.0, commits: 5069

flags: (if any)


System Information:
System name: Linux
Node name: lancelot
Release: 6.10.5-2-cachyos
Version: #1 SMP PREEMPT_DYNAMIC Wed, 14 Aug 2024 16:17:15 +0000


GPU information: 
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 21 [Radeon RX 6800/6800 XT / 6900 XT] [1002:73bf] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="CachyOS Linux"
PRETTY_NAME="CachyOS"
ID=cachyos
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://cachyos.org/"
DOCUMENTATION_URL="https://wiki.cachyos.org/"
SUPPORT_URL="https://forum.cachyos.org/"
BUG_REPORT_URL="https://github.com/cachyos"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=cachyos


plugins:

======Config-Start======


Config File: /home/rai/.config/hypr/hyprland.conf: Read Succeeded
#  _   _                  _                 _  
# | | | |_   _ _ __  _ __| | __ _ _ __   __| | 
# | |_| | | | | '_ \| '__| |/ _` | '_ \ / _` | 
# |  _  | |_| | |_) | |  | | (_| | | | | (_| | 
# |_| |_|\__, | .__/|_|  |_|\__,_|_| |_|\__,_| 
#        |___/|_|                              
#  
# ----------------------------------------------------- 

# ----------------------------------------------------- 
# Monitor
# ----------------------------------------------------- 
source = ~/.config/hypr/conf/monitor.conf

# ----------------------------------------------------- 
# Autostart
# ----------------------------------------------------- 
source = ~/.config/hypr/conf/autostart.conf

# ----------------------------------------------------- 
# Cursor
# ----------------------------------------------------- 
[...]

PR #7806, opened 2024-09-15
Hide cursor on leave

vaxerski:
Hides the hw cursor plane when it leaves the output.

Ref #7412

needs testing

### DIFF
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -246,9 +246,8 @@ void CPointerManager::recheckEnteredOutputs() {
 
             // if we are using hw cursors, prevent
             // the cursor from being stuck at the last point.
-            // if we are leaving it, move it to narnia.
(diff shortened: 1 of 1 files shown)


drammelt:
The hiding part works for moving screens, however when the mouse moves over any window decorations it still appears for a split second then disappears again.

I think it maybe the call from CInputManager::processMouseRequest to CPointerManager::setCursorBuffer as that calls updateCursorBackend -> attemptHardwareCursor which sets the buffer bypassing the screen checks.

vaxerski:
should be fixed, try now

drammelt:
So close, only one thing happening now. when the cursor transitions from one monitor to another and immediately encounters a window decoration or something like waybar the previous screens cursor remains until the current screens cursor leave the decoration/input area.

IMG_20240916_200629_691

leiserfg:
@drammelt That's two cursors for the price of one.

vaxerski:
I'd say that's minor and overall better than main so I'll merge this

PR #7818, opened 2024-09-16
dispatcher: allow moveintogroup when windows are floating

Aqa-Ib:
This allows to use the moveintogroup dispatcher when windows are floating. 

I don't know why was this disabled in the first place though.

Yes


vaxerski:
why would it be useful?

Aqa-Ib:
I was experimenting trying to script a grouped floating windows workflow. Not very useful indeed, but why do we have this limitation though ?

vaxerski:
I think its because it makes little sense

Aqa-Ib:
Well, it is useful for people that like to use floating grouped windows. Also for consistency sake.

vaxerski (approved):
-273C

PR #7851, opened 2024-09-19
core: add HYPRLAND_CONFIG environment variable

amadalusia:

This adds a brand new environment variable named HYPRLAND_CONFIG to the table, this is another fallback config lookup that Hyprland uses in case --config fails.

There are benefits that come from this:

A) You can now just wrap Hyprland itself with your configuration, this solves any issues with wrapping that may come from systemd.

B) While the XDG directory spec cleans up your ~, at least clean up the XDG directories a bit too, god knows they need it...

There's a Debug::log call which may be unnecessary? Other than that nothing else is really notable.

I am also not amazing at C++ but I tried to make the code look C++ which may be a bit more enticing. /s

I have tested the necessary sections of code (trying to w/ & w/o HYPRLAND_CONFIG) with both doing as I expected.

I believe this is ready to be checked at least.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -684,7 +684,7 @@ std::string CConfigManager::getMainConfigPath() {
         return g_pCompositor->explicitConfigPath;
 
     if (const auto CFG_ENV = getenv("HYPRLAND_CONFIG"); CFG_ENV) {
-      return CFG_ENV;
+        return CFG_ENV;
     };
 
     Debug::log(TRACE, "Seems as if HYPRLAND_CONFIG isn't set, let's see what we can do with HOME.");
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -683,6 +683,12 @@ std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
 
+    if (const char *cfg = getenv("HYPRLAND_CONFIG"); cfg != NULL) {
+        return cfg;
+    }
+
+    Debug::log(INFO, "Seems as if HYPRLAND_CONFIG isn't set, let's see what we can do with HOME.");
+
     static const auto paths = Hyprutils::Path::findConfig(ISDEBUG ? "hyprlandd" : "hyprland");
     if (paths.first.has_value()) {
         return paths.first.value();
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -683,10 +683,7 @@ std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
 
-    if (const auto CFG_ENV = getenv("HYPRLAND_CONFIG"); CFG_ENV) {
-        return CFG_ENV;
-    };
-
+    if (const auto CFG_ENV = getenv("HYPRLAND_CONFIG"); CFG_ENV) return CFG_ENV;
     Debug::log(TRACE, "Seems as if HYPRLAND_CONFIG isn't set, let's see what we can do with HOME.");
 
     static const auto paths = Hyprutils::Path::findConfig(ISDEBUG ? "hyprlandd" : "hyprland");
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -683,11 +683,9 @@ std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
 
-    if (const char *cfg = getenv("HYPRLAND_CONFIG"); cfg != NULL) {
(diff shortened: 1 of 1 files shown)


src/config/ConfigManager.cpp
```diff
@@ -683,6 +683,12 @@ std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
 
+    if (const char *cfg = getenv("HYPRLAND_CONFIG"); cfg != NULL) {
+        return cfg;
+    }
```
fufexan, line 688:
```suggestion
    std::string config_env = getenv("HYPRLAND_CONFIG");
    if (!config_env.empty())
        return config_env;
```

src/config/ConfigManager.cpp
```diff
@@ -683,6 +683,12 @@ std::string CConfigManager::getMainConfigPath() {
...
+
+    Debug::log(INFO, "Seems as if HYPRLAND_CONFIG isn't set, let's see what we can do with HOME.");
```
fufexan, line 690: I don't think this is needed, but we can leave it with log level DEBUG.

src/config/ConfigManager.cpp
vaxerski: this will crash idiot

src/config/ConfigManager.cpp
vaxerski: trace

src/config/ConfigManager.cpp
```diff
@@ -683,6 +683,12 @@ std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
 
+    if (const char *cfg = getenv("HYPRLAND_CONFIG"); cfg != NULL) {
```
vaxerski, line 686: no {}

no `!= NULL`

const are CAPS

you can use auto

```
if (const auto CFG_ENV = getenv("HYPRLAND_CONFIG"); CFG_ENV)
```

src/config/ConfigManager.cpp
fufexan: We do a little trolling

src/config/ConfigManager.cpp
amadalusia: should be resolved now

src/config/ConfigManager.cpp
amadalusia: this should also be resolved

src/config/ConfigManager.cpp
amadalusia: Yeah, the string could NULL when there isn't anything there so std::string has a breakdown and crashes and burns and hoo boy a ton of smoke all over the place.

src/config/ConfigManager.cpp
amadalusia: should actually be resolved now

amadalusia:
fatal git error!

vaxerski:
yeah I think you f-ed up the diff

amadalusia:
Diff as seen in the "Files changed" tab
The files changed tab say the diff looks alright, but yeah the force-pushed diff looks wild.

vaxerski:
the diff doesn't look right, you just added an empty if that does nothing

amadalusia:
> the diff doesn't look right, you just added an empty if that does nothing

how do you get the diff? it looks alright to me (though i may not understand what you are implying)

vaxerski:
image
you posted this yourself. Your code does nothing

amadalusia:
I'll retest it when I am back home. I am on the road right now, but I recall the code working for me.

vaxerski:
the if block does nothing

amadalusia:
im reading the code and i genuinely dont have a clue where you got this idea from.

can you tell me what you passed into hyprland? if you didnt set the HYRPLAND_CONFIG environment variable then run hyprland of course nothing is going to change.

amadalusia:
Oh my fucking god LMAO

I read the code you sent wrong, i though you meant just use that and remove the body

src/config/ConfigManager.cpp
```diff
@@ -683,6 +683,12 @@ std::string CConfigManager::getMainConfigPath() {
     if (!g_pCompositor->explicitConfigPath.empty())
         return g_pCompositor->explicitConfigPath;
 
+    if (const auto CFG_ENV = getenv("HYPRLAND_CONFIG"); CFG_ENV) {
```
vaxerski, line 686: no {} around short ifs

vaxerski (commented):
also needs a clang-format

src/config/ConfigManager.cpp
amadalusia: gotcha, will do both

amadalusia: resolved!

vaxerski:
still not clang-formatted

amadalusia:
> still not clang-formatted
my bad, did it in the wrong order :sweat_smile: 
reformatted it, sorry for any pain induced by this pr

vaxerski (approved):

PR #7875, opened 2024-09-22
Fix dpms toggle

Kam1k4dze:
Maybe I'm stupid, but I don't understand the logic behind the current implementation. It is input monitor agnostic and results in the following truth table:
```
+------+------+---------+
| Mon1 | Mon2 |  Enable |
+------+------+---------+
|  Off |  Off |   false |
|   On |  Off |   false |
|  Off |   On |   false |
|   On |   On |    true |
+------+------+---------+
```
This logic does not seem to make sense to me.
So this PR fixes that.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -2399,9 +2399,7 @@ SDispatchResult CKeybindManager::dpms(std::string arg) {
     bool            enable = arg.starts_with("on");
     std::string     port   = "";
 
-    if (arg.starts_with("toggle"))
-        enable = !std::any_of(g_pCompositor->m_vMonitors.begin(), g_pCompositor->m_vMonitors.end(), [&](const auto& other) { return !other->dpmsStatus; }); // enable if any is off
-
+    bool            isToggle = arg.starts_with("toggle");
     if (arg.find_first_of(' ') != std::string::npos)
         port = arg.substr(arg.find_first_of(' ') + 1);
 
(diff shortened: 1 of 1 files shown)


vaxerski:
hm, I don't like that. To me, sounds quite counter-intuitive. Maybe add two separate options?

Kam1k4dze:
What's the current toggle parameter supposed to do? It doesn't seem to do anything for me.

vaxerski:
toggle on if any are off and off otherwise iirc

Kam1k4dze:
Okay, I guess that's very counterintuitive) The current version is still incorrect because the `!` before `std::any_of()` makes it true only if none of `dpmsStatus` are `false`, so it will enable the monitor if all are already enabled.
I could probably add another option, but I think "toggle" most intuitively means that the input monitor would switch from its current state to the opposite one. 
I don't understand why executing `dpms toggle monitor1` would depend on the states of the other monitors. If monitor2 is disabled, this command will enable monitor1 if it is disabled and do nothing if it is already enabled, meaning I could only disable one monitor with toggle. This seems strange to me.
Are there other reasons for this behavior that I'm not considering?

vaxerski:
> I don't understand why executing dpms toggle monitor1 would depend on the states of the other monitors

yeah this is fine to be fixed, but I am talking about a condition where we dont specify a monitor and toggle all

Kam1k4dze:
If no specific monitor is specified and the 'toggle all' option is used, it will toggle the state of all monitors. For example, if all monitors are off, they will be turned on. If monitor1 is off and monitor2 is on, monitor1 will be turned on and monitor2 will be turned off. This behavior seems intuitive. If you want to turn all monitors on or off, we already have `on` and `off` parameters.

vaxerski:
yeah, I am just unsure whether that's the most intuitive way, but sure, let's roll with that.

vaxerski (approved):

PR #7885, opened 2024-09-22
CMake, Meson: suproject improvements

fufexan:

Meson: try to find udis86 through pkgconfig, fallback to subproject. Adds Nix overlay with udis86 fork -> 2 remaining subproject for Nix
Nix: re-add hyprland-protocols -> 1 remaining subprojects for nix
CMake: use hyprland-protocols from pkg-config, fallback to subproject
CMake: use udis86 from pkg-config, fallback to subproject
Meson: add tracy dependency -> no more subproject deps for nix, unless wanted

Tested various toggles, subprojects/no subprojects. If anyone wants to test feel free.

Should be ready.

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -25,4 +25,4 @@ jobs:
           name: hyprland
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
-      - run: nix build 'git+https://github.com/hyprwm/Hyprland?ref=${{ github.ref }}&submodules=1#${{ matrix.package }}' -L --extra-substituters "https://hyprland.cachix.org"
+      - run: nix build 'git+https://github.com/hyprwm/Hyprland?ref=${{ github.ref }}#${{ matrix.package }}' -L --extra-substituters "https://hyprland.cachix.org"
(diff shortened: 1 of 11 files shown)


vaxerski (commented):
are there any version differences? How should we differentiate udis outside of the pc file?

Distro packagers could include it without using the correct fork, or could be using a different fork

fufexan:
https://repology.org/project/udis86/versions reveals that most distros use the original project, with the exception being AUR versions of canihavesomecoffee. I think we're safe. I could also add a message telling users to "MAKE SURE TO USE https://github.com/canihavesomecoffee/udis86".

vaxerski:
sure, I guess

vaxerski:
> reveals that most distros use the original project

doesnt mean they dont add a pc file?

fufexan:
> > reveals that most distros use the original project
> 
> doesnt mean they dont add a pc file?

I've looked through the build files and they don't. Only Alpine uses canihavesomecoffee's fork and already patches Hyprland to make use of the system udis86 (https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/hyprland/use-system-installed-udis86.patch).

vaxerski (approved):
pineapple 2.99

PR #7976, opened 2024-10-02
layout: fix dragging a window into a group after e242694

Aqa-Ib:
The last refactor to e242694 missed this but I didn't realize it until now.
oopsie, sorry ^_^''

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -552,10 +552,10 @@ void IHyprLayout::changeWindowFloatingMode(PHLWINDOW pWindow) {
 
         pWindow->m_vLastFloatingSize = PSAVEDSIZE;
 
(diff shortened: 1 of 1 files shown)


src/layout/IHyprLayout.cpp
```diff
@@ -552,10 +552,10 @@ void IHyprLayout::changeWindowFloatingMode(PHLWINDOW pWindow) {
 
         pWindow->m_vLastFloatingSize = PSAVEDSIZE;
```
Aqa-Ib, line 553: I thought about moving this line below `onWindowCreated(pWindow)` because it applies `pWindow->m_vLastFloatingSize` again, but it does not seem to matter, it is remembered anyway.

vaxerski: ye

vaxerski (approved):

PR #8005, opened 2024-10-06
layout: fix auto_group when opening a new window in a non-focused workspace using window rules

Aqa-Ib:
Fixes auto grouping when using for example these window rules:
```
windowrulev2 = group set,class:discord
windowrulev2 = workspace special:chat silent,class:discord

windowrulev2 = group set,class:telegram
windowrulev2 = workspace special:chat silent,class:telegram
```


Aqa-Ib:
I fucked up the diff, will open anew.

PR #8014, opened 2024-10-07
use clampWindowSize to unify min/maxsize handling

MightyPlaza:
Creates  `clampWindowSize` to unify handling of `minsize` and `maxsize`.
Fixes window position not changing when using `minsize`/`maxsize` (causing window to be placed out of screen for large initial size values).
Seems to work without issues.


vaxerski (approved):
thankx

PR #8023, opened 2024-10-07
Support hyprland-ctm-control-v1

vaxerski:
Adds support for the CTM protocol.

Needs a test client. Will do tomorrow. Or today. God knows.

Needs the protocol merged, otherwise wont build.


vaxerski:
@fufexan I need a nix hl-proto bump + probably make meson require 0.4.0 as well.

PR #8034, opened 2024-10-08
improve version query and some code and add HYPRLAND_VERSION

Toni500github:
~~It's suggested to use `getopt.h` for getting the arguments, it's more reliable and safe and fast than the current implementation. It's also POSIX compliant so no need to worry~~

I have added HYPRLAND_VERSION, which can be used in `--version` and could be useful for system info fetches (neofetch, fastfetch, customfetch, ...)

~~So I wanted to make this PR just for using `getopt_long()`, but then~~ I saw that the code has kinda too much spaghetti, we're on >=C++23 and still using `std::cout` instead of `std::print()` and also we got
```c++
std::string result = "Hyprland, built from branch " + std::string(GIT_BRANCH) + " at commit " + GIT_COMMIT_HASH + " " + GIT_DIRTY + " (" + commitMsg +
            ").\nDate: " + GIT_COMMIT_DATE + "\nTag: " + GIT_TAG + ", commits: " + GIT_COMMITS + std::string{"\nbuilt against aquamarine "} + AQUAMARINE_VERSION + "\n" +
            "\n\nflags: (if any)\n";
```
which is comprehensible but unaccetable in modern c++, imho tho ;p
So I've improved some old code, which is still compatible with c++23

The git diff is kinda unclear and difficult to see the difference IMO, so i'd suggest check directly the forked repo if anything.

It's kinda ready, just wanna get your permission to replace every `std::cout` instance with `std::print()` and also improve some format code like the one above I sent.
Once it's all agreed and set, i'll remove `te.sh` and `compile_flags.txt` and ready to merge


vaxerski:
there are many things smashed here as well as unwanted files added

Toni500github:
> there are many things smashed here as well as unwanted files added

yes there were `te.sh` and `compile_flags.txt` because I use them for compiling and developing. as I already said

> Once it's all agreed and set, i'll remove te.sh and compile_flags.txt and ready to merge

Toni500github:
and also what you mean about things being smashed here and there?
If the diffs in main.cpp are unclear, check the file in my forked repo

vaxerski:
you are making many changes that don't belong in one MR:
 - changing std::cout to std::print (why?)
 - using POSIX getopt (honestly, don't see the benefit either) -> "reliable and safe and fast" source?
 - cleaning up redundant escapes in pluginmgr.cpp (ok)
 - refactoring versionRequest (ok)
 - adding a new void version() (why not just call versionRequest??)

Toni500github:
sorry for late reply, but i'll address those points:
> * changing std::cout to std::print (why?)

`std::print()` is easier to use and read, it also has better performance and reduces boilerplate code https://vitaut.net/posts/2023/print-in-cpp23/

> * using POSIX getopt (honestly, don't see the benefit either) -> "reliable and safe and fast" source?

ok It's not really that much faster, but using `getopt.h` makes parsing args more straightforward and simple, and surely it's better than converting all args in a `std::vector<std::string>`. It's also a standard way across multiple projects

> * adding a new void version() (why not just call versionRequest??)

I was thinking about the same too, but you didn't already do it before too https://github.com/hyprwm/Hyprland/blob/main/src/main.cpp#L118 so I thought to leave it as it was :)

vaxerski:
> std::print() is easier to use and read, it also has better performance and reduces boilerplate code https://vitaut.net/posts/2023/print-in-cpp23/

separate MR please.

> ok It's not really that much faster, but using getopt.h makes parsing args more straightforward and simple, and surely it's better than converting all args in a std::vector<std::string>. It's also a standard way across multiple projects

C projects. This isn't C and I don't like it.

> I was thinking about the same too, but you too didn't already do it before https://github.com/hyprwm/Hyprland/blob/main/src/main.cpp#L118 so I thought to leave it as it was :)

well then you have the perfect opportunity to do a small refactor and minimize the duplication

Toni500github:
> C projects. This isn't C and I don't like it.

C++ projects too such as [nsjail](https://github.com/google/nsjail)\
https://github.com/google/nsjail/blob/583bc3eb24a13061539f65aeb0ba08a1a86cc9dc/logs.h#L25

Toni500github:
but ok fine, I'll minimize the code, and change the title and revert `getopt.h` changes

Toni500github:
@vaxerski can be merged, I'm sorry for wasting any time

scripts/generateVersion.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 cp -fr ./src/version.h.in ./src/version.h
 
+HYPRLAND_VERSION=${HYPRLAND_VERSION-$(cat VERSION)}
```
vaxerski, line 4: instead of this, just pass a `add_compile_definitions` in the cmake

vaxerski (commented):
- I think clang-format needed
- you accidentally changed the ref of subprojects/hyprland-protocols, that needs to be reverted

Toni500github:
Ok, when I get back home I'll do it right away

Toni500github:
@vaxerski just came back, hope it's good. I wouldn't suggest merging with all the commits I made in the git history

vaxerski (approved):
thankx

Toni500github:
nice, i'll start working on the `std::cout` -> `std::print()` transition

vaxerski:
clang-format needed cuz test faul

Toni500github:
Next time I'll directly run `clang-format -i $(find . -name "*.hpp" -o -name "*.cpp")`

vaxerski:
that's the way

vaxerski (approved):

PR #8040, opened 2024-10-08
Make hyprerror reserve space

vaxerski:
Makes hyprerror reserve space

@fufexan needs test on mixed dpi

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -97,6 +97,16 @@ CHyprRenderer::CHyprRenderer() {
         ensureCursorRenderingMode();
     });
 
+    static auto P3 = g_pHookSystem->hookDynamic("focusedMon", [&](void* self, SCallbackInfo& info, std::any param) {
+        g_pEventLoopManager->doLater([this]() {
+            if (!g_pHyprError->active())
+                return;
+            for (auto& m : g_pCompositor->m_vMonitors) {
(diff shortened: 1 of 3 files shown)


fufexan (approved):
Works well on mixed-dpi.

vaxerski:
wooo!

src/render/Renderer.cpp
```diff
@@ -1757,6 +1767,13 @@ void CHyprRenderer::arrangeLayersForMonitor(const MONITORID& monitor) {
 
     CBox usableArea = {PMONITOR->vecPosition.x, PMONITOR->vecPosition.y, PMONITOR->vecSize.x, PMONITOR->vecSize.y};
 
+    if (g_pHyprError->active() && g_pCompositor->m_pLastMonitor == PMONITOR->self) {
+        const auto HEIGHT              = g_pHyprError->height();
+        PMONITOR->vecReservedTopLeft.y = HEIGHT;
```
zjeffer, line 1772: I didn't test it but I assume because of this line the reserved space will always be at the top, even if the user chose to move their error bar to the bottom, as implemented in https://github.com/hyprwm/Hyprland/pull/6111 ?

vaxerski: oh, I forgot about that existing.

vaxerski: done

PR #8060, opened 2024-10-09
implement ext image capture source

ikalco:
implements new ext-image-capture-source-v1 protocol

this is the easy half next will be ext-image-copy-capture
I'll prob end up editing this proto when doing above but I didn't want 1 giant pr

<s>could the multiple listeners in CImageCaptureSource cause a race for its dtor and the event it emits?</s> nvm lol

probably yes

edit: works good 👍
just tested and this pr doesn't break toplevel export or hyprland in general

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,6 +334,7 @@ protocolnew("staging/linux-drm-syncobj" "linux-drm-syncobj-v1" false)
 protocolnew("staging/xdg-dialog" "xdg-dialog-v1" false)
 protocolnew("staging/single-pixel-buffer" "single-pixel-buffer-v1" false)
 protocolnew("staging/security-context" "security-context-v1" false)
+protocolnew("staging/ext-image-capture-source" "ext-image-capture-source-v1" false)
 
 protocolwayland()
 
--- a/protocols/meson.build
+++ b/protocols/meson.build
@@ -66,6 +66,7 @@ protocols = [
 	wayland_protocol_dir / 'staging/xdg-dialog/xdg-dialog-v1.xml',
 	wayland_protocol_dir / 'staging/single-pixel-buffer/single-pixel-buffer-v1.xml',
 	wayland_protocol_dir / 'staging/security-context/security-context-v1.xml',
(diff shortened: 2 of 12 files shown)


ikalco:
nix build failed cause of outdated wayland-protocols, needs to be 1.36 -> 1.37
its currently in nix staging but idk what the means, prob just wait

vaxerski:
generally I don't want this implemented yet tbh. Some reasons:
 - I haven't seen a single client
 - I want to implement a verification framework (akin to xdph where it prompts you) since this is a general protocol

ikalco:
some clients ([wl-screenrec](https://github.com/russelltg/wl-screenrec), [wayvnc](https://github.com/any1/wayvnc/tree/master)) have the code in master, just not released yet
also a guy made a [test client](https://github.com/mahkoh/wayland-copy-capture-test-client/tree/master)

however, looking at some recent issues (212, 220-223 in wl protos issue tracker) with the image-copy-capture, its probably best to wait to implement it

as for a verification framework, id be willing to write it, but
* would this consist a yes/no qt app that's called when a source is created? or something like hyprerrors with a button
* would xdph, when using this protocol, end up having 2 prompts (1st -> choose, 2nd -> yes/no allow)
* ^ if not, would we keep track of xdph wl client id to not ask? or maybe some special token, but how would this proto get the token
* would screencopy and toplevel export also use verification, or we just deprecate them once this becomes more popular

vaxerski:
> as for a verification framework, id be willing to write it, but

don't. I'm waiting for @outfoxxed to finish at least the first parts of shared qml. I'll write the prompt, feel free to improve it later, but I have a baseline decided and it needs to be done according to it.

> would xdph, when using this protocol, end up having 2 prompts (1st -> choose, 2nd -> yes/no allow)

No. On linux, we can track it easily with /proc. On BSD... idk, we can just say fuck it

> would screencopy and toplevel export also use verification, or we just deprecate them once this becomes more popular

Yes.

Obviously verification will be allowed to be disabled for specific binaries in the config, e.g. grim.

PR #8072, opened 2024-10-11
layouts: Nuke no_gaps_when_only

vaxerski:
NGWO has been deprecated for a while in favor of workspace / window rules that achieve the same thing with less bugs.

Can we finally just nuke it as people keep using it?

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1447,12 +1447,6 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .data        = SConfigOptionDescription::SFloatData{1, 0.1, 3},
     },
     SConfigOptionDescription{
-        .value       = "dwindle:no_gaps_when_only",
-        .description = "whether to apply gaps when there is only one window on a workspace, aka. smart gaps. (default: disabled - 0) no border - 1, with border - 2 [0/1/2]",
-        .type        = CONFIG_OPTION_CHOICE,
-        .data        = SConfigOptionDescription::SChoiceData{0, "disabled,no border,with border"},
-    },
-    SConfigOptionDescription{
         .value       = "dwindle:use_active_for_splits",
         .description = "whether to prefer the active window or the mouse position for splits",
         .type        = CONFIG_OPTION_BOOL,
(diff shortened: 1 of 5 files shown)


gksudolol:
Rip it out completely. The same functionality can be replicated with:

```
workspace = w[t1], gapsin:0, gapsout:0, border:0
```

Eclextic:
> Rip it out completely. The same functionality can be replicated with:
> 
> ```
> workspace = w[t1], gapsin:0, gapsout:0, border:0
> ```

Isn't that specifically only for the first workspace then?
So that means if you do the same on the second workspace or a named workspace it won't work?

I really don't care about this feature anyways so... :3

leiserfg:
I use this rule, and it's not buggy for me, is there a replacement that does not imply having to set a rule per workspace?

gksudolol:
> Isn't that specifically only for the first workspace then?

`t` Is for tiled-only clients i.e non-floating/fullscreen. Setting it to 1 because we don't want it to apply if there's more than 1 window in the workspace

> So that means if you do the same on the second workspace or a named workspace it won't work?

By default, that snippet will be all workspaces. But you can set a number, range or named:

```
# Only workspace 1
workspace = 1, w[t1], gapsin:0, gapsout:0, border:0

# Only workspaces from 1 to 3
workspace = r[1-3], w[t1], gapsin:0, gapsout:0, border:0

# Only a specific named workspace
workspace = name:myworkspace, w[t1], gapsin:0, gapsout:0, border:0
```

Honkazel:
> Rip it out completely. The same functionality can be replicated with:
> 
> ```
> workspace = w[t1], gapsin:0, gapsout:0, border:0
> ```

`rounding=0`?

vaxerski:
> But please keep a comment somewhere about how to achieve this.

https://wiki.hyprland.org/Configuring/Workspace-Rules/#smart-gaps

vaxerski:
thanks for the input

nktnet1:
To fully replicate `no_gaps_when_only: true`, In addition to
```
workspace = w[t1], gapsin:0, gapsout:0, border:0
```

I also had to add
```
windowrulev2 = noborder, fullscreen:1
```
for maximised windows when there is more than one in a workspace.

vaxerski:
```
workspace = w[t1], gapsin:0, gapsout:0, border:0
workspace = w[tg1], gapsin:0, gapsout:0, border:0
workspace = f[1], gapsin:0, gapsout:0, border:0
```

As listed on the wiki, enjoy

kRHYME7:
Hi! How can I possibly have my borders/etc back for floating mode.  (Looks weird when BG are the same)

image

vaxerski:
```ini
workspace = w[t1], gapsout:0, gapsin:0
workspace = w[tg1], gapsout:0, gapsin:0
workspace = f[1], gapsout:0, gapsin:0
windowrulev2 = bordersize 0, floating:0, onworkspace:w[t1]
windowrulev2 = rounding 0, floating:0, onworkspace:w[t1]
windowrulev2 = bordersize 0, floating:0, onworkspace:w[tg1]
windowrulev2 = rounding 0, floating:0, onworkspace:w[tg1]
windowrulev2 = bordersize 0, floating:0, onworkspace:f[1]
windowrulev2 = rounding 0, floating:0, onworkspace:f[1]
```

JohnMH:
It would generally be a good idea to have a deprecation warning before a setting/feature is removed so end users are aware.

vaxerski:
as discussed in another issue (I forgot the #) I put those in the release notes.

PR #8083, opened 2024-10-11
Add check for conflicting acquire and release points

Edgars-Cirulis:
adds a check to prevent conflicts between acquire and release points when they are on the same timeline

No known issues or breaking changes

It's ready for merging, I guess.

### DIFF
--- a/src/protocols/DRMSyncobj.cpp
+++ b/src/protocols/DRMSyncobj.cpp
@@ -55,8 +55,13 @@ CDRMSyncobjSurfaceResource::CDRMSyncobjSurfaceResource(SP<CWpLinuxDrmSyncobjSurf
             return;
         }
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
I swear to god I thought we had that. lol

Edgars-Cirulis:
I have a question. I've noticed that flickering occurs whenever you use the mouse, but it doesn't happen when you use only keyboard inputs, like moving around in Minecraft. Does the explicit sync fence gets cleared during mouse movement? I'm still unclear about the source code and still need to learn where everything is.

vaxerski:
> Does the explicit sync fence gets cleared during mouse movement? 

Nope, nothing should be different except for frame pacing, but that shouldn't matter that much... I think

PR #3236, opened 2023-09-10
Add snapping to floating windows

shadowmax31:
#### Describe your PR, what does it fix/add?
Adds a floating window snapping option: https://github.com/hyprwm/Hyprland/issues/3230

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

It adds 3 configuration options:
1. **misc:snap_floating**: Enable snapping to either the windows or the monitor's edge (values: "", monitor, windows)
2. **misc:snap_floating_strength**: determines the distance at which snapping occurs
3. **misc:snap_floating_outside**: if false, windows will only snap to the inside of windows/monitor. If true, windows will snap inside and outside. I disabled it by default, because it reacts a bit weird depending on gap size and strength. Especially if strength is bigger then gap size.

#### Is it ready for merging, or does it need work?
Ready

PR #8088, opened 2024-10-11
add snapping to floating windows

myQwil:

While moving and resizing floating windows, edges and corners will snap to the edges/corners of the monitor or other windows when they get close enough to each other.

It started as an update to @shadowmax31 's PR #3236 , but was eventually changed to behave more similarly to KWin's snapping feature.

It comes with 3 options:

`general:snap:enabled` - whether it's enabled, off by default

`general:snap:window_gap` - minimum gap in pixels between windows before snapping. Setting to 0 effectively turns off this method of snapping.

`general:snap:monitor_gap` - minimum gap in pixels between window and monitor edges before snapping. Again, setting it to 0 effectively turns it off.

Here's a video of it with these settings:
```sh
general:snap {
    enabled = true
    window_gap = 25 # pixels
    monitor_gap = 10 # pixels
}
```

https://github.com/user-attachments/assets/8245a2bf-ff50-4722-b267-e20dfa77c519

Not that I can think of.

Ready to merge.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -396,71 +396,116 @@ void IHyprLayout::onEndDragWindow() {
     g_pInputManager->m_bWasDraggingWindow = false;
 }
 
-inline bool isInSnappingDistance(const double sideA, const double sideB, const double gap) {
+inline bool canSnap(const double sideA, const double sideB, const double gap) {
     return sideB - gap < sideA && sideA < sideB + gap;
 }
 
-// Inner-snap oriented. For outer-snap, swap p1 and p2.
-void snapMove(double& pos, double& len, const double p1, const double p2, const double gap) {
-    if (isInSnappingDistance(pos, p1, gap))
-        pos = p1;
-    else if (isInSnappingDistance(pos + len, p2, gap))
-        pos = p2 - len;
+void snapMoveLeft(double& pos, double& len, const double p) {
+    pos = p;
+}
+void snapMoveRight(double& pos, double& len, const double p) {
+    pos = p - len;
 }
 
-// Inner-snap oriented. For outer-snap, swap p1 and p2.
-void snapResize(double& pos, double& len, const double p1, const double p2, const double gap) {
-    if (isInSnappingDistance(pos, p1, gap)) {
-        len += pos - p1;
-        pos = p1;
+void snapResizeLeft(double& pos, double& len, const double p) {
+    len += pos - p;
+    pos = p;
+}
+void snapResizeRight(double& pos, double& len, const double p) {
+    len = p - pos;
+}
+
+typedef void (*SnapFn)(double& pos, double& len, const double p);
+
+void snapWindows(CBox& wb, PHLWINDOW DRAGGINGWINDOW, const double oldGap, const eMouseBindMode mode, const eRectCorner c) {
+    SnapFn snapUp, snapDown, snapLeft, snapRight;
+    switch (mode) {
+        case MBIND_RESIZE:
+            snapUp = snapLeft = snapResizeLeft;
+            snapDown = snapRight = snapResizeRight;
+            break;
+        default:
+            snapUp = snapLeft = snapMoveLeft;
+            snapDown = snapRight = snapMoveRight;
+            break;
     }
-    if (isInSnappingDistance(pos + len, p2, gap))
-        len = p2 - pos;
-}
 
-typedef void (*SnapFn)(double& pos, double& len, const double p1, const double p2, const double gap);
+    static auto  BORDERSIZE = CConfigValue<Hyprlang::INT>("general:border_size");
+    const int    bord       = *BORDERSIZE;
+    const double gap        = oldGap + bord;
 
-void snapWindows(SnapFn snap, CBox& wb, PHLWINDOW DRAGGINGWINDOW, const double gap, const eRectCorner c) {
-    const auto PID  = DRAGGINGWINDOW->getPID();
-    const auto WSID = DRAGGINGWINDOW->workspaceID();
+    const auto   PID  = DRAGGINGWINDOW->getPID();
+    const auto   WSID = DRAGGINGWINDOW->workspaceID();
 
     for (auto& other : g_pCompositor->m_vWindows) {
         if (other->workspaceID() != WSID || other->getPID() == PID || !other->m_bIsMapped || other->m_bFadingOut || other->m_bIsX11)
             continue;
-
-        const CBox     ob = other->getWindowMainSurfaceBox();
-        const Vector2D ov = ob.extent();
-        Vector2D       wv = wb.extent();
+        const Vector2D wv  = wb.extent();
+        const CBox     box = other->getWindowMainSurfaceBox();
+        const CBox     ob  = {box.x, box.y, box.x + box.w, box.y + box.h};
+        const CBox     bb  = {ob.x - bord, ob.y - bord, ob.w + bord, ob.h + bord};
 
         // only snap windows if their ranges intersect in the opposite axis
-        if (wb.y <= ov.y && ob.y <= wv.y) {
-            snap(wb.x, wb.w, ov.x, ob.x, gap);
-            if (wb.x == ov.x || ob.x == wb.x + wb.w) {
-                if (c) {
-                    if ((c == CORNER_TOPLEFT || c == CORNER_TOPRIGHT) && isInSnappingDistance(wb.y, ob.y, gap)) {
-                        wb.h += wb.y - ob.y;
-                        wb.y = ob.y;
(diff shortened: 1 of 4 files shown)


src/layout/IHyprLayout.cpp
```diff
@@ -443,6 +451,67 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
     if (g_pInputManager->dragMode == MBIND_MOVE) {
 
         CBox wb = {m_vBeginDragPositionXY + DELTA, DRAGGINGWINDOW->m_vRealSize.goal()};
+        if (*SNAPENABLED && !DRAGGINGWINDOW->m_bDraggingTiled) {
```
vaxerski, line 454: arent these two if blocks the same? can't we combine them?

(this and the one further down)

src/layout/IHyprLayout.cpp
```diff
@@ -505,6 +574,75 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
                 newPos = newPos + Vector2D((m_vBeginDragSizeXY - newSize).x, 0.0);
 
             CBox wb = {newPos, newSize};
+            if (*SNAPENABLED) {
+                if (*SNAPWINDOWGAP) {
+                    const int  gap  = *SNAPWINDOWGAP;
+                    const auto PID  = DRAGGINGWINDOW->getPID();
+                    const auto WSID = DRAGGINGWINDOW->workspaceID();
+
+                    for (auto& other : g_pCompositor->m_vWindows) {
```
vaxerski, line 583: needs to ignore windows that:
 - aren't mapped
 - are fading out
 - are XWayland OR

vaxerski (commented):
neat feat

src/layout/IHyprLayout.cpp
myQwil: If `SNAPWINDOWGAP` is zero, there's still `SNAPMONITORGAP` a bit further down.

My thinking was that `enabled` would act as the master switch, and then different snapping methods could be turned on or off with zero/non-zero values.

src/layout/IHyprLayout.cpp
myQwil: something like this?
```cpp
if (other->workspaceID() == WSID && other->getPID() != PID
 && other->m_bIsMapped
 && !other->m_bFadingOut
 && !other->m_bIsX11) {
```

vaxerski: prefer guards.

```cpp
if (bad)
   continue;
```

src/layout/IHyprLayout.cpp
vaxerski: no I mean this entire block and the one at line 577

myQwil: This one is for moving and the one at line 577 is for resizing. They're similar, admittedly. I'm currently working on some refactoring to make both sections more brief.

vaxerski:
tag me for review when ready

src/layout/IHyprLayout.cpp
```diff
@@ -396,6 +396,139 @@ void IHyprLayout::onEndDragWindow() {
...
 }
 
+inline bool canSnap(const double sideA, const double sideB, const double gap) {
+    return std::abs(sideA - sideB) < gap;
+}
+
+void snapMoveLeft(double& pos, double& len, const double p) {
+    pos = p;
+}
+void snapMoveRight(double& pos, double& len, const double p) {
```
vaxerski, line 406: keep one empty newline between functions

src/layout/IHyprLayout.cpp
```diff
@@ -396,6 +396,139 @@ void IHyprLayout::onEndDragWindow() {
...
+
+void snapResizeLeft(double& pos, double& len, const double p) {
+    len += pos - p;
+    pos = p;
+}
+void snapResizeRight(double& pos, double& len, const double p) {
+    len = p - pos;
+}
+
+typedef void (*SnapFn)(double& pos, double& len, const double p);
```
vaxerski, line 418: at this point use a std::function typedef and make all 4 fns static std::function<>s

src/layout/IHyprLayout.cpp
```diff
@@ -396,6 +396,139 @@ void IHyprLayout::onEndDragWindow() {
...
+
+void IHyprLayout::performSnap(Vector2D& pos, Vector2D& size, PHLWINDOW DRAGGINGWINDOW, const eMouseBindMode mode) {
+    static auto SNAPWINDOWGAP  = CConfigValue<Hyprlang::INT>("general:snap:window_gap");
+    static auto SNAPMONITORGAP = CConfigValue<Hyprlang::INT>("general:snap:monitor_gap");
+
+    int         corner, snaps = 0;
```
vaxerski, line 424: !!!!! undefined variable

src/layout/IHyprLayout.cpp
```diff
@@ -396,6 +396,139 @@ void IHyprLayout::onEndDragWindow() {
...
+    SnapFn      snapUp, snapDown, snapLeft, snapRight;
```
vaxerski, line 425: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

src/layout/IHyprLayout.cpp
```diff
@@ -396,6 +396,139 @@ void IHyprLayout::onEndDragWindow() {
...
+        snapDown = snapRight = snapResizeRight;
+    }
+
+    if (*SNAPWINDOWGAP) {
+        const auto PID  = DRAGGINGWINDOW->getPID();
+        const auto WSID = DRAGGINGWINDOW->workspaceID();
+        const int  BORD = DRAGGINGWINDOW->getRealBorderSize();
+
+        for (auto& other : g_pCompositor->m_vWindows) {
+            if (other->workspaceID() != WSID || other->getPID() == PID || !other->m_bIsMapped || other->m_bFadingOut || other->m_bIsX11)
```
vaxerski, line 442: why are we disabling snapping to X11? Only OR should be skipped. `CWindow::isX11OverrideRedirect()`

src/layout/IHyprLayout.cpp
```diff
@@ -396,6 +396,139 @@ void IHyprLayout::onEndDragWindow() {
...
+                continue;
+
+            const int      bord = std::max(BORD, other->getRealBorderSize());
+            const double   gap  = *SNAPWINDOWGAP + bord;
+
+            const CBox     box = other->getWindowMainSurfaceBox();
+            const CBox     ob(box.x, box.y, box.x + box.w, box.y + box.h);
```
vaxerski, line 449: we use {}

src/layout/IHyprLayout.hpp
```diff
@@ -18,11 +19,18 @@ struct SLayoutMessageHeader {
...
-    CORNER_BOTTOMLEFT
+    CORNER_NONE        = 0,
+    CORNER_TOPLEFT     = 1,
+    CORNER_TOPRIGHT    = 2,
+    CORNER_BOTTOMRIGHT = 4,
+    CORNER_BOTTOMLEFT  = 8,
+};
+
+enum eSnapEdge {
+    SNAP_UP    = 1,
```
vaxerski, line 30: SNAP_INVALID = 0

src/layout/IHyprLayout.hpp
```diff
@@ -18,11 +19,18 @@ struct SLayoutMessageHeader {
 enum eFullscreenMode : int8_t;
 
 enum eRectCorner {
-    CORNER_NONE = 0,
-    CORNER_TOPLEFT,
-    CORNER_TOPRIGHT,
-    CORNER_BOTTOMRIGHT,
-    CORNER_BOTTOMLEFT
+    CORNER_NONE        = 0,
+    CORNER_TOPLEFT     = 1,
```
vaxerski, line 23: here and under, use the bitshift notation `(1 << 0)` `(1 << 1)` etc

src/layout/IHyprLayout.hpp
```diff
@@ -206,4 +214,6 @@ class IHyprLayout {
     eRectCorner  m_eGrabbedCorner = CORNER_TOPLEFT;
 
     PHLWINDOWREF m_pLastTiledWindow;
+
+    void         performSnap(Vector2D& pos, Vector2D& size, PHLWINDOW DRAGGINGWINDOW, const eMouseBindMode mode);
```
vaxerski, line 218: at this point, does it need to be a member? Can we just make it `static` in .cpp?

src/layout/IHyprLayout.cpp
myQwil: I declare the variables, and then define them in the if-else clause that immediately follows. That way, only 1 condition needs to be checked to assign multiple variables. They'll never be undefined when actually being called upon.

vaxerski (approved):
lgtm, wiki mr needed

Issue #7271, opened 2024-08-10
hyprctl devices: Getting caps lock and num lock status


KanishakVaidya:
### Description

To get the information about keyboard devices we can do something like
```
hyprctl devices -j
```
getting the following information about every keyboard device:
```
...
{
  "address": "0x<the address>",
  "name": "<the name>",
  "rules": "",
  "model": "",
  "layout": "us",
  "variant": "",
  "options": "",
  "active_keymap": "English (US)",
  "main": true/false
}
...
```
Can we also get the information about the caps lock and num lock status. Weather currently the caps lock is ON or OFF. Or maybe some other way to achieve the same (get caps/num lock status).

vaxerski:
MRs welcome. The current mod state is stored in IKeyboard

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #8145, opened 2024-10-17
hyprctl devices: Get caps/num lock state

czlabinger:
#7271 
Adds caps/num lock state to `hyprctl devices -j`

```
...
{
  "address": "0x<the address>",
  "name": "<the name>",
  "rules": "",
  "model": "",
  "layout": "us",
  "variant": "",
  "options": "",
  "active_keymap": "English (US)",
  "capsLock": true/false,
  "numLock": true/false,
  "main": true/false
}
...
```

No

Ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PREFIX = /usr/local
 
 legacyrenderer:
-	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -DLEGACY_RENDERER:BOOL=true -S . -B ./buildZ
+	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -DLEGACY_RENDERER:BOOL=true -S . -B ./build
 	cmake --build ./build --config Release --target all -j`nproc 2>/dev/null || getconf NPROCESSORS_CONF`
 
 legacyrendererdebug:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PREFIX = /usr/local
 
 legacyrenderer:
-	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -DLEGACY_RENDERER:BOOL=true -S . -B ./build
+	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -DLEGACY_RENDERER:BOOL=true -S . -B ./buildZ
 	cmake --build ./build --config Release --target all -j`nproc 2>/dev/null || getconf NPROCESSORS_CONF`
 
 legacyrendererdebug:
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -586,7 +586,7 @@ std::string devicesRequest(eHyprCtlOutputFormat format, std::string request) {
     }},)#",
                 (uintptr_t)k.get(), escapeJSONStrings(k->hlName), escapeJSONStrings(k->currentRules.rules), escapeJSONStrings(k->currentRules.model),
                 escapeJSONStrings(k->currentRules.layout), escapeJSONStrings(k->currentRules.variant), escapeJSONStrings(k->currentRules.options), escapeJSONStrings(KM),
-                (k->modifiersState.locked & (1 << 1) : "true" ? "false"), (k->modifiersState.locked & (1 << 4) : "true" ? "false"), (k->active ? "true" : "false"));
+                ((k->modifiersState.locked & (1 << 1)) == 2 ? "true" : "false"), ((k->modifiersState.locked & (1 << 4)) == 16 ? "true" : "false"), (k->active ? "true" : "false"));
         }
 
         trimTrailingComma(result);
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -580,13 +580,13 @@ std::string devicesRequest(eHyprCtlOutputFormat format, std::string request) {
         "variant": "{}",
         "options": "{}",
         "active_keymap": "{}",
-        "capsLock: {}",
-        "numLock: {}",
+        "capsLock": {},
+        "numLock": {},
         "main": {}
     }},)#",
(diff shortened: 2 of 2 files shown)


vaxerski (commented):
the JSON is not valid. It should be:
```json
"capsLock": true,
"numLock": false,
```

src/debug/HyprCtl.cpp
```diff
@@ -580,11 +580,13 @@ std::string devicesRequest(eHyprCtlOutputFormat format, std::string request) {
...
         "options": "{}",
         "active_keymap": "{}",
+        "capsLock": {},
+        "numLock": {},
         "main": {}
     }},)#",
                 (uintptr_t)k.get(), escapeJSONStrings(k->hlName), escapeJSONStrings(k->currentRules.rules), escapeJSONStrings(k->currentRules.model),
                 escapeJSONStrings(k->currentRules.layout), escapeJSONStrings(k->currentRules.variant), escapeJSONStrings(k->currentRules.options), escapeJSONStrings(KM),
-                (k->active ? "true" : "false"));
+                ((k->getLEDs().value_or(0) & (1 << 1)) == 2 ? "true" : "false"), ((k->getLEDs().value_or(0) & (1 << 0)) ? "true" : "false"), (k->active ? "true" : "false"));
```
vaxerski, line 589: why LEDs? Just check the modifier mask in the current mod state of the keyboard (`modifiersState`)

LEDs are not reliable

Makefile
```diff
@@ -1,7 +1,7 @@
 PREFIX = /usr/local
 
 legacyrenderer:
-	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -DLEGACY_RENDERER:BOOL=true -S . -B ./buildZ
+	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -DLEGACY_RENDERER:BOOL=true -S . -B ./build
```
vaxerski, line 4: unintended change?

src/debug/HyprCtl.cpp
```diff
@@ -580,11 +580,13 @@ std::string devicesRequest(eHyprCtlOutputFormat format, std::string request) {
         "variant": "{}",
         "options": "{}",
         "active_keymap": "{}",
+        "capsLock": {},
+        "numLock": {},
         "main": {}
     }},)#",
                 (uintptr_t)k.get(), escapeJSONStrings(k->hlName), escapeJSONStrings(k->currentRules.rules), escapeJSONStrings(k->currentRules.model),
                 escapeJSONStrings(k->currentRules.layout), escapeJSONStrings(k->currentRules.variant), escapeJSONStrings(k->currentRules.options), escapeJSONStrings(KM),
-                (k->active ? "true" : "false"));
```
vaxerski, line 587: oh, you forgot to add the props to the non-json output a few dozen lines below. Other than that, lgtm

MahouShoujoMivutilde:
Are you sure that these 1 and 4 is something that won't ever change / be different for some keyboard?

As far as I can see it's just the order from

https://github.com/hyprwm/Hyprland/blob/08cc063e175e48cea44d26b7e3762f4b8611f0c5/src/devices/IKeyboard.cpp#L13-L15

And here

<details><summary>Details</summary>
<p>

https://github.com/hyprwm/Hyprland/blob/08cc063e175e48cea44d26b7e3762f4b8611f0c5/src/devices/IKeyboard.cpp#L131-L134

```log
[LOG] ApplyConfigToKeyboard for "logitech-usb-keyboard", hasconfig: 1
[LOG] Attempting to create a keymap for layout us,ru with variant  (rules: , model: , options: caps:escape,grp:toggle)
[LOG] Updating keyboard 6450bb5c1b00's translation state from a provided keymap
[LOG] xkb: LED index 0 (name Num Lock) got index 1
[LOG] xkb: LED index 1 (name Caps Lock) got index 0
[LOG] xkb: LED index 2 (name Scroll Lock) got index 2
[LOG] xkb: Mod index 0 (name Shift) got index 0
[LOG] xkb: Mod index 1 (name Lock) got index 1
[LOG] xkb: Mod index 2 (name Control) got index 2
[LOG] xkb: Mod index 3 (name Mod1) got index 3
[LOG] xkb: Mod index 4 (name Mod2) got index 4
[LOG] xkb: Mod index 5 (name Mod3) got index 5
[LOG] xkb: Mod index 6 (name Mod4) got index 6
[LOG] xkb: Mod index 7 (name Mod5) got index 7
[LOG] Updating keymap fd for keyboard Logitech USB Keyboard
```

</p>
</details>

it seems to match for me.

But will that always hold true? What if some mod is absent?

vaxerski:
hm, right, worth checking...

czlabinger:
So would this be fine?
```cpp
((k->modifiersState.locked & (1 << xkb_map_mod_get_index(k->xkbKeymap, XKB_MOD_NAME_CAPS))) == 2 ? "true" : "false"), 
((k->modifiersState.locked & (1 << xkb_map_mod_get_index(k->xkbKeymap, XKB_MOD_NAME_NUM))) == 16 ? "true" : "false")
```

MahouShoujoMivutilde:
I think so, yes.

But instead of `== 16` and `== 2`, which are tied to an index, have `> 0`. 

As we do a bitwise AND with a number that's 1 with _INDEX_ zeros after in binary (`1 << INDEX`), either there is 1 on INDEX position in `.locked` and the result will be positive, or there is 0 on that position, so the rest doesn't matter and will be 0 after `&`.

src/debug/HyprCtl.cpp
```diff
@@ -580,10 +580,20 @@ std::string devicesRequest(eHyprCtlOutputFormat format, std::string request) {
         "variant": "{}",
         "options": "{}",
         "active_keymap": "{}",
+        "capsLock": {},
+        "numLock": {},
         "main": {}
     }},)#",
                 (uintptr_t)k.get(), escapeJSONStrings(k->hlName), escapeJSONStrings(k->currentRules.rules), escapeJSONStrings(k->currentRules.model),
                 escapeJSONStrings(k->currentRules.layout), escapeJSONStrings(k->currentRules.variant), escapeJSONStrings(k->currentRules.options), escapeJSONStrings(KM),
+                (((xkb_keymap_mod_get_index(k->xkbKeymap, XKB_MOD_NAME_CAPS)) != XKB_MOD_INVALID &&
```
vaxerski, line 589: can this please be made into a small lambda that takes a SP to the keyboard and the mod name, this is unreadable

put the lambda in the beginning of the func and use it for both json and normal

```cpp
auto getModState = [](SP<IKeyboard> keyboard, uint32_t xkbModName) -> std::string {
   /// ... bla bla bla
   return isEnabled ? "true" : "false";
};
```

then just `getModState(k, XKB_MOD_NAME_CAPS)` for example

vaxerski (approved):
thanks!

PR #8159, opened 2024-10-18
compositor: add group_on_movetoworkspace

Aqa-Ib:
Adds the option "group_on_movetoworkspace" to be able to configure whether using movetoworkspace[silent] will merge the window into the workspace's solitary unlocked group. [Wiki.](https://github.com/hyprwm/hyprland-wiki/pull/812)

Fixes https://github.com/hyprwm/Hyprland/issues/8135

Ready.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2750,6 +2750,7 @@ void CCompositor::moveWindowToWorkspaceSafe(PHLWINDOW pWindow, PHLWORKSPACE pWor
         setWindowFullscreenInternal(pWindow, FSMODE_NONE);
 
     PHLWINDOW  pFirstWindowOnWorkspace = g_pCompositor->getFirstWindowOnWorkspace(pWorkspace->m_iID);
+    int        windowsOnWorkspace      = g_pCompositor->getWindowsOnWorkspace(pWorkspace->m_iID, std::nullopt, std::optional<bool>(true));
     const auto PWINDOWMONITOR          = g_pCompositor->getMonitorFromID(pWindow->m_iMonitorID);
     const auto POSTOMON                = pWindow->m_vRealPosition.goal() - PWINDOWMONITOR->vecPosition;
     const auto PWORKSPACEMONITOR       = g_pCompositor->getMonitorFromID(pWorkspace->m_iMonitorID);
@@ -2761,8 +2762,8 @@ void CCompositor::moveWindowToWorkspaceSafe(PHLWINDOW pWindow, PHLWORKSPACE pWor
     pWindow->m_iMonitorID = pWorkspace->m_iMonitorID;
 
     static auto PGROUPONMOVETOWORKSPACE = CConfigValue<Hyprlang::INT>("group:group_on_movetoworkspace");
-    if (*PGROUPONMOVETOWORKSPACE && pFirstWindowOnWorkspace && pFirstWindowOnWorkspace != pWindow && pFirstWindowOnWorkspace->m_sGroupData.pNextWindow.lock() &&
-        pWindow->canBeGroupedInto(pFirstWindowOnWorkspace)) {
+    if (*PGROUPONMOVETOWORKSPACE && windowsOnWorkspace == 1 && pFirstWindowOnWorkspace && pFirstWindowOnWorkspace != pWindow &&
+        pFirstWindowOnWorkspace->m_sGroupData.pNextWindow.lock() && pWindow->canBeGroupedInto(pFirstWindowOnWorkspace)) {
 
         pWindow->m_bIsFloating = pFirstWindowOnWorkspace->m_bIsFloating; // match the floating state. Needed to group tiled into floated and vice versa.
         if (!pWindow->m_sGroupData.pNextWindow.expired()) {
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -34,6 +34,7 @@
 #include "render/Renderer.hpp"
 #include "xwayland/XWayland.hpp"
 #include "helpers/ByteOperations.hpp"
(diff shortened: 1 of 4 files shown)


Aqa-Ib:
This is somewhat alike the previous behavior that dwindle had. Although I am wondering, should we check if the group is the only window in the workspace? Maybe that makes more sense?

Aqa-Ib:
Done, now its behavior is consistent.
Ready to merge if you like how it works.

vaxerski:
I think it should only merge if the group is the only thing on the ws

Aqa-Ib:
> I think it should only merge if the group is the only thing on the ws

It is done, we are checking if there is only one window in the workspace.

src/Compositor.cpp
```diff
@@ -2748,21 +2749,47 @@ void CCompositor::moveWindowToWorkspaceSafe(PHLWINDOW pWindow, PHLWORKSPACE pWor
     if (FULLSCREEN)
         setWindowFullscreenInternal(pWindow, FSMODE_NONE);
 
-    if (!pWindow->m_bIsFloating) {
+    PHLWINDOW  pFirstWindowOnWorkspace = g_pCompositor->getFirstWindowOnWorkspace(pWorkspace->m_iID);
+    int        windowsOnWorkspace      = g_pCompositor->getWindowsOnWorkspace(pWorkspace->m_iID, std::nullopt, std::optional<bool>(true));
```
vaxerski, line 2753: std::optional here not needed, just `true`

also shouldnt this be const

src/layout/IHyprLayout.cpp
```diff
@@ -352,7 +353,7 @@ void IHyprLayout::onEndDragWindow() {
                     DRAGGINGWINDOW->m_bDraggingTiled    = false;
                 }
 
-                if (DRAGGINGWINDOW->m_sGroupData.pNextWindow.lock()) {
+                if (!DRAGGINGWINDOW->m_sGroupData.pNextWindow.expired()) {
```
vaxerski, line 356: ! ... .expired() redundant

if (DRAGGINGWINDOW->m_sGroupData.pNextWindow)

Aqa-Ib: Pardon my ignorance; would you mind briefly explaining me why please? (I have changed it already though).

src/Compositor.cpp
```diff
@@ -2748,21 +2749,47 @@ void CCompositor::moveWindowToWorkspaceSafe(PHLWINDOW pWindow, PHLWORKSPACE pWor
     if (FULLSCREEN)
         setWindowFullscreenInternal(pWindow, FSMODE_NONE);
 
-    if (!pWindow->m_bIsFloating) {
+    PHLWINDOW  pFirstWindowOnWorkspace   = g_pCompositor->getFirstWindowOnWorkspace(pWorkspace->m_iID);
```
Aqa-Ib, line 2752: should this PHLWINDOW be a const too?

src/Compositor.cpp
```diff
@@ -2748,21 +2749,47 @@ void CCompositor::moveWindowToWorkspaceSafe(PHLWINDOW pWindow, PHLWORKSPACE pWor
     if (FULLSCREEN)
         setWindowFullscreenInternal(pWindow, FSMODE_NONE);
 
-    if (!pWindow->m_bIsFloating) {
+    const PHLWINDOW pFirstWindowOnWorkspace   = g_pCompositor->getFirstWindowOnWorkspace(pWorkspace->m_iID);
```
Aqa-Ib, line 2752: I made this PHLWINDOW a const too. I guess it doesn't matter? do you prefer it like this?

src/layout/IHyprLayout.cpp
vaxerski: operator bool is just better reverse expired

src/Compositor.cpp
vaxerski: it k

src/layout/IHyprLayout.cpp
Aqa-Ib: I don't understand why we don't need to check for the shared pointer here?

Aqa-Ib: May as well just leave it as .lock(), I changed it to expired() because I saw it in other parts of the code base, but as you can see I am noob at C++ memory management.

vaxerski: `.lock()` was there cuz I did a bulk replace

Aqa-Ib: No, I wrote that entire block from scratch two weeks ago: https://github.com/hyprwm/Hyprland/commit/45e82199fb29f422ebbbfe06ce03022dfb6d645e. I wrote .lock() instead of .expired() by ignorance. Now that I have learnt more about Hyprland's C++ code, I think that .expired() would be better suited there, that's why I have requested this change. However, you are the expert here, now that you know the context of the situation, you decide what fits better here. Cheers.

vaxerski: in hyprutils, .expired and operator bool are two different checks, but here will behave the same. That's why the shorter version is better.

`.expired()` checks if it's not possible to make a `SP` anymore.
`operator bool` checks if the pointer is still reference-able.

Difference is in the ~dtor of an object that is owned by a SP. You can't make SPs anymore (because that would resurrect the container) but you can still reference the pointer via `->` because the actual object is not deleted yet.

Aqa-Ib: Done. Thank you very much for your explanation and your patience, really appreciate them.

vaxerski (approved):
dzięki

PR #8164, opened 2024-10-18
config: fix generateConfig loop

PaideiaDilemma:

Two of my friends tried out hyprland recently and got into a infinite loop, where it would write the autogenerated config, trigger the hot reload and write the autogenerated config again.

I struggled quite a bit to find out why that happens, but it turns out that it is because this paths var was declared static. So there weren't any subsequent calls to `findConfig` at all.

I did search for issues but did not find one, although i vaguely remember that someone described this problem somewhere.

No

Yes


vaxerski (approved):
thanks!

PR #8167, opened 2024-10-18
hyprpm: Add option to notify on fail and keep original notify

KAGEYAM4:
Hyprpm fail/pass notification are mutually exclusive.

~~By only generating fail notifications, we reduce UI clutter.~~
Add additional `-nn` option to generate notification only on fail events.

Yes

### DIFF
--- a/hyprpm/src/main.cpp
+++ b/hyprpm/src/main.cpp
@@ -23,7 +23,8 @@ constexpr std::string_view HELP = R"#(┏ hyprpm, a Hyprland Plugin Manager
 ┃
 ┣ Flags:
 ┃
-┣ --notify       | -n    → Send a hyprland notification for important events (e.g. load fail)
+┣ --notify       | -n    → Send a hyprland notification for important events (including both successes and fail events)
+┣ --notify-fail  | -nn   → Send a hyprland notification for fail events only
 ┣ --help         | -h    → Show this menu
 ┣ --verbose      | -v    → Enable too much logging
 ┣ --force        | -f    → Force an operation ignoring checks (e.g. update -f)
@@ -43,7 +44,7 @@ int                        main(int argc, char** argv, char** envp) {
     }
 
     std::vector<std::string> command;
-    bool                     notify = false, verbose = false, force = false, noShallow = false;
+    bool                     notify = false, notifyFail = false, verbose = false, force = false, noShallow = false;
 
     for (int i = 1; i < argc; ++i) {
         if (ARGS[i].starts_with("-")) {
@@ -52,6 +53,8 @@ int                        main(int argc, char** argv, char** envp) {
                 return 0;
             } else if (ARGS[i] == "--notify" || ARGS[i] == "-n") {
                 notify = true;
+            } else if (ARGS[i] == "--notify-fail" || ARGS[i] == "-nn") {
+                notifyFail = notify = true;
             } else if (ARGS[i] == "--verbose" || ARGS[i] == "-v") {
                 verbose = true;
             } else if (ARGS[i] == "--no-shallow" || ARGS[i] == "-s") {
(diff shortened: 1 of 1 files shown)


littleblack111:
i would like this too, but i think we should still keep the option for user too. so maybe like this?
```
diff --git a/hyprpm/src/main.cpp b/hyprpm/src/main.cpp
index 37972443..c5cf1652 100644
--- a/hyprpm/src/main.cpp
+++ b/hyprpm/src/main.cpp
@@ -22,6 +22,7 @@ const std::string HELP = R"#(┏ hyprpm, a Hyprland Plugin Manager
 ┣ Flags:
 ┃
 ┣ --notify       | -n    → Send a hyprland notification for important events (e.g. load fail)
+┣ --notify-fail  | -nn   → Send a hyprland notification for fail events
 ┣ --help         | -h    → Show this menu
 ┣ --verbose      | -v    → Enable too much logging
 ┣ --force        | -f    → Force an operation ignoring checks (e.g. update -f)
@@ -49,6 +50,9 @@ int               main(int argc, char** argv, char** envp) {
                 return 0;
             } else if (ARGS[i] == "--notify" || ARGS[i] == "-n") {
                 notify = true;
+            } else if (ARGS[i] == "--notify-fail" || ARGS[i] == "-nn") {
+                notify = true;
+                notifyFail = true;
             } else if (ARGS[i] == "--verbose" || ARGS[i] == "-v") {
                 verbose = true;
             } else if (ARGS[i] == "--force" || ARGS[i] == "-f") {
@@ -144,7 +148,7 @@ int               main(int argc, char** argv, char** envp) {
                     break;
                 default: break;
             }
-        } else if (notify) {
+        } else if (notify && !notifyFail) {
             g_pPluginManager->notify(ICON_OK, 0, 4000, "[hyprpm] Loaded plugins");
         }
     } else if (command[0] == "list") {
@@ -155,4 +159,4 @@ int               main(int argc, char** argv, char** envp) {
     }
 
     return 0;
-}
\ No newline at end of file
+}
```

KAGEYAM4:
My problem was that on Hyprland first boot i was getting notification and it was ruining the startup animation produced by swww.

I see how how still keeping this can be good. But i am confused wether in UI case do we need to produce notification for pass state. Like UI should be clutter free, and generating notification on fail would be optimal.

@littleblack111 @vaxerski What you wanna do, keep both or keep one.

-------------

Also if we are going to keep both fail and pass, i thought some changes -->
1. `-nn` should be for both pass & fail and `-n` should be for fail. Maybe similar to how we have -vv for more verbose. ( But it's going to be breaking change ) OR maybe we can think of another flag.
2. `-nn` and `-n` help needs to be changed to specify that `-nn` will only produce notification for failed events. `-n` will for both fail&pass.
3. Maybe we can improve the code but at the cost of complexity -->

<details>
<summary> Code </summary>

### At start
```
          } else if (ARGS[i] == "--notify-fail" || ARGS[i] == "-nn") {
                notifyFail = true;
```
      
### we could instead initialize  both
```
 notifyFail = true; notify = true;
```

Then every `(notify || notifyFail)` can be replaced with `(notify)` reducing one check.

And at the end we can do 
 ```
         } else if (notify && !notifyFail) {
            g_pPluginManager->notify(ICON_OK, 0, 4000, "[hyprpm] Loaded plugins");
```

The reson is that, user will only provice one or the other. `notifyfail` and `notify` will produce notification in failed state. Only at the end when we have to generate notification for `pass state` we can check if `notifyFail` is false or not.

</details>

littleblack111:
or we can maybe just use the `notify` var but instead of bool, we can use int or char to represent those

KAGEYAM4:
> or we can maybe just use the `notify` var but instead of bool, we can use int or char to represent those. But that would make different then other vars since they're all bool

Yes, now we are addong two complexities. The one you suggested (int) and the one mine where we still being implicit about checking only one condition. And if we checking both condition `( var1 == <> || var2 == <>)`, we are back to where we started.

vaxerski:
ye but then what will `nnn` be for? no nut november?

jk.

-n and -nn sounds reasonable to me.

KAGEYAM4:
@vaxerski  please review

littleblack111:
no nut November it is. What do u think @vaxerski

vaxerski (approved):

vaxerski:
you dont need to ping me assholes I have all notifications on

Issue #8206, opened 2024-10-23
Swallows don't work in grouped (tabbed) mode


yatinlala:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 4520b30d498daca8079365bdb909a8dea38e8d55  (version: bump to 0.44.1).
Date: Wed Oct 9 12:54:39 2024
Tag: v0.44.1, commits: 5309
built against aquamarine 0.4.2


flags: (if any)


System Information:
System name: Linux
Node name: phoenix
Release: 6.11.5-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Tue, 22 Oct 2024 18:31:38 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation WhiskeyLake-U GT2 [UHD Graphics 620] [8086:3ea0] (rev 02) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/yatin/.config/hypr/hyprland.conf: Read Succeeded
#: Keybinds {{{

$mod = SUPER
$mod2 = ALT
$term = kitty -1
$term2 = foot

bind=$mod SHIFT,n,exec,$HOME/foo.sh
bind=$mod,Return,exec,$term
bind=$mod SHIFT,Return,exec,$term -e muxterm mix
bind=$mod $mod2,Return,exec,$term2

bind=$mod,r,exec, hyprctl reload && killall waybar && waybar &
# TODO fix
bind=$mod $mod2,r,exec,hyprfreeze -a

bind=$mod,space,exec, rofi -show drun
bind=$mod,semicolon,exec, tessen -a both
bind=$mod SHIFT,s,exec, ~/.config/rofi/powermenu/powermenu.sh
[...]

PR #8223, opened 2024-10-24
fix swallowing for groups

Aqa-Ib:
Fixes https://github.com/hyprwm/Hyprland/issues/8206

Ready.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -193,7 +193,11 @@ bool IHyprLayout::onWindowCreatedAutoGroup(PHLWINDOW pWindow) {
     if (pWindow->m_bIsFloating && !OPENINGON->m_bIsFloating)
         denied = true;
 
-    if (*PAUTOGROUP                                      // check if auto_group is enabled.
+    bool swallowing = false;
+    if (pWindow->m_pSwallowed || pWindow->m_bGroupSwallowed)
+        swallowing = true;
+
+    if ((*PAUTOGROUP || swallowing)                      // continue if auto_group is enabled or if dealing with window swallowing.
         && OPENINGON                                     // this shouldn't be 0, but honestly, better safe than sorry.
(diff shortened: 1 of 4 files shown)


Aqa-Ib:
I'm gonna refactor this PR.

Aqa-Ib:
In the end, this is the best approach that I could achieve.
Ready to merge if you like it.

vaxerski:
I'll read this tomorrow. Generally, windows should be separated from containers internally in the codebase, heh, but it's way too late for that.

Aqa-Ib:
> Generally, windows should be separated from containers internally in the codebase, heh, but it's way too late for that.

Better late than never, we can move the swallowing logic to onWindowCreated and onWindowDestroyed if that's what you mean ?

vaxerski:
yeah at this point it'd be really hard + break a lot of plugins, plus prolly introduce tons of bugs. Shit's... good enough I guess. I don't think we'll be doing anything fancier than groups on windows.

src/layout/IHyprLayout.cpp
```diff
@@ -193,7 +193,11 @@ bool IHyprLayout::onWindowCreatedAutoGroup(PHLWINDOW pWindow) {
     if (pWindow->m_bIsFloating && !OPENINGON->m_bIsFloating)
         denied = true;
 
-    if (*PAUTOGROUP                                      // check if auto_group is enabled.
+    bool swallowing = false;
+    if (pWindow->m_pSwallowed || pWindow->m_bGroupSwallowed)
+        swallowing = true;
```
vaxerski, line 198: `const bool SWALLOWING = pWindow->m_pSwallowed || pWindow->m_bGroupSwallowed;` bud..?

Aqa-Ib: Thanks. I have also changed the other variables too for consistency.

vaxerski (approved):

PR #8229, opened 2024-10-24
security-context: avoid UB in C macro

gulafaran:
to safely use wl_container_of with a class the class has to be no virtual functions, no inheritance, and uniform access control (e.g all public)

work around this by putting this into a destroywrapper struct.

just like in https://github.com/hyprwm/hyprwayland-scanner/pull/8

however this warrants for some testing so it doesnt break something. not sure what uses this protocol besides static analyzers showing the issue.


vaxerski:
launch a flatpak app

gulafaran:
> launch a flatpak app

2024-10-25-083405_hyprshot

vaxerski:
cool

vaxerski (approved):

PR #8230, opened 2024-10-24
hyprland-session.service example - add support for xdg autostart

izmyname:
Another small feat, in addition to previous changes to the service - add support for systemd xdg-autostart https://systemd.io/DESKTOP_ENVIRONMENTS/
https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html

No

Yes


fufexan (approved):
I think units can be comma-separated, but this also works.

PR #8266, opened 2024-10-27
Core: move internal structures from monitor IDs to pointers

vaxerski:
Needs some testing, should improve performance a tiny bit now that we can do this.


vaxerski:
ok

PR #8271, opened 2024-10-28
few more marginal optimisations from profiling

gulafaran:
found a few more marginal gains that could be had on high frequently called functions.

reserve the vector sizes to reduce the amount of reallocations that could happend, most of the time it doesnt.
use const ref in the function parameters and reduce function calls and temporar copies in getWindowDecorationExtents
and check if there is even extents to add before adding

### DIFF
--- a/src/render/decorations/DecorationPositioner.cpp
+++ b/src/render/decorations/DecorationPositioner.cpp
@@ -294,13 +294,11 @@ SBoxExtents CDecorationPositioner::getWindowDecorationReserved(PHLWINDOW pWindow
 }
 
 SBoxExtents CDecorationPositioner::getWindowDecorationExtents(PHLWINDOW pWindow, bool inputOnly) {
-    CBox accum = pWindow->getWindowMainSurfaceBox();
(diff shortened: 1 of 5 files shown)


src/render/decorations/DecorationPositioner.cpp
```diff
@@ -294,45 +294,53 @@ SBoxExtents CDecorationPositioner::getWindowDecorationReserved(PHLWINDOW pWindow
...
-        if (decoBox.y + decoBox.h > accum.y + accum.h)
+            needUpdate                 = true;
+        }
+        if (decoBox.y + decoBox.h > accum.y + accum.h) {
             extentsToAdd.bottomRight.y = (decoBox.y + decoBox.h) - (accum.y + accum.h);
+            needUpdate                 = true;
+        }
 
-        accum.addExtents(extentsToAdd);
+        if (needUpdate) {
```
vaxerski, line 338: bro put {} around a short if LMAO

vaxerski (approved):
thanks!

PR #8275, opened 2024-10-28
move setprop into dispatchers

MightyPlaza:
Moves `setprop` to dispatchers, mostly to allow calling it in keybinds without having to use hyprctl.
Didn't find any issues.

wiki: https://github.com/hyprwm/hyprland-wiki/pull/832

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1261,6 +1261,11 @@ std::string dispatchSeterror(eHyprCtlOutputFormat format, std::string request) {
     return "ok";
 }
 
+std::string dispatchSetProp(eHyprCtlOutputFormat format, std::string request) {
+    auto result = g_pKeybindManager->m_mDispatchers["setprop"](request.substr(request.find_first_of(' ') + 1, -1));
+    return "DEPRECATED: use hyprctl dispatch setprop instead" + (result.success ? "" : "\n" + result.error);
+}
+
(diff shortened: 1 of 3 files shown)


vaxerski (dismissed):

vaxerski:
actually, can we make `hyprctl setprop` just redirect to the dispatcher and print something like "deprecated"

src/debug/HyprCtl.cpp
```diff
@@ -1261,6 +1261,11 @@ std::string dispatchSeterror(eHyprCtlOutputFormat format, std::string request) {
     return "ok";
 }
 
+std::string dispatchSetProp(eHyprCtlOutputFormat format, std::string request) {
+    auto result = g_pKeybindManager->m_mDispatchers["setprop"](request.substr(request.find_first_of(' ') + 1, -1));
+    return "DEPRECATED: use hyprctl dispatch setprop instead" + (result.success ? "" : "\n" + result.error);
```
vaxerski, line 1266: worth mentioning it's a dispatcher if someone is running `,exec,hyprctl setprop...`

"use the setprop dispatcher or hyprctl dispatch setprop instead"

MightyPlaza: sure, but if they're running that they won't see the output anyways

vaxerski: right, disregard then

vaxerski (approved):

PR #8289, opened 2024-10-29
snap: add option `border_overlap` and other improvements

myQwil:

I really liked the way borders used to overlap when snapping and how only the window's main surface would snap to the monitor, so I would like to bring that behavior back, but in the form of a config option.

Other improvements include:

- reduced the number of snap functions from 4 down to 2, and only one ever gets declared at any given time.

- border size should not be added to gap size. It seemed like the right way to do it at the time, but it makes snapping feel way stronger than it actually should.

- all `const` variables have been given the all-caps naming convention.

- to avoid excessive casting, border size is declared as a double.

- to avoid excessive `x + w`, `y + h` calculations. I'm using a struct called `Range` and working only with start and end values until the very end of the function.

- check for both monitor snapping as well as reserved monitor space
snapping in a relatively efficient way.

Not that I can think of.

It's ready to merge.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -392,91 +392,91 @@ void IHyprLayout::onEndDragWindow() {
     g_pInputManager->m_bWasDraggingWindow = false;
 }
 
-static inline bool canSnap(const double sideA, const double sideB, const double gap) {
-    return std::abs(sideA - sideB) < gap;
+static inline bool canSnap(const double SIDEA, const double SIDEB, const double GAP) {
+    return std::abs(SIDEA - SIDEB) < GAP;
 }
 
-static void snapMoveLeft(double& pos, double& len, const double p) {
-    pos = p;
+static void snapMove(double& start, double& end, const double P) {
+    end   = P + (end - start);
+    start = P;
 }
 
-static void snapMoveRight(double& pos, double& len, const double p) {
-    pos = p - len;
-}
-
-static void snapResizeLeft(double& pos, double& len, const double p) {
-    len += pos - p;
-    pos = p;
-}
-
-static void snapResizeRight(double& pos, double& len, const double p) {
-    len = p - pos;
+static void snapResize(double& start, double& end, const double P) {
+    start = P;
 }
 
 typedef std::function<void(double&, double&, const double)> SnapFn;
 
-static void performSnap(Vector2D& sourcePos, Vector2D& sourceSize, PHLWINDOW DRAGGINGWINDOW, const eMouseBindMode mode, const int corner, const Vector2D& beginSize) {
-    static auto  SNAPWINDOWGAP  = CConfigValue<Hyprlang::INT>("general:snap:window_gap");
-    static auto  SNAPMONITORGAP = CConfigValue<Hyprlang::INT>("general:snap:monitor_gap");
+static void performSnap(Vector2D& sourcePos, Vector2D& sourceSize, PHLWINDOW DRAGGINGWINDOW, const eMouseBindMode MODE, const int CORNER, const Vector2D& BEGINSIZE) {
+    static auto  SNAPWINDOWGAP     = CConfigValue<Hyprlang::INT>("general:snap:window_gap");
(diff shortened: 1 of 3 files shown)


src/layout/IHyprLayout.cpp
```diff
@@ -392,141 +392,141 @@ void IHyprLayout::onEndDragWindow() {
...
-    int          snaps     = 0;
+    const SnapFn SNAP  = (MODE == MBIND_MOVE) ? snapMove : snapResize;
+    int          snaps = 0;
 
+    const bool   OVERLAP            = *SNAPBORDEROVERLAP;
     const int    DRAGGINGBORDERSIZE = DRAGGINGWINDOW->getRealBorderSize();
-    CBox         sourceBox          = CBox{sourcePos, sourceSize}.expand(DRAGGINGBORDERSIZE);
+
+    struct Range {
+        double start;
```
vaxerski, line 422: ALARM UNINITIALIZED VALUE DETECTED

src/layout/IHyprLayout.cpp
```diff
@@ -392,141 +392,141 @@ void IHyprLayout::onEndDragWindow() {
...
-    const SnapFn snapUp    = snapLeft;
-    int          snaps     = 0;
+    const SnapFn SNAP  = (MODE == MBIND_MOVE) ? snapMove : snapResize;
+    int          snaps = 0;
 
+    const bool   OVERLAP            = *SNAPBORDEROVERLAP;
     const int    DRAGGINGBORDERSIZE = DRAGGINGWINDOW->getRealBorderSize();
-    CBox         sourceBox          = CBox{sourcePos, sourceSize}.expand(DRAGGINGBORDERSIZE);
+
+    struct Range {
```
vaxerski, line 421: SRange

src/layout/IHyprLayout.cpp
```diff
@@ -392,141 +392,141 @@ void IHyprLayout::onEndDragWindow() {
...
+        const double GAPSIZE = *SNAPWINDOWGAP;
+        const auto   WSID    = DRAGGINGWINDOW->workspaceID();
 
         for (auto& other : g_pCompositor->m_vWindows) {
             if (other == DRAGGINGWINDOW || other->workspaceID() != WSID || !other->m_bIsMapped || other->m_bFadingOut || other->isX11OverrideRedirect())
                 continue;
 
-            const int      BORDERSIZE = other->getRealBorderSize();
-            const double   GAPSIZE    = *SNAPWINDOWGAP + BORDERSIZE;
+            const double BORDERSIZE = OVERLAP ? std::max(DRAGGINGBORDERSIZE, other->getRealBorderSize()) : (other->getRealBorderSize() + DRAGGINGBORDERSIZE);
```
vaxerski, line 436: I don't get why you do std::max here. Both windows can have different border sizes, in which case you have to add both of them together, not max them?

src/layout/IHyprLayout.cpp
myQwil: that's a field of a local struct.

src/layout/IHyprLayout.cpp
myQwil: This feature is all about determining how much space there should be between the main surfaces of two windows.

If `border_overlap` is off (default), then the space between them is both borders added together.

If `border_overlap` is on, then the space between them is the larger of the two borders, allowing some border overlap, but not enough to actually start covering over the main surface of either window.

src/layout/IHyprLayout.cpp
myQwil: I can change the name, but what does the 'S' stand for in this context?

myQwil:
Here is a video demonstrating what these changes look like:

https://github.com/user-attachments/assets/e248a568-1dbc-48e2-9545-58f3e3f9230b

src/layout/IHyprLayout.cpp
vaxerski: STRUCT

src/layout/IHyprLayout.cpp
vaxerski: makes sense.

vaxerski (approved):
thanks! wiki mr needed

Issue #8152, opened 2024-10-17
Bi-directional floating resize


Vladimir-csp:
### Description

Dispatcher to resize floating window along an axis while maintaining center position. Something like `resizeactiveaxis, x 10`, `resizeactiveaxis, y -5`

I tried to fabricate this behavior using multiple actions, but only a single `binde` seems to fire on repeat.
```
binde = SUPER SHIFT, left, moveactive, 5 0                                                                                                                                        
binde = SUPER SHIFT, left, resizeactive, -10 0                                                                                                                                    
binde = SUPER SHIFT, right, moveactive, -5 0                                                                                                                                      
binde = SUPER SHIFT, right, resizeactive, 10 0                                                                                                                                    
binde = SUPER SHIFT, up, moveactive, 0 5                                                                                                                                          
binde = SUPER SHIFT, up, resizeactive, 0 -10                                                                                                                                      
binde = SUPER SHIFT, down, moveactive, 0 -5                                                                                                                                       
binde = SUPER SHIFT, down, resizeactive, 0 10
```

Vladimir-csp:
Retested with v0.44.1, still repeats only one `binde`.

Issue #4248, opened 2023-12-25
resizewindow on both sides


vincentaxhe:
### Description

when a window in the middle ,resizewindow cannot specify sides, in bspwm when I want to expand a window, use `bspc node -z {left -30 0,bottom 0 30,top 0 -30,right 30 0}` when shrink a window use `bspc node -z {right -30 0,top 0 30,bottom 0 -30,left 30 0}`,hyprland simplify it ,but I think bspwm have a more flawless strategy because firstly user determined is expand or shrink a window, but resizewindow in hyprland cannot always got what users want, there should be a sub definition to deal with resizewindow.

dranull:
You can move and resize the window at the same time, which gives you the effect of expanding it to the opposite direction. Here's an example:

```
$isFloat = hyprctl activewindow | grep -q 'floating: 1'
$mpx = 20

binde = $mainMod SHIFT, H, exec, $isFloat && hyprctl --batch "dispatch resizeactive $mpx 0; dispatch moveactive -$mpx 0" || hyprctl dispatch resizeactive -$mpx 0
binde = $mainMod SHIFT, L, resizeactive, $mpx 0
binde = $mainMod SHIFT, J, resizeactive, 0 $mpx
binde = $mainMod SHIFT, K, exec, $isFloat && hyprctl --batch "dispatch resizeactive 0 $mpx; dispatch moveactive 0 -$mpx" || hyprctl dispatch resizeactive 0 -$mpx

binde = $mainMod CTRL SHIFT, L, exec, $isFloat && hyprctl --batch "dispatch resizeactive -$mpx 0; dispatch moveactive $mpx 0" || hyprctl dispatch resizeactive $mpx 0
binde = $mainMod CTRL SHIFT, H, resizeactive, -$mpx 0
binde = $mainMod CTRL SHIFT, K, resizeactive, 0 -$mpx
binde = $mainMod CTRL SHIFT, J, exec, $isFloat && hyprctl --batch "dispatch resizeactive 0 -$mpx; dispatch moveactive 0 $mpx" || hyprctl dispatch resizeactive 0 $mpx

```

If you want to resize both sides at the same time, then you have to move the window twice as far. Or maybe I didn't understand your issue...

vincentaxhe:
I mean when the middle window is active ,resizeactive can only move the left side left or right ,not the right side,It's not about floating,though movewindow on a floating window pushes it to the the edge seems not exactly what I think about "movewindow".
[...]

PR #8290, opened 2024-10-29
allow repeating multiple binds

MightyPlaza:
allows repeating multiple binds
closes: https://github.com/hyprwm/Hyprland/issues/8152
and maybe other issues
seems to work fine

closes: https://github.com/hyprwm/Hyprland/issues/8244
removes need for script in: https://github.com/hyprwm/Hyprland/issues/4248

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -147,19 +147,18 @@ CKeybindManager::CKeybindManager() {
     m_pRepeatKeyTimer = makeShared<CEventLoopTimer>(
         std::nullopt,
         [this](SP<CEventLoopTimer> self, void* data) {
+            if (m_vActiveKeybinds.size() == 0 || g_pSeatManager->keyboard.expired())
+                return;
 
-        if (m_vActiveKeybinds.size() == 0 || g_pSeatManager->keyboard.expired())
-            return;
+            for (SKeybind* k : m_vActiveKeybinds) {
+                const auto DISPATCHER = g_pKeybindManager->m_mDispatchers.find(k->handler);
 
-        for (SKeybind* k : m_vActiveKeybinds) {
-            const auto DISPATCHER = g_pKeybindManager->m_mDispatchers.find(k->handler);
+                Debug::log(LOG, "Keybind repeat triggered, calling dispatcher.");
+                DISPATCHER->second(k->arg);
+            }
 
-            Debug::log(LOG, "Keybind repeat triggered, calling dispatcher.");
-            DISPATCHER->second(k->arg);
-        }
-
-        const auto PACTIVEKEEB = g_pSeatManager->keyboard.lock();
-        self->updateTimeout(std::chrono::milliseconds(1000 / PACTIVEKEEB->repeatRate));
+            const auto PACTIVEKEEB = g_pSeatManager->keyboard.lock();
+            self->updateTimeout(std::chrono::milliseconds(1000 / PACTIVEKEEB->repeatRate));
         },
         nullptr);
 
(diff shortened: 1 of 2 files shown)


MightyPlaza:
I might need to make sure event loop only gets added once

src/managers/KeybindManager.hpp
```diff
@@ -119,7 +119,7 @@ class CKeybindManager {
 
     inline static std::string       m_szCurrentSelectedSubmap = "";
 
-    SKeybind*                       m_pActiveKeybind = nullptr;
+    std::vector<SKeybind*>          m_pActiveKeybinds;
```
vaxerski, line 122: not a `p` anymore, now a `v`

src/managers/KeybindManager.cpp
```diff
@@ -732,12 +734,14 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
...
-            m_pActiveKeybindEventSource = wl_event_loop_add_timer(g_pCompositor->m_sWLEventLoop, repeatKeyHandler, &m_pActiveKeybind);
+            m_pActiveKeybinds.push_back(&k);
+            if (!m_pActiveKeybindEventSource) {
+                m_pActiveKeybindEventSource = wl_event_loop_add_timer(g_pCompositor->m_sWLEventLoop, repeatKeyHandler, &m_pActiveKeybinds);
 
-            const auto PACTIVEKEEB = g_pSeatManager->keyboard.lock();
+                const auto PACTIVEKEEB = g_pSeatManager->keyboard.lock();
 
-            wl_event_source_timer_update(m_pActiveKeybindEventSource, PACTIVEKEEB->repeatDelay);
+                wl_event_source_timer_update(m_pActiveKeybindEventSource, PACTIVEKEEB->repeatDelay);
```
vaxerski, line 743: can we please get rid of the C timer and use our C++ one? managers/eventLoop

vaxerski (commented):
note to self: refactor to not use raw-ass pointers in keybinds

MightyPlaza:
seems to be working fine

vaxerski:
but doesnt pass style xd

src/managers/KeybindManager.cpp
```diff
@@ -144,11 +144,30 @@ CKeybindManager::CKeybindManager() {
         },
         nullptr);
 
+    m_pRepeatKeyTimer = makeShared<CEventLoopTimer>(
+        std::nullopt,
+        [this](SP<CEventLoopTimer> self, void* data) {
+            if (m_vActiveKeybinds.size() == 0 || g_pSeatManager->keyboard.expired())
+                return;
+
+            for (SKeybind* k : m_vActiveKeybinds) {
```
vaxerski, line 153: what if two keybinds are separate?

consider one keybind for A, and one for B.

When pressing A, then B, both will start spamming at the same time. When they are spamming, pressing C will make it _instantly_ spam too.

Is this what we want?

MightyPlaza: despite the name m_vActiveKeybinds is only used/added for repeat keybinds

additionally in the case where there 2 binds for different keys
I believe that would/should require using `bindte`
but even with that can't make it work (it prolly should, but maybe for another time)
(just the set of keybinds for 1 key gets ran)

likely releated to the vector being cleared in multiple places, for example `onKeyEvent()`
which should for now handle edge cases where multiple keybinds for different keys would get matched
since it seems to only run the last one

vaxerski: mhm, okay

vaxerski (approved):

PR #8300, opened 2024-10-30
more profiling less calls and local copies

gulafaran:
more profiling of erraticly moving mouse and generally looking at the most cpu demanding functions appearing doing so.

const ref the things we can, move various objects and variables down in scope to where its actually being used, reduce amount of calls to functions and try reuse already looped and fetched object. mousemoveunified itself could use a refactor but its scary big and i will eventually break something, maybe il attempt it in the future.

add one more for loop in renderWorkspaceWindows and store the windows in a vector with reference wrappers that should be  rendered, this adds a loop but reduces the amount of repeated calls to shouldRenderWindow and also makes the rest of the loops go over way smaller vector when many windows exist.

this could use some testing so i didnt actually break something non obvious, works here locally so far. and made marginal gains in the profiling.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -158,9 +158,7 @@ static void renderSurface(SP<CWLSurfaceResource> surface, int x, int y, void* da
     if (!surface->current.texture)
         return;
 
-    const auto& TEXTURE                     = surface->current.texture;
-    const auto  RDATA                       = (SRenderData*)data;
-    const auto  INTERACTIVERESIZEINPROGRESS = RDATA->pWindow && g_pInputManager->currentlyDraggedWindow && g_pInputManager->dragMode == MBIND_RESIZE;
+    const auto& TEXTURE = surface->current.texture;
 
     // this is bad, probably has been logged elsewhere. Means the texture failed
     // uploading to the GPU.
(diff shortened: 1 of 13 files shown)


src/render/Renderer.cpp
```diff
@@ -484,17 +484,25 @@ void CHyprRenderer::renderWorkspaceWindows(PHLMONITOR pMonitor, PHLWORKSPACE pWo
 
     EMIT_HOOK_EVENT("render", RENDER_PRE_WINDOWS);
 
-    // Non-floating main
+    std::vector<std::reference_wrapper<PHLWINDOW const>> windows;
```
vaxerski, line 487: instead of a ref wrapper, why not just use a WP? Would it be that much slower?

Reasoning is, since we never render a window twice, you could set the weak_ptr to a nullptr after every render and that would provide a very fast bail-out in further loops (if (!w) continue)

vaxerski:
also wondering, how are you profiling? callgrind has been kinda dog on my end, and tracy broke a while back.

gulafaran:
> also wondering, how are you profiling? callgrind has been kinda dog on my end, and tracy broke a while back.

pretty much like this, 

https://github.com/user-attachments/assets/854ea040-26b6-40cc-9497-b2a67bf60e57

keep in mind it has tons of various parameters and options to fiddle with, but basicly get pid , `perf top -K -p $PID` and you can step into the various functions and calls that is averaging the most and trickle down why or what its doing.

vaxerski:
dope, thanks

src/render/Renderer.cpp
gulafaran: oh true thats also one way to tackle it. nah WP is fine, i just tried different things and got down to the least expensive way, but didnt consider that way. heh.

vaxerski: yes. In the first loop if (!w) is always false tho can be dropped

vaxerski (approved):
tack!

PR #8323, opened 2024-11-01
xwayland: minor fixups for stability

Trimutex:
- Fix lock file permissions
- Ensure sockets are closed safely using inline methods
- Attempt to recreate XWayland instance on crash

Any socket closes swapped from close() to closeSocketSafely() were differences found in wlroots's method.

Ready

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -13,6 +13,7 @@
 #include "../defines.hpp"
 #include "../Compositor.hpp"
 #include "../protocols/core/Seat.hpp"
+#include "../managers/eventLoop/EventLoopManager.hpp"
 #include "../managers/SeatManager.hpp"
 #include "../protocols/XWaylandShell.hpp"
 #include "../protocols/core/Compositor.hpp"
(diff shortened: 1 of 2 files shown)


src/xwayland/XWM.cpp
```diff
@@ -691,6 +691,8 @@ int CXWM::onEvent(int fd, uint32_t mask) {
         Debug::log(CRIT, "XWayland has yeeten the xwm off?!");
         g_pXWayland->pWM.reset();
         g_pXWayland->pServer.reset();
+        // Attempt to create fresh instance
+        g_pXWayland = std::make_unique<CXWayland>(true);
```
vaxerski, line 695: wont this already happen elsewhere?

Trimutex: I don't think so as every time my XWayland crashes (a separate thing I haven't reasoned out yet) it requires disabling and re-enabling before it works again. Programs instead just pop up as if XWayland were disabled.

Trimutex: If you have an area where you think this should be put instead I'm down take a look into it

vaxerski: ah.. right. I see. This should probably be done in a `doLater`, though. Making it instantly here would be an awkward codepath.

Trimutex: Good point, newest commit should resolve that

vaxerski (approved):
thanks!

Issue #6889, opened 2024-07-15
Add option to map tablets across monitors using input:tablet:output=ALL in hyprland.conf


Da-Boom:
### Description

In relation to The changes made in #5902 and the bug at #6023

Its well known that Hyprland gives the option `input:tablet:output=`  which allows me to specify a monitor to bind a tablet to. This is great for regular drawing tablets, as it simplifies things. By default, when not specified it maps to whichever monitor the cursor is currently on. 

But there's a situation where this setup not great. (not counting the obvious oddly shaped tablets and monitors like a 16:10 tablet on a 16:9 display, which has obvious implications, but can also be manipulated with other input settings for tablets.)

when i have an abnormal pointing devices that doesn't follow these conventions. like in the case of `wlx-overlay-s` made by galister (https://github.com/galister/wlx-overlay-s) or say if someone were to try using a wii remote as a pointing device across monitors - (`cwiid` is a thing)

`wlx-overlay-s` designed to work with `wlroots` based compositors. It allows using desktop portals to preview the screens onto surfaces on a HMD (head mounted display) it simulates a pointer device using `uinput` and maps using the regular tablet surface - which is defined by `input:tablet:output=`  in hyprland. 

This leads to a problem. Given that it expects to get the entire desktop canvass for BOTH monitors, as that's what most other compositors do. it ends up inadvertently mapping one monitor across both the virtual screens in vr. 

 - This leads to me, with my cursor on the right screen, pointing my VIVE wand at the left, and finding the cursor moving about on the left half of the right screen.  If i point at the right screen, i get control of the right half of the right screen.
 
 i run 2 1440p displays (resulting resolution 5120x1440). i imagine it would behave even more wacky if they were different resolutions, or if i had a third display.
 
what i want in this feature request is the simple ability to supply `ALL` to the `input:tablet:output=` setting, allowing me to map the virtual device across all existing separate displays at the same time.
 
I am willing to test this as i have my `monado` setup with galisters' patches functioning nicely, a HMD and i already have `wlx-overlay-s` installed.

CobaltSpace:
I was doing some testing. Using my drawing tablet, I tested the following:
```hyprlang
input:tablet {
  output = $top_left_output
  region_size = $layout_width, $layout_height
}
```
This worked for my tablet, but when I launched wlx-overlay-s, it was still broken. I then tried
```hyprlang
device {
  name = wlxoverlay-s-keyboard-mouse-hybrid-thing-1
[...]

Issue #6023, opened 2024-05-11
Tablet no longer maps to area correctly


Stoppedpuma:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 494b9415a1157279a1e1782ba635fc2ef6a18155  (layersurface: avoid restack on identical layers).
Tag: v0.40.0-62-g494b9415, commits: 4668

flags: (if any)

System Information:
System name: Linux
Release: 6.8.9-273-tkg-eevdf-llvm
Version: #1 SMP PREEMPT_DYNAMIC TKG

GPU information: 
0c:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M/6850M XT] [1002:73df] (rev c5) (prog-if 00 [VGA controller])
```

</details>

### Bug or Regression?

Regression

### Description

Caused by #5902 I'm assuming.

Tablet area and display area in OpenTabletDriver no longer map correctly, neither the tablet area or display area seem to work correctly.

Possible reason this might not have been reported in the PR is that I'm using two displays?

### Crash reports, logs, images, videos

Video showing cursor stopping near the middle of the screen:

https://github.com/hyprwm/Hyprland/assets/58333920/450745f4-41ec-4898-a879-1fdd1cdccb0e

Screenshot of the display area in OpenTabletDriver:

![1715453971](https://github.com/hyprwm/Hyprland/assets/58333920/dcd9c528-217e-4f1a-95e8-be57f9d9d1fd)

Agent00Ming:
Yeah, I can replicate. Hyprland ignores what opentabletdriver is mapped to and automatically maps to the display your cursor is on. You can leave the opentabletdriver mapping to default (whole area) and Hyprland will do it's thing. You can set which monitor to map your tablet to in `input:tablet:output=`

Stoppedpuma:
> Hyprland ignores what opentabletdriver is mapped to

Is this the way planned forward or just an oversight? I don't think that ignoring the drivers configuration is the best way of handling multiple displays?

Agent00Ming:
"Be the change you want to see". I'll look at the code and see if there's a better way to manage this (add a config maybe?). Stock behaviour has definitely changed from before #5902 and vaxry doesn't even have a tablet to test with :stuck_out_tongue_closed_eyes:

vaxerski:
[...]

PR #7586, opened 2024-08-30
[Tablet] Allow mapping to entire monitor layout

Agent00Ming:
#### Describe your PR, what does it fix/add?
Fixes #6023. 
Moves monitor layout entire area search to a function and enables tablets to map to the entire monitor layout. This allows OpenTabletDriver's monitor mapping settings to work on artist mode.

Optionally enabled via `input:tablet:output` = `entire`

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Tested working with OTD on artist mode. Does not work with OTD absolute mode (virtual device is listed alongside normal mice in `hyprctl devices`). Related: #6889 

#### Is it ready for merging, or does it need work?
Ready.

PR #8352, opened 2024-11-04
pointer: map devices across all outputs by default

JManch:
This fixes a problem with [wlx-overlay-s](https://github.com/galister/wlx-overlay-s?tab=readme-ov-file#mouse-is-not-where-it-should-be) where multi-monitor overlay pointer movement is broken. Somewhat related issues: #6023 #6889

I'm not sure if it's correct to map across the entire space by default.

Ready

### DIFF
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -668,9 +668,7 @@ void CPointerManager::move(const Vector2D& deltaLogical) {
 }
 
 void CPointerManager::warpAbsolute(Vector2D abs, SP<IHID> dev) {
-
-    PHLMONITOR currentMonitor = g_pCompositor->m_pLastMonitor.lock();
-    if (!currentMonitor || !dev)
+    if (!dev)
         return;
 
     if (!std::isnan(abs.x))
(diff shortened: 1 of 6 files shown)


Agent00Ming:
Mapping to all outputs by default only for pointer devices makes OpenTabletDriver's absolute mode (pointer device) and artist mode (tablet device) map differently despite otherwise identical internal mapping configurations.

vaxerski:
isnt that just #7586

Agent00Ming:
Not exactly, this PR changes the default mapping for pointer devices only while #7586 only gives the option for tablet devices to map to all outputs.

JManch:
#7586 doesn't fix the problem since `boundOutput` is empty. Would it make sense to change the default mapped area to all outputs instead of just the current monitor?
https://github.com/hyprwm/Hyprland/blob/88e9e0394541a853600bc2c910005c05fa156269/src/managers/PointerManager.cpp#L682

JManch:
I've changed it so that tablet, touch and points devices map across all outputs unless they are explicitly bound to an output. This changes how tablet regions are applied. When bound to an output, region_position is relative to the bound output, otherwise, it's an absolute position.

This fixes wlx-overlay-s whilst hopefully maintaining consistent behaviour between absolute mode and artist mode for tablets. I've done some brief testing with my tablet and it seems ok but I don't have a way to test touch devices.

src/managers/PointerManager.cpp
```diff
@@ -668,68 +668,71 @@ void CPointerManager::move(const Vector2D& deltaLogical) {
...
+                    // apply the bound box relative to the bound output
+                    mappedArea.translate(TAB->boundBox.pos());
+                    if (!TAB->boundBox.empty()) {
+                        mappedArea.w = TAB->boundBox.w;
+                        mappedArea.h = TAB->boundBox.h;
+                    }
+                }
+            } else {
+                mappedArea.x = TAB->boundBox.x;
+                mappedArea.y = TAB->boundBox.y;
```
Agent00Ming, line 712: This causes the origin to move to (0, 0) and make monitor regions in negative coordinates deadzones.

Agent00Ming (changes requested):

src/managers/PointerManager.cpp
JManch: oh yeah forgot about that. I'd have to add another option like `bool custom_region_position` or just always treat the region as relative.

Do you think it's worth it? I'd like being able to set an absolute tablet region because it makes mapping a tablet to a window really easy by just using the window's position and size from `hyprctl clients`

JManch:
I've added an `absolute_region_position` option that, when true, treats the `region_position` as absolute. Otherwise it's treated as relative like before. It doesn't apply when the tablet is bound to an output.

JManch:
Just tested wlx-overlay-s with one of my monitors in negative space and the overlay pointer problem is back so need to figure out what's going on there...

JManch:
Turns out it was an upstream issue galister/wlx-capture#3 so this is ready for review

src/managers/PointerManager.cpp
```diff
@@ -668,30 +668,46 @@ void CPointerManager::move(const Vector2D& deltaLogical) {
...
 
     if (!std::isnan(abs.x))
         abs.x = std::clamp(abs.x, 0.0, 1.0);
     if (!std::isnan(abs.y))
         abs.y = std::clamp(abs.y, 0.0, 1.0);
 
-    // in logical global
-    CBox mappedArea = currentMonitor->logicalBox();
+    // find x and y size of the entire space
+    Vector2D bottomRight = {-9999999, -9999999}, topLeft = {9999999, 9999999};
```
vaxerski, line 680: g_pCompositor->m_vMonitors.at(0).vecPosition, g_pCompositor->m_vMonitors.at(0).vecPosition + g_pCompositor->m_vMonitors.at(0).vecSize

then skip iterating the first one

JManch:
I've added the ability to set `input:tablet:output` to `current` for replicating the old behaviour.

I could make `current` the default to keep default behaviour the same as it is now but, as @Agent00Ming said, this would make the OTD absolute mode map differently to artist mode.

Agent00Ming (approved):
I think that's a reasonable compromise for now as pointer devices lack the configuration granularity that tablet devices have. Tablet tested.

vaxerski (approved):
wiki mr needed

vaxerski:
oops conflict

vaxerski (approved):

vaxerski:
clang-format missing xdddddd

JManch:
I don't think it's from my diff

izmyname:
> I don't think it's from my diff

0.45 broke the stuff, prolly

JManch:
Rotated tablet works for me. In what way does it not work?

Fxzzi:
how i have rotated it on that screenshot causes the mouse cursor to still be able to reach my monitor on the left, which should not be possible considering how I have mapped the tablet area to the display. Moving the area over to the bottom right of the tablet (top left physically, as it is rotated), seems to work fine: 
image

PR #8360, opened 2024-11-05
default/config: move shadow options to decoration:shadow

Kam1k4dze:
Yes

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -86,10 +86,12 @@ decoration {
     active_opacity = 1.0
     inactive_opacity = 1.0
 
-    drop_shadow = true
-    shadow_range = 4
-    shadow_render_power = 3
-    col.shadow = rgba(1a1a1aee)
+    shadow {
+        enabled = true
+        range = 4
+        render_power = 3
+        color = rgba(1a1a1aee)
+    }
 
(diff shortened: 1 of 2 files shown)


izmyname:
Is this gonna break existing configs?

Kam1k4dze:
No, It just fixes the default config. After https://github.com/hyprwm/Hyprland/commit/d1638a09bacd84b994de3f77746b74f427b9d41e, it gives errors because shadow options have now moved to `decoration:shadow`.

vaxerski (approved):
right, forgot. Thanks!

cowboyai:
it totally breaks existing configs that auto-populate hyprland.conf, such as stylix.
see [Stylix PR](https://github.com/danth/stylix/pull/623)

vaxerski:
that is their problem, not ours, no?

cowboyai:
indeed it is, but it does break existing things without issuing a warning of the change first and supporting both models during the transition. For Nix, this seems to be a worthwhile model, if you detect the old model in Nix, convert it to the new model and issue a warning for a version or two. I see this in most things supporting nixpkgs.

vaxerski:
supporting both would be a PITA. You can always query version with `hyprctl version` :)

PR #8370, opened 2024-11-06
defaultConfig: improve default groupbar appearance

Aqa-Ib:
This is subjective of course (this is my personal config), but I feel like this would be better for new users than the current one (suggestions welcome).
Cheers

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -155,6 +155,25 @@ master {
     new_status = master
 }
 
+# See https://wiki.hyprland.org/Configuring/Variables/#group for more variables to configure groups behavior and appearance
+group {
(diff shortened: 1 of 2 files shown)


vaxerski:
groupbar generally needs some improvements heh

vaxerski:
idk imho its not much better, its almost the same

Aqa-Ib:
Yeah it is not that great. Maybe some ricer on discord or something could help improving this, I will ask.

Aqa-Ib:
I have created an issue to discuss this. Maybe rather than modifying the defaultConfig, would be better to change the default values for the involved variables when we settle on a good rice. https://github.com/hyprwm/Hyprland/issues/8375

Issue #8400, opened 2024-11-09
Overuse of execAndGet


WavyEbuilder:
### Description

Hi,

Marking as a feature request as this isn't necessarily a bug. I thought I'd open an issue to gauge your opinion before spending a large amount of time, hope you understand.

While reading Hyprland's source code, there seems to be a lot of use of `execAndGet` for `git` (where I think libgit2 should probably be used instead), for `grep`, and for `cat`, and for getting the date using `date` (instead of `std::chrono`). While not necessarily an outright issue, I think it's worth using some more standard language tooling from STL instead of shelling out in a lot of these cases.

I'd be happy to work a bit on this myself later if that's of interest.

Thanks.

vaxerski:
completely fine, I just used them cuz it was faster to do so (development-wise) and the small performance differences did not matter

WavyEbuilder:
Perfect, thanks for the response. I'll send a few patches through shortly

izmyname:
Should the issue be open?

vaxerski:
I'd assume OP wants to get rid of all of them, so no

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #8401, opened 2024-11-09
debug: clean up opening of files in HyprCtl

WavyEbuilder:
`std::ifstream` is more suited than `execAndGet` here.

Works on part of https://github.com/hyprwm/Hyprland/issues/8400

Just a quick note:

Regarding `result += "error";` for the case where `/proc/driver/nvidia/version` or `/etc/os-release` could not be opened, I'm attempting to match what `execAndGet` does:
```c++
// Execute a shell command and get the output
std::string execAndGet(const char* cmd) {
    CProcess proc("/bin/sh", {cmd});

    if (!proc.runSync())
        return "error";

    return proc.stdOut();
}
```
i.e. by appending `"error"` to `result`.

I don't mind adjusting this to be a bit more detailed or to print something else if that would be desired. 

Should be good other than the above.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1,6 +1,7 @@
 #include "HyprCtl.hpp"
 
 #include <format>
+#include <fstream>
 #include <netinet/in.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


src/debug/HyprCtl.cpp
```diff
@@ -949,11 +950,29 @@ std::string systemInfoRequest(eHyprCtlOutputFormat format, std::string request)
...
+                }
+            }
+            file.close();
+        } else
+            result += "error";
+    }
     result += "\n\n";
 
-    result += "os-release: " + execAndGet("cat /etc/os-release") + "\n\n";
+    std::ifstream file("/etc/os-release");
```
vaxerski, line 969: this file is not closed properly (will linger until the end of the func), and it should be local probably.

```cpp
if (std::ifstream file("/etc/os-release"); file.is_open()) {
   // ...
}
```

src/debug/HyprCtl.cpp
```diff
@@ -949,11 +950,29 @@ std::string systemInfoRequest(eHyprCtlOutputFormat format, std::string request)
...
 #endif
     result += "GPU information: \n" + GPUINFO;
-    if (GPUINFO.contains("NVIDIA") && std::filesystem::exists("/proc/driver/nvidia/version"))
-        result += execAndGet("cat /proc/driver/nvidia/version | grep NVRM");
+    if (GPUINFO.contains("NVIDIA") && std::filesystem::exists("/proc/driver/nvidia/version")) {
+        std::ifstream file("/proc/driver/nvidia/version");
+        std::string   line;
+        if (file.is_open()) {
+            while (std::getline(file, line)) {
+                if (line.find("NVRM") != std::string::npos) {
```
vaxerski, line 958:
```cpp
if (!line.contains("NVRM"))
   continue;
```

src/debug/HyprCtl.cpp
```diff
@@ -949,11 +950,29 @@ std::string systemInfoRequest(eHyprCtlOutputFormat format, std::string request)
...
+                    result += line;
+                    result += "\n";
+                }
+            }
+            file.close();
```
vaxerski, line 963: redundant?

WavyEbuilder: yea, deconstructor calls it automatically. Sorted

src/debug/HyprCtl.cpp
WavyEbuilder: Done

src/debug/HyprCtl.cpp
WavyEbuilder: Done

vaxerski (approved):
thanks!

PR #8413, opened 2024-11-10
debug: clean up fetching of the contents of /proc/device-tree

WavyEbuilder:

Works a bit more on https://github.com/hyprwm/Hyprland/issues/8400, should basically get done what can reasonably be cleaned up for HyprCtl.cpp

I attempted to match the behavior of cat again, but I'm not sure if you want an error printing if `/proc/soc*/gpu*` can't be matched. Happy to add that if desired.

Should be ready.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1,7 +1,9 @@
 #include "HyprCtl.hpp"
 
+#include <algorithm>
 #include <format>
 #include <fstream>
+#include <iterator>
 #include <netinet/in.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


src/debug/HyprCtl.cpp
```diff
@@ -945,7 +947,22 @@ std::string systemInfoRequest(eHyprCtlOutputFormat format, std::string request)
 #if defined(__DragonFly__) || defined(__FreeBSD__)
     const std::string GPUINFO = execAndGet("pciconf -lv | fgrep -A4 vga");
 #elif defined(__arm__) || defined(__aarch64__)
-    const std::string GPUINFO = execAndGet("cat /proc/device-tree/soc*/gpu*/compatible");
+    std::string                 GPUINFO;
+    const std::filesystem::path dev_tree = "/proc/device-tree";
+    if (std::filesystem::exists(dev_tree) && std::filesystem::is_directory(dev_tree)) {
```
vaxerski, line 952: all of this might throw so probably should be in a try catch

WavyEbuilder: The entire `if` block? Just catch `...` (and I guess like `execAndGet` set `GPUINFO` to `"error"`) okay?

vaxerski: every std::filesystem call can throw unless supplied with a std::error_code as the second arg, so the entire thing

WavyEbuilder: Alright, will do

WavyEbuilder: Done

Issue #1147, opened 2022-12-02
Pinned window cannot be made fullscreen without unpinning first


mcginty:
This is at least true with mpv: if I'm playing a video that I've made floating, I can fullscreen. However, if I then pin it and try to fullscreen it won't work until I unpin it first.

It *seems* like it should be possible to fullscreen while pinned.

Thanks for making Hyprland!

vaxerski:
intentional design.

Fullscreening a pinned window would have to remove its pinned status. More intuitive to leave as is.

maximbaz:
Hi @vaxerski, there are a couple downsides with leaving it as it is that I wanted to share with you: 

- the native app's bindings to toggle fullscreen status (mpv's `f`, in this case) "suddenly" don't work if window is pinned
- it isn't possible to make a hyprland keybinding that would just toggle fullscreen status of floating pinned window

For the latter, instead of `pin` dispatcher which currently _toggles_ pinned status, we'd have to do something like:

```
bind SUPER, f, pin, disable (not toggle, always disable)
bind SUPER, f, fullscreen
bind SUPER, f, pin, enable
```

And use this binding (also for non-floating windows, which feels a bit weird, but I wouldn't want to have two separate shortcuts to toggle fullscreen).
This idea also assumes, that I want _all_ floating windows to be pinned, which might not be the case (I could personally live with it).

Maybe Hyprland could do this logic internally on `fullscreen` dispatcher? But to make it proper, it'd have to keep a state of whether the app was pinned, before it enters fullscreen, and restore that status upon exit?

Coming from sway, I want also to share how sway does this: it simply doesn't remove the pinned state from the fullscreen window! In other words, if you are on a workspace 1, you fullscreen a pinned window, press binding to go to workspace 2, you still see the fullscreen video, but when you unfullscreen it, you'll see that you are in fact on workspace 2 now.

vaxerski:
> Coming from sway, I want also to share how sway does this: it simply doesn't remove the pinned state from the fullscreen window! In other words, if you are on a workspace 1, you fullscreen a pinned window, press binding to go to workspace 2, you still see the fullscreen video, but when you unfullscreen it, you'll see that you are in fact on workspace 2 now.

nope it's definitely not gonna work like that.

I can make it remove the pin status, but definitely not this.

maximbaz:
[...]

PR #8426, opened 2024-11-11
binds: option to fullscreen pinned window

littleblack111:
fixes https://github.com/hyprwm/Hyprland/issues/1147

if u need, i can store the pin state somewhere when fullscreened and restore it back afterwards

ready


vaxerski:
this is not that easy and will break

PR #8431, opened 2024-11-11
Enable explicit out fence in AQ

UjinT34:
Enables explicit out fences

Requires https://github.com/hyprwm/aquamarine/pull/105


vaxerski:
@fufexan nix needs aq bump

vaxerski (approved):

Fxzzi:
After this MR, osu!lazer is now locked to my screen refresh rate of 170Hz. Whilst previously it could go above to 340fps, it can no longer and it always stays below that.

UjinT34:
https://github.com/hyprwm/Hyprland/issues/8482 kms explicit sync prevents tearing

Fxzzi:
> #8482 kms explicit sync prevents tearing

Ty, is this intended behaviour? Id assume not

UjinT34:
Not really. It is unclear how much of the explicit sync should be disabled and when. The issue comes from the explicit out fence support. It could be conditionally disabled when tearing is active but it doesn't feel right.

ferrreo:
@UjinT34 kwin seems to handle this fine, it looks like it doesn't fence while tearing: https://invent.kde.org/plasma/kwin/-/merge_requests/4800/diffs

PR #8476, opened 2024-11-14
config: update the configStringToInt implementation

PaideiaDilemma:
Copied from hyprlang and removed std::expected.

Allows to use `rgb(a)` with decimal values for some options. For example gradients.

I did not test all possible strings that should parse to an int. But I tested all the obvious ones.

Ready.


vaxerski (approved):
wiki mr

Issue #8443, opened 2024-11-12
mouse clicks registering behind topmost floating window


warrenfalk:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.45.0 built from branch  at commit a29cfa78431a054a093f3c843228bac6783a6d33  ().
Date: 2024-11-12
Tag: , commits: 5449
built against aquamarine 0.4.3


flags set:
debug


System Information:
System name: Linux
Node name: warren-desktop
Release: 6.11.7
Version: #1-NixOS SMP PREEMPT_DYNAMIC Fri Nov  8 15:31:04 UTC 2024


GPU information: 
2d:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] [1002:67df] (rev ef) (prog-if 00 [VGA controller])


os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.11.20241109.76612b1"
CPE_NAME="cpe:/o:nixos:nixos:24.11"
DEFAULT_HOSTNAME="warren-desktop"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
ID_LIKE=""
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.11 (Vicuna)"
SUPPORT_URL="https://nixos.org/community.html"
VARIANT=""
VARIANT_ID=""
VENDOR_NAME=NixOS
VENDOR_URL="https://nixos.org/"
VERSION="24.11 (Vicuna)"
VERSION_CODENAME=vicuna
VERSION_ID="24.11"


plugins:

======Config-Start======


Config File: /home/warren/.config/hypr/hyprland.conf: Read Succeeded
#
# Please note not all available settings / options are set here.
# For a full list, see the wiki
#
exec=dbus-hyprland-environment
exec=hyprpaper --config /home/warren/.config/hyprpaper.conf

# See https://wiki.hyprland.org/Configuring/Monitors/
monitor=DP-2,3840x2160@120,auto,1.25
monitor=DP-3,3840x2160@120,auto,1.25

[...]

PR #8494, opened 2024-11-16
fix changeWindowZOrder

Aqa-Ib:
Fixes https://github.com/hyprwm/Hyprland/issues/8443

ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1389,6 +1389,9 @@ void CCompositor::changeWindowZOrder(PHLWINDOW pWindow, bool top) {
     if (!validMapped(pWindow))
         return;
 
+    if (top)
+        pWindow->m_bCreatedOverFullscreen = true;
+
     if (pWindow == (top ? m_vWindows.back() : m_vWindows.front()))
         return;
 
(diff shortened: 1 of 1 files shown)


Aqa-Ib:
This was introduced in https://github.com/hyprwm/Hyprland/pull/6877
@UjinT34 do you have an insight ?

UjinT34:
Without this commit https://github.com/hyprwm/Hyprland/pull/6877/commits/507df67fa504522bd907ad1e899fefeb5364bdf3 every click on a fullscreen floating window will result in unneeded reordering which will add damage to monitor and schedule an unwanted frame. Maybe it returns too early. `g_pHyprRenderer->damageMonitor` shouldn't be called when the clicked window is already on top and fullscreen.

Aqa-Ib:
OK, I found that it can't return before executing `pWindow->m_bCreatedOverFullscreen = true;` because `CInputManager::mouseMoveUnified` needs that in order to work properly.

So I moved that operation before the early return. I think that this is fine in terms of performance; could it be done better or is it good enough?

vaxerski (approved):

vaxerski:
@UjinT34 is this version fine? (vrr stuff)

UjinT34:
It's fine

PR #8514, opened 2024-11-18
Adding nvidia variables as coments to hyprland.conf in case user uses…

MrDraxs:
… nvidia

simplifies the process of setting nvidia on hyperland for users

i believe is ready for merging, i may be wrong tough.

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -51,6 +51,17 @@ $menu = wofi --show drun
 env = XCURSOR_SIZE,24
 env = HYPRCURSOR_SIZE,24
 
+##############
+### NVIDIA ###
+##############
+
+# if you use nvidia is recomended to uncoment this section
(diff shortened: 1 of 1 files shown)


fufexan (commented):
I'd rather not include this. It's duplicating work from the wiki. Also, not all the variables are correct (and that should be updated in the wiki if it wasn't already).

vaxerski:
as above. Especially with nvidia, we really want to encourage reading the wiki, as it's important

MrDraxs:
> I'd rather not include this. It's duplicating work from the wiki. Also, not all the variables are correct (and that should be updated in the wiki if it wasn't already).

what variables are wrong?

fufexan:
There have been issues regarding the GBM_BACKEND variable. Not sure about other variables.

PR #8536, opened 2024-11-21
draft: begin using CFileDescriptor instead of manual fd handling

gulafaran:
depends on https://github.com/hyprwm/hyprutils/pull/21

so far changed most non protocol fd handling to CFileDescriptors, the protocols il leave for future PR when AQ gets same treatment aswell since some fds are cojoined into AQ and viceversa there.

gonna need a bit of lookover and testing.


gulafaran:
Gonna poke around in AQ before, hitting a few struggles.

PR #8538, opened 2024-11-21
actually suppress internal keybinds instead of passing them along

m4rch3n1ng:

previously when switching vt hyprland would print bullshit to the terminal[^1], because it changes what a `true` boolean means somewhere in the middle without flipping one of them.

in [`CKeybindManager::handleVT`](https://github.com/hyprwm/Hyprland/blob/940f7aa990dbc99815bab8d355999d8277534b17/src/managers/KeybindManager.cpp#L831-L832) it returns a `true` when the key should be intercepted, which is passed up by [`CKeybindManager::handleInternalKeybinds`](https://github.com/hyprwm/Hyprland/blob/940f7aa990dbc99815bab8d355999d8277534b17/src/managers/KeybindManager.cpp#L864-L866), and then handled in [`CKeybindManager::onKeyEvent`](https://github.com/hyprwm/Hyprland/blob/940f7aa990dbc99815bab8d355999d8277534b17/src/managers/KeybindManager.cpp#L426-L427), where it also returns a `true`. but [`CInputManager::onKeyboardKey`](https://github.com/hyprwm/Hyprland/blob/940f7aa990dbc99815bab8d355999d8277534b17/src/managers/input/InputManager.cpp#L1307) interprets a `true` to mean that it should [pass the event along](https://github.com/hyprwm/Hyprland/blob/940f7aa990dbc99815bab8d355999d8277534b17/src/managers/input/InputManager.cpp#L1311-L1323), which is not what we want for internal keybinds.

i didn't check if the boolean stuff was correct for everything else

ready

[^1]: image 
image

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -423,8 +423,10 @@ bool CKeybindManager::onKeyEvent(std::any event, SP<IKeyboard> pKeyboard) {
     const xkb_keysym_t keysym         = xkb_state_key_get_one_sym(pKeyboard->resolveBindsBySym ? pKeyboard->xkbSymState : m_pXKBTranslationState, KEYCODE);
     const xkb_keysym_t internalKeysym = xkb_state_key_get_one_sym(pKeyboard->xkbState, KEYCODE);
 
+    // handleInternalKeybinds returns true when the key should be suppressed,
+    // while this function returns true when the key event should be sent
     if (handleInternalKeybinds(internalKeysym))
-        return true;
+        return false;
 
     const auto MODS = g_pInputManager->accumulateModsFromAllKBs();
 


vaxerski (approved):

vaxerski:
thanks!

msmafra:
Should it have fixed this?

image

Pressing non-assigned keybinds, in the example SUPER+Q, SUPER+], SUPER+' etc.

m4rch3n1ng:
no that's intentional: if a keybind is unbound it is not a keybind and should be forwarded to the application to handle. kitty will then print the escape sequence. other terminal emulators like alacritty and ghostty will not print the escape sequence and instead just "ignore" the modifiers and just print q, ' or ].

what this pr fixes, is that internal keybinds (like switching to a different vt), that should have been intercepted and not sent to the application, (abecause they are handled by the compositor, were not actually intercepted.

Issue #8570, opened 2024-11-24
Hyprland crashes when client fails to send the correct SHM size to wl_shm_create_pool


nabil-otsmane:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.45.0 built from branch main at commit 940f7aa990dbc99815bab8d355999d8277534b17  (renderer: fixup blur optimization considitons).
Date: Wed Nov 20 15:02:21 2024
Tag: v0.45.0-51-g940f7aa9, commits: 5484
built against aquamarine 0.5.0


flags set:
debug


System Information:
System name: Linux
Node name: nabil-20vd
Release: 6.10.13-3-MANJARO
Version: #1 SMP PREEMPT_DYNAMIC Tue Oct  8 03:24:49 UTC 2024


GPU information: 
0000:00:02.0 VGA compatible controller [0300]: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] [8086:9a49] (rev 01) (prog-if 00 [VGA controller])


os-release: NAME="Manjaro Linux"
PRETTY_NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://forum.manjaro.org/"
BUG_REPORT_URL="https://docs.manjaro.org/reporting-bugs/"
PRIVACY_POLICY_URL="https://manjaro.org/privacy-policy/"
LOGO=manjarolinux


plugins:

======Config-Start======


Config File: /home/nabil/.config/hypr/hyprland.conf: Read Succeeded

# #######################################################################################
# AUTOGENERATED HYPR CONFIG.
# PLEASE USE THE CONFIG PROVIDED IN THE GIT REPO /examples/hypr.conf AND EDIT IT,
# OR EDIT THIS ONE ACCORDING TO THE WIKI INSTRUCTIONS.
# #######################################################################################

# autogenerated = 1 # remove this line to remove the warning

# This is an example Hyprland config file.
# Refer to the wiki for more information.
# https://wiki.hyprland.org/Configuring/Configuring-Hyprland/

# Please note not all available settings / options are set here.
# For a full list, see the wiki

# You can split this configuration into multiple files
[...]

PR #8572, opened 2024-11-24
protocols: fix shm fd size check before creating or resizing shm_pool

nabil-otsmane:
Hyprland is crashing when trying to access SHM pool buffer because of the lack of size validation on the fd sent from the client.

To reproduce this bug please check this issue #8570

The code might need some refactoring, and the check function can be moved somewhere else.

Please let me know anything needs to be changed.

Yes

### DIFF
--- a/src/protocols/core/Shm.cpp
+++ b/src/protocols/core/Shm.cpp
@@ -100,7 +100,7 @@ void CSHMPool::resize(size_t size_) {
         LOGM(ERR, "Couldn't mmap {} bytes from fd {} of shm client", size, fd);
 }
 
-int shmIsSizeValid(int fd, size_t size) {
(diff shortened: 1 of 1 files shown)


src/protocols/core/Shm.cpp
```diff
@@ -99,10 +100,25 @@ void CSHMPool::resize(size_t size_) {
         LOGM(ERR, "Couldn't mmap {} bytes from fd {} of shm client", size, fd);
 }
 
+int shmIsSizeValid(int fd, size_t size) {
```
vaxerski, line 103: this is fine, just make it `static` please

nabil-otsmane: Oh yes of course

vaxerski (approved):

PR #8584, opened 2024-11-25
sessionLock: don't send motion events on every surface commit

AThePeanut4:

This fixes two bugs that cause https://github.com/jirutka/swaylock-effects/issues/68:
- In a multi-monitor setup, the `pWlrSurface != g_pCompositor->m_pLastFocus` check (added in #7647) will always be false for all but the currently focused surface, meaning that a motion event is sent (through the `simulateMouseMovement()` call) after a commit to any non-focused surface. In the case of swaylock, this results in a motion event around once every second, which is treated as a mouse movement thus preventing `--grace` from working properly. The fix is to only call `simulateMouseMovement()` from the surface map/commit listeners if there is no currently focused surface.
- When the session lock surface is first mapped, it unconditionally receives a pointer motion event, due to #6658. Swaylock again treats this as a normal mouse movement, and therefore unlocks itself immediately after locking if `--grace` is specified. The fix here is to not call `sendPointerMotion()` when the first surface is focused.

As mentioned in https://github.com/jirutka/swaylock-effects/issues/68, the actual issue can be easily solved by just ignoring the motion event if the cursor didn't actually move - I believe this is what hyprlock does. However, it doesn't really make sense to send a motion event after the pointer enter event because the pointer didn't actually move. I think it could be useful to combine `setPointerFocus` and `sendPointerMotion` into one method that either sends exit/enter events if the surface is not focused, or just sends a motion event if it is.

Ready

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -31,7 +31,7 @@ SSessionLockSurface::SSessionLockSurface(SP<CSessionLockSurface> surface_) : sur
     listeners.commit = surface_->events.commit.registerListener([this](std::any data) {
         const auto PMONITOR = g_pCompositor->getMonitorFromID(iMonitorID);
 
-        if (mapped && pWlrSurface != g_pCompositor->m_pLastFocus)
+        if (mapped && !g_pCompositor->m_pLastFocus)
             g_pInputManager->simulateMouseMovement();
 
         if (PMONITOR)


vaxerski:
> the actual issue can be easily solved by just ignoring the motion event if the cursor didn't actually move - I believe this is what hyprlock does

The actual issue can be easily solved by making the "grace" option not stupid and take a delta instead of just blindly reacting to any movement. Hyprlock IIRC will unlock in grace if Δ > 5px

src/managers/SessionLockManager.cpp
```diff
@@ -13,7 +13,8 @@ SSessionLockSurface::SSessionLockSurface(SP<CSessionLockSurface> surface_) : sur
     listeners.map = surface_->events.map.registerListener([this](std::any data) {
         mapped = true;
 
-        g_pInputManager->simulateMouseMovement();
+        if (!g_pCompositor->m_pLastFocus)
```
vaxerski, line 16: doing this may prevent the lock surface from being focused

src/managers/input/InputManager.cpp
```diff
@@ -216,6 +216,8 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus) {
         g_pCompositor->setActiveMonitor(PMONITOR);
 
     if (g_pSessionLockManager->isSessionLocked()) {
+        const bool notFirstFocus = g_pCompositor->m_pLastFocus;
```
vaxerski, line 219: what if going from focused window/layer/whatever -> session lock?

src/managers/SessionLockManager.cpp
AThePeanut4: The focus is reset when the session is locked here: https://github.com/hyprwm/Hyprland/blob/1fb720b62aeb474873ba43426ddc53afde1e6cdd/src/managers/SessionLockManager.cpp#L88
so `g_pCompositor->m_pLastFocus` should always be null after the session is locked.

But I'm happy to remove the check, it doesn't affect the swaylock issue and just avoids extra unnecessary motion events.

src/managers/input/InputManager.cpp
AThePeanut4: For the same reason as above, `g_pCompositor->m_pLastFocus` is null after the session is locked so this will always be false until the lock surface is focused for the first time.

AThePeanut4:
> > the actual issue can be easily solved by just ignoring the motion event if the cursor didn't actually move - I believe this is what hyprlock does
> 
> The actual issue can be easily solved by making the "grace" option not stupid and take a delta instead of just blindly reacting to any movement. Hyprlock IIRC will unlock in grace if Δ > 5px

I think it's reasonable to expect motion events to mean the mouse/pointer actually moved rather than some internal "simulated" zero-distance mouse movement (i.e. `simulateMouseMovement()`). But if you don't agree no problem, I'm happy to undo ac7dadc (and add a fix to swaylock). But the multi-monitor motion event on every commit issue I think is pretty clearly a bug.

src/managers/SessionLockManager.cpp
AThePeanut4: Actually nevermind just checked, removing it results in extra motion events if the first surface mapped was the one under the cursor and was thus focused (the subsequent calls to `mouseMoveUnified` will have `notFirstFocus == true` and will therefore produce motion events).

vaxerski:
I am generally unsure this is a problem in hyprland, like, at all. The protocol does not specify any requirements for motion events, so it's up to us to determine how and when to use them

AThePeanut4:
Alright, fair. I've removed everything but the fix for motion events on every surface commit in a multi monitor setup.

Issue #8434, opened 2024-11-12
dwindle: window splits can result in windows that are higher than wide


Atemu:
### Description

Currently, the split is determined based on the W/H ratio of the current (then parent) node:

> The split is determined dynamically with the W/H ratio of the parent node. If W > H, it’s side-by-side. If H > W, it’s top-and-bottom.

With windows that are quite square already however, this can lead to windows being spawned that are quite a bit higher than they are wide. For instance: The default multiplier splits a `1716,1438` into two `856,1438` windows.

This is because it doesn't take into account that a vertical split will result in windows that are half as wide. You can work around that by setting `dwindle:split_width_multiplier = 2`.  
This would arguably make for a better default value for that variable; should we change it?  

This is imperfect though as it doesn't take window borders and spacing into account which can can differ based on direction. 

The perfect solution would instead simulate what the split would actually do to the ratios of the resulting windows and then choose a vertical split only when it results in two windows where W >= H.

vaxerski:
MRs welcome

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #8598, opened 2024-11-27
dwindle: split windows horizontally if height becomes greater than width

poulsena:
Changes dwindle layout to split horizontally when the new windows height would become greater than its width.
Addresses #8434 

The issue mentions changing the default value of dwindle:split_width_multiplier.
This PR does not do that.

Ready


vaxerski:
I think this solution sucks, personally. You:
- nuke people's muscle memory
- this is not how most bsp tilers work
- this can be achieved with split_width_multiplier if desired

PR #8601, opened 2024-11-28
Modified the  `movefocus` dispatcher to prioritize focus change within groups. 

normaltaro:
It modifies the `movefocus` dispatcher logic to switch directional focus within groups first:
- Check if the current window is part of a Group.
- `movefocus l/r`  can now be used to switch tabs within the Group
- If the current window is the first/last tab of its group OR `movefocus u/d` is pressed, then the usual directional focus logic applies.

In other words, a sideways directional focus change is prioritized within groups first.

##### How I think it benefits users?
  + I think most users (like me) expect the directional flow of `movefocus` dispatcher to be followed within groups as it does in other window managers. I felt locked out of tabbed groups and never really used them because of this.
  + It is not possible to do so using `changegroupactive` along with `movefocus` as they require separate keybinds which do not conflict.
  + This keeps the last active index intact when focus is moved out of the group.
  + Personally, I think the overall experience with this patch is much better compared to other window managers regarding tabbed groups (like `i3`)
+ Example:
  + `mod h/j/k/l` with `movefocus l/r/u/d` is intuitive for groups too.

I haven't found any bugs. It works as per the logic described above.

I have tried with all kinds of group orientations and it works good. It is a single block of 13 lines and could be merged as-is. :)


vaxerski:
`hyprctl dispatch changegroupactive f | grep "next" && hyprctl dispatch movefocus r`?

vaxerski:
should be an option, some people might not like this

normaltaro:
how about another dispatcher, like `movefocusingroups`?

vaxerski:
I was thinking about simply a bool flag in binds:

normaltaro:
added bool option variable `movefocus_cycles_groupfirst` to `binds`.

vaxerski (approved):
thanks! just a wiki mr needed

izmyname:
Just wanna say thanks. The only feature I missed when I switched from i3 to Hyprland.

izmyname:
Also, some thoughts. Would it be possible to add something like tab focus cycling if a group window is the only window on a workspace? For example:  right now, if a windowgroup has 3 tabs, using 'movefocus, r' goes like 1>2>3 instead of 1>2>3>1>2>3.

normaltaro:
> Also, some thoughts. Would it be possible to add something like tab focus cycling if a group window is the only window on a workspace? For example: right now, if a windowgroup has 3 tabs, using 'movefocus, r' goes like 1>2>3 instead of 1>2>3>1>2>3.

damn, i didn't think of that. its probably a small change in the `if` condition too. i'll try for another pr to fix/add it.

PR #8623, opened 2024-11-30
Add pseudo field for windowrulev2

fliplus:
As the title says, this adds a "pseudo:" field for windowrulev2.

It's ready.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1236,6 +1236,7 @@ std::vector<SWindowRule> CConfigManager::getMatchingRules(PHLWINDOW pWindow, boo
 
     // since some rules will be applied later, we need to store some flags
     bool hasFloating   = pWindow->m_bIsFloating;
+    bool hasPseudo     = pWindow->m_bIsPseudotiled;
     bool hasFullscreen = pWindow->isFullscreen();
 
     // local tags for dynamic tag rule match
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
hm. I don't like this. Pseudo is... technically layout-specific. Maybe we should add some method to define layout-specific windowrules? like `layout:...`?

fliplus:
> hm. I don't like this. Pseudo is... technically layout-specific. Maybe we should add some method to define layout-specific windowrules? like `layout:...`?

Indeed, `pseudo:` is technically layout-specific, but the window still holds the `pseudo` attribute on every layout.
That being said, I [added](https://github.com/hyprwm/Hyprland/pull/8623/commits/72435a14b7b34d702384fbc78f0982f2d829d6e8) a `layout:` field and made so `updateDynamicRules()` is called on all windows every time `handleWindowRuleV2()` is ran.

vaxerski:
hm, I am unsure how to deal with this... nicely. Considering pseudo is exposed to everything, I'd say let's just keep it the old way for now I guess.. I'll do something in the future with this

zakk4223:
Stupid question/suggestion. Can't you just use window tagging for some of this? Layouts change behavior/do magic stuff based on existence of some window tag(s). Users can tag windows however they want (rules, manually etc) and stuff Just Works[tm].

Just document which tags do what for the layouts.

I dunno, just seems like the tag system could lead to some pretty useful/powerful stuff layout wise if the layouts wanted to take advantage. I guess the one downside is tags can't have values, hmmm.

You'd still have the 'layout specific and possibly colliding' thing either way tho. Although some tags might make sense to share between layouts (all master/stack based layouts could use a 'master' tag on windows to place them into the master stack, etc)

vaxerski:
closing due to inactivity

PR #8625, opened 2024-12-01
Ignore reserved areas when centering master window

UjinT34:
Adds an option `master:center_ignores_reserved` to ignore reserved areas while calculating master window size and position. Makes master window always centered on monitor. Reserved areas on left and right will affect slave columns only.
Needs `master:orientation = center`.

With very high `master:mfact` and large reserved areas the master window size and position might overlap with reserved areas. I don't think that such a situation can happen in any usable setup.

Ready


vaxerski (approved):
k

PR #8634, opened 2024-12-03
hyprpm: target installed instead of running version

RoootTheFox:
After updating Hyprland, `hyprpm update` will still build plugins against the currently running version of Hyprland.
This is a little annoying, because you first have to restart Hyprland and *then* run `hyprpm update` for your plugins to be built for the correct version.
This PR makes `hyprpm` build against the *installed* version of Hyprland instead of the running version, and shows a popup notification if the running Hyprland version differs from the installed one after updating, telling users to restart Hyprland.
example

When I tested it on my setup (Arch) everything worked fine, so no I don't think so. Not sure how Nix handles stuff, but it should work there too I suppose.
I did change the signature of hyprpm's `getHyprlandVersion` func, but gave it a default value of true that corresponds to the original behavior.

Should be ready for merging, although if you have any feedback (e.g. UX wise) go for it

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -50,7 +50,7 @@ static std::string getTempRoot() {
     return STR;
 }
 
-SHyprlandVersion CPluginManager::getHyprlandVersion(bool running = true) {
+SHyprlandVersion CPluginManager::getHyprlandVersion(bool running) {
     static bool             onceRunning   = false;
     static bool             onceInstalled = false;
     static SHyprlandVersion verRunning;
@@ -868,10 +868,12 @@ ePluginLoadStateReturn CPluginManager::ensurePluginsLoadState() {
             if (std::find_if(loadedPlugins.begin(), loadedPlugins.end(), [&](const auto& other) { return other == p.name; }) != loadedPlugins.end())
                 continue;
 
-            if (!loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true))
+            if (!loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true)) {
+                std::println("{}", infoString("{} will be loaded after restarting Hyprland", p.name));
                 hyprlandVersionMismatch = true;
-
-            std::println("{}", successString("Loaded {}", p.name));
+            } else {
+                std::println("{}", successString("Loaded {}", p.name));
+            }
         }
     }
 
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -50,17 +50,26 @@ static std::string getTempRoot() {
     return STR;
 }
 
-SHyprlandVersion CPluginManager::getHyprlandVersion() {
-    static SHyprlandVersion ver;
-    static bool             once = false;
+SHyprlandVersion CPluginManager::getHyprlandVersion(bool running = true) {
+    static bool             once_running   = false;
+    static bool             once_installed = false;
+    static SHyprlandVersion ver_running;
+    static SHyprlandVersion ver_installed;
 
-    if (once)
-        return ver;
+    if (once_running && running)
+        return ver_running;
 
-    once                 = true;
-    const auto HLVERCALL = execAndGet("hyprctl version");
+    if (once_installed && !running)
+        return ver_installed;
+
+    if (running)
+        once_running = true;
+    else
+        once_installed = true;
+
+    const auto HLVERCALL = running ? execAndGet("hyprctl version") : execAndGet("Hyprland --version");
     if (m_bVerbose)
-        std::println("{}", verboseString("version returned: {}", HLVERCALL));
+        std::println("{}", verboseString("{} version returned: {}", running ? "running" : "installed", HLVERCALL));
 
     if (!HLVERCALL.contains("Tag:")) {
         std::println(stderr, "\n{}", failureString("You don't seem to be running Hyprland."));
@@ -91,7 +100,13 @@ SHyprlandVersion CPluginManager::getHyprlandVersion() {
     if (m_bVerbose)
         std::println("{}", verboseString("parsed commit {} at branch {} on {}, commits {}", hlcommit, hlbranch, hldate, commits));
 
-    ver = SHyprlandVersion{hlbranch, hlcommit, hldate, commits};
+    auto ver = SHyprlandVersion{hlbranch, hlcommit, hldate, commits};
+
+    if (running)
+        ver_running = ver;
+    else
+        ver_installed = ver;
+
     return ver;
 }
 
@@ -356,7 +371,7 @@ bool CPluginManager::removePluginRepo(const std::string& urlOrName) {
 }
 
 eHeadersErrors CPluginManager::headersValid() {
-    const auto HLVER = getHyprlandVersion();
+    const auto HLVER = getHyprlandVersion(false);
 
     if (!std::filesystem::exists(DataState::getHeadersPath() + "/share/pkgconfig/hyprland.pc"))
         return HEADERS_MISSING;
@@ -418,7 +433,7 @@ bool CPluginManager::updateHeaders(bool force) {
 
     DataState::ensureStateStoreExists();
 
-    const auto HLVER = getHyprlandVersion();
+    const auto HLVER = getHyprlandVersion(false);
 
     if (!hasDeps()) {
         std::println("\n{}", failureString("Could not update. Dependencies not satisfied. Hyprpm requires: cmake, meson, cpio, pkg-config"));
@@ -580,7 +595,7 @@ bool CPluginManager::updatePlugins(bool forceUpdateAll) {
         return true;
     }
 
-    const auto   HLVER = getHyprlandVersion();
+    const auto   HLVER = getHyprlandVersion(false);
 
     CProgressBar progress;
     progress.m_iMaxSteps        = REPOS.size() * 2 + 2;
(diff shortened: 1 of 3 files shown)


vaxerski:
sooo... essentially then it builds for the installed version right?

We also need a safeguard in `enable` for this, don't we?

RoootTheFox:
> sooo... essentially then it builds for the installed version right?

yeah.

> We also need a safeguard in `enable` for this, don't we?

i mean, the actual safeguard is in `ensurePluginsLoadState` which is used by enable too where (from my understanding) the actual plugin loading (and unloading) happens

i suppose a notify would be neat though (like how i have it impld in `update`)

i'll go and update upstream rq and then impl that

hyprpm/src/main.cpp
```diff
@@ -114,7 +114,10 @@ int                        main(int argc, char** argv, char** envp) {
 
             auto ret2 = g_pPluginManager->ensurePluginsLoadState();
 
-            if (ret2 != LOADSTATE_OK)
+            if (ret2 == LOADSTATE_HYPRLAND_UPDATED)
+                g_pPluginManager->notify(ICON_INFO, 0, 10000, "[hyprpm] Updated plugins, but Hyprland was updated. Please restart Hyprland.");
+
+            if (ret2 != LOADSTATE_OK && ret2 != LOADSTATE_HYPRLAND_UPDATED)
```
vaxerski, line 120: I think it should return 1 if it's updated too. Don't return success when we didnt succeed

hyprpm/src/main.cpp
```diff
@@ -130,7 +133,11 @@ int                        main(int argc, char** argv, char** envp) {
         }
 
         auto ret = g_pPluginManager->ensurePluginsLoadState();
-        if (ret != LOADSTATE_OK)
+
+        if (ret == LOADSTATE_HYPRLAND_UPDATED)
+            g_pPluginManager->notify(ICON_INFO, 0, 10000, "[hyprpm] Enabled plugin, but Hyprland was updated. Please restart Hyprland.");
+
+        if (ret != LOADSTATE_OK && ret != LOADSTATE_HYPRLAND_UPDATED)
```
vaxerski, line 140: ditto

hyprpm/src/core/PluginManager.cpp
```diff
@@ -50,17 +50,26 @@ static std::string getTempRoot() {
...
 }
 
-SHyprlandVersion CPluginManager::getHyprlandVersion() {
-    static SHyprlandVersion ver;
-    static bool             once = false;
+SHyprlandVersion CPluginManager::getHyprlandVersion(bool running = true) {
+    static bool             once_running   = false;
+    static bool             once_installed = false;
+    static SHyprlandVersion ver_running;
+    static SHyprlandVersion ver_installed;
```
vaxerski, line 57: what is all this snake case, yuck!

RoootTheFox: oops, habits

hyprpm/src/main.cpp
RoootTheFox: done!

RoootTheFox:
somehow missed the other 2 snake_case vars i still had, oops
if you want i can squash the commits into one single commit too

vaxerski:
I always squash dw

hyprpm/src/core/PluginManager.cpp
```diff
@@ -50,17 +50,26 @@ static std::string getTempRoot() {
     return STR;
 }
 
-SHyprlandVersion CPluginManager::getHyprlandVersion() {
-    static SHyprlandVersion ver;
-    static bool             once = false;
+SHyprlandVersion CPluginManager::getHyprlandVersion(bool running = true) {
```
vaxerski, line 53: default params in the header pls not source

Furthermore, why is it true by default when it looks to me like you use false more?

hyprpm/src/core/PluginManager.cpp
```diff
@@ -847,17 +868,27 @@ ePluginLoadStateReturn CPluginManager::ensurePluginsLoadState() {
             if (std::find_if(loadedPlugins.begin(), loadedPlugins.end(), [&](const auto& other) { return other == p.name; }) != loadedPlugins.end())
                 continue;
 
-            loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true);
+            if (!loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true))
+                hyprlandVersionMismatch = true;
```
vaxerski, line 872: this will still make hyprpm print "loaded" when it didn't?

hyprpm/src/core/PluginManager.cpp
RoootTheFox: funnily enough, i had it in the header before. no idea why i moved it

its true by default as that corresponds to the default behavior that the function had before to not accidentally cause incorrect behaviour somewhere else if i missed some statement

hyprpm/src/core/PluginManager.cpp
RoootTheFox: do you prefer to have it print nothing instead, or have it print a warning message?

RoootTheFox: i made it print an info message for now so ill  mark this as resolved, if u want it changed just review it again or smth

hyprpm/src/core/PluginManager.cpp
vaxerski: ok

hyprpm/src/core/PluginManager.cpp
```diff
@@ -847,17 +868,29 @@ ePluginLoadStateReturn CPluginManager::ensurePluginsLoadState() {
             if (std::find_if(loadedPlugins.begin(), loadedPlugins.end(), [&](const auto& other) { return other == p.name; }) != loadedPlugins.end())
                 continue;
 
-            loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true);
-            std::println("{}", successString("Loaded {}", p.name));
+            if (!loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true)) {
+                std::println("{}", infoString("{} will be loaded after restarting Hyprland", p.name));
+                hyprlandVersionMismatch = true;
+            } else {
```
vaxerski, line 874: ahem no {} around short ifs

hyprpm/src/core/PluginManager.cpp
```diff
@@ -829,11 +844,17 @@ ePluginLoadStateReturn CPluginManager::ensurePluginsLoadState() {
...
+    // bcs that means the header version doesn't match the running version
+    // (and Hyprland needs to restart)
+    bool hyprlandVersionMismatch = false;
+
     // unload disabled plugins
     for (auto const& p : loadedPlugins) {
         if (!enabled(p)) {
             // unload
-            loadUnloadPlugin(HYPRPMPATH + repoForName(p) + "/" + p + ".so", false);
+            if (!loadUnloadPlugin(HYPRPMPATH + repoForName(p) + "/" + p + ".so", false))
```
vaxerski, line 856: forgot here

hyprpm/src/core/PluginManager.cpp
RoootTheFox: it won't compile if i get rid of the braces like so:
```
if (!loadUnloadPlugin(HYPRPMPATH + repoForName(p.name) + "/" + p.filename, true))
    std::println("{}", infoString("{} will be loaded after restarting Hyprland", p.name));
    hyprlandVersionMismatch = true;
else
    std::println("{}", successString("Loaded {}", p.name));
```
so i kept them. clang does not know how to deal with this and frankly it looks like python. ill remove them from the lower part tho

hyprpm/src/core/PluginManager.cpp
RoootTheFox: oops

hyprpm/src/core/PluginManager.cpp
vaxerski: just else

hyprpm/src/main.cpp
```diff
@@ -114,7 +114,10 @@ int                        main(int argc, char** argv, char** envp) {
 
             auto ret2 = g_pPluginManager->ensurePluginsLoadState();
 
-            if (ret2 != LOADSTATE_OK)
+            if (ret2 == LOADSTATE_HYPRLAND_UPDATED)
+                g_pPluginManager->notify(ICON_INFO, 0, 10000, "[hyprpm] Updated plugins, but Hyprland was updated. Please restart Hyprland.");
+
+            if (ret2 != LOADSTATE_OK || ret2 == LOADSTATE_HYPRLAND_UPDATED)
```
vaxerski, line 120: why the ||ret2 == LOADSTATE_HYPRLAND_UPDATED  it's redundant lol

hyprpm/src/main.cpp
```diff
@@ -130,7 +133,11 @@ int                        main(int argc, char** argv, char** envp) {
         }
 
         auto ret = g_pPluginManager->ensurePluginsLoadState();
-        if (ret != LOADSTATE_OK)
+
+        if (ret == LOADSTATE_HYPRLAND_UPDATED)
+            g_pPluginManager->notify(ICON_INFO, 0, 10000, "[hyprpm] Enabled plugin, but Hyprland was updated. Please restart Hyprland.");
+
+        if (ret != LOADSTATE_OK || ret == LOADSTATE_HYPRLAND_UPDATED)
```
vaxerski, line 140: same

vaxerski (commented):
rest lgtm

hyprpm/src/main.cpp
RoootTheFox: oop right

RoootTheFox:
fixed :)

vaxerski:
@RoootTheFox not just for me but the message pops up every hyprpm update

I don't think that's intended
image

RoootTheFox:
yeah, oops, that's definitely not intended. i'll look into fixing that once i get time

vaxerski:
thanks

PR #8637, opened 2024-12-03
snap: check which corner is being grabbed for monitor snapping

myQwil:

Floating windows were snapping to monitor edges even when resizing away from the monitor edge.

Nope

Ready for merging

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -487,19 +487,23 @@ static void performSnap(Vector2D& sourcePos, Vector2D& sourceSize, PHLWINDOW DRA
         SRange       monX = {MON->vecPosition.x + BORDERSIZE, MON->vecPosition.x + MON->vecSize.x - BORDERSIZE};
         SRange       monY = {MON->vecPosition.y + BORDERSIZE, MON->vecPosition.y + MON->vecSize.y - BORDERSIZE};
 
-        if (canSnap(sourceX.start, monX.start, GAPSIZE) || canSnap(sourceX.start, (monX.start += MON->vecReservedTopLeft.x + BORDERDIFF), GAPSIZE)) {
+        if (CORNER & (CORNER_TOPLEFT | CORNER_BOTTOMLEFT) &&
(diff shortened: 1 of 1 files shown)


soufhehe:
maybe not that important but I just found a case where it still occurs : 

https://github.com/user-attachments/assets/2fea264c-08be-4b0c-9196-841b6cda82f4

I don't know if I need to create a new issue

myQwil:
What are you expecting to happen in that scenario?

Also, what is your `monitor_gap` set to?

soufhehe:
> What are you expecting to happen in that scenario?

If I drag horizontally, I expect the window not to snap to the monitor. It works as expected if the cursor is at the bottom of the window (it doesn't snap), but not if it's at the top (as shown in the video).
 
> Also, what is your `monitor_gap` set to?

30, and the height of my bar is 18

> If you want, I can try adding an option that doesn't (or does) allow snapping beyond the reserved monitor space. I thought it would be a good idea to allow snapping to both the reserved space and the monitor edge.

Not necessary, it's a good idea to allow both. Thanks for bothering, but it's really just a tiny detail. If it's not easy to fix, there's no need to

myQwil:
You can drag horizontally, but it's always going to be a corner that you grab. You currently can't grab just one side. So in your example, you've grabbed one of the upper corners, which signals to the snap logic that if there's something above the window that it's close enough to snap to, then it should snap to it.

Your `monitor_gap` is big enough that it evidently reaches past the height of your status bar. It may seem abrupt or buggy but that's currently how it's supposed to work at this point.

Users should probably be able to have the option to prevent any snapping that causes the window to overlap the status bar like that, so I'm still going to look into adding that option.

Kommynct:
the monitor_gap should probably be affected by reserved areas by default, the edge of the monitor should probably effectively be the reserved area. Kinda seems silly to do it any other way.

myQwil:
> the monitor_gap should probably be affected by reserved areas by default

That would mess up `border_overlap`, but there's probably a better way of doing it than how it's currently being done. Maybe something like:
```
if reserved space > 0:
    bias towards reserved space
else:
    bias towards monitor edge
```
That approach might work well enough, and it wouldn't require another config option.

Issue #8482, opened 2024-11-15
Fullscreen apps cap to refresh rate when tearing is enabled


ethanh20009:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.45.0 built from branch main at commit 098e491a43e8b26f4382b48651a4131464bf6a2f  (protocols: mark primarySelection as not privileged).
Date: Fri Nov 15 00:47:34 2024
Tag: v0.45.0-23-g098e491a, commits: 5456
built against aquamarine 0.4.5


flags set:
debug


System Information:
System name: Linux
Node name: archlinux
Release: 6.12.0-rc7-1-cachyos-rc
Version: #1 SMP PREEMPT_DYNAMIC Mon, 11 Nov 2024 13:14:27 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD103 [GeForce RTX 4080] [10de:2704] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  565.57.01  Release Build  (cachyos-builder@)  


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/ethan/.config/hypr/hyprland.conf: Read Succeeded
#  _   _                  _                 _  
# | | | |_   _ _ __  _ __| | __ _ _ __   __| | 
# | |_| | | | | '_ \| '__| |/ _` | '_ \ / _` | 
# |  _  | |_| | |_) | |  | | (_| | | | | (_| | 
# |_| |_|\__, | .__/|_|  |_|\__,_|_| |_|\__,_| 
#        |___/|_|                              
#  
# ----------------------------------------------------- 

# ----------------------------------------------------- 
# Monitor
# ----------------------------------------------------- 
[...]

PR #8645, opened 2024-12-04
Don't use explicit sync fences if we can tear

ferrreo:
Fixes https://github.com/hyprwm/Hyprland/issues/8482

This fixes the issue where with explicit sync on no tearing is possible even when the requirements for it are fully met.

This is similar to how KWIN handles it.

I am not 100% sure this is ultimately the correct fix but it retains fences and allows tearing to happen when requested. It also seems to be what other compositors are doing too.

Also been running this patch on my distro with a good number of users with no issues reported for around 3 weeks.

Yea do it.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -1531,7 +1531,7 @@ bool CHyprRenderer::commitPendingAndDoExplicitSync(PHLMONITOR pMonitor) {
     if (inFD >= 0)
         pMonitor->output->state->setExplicitInFence(inFD);
     auto explicitOptions = getExplicitSyncSettings();
-    if (explicitOptions.explicitEnabled && explicitOptions.explicitKMSEnabled)
+    if (explicitOptions.explicitEnabled && explicitOptions.explicitKMSEnabled && !pMonitor->tearingState.canTear)
         pMonitor->output->state->enableExplicitOutFenceForNextCommit();
 
     if (pMonitor->ctmUpdated) {


src/render/Renderer.cpp
```diff
@@ -1531,7 +1531,7 @@ bool CHyprRenderer::commitPendingAndDoExplicitSync(PHLMONITOR pMonitor) {
     if (inFD >= 0)
         pMonitor->output->state->setExplicitInFence(inFD);
     auto explicitOptions = getExplicitSyncSettings();
-    if (explicitOptions.explicitEnabled && explicitOptions.explicitKMSEnabled)
+    if (explicitOptions.explicitEnabled && explicitOptions.explicitKMSEnabled && !pMonitor->tearingState.canTear)
```
vaxerski, line 1534: this should be activelyTearing not canTear.

vaxerski (commented):
can I get a ref on kwin for this one?

vaxerski:
yeah looks like it. we release when backend releases regardless

src/render/Renderer.cpp
ferrreo: I tried that first and it didn't work

ferrreo:
> can I get a ref on kwin for this one?

https://invent.kde.org/plasma/kwin/-/merge_requests/4800/diffs#f4ee7be3c0eb6f483546d5b8823ff33266747680_76_75

vaxerski:
1) that's IN_FENCE? And you are disabling OUT_FENCE?

UjinT34:
What does `explicitOutFence` really fences and when we should wait for it?

> This Sync File contains the CRTC fence that will be signaled when all framebuffers present on the Atomic Commit * request for that given CRTC are scanned out on the screen.

https://drmdb.emersion.fr/properties/3435973836/OUT_FENCE_PTR

Seems reasonable not to use it with tearing. 

kwin doesn't use `OUT_FENCE_PTR` at all and disables `IN_FENCE_FD` to avoid some bugs.

vaxerski:
we could disable IN_FENCE, but I don't think that's a good idea: explicit sync + scanout without IN_FENCE _will_ mess up.

UjinT34:
Is it possible to get DS with tearing?
https://github.com/hyprwm/Hyprland/blob/f9e4998a6d7713b19947b0db2c43ad88c5b71d80/src/render/Renderer.cpp#L1315

vaxerski:
hm, in our code, maybe not, but ideally yea it should be perfectly possible

src/render/Renderer.cpp
vaxerski: canTear is definitely wrong, but I don't thing we even use the out fence, can we drop it?

vaxerski:
[patch.txt](https://github.com/user-attachments/files/18028375/patch.txt)
can't we just do this? (path for main). This just drops out fd, I don't think we actually use it at all.

ferrreo:
Doesn't that patch undo https://github.com/hyprwm/Hyprland/commit/967fe76a60ec6b2928b495ed202f772beca05275?

vaxerski:
the cursor does look choppy when tearing, but oh well.

UjinT34:
With explicit sync enabled HL expects to receive the fence from AQ and complains
`Aquamarine did not return an explicit out fence`

I don't know how this fence should be used. Looks like it should be waited either before committing the next frame or before reusing the current framebuffer. Waiting for it immediately after the commit feels too early. Not using it at all is probably the same as waiting for it before reusing the fb because it is very unlikely to block at that point. Maybe there is no point in this fence with double buffering.

vaxerski:
yeah I don't think there is.

vaxerski:
We can go with my patch, do you want me to commit it to main or will you apply it here and we roll with this MR?

ferrreo:
> We can go with my patch, do you want me to commit it to main or will you apply it here and we roll with this MR?

I think Just stick yer patch in main and I'll close this

PR #8655, opened 2024-12-06
Add window rules for mouse and touchpad scroll factors

myQwil:

There are scenarios where I want the scroll factor to be different on a per-window/application basis. This change adds window rules for overriding the global scroll factors. There is one for the mouse scroll factor (`scrollmouse`) and one for the touchpad (`scrolltouchpad`)

Example:

`hyprland.conf`:
```ini
# Faster scrolling in Dolphin
windowrulev2 = scrollmouse 3, class:org.kde.dolphin

# Even faster scrolling in Sublime Text
windowrulev2 = scrollmouse 5, class:sublime_text
```

Not that I can think of.

Ready for merging.

### DIFF
--- a/src/config/ConfigManager.hpp
+++ b/src/config/ConfigManager.hpp
@@ -263,6 +263,10 @@ class CConfigManager {
         {"bordersize", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.borderSize; }},
     };
 
+    std::unordered_map<std::string, std::function<CWindowOverridableVar<float>*(PHLWINDOW)>> mfWindowProperties = {
+        {"scrollmouse", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.scrollMouse; }},
+        {"scrolltouchpad", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.scrollTouchpad; }}};
+
     bool m_bWantsMonitorReload = false;
     bool m_bForceReload        = false;
     bool m_bNoMonitorReload    = false;
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -3014,6 +3014,13 @@ SDispatchResult CKeybindManager::setProp(std::string args) {
                 search->second(PWINDOW)->unset(PRIORITY_SET_PROP);
             else if (const auto V = configStringToInt(VAL); V)
                 *(search->second(PWINDOW)) = CWindowOverridableVar((int)*V, PRIORITY_SET_PROP);
+        } else if (auto search = g_pConfigManager->mfWindowProperties.find(PROP); search != g_pConfigManager->mfWindowProperties.end()) {
+            if (VAL == "unset")
+                search->second(PWINDOW)->unset(PRIORITY_SET_PROP);
+            else {
+                const auto V               = std::stof(VAL);
+                *(search->second(PWINDOW)) = CWindowOverridableVar(V, PRIORITY_SET_PROP);
+            }
         } else
             return {.success = false, .error = "Prop not found"};
     } catch (std::exception& e) { return {.success = false, .error = std::format("Error parsing prop value: {}", std::string(e.what()))}; }
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -761,7 +761,8 @@ void CInputManager::onMouseWheel(IPointer::SAxisEvent e) {
     static auto PEMULATEDISCRETE      = CConfigValue<Hyprlang::INT>("input:emulate_discrete_scroll");
     static auto PFOLLOWMOUSE          = CConfigValue<Hyprlang::INT>("input:follow_mouse");
 
-    auto        factor = (*PTOUCHPADSCROLLFACTOR <= 0.f || e.source == WL_POINTER_AXIS_SOURCE_FINGER ? *PTOUCHPADSCROLLFACTOR : *PINPUTSCROLLFACTOR);
+    const bool  ISTOUCHPADSCROLL = *PTOUCHPADSCROLLFACTOR <= 0.f || e.source == WL_POINTER_AXIS_SOURCE_FINGER;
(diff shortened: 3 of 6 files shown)


src/managers/input/InputManager.cpp
```diff
@@ -776,6 +776,7 @@ void CInputManager::onMouseWheel(IPointer::SAxisEvent e) {
         const auto PWINDOW     = g_pCompositor->vectorToWindowUnified(MOUSECOORDS, RESERVED_EXTENTS | INPUT_EXTENTS | ALLOW_FLOATING);
 
         if (PWINDOW) {
+            factor *= PWINDOW->getScrollFactor();
```
fufexan, line 779: This multiplies the compositor-wide factor instead of overriding it entirely. This means that if I have the global factor set to `0.1`, I'd have to set the window scrollfactor to 10 to get a final factor of 1.

This might be correct, but I'm unsure.

myQwil: That's correct.

Originally, I was going to use the override approach, but there are two different scroll factors, one for mouse and one for touchpad, and if I were to add a window rule that overrides one, then I would feel obligated to add a second one that overrides the other.

This approach allows it to be either or.

fufexan: Sure, this makes sense. But doesn't that still pose the problem that the touchpad/mouse scroll factors are different, this also affects them differently?

A concrete example I have: global touchpad scroll factor is 0.1 due to GTK3 having a way too fast scroll. foot is not affected and could use a scroll factor of 1. This means if I set the scrollfactor windowrule to 10, it would also affect mouse scrolling, which I don't want it to.

myQwil: Alright, you've convinced me. I will use the override approach and add a window rule for touchpad scroll factor.

fufexan (approved):
Just tested on my touchpad, lgtm.

@vaxerski

myQwil:
Something's still off. The touchpad scroll factor is using `input:touchpad:scroll_factor` but then doesn't get overridden by `scrolltouchpad` window rules. It does, however, get overridden by `scrollmouse` window rules.

src/config/ConfigManager.hpp
```diff
@@ -263,6 +263,10 @@ class CConfigManager {
         {"bordersize", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.borderSize; }},
     };
 
+    std::unordered_map<std::string, std::function<CWindowOverridableVar<float>*(PHLWINDOW)>> mfWindowProperties = {
+        {"scrollmouse", [](const PHLWINDOW pWindow) { return &pWindow->m_sWindowData.scrollMouse; }},
+        {"scrolltouchpad", [](const PHLWINDOW pWindow) { return &pWindow->m_sWindowData.scrollTouchpad; }}};
```
vaxerski, line 268: const PHLWINDOW&

src/managers/KeybindManager.cpp
```diff
@@ -3014,6 +3014,12 @@ SDispatchResult CKeybindManager::setProp(std::string args) {
                 search->second(PWINDOW)->unset(PRIORITY_SET_PROP);
             else if (const auto V = configStringToInt(VAL); V)
                 *(search->second(PWINDOW)) = CWindowOverridableVar((int)*V, PRIORITY_SET_PROP);
+        } else if (auto search = g_pConfigManager->mfWindowProperties.find(PROP); search != g_pConfigManager->mfWindowProperties.end()) {
+            char* end_ptr;
+            if (VAL == "unset")
+                search->second(PWINDOW)->unset(PRIORITY_SET_PROP);
+            else if (const auto V = std::strtof(VAL.c_str(), &end_ptr); end_ptr != VAL.c_str() && *end_ptr == '\0')
```
vaxerski, line 3021: what the f is this? why not std::stof in a try catch???

src/managers/input/InputManager.cpp
```diff
@@ -776,6 +777,7 @@ void CInputManager::onMouseWheel(IPointer::SAxisEvent e) {
         const auto PWINDOW     = g_pCompositor->vectorToWindowUnified(MOUSECOORDS, RESERVED_EXTENTS | INPUT_EXTENTS | ALLOW_FLOATING);
 
         if (PWINDOW) {
+            factor = (ISTOUCHPADSCROLL) ? PWINDOW->getScrollTouchpad() : PWINDOW->getScrollMouse();
```
vaxerski, line 780: () redundant?

src/managers/input/InputManager.cpp
```diff
@@ -761,7 +761,8 @@ void CInputManager::onMouseWheel(IPointer::SAxisEvent e) {
     static auto PEMULATEDISCRETE      = CConfigValue<Hyprlang::INT>("input:emulate_discrete_scroll");
     static auto PFOLLOWMOUSE          = CConfigValue<Hyprlang::INT>("input:follow_mouse");
 
-    auto        factor = (*PTOUCHPADSCROLLFACTOR <= 0.f || e.source == WL_POINTER_AXIS_SOURCE_FINGER ? *PTOUCHPADSCROLLFACTOR : *PINPUTSCROLLFACTOR);
+    const bool  ISTOUCHPADSCROLL = (*PTOUCHPADSCROLLFACTOR <= 0.f || e.source == WL_POINTER_AXIS_SOURCE_FINGER);
+    auto        factor           = (ISTOUCHPADSCROLL ? *PTOUCHPADSCROLLFACTOR : *PINPUTSCROLLFACTOR);
```
vaxerski, line 765: () redundant

myQwil:
Okay, made the following changes:
- `ConfigManager.hpp`: made the `PHLWINDOW` parameters refs
- `InputManager.cpp`: removed redundant parentheses
- `KeybindManager.cpp`: switched to using `std::stof`
  - was going to add a try catch, but it's already in one for catching parse errors.

myQwil:
Moved the factor override past all the potential early return cases.

vaxerski (approved):
lgtm, thanks!

vaxerski:
just a wiki mr needed

PR #8664, opened 2024-12-06
core: Add clang-tidy

vaxerski:
This adds a nice clang-tidy config. Partially stolen from @outfoxxed.

Also fixes a lot of style issues with old MRd code. There are still a bunch of "errors" in some files. These might be up to either ignoring or fixing.

Probably should be done gradually, a lot of work all at once.


vaxerski:
some "can be made const" stuff isn't correct I believe, so we might nuke that.

Issue #7737, opened 2024-09-10
DND from menus to quicklaunch or desktop broken in lxqt-panel


stefonarch:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 0f594732b063a90d44df8c5d402d658f27471dfe  (props: bump version to 0.43.0).
Date: Sun Sep 8 16:48:21 2024
Tag: v0.43.0, commits: 5196

flags: (if any)


System Information:
System name: Linux
Node name: archlinux
Release: 6.6.49-1-lts
Version: #1 SMP PREEMPT_DYNAMIC Wed, 04 Sep 2024 13:18:51 +0000


GPU information: 
02:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Picasso/Raven 2 [Radeon Vega Series / Radeon Vega Mobile Series] [1002:15d8] (rev c4) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/stef/.config/lxqt/wayland/lxqt-hyprland.conf: Read Succeeded
# This is an example Hyprland config file.
# Refer to the wiki for more information.                                                                                         
# https://wiki.hyprland.org/Configuring/Configuring-Hyprland/                                                                     
                                                                                                                                  
# Please note not all available settings / options are set here.                                                                  
# For a full list, see the wiki

# You can split this configuration into multiple files
# Create your files separately and then link them to this file like this:
# source = ~/.config/hypr/myColors.conf


################
### MONITORS ###
################

# See https://wiki.hyprland.org/Configuring/Monitors/
monitor=,preferred,auto,1


###################
### MY PROGRAMS ###
###################

# See https://wiki.hyprland.org/Configuring/Keywords/

[...]

PR #8707, opened 2024-12-13
core/data: Use pointer focus for DnD operations

vaxerski:
This is mostly to fix stuff from #7737

Cases:
 - [x] `pcmanfm-qt --desktop` -> panel
 - [x] `pcmanfm-qt` -> panel
 - [x] menu -> panel

Unsure why the last one doesn't work ATM.

### DIFF
--- a/src/managers/SeatManager.cpp
+++ b/src/managers/SeatManager.cpp
@@ -14,8 +14,7 @@ CSeatManager::CSeatManager() {
     listeners.newSeatResource = PROTO::seat->events.newSeatResource.registerListener([this](std::any res) { onNewSeatResource(std::any_cast<SP<CWLSeatResource>>(res)); });
 }
 
-CSeatManager::SSeatResourceContainer::SSeatResourceContainer(SP<CWLSeatResource> res) {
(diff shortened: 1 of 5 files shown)


stefonarch:
Fine:
* DND in taskbar works now
* DND from desktop or filemanagers works
* DND from menu not
* Ctrl+DND in quicklauch (to reorder) not

vaxerski:
what does "ctrl+dnd" do that dnd doesnt? We don't send different actions based on modifiers pressed.

vaxerski:
menu->panel is fixed.

vaxerski:
I'd say the reorder might be an lxqt-panel bug, I don't see what is to be done on our side. Anyways, I am happy with the state of this MR.

stefonarch:
> what does "ctrl+dnd" do that dnd doesnt? We don't send different actions based on modifiers pressed.

Prevents accidental dragging.

vaxerski:
I meant more like in lxqt-panel. How do you detect it's ctrl+drag and not just drag.

PR #8714, opened 2024-12-14
`movefocus_cycle_groupfirst`: cycle within group on single monitor if no window found in the argument direction.

normaltaro:
+ Modified the logic of `movefocus_cycle_groupfirst` to keep cycling within the group if there is no window found in the argument direction for single monitor users. Example cases:
  + For single-monitor users
    + if there's only a single group in the workspace, focus will keep cycling to the other end instead of ending on one side. https://github.com/hyprwm/Hyprland/pull/8601#issuecomment-2543070992
    + If the there is no window to the left of the group and the user tries to go left, it will cycle back into the tail of the group instead of trying to look for the next window on the workspace.
  + For multi-monitor users, the logic remains same as the previous.

OBS keeps crashing after I installed hyprland-git and I haven't looked into it so I can't post a video, sorry.

I don't have multiple monitors so I haven't tested for it. I think the logic should work as intended.

Changes are limited to the original code. Ready.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1395,17 +1395,16 @@ SDispatchResult CKeybindManager::moveFocusTo(std::string args) {
         g_pCompositor->getWindowInDirection(PLASTWINDOW, arg);
 
     // Prioritize focus change within groups if the window is a part of it.
-    if (*PGROUPCYCLE) {
-        if (!PLASTWINDOW->m_sGroupData.pNextWindow.expired()) {
-            if (arg == 'l' && PLASTWINDOW != PLASTWINDOW->getGroupHead()) {
-                PLASTWINDOW->setGroupCurrent(PLASTWINDOW->getGroupPrevious());
-                return {};
-            }
+    if (*PGROUPCYCLE && (!PLASTWINDOW->m_sGroupData.pNextWindow.expired())) {
+        auto isTheOnlyGroupOnWs = (PWINDOWTOCHANGETO == nullptr);
+        if (arg == 'l' && (PLASTWINDOW != PLASTWINDOW->getGroupHead() || isTheOnlyGroupOnWs)) {
+            PLASTWINDOW->setGroupCurrent(PLASTWINDOW->getGroupPrevious());
+            return {};
+        }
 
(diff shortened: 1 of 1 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -1395,17 +1395,16 @@ SDispatchResult CKeybindManager::moveFocusTo(std::string args) {
         g_pCompositor->getWindowInDirection(PLASTWINDOW, arg);
 
     // Prioritize focus change within groups if the window is a part of it.
-    if (*PGROUPCYCLE) {
-        if (!PLASTWINDOW->m_sGroupData.pNextWindow.expired()) {
-            if (arg == 'l' && PLASTWINDOW != PLASTWINDOW->getGroupHead()) {
-                PLASTWINDOW->setGroupCurrent(PLASTWINDOW->getGroupPrevious());
-                return {};
-            }
+    if (*PGROUPCYCLE && (!PLASTWINDOW->m_sGroupData.pNextWindow.expired())) {
```
vaxerski, line 1398: .expired() and parentheses redundant

src/managers/KeybindManager.cpp
```diff
@@ -1395,17 +1395,16 @@ SDispatchResult CKeybindManager::moveFocusTo(std::string args) {
...
+        auto isTheOnlyGroupOnWs = (PWINDOWTOCHANGETO == nullptr) && (g_pCompositor->m_vMonitors.size() == 1);
```
vaxerski, line 1399: whats with all those parentheses. Also, `== nullptr` is the same as `!var`

normaltaro: sorry, im new to this. i didn't see it. i'll fix it.

normaltaro: done

src/managers/KeybindManager.cpp
vaxerski: you havent removed the expired call.

normaltaro: i see now, it is done

normaltaro: i didn't know what it actually did earlier. done

PR #8735, opened 2024-12-16
core: Optimize window/layer rule application and scanning

vaxerski:
This generally rewrites and optimizes parts of windowrule/layerrule handling.

Changes:
 - window and layer rules now do verification themselves and are classes
 - each rule scans itself on creation and assigns itself a numeric type
   - this means no more string comparisons later down the line
 - rules are now shared pointers, no more massive copying when vec resizes, it's passed, etc 
 
ref #8732

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -736,7 +736,7 @@ std::optional<std::string> CConfigManager::verifyConfigExists() {
     return {};
 }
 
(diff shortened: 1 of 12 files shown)


vaxerski:
in many cases we could precompute the result too (then ->apply() which has a cached result, e.g. with gradients and shit) but I think that's for another rainy day, plus at this point regex takes a lot longer...

PR #8762, opened 2024-12-18
Unbreak build on FreeBSD

jbeich:
Regressed by #8664. See #1708 for more details.


vaxerski (dismissed):

vaxerski:
missing clang-format

src/signal-safe.cpp
```diff
@@ -7,6 +7,11 @@
 #include <unistd.h>
 #include <cstring>
 
+// clang-format off
```
jbeich, line 10: I have no clue why clang-format mangles unrelated lines here but not in https://github.com/hyprwm/Hyprland/blob/80a4852266643c622ec68969b39c5abc3c610f30/src/config/ConfigManager.cpp#L36-L41

jbeich: Wouldn't that disable `NOLINTNEXTLINE`? I can't test clang-tidy atm.

vaxerski: if you put it _below_ extern C, no

vaxerski:
I fixed it for ya

vaxerski (approved):

PR #8767, opened 2024-12-19
Make static analysis more happy

ikalco:
* moves single file functions to static
* fixes a misuse of substr
* default initializer for borderGradient, kinda nitpicky but improvement nonetheless
* RASSERT uses std::print instead of printf

just opened my editor and fixed some of the red, might do more later

yes


vaxerski (approved):
naisu :+1:

PR #8775, opened 2024-12-19
Remove non-standard value from example config

StickMick:
While `enabled = yes, please :)` might work the same as true: the examples should show best practice rather than humor

This affects example config only.

Ready for merge.


vaxerski:
no, this is there so you know it's valid

Eclextic:
HUH? Commas are just inline comments or what lol?

vaxerski:
it just matches the beginning. You can put anything after `yes` `no` `true` etc

`yesamogus` is valid

PR #8788, opened 2024-12-20
add "disable_assets_check"

MightyPlaza:
adds "disable_assets_check"
https://github.com/hyprwm/hyprland-wiki/pull/910


vaxerski (commented):
why?

MightyPlaza:
Some users don't want to install the assets and don't want to have a patched version of Hyprland either
[Alpine](https://wiki.alpinelinux.org/wiki/Hyprland) for example has a different package for the wallpapers

vaxerski:
That's for wallpapers. What about crashed lock screens? I don't want people crying that the screen is black or whatever.

MightyPlaza:
> That's for wallpapers. What about crashed lock screens? I don't want people crying that the screen is black or whatever.
> 
> We can exclude the wallpaper check but other resources are important

they would still have to set the variable to true in the config, but in the end the user is probably the one that should choose what they want
another possibility could be a fallback lockscreen crash screen being added just with basic text rendering

vaxerski:
> the user is probably the one that should choose what they want

the user is stupid more times than not

PR #8860, opened 2024-12-26
update ForeignTopLevel on null window focus

user111111111111111111111111111111111:
removes check for valid window so ForeignTopLevel gets updated on null window focus
don't see any crashes or side effects, but haven't checked if everything is properly handled in `onNewFocus`

### DIFF
--- a/src/protocols/ForeignToplevelWlr.cpp
+++ b/src/protocols/ForeignToplevelWlr.cpp
@@ -347,9 +347,6 @@ CForeignToplevelWlrProtocol::CForeignToplevelWlrProtocol(const wl_interface* ifa
     static auto P3 = g_pHookSystem->hookDynamic("activeWindow", [this](void* self, SCallbackInfo& info, std::any data) {
         const auto PWINDOW = std::any_cast<PHLWINDOW>(data);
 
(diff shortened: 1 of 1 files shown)


src/protocols/ForeignToplevelWlr.cpp
```diff
@@ -347,9 +347,6 @@ CForeignToplevelWlrProtocol::CForeignToplevelWlrProtocol(const wl_interface* ifa
     static auto P3 = g_pHookSystem->hookDynamic("activeWindow", [this](void* self, SCallbackInfo& info, std::any data) {
         const auto PWINDOW = std::any_cast<PHLWINDOW>(data);
 
-        if (!windowValidForForeign(PWINDOW))
```
vaxerski, line 350: this will break if we have a window that is invalid for foreign. Should probably be `!valid() && PWINDOW`

MightyPlaza: done

user111111111111111111111111111111111: done

vaxerski (approved):
thanks!

Issue #8855, opened 2024-12-26
Not able to reduce the size of fullscreen groupped-floating-windows with dispatcher fullscreen (if changed the active grouped window)


khushal-solves:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

Yes, No

### System Info and Version

<details> <summary>System/Version info</summary>

> Using Default Config to ensure this bug has nothing to do with my configuration.

```sh
# [works@archlinux ~]$ hyprctl systeminfo -c
Hyprland 0.46.0 built from branch main at commit 71dc9f6128b8d2e382b2a574d2d5f15e2d907f3a  (pluginAPI: add register dispatcher v2).
Date: Fri Dec 20 01:53:32 2024
Tag: v0.46.0-18-g71dc9f61, commits: 5573
built against:
 aquamarine 0.5.0
 hyprlang 0.6.0
 hyprutils 0.2.6
 hyprcursor 0.1.10
 hyprgraphics 0.1.1


flags set:
debug


System Information:
System name: Linux
Node name: archlinux
Release: 6.6.65-1-lts
Version: #1 SMP PREEMPT_DYNAMIC Wed, 11 Dec 2024 15:35:54 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Iris Plus Graphics G1 (Ice Lake) [8086:8a56] (rev 07) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/works/.config/hypr/hyprland.conf: Read Succeeded

# #######################################################################################
# AUTOGENERATED HYPRLAND CONFIG.
# PLEASE USE THE CONFIG PROVIDED IN THE GIT REPO /examples/hyprland.conf AND EDIT IT,
# OR EDIT THIS ONE ACCORDING TO THE WIKI INSTRUCTIONS.
# #######################################################################################

[...]

PR #8865, opened 2024-12-27
Updated window position/size for grouped fullscreen-windows in setGroupCurrent fn

khushal-solves:
# Change

In `setGroupCurrent` function in `src/desktop/window.cpp` file,
Fullscreen windows were not able to get effective values for `m_vRealSize` and `m_vRealPosition`
With this PR, both size and position are updated after changing fullscreenstate in the same function.

# Effect
This effects fullscreen group of floating windows. 

Issues fixed:
- If a group of floating windows is fullscreen with static window rule `group` set new added windows in group were not able to toggle fullscreen
- If a group of floating windows is fullscreen then after switching to next window in group with `changegroupactive` any window in group was not able to toggle fullscreen.

**With this PR, both of these issues are now fixed and Issue #8855 can be closed.**

I hope i've not misunderstood something.
In case you are not interested, feel also free to close this PR :)

Cheers!

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1026,13 +1026,13 @@ void CWindow::setGroupCurrent(PHLWINDOW pWindow) {
     const auto WORKSPACE  = PCURRENT->m_pWorkspace;
     const auto MODE       = PCURRENT->m_sFullscreenState.internal;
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
lgtm

PR #8868, opened 2024-12-27
core: move parts of the animation system to hyprutils

PaideiaDilemma:
Requires https://github.com/hyprwm/hyprutils/pull/27

Moves core components of the animation system to hyprutils. (In order to use it in hyprlock as well)
Introduces smart pointers for animated variables and animation properties.

Interface changes:
- AnimatedVaraibles are now created/initialized with `g_pAnimationManager->createAnimation`. That function takes a value to always have AnimatedVaraibles explicitly initialized to a value.
- All AnimatedVariables are now SPs, resulting in a required operator* for assignments (e.g `*m_vRealSize = Vector2D(42,42)` and in a change from `av.setValue` to `av->setValue`
- Animation config is now updated with `av->setConfig`.
- `CAnimatedVaraible` now has member functions `enabled`, `getBezierName` and `getStyle` to get their config properties more easily.

Needs testing.
~~The damage handling was refactored a bit. Previously it damaged before and after a variable update in `CAnimationManager::tick`. I don't think that is needed and from what i tested it works the same as before.
Possibly needs to be reverted to how it worked before, but the damage handling needs to be moved into functions, because we need to upcast first to get the context of animated variables.~~

Need some testing and review.


vaxerski:
I've pushed hyprutils 0.3.1, cmakelists can be updated

vaxerski:
with HU 0.3.2 https://github.com/hyprwm/hyprutils/releases/tag/v0.3.2 I assume this should also move to the config tree before mergin?

PaideiaDilemma:
Yes sir, i think that would be best

src/config/ConfigManager.cpp
```diff
@@ -1641,8 +1615,8 @@ void CConfigManager::ensureVRR(PHLMONITOR pMonitor) {
     }
 }
 
-SAnimationPropertyConfig* CConfigManager::getAnimationPropertyConfig(const std::string& name) {
-    return &animationConfig[name];
+SP<SAnimationPropertyConfig> CConfigManager::getAnimationPropertyConfig(const std::string& name) {
+    return m_AnimationTree.getConfig(name);
 }
```
PaideiaDilemma, line 1620: Do we want to keep the existing interface, or make `m_AnimationTree` public and access it directly?

src/config/ConfigManager.cpp
```diff
@@ -2071,48 +2034,41 @@ std::optional<std::string> CConfigManager::handleAnimation(const std::string& co
...
-    if (PANIM == animationConfig.end())
+    if (!m_AnimationTree.nodeExists(ANIMNAME))
         return "no such animation";
 
-    PANIM->second.overridden = true;
-    PANIM->second.pValues    = &PANIM->second;
-
     // This helper casts strings like "1", "true", "off", "yes"... to int.
-    int64_t enabledInt = configStringToInt(ARGS[1]).value_or(0) == 1;
+    int64_t enabledInt = configStringToInt(ARGS[1]).value_or(0);
```
PaideiaDilemma, line 2041: note: idk why `== 1` was there. I removed it cause i thought it was intended to have the plain int here.

PaideiaDilemma: found why it was added. keeping it.

src/config/ConfigManager.cpp
vaxerski: keep it

PaideiaDilemma:
I am going to rebase and cleanup the commit history of this PR so that all updates regarding the usage of animated variables (so adding `*` and changing `.` to `->`) are in a separate commit.

Hopefully that will make it a bit easier to review.

vaxerski:
thanks :) tag me when ready

PaideiaDilemma:
@vaxerski ready.

vaxerski (approved):
thanks!

PR #8877, opened 2024-12-28
protocols: add hyprland_surface_v1 implementation

outfoxxed:
Implements hyprland_surface_v1, added here: https://github.com/hyprwm/hyprland-protocols/pull/11.
Reasoning is in the linked protocol MR.

Tested against a client implementation, works as expected.

Ready to merge, however the nix lockfile will need to be updated after the protocol is merged.

### DIFF
--- a/src/protocols/HyprlandSurface.cpp
+++ b/src/protocols/HyprlandSurface.cpp
@@ -1,0 +1,111 @@
+#include "HyprlandSurface.hpp"
+#include "../desktop/WLSurface.hpp"
+#include "../render/Renderer.hpp"
+#include "core/Compositor.hpp"
+#include "hyprland-surface-v1.hpp"
+
+CHyprlandSurface::CHyprlandSurface(SP<CHyprlandSurfaceV1> resource, SP<CWLSurfaceResource> surface) : m_pSurface(surface) {
+    setResource(std::move(resource));
+}
+
+bool CHyprlandSurface::good() const {
+    return m_pResource->resource();
+}
+
+void CHyprlandSurface::setResource(SP<CHyprlandSurfaceV1> resource) {
+    m_pResource = std::move(resource);
+
+    if (!m_pResource->resource())
+        return;
+
+    m_pResource->setDestroy([this](CHyprlandSurfaceV1* resource) { destroy(); });
+    m_pResource->setOnDestroy([this](CHyprlandSurfaceV1* resource) { destroy(); });
+
+    m_pResource->setSetOpacity([this](CHyprlandSurfaceV1* resource, uint32_t opacity) {
+        if (!m_pSurface) {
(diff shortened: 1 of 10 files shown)


outfoxxed:
*fixed meson

outfoxxed:
Updated based on proto changes

outfoxxed:
Made destroy atomic

src/protocols/HyprlandSurface.cpp
```diff
@@ -0,0 +1,113 @@
...
+            if (box.has_value())
+                g_pHyprRenderer->damageBox(&*box);
+
+            if (!m_pResource)
+                PROTO::hyprlandSurface->destroySurface(this);
+        }
+    });
+
+    listeners.surfaceDestroyed = m_pSurface->events.destroy.registerListener([this](std::any data) {
+        m_pSurface.reset();
```
vaxerski, line 55: useless

src/render/OpenGL.cpp
```diff
@@ -2045,7 +2045,8 @@ bool CHyprOpenGLImpl::shouldUseNewBlurOptimizations(PHLLS pLayer, PHLWINDOW pWin
     return false;
 }
 
-void CHyprOpenGLImpl::renderTextureWithBlur(SP<CTexture> tex, CBox* pBox, float a, SP<CWLSurfaceResource> pSurface, int round, bool blockBlurOptimization, float blurA) {
+void CHyprOpenGLImpl::renderTextureWithBlur(SP<CTexture> tex, CBox* pBox, float a, SP<CWLSurfaceResource> pSurface, int round, bool blockBlurOptimization, float blurA,
```
vaxerski, line 2048: why this change? why not just multiply it beforehand?

src/protocols/HyprlandSurface.cpp
outfoxxed: Since its a WP?

src/render/OpenGL.cpp
outfoxxed: Because overallA multiplies both the texture and blur texture A, regardless of PBLURIGNOREOPACITY .

outfoxxed: without that we might multiply by overallA^2 if !PBLURIGNOREOPACITY

outfoxxed:
rebased and fixed submodules+nix

src/protocols/HyprlandSurface.cpp
vaxerski: surface calls destroy when its about to be freed so the WP will be dead after this is called

src/render/OpenGL.cpp
vaxerski: makes sens

src/protocols/HyprlandSurface.cpp
outfoxxed: fixed

vaxerski (approved):

Issue #4731, opened 2024-02-17
Hyprpm: make `reload -f` unload & load the plugin. 


zjeffer:
### Description

Hyprpm reload doesn't seem to actually unload and reload the plugin? It only loads plugins that are enabled but not currently loaded. I would expect that the PLUGIN_EXIT and PLUGIN_INIT functions of each plugin would be called.

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #8883, opened 2024-12-29
make `hyprpm reload -f` unload & load all enabled plugins

zjeffer:
Implements https://github.com/hyprwm/Hyprland/issues/4731

I've only tested it with one plugin enabled, will test a bit more first and then mark it as ready to review.

EDIT: tested with multiple plugins, only noticed one problem: the Hyprtrails plugin will crash Hyprland when unloading them:(https://github.com/hyprwm/hyprland-plugins/issues/264). Not a problem with this PR, so ready to merge.


vaxerski (approved):
as node-js tends to say
`Using --force, I hope you know what you are doing.`

so the hyprtrails issue can be ignored for now :)

Issue #8893, opened 2024-12-30
Problem: github issues are not really searchable anymore


rszyma:
### Description

Currently it's impossible to search anything in issues anymore since people embed their full configs in dropdown instead of attaching. For example, I'm trying to search for an issue now and it doesn't get filtered at all [https://github.com/hyprwm/Hyprland/issues?q="workspaces"+"fade"+"animation"+is%3Aissue](https://github.com/hyprwm/Hyprland/issues?q=%22workspaces%22+%22fade%22+%22animation%22+is%3Aissue). No way I'm going through 300 open issues to check duplicates

Ghosthree3:
Well, even if the rules were changed and people started attaching their configs instead of embedding them, the damage is already far too gone to fix the problem.

rszyma:
> I am not sure how effective this is But I personally found logical operators very effective.
> 
> for example something like following in `filter`:
> 
> > is:issue, in:title "workspace animation"  OR "animation fade"
> 
> Gives `only 10 Open Issues` with 18 Closed Issues .. [Here](https://github.com/hyprwm/Hyprland/issues?q=sort%3Aupdated-desc+is%3Aissue%2C+in%3Atitle+%22workspace+animation%22+OR+%22animation+fade%22)

yea seems like the only viable option now. But yea, it will not search description which often rewords title and adds more related keywords

rszyma:
> Well, even if the rules were changed and people started attaching their configs instead of embedding them, the damage is already far too gone to fix the problem.

Is it not? Maybe it would be possible to write a script that iterates over existing bug reports and replaces the config in op with link.

2 problems though:
- github seems to doesn't provide api for file uploads for attachments (see https://github.com/orgs/community/discussions/46951)
- not sure if the edited issues will get reindexed to not pop up when searching keywords that were removed in edit (probably yes?)

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #6160, opened 2024-05-19
#6098: Add Config Flag to `hyprctl systeminfo`

The-Briel-Deal:
Closes #6098

#### Describe your PR, what does it fix/add?
Frequently I will look through issues on Github and be unable to repro the issue because there is no config attached to the issue. My hope is that by adding a --config/-c flag to systeminfo and asking for config files will significantly improve this problem and hopefully save developer time from time wasted trying to repro unclear issues.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Should be fine. Let me know if I should correct wording in the template.

#### Is it ready for merging, or does it need work?
Should be ready.

PR #8894, opened 2024-12-30
bug issue template improvements

rszyma:

Attempts to remedy #8893 by bringing back attaching as primary way of sending text files (logs, crashes, config).

Just so everbody reading this know, we used to have bug template that incentivized attaching instead of pasting, but it was not really effective. Anyway it was changed in https://github.com/hyprwm/Hyprland/pull/6160 so that user always pastes full config in plain text, and the issue search clutter began because of that. 

Compared to before #6160, hopefully now it's clearer that before that every text file must be attached. 

Also added more variants to "is it regression" dropdown, now I feel like response from that field will be more useful.

rendered issue looks like this:
- https://github.com/rszyma/Hyprland/issues/4

template looks like this:
- https://github.com/rszyma/Hyprland/issues/new?assignees=&labels=bug&projects=&template=bug.yml

nope

potentially needs more discussion, but I went a couple of iterations myself and now I feel like it's good now as is

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -79,9 +79,11 @@ body:
 
         When including text files (such as logs or config), please **always ATTACH** them, and not paste them directly.
 
-        <u>This is important</u> because pasting text directly clutters the issue with unnecessary keywords, making github issue search useless.
+        This is important to avoid clutter, spam, and make the issues more readable.
         Thanks for your understanding.
 
+  # The main reason to disallow pasting directly or in a dropdown, is to not clutter
+  # the issue with unnecessary keywords, making the github issue search useless.
   - type: checkboxes
     attributes:
       label: Attach not paste
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -5,31 +5,34 @@ body:
   - type: checkboxes
     attributes:
       label: Already reported ? *
-      description: Before opening a new bug report, please take a moment to search through the current open and closed issues to check if it already exists.
+      description: Before opening a new bug report, please take a moment to search through the current open issues. If same bug is already reported, don't open new issue - instead go upvote/comment on existing one.
       options:
-      - label: I have searched the existing open and closed issues.
-        required: true
+        - label: I have searched the existing open and closed issues.
+          required: true
 
   - type: dropdown
     id: type
     attributes:
-      label: Regression?
+      label: Is this bug a regression?
       description: |
         Regression means that something used to work but no longer does.
         **BEFORE CONTINUING**, please check if this bug is a regression or not, and if it is, we need you to bisect with the help of the wiki: https://wiki.hyprland.org/Crashes-and-Bugs/#bisecting-an-issue
       multiple: true
       options:
-        - "Yes"
-        - "No"
+        - "Definitely a regression - something broke after update (requires bisect)"
+        - "Probably not a regression / I don't remember it happening before"
+        - "Not a regression - it's bug regarding new feature"
+        - "Not a regression - it's an old bug"
+        - "I don't know, I started using Hyprland only recently"
     validations:
       required: true
 
   - type: textarea
     id: ver
     attributes:
-      label: System Info and Version
+      label: System Info and Hyprland Version
       description: |
-        Paste the output of `hyprctl systeminfo -c` here. If you can't
+        Paste the output of `hyprctl systeminfo` here. If you can't
         launch Hyprland, paste the output of `Hyprland --systeminfo`.
         If `Hyprland --systeminfo` errors out (added in 0.44.0), find
         and paste the Hyprland version manually.
@@ -37,8 +40,8 @@ body:
         <summary>System/Version info</summary>
 
 
-        ```sh
-        
+        ```
+
         <Paste the output of the command here>
 
         ```
@@ -61,15 +64,55 @@ body:
     attributes:
       label: How to reproduce
       description: "How can someone else reproduce the issue?"
+      placeholder: |
+        1. ...
+        2. ...
+        3. ...
+
     validations:
       required: true
+
+  - type: markdown
+    attributes:
+      value: |
+        ## Additional info section
+
+        In the section below you will be asked to upload some files.
+
+        When including text files (such as logs or config), please **always ATTACH** them, and not paste them directly.
+
+        <u>This is important</u> because pasting text directly clutters the issue with unnecessary keywords, making github issue search useless.
+        Thanks for your understanding.
+
+  - type: checkboxes
+    attributes:
+      label: Attach not paste
+      options:
+        - label: I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam
+          required: true
+
+  - type: markdown
+    attributes:
+      value: >-
+        Please be sure to upload the following files below:
+
+          - Logs can be found in $XDG_RUNTIME_DIR/hypr
+          - Crash reports are stored in ~/.cache/hyprland or $XDG_CACHE_HOME/hyprland
+          - Hyprland config files - `hyprctl systeminfo -c > /tmp/hyprland_config_dump.txt` use this command to dump full configuration to a single file.
+
+  - type: checkboxes
+    attributes:
+      label: Checklist of files to include below
+      options:
+        - label: Hyprland config - `hyprctl systeminfo -c` (always include)
+          required: true
+        - label: Crash report (always include in case of crash)
+        - label: Video (always include in case of a visual bug)
+        - label: Logs (might contain useful info such as errors)
 
   - type: textarea
     id: logs
     attributes:
-      label: Crash reports, logs, images, videos
+      label: Additional info & File uploads
       description: |
-        Anything that can help. Please always ATTACH and not paste them.
-        Logs can be found in $XDG_RUNTIME_DIR/hypr
-        Crash reports are stored in ~/.cache/hyprland or $XDG_CACHE_HOME/hyprland
-
+        Tip: You can attach files by clicking this area to highlight it and then dragging files in.
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -105,7 +105,6 @@ body:
       label: Checklist of files to include below
       options:
         - label: Hyprland config - `hyprctl systeminfo -c` (always include)
-          required: true
         - label: Crash report (always include in case of crash)
         - label: Video (always include in case of a visual bug)
         - label: Logs (might contain useful info such as errors)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,40 +5,43 @@ body:
...
-      - label: I have searched the existing open and closed issues.
-        required: true
+        - label: I have searched the existing open and closed issues.
+          required: true
 
   - type: dropdown
     id: type
     attributes:
-      label: Regression?
+      label: Is this bug a regression?
```
littleblack111, line 16:
```suggestion
      label: Is this bug due to regression?
```

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,40 +5,43 @@ body:
   - type: checkboxes
     attributes:
       label: Already reported ? *
-      description: Before opening a new bug report, please take a moment to search through the current open and closed issues to check if it already exists.
+      description: Before opening a new bug report, please take a moment to search through the current open issues. If same bug is already reported, don't open new issue - instead go upvote/comment on existing one.
```
littleblack111, line 8:
```suggestion
      description: Before opening a new bug report, please take a moment to search through the issues. If same bug is already reported, don't open new issue - instead go upvote/comment on existing one.
```

.github/ISSUE_TEMPLATE/bug.yml
rszyma: Or maybe "is this issue a regression?" 
Regression may not be necessarily a  bug, e.g. perf degradation, but still someone might want to report it

littleblack111: sure, go ahead

rszyma: reverted to the original to not distract from the meat of this pr

.github/ISSUE_TEMPLATE/bug.yml
rszyma: I would leave as is. I don't want to get into details why, to not go too much off the main pr topic, it's also not as important. In short, explicit > implicit, also "open issues" suggests better to a complete noob that doesn't even know that "issues" means issue listing not "issues" as "problems" (this is going a bit too deep already, sigh...)

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -61,15 +64,55 @@ body:
...
+    attributes:
+      label: Attach not paste
+      options:
+        - label: I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam
+          required: true
+
+  - type: markdown
+    attributes:
+      value: >-
+        Please be sure to upload the following files below:
```
vaxerski, line 97: _if they are relevant to the issue_. Not all bugs are crashes

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -61,15 +64,55 @@ body:
...
+
+          - Logs can be found in $XDG_RUNTIME_DIR/hypr
```
vaxerski, line 99: you can add a hint to (sort by date to grab the latest)

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -61,15 +64,55 @@ body:
...
+  - type: markdown
+    attributes:
+      value: |
+        ## Additional info section
+
+        In the section below you will be asked to upload some files.
+
+        When including text files (such as logs or config), please **always ATTACH** them, and not paste them directly.
+
+        <u>This is important</u> because pasting text directly clutters the issue with unnecessary keywords, making github issue search useless.
```
vaxerski, line 84: too long. "This is important to avoid clutter, spam, and make the issues more readable".

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -19,26 +19,29 @@ body:
...
   - type: textarea
     id: ver
     attributes:
-      label: System Info and Version
+      label: System Info and Hyprland Version
       description: |
-        Paste the output of `hyprctl systeminfo -c` here. If you can't
+        Paste the output of `hyprctl systeminfo` here. If you can't
         launch Hyprland, paste the output of `Hyprland --systeminfo`.
         If `Hyprland --systeminfo` errors out (added in 0.44.0), find
```
vaxerski, line 37: we can remove this line by now. 0.44 is mega old

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,10 +5,10 @@ body:
   - type: checkboxes
     attributes:
       label: Already reported ? *
-      description: Before opening a new bug report, please take a moment to search through the current open and closed issues to check if it already exists.
+      description: Before opening a new bug report, please take a moment to search through the current open issues. If same bug is already reported, don't open new issue - instead go upvote/comment on existing one.
```
vaxerski, line 8: if _the_ same bug ...

comment on _an_ existing one

.github/ISSUE_TEMPLATE/bug.yml
rszyma: fbb4c58

.github/ISSUE_TEMPLATE/bug.yml
rszyma: added

.github/ISSUE_TEMPLATE/bug.yml
rszyma: yes chef

.github/ISSUE_TEMPLATE/bug.yml
rszyma: fbb4c58

.github/ISSUE_TEMPLATE/bug.yml
rszyma: I really wanted to include this line explaining why exactly it's a problem, and not just say "it's better trust me bro". Otherwise someone will question it "I'm too lazy to put it in file first, how is copy-pasting a spam anyway? If it's too long I'll put it in dropdown" you get the idea

vaxerski: If you explicitly state that failing to do so will get your issue removed they will do it I'd say

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -19,26 +19,27 @@ body:
...
-        and paste the Hyprland version manually.
       value: "<details>
         <summary>System/Version info</summary>
 
 
-        ```sh
-        
+        ```
+
         <Paste the output of the command here>
```
vaxerski, line 43: can we clarify this? Some people paste over the entire box. `Paste the output of the command here, without removing any formatting around this.` or something

.github/ISSUE_TEMPLATE/bug.yml
rszyma: yea sure. Changed to this d3b52c1

.github/ISSUE_TEMPLATE/bug.yml
rszyma: how about this?  50c5b9434ff16ecd6049d928448897da52def5f7

image

rszyma: what the heck, why are there's duplicate lines issue when using pre tags but not when using triple backticks ...
image

vaxerski: no clue lmao

rszyma: using double newlines inside dropdown in pre tag is broken on github smh. I'll just do what you proposed then.

5ad89b5

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -19,27 +19,28 @@ body:
         **BEFORE CONTINUING**, please check if this bug is a regression or not, and if it is, we need you to bisect with the help of the wiki: https://wiki.hyprland.org/Crashes-and-Bugs/#bisecting-an-issue
       multiple: true
       options:
-        - "Yes"
-        - "No"
+        - "Definitely a regression - something broke after update (requires bisect)"
```
vaxerski, line 22: why can we select multiple here btw? It should be single

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -61,15 +62,57 @@ body:
...
+          - Logs can be found in $XDG_RUNTIME_DIR/hypr (sort by date to grab the latest)
+          - Crash reports are stored in ~/.cache/hyprland or $XDG_CACHE_HOME/hyprland
+          - Hyprland config files - `hyprctl systeminfo -c > /tmp/hyprland_config_dump.txt` use this command to dump full configuration to a single file.
+
+  - type: checkboxes
+    attributes:
+      label: Checklist of files to include below
+      options:
+        - label: Hyprland config - `hyprctl systeminfo -c` (always include)
+          required: true
```
vaxerski, line 108: I dont think this should be required.

vaxerski (commented):
rest ok

.github/ISSUE_TEMPLATE/bug.yml
rszyma: it was like that before too. I guess the reason is that single-selection form has to have one of values selected on by default - and then it doesn't matter if that the field is marked as required, and people gonna miss it and leave default value

.github/ISSUE_TEMPLATE/bug.yml
rszyma: yea sure 7cb93c1

.github/ISSUE_TEMPLATE/bug.yml
rszyma: it was discussed before - see https://github.com/hyprwm/Hyprland/pull/6520#discussion_r1644167564

vaxerski: ok, sure

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -61,15 +62,56 @@ body:
...
+
+          - Logs can be found in $XDG_RUNTIME_DIR/hypr (sort by date to grab the latest)
+          - Crash reports are stored in ~/.cache/hyprland or $XDG_CACHE_HOME/hyprland
+          - Hyprland config files - `hyprctl systeminfo -c > /tmp/hyprland_config_dump.txt` use this command to dump full configuration to a single file.
+
+  - type: checkboxes
+    attributes:
+      label: Checklist of files to include below
+      options:
+        - label: Hyprland config - `hyprctl systeminfo -c` (always include)
```
vaxerski, line 107: you still left the (always include)

rszyma: Yea, i assumed you only want relax the requirement not remove it conpletely.

If we communicate here that the config is optional we would effectively revert https://github.com/hyprwm/Hyprland/pull/6160, you sure we want that?

vaxerski: actually, it's fine how it is.

vaxerski (approved):
thanks!

PR #8913, opened 2024-12-31
Share hyprctl flags with hyprpaper

theaino:
To allow [json output for hyprpaper](https://github.com/hyprwm/hyprpaper/issues/186), sharing the flags with hyprpaper is needed. This change just appends the flags **at the end of the request**, with a slash. This allows it to still be compatible with hyprpaper which doesn't handle the flags.

nope

Maybe needs some discussion, but since it's a really simple fix I think it's ready.

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -255,8 +255,10 @@ int requestHyprpaper(std::string arg) {
         return 3;
     }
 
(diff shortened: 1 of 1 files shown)


theaino:
Oh - this doesn't work when having args... I don't think it's possible without breaking compatibility with the current hyprpaper version.

Issue #8920, opened 2025-01-01
Make `v2` version of `focusedmon` event.


jbirnick:
### Description

For some of the events that the second socket emits, it seems like they at some point got a second version `...v2` which includes IDs and addresses instead of just titles/names.

However, it seems like not all events received a `v2` version. For example, `focusedmon` only provides the workspace name, not the workspace ID. Would it be possible to add this?

PR #8921, opened 2025-01-01
socket2: add focusedmonv2 event

davc0n:
Closes https://github.com/hyprwm/Hyprland/issues/8920, not sure if there's any other info which makes sense to add (neither about that one tbh).

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2591,7 +2591,7 @@ void CCompositor::setActiveMonitor(PHLMONITOR pMonitor) {
     const auto WORKSPACE_NAME = PWORKSPACE ? PWORKSPACE->m_szName : "?";
 
     g_pEventManager->postEvent(SHyprIPCEvent{"focusedmon", pMonitor->szName + "," + WORKSPACE_NAME});
-    g_pEventManager->postEvent(SHyprIPCEvent{"focusedmonv2", pMonitor->szName + "," + WORKSPACE_ID + "," + WORKSPACE_NAME});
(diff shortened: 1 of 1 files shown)


src/Compositor.cpp
```diff
@@ -2587,7 +2587,12 @@ void CCompositor::setActiveMonitor(PHLMONITOR pMonitor) {
 
     const auto PWORKSPACE = pMonitor->activeWorkspace;
 
-    g_pEventManager->postEvent(SHyprIPCEvent{"focusedmon", pMonitor->szName + "," + (PWORKSPACE ? PWORKSPACE->m_szName : "?")});
+    const auto WORKSPACE_ID   = PWORKSPACE ? std::to_string(PWORKSPACE->m_iID) : std::to_string(WORKSPACE_INVALID);
+    const auto WORKSPACE_NAME = PWORKSPACE ? PWORKSPACE->m_szName : "?";
+
+    g_pEventManager->postEvent(SHyprIPCEvent{"focusedmon", pMonitor->szName + "," + WORKSPACE_NAME});
+    g_pEventManager->postEvent(SHyprIPCEvent{"focusedmonv2", pMonitor->szName + "," + WORKSPACE_ID + "," + WORKSPACE_NAME});
```
vaxerski, line 2594: sending name here makes no sense:
 - firstly, we sent it before
 - secondly, can be queried with hyprctl

davc0n: Fixed (kept it because it's like that in other cases such as `createworkspacev2`)

vaxerski (approved):
thanks! wiki mr needed

PR #8926, opened 2025-01-01
master: make loop around optional when cycling

lonyelon:

Implement the option to not loop around when cycling windows in the master layout (i.e. going from master to the last slave or vice-versa), replicating this layout's behavior in DWL.

No default behavior was modified, and the option `noLoopAround` option was added to `cyclenext`, `cycleprev`, `swapnext` and `swapprev`. This ensures no current setups are broken.

Example usage:
- Default, can cycle around as always:
  ```sh
  bind = $mainMod, up,    layoutmsg, cycleprev
  bind = $mainMod, down,  layoutmsg, cyclenext
  ```
- Same as default but with explicit option set:
  ```sh
  bind = $mainMod, up,    layoutmsg, cycleprev loop
  bind = $mainMod, down,  layoutmsg, cyclenext loop
  ```
- No looping around (as in DWM):
  ```sh
  bind = $mainMod, up,    layoutmsg, cycleprev noloop
  bind = $mainMod, down,  layoutmsg, cyclenext noloop
  ```

I think it is, I tested it and it works, but I am new in this codebase so some important test may be missing.

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -978,7 +978,7 @@ void CHyprMasterLayout::alterSplitRatio(PHLWINDOW pWindow, float ratio, bool exa
     recalculateMonitor(pWindow->monitorID());
 }
 
-PHLWINDOW CHyprMasterLayout::getNextWindow(PHLWINDOW pWindow, bool next) {
+PHLWINDOW CHyprMasterLayout::getNextWindow(PHLWINDOW pWindow, bool next, bool loop) {
     if (!isWindowTiled(pWindow))
         return nullptr;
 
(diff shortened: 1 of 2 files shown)


src/layout/MasterLayout.cpp
```diff
@@ -1110,15 +1117,17 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
         if (!PWINDOW)
             return 0;
 
-        const auto PNEXTWINDOW = getNextWindow(PWINDOW, true);
+        const bool noLoopAround = vars.size() >= 2 && vars[1] == "noLoopAround";
```
vaxerski, line 1120: noLoopAround is excessively verbose I'd say, why not just `noloop`?

also const vars in CAPS pls

lonyelon: Yeah, way better. It was like that at the beginning but I felt it was more descriptive like that.

Would it be better to add a new layout command named `cyclenextnoloop` rather than making this a param for `cyclenext`?

Thanks for the feedback!

lonyelon:
Hello vaxry, 

Would it be better to add a new layout command named `cyclenextnoloop` rather than making this a param for `cyclenext`? It would be like `movetoworkspace` and `movetoworkspacesilent`.

vaxerski:
no, param is better imo

lonyelon:
Ok so we can leave it like this.

vaxerski (dismissed):
yep this lgtm, just a wiki mr needed

lonyelon:
Added in https://github.com/hyprwm/hyprland-wiki/pull/926!

lonyelon:
clang style issues fixed.

vaxerski (approved):
thanks!

PR #8940, opened 2025-01-03
Add option to  show slaves on left in master layout center orientation

Zetta1Reid0:
Adds the option to use right master orientation when slave_count_for_center_master > 2 ,
also makes windows tile to left side first then to the right side when option is set.

I think it should e finished. I tested it, everything works as expected.

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1627,6 +1627,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .data        = SConfigOptionDescription::SRangeData{2, 0, 10}, //##TODO RANGE?
     },
     SConfigOptionDescription{
+        .value       = "master:center_master_slaves_on_right",
+        .description = "set if the slaves should appear on right of master when slave_count_for_center_master > 2",
+        .type        = CONFIG_OPTION_BOOL,
+        .data        = SConfigOptionDescription::SBoolData{true},
+    },
+    SConfigOptionDescription{
         .value       = "master:center_ignores_reserved",
         .description = "centers the master window on monitor ignoring reserved areas",
         .type        = CONFIG_OPTION_BOOL,
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -478,6 +478,7 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("master:mfact", {0.55f});
     m_pConfig->addConfigValue("master:new_status", {"slave"});
     m_pConfig->addConfigValue("master:slave_count_for_center_master", Hyprlang::INT{2});
+    m_pConfig->addConfigValue("master:center_master_slaves_on_right", Hyprlang::INT{1});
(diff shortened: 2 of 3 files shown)


src/config/ConfigDescriptions.hpp
```diff
@@ -1626,6 +1626,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_INT,
         .data        = SConfigOptionDescription::SRangeData{2, 0, 10}, //##TODO RANGE?
     },
+    SConfigOptionDescription{
+        .value       = "master:center_master_slaves_on_right",
+        .description = "set if the slaves should appear on rigth of master when slave_count_for_center_master > 2",
```
vaxerski, line 1631: the right

src/config/ConfigManager.cpp
```diff
@@ -478,6 +478,7 @@ CConfigManager::CConfigManager() {
     m_pConfig->addConfigValue("master:mfact", {0.55f});
     m_pConfig->addConfigValue("master:new_status", {"slave"});
     m_pConfig->addConfigValue("master:slave_count_for_center_master", Hyprlang::INT{2});
+    m_pConfig->addConfigValue("master:center_master_slaves_on_right", Hyprlang::INT{0});
```
vaxerski, line 481: this should default to 1 no?

Zetta1Reid0: I wanted , to keep default behavior before the option , so people wouldn't really have to change their config if they want to keep the old behavior.

src/config/ConfigDescriptions.hpp
Zetta1Reid0: Fixed

khushal-solves:
Why only left or right.. why not top and bottom?

Zetta1Reid0:
> Why only left or right.. why not top and bottom?

Couse, it doesn't only set the layout used if slave_counter_for_center_master > 2. It also makes windows spawn from left to right instead of right to left when masters are centered, Can't do that in 1 property, since we can't decide where should slaves spawn if we are using top or bottom.

src/config/ConfigManager.cpp
vaxerski: then why is it 1 in config descriptions

Zetta1Reid0: Becouse I am obviously blind.

src/layout/MasterLayout.cpp
```diff
@@ -515,15 +519,23 @@ void CHyprMasterLayout::calculateWorkspace(PHLWORKSPACE pWorkspace) {
         float       nextY       = 0;
         float       nextYL      = 0;
         float       nextYR      = 0;
-        bool        onRight     = true;
+        bool        onRight     = *CMSLAVESONRIGHT;
+        int         slavesLeftL;
```
vaxerski, line 523: bugprone: uninitializes primitives.

vaxerski (commented):
rest ok, wiki mr needed

vaxerski (approved):
thanks!

Issue #3804, opened 2023-11-08
Squircles


nabato:
Does it support squircles? Please, add a definitive answer to the docs if possible.

vaxerski:
No it doesn't.

> definitive answer to the docs if possible.

if it's not explicitly listed as supported, it's not. That's how docs work. Should I also mention it can't make a sandwich?

nabato:
@vaxerski, `awesome-wm` or `i3` than. Thank you nonetheless.

woutdp:
Just leaving this here for anyone who wants to implement Squircles, this code is the shader code to make squircles, it replaces the regular corner code. You'd still need to keep the previous code for the regular corners as squircles should be an option and not the default. `squirclePolynomial` might be interesting to keep as a config setting, something between 2 and 8 gives good results.

https://github.com/hyprwm/Hyprland/blob/main/src/render/shaders/Textures.hpp#L5-L33
```
inline static constexpr auto ROUNDED_SHADER_FUNC = [](const std::string colorVarName) -> std::string {
    return R"#(

    // branchless baby!
    highp vec2 pixCoord = vec2(gl_FragCoord);
    pixCoord -= topLeft + fullSize * 0.5;
    pixCoord *= vec2(lessThan(pixCoord, vec2(0.0))) * -2.0 + 1.0;
    pixCoord -= fullSize * 0.5 - radius;
    pixCoord += vec2(1.0, 1.0) / fullSize; // center the pix dont make it top-left

    if (pixCoord.x + pixCoord.y > radius) {
      float squirclePolynomial = 4.0;
      float squircle = pow(abs(pixCoord.x / radius), squirclePolynomial) + pow(abs(pixCoord.y / radius), squirclePolynomial) - 1.0;
      float edgeTransition = 1.0 - smoothstep(-0.02, 0.02, squircle);
      )#" + colorVarName + R"#( = )#" + colorVarName + R"#( * edgeTransition;
    }
    )#";
};
```

Won't pursue this further but migh be helpful to someone.

This is only the code for the texture, you'd still need to make the border work.

Issue #3048, opened 2023-08-21
More corner options besides just round (diagonal, different amount, etc)


REALERvolker1:
### Description

Hello. What if your hyprland window corners were diagonal?

Think about it for a minute. This could be a subtle enhancement for hyprland that makes it look so much better!

I propose adding the following keywords in the `decoration` subcategory:
- corners
  - specifies the number of "corners" of each window corner
- radius
  - specifies the radius oof the corner circle that this is all calculated from
- rounding
  - floating-point number that controls how "round" the corner is

I made a quick visual in kolourpaint of what I am imagining

![image](https://github.com/hyprwm/Hyprland/assets/73304952/bc41f0a1-0bde-4f02-8360-cb6435e426ac)

What say ye?

vaxerski:
how high are you?

REALERvolker1:
idk, I think it is a good idea and would make hyprland rices way better + more unique

ferrreo:
What happens if I set 40 corners

REALERvolker1:
you would probably only see a difference if your radius property is set to a high integer as well. It would not be advisable to do so for performance reasons (calculating all those angles)

PR #8943, opened 2025-01-03
feat: add support for supercircular (e.g. squircles) window corner rounding

polluxcodes:

This PR adds support for using generalized supercircles for rounding the corners of windows.

Supercircles are curves defined by equations of the form $x^n + y^n = R^n$, which includes regular circles ($n=2$) and squircles ($n\approx4$). Many consider these to be a more aesthetically pleasing way to round the corners of rectangles, when compared to circles, since they don't introduce curvature discontinuities for $n\geq3$. (See the image below.)

It adds a single configuration parameter `decoration:rounding_power`, which is used as the value of $n$ and defaults to 2 to conserve existing behavior.

Here is a comparison showing the difference between a rounding power of 2 (left, current behavior) and a rounding power of 4 (right):
hyprland-corners-comparison

This modification preserves existing behavior, although it may have a slight performance impact. I am open to benchmarking the code, but would need some tips on how to do so.

It should be ready for merging.

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -824,7 +824,8 @@ void CWindow::updateDynamicRules() {
 // it is assumed that the point is within the real window box (m_vRealPosition, m_vRealSize)
 // otherwise behaviour is undefined
 bool CWindow::isInCurvedCorner(double x, double y) {
-    const int ROUNDING = rounding();
+    const int ROUNDING      = rounding();
+    const int ROUNDINGPOWER = roundingPower();
     if (getRealBorderSize() >= ROUNDING)
         return false;
 
@@ -835,16 +836,16 @@ bool CWindow::isInCurvedCorner(double x, double y) {
     double y1 = m_vRealPosition.value().y + m_vRealSize.value().y - ROUNDING;
 
     if (x < x0 && y < y0) {
-        return Vector2D{x0, y0}.distance(Vector2D{x, y}) > (double)ROUNDING;
+        return std::pow(x0 - x, ROUNDINGPOWER) + std::pow(y0 - y, ROUNDINGPOWER) > std::pow((double)ROUNDING, ROUNDINGPOWER);
     }
     if (x > x1 && y < y0) {
-        return Vector2D{x1, y0}.distance(Vector2D{x, y}) > (double)ROUNDING;
+        return std::pow(x - x1, ROUNDINGPOWER) + std::pow(y0 - y, ROUNDINGPOWER) > std::pow((double)ROUNDING, ROUNDINGPOWER);
     }
     if (x < x0 && y > y1) {
-        return Vector2D{x0, y1}.distance(Vector2D{x, y}) > (double)ROUNDING;
+        return std::pow(x0 - x, ROUNDINGPOWER) + std::pow(y - y1, ROUNDINGPOWER) > std::pow((double)ROUNDING, ROUNDINGPOWER);
     }
     if (x > x1 && y > y1) {
-        return Vector2D{x1, y1}.distance(Vector2D{x, y}) > (double)ROUNDING;
+        return std::pow(x - x1, ROUNDINGPOWER) + std::pow(y - y1, ROUNDINGPOWER) > std::pow((double)ROUNDING, ROUNDINGPOWER);
     }
 
     return false;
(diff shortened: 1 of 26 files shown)


vaxerski:
wasn't there an open issue to add this that we could close with this one?

Anyways I'll review tomorrow it's way too late for me to review gpu code

polluxcodes:
> wasn't there an open issue to add this that we could close with this one?

Are you referring to #3804 or #3048?

vaxerski:
I guess yea

leiserfg:
This might result interesting https://iquilezles.org/articles/roundedboxes/

polluxcodes:
> This might result interesting https://iquilezles.org/articles/roundedboxes/

Hmmm, an SDF might be a pretty good way to do this. Although I don't think any of the solutions offered there work better than supercircles/superellipses here, since they either a) don't remove the discontinuity fully, or b) require a numerical solver to calculate.

Now that I think about it, I don't think an exact SDF can possibly satisfy both of the above conditions, since any curve that removes the discontinuity must be at least third order, so finding it's SDF must necessarily involve solving a quintic equation, which requires a numerical solver because of the [Abel–Ruffini theorem](https://en.wikipedia.org/wiki/Abel%E2%80%93Ruffini_theorem).

Currently, I'm sort of unintentionally using an inexact SDF given by $SDF=\left(x^n+y^n\right)^\frac{1}{n} - D$. It's not perfect, but it's close enough for its flaws to be hard to see under normal use. If you look really closely at the example on the right, the border gets slightly thicker in the corner than along the sides.

polluxcodes:
My bad if the force push wasn't done properly. I'm new to using git collaboratively.

vaxerski:
whoa calm down there nerd patrol what is all that math in my git repo

src/desktop/Window.hpp
```diff
@@ -414,6 +415,7 @@ class CWindow {
     Vector2D               middle();
     bool                   opaque();
     float                  rounding();
+    float                  roundingPower();
```
vaxerski, line 418: why should the rounding power be a member? If you can't override it with a window rule, there is no point?

You can either add a windowrule (like rounding) for it, and then this makes sense, or just access the config value directly

vaxerski (commented):
at some point I should start passing structs to render functions the amount of args is getting ridiculous, lol.

Nice work :)

src/desktop/Window.hpp
polluxcodes: Working on adding a window rule. Just need to test it.

vaxerski: alrighty

leiserfg:
@polluxcodes I was talking (sorry for just throwing a link) about the Cubic sdf, which keeps the edges parallel but still removes the discontinuity in the derivative.
 
image
Image borrowed from the article where the all the corners but the bottom right, are using cubic.

src/desktop/Window.hpp
polluxcodes: Added in [30f752a](https://github.com/hyprwm/Hyprland/pull/8943/commits/30f752a460089c0d980121dccde85d62a92f151e).

polluxcodes:
@leiserfg I think there is a small misunderstanding, I'm not making the whole window a single superellipse, I'm only replacing the corners with quarter-superellipses. This way the windows still have parallel edges, while also not having any discontinuities.

The only benefits the cubic rounding could have is a nicer shape (I compared them on a graphing calculator, they're nearly identical if $n\approx3.8$), or better performance, and I don't think the performance would be better since the cubic SDF uses a numerical solver.

src/desktop/Window.cpp
```diff
@@ -1164,13 +1165,23 @@ bool CWindow::opaque() {
...
 
     return m_sWindowData.noRounding.valueOrDefault() ? 0 : rounding;
 }
 
+float CWindow::roundingPower() {
+    static auto PROUNDINGPOWER = CConfigValue<Hyprlang::FLOAT>("decoration:rounding_power");
+
+    float       roundingPower = m_sWindowData.roundingPower.valueOr(*PROUNDINGPOWER);
+
+    return roundingPower;
```
vaxerski, line 1182: why the local instead of a simple ternary?

vaxerski (commented):
rest lgtm, needs a wiki mr

src/desktop/Window.cpp
polluxcodes: Forgot ternaries existed tbh

polluxcodes: Wait, did you mean why not just directly return? I'll fix that.

polluxcodes:
Squashed commits and removed fixup commit messages.

src/desktop/Window.cpp
vaxerski: ye, sorry, 1am comments

vaxerski (approved):
lgtm tanks

Eclextic:
More features is more features I guess, but this seems a little pedantic no?

polluxcodes:
@Eclextic I agree the difference is fairly subtle, but I wouldn't say it's pedantic. I think the rounded corners on the right example look significantly better than the left, even at a glance, and though I can't necessarily speak for anyone else, given that this feature has been requested before (#3804) and the positive reactions from people to this PR, it seems at least a few others agree with me.

Also, I'm not sure I would consider the rounding in the example to be all that high; I have it set to a pretty similar value in my actual hyprland config, though maybe it is just me.

polluxcodes:
> Why don't we then just make that the default? Just remove the other rounding calculation honestly... Sure it does look "better" but to such a negligible degree, that I would assume, it can just become the way how rounding looks anyways. No confit option no nothing.
> 
> Honestly I would advocate for it to just become the default rounding style.

I think it is important to conserve behavior, especially when it comes to design choices, which are very personal. If we changed the default corner style, and provided no option to change it back, then I garuntee people will create issues asking to add an option to do so.

There are plenty of situations where circular corners do look better, especially when placed in context of a desktop with other UI elements. For example, many people have rices with GTK bars that float slightly away from the edges of their screen and have rounded corners. CSS doesn't support squircle rounding easily. (You could maybe do it with `clip-path` and an SVG but it would be a fair bit more work.) If we provided no way to keep the existing corner style, they would have to put up with different kinds of rounded corners on different applications. I can confidently say that that would drive me insane.

vaxerski:
Most of the lines in this MR are small style changes due to clang-format realigning members. 100 lines of code to add this won't hurt anyone, we already have 75k.

PukPukov:
Just curious, how does the squircle can look better? It feels so off.

Issue #8834, opened 2024-12-23
First slave window has weird tall and narrow size when switched to floating.


Lamampis:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.45.2 built from branch  at commit 12f9a0d0b93f691d4d9923716557154d74777b0a  ([gha] Nix: update inputs).
Date: Tue Nov 19 21:47:18 2024
Tag: v0.45.2, commits: 5451
built against aquamarine 0.5.0


flags set:
debug


System Information:
System name: Linux
Node name: thinkpad
Release: 6.12.4-gentoo-krunking
Version: #1 SMP PREEMPT_DYNAMIC Thu Dec 12 20:45:11 EET 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation 3rd Gen Core processor Graphics Controller [8086:0166] (rev 09) (prog-if 00 [VGA controller])


os-release: NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
VERSION_ID="2.17"


plugins:

======Config-Start======


Config File: /home/lampis/.config/hypr/hyprland.conf: Read Succeeded
# See https://wiki.hyprland.org/Configuring/Monitors/
monitor=,preferred,auto,auto
#monitor=,addreserved,0,6,38,38
monitor=,addreserved,0,8,8,8

# Startup
exec-once=/usr/libexec/polkit-gnome-authentication-agent-1 &
exec-once=hyprpaper
exec-once=udiskie &
exec-once = dunst &
exec-once = cat ~/.cache/wal/sequences &
exec-once = dbus-launch
exec-once=dbus-update-activation-environment DISPLAY
exec-once=sleep 0.7 && waybar
exec-once = wl-paste --type text --watch cliphist store #Stores only text data
exec-once = wl-paste --type image --watch cliphist store #Stores only image data

# Set programs that you use
$terminal = foot
$fileManager = nemo 
$menu = tofi-drun 
$volup = pulseaudio-ctl up
$voldown = pulseaudio-ctl down
$mute = pulseaudio-ctl mute
[...]

PR #8951, opened 2025-01-04
Fixed weird size issue of 1st foot slave when set floating

khushal-solves:
First slave window of foot terminal always had desired width = 14
Hyprland allowed all desired sizes with width & height greater than 5 so foot window saved 14 as width in last saved floating size.
So, later when this window is set floating it would have a width of 14 which looks very weird.

This only happens with 1st slave window of foot terminal, [here is a video demo](https://youtu.be/9XPqwA03uAw)
I modified the code to allow all desired sizes with width & height more than 14.
Now, desired width 14 for first slave window of foot terminal becomes invalid and Hyprland resizes it appropriately. [which looks normal as shown in this video](https://youtu.be/mjvs033lJDo)

It relates to issue #8834 but i was only able to replicate the very tall and narrow window (when 1st slave is set floating) for foot terminal only

- This only happens with foot terminal and that too with 1st slave window which is very weird.
- Should this be treated as bug of foot terminal?
Although minimum allowed size should be not related to how to different applications behave, width/height 14 is still very small. So it should not be allowed. 

It requires review.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -12,7 +12,7 @@ void IHyprLayout::onWindowCreated(PHLWINDOW pWindow, eDirection direction) {
     CBox desiredGeometry = {};
     g_pXWaylandManager->getGeometryForWindow(pWindow, &desiredGeometry);
 
-    if (desiredGeometry.width <= 5 || desiredGeometry.height <= 5) {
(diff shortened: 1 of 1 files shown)


vaxerski:
why does foot request width of 14?

khushal-solves:
Thanks, I will try to find the root cause.

PR #8952, opened 2025-01-04
Bug fix for hyprctl --batch doesn't work with exec rules

### LINKED ISSUE #1820: hyprctl --batch doesn't work with exec rules
I'm trying to do something like this:
```bash
hyprctl --batch "\
dispatch exec [noanim;float;workspace name:$SELECTED silent] kitty;\
etc..."
```

The rules for exec in square brackets don't work because the semicolon is used as delimiter for the batch command as well.

Jakenv:
Fixes #1820 by ignoring ';' inside brackets [ ]

For some reason the only app that was problematic when testing using commands like this -> 
`hyprctl --batch 'dispatch exec [noanim;float;workspace 7 silent] appname; dispatch exec [noanim;float;workspace 6 silent] appname; dispatch exec appname;'` 
was Wezterm, it would spawn on whatever workspace, other than that everything else worked without issues

Yeah, I was testing it with different apps and different combinations of commands and as mentioned above only Wezterm was problematic

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1150,34 +1150,28 @@ static std::string cursorPosRequest(eHyprCtlOutputFormat format, std::string req
 }
 
 static std::string dispatchBatch(eHyprCtlOutputFormat format, std::string request) {
-    // split by ;
+    // split by ; ignores ; inside [] and adds ; on last command
 
-    request             = request.substr(9);
-    std::string curitem = "";
-    std::string reply   = "";
-
-    auto        nextItem = [&]() {
-        auto idx = request.find_first_of(';');
-
-        if (idx != std::string::npos) {
-            curitem = request.substr(0, idx);
-            request = request.substr(idx + 1);
-        } else {
-            curitem = request;
-            request = "";
-        }
-
-        curitem = trim(curitem);
-    };
-
-    nextItem();
-
+    request                     = request.substr(9);
+    std::string       curitem   = "";
+    std::string       reply     = "";
     const std::string DELIMITER = "\n\n\n";
+    int               bracket   = 0;
 
-    while (curitem != "" || request != "") {
-        reply += g_pHyprCtl->getReply(curitem) + DELIMITER;
-
-        nextItem();
+    for (size_t i = 0; i <= request.size(); ++i) {
+        char ch = (i < request.size()) ? request[i] : ';';
+        if (ch == '[')
+            ++bracket;
+        else if (ch == ']')
+            --bracket;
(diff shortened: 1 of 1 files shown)


src/debug/HyprCtl.cpp
```diff
@@ -1150,34 +1150,28 @@ static std::string cursorPosRequest(eHyprCtlOutputFormat format, std::string req
...
+        else if (ch == ']')
+            --bracket;
+        else if (ch == ';' && bracket == 0) {
+            curitem = trim(curitem);
+            if (!curitem.empty())
+                reply += g_pHyprCtl->getReply(curitem).append(DELIMITER);
+            curitem.clear();
+            continue;
+        }
+        curitem += ch;
```
vaxerski, line 1174: this will do a lot of potential reallocs, wouldn't it be better to keep an idx and substring when encountering a ';'?

Jakenv: Yeah, valid. I updated it and removed the constant clearing in favor of substrings. Dunno why I didn't think of it in the first place

src/debug/HyprCtl.cpp
```diff
@@ -1150,34 +1150,26 @@ static std::string cursorPosRequest(eHyprCtlOutputFormat format, std::string req
...
+
+    for (size_t i = 0; i <= request.size(); ++i) {
+        char ch = (i < request.size()) ? request[i] : ';';
+        if (ch == '[')
+            ++bracket;
+        else if (ch == ']')
+            --bracket;
+        else if (ch == ';' && bracket == 0) {
+            if (idx < i)
+                reply += g_pHyprCtl->getReply(trim(request.substr(idx, i - idx))).append(DELIMITER);
```
vaxerski, line 1169: what about the last argument?
`cmda ; cmdb ; cmdc`
won't cmdc not be parsed? it doesn't have a `;` at the end that we can find.

Jakenv: That is why on line 1162 there is ternary to add virtual ';' for last command, without it last command would get ignored. From my testing everything from commands that end in no ';' or even with multiple ';' in case of miss input will just work.

So stuff like `cmda ; cmdb ; cmdc` or `cmda ; cmdb ; cmdc;;;;;;;` will return "ok" 3 times, ofc if the commands itself do have proper syntax

Jakenv: Let me know if this approach in general is ok for you for this bug fix. This is my first pull request ever so I kinda don't know the dynamics of this lol

vaxerski: right, I am blind.

vaxerski (approved):

PR #8953, opened 2025-01-04
exec(-onec) with rules and execr(-once)

littleblack111:
currently both exec-once and exec does not support window rule like their dispatchers, i've added that feature and also added execr and execr-once

for the change in firstExecRequests, not sure if i shall also do that to finalExecRequests, i think it's better to keep it as it is, since exec-shutdown is executed when shutting down thus it's not necessary for window rule?

ready

### DIFF
--- a/src/config/ConfigManager.hpp
+++ b/src/config/ConfigManager.hpp
@@ -285,11 +285,11 @@ class CConfigManager {
     std::vector<SP<CLayerRule>>                      m_vLayerRules;
     std::vector<std::string>                         m_dBlurLSNamespaces;
 
-    bool                                                      firstExecDispatched     = false;
-    bool                                                      m_bManualCrashInitiated = false;
+    bool                                             firstExecDispatched     = false;
+    bool                                             m_bManualCrashInitiated = false;
 
-    std::vector<SFirstExecRequest>                            firstExecRequests; // bool is for if with rules
-    std::vector<std::string>                                  finalExecRequests;
+    std::vector<SFirstExecRequest>                   firstExecRequests; // bool is for if with rules
+    std::vector<std::string>                         finalExecRequests;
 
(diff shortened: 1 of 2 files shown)


src/config/ConfigManager.hpp
```diff
@@ -293,7 +295,7 @@ class CConfigManager {
 
     bool                                                      firstExecDispatched     = false;
     bool                                                      m_bManualCrashInitiated = false;
-    std::vector<std::string>                                  firstExecRequests;
+    std::vector<std::pair<std::string, bool>>                 firstExecRequests; // bool is for if with rules
```
vaxerski, line 298: at this point I'd just make a small struct so that the comment is not necessary

littleblack111: so.. do you want me to make a struct? inside the vector?

src/config/ConfigManager.hpp
```diff
@@ -146,6 +146,11 @@ struct SConfigOptionDescription {
     std::variant<SBoolData, SRangeData, SFloatData, SStringData, SColorData, SChoiceData, SGradientData, SVectorData> data;
 };
 
+struct SfirstExecRequest {
+    std::string exec;
+    bool        withRules;
```
vaxerski, line 151: uninitialized primitive please initialize it

src/config/ConfigManager.hpp
```diff
@@ -146,6 +146,11 @@ struct SConfigOptionDescription {
     std::variant<SBoolData, SRangeData, SFloatData, SStringData, SColorData, SChoiceData, SGradientData, SVectorData> data;
 };
 
+struct SfirstExecRequest {
```
vaxerski, line 149: capitalization fail: `SFirstExecRequest`

vaxerski (commented):
rest ok

vaxerski:
conflict

vaxerski (approved):
thanks!

vaxerski:
wiki mr pls

PR #8961, opened 2025-01-05
fix: forcerelease when locked && captured

littleblack111:


littleblack111:
omg gh cli sucks, i did --base 3l0w:feat/input-capture-impl

PR #8962, opened 2025-01-05
windows destruction dispatchers

littleblack111:
atm kill(active,window) don't actually kill it but close it instead, this fixes that
also adds real kill(active,window) which actually kills it

should probably be in breaking changes

ready

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -12,6 +12,7 @@
 #include "eventLoop/EventLoopManager.hpp"
 #include "debug/Log.hpp"
 #include "helpers/varlist/VarList.hpp"
+#include "../helpers/signal/Signal.hpp"
 
 #include <optional>
 #include <iterator>
@@ -63,8 +64,10 @@ CKeybindManager::CKeybindManager() {
 
     m_mDispatchers["exec"]                           = spawn;
     m_mDispatchers["execr"]                          = spawnRaw;
+    m_mDispatchers["closeactive"]                    = closeActive;
     m_mDispatchers["killactive"]                     = killActive;
-    m_mDispatchers["closewindow"]                    = kill;
+    m_mDispatchers["closewindow"]                    = closeWindow;
+    m_mDispatchers["killwindow"]                     = killWindow;
     m_mDispatchers["togglefloating"]                 = toggleActiveFloating;
     m_mDispatchers["setfloating"]                    = setActiveFloating;
     m_mDispatchers["settiled"]                       = setActiveTiled;
@@ -978,12 +981,18 @@ uint64_t CKeybindManager::spawnRawProc(std::string args, PHLWORKSPACE pInitialWo
 }
 
 SDispatchResult CKeybindManager::killActive(std::string args) {
+	kill(g_pCompositor->m_pLastWindow.lock()->getPID(), SIGKILL);
+
+    return {};
+}
+
+SDispatchResult CKeybindManager::closeActive(std::string args) {
     g_pCompositor->closeWindow(g_pCompositor->m_pLastWindow.lock());
 
     return {};
 }
 
-SDispatchResult CKeybindManager::kill(std::string args) {
+SDispatchResult CKeybindManager::closeWindow(std::string args) {
     const auto PWINDOW = g_pCompositor->getWindowByRegex(args);
 
     if (!PWINDOW) {
@@ -992,6 +1001,19 @@ SDispatchResult CKeybindManager::kill(std::string args) {
     }
 
     g_pCompositor->closeWindow(PWINDOW);
+
+    return {};
+}
+
+SDispatchResult CKeybindManager::killWindow(std::string args) {
+    const auto PWINDOW = g_pCompositor->getWindowByRegex(args);
+
+    if (!PWINDOW) {
+        Debug::log(ERR, "kill: no window found");
+        return {.success = false, .error = "kill: no window found"};
+    }
+
+	kill(PWINDOW->getPID(), SIGKILL);
 
     return {};
 }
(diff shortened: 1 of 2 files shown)


izmyname:
Isn't the whole point of `killactive` is to gracefully close the client, akin to 'x' button on stacking compositors?

littleblack111:
> Isn't the whole point of `killactive` is to gracefully close the client, akin to 'x' button on stacking compositors?
> 
> Also, what about daemon processes, like `easyeffects`? For example, won't 'killing' `easyeffects` client window also kill `---gapplication-service` process?

imo kill doesn't sound that graceful, but close is. you can still use the killactive the same, just now renamed to closeactive

vaxerski (commented):
silent config break, I don't like those.

Is this really needed?

izmyname:
It's not about how it sounds, but what it does.



> you can still use the killactive the same, just now renamed to closeactive

Honestly, I don't see the point. Besides, this is a breaking change and I can predict many 'help when I close the window - my daemon processes die too' issues.

littleblack111:
Just an feature I used to have but now have to replace via scripting, and I do use it pretty often(especially to applications that are poorly coded and mainly focus on windows, usually sometimes just hang for 1-2min after giving killactive. I guess kill didn’t rly make sense in the first place if it’s not really killing it but closing the window gracefully. Also, since we have that dialog that shows after every update, should be fine right

izmyname:
> I guess the naming of “kill”active didn’t rly make sense in the first place

It's a common name for this function - you can find this in sway, i3 and other compositors/xorg wms. Keeping things standardized is a good thing.




> Also, since we have that dialog that shows after every update, should be fine right

That dialog pops up after a version bump - not git.



> edit: if you have any better ways of doing this without a silent config break. I’d be happy to do

Not changing `kill`, `killactive` behavior.



> don’t see why would it kill the daemon? It sends a kill to the pid

Okay, but what about the apps that aren't meant to be killed upon closing a window? The ones with a systray, for example?

Overall, I still don't see the point. To me - it looks like you want to break the most common must-have dispatcher, everyone uses, just because you, personally, prefer killing clients, instead of closing them.

izmyname:
> don’t see why would it kill the daemon? It sends a kill to the pid, not a killall to the binary(name)

Also, I've just checked - easyeffects flatpak (and non-flatpak, I guess) have just one pid for both - `gapplication-service` and an opened client window and, I assume, the same is true for all GTK apps with services. So,  my point till stands.

TAforever:
Using a real kill by default instead of close is a bad idea, but having a separate dispatcher for kill would be nice. bspwm has this option 
```bash 
bspc -c #close
bspc -k #kill
 ```
Currently in hyprland I use 
 `bind = $mainMod SHIFT, Q, exec, kill -9 $(hyprctl activewindow | grep pid | tail -1 | awk '{print $2}')`

Also it would close this issue  https://github.com/hyprwm/Hyprland/issues/8212

**Comment by izmyname** [comment]
You could leave `kill(active)`, as it already is, a name the new dispatcher something like `pid_kill(active)` or `kill(active)_proc`. Or `pkill(active)`. A desired function without breaking things.

**Comment by leiserfg** [comment]
I agree with @izmyname, repurposing `kill` will break old configs (even when IMHO the name was confusing from the beginning). 
Maybe an option to consider could be  `signal` allowing to pass https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html

**Comment by littleblack111** [comment]
Already, will use forcekillactive then?

**Comment by littleblack111** [comment]
> I agree with @izmyname, repurposing `kill` will break old configs (even when IMHO the name was confusing from the beginning). Maybe an option to consider could be `signal` allowing to pass https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html
> 
> So @TAforever 's script : `bind = $mainMod SHIFT, Q, exec, kill -9 $(hyprctl activewindow | grep pid | tail -1 | awk '{print $2}')`
> 
> could be replaced by: `bind = $mainMod SHIFT, Q, signal, 9` or: `bind = $mainMod SHIFT, Q, signal, 15`
> 
> depending on how the user wants to deal with the processes

alright, here you go.

you can do forcekillactive
or signal 9

signalwindow is also added

File: src/managers/KeybindManager.cpp [hunk 1]
```diff
@@ -978,24 +983,71 @@ uint64_t CKeybindManager::spawnRawProc(std::string args, PHLWORKSPACE pInitialWo
...
+    kill(PWINDOW->getPID(), SIGKILL);
+
+    return {};
+}
+
+SDispatchResult CKeybindManager::signalActive(std::string args) {
+    if (!std::all_of(args.begin(), args.end(), ::isdigit))
+        return {.success = false, .error = "signalActive: signal has to be int"};
+
+    kill(g_pCompositor->m_pLastWindow.lock()->getPID(), std::stoi(args));
```
vaxerski, line 1027: stoi can throw + no verif for idiotic values like 0 or 999999

src/managers/KeybindManager.cpp
```diff
@@ -978,24 +983,71 @@ uint64_t CKeybindManager::spawnRawProc(std::string args, PHLWORKSPACE pInitialWo
...
+
+    if (!PWINDOW) {
+        Debug::log(ERR, "signalWindow: no window");
+        return {.success = false, .error = "signalWindow: no window"};
+    }
+
+    if (!std::all_of(SIGNAL.begin(), SIGNAL.end(), ::isdigit))
+        return {.success = false, .error = "signalWindow: signal has to be int"};
+
+    kill(PWINDOW->getPID(), std::stoi(SIGNAL));
```
vaxerski, line 1046: same

vaxerski (commented):
wiki mr pls

izmyname:
> forcekill(active)

Imo, not the best name as it looks similar to already existing dispatcher, but  'force' is ambiguous and doesn't explain the difference.

src/managers/KeybindManager.cpp
littleblack111: > stoi can throw

That's why I wrote L1024

vaxerski: I know

vaxerski (approved):
thanks!

Issue #6844, opened 2024-07-11
GTK2 / Eclipse-based apps flicker in Wayland native mode


lufjo:
### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 918d8340afd652b011b937d29d5eea0be08467f5  (flake.lock: update).
Date: Tue Jun 25 12:06:02 2024
Tag: v0.41.2, commits: 4886

flags: (if any)


System Information:
System name: Linux
Node name: arch
Release: 6.9.8-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Fri, 05 Jul 2024 22:11:24 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU117M [GeForce GTX 1650 Ti Mobile] [10de:1f95] (rev a1) (prog-if 00 [VGA controller])
05:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Renoir [Radeon Vega Series / Radeon Vega Mobile Series] [1002:1636] (rev c6) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/lufjo/.config/hypr/hyprland.conf: Read Succeeded
#
# Please note not all available settings / options are set here.
# For a full list,see the wiki
#

source=~/.config/hypr/conf/basic.conf
source=~/.config/hypr/conf/rules.conf
source=~/.config/hypr/conf/env.conf
source=~/.config/hypr/conf/binds.conf

exec-once=fcitx5 -d
# exec-once=wl-paste --type text --watch cliphist store
exec-once=copyq --start-server
exec-once=~/.config/hypr/scripts/boot.sh


Config File: /home/lufjo/.config/hypr/conf/basic.conf: Read Succeeded
[...]

PR #8399, opened 2024-11-09
drm dumb buffers for cursors

vaxerski:
This adds support for aq's drm_dumb buffers to render cursors via the cpu.

For me on amd this nukes the entire system atm idk will have to debug

deps on https://github.com/hyprwm/aquamarine/pull/104

gated behind `cursor:use_cpu_buffer`

PR #7110, opened 2024-07-30
Buffer lock/release fixes

vaxerski:
Fixes a few edge cases with buffer locking and releasing. Needs testing.

Fixes #7091

Possibly impacted:
 - direct scanout
 - cursors

Untested:
 - xwayland

Issue #6912, opened 2024-07-17
dbeaver will cause memory overflow


Liu-WeiHu:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
<Paste the output of the command here>
```

</details>
[systeminfo.txt](https://github.com/user-attachments/files/16263802/systeminfo.txt)

### Description

https://github.com/user-attachments/assets/839e6af9-183b-4703-9b32-4c5cac355f7b

### How to reproduce

Just watch the video

Liu-WeiHu:
Every time a visual database table is opened, the memory of Hyprland will increase. When the Dbeaver program is closed, the memory will drop to about 100M.

fufexan:
You said it is a regression. Which version did it used to work fine on? Can you bisect?

vaxerski:
can you test on #6608

nzlov:
```
Hyprland, built from branch  at commit 4520b30d498daca8079365bdb909a8dea38e8d55  (version: bump to 0.44.1).
Date: Wed Oct 9 12:54:39 2024
Tag: v0.44.1, commits: 5309
built against aquamarine 0.4.3
```

```
NVIDIA-SMI 565.57.01              Driver Version: 565.57.01      CUDA Version: 12.7e
```

This still happens.

UjinT34:
Probably same as https://github.com/hyprwm/Hyprland/issues/6844

vaxerski:
likely

PR #8966, opened 2025-01-05
Fix a bunch of issues with dropCurrentBuffer

UjinT34:
Moved `dropPendingBuffer` closer to a point where we stop using it.
Commented out `dropCurrentBuffer` for sync buffers since it causes more trouble than solves.
Looks like it it either wrong to drop the buffer so early or there is a bug somewhere else. Gamescope complains about incorrect buffer releases from time to time (definitely not on every release) https://github.com/ValveSoftware/gamescope/blob/f873ec7868fe84d2850e91148bcbd6d6b19a7443/src/Backends/WaylandBackend.cpp#L786 For some reason it fails to detect that the buffer was acquired by the compositor. I guess that some other clients have similar issues but they are not smart enough to handle the release correctly and destroy the buffer upon receiving a release while they still need it for something internally.
This fixes 
https://github.com/hyprwm/Hyprland/issues/6844
https://github.com/hyprwm/Hyprland/issues/6912
https://github.com/hyprwm/Hyprland/pull/8399#issuecomment-2468427288 and bunch of other corrupted/invisible cursors with `cursor:use_cpu_buffer`

https://github.com/hyprwm/hyprpicker/issues/85 seems to work fine. Couldn't reproduce https://github.com/hyprwm/Hyprland/issues/7091 but if it'll come back with this PR I'd rather have some minor green artifacts while having a stroke trying to resize kitty than issues mentioned above.

Tested it for a long time since https://github.com/hyprwm/Hyprland/issues/6844#issuecomment-2473766368 without an issue. Still might need more testing, haven't checked everything related to https://github.com/hyprwm/Hyprland/pull/7110. Might behave differently with explicit sync disabled.
Maybe `isSynchronous` isn't always correct? Or some clients have a different idea about buffer sync? Or some events should be sent before the release?

Ready.

### DIFF
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -428,6 +428,7 @@ void CWLSurfaceResource::commitPendingState() {
     pending.damage.clear();
     pending.bufferDamage.clear();
     pending.newBuffer = false;
+    dropPendingBuffer(); // at this point current.buffer holds the same SP and we don't use pending anymore
 
     events.roleCommit.emit();
 
@@ -448,10 +449,9 @@ void CWLSurfaceResource::commitPendingState() {
 
         // release the buffer if it's synchronous as update() has done everything thats needed
         // so we can let the app know we're done.
-        if (current.buffer->buffer->isSynchronous()) {
-            dropCurrentBuffer();
-            dropPendingBuffer(); // pending atm is just a copied ref of the current, drop it too to send a release
-        }
+        // Some clients aren't ready to receive a release this early. Should be fine to release it on the next commitPendingState.
+        // if (current.buffer->buffer->isSynchronous())
+        //     dropCurrentBuffer();
     }
 
     // TODO: we should _accumulate_ and not replace above if sync


vaxerski:
oh, no, the kitty thing is a very bad problem (it's when we drop a dma buffer before we are done sampling from it)

vaxerski:
IIRC the early drop was for foot...?

vaxerski (approved):
tested, lgtm

goggi:
Hi! This is kind of strange, but after this commit I got stuttering in animations when a firefox based browser is present. It works fine if I revert back to https://github.com/hyprwm/Hyprland/commit/602d6b73563e56fe003bbf3d25b45a53c51bad24

UjinT34:
Doesn't happen to me with FF 128.5.0esr. Might be specific to some video driver and explicit sync combo.

goggi:
> Doesn't happen to me with FF 128.5.0esr. Might be specific to some video driver and explicit sync combo.
If I remove the comments for:
        // if (current.buffer->buffer->isSynchronous())
             // dropCurrentBuffer();
it works like expected again. 

Its a very minor shuttering in the animation. If I have the browser on workspace 1 and then move to workspace 2 that contains vscode, the animation shutters. If I move back and forward fast between the workspace then i get around 10 fps drop. This does not happen if  I move between workspace 2 (vscode) and workspace 3 (chrome)

UjinT34:
I guess dropping it there is too early for some scenarios and dropping it on the next `commitPendingState` is too late for other scenarios.
Couldn't reproduce it with intel and floorp 11.21.0.

UjinT34:
Try toggling explicit_sync and explicit_sync_kms. Looks like an amd bug or some specific config issue.
Basically client has two buffers A and B.
With this line
1. client submits A
2. hl does it's stuff with A
3. hl releases A
4. client is ready to handle the release of A
5. client submits B
6. client needs to work with A
7. hl does it's stuff with B
8. and so on...

Without this line
1. client submits A
2. hl does it's stuff with A
3. client is ready to handle the release of A
4. client submits B
5. client needs to work with A
6. hl releases A
7. hl does it's stuff with B
8. and so on...

Some clients are fine when A is released early. Some client are fine when A is released late. For everything to work fine we need to release A between 3 and 5. No idea where this point is... Maybe after `presentFeedback`? At least gamescope is ready to receive a release after some stuff related to `Present`.

PR #8981, opened 2025-01-07
Add a periodic donation nag

vaxerski:
This adds a periodic donation nag. It will fire two times a year: around the second half of july, and in december.

Additionally, will not fire if:
 - the user has first used hyprland less than a month prior
 - disabled in the config
 - directly after an update if the update popup came up
 - your qtutils is outdated or missing

image

It will only fire _once_ and go away until the next "nag time" (july, december)

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -9,6 +9,7 @@
 #include "managers/PointerManager.hpp"
 #include "managers/SeatManager.hpp"
 #include "managers/VersionKeeperManager.hpp"
+#include "managers/DonationNagManager.hpp"
 #include "managers/eventLoop/EventLoopManager.hpp"
 #include <aquamarine/output/Output.hpp>
 #include <bit>
@@ -544,6 +545,8 @@ void CCompositor::cleanup() {
     g_pSeatManager.reset();
     g_pHyprCtl.reset();
     g_pEventLoopManager.reset();
+    g_pVersionKeeperMgr.reset();
+    g_pDonationNagManager.reset();
 
     if (m_pAqBackend)
         m_pAqBackend.reset();
@@ -644,6 +647,9 @@ void CCompositor::initManagers(eManagersInitStage stage) {
 
             Debug::log(LOG, "Creating the VersionKeeper!");
             g_pVersionKeeperMgr = std::make_unique<CVersionKeeperManager>();
+
+            Debug::log(LOG, "Creating the DonationNag!");
+            g_pDonationNagManager = std::make_unique<CDonationNagManager>();
 
             Debug::log(LOG, "Starting XWayland");
             g_pXWayland = std::make_unique<CXWayland>(g_pCompositor->m_bEnableXwayland);
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -25,6 +25,7 @@
 #endif
 #include <ranges>
 #include "helpers/varlist/VarList.hpp"
+#include "helpers/fs/FsUtils.hpp"
 #include "protocols/FractionalScale.hpp"
 #include "protocols/PointerConstraints.hpp"
 #include "protocols/LayerShell.hpp"
@@ -2629,7 +2630,7 @@ void CCompositor::performUserChecks() {
     }
 
     if (!*PNOCHECKQTUTILS) {
-        if (!executableExistsInPath("hyprland-dialog")) {
+        if (!NFsUtils::executableExistsInPath("hyprland-dialog")) {
             g_pHyprNotificationOverlay->addNotification(
                 "Your system does not have hyprland-qtutils installed. This is a runtime dependency for some dialogs. Consider installing it.", CHyprColor{}, 15000, ICON_WARNING);
         }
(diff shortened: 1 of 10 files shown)


izmyname:
Thunderbird vibe.

>led by one person

Seriously, though - try inviting existing maintainers to the team, so they could help. A one-man big project, without a clear roadmap and plans to counter so-called 'The Bus factor' is never a good thing.

Apologies for the offtopic.

vaxerski:
> try inviting existing maintainers to the team

I do occasionally, but I don't really have many people willing to work on the project in a consistent matter

izmyname:
> > try inviting existing maintainers to the team
> 
> I do occasionally, but I don't really have many people willing to work on the project in a consistent matter

I see. Still, some sort of roadmap and plans for a long-term maintenance would be a good thing, but anyway.

vaxerski:
roadmap:
- stuff

kurtbahartr:
I think this should be done more in osu!lazer way. You know how it pops up "Consider becoming an osu!supporter to help support osu!'s development :heart:" after the intro at the bottom left corner for a moment after each launch ("Consider donating to help support Hyprland's development" anyone?). The popup for hyprpm can be used for that if you ask me. And instead of a full-fledged "Donate" button, a portion of the text can be turned into a clickable link.

The timing is reasonable. It makes it feel like an Easter egg of sorts. But yeah, doing this instead of a window should make it feel much less like Windows (Activate Windows flashbacks).

vaxerski:
it's a small window twice a year. Come on.

kurtbahartr:
Even *that* could be annoying to people. I don't want to boot up my laptop and see such a window up and front. Why not use an existing code instead of adding another requirement within the code (`qtutils` in this case)?

vaxerski:
you can disable it.

kurtbahartr:
I mean yeah, I can disable it, but what if someone doesn't really follow up on the update changelogs and ends up getting hit by the window (*I'm looking at you, Arch and Gentoo users.*)? I'm sure not everyone is willing to check the wiki all the time.

alexlnkp:
> Even _that_ could be annoying to some niche group of people, such as those like me or those who had enough of Windows. I don't want to boot up my laptop and see such a window up and front. Why not use an existing code instead of adding another requirement within the code (`qtutils` in this case)?

Even KDE asks for donations. Developers need support.

myamusashi:
> Even _that_ could be annoying to some niche group of people, such as those like me or those who had enough of Windows. I don't want to boot up my laptop and see such a window up and front. Why not use an existing code instead of adding another requirement within the code (`qtutils` in this case)?

It's not that hard man, just disable and go "But what about people who don't look at the wiki all the time", just searching on google and boom congratulations you solved it,  it's not that hard man.

nnyyxxxx:
> I mean yeah, I can disable it, but what if someone doesn't really follow up on the update changelogs and ends up getting hit by the window (_I'm looking at you, Arch and Gentoo users._)? I'm sure not everyone is willing to check the wiki all the time.

This is FOSS, be grateful that we have Hyprland. If their is a donation nag, so be it; nobody is getting paid for this, unless through donations.

We're lucky to have graphical environments, stuff like this isn't easy to make.

kurtbahartr:
> > Even _that_ could be annoying to some niche group of people, such as those like me or those who had enough of Windows. I don't want to boot up my laptop and see such a window up and front. Why not use an existing code instead of adding another requirement within the code (`qtutils` in this case)?
> 
> Even KDE asks for donations. Developers need support.

And they do it in form of a notification, not with a window.

> It's not that hard man, just disable and go "But what about people who don't look at the wiki all the time", just searching on google and boom congratulations you solved it, it's not that hard man.

image

I get where you're coming from, but give it a month or two before this gets dumped down by Google's cruel SEO shit.

> This is FOSS, be grateful that we have Hyprland. If their is a donation nag, so be it; nobody is getting paid for this, unless through donations.
> 
> We're lucky to have graphical environments, stuff like this isn't easy to make.

My issue is not the fact that it exists. This is my very first time complaining about a donation nag after having seen it countless times on osu! and other freewares.

My issue is the fact that it's made in a way that literally makes me forget what I was about to do on my computer when I boot it up. Why not improve/revise something for the better when you can do exactly that? Is it really too hard to turn this thing into a notification like done with HyprPM needing a Hyprland restart after `hyprpm update`? Do I really have to plan ahead of my already-unbelievably-tight schedule that I'm cutting my sleep short for and spare time just to make this?

Let me make this clear. I'm not saying that this PR should be reverted - It shouldn't because it's a work of community after all. I'm saying that it could've been made miles better by turning this into a notification like done with HyprPM stuff. Same with update notifications. Same code but with wider range of usage. You guys are over-exaggerating it and making it seem like I hate Hyprland as a whole - While I myself am a Hyprland user!

image

^^^^^ See this? This is what I was trying to say for the donation nag as well!

vaxerski:
KDE has a built-in notification center. Hyprland doesn't. If someone doesn't notice, or doesn't manage to read it in time, it's gone.

Plus, these notifications don't support buttons, like KDE's do.

You are very unlikely to get this changed, I'll tell you that. It's not like the design decision here was "meh I thought so", I did think through the options we have.

ca5ua1:
I will input my 2c:
Maybe set donation reminder as "text at the bottom of default wallpaper"?

When I change wallpaper I 10/10 times read it.
And it's not as annoying as pop-up dialog box.

If it have to be done, I would want it this way

vaxerski:
> When I change wallpaper I 10/10 times read it.

People don't do that almost at all. The splashes are random.

Please consider being human for a second, thank you. You're getting free software. You can disable the popup entirely within 2 minutes.

Issue #8977, opened 2025-01-07
Add tag: to the window param type for use with dispatchers


Ringflash:
### Description

It would be very useful to tag a window with shortcut and then use the dispatcher **focuswindow** to focus tagged one.

For example tag different terminals with tags and quickly jump to the specific one with a shortcut or tag apps that are currently in use for some project (e.g. two terminals, two browsers, IDE, etc)

Truenya:
not done yet? would be nice to close

vaxerski:
done, idk why it didnt close

PR #8985, opened 2025-01-08
Add tag: to window parameters for matching tag(s) by regex

zakk4223:
Adds a tag: argument to dispatcher window params. So you can do stuff like focuswindow/closewindow by tag.
See #8977 

Nope.

Tis ready


zakk4223:
I thought about adding a way to match multiple tags with this (window must have tags A and B and C) but I cannot think of any way to do that without the config file/dispatcher syntax being absolutely terrible

vaxerski (approved):
dont then :D

zakk4223:
https://github.com/hyprwm/hyprland-wiki/pull/934

PR #8998, opened 2025-01-09
Add `movecursortomonitor` dispatcher

Insprill:
Adds the `movecursortomonitor` dispatcher to move the cursor to a specific monitor.

I think I did the hyprctl completions correctly, but please double-check.

Yes


vaxerski:
...what's the point?

Insprill:
My main use case is part of a workaround for the [split-monitor-workspaces](https://github.com/Duckonaut/split-monitor-workspaces) plugin I use. I have the `cursor:default_monitor` config value set, but when the plugin loads, it moves the default workspace (and cursor) to a different monitor, making that config option useless. With this MR, I used a hyprctl command to set the cursor back to my default display after loading plugins.

Since there are already dispatchers for moving the cursor to an arbitrary position and the corner of a window, as well as for moving workspaces to specific monitors, I thought adding one to move the cursor to a monitor was fitting. If it's unwanted, I'll fix it in the plugin itself (which I might do anyways), but I figured I'd try to upstream something first in case it'd help others, too.

vaxerski:
why not use `focusmonitor` instead though?

Insprill:
Yikes, that does what I need. No clue how I missed that dispatcher. Sorry :sweat_smile:

PR #9006, opened 2025-01-09
nix/module.nix: expand nixos module for configuring hyprland

bloxx12:
Expand the nixos module to be able to configure hyprland, just like the current home-manager module does. I think this would be neat, would probably not be upstreamed into nixpkgs though.  

@fufexan  and I already talked a bit about this, I got around to making a PR only now though.

It works on my machine™, feel free to test.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -7,12 +7,7 @@ inputs: {
   inherit (pkgs.stdenv.hostPlatform) system;
   cfg = config.programs.hyprland;
 
-  package = inputs.self.packages.${system}.hyprland;
-  portalPackage = inputs.self.packages.${system}.xdg-desktop-portal-hyprland.override {
-    hyprland = cfg.finalPackage;
-  };
-
-# basically 1:1 taken from https://github.com/nix-community/home-manager/blob/master/modules/services/window-managers/hyprland.nix
+  # basically 1:1 taken from https://github.com/nix-community/home-manager/blob/master/modules/services/window-managers/hyprland.nix
   toHyprconf = {
     attrs,
     indentLevel ? 0,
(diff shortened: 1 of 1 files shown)


nix/module.nix
```diff
@@ -11,11 +11,209 @@ inputs: {
...
+        in
+          valueType;
+        default = {};
+        description = ''
+          Hyprland configuration written in Nix. Entries with the same key
+          should be written as lists. Variables' and colors' names should be
+          quoted. See <https://wiki.hyprland.org> for more examples.
+
+          ::: {.note}
+          Use the [](#opt-wayland.windowManager.hyprland.plugins) option to
```
fufexan, line 115:
```suggestion
          Use the [](#programs.hyprland.plugins) option to
```

nix/module.nix
```diff
@@ -11,11 +11,209 @@ inputs: {
...
+          submap = resize
+          binde = , right, resizeactive, 10 0
+          binde = , left, resizeactive, -10 0
+          binde = , up, resizeactive, 0 -10
+          binde = , down, resizeactive, 0 10
+          bind = , escape, submap, reset
+          submap = reset
+        '';
+        description = ''
+          Extra configuration lines to add to `~/.config/hypr/hyprland.conf`.
```
fufexan, line 155:
```suggestion
          Extra configuration lines to add to `/etc/xdg/hypr/hyprland.conf`.
```

fufexan (commented):
From a glance it looks alright, but I'll need to test it before approving.

fufexan:
This might be a skill issue on my end, but setting plugins causes this error. I'll keep investigating.
[stacktrace.txt](https://github.com/user-attachments/files/18389665/stacktrace.txt)

fufexan:
I've finally had time to debug this. Turns out the `package` definition in the top-most `let...in` was shadowing `with lib.types; ... package`. I've moved the whole thing down and added a `lib.mkMerge` for legibility.
Also removed adding xwayland as that's already handled.

fufexan (approved):
Currently running the config generated by this module. All seems well.

LGTM

PR #9014, opened 2025-01-09
Do not set hdr metadata unless needed

UjinT34:
Create and send new hdr metadata only when it changes. Should fix increased cpu usage on some hw configurations.

Ready. Requires https://github.com/hyprwm/aquamarine/pull/95


Trimutex:
Confimed this PR + AQ PR fixes the CPU issue, thanks!

There's now a heavy performance hit on games, but I think its a driver issue at this point. TYVM

vaxerski (approved):

PR #9016, opened 2025-01-09
hyprctl: add lastwindowclass to workspaces

LivingIchor:

Just adds the field "lastwindowclass" to the out put of `hyprctl workspaces`

No

Yes

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -308,15 +308,17 @@ std::string CHyprCtl::getWorkspaceData(PHLWORKSPACE w, eHyprCtlOutputFormat form
     "windows": {},
     "hasfullscreen": {},
     "lastwindow": "0x{:x}",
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
useless. Query the window from hyprctl clients...?

LivingIchor:
While hyprctl clients can be used, I don't understand why there should be the misdirection of including lastwindowtitle in workspaces but not including the class.

vaxerski:
because we are trashing the output for no reason imo

LivingIchor:
The reason I was wanting the change was for getting the icon of the last focused window on all the separate workspaces, maybe I'm using the class wrong, but it seems to align with the app name. I was just confused as to why lastwindowclass wasn't included with hyprctl workspaces. But if you think adding it is too big a breaking change I can just keep self-aligning workspaces and clients output

PR #9030, opened 2025-01-11
animation: fixup adding animvars during ::tick

PaideiaDilemma:

Removes the FIXME regarding adding animvars via callbacks during a tick.
Looping via indices resolves the problem when vars get added to active via a callback.

Vars are only actively removed from `m_vActiveAnimatedVaraibles` via the dtor of `CBaseAnimatedVariable`. But a `CBaseAnimatedVariable` can't be deleted during a callback, since we own a SP to it when calling the callback in the first place. So having elements of `m_vActiveAnimatedVariables` be removed via a callback during `::tick` is currently not possible anyways.

No.

Ready.


vaxerski (approved):
39

PR #9049, opened 2025-01-13
protocols: immediately copy toplevel content when ignoreDamage set

outfoxxed:
Makes hyprland_toplevel_export_frame_v1.copy attempt to capture a frame immediately if ignoreDamage was set. Additionally makes the field of the same name actually do something, as it was previously never set.

I'm not sure if anything else needs to be done to ensure the window is in a good state to capture, e.g. damage the monitor or the window, but it seems to work as is.

Ready to merge.

### DIFF
--- a/src/protocols/ToplevelExport.cpp
+++ b/src/protocols/ToplevelExport.cpp
@@ -202,7 +202,12 @@ void CToplevelExportFrame::copy(CHyprlandToplevelExportFrameV1* pFrame, wl_resou
 
     buffer = PBUFFER->buffer;
 
-    PROTO::toplevelExport->m_vFramesAwaitingWrite.emplace_back(self);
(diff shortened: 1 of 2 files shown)


src/protocols/ToplevelExport.cpp
```diff
@@ -202,7 +202,12 @@ void CToplevelExportFrame::copy(CHyprlandToplevelExportFrameV1* pFrame, wl_resou
 
     buffer = PBUFFER->buffer;
 
-    PROTO::toplevelExport->m_vFramesAwaitingWrite.emplace_back(self);
+    m_ignoreDamage = ignoreDamage;
+    if (ignoreDamage && validMapped(pWindow)) {
```
vaxerski, line 206: what is this {} fox

vaxerski (approved):

Issue #9061, opened 2025-01-14
Unify member variable naming


vaxerski:
### Description

If someone is bored, can you unify member variable naming? I've initially used hungarian + m, `m_bBooleanValue` but then I stopped, if someone can just move it to camel + m_ for members it would be nice. (`m_value`, `m_speed`, `m_amogus`) 

obviously m_ for members of classes / structs

ninetailedtori:
I could for sure :)
Lemme go through it all with a linter or smth, I'll write up a linter format for hypr :)

vaxerski:
we have clang-tidy, if you know how to write a lint rule for that and auto apply it, would be appreciated

vaxerski:
@ninetailedtori any progress on that?

ninetailedtori:
Yep, it's nearly done. Lemme commit it to my fork in a bit :)

vaxerski:
thanks! no hurries, just wanted to check in

ninetailedtori:
Yep, after my clang-tidy and redo of several grammatical fixes...there's about 850 renames I'm about to push to my fork...heheh. Do you want ALL global pointers to have g_, or only the singleton pointers? If all, I can more easily do the formatting, but if not, then I'll need to be a little more careful.

vaxerski:
are there any other global pointers? all is fine.

ninetailedtori:
I'll be moving all statics and functions in named namespaces to ones in anonymous ones, in line with misc-use-internal-linkage, if that's ok?

ninetailedtori:
misc-use-internal-linkage wants us to move a bunch of functions into a single file, resolving to use an anonymous namespace instead of a named one. Would you like me to do this as well?

ninetailedtori:
To be honest, hyprctl is already in a single file, and can be statically linked, or anonymous namespaced with ease. It might be worth moving some of the others over too, like hyprpm and the other modules.

vaxerski:
ignore clang-tidy for now, just change the variable naming. Other clang-tidy stuff can be dealt with later.

ninetailedtori:
Yep [it's bed time](https://github.com/ninetailedtori/Hyprland/commit/6b18d956cda29e4916468860291203186df3e261), I'm gonna sort the rest out tmr and hopefully can get the whole repo sorted within a few days. We'll have to do the others as well though...aquamarine especially.

vaxerski:
[...]

PR #9077, opened 2025-01-15
Check that Xwayland exists before enabling it

earboxer:

Fixes launching Hyprland when you don't have Xwayland installed

Hyprland 0.46 would crash almost immediately if Xwayland wasn't installed unless you put `xwayland:enabled = false` in your config.

I think this is ready for merging

### DIFF
--- a/src/xwayland/XWayland.cpp
+++ b/src/xwayland/XWayland.cpp
@@ -1,21 +1,39 @@
 #include "XWayland.hpp"
+#include "../Compositor.hpp"
 #include "../debug/Log.hpp"
+#include "../helpers/fs/FsUtils.hpp"
 
-CXWayland::CXWayland(const bool enabled) {
+CXWayland::CXWayland(const bool wantsEnabled) {
 #ifndef NO_XWAYLAND
+    // Disable Xwayland and clean up if the user disabled it.
+    if (!wantsEnabled) {
+        Debug::log(LOG, "XWayland has been disabled, cleaning up...");
+        for (auto& w : g_pCompositor->m_vWindows) {
+            if (!w->m_bIsX11)
+                continue;
+            g_pCompositor->closeWindow(w);
+        }
+        unsetenv("DISPLAY");
+        m_enabled = false;
+        return;
+    }
+
+    if (!NFsUtils::executableExistsInPath("Xwayland")) {
+        // If Xwayland doesn't exist, don't try to start it.
+        Debug::log(LOG, "Unable to find XWayland; not starting it.");
+        return;
+    }
+
     Debug::log(LOG, "Starting up the XWayland server");
 
     pServer = std::make_unique<CXWaylandServer>();
-
-    if (!enabled) {
-        unsetenv("DISPLAY");
-        return;
-    }
 
     if (!pServer->create()) {
         Debug::log(ERR, "XWayland failed to start: it will not work.");
         return;
     }
+
+    m_enabled = true;
 #else
     Debug::log(LOG, "Not starting XWayland: disabled at compile time");
 #endif
@@ -30,4 +48,8 @@ void CXWayland::setCursor(unsigned char* pixData, uint32_t stride, const Vector2
 
     pWM->setCursor(pixData, stride, size, hotspot);
 #endif
+}
+
+bool CXWayland::enabled() {
+    return m_enabled;
 }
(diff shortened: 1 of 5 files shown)


littleblack111:
wouldn't it be better for it to be set in compile time?

vaxerski (commented):
wouldnt it be better for this check to happen in the ctor of cxwayland so that it skips server

vaxerski:
@littleblack111 no

earboxer:
> wouldnt it be better for this check to happen in the ctor of cxwayland so that it skips server

Can any of Xwayland work if it's not installed?

I was thinking with putting it here that if it wasn't installed, one could install it, then reload the config and hyprland would initialize xwayland, seeing that `prevEnabledXwayland` is not the same as `PENABLEXWAYLAND`.

vaxerski:
I was talking about moving the executable check to the ctor of CXWayland. By the way we could also move m_bEnableXWayland -> CXWayland::enabled() (which is based on the ctor arg `const bool enabled`)

src/xwayland/XWayland.hpp
```diff
@@ -17,18 +17,22 @@ class CXWM;
...
+    bool enabled();
 
     void setCursor(unsigned char* pixData, uint32_t stride, const Vector2D& size, const Vector2D& hotspot);
 
     struct {
         CSignal newSurface;
     } events;
+
+  private:
+    bool m_benabled;
```
vaxerski, line 35: `m_enabled`. We are moving to drop hungarian see #9061

also please initialize it

src/xwayland/XWayland.hpp
```diff
@@ -17,18 +17,22 @@ class CXWM;
 
 class CXWayland {
   public:
-    CXWayland(const bool enabled);
+    CXWayland(const bool wants_enabled);
```
vaxerski, line 20: no snake. `wantsEnabled`

src/xwayland/XWayland.cpp
```diff
@@ -1,21 +1,36 @@
...
-    Debug::log(LOG, "Starting up the XWayland server");
-
-    pServer = std::make_unique<CXWaylandServer>();
-
-    if (!enabled) {
+    // Disable Xwayland and clean up if the user disabled it.
+    if (!wants_enabled) {
+        Debug::log(LOG, "XWayland has been disabled, cleaning up...");
+        for (auto& w : g_pCompositor->m_vWindows) {
+            if (w->m_pXDGSurface || !w->m_bIsX11)
```
vaxerski, line 12: one excludes the other. (just !w->m_bIsX11 is ok)

src/xwayland/XWayland.cpp
```diff
@@ -1,21 +1,36 @@
...
+                continue;
+            g_pCompositor->closeWindow(w);
+        }
         unsetenv("DISPLAY");
+        m_benabled = false;
+        return;
+    }
+    if (!NFsUtils::executableExistsInPath("Xwayland")) {
```
vaxerski, line 20: can you put a newline above this if? Having } and if without a break is a bit cluttery

src/xwayland/XWayland.cpp
```diff
@@ -1,21 +1,36 @@
...
+        // If Xwayland doesn't exist, don't try to start it.
+        Debug::log(LOG, "Unable to find XWayland; not starting it.");
         return;
     }
+    Debug::log(LOG, "Starting up the XWayland server");
```
vaxerski, line 25: same here and below with m_benabled = true

vaxerski (commented):
some style rest good

src/xwayland/XWayland.hpp
```diff
@@ -17,18 +17,22 @@ class CXWM;
...
+    bool enabled();
 
     void setCursor(unsigned char* pixData, uint32_t stride, const Vector2D& size, const Vector2D& hotspot);
 
     struct {
         CSignal newSurface;
     } events;
+
+  private:
+    bool m_enabled;
```
vaxerski, line 35: still not initializeddd

earboxer: Sorry, I don't write c++ as my day job :sweat_smile: Hopefully good this time

vaxerski:
we got a bit of a conflict here

vaxerski (approved):
thanks!

Issue #4272, opened 2023-12-27
Change binary name from Hyprland -> hyprland


hughesjs:
Most linux binaries are lowercased by convention.

It's quite irritating to have to remember that `Hyprland` has a capital `H`.

Could we change this? 

It would also then line up with `hyprctl`.

MightyPlaza:
alias ?

vaxerski:
alias?

hughesjs:
Like I know I *can* alias. I have aliased it... However, it just feels to me like it probably shouldn't be necessary and this would be a nice little QoL improvement

vaxerski:
yeah but now that it's been Hyprland so long it would be a problem the other way, for all the people who are used to the capital

hughesjs:
I think people will be generally fairly happy to revert to the norm, even if they are used to the capital. Unlearning an exception to a rule is generally easier than learning an exception

vaxerski:
what if we ship an `ln -s` for the time being?

correct me if I am wrong but I think you can make relative links like `./Hyprland`

vincentaxhe:
Hyprland reminds  people he is the one and true supervisor.

vaxerski:
added a link in the makefile.

hughesjs:
I think that's a reasonable compromise, is that going to be a long-term thing or will it be deprecated in a bit?

vaxerski:
idk, either way it's minor, if anything.

tienbuigia:
I support this lowercase change. And It would be better to add a link to the capital case just so old user doesn't need to change their config, but the official should be lowercase.

vaxerski:
what's the difference...?

tienbuigia:
So we can deprecate the capital one and remove it latter on.

vaxerski:
which one is the symlink and which one is not does not change that.

tienbuigia:
You are right.

hughesjs:
> which one is the symlink and which one is not does not change that.

I don't think it makes a practical difference. It does make a signalling difference though. Hugely minor mind.

hholst80:
Instead of low prio label it as "good First issue" and invite people to help

vaxerski:
what is there to help with??

hughesjs:
> added a link in the makefile.

> Instead of low prio label it as "good First issue" and invite people to help

[...]

PR #9091, opened 2025-01-16
Meson: properly install 'hyprland' symlink

fufexan:

Fixes https://github.com/hyprwm/Hyprland/issues/4272#issuecomment-2594539485.

Looks like it works, but untested.

Should be ready. Please test and report back.


fufexan:
CMake just doesn't want to install the symlink, so we'll probably fallback to the old mechanism, which doesn't work when you invoke it with `-P`.

Issue #9057, opened 2025-01-13
Wrong initial window size on unfocused monitor


Kam1k4dze:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

Probably not a regression / I don't remember it happening before

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.46.0 built from branch main at commit a6b263713a2b862ed41362082e2147e081934077  (protocols: allow hyprland-toplevel-export to capture hidden windows (9041)).
Date: Sun Jan 12 22:10:36 2025
Tag: v0.46.0-115-ga6b26371, commits: 5670
built against:
 aquamarine 0.7.1
 hyprlang 0.6.0
 hyprutils 0.3.3
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set


System Information:
System name: Linux
Node name: arch
Release: 6.12.9-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 10 Jan 2025 00:39:35 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA104 [GeForce RTX 3060] [10de:2487] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  565.77  Release Build  (archlinux-builder@)  


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Description

When a window opens while the current focus is on a different monitor, it briefly appears at an incorrect size. This issue occurs regardless of whether animations are enabled or disabled.

### How to reproduce

1. Focus a window on one monitor.
2. Attempt to open a window on a different monitor.

### Attach not paste

- [X] I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam

### Checklist of files to include below

[...]

PR #9116, opened 2025-01-18
Fix size prediction when the focused window and cursor are on different monitors

Kam1k4dze:
Fixes https://github.com/hyprwm/Hyprland/issues/9057

No

Ready

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -139,10 +139,10 @@ animations {
 # uncomment all if you wish to use that.
 # workspace = w[tv1], gapsout:0, gapsin:0
 # workspace = f[1], gapsout:0, gapsin:0
-# windowrulev2 = bordersize 0, floating:0, onworkspace:w[tv1]
-# windowrulev2 = rounding 0, floating:0, onworkspace:w[tv1]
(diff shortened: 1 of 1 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -1084,7 +1084,7 @@ Vector2D CHyprDwindleLayout::predictSizeForNewWindowTiled() {
     // create a fake node
     SDwindleNodeData node;
 
-    if (!candidate)
+    if (!candidate || candidate->monitorID() != g_pCompositor->m_pLastMonitor->ID)
```
vaxerski, line 1087: this will always predict the full monitor even if it is populated. That's not the way to do it, we probably should find a candidate on the other monitor and go from that

Kam1k4dze: You're right and this is the right way, but even when it predicts bigger than needed, I don't see any artifacts or bugs. Maybe I'm not seeing something because logically you should expect something to go wrong if the predicted size is wrong.

vaxerski: you do see bugs you just dont notice them cuz they arent as visible

PR #9118, opened 2025-01-19
reserve vector sizes as much as we can

gulafaran:
avoid reallocations as much as possible with a few edge cases where the reservation overshoots a tiny bit. but a few bytes of memory short term is better used then the overhead of potential reallocation.


vaxerski (approved):
thanks!

PR #9121, opened 2025-01-20
Touchpad gestures: workspace speed multiplier

thicileares:
- When using the workspace swipe gestures, the movement was too fast for my liking. 
- So I added a config option to slow down (or speed up) the movement by an arbitrary factor.

For example:

- `workspace_swipe_speed_multiplier = 1`: windows will move one-to-one with touchpad movement (current, default behavior)
- `workspace_swipe_speed_multiplier = 0.5`: windows will move half as fast when swiping on the touchpad
- `workspace_swipe_speed_multiplier = 2`: windows will move twice as fast

Pretty basic addition and I think it would be useful for others.

Ready to merge (as far as I can tell).

### DIFF
--- a/src/managers/input/Swipe.cpp
+++ b/src/managers/input/Swipe.cpp
@@ -4,8 +4,6 @@
 #include "../../config/ConfigValue.hpp"
 #include "../../managers/HookSystemManager.hpp"
 #include "../../render/Renderer.hpp"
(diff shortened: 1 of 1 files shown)


earboxer:
Is this just the inverse of `workspace_swipe_distance`?

If so, I think this should replace that, unless there's some reasonable reason to have both....

PR #9134, opened 2025-01-22
xwayland: fix sending large clipboard data

DDoSolitary:

Wayland to X11:
Wayland clients might not expect a non-blocking fd. e.g. KSystemClipboard in KGuiAddons (used by Flameshot) [uses a single write to the fd and doesn't check the return value](https://github.com/KDE/kguiaddons/blob/35e508e2061e4c3d843a4be986613fb3bc954835/src/systemclipboard/waylandclipboard.cpp#L379). If we pass a non-blocking fd, the write call will return early and result in partially transferred data.
This PR removes the O_NONBLOCK flag from the pipe fd passed to Wayland clients. Clients requiring non-blocking I/O can still set the flag themselves.

X11 to Wayland:
The fd received from Wayland clients [has O_NONBLOCK set](https://github.com/hyprwm/Hyprland/blob/ce48bc540824fcaefd777959703df9908e258d0a/src/xwayland/XDataSource.cpp#L83), but we only write to it once and error out if data is transferred partially. Instead we should use the Wayland event loop to keep writing to the fd until all data is transferred.

INCR is still not implemented so the size of clipboard data is still limited by x11 request size, which is 16MB by default.

Y

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -1163,6 +1163,11 @@ void CXWM::setClipboardToWayland(SXSelection& sel) {
     g_pSeatManager->setCurrentSelection(sel.dataSource);
 }
 
+static int writeDataSource(int fd, uint32_t mask, void* data) {
+    auto selection = (SXSelection*)data;
+    return selection->onWrite();
+}
+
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
thanks!

PR #9137, opened 2025-01-22
xwayland: fix clipboard mime name and atom mismatch

DDoSolitary:

The TARGETS and TIMESTAMP targets are skipped when processing the target list returned from X11 clipboard sources, but still added to the mimeAtoms list erroneously. As a result, the mimeAtoms and mimeTypes lists are out of sync and the target atom later passed to xcb_convert_selection will be incorrect if an mime type appearing after the aforementioned two targets is used.

N/A

Y


vaxerski (approved):
thanks!

PR #9145, opened 2025-01-23
ci: clang-format fix

littleblack111:
fix again, (see https://discord.com/channels/961691461554950145/1070436481912549497/1331967678763499580)

used the second method:
> You can try switching to use pull\_request\_target as the event, which will cause the workflow to run in your branch context — the target of the PR itself. I think that will get the default GITHUB\_TOKEN working, or you could optionally provide your own token as a secret on your repo and it will get picked up even if a fork triggered the run.

> One minor caveat to pull\_request\_target that I've noticed is that it makes workflow development somewhat more difficult, since the workflow and action changes in an active PR won't be picked up. Not a huge issue, but worth being aware of.

yes


vaxerski (commented):
This nukes c-f checks on main though, which also need to happen, just without any comments

littleblack111:
> This nukes c-f checks on main though, which also need to happen, just without any comments

mb i forgot to read, however, i did found an action that can add comment to commit. or should i just let it fail

vaxerski:
Just fail it, no need to add a comment to a commit.

Also can you rebase so that ci doesnt fail

littleblack111:
done

littleblack111:
cc: @vaxerski

vaxerski (approved):

Issue #9065, opened 2025-01-14
Option to enable border gradient angle animation upon focus change/refocus


caffeine01:
### Description

At the moment, the two options for border animations are simply "once" or "loop".
I feel as though it would be cool to have an option where upon focus change, the new focused window has its border re-animated once for example.

vaxerski:
isn't that what happens with `once`?

caffeine01:
> isn't that what happens with `once`?

unfortunately not... either im a dumbass or u did a fuckup at some point and just didn't notice.
i'm willing to accept either, i jus want the funny animations lol.

nishiiko:
to me it looks like it is trying to play an animation with ``once`` but its bugged
opening windows plays the animations fine, but refocusing plays the animation instantly
i made the borders massive to make it easier to see here

https://github.com/user-attachments/assets/6831fd74-3d01-42df-91dd-218939b5c536

caffeine01:
why are your window borders so fuckin massive bro

gksudolol:
> why are your window borders so fuckin massive bro

*"i made the borders massive to make it easier to see here"*

caffeine01:
> > why are your window borders so fuckin massive bro
> 
> _"i made the borders massive to make it easier to see here"_

shut up

vaxerski:
might be bugged then.

@caffeine01 chill out.

vaxerski:
@PaideiaDilemma might be damage? Can you check its not your MR that broke this?

I am fairly certain it should anim on ofcus change

caffeine01:
> ofcus change

vaxerski:
paideia's move animations to hyprutils one

caffeine01:
>I am fairly certain it should anim on ofcus change

***just to reiterate what vax said just a few comments ago***

*My opinion noone asked for but im gonna say anyway:*
Anim each focus > anim on creation anyways. Either way there should at least be a toggle between them in this case, because i can see why people would want either option.
like, `borderangle, a, b, default, focus` and  `borderangle, a, b, default, once`

vaxerski:
it should play the anim on focus change

caffeine01:
mind blowing

PR #9149, opened 2025-01-23
animation: fix borderangle once

PaideiaDilemma:

Closes https://github.com/hyprwm/Hyprland/issues/9065

Personally I would prefer if it we would increment the progress by 1 on a focus change. That way it would always rotate by 360 degrees when the focus changes. But it would mean that the angle is not static and can shift, so the way it is now probably makes more sense.

Ready


vaxerski:
it should animate once on focus change not all the time

vaxerski:
hm, probably. This would be confusing. Maybe make a func to update border anim and call in focusWindow?

PaideiaDilemma:
Done

vaxerski (approved):
thanks!

caffeine01:
pog

caffeine01:
> been watching this pull for an hour now

PR #9156, opened 2025-01-24
layershell: check if layer is valid

heather7283:
Fixes compositor crash when client tried to create a layer surface with invalid layer argument

When playing around with wayland clients I noticed that hyprland would crash when I try to pass an invalid layer param, turns out it was causing out of bounds array access in https://github.com/hyprwm/Hyprland/blob/57a39984dddd00fd1aca436e149b7566e5e48d95/src/desktop/LayerSurface.cpp#L37

Basically copy pasted the check from https://github.com/hyprwm/Hyprland/blob/57a39984dddd00fd1aca436e149b7566e5e48d95/src/protocols/LayerShell.cpp#L147

Nope

Ready

### DIFF
--- a/src/protocols/LayerShell.cpp
+++ b/src/protocols/LayerShell.cpp
@@ -233,6 +233,11 @@ void CLayerShellProtocol::onGetLayerSurface(CZwlrLayerShellV1* pMgr, uint32_t id
         return;
     }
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks!

Issue #8826, opened 2024-12-23
Window briefly visible when using movetoworkspacesilent to move a window from a non-visible workspace to a non-visible workspace


soufhehe:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.46.0 built from branch main at commit 5f7ad767dbf0bac9ddd6bf6c825fb9ed7921308a  (flake.lock: update).
Date: Sun Dec 22 23:39:17 2024
Tag: v0.46.0-27-g5f7ad767, commits: 5582
built against:
 aquamarine 0.5.1
 hyprlang 0.6.0
 hyprutils 0.3.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


flags set:
debug
```

</details>

### Description

I don't know if it's worth creating a bug report as it has no real use, feel free to close if you dont care

When a window is in another workspace and you move it to another workspace (not the currently active), the window appears briefly in the active workspace

### How to reproduce

foot in workspace 1
you are in workspace 2
`hyprctl dispatch movetoworkspacesilent 3,class:foot`

### Crash reports, logs, images, videos

https://github.com/user-attachments/assets/2ad099eb-d998-4583-9a58-958d621d9b3c

vaxerski:
can't repro on git

PaideiaDilemma:
It's only when moving to a workspace that does not exit yet and only properly visible with floating.

PR #9160, opened 2025-01-24
window: only set m_iMonitorMovedFrom, when moving to a different monitor

PaideiaDilemma:

Closes https://github.com/hyprwm/Hyprland/issues/8826

I think `m_iMonitorMovedFrom` was supposed to remain `-1` in case the window gets moved to another workspace on the same montior. That fixes the issue.

No

Ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1867,7 +1867,6 @@ void CCompositor::updateWindowAnimatedDecorationValues(PHLWINDOW pWindow) {
     }
 
     // opacity
-    const auto PWORKSPACE = pWindow->m_pWorkspace;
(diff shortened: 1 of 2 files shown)


src/Compositor.cpp
```diff
@@ -1867,7 +1867,6 @@ void CCompositor::updateWindowAnimatedDecorationValues(PHLWINDOW pWindow) {
     }
 
     // opacity
-    const auto PWORKSPACE = pWindow->m_pWorkspace;
```
PaideiaDilemma, line 1870: Unused

vaxerski (approved):
tanku!!!!

vaxerski:
I should test changes better this obviously "fixes" the issue by breaking the feature altogether.

PR #9161, opened 2025-01-24
compositor: fix incorrect cast, use lambda capture instead

heather7283:
Fixes a very sus `*(bool*)data` cast (which should've been `((std::pair<SP<CWLSurfaceResource>, bool>*)data)->second`) that was causing segfaults from time to time (haven't found a consistent way to reproduce it though), and replaces it with cleaner lambda capture.

No

Ready


vaxerski (approved):
that was a sus cast indeed, thanks :)

PR #9183, opened 2025-01-26
update groups on movewindow

user111111111111111111111111111111111:
fixes `movewindow` between monitors for grouped windows making hidden (when moved) group windows invisible and belonging to the wrong monitor.


vaxerski (approved):
thanks!

PR #9190, opened 2025-01-27
xwayland: respect window size set by configure requests

DDoSolitary:

Currently for XWayland windows, base_width & base_height from WM_NORMAL_HINTS is used as initial size if available. This is undesirable as some X11 applications (including VLC and LibreOffice) set initial window size by sending ConfigureWindow **before** MapWindow, which gets overridden by base size in handleMapRequest. Instead we should keep CXWaylandSurface::geometry, which is set by CreateWindow or ConfigureWindow, as is when possible.

Also, according to ICCCM:

>  The x, y, width, and height fields have been removed from the WM_NORMAL_HINTS property and replaced by pad fields. Values set into these fields will be ignored. The position and size of the window should be set by setting the appropriate window attributes. 

> The base_width and base_height elements in conjunction with width_inc and height_inc define an arithmetic progression of preferred window widths and heights for non-negative integers i and j

>  Window managers that honor aspect ratios should take into account the base size in determining the preferred window size. If a base size is provided along with the aspect ratio fields, the base size should be subtracted from the window size prior to checking that the aspect ratio falls in range.

We are not implementing constraints described in the latter two paragraphs, and the deprecated size fields are replaced by "setting the appropriate window attributes", which I think are the fields in CreateWindow and ConfigureWindow requests.

VLC and LibreOffice (run under XWayland) having bad initial window size:

image

N/A

Y

### DIFF
--- a/src/managers/XWaylandManager.cpp
+++ b/src/managers/XWaylandManager.cpp
@@ -82,29 +82,9 @@ CBox CHyprXWaylandManager::getGeometryForWindow(PHLWINDOW pWindow) {
 
     CBox box;
 
-    if (pWindow->m_bIsX11) {
-        const auto SIZEHINTS = pWindow->m_pXWaylandSurface->sizeHints.get();
-
-        if (SIZEHINTS && !pWindow->isX11OverrideRedirect()) {
-            // WM_SIZE_HINTS' x,y,w,h is deprecated it seems.
-            // Source: https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#wm_normal_hints_property
-            box.x = pWindow->m_pXWaylandSurface->geometry.x;
-            box.y = pWindow->m_pXWaylandSurface->geometry.y;
-
-            constexpr int ICCCM_USSize = 0x2;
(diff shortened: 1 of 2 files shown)


DDoSolitary:
The failing clang format check is not introduced in this pull request and I think should be addressed elsewhere.

DDoSolitary:
Also I noticed that there are commits pushed to main yesterday mentioning fixing initial window size: 445acec2a29adbb30196d8b92f86270a038b7ee5. However, among all applications that I noticed to have bad initial window size under XWayland: Firefox, Chromium, VLC, LibreOffice, only Firefox is fixed by that commit, while this PR should fix all these apps.

vaxerski:
that commit was relating to some windows (notably popup dialogs in e.g. filezilla, ATLauncher, etc) having incorrect sizes on launch that desync'd with hyprland, causing them to be blurry. Alongside that I did a small refactor (I want to get rid of CXWaylandManager cuz its a confusing name and stupid manager)
Your previous MR with a fake configure request might've solved that as well.

vaxerski (approved):
thanks!

PR #9193, opened 2025-01-27
xwayland: send synthetic configure events

DDoSolitary:

ICCCM says that when we move a window or receive ConfigureRequest but decide not to change the window size, a synthetic ConfigureNotify event should be sent in addition to the real event possibly generated by the X server.

This is mainly because:
1. For reparenting WMs, the event generated by the X server contains coordinates relative to the parent window, which are not quite useful, while the synthetic event should contain coordinates relative to the root window. Hyprland's XWM is not reparenting, so this should be irrelevant here.
2. If we decide not to change window geometry despite client's request (because the window is tiled for example), the client should be informed but the X server will not generate ConfigureNotify events if we call xcb_configure_window with the old geometry.

This should fix the problem that, when GIMP's window is tiled and wants to extend itself to display a large image, it will become unresponsive, waiting for a ConfigureNotify event.

N/A

Y

### DIFF
--- a/src/xwayland/XSurface.cpp
+++ b/src/xwayland/XSurface.cpp
@@ -169,6 +169,22 @@ void CXWaylandSurface::configure(const CBox& box) {
     uint32_t values[] = {box.x, box.y, box.width, box.height, 0};
     xcb_configure_window(g_pXWayland->pWM->connection, xID, mask, values);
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks! god-forsaken X.

PR #5595, opened 2024-04-15
Unbreak build on FreeBSD by adjusting dependencies

jbeich:
CMake (unlike Meson) doesn't add `-Wl,--as-needed` by default, so a bit more verbose.

PR #9197, opened 2025-01-27
Unbreak build on FreeBSD via libinotify-kqueue

jbeich:
Similar to #5595 (timerfd). Also used by Wayfire, pipewire, etc.


fufexan (approved):

vaxerski (approved):
thanks as always!

PR #9198, opened 2025-01-27
clear keyboard focus if found surface can't be focused

user111111111111111111111111111111111:
clears keyboard focus if found surface can't be keyboard focused
fixes active window not being set to null after closing a special workspace if cursor is above layer that's not keyboard focusable

unsure what side effects this can cause


vaxerski (commented):
if it's not keyboard focusable it can still be pointer focusable

user111111111111111111111111111111111:
> if it's not keyboard focusable it can still be pointer focusable

shouldn't mess with pointer focus now

vaxerski (commented):
it will, layers

PR #9201, opened 2025-01-28
fix two minor asan reported leaks

gulafaran:
free xkbSymState on destruction or call to clearManuallyAllocd if its allocated, otherwise we leak it.

globfree seems to only free internal resources allocated by glob() so lets free the glob_t structure aswell.


vaxerski (approved):
thanks!

PR #9219, opened 2025-01-29
src/config: add misc:watch_symlinks option

fufexan:

Adds `misc:watch_symlinks`, which makes inotify watch changes in the symlink, not in the file the symlink points to. This is especially useful on NixOS, and that's why it's disabled by default.

Wiki PR: TODO.

Should be good to go. Tested through `hyprctl keyword` as well.

Ready.


bloxx12 (approved):

vaxerski:
why not just make it the default, and watch both the symlink itself and the pointed file?

fufexan:
@vaxerski how bout this

src/config/ConfigWatcher.cpp
```diff
@@ -46,7 +47,11 @@ void CConfigWatcher::setWatchList(const std::vector<std::string>& paths) {
     // add new paths
     for (const auto& path : paths) {
         m_watches.emplace_back(SInotifyWatch{
-            .wd   = inotify_add_watch(m_inotifyFd, path.c_str(), IN_MODIFY),
+            .wd   = inotify_add_watch(m_inotifyFd, path.c_str(), IN_MODIFY | IN_DONT_FOLLOW),
+            .file = path,
+        });
+        m_watches.emplace_back(SInotifyWatch{
+            .wd   = inotify_add_watch(m_inotifyFd, (std::filesystem::canonical(path)).c_str(), IN_MODIFY),
```
vaxerski, line 54: canonical can throw. Needs an `std::error_code` and handling

src/config/ConfigWatcher.cpp
```diff
@@ -43,9 +44,16 @@ void CConfigWatcher::setWatchList(const std::vector<std::string>& paths) {
     // add new paths
     for (const auto& path : paths) {
         m_watches.emplace_back(SInotifyWatch{
-            .wd   = inotify_add_watch(m_inotifyFd.get(), path.c_str(), IN_MODIFY),
+            .wd   = inotify_add_watch(m_inotifyFd.get(), path.c_str(), IN_MODIFY | IN_DONT_FOLLOW),
             .file = path,
         });
+        std::error_code ec;
+        if (const auto CANONICAL = std::filesystem::canonical(path, ec); !ec) {
```
vaxerski, line 51: you should also check that it is a symlink at all, otherwise you get 2 watches for one file for no reason

vaxerski (approved):

erycwang:
For what it's worth, I'm currently using hyprland on CachyOS. I moved my files out of ~/.config/hypr to another folder so I could consolidate all my .dotfiles - when I symlinked the default folder to the canonical location, the auto-refresh on write stopped working. It works exactly once on the first write; but doesn't work after. I have tried using backup=no and yes for nvim for this. 

I wonder if anyone else has this issue?

fufexan:
> For what it's worth, I'm currently using hyprland on CachyOS. I moved my files out of ~/.config/hypr to another folder so I could consolidate all my .dotfiles - when I symlinked the default folder to the canonical location, the auto-refresh on write stopped working. It works exactly once on the first write; but doesn't work after. I have tried using backup=no and yes for nvim for this.
> 
> I wonder if anyone else has this issue?

Check logs with debug output enabled. See if there are any warnings or at least if the watcher detects changes.

PR #9221, opened 2025-01-29
nix/module: toHyprconf -> toHyprlang

fufexan:

Reworks the current Nix to hyprland.conf generator implementation by using a more generic Nix 2 Hyprlang generator.
This allows any application using hyprlang to use this for easy config generation.

This is the second step towards getting this module into Nixpkgs.

Considerations:
- [x] Rename `section` -> `category`, to be in line with the [documentation](https://wiki.hyprland.org/Hypr-Ecosystem/hyprlang/#categories).
- [x] Use `category[special]:field = value` for special categories. This reduces the complexity in finding the `special` key inside the attrset and then properly ordering it as the first in the generated category. But this poses a question: how should we let the user define these categories?
  - [ ] Write them verbatim (`"device[some-device:0baf3-952c]" = { sensitivity = 0.3; }`)
  - [ ] Parse a list of devices in the module, rewrite the attrset and pass it to `toHyprlang`:
      ```nix
      device = [
        {
          name = "some-device:0baf3-952c";
          sensitivity = 0.3;
        }
        {
          # ...
        }
      ];

      # this will be transformed to the syntax in the point above
      ```
- [x] Ensure nested categories are handled by recursing over them.
- [x] Use `category:field = value` for everything instead of nested categories (`category {...}`). This further reduces complexity by having only one code path for all categories.
- [x] Remove `indentLevel`, no longer needed without expanded categories.
- [x] Change `importantPrefixes` -> `topCommandsPrefixes` and also add `bottomCommandsPrefixes` for greater flexibility. Group these as the first argument (called `options` from now on).
- [x] Separate `attrs` into a standalone argument, after the `options` arg. This is the actual config that will be translated to hyprlang.

CC: @bloxx12 @donovanglover @JohnRTitor @khaneliman @notashelf

Needs testing. Different ideas and constructive criticism welcome.

Needs testing and discussion.

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -1,0 +1,59 @@
+lib: with lib; {
+  toHyprlang = {
+    topCommandsPrefixes ? ["$"],
+    bottomCommandsPrefixes ? [],
+  }: attrs: let
+    toHyprlang' = attrs: let
+      toStr = x:
+        if isBool x
+        then boolToString x
+        else toString x;
+
+      categories = filterAttrs (n: v: isAttrs v || (isList v && all isAttrs v)) attrs;
+      mkCategory = parent: attrs:
+        if lib.isList attrs
+        then concatMapStringsSep "\n" (a: mkCategory parent a) attrs
+        else
+          concatStringsSep "\n" (
+            mapAttrsToList (
+              k: v:
+                if isAttrs v
+                then mkCategory "${parent}:${k}" v
+                else if isList v
+                then concatMapStringsSep "\n" (item: "${parent}:${k} = ${toStr item}") v
+                else "${parent}:${k} = ${toStr v}"
+            )
+            attrs
+          );
+
+      mkCommands = toKeyValue {
+        mkKeyValue = mkKeyValueDefault {} " = ";
+        listsAsDuplicateKeys = true;
+        indent = "";
+      };
+
+      allCommands = filterAttrs (n: v: !(isAttrs v || (isList v && all isAttrs v))) attrs;
+
+      filterCommands = list: n: foldl (acc: prefix: acc || hasPrefix prefix n) false list;
+
+      # Get topCommands attr names
+      result = partition (filterCommands topCommandsPrefixes) (attrNames allCommands);
+      # Filter top commands from all commands
+      topCommands = filterAttrs (n: _: (builtins.elem n result.right)) allCommands;
+      # Remaining commands = allcallCommands - topCommands
+      remainingCommands = removeAttrs allCommands result.right;
+
+      # Get bottomCommands attr names
+      result2 = partition (filterCommands bottomCommandsPrefixes) result.wrong;
+      # Filter bottom commands from remainingCommands
+      bottomCommands = filterAttrs (n: _: (builtins.elem n result2.right)) remainingCommands;
+      # Regular commands = allCommands - topCommands - bottomCommands
+      regularCommands = removeAttrs remainingCommands result2.right;
+    in
+      mkCommands topCommands
+      + concatStringsSep "\n" (mapAttrsToList mkCategory categories)
+      + mkCommands regularCommands
+      + mkCommands bottomCommands;
+  in
+    toHyprlang' attrs;
+}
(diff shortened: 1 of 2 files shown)


nix/module.nix
```diff
@@ -182,29 +214,31 @@ in {
       environment.etc."xdg/hypr/hyprland.conf" = let
         shouldGenerate = cfg.extraConfig != "" || cfg.settings != {} || cfg.plugins != [];
```
JohnRTitor, line 215: Does this mean we can have distro specific configuration now? (which and user config would work together/be merged and then evaluated?)

fufexan: This is somewhat out of the scope of this PR, but I'm curious what @vaxerski would say about this.

vaxerski: eee whatev

fufexan:
Even simpler now, in case y'all found this too daunting to review :P

After implementing `flattenAttrs` I searched Nixpkgs PRs and found [this](https://github.com/NixOS/nixpkgs/pull/221608). :man_shrugging:

nix/lib.nix
```diff
@@ -0,0 +1,188 @@
+lib: with lib; let
```
NotAShelf, line 1: Please don't use `with lib;`

nix/lib.nix
```diff
@@ -0,0 +1,188 @@
...
+      # and a legible key-value separator.
+      mkCommands = generators.toKeyValue {
+        mkKeyValue = generators.mkKeyValueDefault {} " = ";
+        listsAsDuplicateKeys = true;
+        indent = ""; # No indent, since we don't have nesting
+      };
+
+      # Flatten the attrset, combining keys in a "path" like `"a:b:c" = "x"`.
+      # Uses `flattenAttrs` with a colon separator.
+      commands = flattenAttrs (p: k: "${p}:${k}") attrs;
```
NotAShelf, line 86: Can we guarantee that`p` and `k` are always strings? Other types might create issues while coercing I think.

I might also be overthinking this, ignore this comment if that's the case.

nix/lib.nix
fufexan: Sure, will use `inherit`. For some reason I remembered that `lib/generators.nix` uses `with lib;`, but it does not.

nix/lib.nix
fufexan: `p` and `k` are always the keys in an attrset, which have to be strings.

I think we should check that we're actually consuming an attrset, though I'm not sure what the policy is in Nixpkgs.

nix/lib.nix
fufexan: Done.

nix/lib.nix
fufexan:
```nix
nix-repl> l.toHyprlang {} []
error:
       … while calling the 'concatStringsSep' builtin
         at /nix/store/kdynjy1mbgkdg4p196v9gx6ljpf7q4nk-source/lib/strings.nix:95:31:
           94|   */
           95|   concatMapStrings = f: list: concatStrings (map f list);
             |                               ^
           96|

       … while calling the 'concatStringsSep' builtin
         at /nix/store/kdynjy1mbgkdg4p196v9gx6ljpf7q4nk-source/lib/generators.nix:183:13:
          182|         else k: v: [ (mkLine k v) ];
          183|   in attrs: concatStrings (concatLists (mapAttrsToList mkLines attrs));
             |             ^
          184|

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: expected a set but found a list: [ ]
```

Looks like Nix errors out anyway, and I don't see type enforcing in generators.nix. Can ask about that when I PR this to Nixpkgs.

NotAShelf (approved):
I have nothing left to add, LGTM.

One thing that might possibly come to bite you in the back is the fact that  nixpkgs requires test for new library functions, but since this'll be put in generators that might be necessary.

fufexan:
> One thing that might possibly come to bite you in the back is the fact that nixpkgs requires test for new library functions, but since this'll be put in generators that might be necessary.

Tests for these functions should be relatively easy to implement.

fufexan:
> One thing that might possibly come to bite you in the back is the fact that nixpkgs requires test for new library functions, but since this'll be put in generators that might be necessary.

Tests for these functions should be relatively easy to implement.

PR #9223, opened 2025-01-29
Extract IPC code to separate method

aaronjamt:
This makes it possible to use the same IPC code for more projects in the future.

This doesn't do much yet, but I'm planning to add IPC support to Hyprsunset, and this will be useful for that (as well as for any other Hypr* projects that need IPC in the future)

Yes, it should be ready to merge.

### DIFF
--- a/hyprctl/main.cpp
+++ b/hyprctl/main.cpp
@@ -228,7 +228,7 @@ int request(std::string arg, int minArgs = 0, bool needRoll = false) {
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


vaxerski:
yeah, but why? For now we don't

aaronjamt:
I'm not sure what you mean. I'm currently working on adding IPC support to Hyprsunset, this PR prepares for when that's actually ready to merge.

vaxerski (approved):

Issue #7621, opened 2024-09-02
Window Rule for VRR


heywatsuphello:
### Description

VRR causes noticeable flickering on OLED displays in dark content, this is mostly apparently when viewing 4:3 content with black bars in fullscreen.

I think a window rule would simplify things so people can disable VRR on their browser or media players specifically, so they don't have to keep remembering to toggle it on/off depending on what they're doing

Gliczy:
I would also like to have a window rule to enable VRR.

VRR doesn't work at all in some games without `cursor:no_break_fs_vrr`, but this same option makes the cursor laggy/stutter outside of games.

A window rule would allow us to only enable VRR for games or content that doesn't cause problems.

8Vinso8:
Lack of this stops me from using hyprland. 60 fps locked games like Factorio make cursor laggy for no benefit and in plasma i can disable vrr for them specifically.

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #9226, opened 2025-01-29
Support content-type-v1 proto

UjinT34:

- [x] add content-type-v1 proto support 
- [x] `cursor:no_break_fs_vrr` now defaults to 2 - auto: enabled for content type "game" https://github.com/hyprwm/Hyprland/issues/7621
- [x] `render:direct_scanout` can be set to 2 - auto: enabled for content type "game"
- [x] window field matching content type `content:TYPE`
- [x] window rule to set content type `content TYPE`
- [x] set drm property
- [x] check AQ version in preprocessor

`TYPE` is `none`, `photo`, `video` or `game`

If there are any clients which support this proto and set the type to "game" than new defaults might be unexpected.
Window rule is static. Any reason to make it dynamic?
Needs some testing with different AQ versions. Should be able to build with and without `output->state->setContentType` support.
Meson build is untested. 

Ready. drm part is not important, requires https://github.com/hyprwm/aquamarine/pull/140 with a version bump. Assumes that 0.7.3 will have it merged.


vaxerski:
DS is not ready for auto enabling. It has buffer release / timing issues (esp on amd)

Kommynct:
what about enabling tearing for games?

vaxerski:
we already have the tearing hint protocol for that

src/protocols/types/ContentType.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "content-type-v1.hpp"
+#include <cstdint>
+
+namespace NContentType {
+    wpContentTypeV1Type fromString(const std::string name);
```
vaxerski, line 7: I wouldn't use `wpContentTypeV1Type` for internal handling, it's generated... Better to have our own internal one and `to`/`from` fns

src/render/Renderer.cpp
```diff
@@ -1508,6 +1511,14 @@ bool CHyprRenderer::commitPendingAndDoExplicitSync(PHLMONITOR pMonitor) {
         }
     }
 
+#if AQUAMARINE_VERSION_NUMBER > 702 // >0.7.2
```
vaxerski, line 1514: uhhh what will happen if we do `.10` at some point?

UjinT34: up to .99 is fine. 100 will cause trouble.

vaxerski: oh. 7 and then 02 is the minor rev, ok

vaxerski (approved):

src/desktop/Window.hpp
```diff
@@ -19,6 +19,7 @@
 #include "WLSurface.hpp"
 #include "Workspace.hpp"
 #include "WindowRule.hpp"
+#include "protocols/types/ContentType.hpp"
```
zjeffer, line 22: My plugin doesn't build because of this line, with the following error:

```
╰─❯ make all
Hyprland headers found.
Building with XWayland support.
g++ -shared -g -fPIC --no-gnu-unique -std=c++23 `pkg-config --cflags pixman-1 libdrm hyprland` -Iinclude -DWLR_USE_UNSTABLE src/main.cpp -o split-monitor-workspaces.so
In file included from /usr/include/hyprland/src/Compositor.hpp:8,
                 from src/main.cpp:3:
/usr/include/hyprland/src/desktop/Window.hpp:22:10: fatal error: protocols/types/ContentType.hpp: No such file or directory
   22 | #include "protocols/types/ContentType.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:38: split-monitor-workspaces.so] Error 1
```
It does build if I change the line to `#include "../protocols/types/ContentType.hpp"` (which is also where the actual file is). Any ideas why this happens?

UjinT34: It might be a correct path when building from the source directory. Otherwise it's strange that CI didn't catch it. Most of the imports in this PR are generated by IDE.

zjeffer: > Most of the imports in this PR are generated by IDE.

Did you open the src/ folder in your IDE while developing this, instead of the root folder?

vaxerski: oh yeah I missed this... we can't have imports using src paths
always relative

vaxerski: fixed

zjeffer: thx

zjeffer: You missed one in `src/desktop/Window.hpp`

zjeffer: and in `src/events/Windows.cpp`
and likely some other places ;)

vaxerski: better?

zjeffer: Yes, ty

PR #9241, opened 2025-01-30
refactor: clang-tidy in compositor

Truenya:
clang-tidy fixes

nothing

ready for merging

### DIFF
--- a/src/Compositor.hpp
+++ b/src/Compositor.hpp
@@ -1,15 +1,12 @@
 #pragma once
 
-#include <list>
 #include <sys/resource.h>
 
-#include "defines.hpp"
 #include "managers/XWaylandManager.hpp"
 #include "managers/KeybindManager.hpp"
 #include "managers/SessionLockManager.hpp"
 #include "desktop/Window.hpp"
 #include "protocols/types/ColorManagement.hpp"
-#include "helpers/memory/Memory.hpp"
 
 #include <aquamarine/backend/Backend.hpp>
 #include <aquamarine/output/Output.hpp>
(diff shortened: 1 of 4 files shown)


src/Compositor.hpp
```diff
@@ -170,7 +167,7 @@ class CCompositor {
 
     uint64_t         m_iHyprlandPID    = 0;
     wl_event_source* m_critSigSource   = nullptr;
-    rlimit           m_sOriginalNofile = {0};
+    rlimit           m_sOriginalNofile = {.rlim_cur = 0};
```
vaxerski, line 170: not the same!!!!!!

Truenya: hmmmm?

first elem is rlim_cur on my machine. Or it is intentional UB here?
```
struct rlimit
  {
    /* The current (soft) limit.  */
    rlim_t rlim_cur;
    /* The hard limit.  */
    rlim_t rlim_max;
  };
```

Truenya: I just want to figure it out before making edits.

vaxerski: woops, my bad. read up and it apparently is the same, but you can also just do `{}` and it will get zeroes.

Truenya: it probably can be written as
```
    rlimit           m_sOriginalNofile;
```
in this case

Truenya: according to https://en.cppreference.com/w/cpp/language/zero_initialization

vaxerski: no, that will leave it uninitialized.

vaxerski: can you just make it {}

Truenya: oh sorry, just forgot to push

vaxerski (approved):

PR #9247, opened 2025-01-31
desktop: fix segfault when destroying a partially create layer surface

BlueZeeKing:

This fixes the bug that causes Hyprland to crash after monitors change. The bug
was caused by a layer surface being created for an invalid output, which causes
the constructor to fail early. This left the fade in animation variable
uninitialized, so when the surface was destroyed and the variable accessed,
Hyprland segfaulted. 

No

No


vaxerski (approved):
thanks!

PR #9297, opened 2025-02-02
Core: Add a test suite

vaxerski:
Currently we only test whether hyprland compiles, but we don't actually test... well, anything else.

This MR is a WIP that adds a test suite.

For now it runs, works, and tests spawing two kitties with smart gaps, that's about it.

TODOs:
 - [ ] Test... well, more.
 - [ ] Add a plugin that will be built alongside to allow for simulating keystrokes mouse strokes and other stuff
 - [ ] consider thinking about visual tests?

@fufexan nix, meson, ci pls? Tester _needs_ to be ran from `/hyprtester`, btw.

cc @PaideiaDilemma @gulafaran

### DIFF
--- a/hyprtester/src/hyprctlCompat.cpp
+++ b/hyprtester/src/hyprctlCompat.cpp
@@ -1,0 +1,137 @@
+#include "hyprctlCompat.hpp"
+#include "shared.hpp"
+
+#include <pwd.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/un.h>
+#include <unistd.h>
+
+#include <filesystem>
+#include <fstream>
+#include <algorithm>
+#include <csignal>
+#include <cerrno>
+#include <print>
+
+static int getUID() {
+    const auto UID   = getuid();
+    const auto PWUID = getpwuid(UID);
+    return PWUID ? PWUID->pw_uid : UID;
+}
+
+static std::string getRuntimeDir() {
+    const auto XDG = getenv("XDG_RUNTIME_DIR");
+
+    if (!XDG) {
+        const std::string USERID = std::to_string(getUID());
+        return "/run/user/" + USERID + "/hypr";
+    }
+
+    return std::string{XDG} + "/hypr";
+}
+
+std::vector<SInstanceData> instances() {
+    std::vector<SInstanceData> result;
+
+    try {
+        if (!std::filesystem::exists(getRuntimeDir()))
+            return {};
+    } catch (std::exception& e) { return {}; }
+
+    for (const auto& el : std::filesystem::directory_iterator(getRuntimeDir())) {
+        if (!el.is_directory() || !std::filesystem::exists(el.path().string() + "/hyprland.lock"))
+            continue;
+
+        // read lock
+        SInstanceData* data = &result.emplace_back();
+        data->id            = el.path().filename().string();
+
+        try {
+            data->time = std::stoull(data->id.substr(data->id.find_first_of('_') + 1, data->id.find_last_of('_') - (data->id.find_first_of('_') + 1)));
+        } catch (std::exception& e) { continue; }
+
+        // read file
+        std::ifstream ifs(el.path().string() + "/hyprland.lock");
+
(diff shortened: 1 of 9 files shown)


fufexan:
Good idea, but I suggest we use https://nix.dev/tutorials/nixos/integration-testing-using-virtual-machines.html. These are much simpler to use since we don't need to build our own testing framework, and there are plenty of examples to help.

I'll try to cook something up today or tomorrow if I have time.

vaxerski:
what?

hyprtester/src/hyprctlCompat.cpp
```diff
@@ -0,0 +1,136 @@
...
+                    data->pid = std::stoull(line);
+                } catch (std::exception& e) { continue; }
+            } else if (i == 1) {
+                data->wlSocket = line;
+            } else
+                break;
+        }
+
+        ifs.close();
+    }
```
nnyyxxxx, line 71: redundant?

nnyyxxxx: ifs.close()

nnyyxxxx: can u flatten this shit

nnyyxxxx: hard to follow

nnyyxxxx: why are we using a manual counter for this?

nnyyxxxx:
there are a lot of magic numbers, should they be extracted?

hyprtester/src/hyprctlCompat.cpp
vaxerski: c+p from hyprctl

nnyyxxxx: top tier laziness @vaxerski

vaxerski: that's what programming is all about

PaideiaDilemma:
This is cool ; )

Do we want something like `swapConfig(file)` in hyprtester, so that the same test can be executed with different configurations, or do we just change config options via IPC?

Also is there a reason why we don't want to allow passing the path to Hyprland binary and a path to the test suite directory right away?

nnyyxxxx:
swap config sounds kinda sus ngl

nnyyxxxx (commented):
why cant we just use the config from the examples, everything else can be done via ipc like paideia mentioned

https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.conf

vaxerski:
> Do we want something like swapConfig(file) in hyprtester, so that the same test can be executed with different configurations, or do we just change config options via IPC?

Probably not, use the config which is in the tester directory and the rest can be IPC

> Also is there a reason why we don't want to allow passing the path to Hyprland binary and a path to the test suite directory right away?

No, just laziness.

vaxerski:
> why cant we just use the config from the examples, everything else can be done via ipc like paideia mentioned

easier and faster to just append a few things to a special config file

mhydock:
These tests look like they're behavior/integration tests, which are undoubtedly valuable, but is there any interest in adding unit tests? Additionally, instead of hand-rolling a test framework, perhaps consider using an already established test framework? I personally like the look of [Catch2](https://github.com/catchorg/Catch2) (at least for unit tests), but I'm ok with anything as long as it a) has mass appeal b) autowires tests instead of needing to explicitly list them and c) is easy to read.

fufexan:
The current impl doesn't pass tests because the plugin does not load in Hyprland when built in a clean repo. A version mismatch. Though it loads fine when built in a dirty repo. If anyone with Nix knowledge can look into this, I'd be grateful, as I don't have time to debug this in a timely manner.

vaxerski:
@fufexan can you fix the nix conflicts

vaxerski:
we got some tests in, they seem to be working reliably. More tests will be added as time goes on. As a baseline, this framework LGTM

malikwirin:
A new testing framework called hyprtest??
When can I replace Gtests with it in my projects?

PR #9317, opened 2025-02-04
Add new config value "misc:splashes_file" to allow users to create their own splash mesages.

ljfp:

> "stop playing league loser" - hyprBot

I don't play league, I play Dota.

If you're like me and don't like being berated by hyprBot, this patch will allow you to create your own splash messages.
You just need to add the following line to your hyprland.conf file: 
`misc:splashes_file = /path/to/your/hypr/splashes/file`

And, of course, create a file in said path with a list of splashes (one per line) ending with a newline character.

This annoys me a lot:
```
    void                                  initAllSignals();
    void                                  removeAllSignals();
    void                                  cleanEnvironment();
    void                                  setRandomSplash();
    void                                  initManagers(eManagersInitStage stage);
    void                                  prepareFallbackOutput();
    static std::vector<std::string>       loadCustomSplashes(const std::string&);
```
But I figured it was better than making the function inline. If you have a better idea, I'm open to change it.

There's room for improvement but it can be merged.


vaxerski:
already has been rejected before

Eclextic:
> already has been rejected before

But why? I mean it is actually a smart solution for those who want their own splash messages... And we keep the default ones, when the file is not created.

Eclextic:
Just a disclaimer: I use wpaper so I don't see them anyways, but I mean, more choice no?

Issue #9331, opened 2025-02-05
direct_scanout=2 crashes


zakk4223:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

Not a regression - it's bug regarding new feature

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.47.0 built from branch main at commit 3b99e906df8b439d65e740301940e57efc057012  (compositor: don't iterate over unmapped ls-es in vectorToLS).
Date: Tue Feb 4 05:18:08 2025
Tag: v0.47.0-45-g3b99e906, commits: 5787
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set


System Information:
System name: Linux
Node name: blkheart
Release: 6.13.1-1-cachyos-eevdf
Version: #1 SMP PREEMPT_DYNAMIC Mon, 03 Feb 2025 21:57:32 +0000


GPU information: 
0d:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 21 [Radeon RX 6800/6800 XT / 6900 XT] [1002:73bf] (rev c0) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Description

If you set direct_scanout=2 hyprland crashes. It will then crash on startup until you change the value of direct_scanout to 0 or 1.

### How to reproduce

Just setup direct_scanout=2

### Attach not paste

- [x] I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam

### Checklist of files to include below

- [ ] Hyprland config - `hyprctl systeminfo -c` (always include)
- [ ] Crash report (always include in case of crash)
- [ ] Video (always include in case of a visual bug)
[...]

Issue #9230, opened 2025-01-30
VRR Pink Screen Flicker


jncsotk30:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

I don't know, I started using Hyprland only recently

### System Info and Hyprland Version

<details> 

<summary>System/Version info</summary>

```
❯ Hyprland --version
Hyprland 0.47.0 built from branch  at commit 2a478c30cab89c8c608936ca966155da5b16728d  ().
Date: 2025-01-27
Tag: v0.47.0, commits:
built against:
 aquamarine 0.7.1
 hyprlang 0.6.0
 hyprutils 0.4.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set
```

```
Linux <redacted> 6.12.7 #1-NixOS SMP PREEMPT_DYNAMIC Fri Dec 27 13:02:20 UTC 2024 x86_64 GNU/Linux
```

System is a Zephyrus G14 2024 with an `NVIDIA GeForce RTX 4070 Max-Q / Mobile` GPU. It's running on a 4K monitor (`3840x2160`) at 240Hz with `1.333333` scaling.

</details>

### Description

When setting the `vrr` property to `2` and rebuilding config (NixOS), if I switch between workspaces quickly, there is an obvious pink flash artifacting seen. If I disable `vrr` by setting to `0`, the issues goes away. See video clip for example. 

**WARNING FOR EPILEPTICS**: Video clip contains quick flashing changes on screen.

### How to reproduce

1. Set `vrr` property to `2`
2. Reload the config
3. Quickly switch between workspaces with open clients.
4. Set `vrr` to `0` to eliminate issue.

### Attach not paste

- [x] I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam

### Checklist of files to include below

- [ ] Hyprland config - `hyprctl systeminfo -c` (always include)
- [ ] Crash report (always include in case of crash)
- [x] Video (always include in case of a visual bug)
- [ ] Logs (might contain useful info such as errors)

### Additional info & File uploads

https://github.com/user-attachments/assets/dbb9f3c9-08f1-4cce-9dfc-2fdb88502238

jncsotk30:
[2025-01-29_hyprland-config.txt](https://github.com/user-attachments/files/18595791/2025-01-29_hyprland-config.txt)

UjinT34:
[...]

PR #9332, opened 2025-02-05
fix(render): add null checks to prevent crashes

nnyyxxxx:
fixes #9331

PR #9334, opened 2025-02-05
Fix DS and VRR automation

UjinT34:
Fix DS automation conditions. Fixes https://github.com/hyprwm/Hyprland/issues/9331
Fix VRR auto activation. Might fix https://github.com/hyprwm/Hyprland/issues/9230
Add `misc:vrr = 3` to auto enable VRR for game or video content types. Can be used instead of a window rule https://github.com/hyprwm/Hyprland/issues/7621

There is no need to commit VRR state and deal with explicit sync inside `ensureVRR`. Next rendered frame should pickup new desired state. Legacy renderer is untested.
Needs more testing with `vrr = 2` or `vrr = 3` and monitors without VRR support.

Ready

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -1216,7 +1216,8 @@ void CHyprRenderer::renderMonitor(PHLMONITOR pMonitor) {
     pMonitor->tearingState.activelyTearing = shouldTear;
 
     if ((*PDIRECTSCANOUT == 1 ||
-         (*PDIRECTSCANOUT == 2 && pMonitor->activeWorkspace->getFullscreenWindow() && pMonitor->activeWorkspace->getFullscreenWindow()->getContentType() == CONTENT_TYPE_GAME)) &&
+         (*PDIRECTSCANOUT == 2 && pMonitor->activeWorkspace && pMonitor->activeWorkspace->m_bHasFullscreenWindow &&
+          pMonitor->activeWorkspace->m_efFullscreenMode == FSMODE_FULLSCREEN && pMonitor->activeWorkspace->getFullscreenWindow()->getContentType() == CONTENT_TYPE_GAME)) &&
         !shouldTear) {
         if (pMonitor->attemptDirectScanout()) {
             return;
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1054,9 +1054,9 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     },
     SConfigOptionDescription{
         .value       = "misc:vrr",
-        .description = "	controls the VRR (Adaptive Sync) of your monitors. 0 - off, 1 - on, 2 - fullscreen only [0/1/2]",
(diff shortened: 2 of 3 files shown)


src/render/Renderer.cpp
```diff
@@ -1196,7 +1194,10 @@ void CHyprRenderer::renderMonitor(PHLMONITOR pMonitor) {
 
     pMonitor->tearingState.activelyTearing = shouldTear;
 
-    if ((*PDIRECTSCANOUT == 1 || (*PDIRECTSCANOUT == 2 && pMonitor->activeWorkspace->getFullscreenWindow()->getContentType() == CONTENT_TYPE_GAME)) && !shouldTear) {
+    if ((*PDIRECTSCANOUT == 1 ||
+         (*PDIRECTSCANOUT == 2 && pMonitor->activeWorkspace && pMonitor->activeWorkspace->m_bHasFullscreenWindow &&
+          pMonitor->activeWorkspace->m_efFullscreenMode == FSMODE_FULLSCREEN && pMonitor->activeWorkspace->getFullscreenWindow()->getContentType() == CONTENT_TYPE_GAME)) &&
+        !shouldTear) {
```
nnyyxxxx, line 1200: https://github.com/hyprwm/Hyprland/pull/9332/

dupes fixes from 9332?

nnyyxxxx:
might wanna drop DS changes

src/render/Renderer.cpp
UjinT34: Slightly different check. `getFullscreenWindow` might return a window with a fake fullscreen mode. This checks for a real one.

nnyyxxxx: does it matter if its ffm? seems redundant since if something is trying to request fullscreen its most likely doing it correctly; what if a window is fullscreen but not FSMODE_FULLSCREEN, maybe we still want DS for performance?

nnyyxxxx: arent DS requirements for buffer requirements and state not content?

nnyyxxxx: https://github.com/hyprwm/Hyprland/blob/84c9baecc6b73f6321b2739061e3c232710f73bf/src/helpers/Monitor.cpp#L1266

nnyyxxxx: fn for reference

UjinT34: ds = 2 is for content

nnyyxxxx: > ds = 2 is for content

the check for games is just a policy decision on top of the other requirements, so in turn its only **WHEN** direct scanout should be used not if it can technically work

nnyyxxxx: the solitary checks are already handling this

zakk4223:
Isn't this all supposed to be gated behind the window being *solitary* anyways?  DS requires a solitary window on the monitor and I assume so should vrr. PMONITOR->solitaryClient

UjinT34:
> Isn't this all supposed to be gated behind the window being _solitary_ anyways? DS requires a solitary window on the monitor and I assume so should vrr. PMONITOR->solitaryClient

`direct_scanout = 1` works this way.
`direct_scanout = 2` also checks for a fullscreen window with game content type to avoid DS activation for apps that don't work well with DS.

gulafaran:
needs a rebase

vaxerski (approved):

PR #9343, opened 2025-02-06
toplevelexport: fix transformed origin for shm buffers

WhySoBad:

When using a shm buffer for the toplevel export of a window which is on a transformed output, the origin for copying the frame into the buffer is not transformed which can produce black spots or entirely black images. I've implemented the origin transformations for all allowed output transformations.

I don't know whether it's intended that frames captured using the toplevel export protocol do not respect transformations of the outputs their on but since I've found this code I assume it isn't intended:

https://github.com/hyprwm/Hyprland/blob/8a6778f0a087cdfc4bc1d3751b0be2c2bf3322aa/src/protocols/ToplevelExport.cpp#L122-L124

however, since `pWindow->m_vRealSize` is the size of the window with transformations taken into account, transforming the box can cause undesired box dimensions (in my case of `WL_OUTPUT_TRANSFORM_270` the height and width are swapped). 

In my opinion there are two possible solutions to "fixing" this:
1. **Inform the client about the transformation**: In this case, we would ideally change the `hyprland_toplevel_export_frame_v1` protocol to inform the client (in the buffer event?) about the transformation of the output on which the window is, such that the client can implement the transformation itself. Alternatively, there is already the option for the client to get the output for the window by it's address handle using e.g. the hyprland socket and read the transformation from there. In this case the problem is already kind of solved.

2. **Transform the buffer on the server**: Since the `renderWindow` method doesn't care about the transformation of the output on which the window is, we would need to read the pixels into a temporary buffer from which we can transform them accordingly. For the dma buffer something similar would need to be implemented

Let me know if I missed something or got something wrong. 

I think solving the transformation problem (should it even be one) would be better off in a separate issue. In this case this pull request is ready for merging

### DIFF
--- a/src/protocols/ToplevelExport.cpp
+++ b/src/protocols/ToplevelExport.cpp
@@ -289,7 +289,30 @@ bool CToplevelExportFrame::copyShm(timespec* now) {
     glPixelStorei(GL_PACK_ALIGNMENT, 1);
 
     auto glFormat = PFORMAT->flipRB ? GL_BGRA_EXT : GL_RGBA;
-    glReadPixels(0, 0, box.width, box.height, glFormat, PFORMAT->glType, pixelData);
+
+    int  origin_x = 0;
+    int  origin_y = 0;
+    switch (PMONITOR->transform) {
+        case WL_OUTPUT_TRANSFORM_FLIPPED_180:
+        case WL_OUTPUT_TRANSFORM_90: {
+            origin_y = PMONITOR->vecPixelSize.y - box.height;
+            break;
+        }
+        case WL_OUTPUT_TRANSFORM_FLIPPED_270:
+        case WL_OUTPUT_TRANSFORM_180: {
+            origin_x = PMONITOR->vecPixelSize.x - box.width;
+            origin_y = PMONITOR->vecPixelSize.y - box.height;
+            break;
+        }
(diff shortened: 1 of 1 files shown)


src/protocols/ToplevelExport.cpp
```diff
@@ -289,7 +289,30 @@ bool CToplevelExportFrame::copyShm(timespec* now) {
     glPixelStorei(GL_PACK_ALIGNMENT, 1);
 
     auto glFormat = PFORMAT->flipRB ? GL_BGRA_EXT : GL_RGBA;
-    glReadPixels(0, 0, box.width, box.height, glFormat, PFORMAT->glType, pixelData);
+
+    int  origin_x = 0;
```
vaxerski, line 293: why not use a vec2d?

also no snake

WhySoBad: done

vaxerski (approved):
thanaks!

WhySoBad:
What are your thoughts about applying output transformations to the buffers returned by the toplevel export? 
Is it intended to to not transform the buffers and not inform the client about any transformations or should I create a new issue which addresses this problem?

vaxerski:
good question. We might want to untransform them to be always "normal"

vaxerski:
that's for a separate MR tho

Issue #4527, opened 2024-01-24
The animation when scrolling through workspace is going into the wrong direction when going between first and last workspace


Azercoco:
### Hyprland Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch HEAD at commit 03ebbe18ed8517ee22591eac82cd54322f42cb7d  (props: bump ver to 0.34.0).
Date: Mon Jan 1 12:03:15 2024
Tag: v0.34.0
```

</details>

### Bug or Regression?

Bug

### Description

When scrolling through the workspace using the `workspace, e +/- 1 `, the animation direction of going from workspace 1 to the the last is the same than going from 1 to 2. The same applies from going to the last to the first which is also reversed.

### How to reproduce

* Open at least 3 workspaces
* Scroll between the first and the last 
* The animation is reversed which is confusing for navigation

pdamianik:
This animation direction is intentional, as the last workspace is in the same direction as the second workspace from the perspective of the first workspace.
![image](https://github.com/hyprwm/Hyprland/assets/39028343/71fda825-0b3b-4d22-81e6-e0cb5c12c238)

As far as I understood you want the option to have the following behavior instead:
![image](https://github.com/hyprwm/Hyprland/assets/39028343/a1b277f8-278a-4e92-ae6c-6155e7e6e475)

The relevant code section determining the direction of this animation is in [`src/helpers/Monitor.cpp:527`](https://github.com/hyprwm/Hyprland/blob/334a0f03ee2f80118418c16cc06005a1fe8cfd60/src/helpers/Monitor.cpp#L527) in the `changeWorkspace` function which gets called in [`src/managers/KeybindManager.cpp:929`](https://github.com/hyprwm/Hyprland/blob/334a0f03ee2f80118418c16cc06005a1fe8cfd60/src/managers/KeybindManager.cpp#L929). I won't implement a Pull Regest for this for now as it is quite late, but I might take another look at it tomorrow.

pdamianik:
PR draft is ready: #4684. The option is named `binds:animate_workspace_band` for now. I am open for suggestions.

pdamianik:
#4684 and hyprwm/hyprland-wiki#477 are ready. I am just waiting for feedback from @Azercoco.

KoyeBosh:
I propose `animations:workspace_wraparound` instead, it's more consistent with where other workspace animation setting are configured and imo is easier to understand

pdamianik:
[...]

PR #9368, opened 2025-02-08
Update persistence on workspace rename

hariseldon78:
The persistent workspace property is defined with workspace rules. 
At the moment there is no way to remove or edit a workspace rule. One consequence of this is that when a workspace is made persistent there is no way to remove that persistent property, and thus there is no way to remove the workspace.

**This pull request reevaluate the workspace rule when a workspace is renamed**, making the workspace rule to be respected. As a side effect it allows to rename the workspace in order to remove the persistence. 

I also added a line in the workspace info to show the status of persistence for the workspace.

There are other ways to do this, I just thought this is a change little enough to not have too much code change. 
Other possible ways: 
1) implement a way to remove or edit a workspace rule
2) [probably ugly] implement a deleteworkspace which will remove the workspace even if it is persistent (but still must be empty)
3) extending this pr by enforcing all the properties of a workspace rule, not only the persistent bit. I think it would just be a matter of invoking `updateWindowData()`, but maybe optimizing a bit to only fetch the workspace rule once.

Anyway I still think that people would expect the workspace rules to be an "invariant", which is enforced at all time, not only at workspace creation

Ready and tested.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -3040,7 +3040,13 @@ void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspace
             continue;
         }
 
-        auto PWORKSPACE = pWorkspace;
+        PHLWORKSPACE PWORKSPACE = nullptr;
+        if (pWorkspace) {
+            if (pWorkspace->matchesStaticSelector(rule.workspaceString))
+                PWORKSPACE = pWorkspace;
+            else
+                continue;
+        }
 
         if (!PWORKSPACE) {
             WORKSPACEID id     = rule.workspaceId;
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -3040,11 +3040,12 @@ void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspace
             continue;
         }
 
-        auto        PWORKSPACE = pWorkspace;
+        auto PWORKSPACE = pWorkspace;
 
-        WORKSPACEID id     = rule.workspaceId;
-        std::string wsname = rule.workspaceName;
         if (!PWORKSPACE) {
+            WORKSPACEID id     = rule.workspaceId;
+            std::string wsname = rule.workspaceName;
+
             if (id == WORKSPACE_INVALID) {
                 const auto R = getWorkspaceIDNameFromString(rule.workspaceString);
                 id           = R.id;
@@ -3056,6 +3057,8 @@ void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspace
                 continue;
             }
             PWORKSPACE = getWorkspaceByID(id);
+            if (!PWORKSPACE)
+                createNewWorkspace(id, PMONITOR ? PMONITOR : m_pLastMonitor.lock(), wsname, false);
         }
         if (PWORKSPACE) {
             PWORKSPACE->m_bPersistent = true;
@@ -3069,8 +3072,7 @@ void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspace
             Debug::log(LOG, "ensurePersistentWorkspacesPresent: workspace persistent {} not on {}, moving", rule.workspaceString, PMONITOR->szName);
             moveWorkspaceToMonitor(PWORKSPACE, PMONITOR);
             continue;
-        } else
-            createNewWorkspace(id, PMONITOR ? PMONITOR : m_pLastMonitor.lock(), wsname, false);
+        }
     }
 
     // cleanup old
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -3027,7 +3027,8 @@ bool CCompositor::shouldChangePreferredImageDescription() {
     return false;
 }
 
-void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspaceRule>& rules) {
+void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspaceRule>& rules, PHLWORKSPACE pWorkspace) {
+
     for (const auto& rule : rules) {
         if (!rule.isPersistent)
             continue;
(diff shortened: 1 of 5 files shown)


src/debug/HyprCtl.cpp
```diff
@@ -319,15 +319,17 @@ std::string CHyprCtl::getWorkspaceData(PHLWORKSPACE w, eHyprCtlOutputFormat form
...
     "hasfullscreen": {},
     "lastwindow": "0x{:x}",
-    "lastwindowtitle": "{}"
+    "lastwindowtitle": "{}",
+    "ispersistent": {}
 }})#",
                            w->m_iID, escapeJSONStrings(w->m_szName), escapeJSONStrings(PMONITOR ? PMONITOR->szName : "?"),
                            escapeJSONStrings(PMONITOR ? std::to_string(PMONITOR->ID) : "null"), w->getWindows(), ((int)w->m_bHasFullscreenWindow == 1 ? "true" : "false"),
-                           (uintptr_t)PLASTW.get(), PLASTW ? escapeJSONStrings(PLASTW->m_szTitle) : "");
+                           (uintptr_t)PLASTW.get(), PLASTW ? escapeJSONStrings(PLASTW->m_szTitle) : "", ((int)w->m_bPersistent == 1 ? "true" : "false"));
```
vaxerski, line 327: what is this goofy comparison just `w->m_bPersistent ? true : false`

src/desktop/Workspace.cpp
```diff
@@ -648,6 +648,9 @@ void CWorkspace::rename(const std::string& name) {
     Debug::log(LOG, "CWorkspace::rename: Renaming workspace {} to '{}'", m_iID, name);
     m_szName = name;
 
+    const auto WORKSPACERULE = g_pConfigManager->getWorkspaceRuleFor(m_pSelf.lock());
+    m_bPersistent            = WORKSPACERULE.isPersistent;
```
vaxerski, line 652: wouldnt it be better to call recheck persistent?

src/debug/HyprCtl.cpp
hariseldon78: You are right, i'll fix the fullscreen check too. I'm tempted to also return the booleans directly, instead of converting to int, in the non-json branch, but maybe that would break users scripts?

src/desktop/Workspace.cpp
hariseldon78: Sure, I see that it's a convention in the project. And how do you feel about also calling updateWindowData from the renaming handler?

littleblack111:
> [probably ugly] implement a deleteworkspace which will remove the workspace even if it is persistent (but still must be empty)

maybe just add an option to forcefully remove it which closes everything in the workspace

also, it'd be great if we get a `presworkspace` dispatcher to switch to other presistance workspaces(for people who prefer static workspaces) can just make like 10 presistant workspaces and will be able to scroll through it (see img below, stolen from https://github.com/hyprwm/Hyprland/issues/4527#issuecomment-1937365283) to replace `workspace +1/-1`

image

hariseldon78:
> > [probably ugly] implement a deleteworkspace 
> maybe just add an option/arg to forcefully remove it which closes everything in the workspace

This is an interesting take, although IMO dangerous, and still closing some windows is not a guaranteed action, some presents popups and some are just stuck and don't react to the kill signal. It's better to only have basic building blocks as dispatchers that can be merged by user scripts: just fetch the windows list, kill all one by one, decide what to do if they don't close, and then delete the workspace.


> also, it'd be great if we get a `persisworkspace` dispatcher to switch to other presistance workspaces(for people who prefer static workspaces) 

Again, it's easy to implement a script that will fetch the list of workspace, determine the next "persistent" one (with the debug info from this PR), and switch to it. I don't think there is the need for a specific dispatcher.

src/desktop/Workspace.cpp
```diff
@@ -648,6 +653,8 @@ void CWorkspace::rename(const std::string& name) {
     Debug::log(LOG, "CWorkspace::rename: Renaming workspace {} to '{}'", m_iID, name);
     m_szName = name;
 
+    recheckPersistent();
```
vaxerski, line 656: I meant `CCompositor::ensurePersistentWorkspacesPresent`

hariseldon78: No, that method does not touch the workspace persistent property in any way. If i call it without calling "recheckPersistent" the workspace stays alive after renaming, even when empty and no persistent rule on its name.

vaxerski: well then it should, no?

hariseldon78: hmm, at a first look it would be extremely inefficient: that method doesn't not know which workspace have changed, so it needs to check all of them, against each rule, and we already know there is only one change, so it becomes o(rules * workspaces) instead of o(rules).

we could mitigate this by passing an optional workspace argument, so the method can restrict the check only to it. do you like this idea?

vaxerski: that is fine ye

hariseldon78:
ok, i think i got something decent: 
- i still had to do the `m_bPersistent = WORKSPACERULE.isPersistent` at the rename method, to make it false if the rule is not persistent anymore
- by calling the ensurePersistentWorkspacesPresent we make sure the renamed workspace move to the correct monitor if needed
- the ensurePersistentWorkspacesPresent method became a bit more complex to handle the passed optional workspace, but I think it's a reasonable compromise

src/Compositor.cpp
```diff
@@ -3039,31 +3040,39 @@ void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspace
...
         }
 
-        WORKSPACEID id     = rule.workspaceId;
-        std::string wsname = rule.workspaceName;
-        if (id == WORKSPACE_INVALID) {
-            const auto R = getWorkspaceIDNameFromString(rule.workspaceString);
-            id           = R.id;
-            wsname       = R.name;
-        }
+        auto PWORKSPACE = pWorkspace;
```
vaxerski, line 3043: I don't think this is correct?

Let's say we want to recheck Workspace `1`, so pWorkspace is `1`

Then, we have a rule
```
workspace = 2, persistent:1, monitor:DP-2
```

we get to this spot with our rule, but set `1`'s persistent to true...

hariseldon78: I am calling this method passing a single rule that i already derived from the workspace from Workspace.cpp:651 'const auto WORKSPACERULE = g_pConfigManager->getWorkspaceRuleFor(m_pSelf.lock());'. 
So i'm already sure that the workspace match the rule, but I agree with you that the method signature could be misleading without this info. I'll add a check to only apply the matching rules.

src/Compositor.cpp
```diff
@@ -3039,31 +3040,45 @@ void CCompositor::ensurePersistentWorkspacesPresent(const std::vector<SWorkspace
...
         }
 
-        WORKSPACEID id     = rule.workspaceId;
-        std::string wsname = rule.workspaceName;
-        if (id == WORKSPACE_INVALID) {
-            const auto R = getWorkspaceIDNameFromString(rule.workspaceString);
-            id           = R.id;
-            wsname       = R.name;
+        PHLWORKSPACE PWORKSPACE = nullptr;
+        if (pWorkspace) {
```
vaxerski, line 3044: this check can be moved above PMONITOR finding a monitor for better optimization

vaxerski (commented):
rest looks ok

src/Compositor.cpp
```diff
@@ -3027,43 +3027,60 @@ bool CCompositor::shouldChangePreferredImageDescription() {
...
+            }
+            PWORKSPACE = getWorkspaceByID(id);
+            if (!PWORKSPACE)
+                createNewWorkspace(id, PMONITOR ? PMONITOR : m_pLastMonitor.lock(), wsname, false);
         }
 
-        if (id == WORKSPACE_INVALID) {
-            Debug::log(ERR, "ensurePersistentWorkspacesPresent: couldn't resolve id for workspace {}", rule.workspaceString);
+        if (PWORKSPACE)
+            PWORKSPACE->m_bPersistent = true;
```
hariseldon78, line 3066: i had to isolate this check and do it before handling the monitor to also set the bit correctly for rules not setting a target monitor.

vaxerski (approved):

vaxerski:
tanks!

PR #9398, opened 2025-02-13
fix(render): disable explicit if aquamarine output doesn't support it

### LINKED ISSUE #9396: hyprland needs to check DRM_CAP_SYNCOBJ_TIMELINE
### Already reported ? *

### Regression?

Not a regression - it's bug regarding new feature

current ToT, on any driver that doesn't support `DRM_CAP_SYNCOBJ_TIMELINE`

### Description

AFAICT hyprland isnt' even checking for basic syncobj support (`DRM_CAP_SYNCOBJ`) let alone timeline semaphore support (`DRM_CAP_SYNCOBJ_TIMELINE`), which is a requirement for `linux-drm-syncobj-v1`.  You should be checking along the lines of:
```c
   uint64_t cap;
   int ret = drmGetCap(drm_fd, DRM_CAP_SYNCOBJ_TIMELINE, &cap);
   supported = cap && !ret;
```

see: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/90#note_2776458

### How to reproduce

Probably removing `DRIVER_SYNCOBJ_TIMELINE` from `drm_driver::driver_features` on the kernel side for any driver would be enough to reproduce the unsynchronized rendering

### Attach not paste

- [ ] Hyprland config - `hyprctl systeminfo -c` (always include)
- [ ] Logs (might contain useful info such as errors)

_No response_

aruhier:

Fixes #9396.

The explicit settings ignore the aquamarine output.supportsExplicit attribute, which creates glitches on drivers not supporting explicit sync (example: freedreno).

If the output has been set as not supporting explicit, disable the explicit settings.

Ready to merge.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1312,7 +1312,7 @@ bool CMonitor::attemptDirectScanout() {
         return false;
     }
 
-    auto explicitOptions = g_pHyprRenderer->getExplicitSyncSettings();
+    auto explicitOptions = g_pHyprRenderer->getExplicitSyncSettings(output);
 
     // wait for the explicit fence if present, and if kms explicit is allowed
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
thanks!

PR #9422, opened 2025-02-16
feat(window): add option to allow size persistence between app launches

### LINKED ISSUE #9171: Have option to not tell windows they are maximized
### Description

A common issue with Hyprland is that some windows don't remember their size. That is, if you use windows in floating mode and adjust them to some size, then after restarting they _should_ open with the same size again (because they have internal mechanisms to remember the size and then give a size hint to Hyprland), but they don't.

The reason for this is that Hyprland tells windows always that they are maximized, and then the windows don't bother remembering their size. The reason that Hyprland does this is to force windows into not drawing e.g. shadows.

However, not having windows remember their size is a significant drawback, especially given that for many windows the shadow thing _isn't even a problem_, because they wouldn't draw any decorations anyway. Other windows managers (e.g. river) also don't do this maximizing "trick".

So I propose to add a window rule, which would allow the user to disable this tell-them-they-are-maximized-trick for certain windows.

Perhaps one could even add an option to set the default, so that one could set to default to "disable the trick", but then enable it for some special windows that draw otherwise would draw shadows.

nnyyxxxx:
fixes #9171

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -2952,12 +2952,12 @@ void CConfigManager::ensurePersistentWorkspacesPresent() {
 
 void CConfigManager::storeFloatingSize(const std::string& szClass, const std::string& szTitle, const Vector2D& size) {
     Debug::log(LOG, "storing floating size {}x{} for window {}::{}", size.x, size.y, szClass, szTitle);
-    FloatCache id{szClass, szTitle};
+    SFloatCache id{szClass, szTitle};
     m_mStoredFloatingSizes[id] = size;
 }
 
 std::optional<Vector2D> CConfigManager::getStoredFloatingSize(const std::string& szClass, const std::string& szTitle) {
-    FloatCache id{szClass, szTitle};
+    SFloatCache id{szClass, szTitle};
     if (m_mStoredFloatingSizes.contains(id)) {
         Debug::log(LOG, "got stored size {}x{} for window {}::{}", m_mStoredFloatingSizes[id].x, m_mStoredFloatingSizes[id].y, szClass, szTitle);
         return m_mStoredFloatingSizes[id];
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -2951,14 +2951,15 @@ void CConfigManager::ensurePersistentWorkspacesPresent() {
 
 void CConfigManager::storeFloatingSize(const std::string& szClass, const std::string& szTitle, const Vector2D& size) {
     Debug::log(LOG, "storing floating size {}x{} for window {}::{}", size.x, size.y, szClass, szTitle);
-    m_mStoredFloatingSizes[szClass + "|||" + szTitle] = size;
+    FloatCache id{szClass, szTitle};
+    m_mStoredFloatingSizes[id] = size;
 }
 
 std::optional<Vector2D> CConfigManager::getStoredFloatingSize(const std::string& szClass, const std::string& szTitle) {
-    const auto KEY = szClass + "|||" + szTitle;
-    if (m_mStoredFloatingSizes.contains(KEY)) {
-        Debug::log(LOG, "got stored size {}x{} for window {}::{}", m_mStoredFloatingSizes[KEY].x, m_mStoredFloatingSizes[KEY].y, szClass, szTitle);
-        return m_mStoredFloatingSizes[KEY];
+    FloatCache id{szClass, szTitle};
+    if (m_mStoredFloatingSizes.contains(id)) {
+        Debug::log(LOG, "got stored size {}x{} for window {}::{}", m_mStoredFloatingSizes[id].x, m_mStoredFloatingSizes[id].y, szClass, szTitle);
+        return m_mStoredFloatingSizes[id];
     }
     return std::nullopt;
 }
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -2950,16 +2950,16 @@ void CConfigManager::ensurePersistentWorkspacesPresent() {
     g_pCompositor->ensurePersistentWorkspacesPresent(m_vWorkspaceRules);
 }
 
-void CConfigManager::storeFloatingSize(const std::string& szClass, const std::string& szTitle, const Vector2D& size) {
-    Debug::log(LOG, "storing floating size {}x{} for window {}::{}", size.x, size.y, szClass, szTitle);
-    SFloatCache id{szClass, szTitle};
+void CConfigManager::storeFloatingSize(PHLWINDOW window, const Vector2D& size) {
+    Debug::log(LOG, "storing floating size {}x{} for window {}::{}", size.x, size.y, window->m_szClass, window->m_szTitle);
+    SFloatCache id{window};
     m_mStoredFloatingSizes[id] = size;
 }
 
-std::optional<Vector2D> CConfigManager::getStoredFloatingSize(const std::string& szClass, const std::string& szTitle) {
-    SFloatCache id{szClass, szTitle};
+std::optional<Vector2D> CConfigManager::getStoredFloatingSize(PHLWINDOW window) {
+    SFloatCache id{window};
     if (m_mStoredFloatingSizes.contains(id)) {
-        Debug::log(LOG, "got stored size {}x{} for window {}::{}", m_mStoredFloatingSizes[id].x, m_mStoredFloatingSizes[id].y, szClass, szTitle);
+        Debug::log(LOG, "got stored size {}x{} for window {}::{}", m_mStoredFloatingSizes[id].x, m_mStoredFloatingSizes[id].y, window->m_szClass, window->m_szTitle);
         return m_mStoredFloatingSizes[id];
     }
     return std::nullopt;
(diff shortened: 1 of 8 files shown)


nnyyxxxx:
should this be per window or windows in general

vaxerski:
windowrule, preferably

nnyyxxxx:
> windowrule, preferably

okay...

nnyyxxxx:
not sure this is possible

ive tried some shit but apps start flipping out with their scaling and shiz

nnyyxxxx:
can you give me opinions on this @vaxerski

littleblack111:
can't you just gate `changeWindowFullscreenModeClient` or `setWindowFullscreenClient` from actually executing, with the windowrule?

nnyyxxxx:
if we stop telling apps they are maximized then some apps like firefox dont scale properly

nnyyxxxx:
and they dont even remember their previous floating size

vaxerski:
> can you give me opinions on this @vaxerski

:+1: 

> not sure this is possible

should easily be

nnyyxxxx:
> if we stop telling apps they are maximized then some apps like firefox dont scale properly

what about this

vaxerski:
might be our UV calculations not liking CSD.

Some apps, notably gtk ones, will draw csd no matter what unless they are maximized (because then the protocol doesn't allow them)

I would love if the xdg_shell spec said the same about tiled modes but it doesn't and I can't suggest the change due to being banned.

nnyyxxxx:
> I can't suggest the change due to being banned.

damn bro is banned :skull: thats crazy...

littleblack111:
> might be our UV calculations not liking CSD.
> 
> Some apps, notably gtk ones, will draw csd no matter what unless they are maximized (because then the protocol doesn't allow them)
> 
> I would love if the xdg_shell spec said the same about tiled modes but it doesn't and I can't suggest the change due to being banned.

freedesktop shenanigans

nnyyxxxx:
wiki : https://github.com/hyprwm/hyprland-wiki/pull/983

src/desktop/WindowRule.cpp
```diff
@@ -29,6 +29,8 @@ CWindowRule::CWindowRule(const std::string& rule, const std::string& value, bool
         ruleType = RULE_FULLSCREEN;
     else if (rule == "maximize")
         ruleType = RULE_MAXIMIZE;
+    else if (rule == "disablemaximizedhints")
```
littleblack111, line 32: nomaximizehints maybe fit better? (as we have noinitialfocus)

littleblack111 (approved):
👍

src/desktop/WindowRule.cpp
nnyyxxxx: ok

nnyyxxxx: done [5f95671](https://github.com/hyprwm/Hyprland/pull/9422/commits/5f95671e81565eb872cdb0b11c99ae4bab93f8a7)

littleblack111: resolve it :)

nnyyxxxx:
we could do this for x windows as well, though i dont know if that will blow some app up lol or if its even necessary

vaxerski (commented):
I don't see where this disables the maximized hints. It doesn't, in fact.

nnyyxxxx:
I'll just rename the window rule

littleblack111:
> > I don't see where this disables the maximized hints. It doesn't, in fact.
> 
> yeah actually we dont need to disable maximize hints, if we do, some apps like firefox wont scale properly.
> 
> if we do it this way, apps like firefox will scale properly, and all apps will remember their size after app restart.
> 
> do you want me to rename the window rule to somethin else?

wdym by not scaling properly? do they try to have a full-screen scale? can you maybe lie to them about what is the geometry of that too?

nnyyxxxx:
i would love to send an image for reference but latest -git keeps crashing for me whenever i open a window.. does it crash for you as well? @littleblack111

nnyyxxxx:
nvm must just be me

nnyyxxxx:
here's that screenshot, if we dont send a hint, the app doesnt scale. unless im just stupid

image

littleblack111:
> here's that screenshot, if we dont send a hint, the app doesnt scale. unless im just stupid
> 
> image

cant you send a resize event or sth to force the window to resize/scale to the fullscreen size

nnyyxxxx:
nah its just stuck like that

nnyyxxxx:
> it will not work across reboots, though I don't really have a great solution for that.

wait what :skull: what do you mean? this is just session based :skull: 

bros expecting too much of me

nnyyxxxx:
i dont even know how we could make it persist across reboots. maybe manually storing values in a file in .cache/hyprland or somethin and read off of that

vaxerski:
> i dont even know how we could make it persist across reboots. maybe manually storing values in a file in .cache/hyprland or somethin and read off of that

sqlite3 it's called but a hard dep on that would be a bit funky.

Anyways, yeah, session based better than nothing I guess

src/config/ConfigManager.cpp
```diff
@@ -2948,3 +2948,17 @@ std::string SConfigOptionDescription::jsonify() const {
 void CConfigManager::ensurePersistentWorkspacesPresent() {
     g_pCompositor->ensurePersistentWorkspacesPresent(m_vWorkspaceRules);
 }
+
+void CConfigManager::storeFloatingSize(const std::string& szClass, const std::string& szTitle, const Vector2D& size) {
```
vaxerski, line 2952: use a data struct that is constructed with a `PHLWINDOW` as a parameter, and implement std::hash for it. Then you can just store those in an unordered_map

nnyyxxxx: 61edff427531b3b57c053b0df572b3405c1a1834 is this good

src/config/ConfigManager.hpp
```diff
@@ -138,6 +138,27 @@ struct SFirstExecRequest {
...
 };
 
+struct FloatCache {
+    std::string class_name;
+    std::string title;
+
+    FloatCache(PHLWINDOW window) : class_name(window->m_szClass), title(window->m_szTitle) {}
+    FloatCache(const std::string& class_name_, const std::string& title_) : class_name(class_name_), title(title_) {}
+
+    bool operator==(const FloatCache& other) const {
```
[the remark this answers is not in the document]
nnyyxxxx: wdym? thats being used

src/config/ConfigManager.hpp
```diff
@@ -138,6 +138,27 @@ struct SFirstExecRequest {
     bool        withRules = false;
 };
 
+struct FloatCache {
+    std::string class_name;
+    std::string title;
```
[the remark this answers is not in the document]
nnyyxxxx: bros speaking some magic words that i cant quite understand right now

src/config/ConfigManager.cpp
```diff
@@ -2948,3 +2948,18 @@ std::string SConfigOptionDescription::jsonify() const {
...
```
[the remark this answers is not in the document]
nnyyxxxx: [8f5ec5a](https://github.com/hyprwm/Hyprland/pull/9422/commits/8f5ec5ad8b1bd49b99d6b84e7bbef5f476d9f985)

src/config/ConfigManager.hpp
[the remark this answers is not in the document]
nnyyxxxx: [8f5ec5a](https://github.com/hyprwm/Hyprland/pull/9422/commits/8f5ec5ad8b1bd49b99d6b84e7bbef5f476d9f985)

nnyyxxxx:
so ive found a bug with this, or is it a feature instead?

open 1 window resize it down to a size, and keep it open

open a new window, make it float, see that it has become the same size as window 1

now resize window 2 to a different size, and then reopen it.

window 2 will not keep the new size, instead it will keep window 1's size until window 1 has been closed

nnyyxxxx:
this doesnt happen for different apps. so not an issue in that department

nnyyxxxx:
the current implementation treats windows differently based on class+title, so if a user matches with an asterisk. this would become a problem in the users perspective?

nnyyxxxx:
any opinions on this "bug"?

this is ready to be merged otherwise

src/config/ConfigManager.hpp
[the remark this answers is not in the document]
vaxerski: wouldnt hash be used? eh, whatevr.

nnyyxxxx:
conflicts

nnyyxxxx:
ah yes https://github.com/hyprwm/Hyprland/actions/runs/13611130054/job/38048422431

nnyyxxxx:
ignore https://github.com/hyprwm/Hyprland/actions/runs/13615419769/job/38057849382

nnyyxxxx:
i dont think thats really fixable with our current approach

vaxerski:
yeah, the reason is I forgor

nnyyxxxx:
https://github.com/hyprwm/Hyprland/pull/9422#issuecomment-2677503965

jbirnick:
As already mentioned in #9171 , this doesn't solve the issue, and in general I think it's the wrong approach.

There should just be an option to not tell windows that they are maximized. No workaround with storing sizes or whatever.

vaxerski:
@jbirnick our uv calculations are already a complete spaghetti due to there being _so many things_ that can alter a surface's presentation wrt. the texture, and it seems that adding xdg geometries breaks the windows. 

The problem is that `TILED` does not mandate no csd, which it should, but I am not able to even suggest that be done because I am banned from fdo.

Only maximized does that, which means we have to rely on it to avoid weird textures for now.

Issue #8506, opened 2024-11-17
Flicking Camera Position In FFXIV


ToxicMushroom:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.45.0 built from branch main at commit af83c825138386d269d69b3ef755b844a2eacb22 dirty (hyprctl: add json output on hyprctl -j plugins list (8480)).
Date: Sun Nov 17 00:18:30 2024
Tag: v0.45.0-29-gaf83c825, commits: 5462
built against aquamarine 0.5.0


flags set:
debug


System Information:
System name: Linux
Node name: karamba2000
Release: 6.11.8-arch1-2
Version: #1 SMP PREEMPT_DYNAMIC Fri, 15 Nov 2024 15:35:07 +0000


GPU information: 
07:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6650 XT / 6700S / 6800S] [1002:73ef] (rev c1) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/merlijn/.config/hypr/hyprlandd.conf: Read Succeeded
########################################################################################

#
# Please note not all available settings / options are set here.
# For a full list, see the wiki
#

debug {
    disable_logs = false
}

# See https://wiki.hyprland.org/Configuring/Monitors/
monitor=DP-1,1920x1080@144,0x0,1
monitor=DP-2,1920x1080@144,1920x0,1

# See https://wiki.hyprland.org/Configuring/Keywords/ for more

# Execute your favorite apps at launch
exec-once = ironbar & hyprpaper

# Source a file (multi-file configs)
# source = ~/.config/hypr/myColors.conf

# Some default env vars.
[...]

PR #9427, opened 2025-02-17
fix PointerConstraints activation occurring before attached to InputManager

bk304:

Fixes  [#8506](https://github.com/hyprwm/Hyprland/issues/8506)

It is possible for the CInputManager::simulateMouseMovement() method to be called via CPointerConstraint::activate() without the constraint being present in InputManager.m_vConstraints.

This occurs when the constraint is activated before CPointerConstraintsProtocol::onNewConstraint is called, which is what places it in InputManager.m_vConstraints.

The lack of the instance of CPointerConstraint in m_vConstraints causes the CInputManager::isConstrained() method to fail and return false. As a result, the CInputManager::mouseMoveUnified method is run erroneously until it's end when the constraint is activated.

This does not seem to be very common behavior, but the FFXIV game running via wine and xwayland can trigger this.

As a solution, I found it ideal to postpone the activation of the constraint until after the onNewConstraint call.

This change assumes that it is true that a constraint cannot be activated before it is bound to the Input Manager. I don't see any sense in which this could be wrong, but it would be good to have a sanity check.

Later I will test other games that restrict the mouse to see if anything obviously broke.

Ready for merging.

### DIFF
--- a/src/protocols/PointerConstraints.cpp
+++ b/src/protocols/PointerConstraints.cpp
@@ -111,7 +111,8 @@ void CPointerConstraint::deactivate() {
     else
         resourceC->sendUnconfined();
 
-    active = false;
+    pendingActivation = false;
+    active            = false;
 
     if (lifetime == ZWP_POINTER_CONSTRAINTS_V1_LIFETIME_ONESHOT) {
         dead = true;
@@ -127,7 +128,13 @@ void CPointerConstraint::activate() {
     if (dead || active)
         return;
 
-    // TODO: hack, probably not a super duper great idea
+    if (!attached) {
+        pendingActivation = true;
+        return;
+    }
+    pendingActivation = false;
+
+    //TODO: hack, probably not a super duper great idea
     if (g_pSeatManager->state.pointerFocus != pHLSurface->resource()) {
         const auto SURFBOX = pHLSurface->getSurfaceBoxGlobal();
         const auto LOCAL   = SURFBOX.has_value() ? logicPositionHint() - SURFBOX->pos() : Vector2D{};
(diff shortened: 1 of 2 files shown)


vaxerski:
> This occurs when the constraint is activated before CPointerConstraintsProtocol::onNewConstraint is called, which is what places it in InputManager.m_vConstraints.

how can a constraint send activate if it's not created yet?

bk304:
> > This occurs when the constraint is activated before CPointerConstraintsProtocol::onNewConstraint is called, which is what places it in InputManager.m_vConstraints.
> 
> how can a constraint send activate if it's not created yet?

I just checked and it is the CPointerConstraint constructor that calls activate().

The activate() method expects the constraint to be ready, but this is not true during the execution of the constructor. It has not yet been added to the m_vConstraints of the Input Manager.

Perhaps, this activate() call could be moved from the constructor to onNewConstraint(). But I am unsure if this would not cause some problem.

bk304:
> Perhaps, this activate() call could be moved from the constructor to onNewConstraint(). But I am unsure if this would not cause some problem.

It's okay to move it. It'll be much cleaner that way too.

vaxerski:
> Perhaps, this activate() call could be moved from the constructor to onNewConstraint(). But I am unsure if this would not cause some problem.

yes, or, you could also `g_pEventLoopManager->doLater` it

keifufu:
I'm not entirely sure this is the same issue as this PR intends to fix but with this patch I still have my camera snap down randomly when spamming left/right click.

This is with wine-xiv 8.5 on hyprland `b21edb1a9753ba8c236302c1d6dba9ea2418b1d3` with this patch.

keifufu:
Hi @bk304, is this PR abandoned?

bk304:
@keifufu 

Ooohhh, I didn't mean to close this PR. I didn't know or was told that renaming the branch would cause this PR to be closed.

The issue I set out to solve has been resolved.
There are still some instances of the warp issue when the pointer constraint is activated by a workspace switch, but that's because of a problem elsewhere in the code and I don't plan on looking into it.

From what I can see, Github won't let me reopen this PR since I renamed the branch. So I'll create another PR.

Issue #3861, opened 2023-11-15
Per-Window Scaling


chaorace:
### Description

I'd like to be able to set custom scaling for specific windows. In other words: I'd like to be able to use different Wayland scaling multipliers between individual windows within the same workspace.

I feel that such a feature would improve workspace management when using applications which support Wayland scaling but otherwise lack a built-in "zoom" mechanism. The value-add would come from being able to fit otherwise unreasonably bulky applications into a given workspace layout without sacrificing unacceptable amounts of screen real-estate.

Off the top of my head, I can imagine the following possible implementations:

**A.** As a dispatcher (e.g.: `scalewindow +0.2` / `scalewindow exact 2.0`) -- does the spec allow for changing scale on the fly?
**B.** As a window rule (e.g.: `scale 2.0` / `scale 0.5`)

vaxerski:
worth noting this is impossible for xwayland windows

loki1368:
I'd really like "PerWindow scaling" ability too. Because wlfreerdp works bad with global scaling(i have 1.75) and i need an option to drop scale to 1.0 for it(per-window or per-workspace).

DrShadow34:
I second that, this feature would be nice to see. Right now I'm using some wacky stuff with script combined with keybinding hotkeys, to change scaling in config on the fly, or for xwindow I'm just using gamescope with custom resolution. All of this could be avoided if I could set for xwindow and certain apps own scaling.

ildar:
this FR is really critical for those using RDP.
I just have to manually set the scale globally for the moments I use RDP. 😞

ildar:
BTW, can a wayland application choose its scaling factor?

vaxerski:
yes

1over137:
One more potential use of this is to use the native fractional scaling present in some programs (firefox's `layout.css.devPixelsPerPx`) instead of Wayland's 2x then downscale method, which can save resources and produce clearer text.

droc12345:
I started out, in wayland, using wayfire and there are 2 features that I really liked and used a fair amount,
One was being able to set alpha per window, binding to a key and I can make that work with hyprland.
The other was the built in zoom (magnification) where it would apply to a window or even the wallpaper.
That would be similar to what's being asked for here, and I'm not sure how easy it is to implement.

1over137:
[...]

PR #9428, opened 2025-02-17
Add per-window content scaling

dawsers:
#### Add a new dispatcher `scaleactive` that scales the content of the active window

per-window_scaling

Each window can have its contents with a different scale. This can be useful for background apps/jobs you want to control and you don't want them to take much space, pinned windows etc. You can also use it for applications that have their UI too big or too small and don't support native re-scaling.

I don't think I have covered every case yet, but I wanted to propose this initial PR so you guys tell me if there is any real interest for this before I spend more time in it. I was going to implement this in [hyprscroller](https://github.com/dawsers/hyprscroller), but it made more sense to integrate the feature in Hyprland.

There is currently a bug that I haven't figured out: Sometimes the client renders popups to be "inside the monitor", so despite the window being visible and interactive, some popups render in what would be the non-scaled monitor extents.  I can attach an image if there is any interest to help.

The change should be transparent. By default the scale is 1.0, so it shouldn't interfere with Hyprland unless changed.

To try it out, add something like this to your config:

```
# Scaling submap
# will switch to a submap called scaling
bind = $mainMod, period, submap, scaling
# will start a submap called "scaling"
submap = scaling
# sets repeatable binds for scaling the content of the active window
bind = , 1, scaleactive, 0.4 
bind = , 1, submap, reset
bind = , 2, scaleactive, 0.6
bind = , 2, submap, reset
bind = , 3, scaleactive, 0.8
bind = , 3, submap, reset
bind = , 4, scaleactive, 1.25
bind = , 4, submap, reset
bind = , 5, scaleactive, 1.5
bind = , 5, submap, reset
bind = , 6, scaleactive, 2.0
bind = , 6, submap, reset
bind = , 0, scaleactive, -1.0
bind = , 0, submap, reset
bind = , minus, scaleactive, -0.05
bind = , minus, submap, reset
bind = , equal, scaleactive, +0.05
bind = , equal, submap, reset
# use reset to go back to the global submap
bind = , escape, submap, reset
# will reset the submap, meaning end the current one and return to the global one
submap = reset

#bind = $mainMode, period, scroller:scaleactive, 0.5
bind = $mainMode, comma, scaleactive, -1
```

The dispatcher accepts any positive floating point number as scale. Using a negative number (-1 by convention) will reset the scale of the content of that window. The scale has the same meaning as the fractional value for the monitor: 2 would make the content 2x2 times bigger. 0.5 would make it smaller.

Thanks!

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1750,7 +1750,7 @@ void CWindow::sendWindowSize(bool force) {
     // TODO: this should be decoupled from setWindowSize IMO
     const auto REPORTPOS = realToReportPosition();
 
-    const auto REPORTSIZE = realToReportSize();
+    const auto REPORTSIZE = realToReportSize() * m_fContentScale;
 
     if (!force && m_vPendingReportedSize == REPORTSIZE && (m_vReportedPosition == REPORTPOS || !m_bIsX11))
         return;
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -13,6 +13,7 @@
 #include "../config/ConfigValue.hpp"
 #include "../managers/TokenManager.hpp"
 #include "../managers/AnimationManager.hpp"
+#include "../managers/ANRManager.hpp"
 #include "../protocols/XDGShell.hpp"
 #include "../protocols/core/Compositor.hpp"
 #include "../protocols/ContentType.hpp"
@@ -48,6 +49,7 @@ PHLWINDOW CWindow::create(SP<CXWaylandSurface> surface) {
     g_pAnimationManager->createAnimation(0.f, pWindow->m_fDimPercent, g_pConfigManager->getAnimationPropertyConfig("fadeDim"), pWindow, AVARDAMAGE_ENTIRE);
     g_pAnimationManager->createAnimation(0.f, pWindow->m_fMovingToWorkspaceAlpha, g_pConfigManager->getAnimationPropertyConfig("fadeOut"), pWindow, AVARDAMAGE_ENTIRE);
     g_pAnimationManager->createAnimation(0.f, pWindow->m_fMovingFromWorkspaceAlpha, g_pConfigManager->getAnimationPropertyConfig("fadeIn"), pWindow, AVARDAMAGE_ENTIRE);
+    g_pAnimationManager->createAnimation(0.f, pWindow->m_notRespondingTint, g_pConfigManager->getAnimationPropertyConfig("fade"), pWindow, AVARDAMAGE_ENTIRE);
 
     pWindow->addWindowDeco(makeUnique<CHyprDropShadowDecoration>(pWindow));
     pWindow->addWindowDeco(makeUnique<CHyprBorderDecoration>(pWindow));
@@ -71,6 +73,7 @@ PHLWINDOW CWindow::create(SP<CXDGSurfaceResource> resource) {
     g_pAnimationManager->createAnimation(0.f, pWindow->m_fDimPercent, g_pConfigManager->getAnimationPropertyConfig("fadeDim"), pWindow, AVARDAMAGE_ENTIRE);
     g_pAnimationManager->createAnimation(0.f, pWindow->m_fMovingToWorkspaceAlpha, g_pConfigManager->getAnimationPropertyConfig("fadeOut"), pWindow, AVARDAMAGE_ENTIRE);
     g_pAnimationManager->createAnimation(0.f, pWindow->m_fMovingFromWorkspaceAlpha, g_pConfigManager->getAnimationPropertyConfig("fadeIn"), pWindow, AVARDAMAGE_ENTIRE);
+    g_pAnimationManager->createAnimation(0.f, pWindow->m_notRespondingTint, g_pConfigManager->getAnimationPropertyConfig("fade"), pWindow, AVARDAMAGE_ENTIRE);
 
     pWindow->addWindowDeco(makeUnique<CHyprDropShadowDecoration>(pWindow));
     pWindow->addWindowDeco(makeUnique<CHyprBorderDecoration>(pWindow));
@@ -1807,4 +1810,8 @@ void CWindow::deactivateGroupMembers() {
         if (curr == getGroupHead())
             break;
     }
+}
+
+bool CWindow::isNotResponding() {
+    return g_pANRManager->isNotResponding(m_pSelf.lock());
 }
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -113,7 +113,7 @@ CWindow::CWindow(SP<CXWaylandSurface> surface) : m_pXWaylandSurface(surface) {
 }
 
 CWindow::~CWindow() {
-    if (g_pCompositor->m_pLastWindow.lock().get() == this) {
+    if (g_pCompositor->m_pLastWindow == m_pSelf) {
         g_pCompositor->m_pLastFocus.reset();
         g_pCompositor->m_pLastWindow.reset();
     }
@@ -124,7 +124,7 @@ CWindow::~CWindow() {
         return;
 
     g_pHyprRenderer->makeEGLCurrent();
-    std::erase_if(g_pHyprOpenGL->m_mWindowFramebuffers, [&](const auto& other) { return !other.first.lock() || other.first.lock().get() == this; });
+    std::erase_if(g_pHyprOpenGL->m_mWindowFramebuffers, [&](const auto& other) { return other.first.expired() || other.first.get() == this; });
 }
 
 SBoxExtents CWindow::getFullWindowExtents() {
@@ -415,7 +415,7 @@ void CWindow::moveToWorkspace(PHLWORKSPACE pWorkspace) {
             if (*PINITIALWSTRACKING == 2) {
                 // persistent
                 SInitialWorkspaceToken token = std::any_cast<SInitialWorkspaceToken>(TOKEN->data);
-                if (token.primaryOwner.lock().get() == this) {
+                if (token.primaryOwner == m_pSelf) {
                     token.workspace = pWorkspace->getConfigName();
                     TOKEN->data     = token;
                 }
@@ -505,7 +505,7 @@ void CWindow::onUnmap() {
             if (*PINITIALWSTRACKING == 2) {
                 // persistent token, but the first window got removed so the token is gone
                 SInitialWorkspaceToken token = std::any_cast<SInitialWorkspaceToken>(TOKEN->data);
-                if (token.primaryOwner.lock().get() == this)
+                if (token.primaryOwner == m_pSelf)
                     g_pTokenManager->removeToken(TOKEN);
             }
         }
@@ -513,7 +513,7 @@ void CWindow::onUnmap() {
 
     m_iLastWorkspace = m_pWorkspace->m_iID;
 
-    std::erase_if(g_pCompositor->m_vWindowFocusHistory, [&](const auto& other) { return other.expired() || other.lock().get() == this; });
+    std::erase_if(g_pCompositor->m_vWindowFocusHistory, [this](const auto& other) { return other.expired() || other == m_pSelf; });
 
     if (*PCLOSEONLASTSPECIAL && m_pWorkspace && m_pWorkspace->getWindows() == 0 && onSpecialWorkspace()) {
         const auto PMONITOR = m_pMonitor.lock();
@@ -523,7 +523,7 @@ void CWindow::onUnmap() {
 
     const auto PMONITOR = m_pMonitor.lock();
 
-    if (PMONITOR && PMONITOR->solitaryClient.lock().get() == this)
+    if (PMONITOR && PMONITOR->solitaryClient == m_pSelf)
         PMONITOR->solitaryClient.reset();
 
     if (m_pWorkspace) {
@@ -609,9 +609,8 @@ void CWindow::onBorderAngleAnimEnd(WP<CBaseAnimatedVariable> pav) {
 void CWindow::setHidden(bool hidden) {
     m_bHidden = hidden;
 
-    if (hidden && g_pCompositor->m_pLastWindow.lock().get() == this) {
+    if (hidden && g_pCompositor->m_pLastWindow == m_pSelf)
         g_pCompositor->m_pLastWindow.reset();
-    }
 
     setSuspended(hidden);
 }
@@ -785,15 +784,15 @@ void CWindow::applyDynamicRule(const SP<CWindowRule>& r) {
         }
         case CWindowRule::RULE_PROP: {
             const CVarList VARS(r->szRule, 0, ' ');
-            if (auto search = g_pConfigManager->miWindowProperties.find(VARS[1]); search != g_pConfigManager->miWindowProperties.end()) {
+            if (auto search = NWindowProperties::intWindowProperties.find(VARS[1]); search != NWindowProperties::intWindowProperties.end()) {
                 try {
                     *(search->second(m_pSelf.lock())) = CWindowOverridableVar(std::stoi(VARS[2]), priority);
                 } catch (std::exception& e) { Debug::log(ERR, "Rule \"{}\" failed with: {}", r->szRule, e.what()); }
-            } else if (auto search = g_pConfigManager->mfWindowProperties.find(VARS[1]); search != g_pConfigManager->mfWindowProperties.end()) {
+            } else if (auto search = NWindowProperties::floatWindowProperties.find(VARS[1]); search != NWindowProperties::floatWindowProperties.end()) {
                 try {
                     *(search->second(m_pSelf.lock())) = CWindowOverridableVar(std::stof(VARS[2]), priority);
                 } catch (std::exception& e) { Debug::log(ERR, "Rule \"{}\" failed with: {}", r->szRule, e.what()); }
-            } else if (auto search = g_pConfigManager->mbWindowProperties.find(VARS[1]); search != g_pConfigManager->mbWindowProperties.end()) {
+            } else if (auto search = NWindowProperties::boolWindowProperties.find(VARS[1]); search != NWindowProperties::boolWindowProperties.end()) {
                 try {
                     *(search->second(m_pSelf.lock())) = CWindowOverridableVar(VARS[2].empty() ? true : (bool)std::stoi(VARS[2]), priority);
                 } catch (std::exception& e) { Debug::log(ERR, "Rule \"{}\" failed with: {}", r->szRule, e.what()); }
@@ -909,7 +908,7 @@ void CWindow::createGroup() {
 }
 
 void CWindow::destroyGroup() {
-    if (m_sGroupData.pNextWindow.lock().get() == this) {
+    if (m_sGroupData.pNextWindow == m_pSelf) {
         if (m_eGroupRules & GROUP_SET_ALWAYS) {
             Debug::log(LOG, "destoryGroup: window:{:x},title:{} has rule [group set always], ignored", (uintptr_t)this, this->m_szTitle);
             return;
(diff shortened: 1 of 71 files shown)


vaxerski (commented):
some general design remarks:
- consider adding a window rule for this (see prop window rules) then the dispatcher would be to override it.
- we're adding the _conentScale_ param everywhere - no better method? in most cases, it could be read from the window param.

No style nits for now, general design needs some work

sqlerrorthing (approved):

SNAKESLIGHT:
Would love to see the ability to use a relative float such as +0.2, similar to what the splitratio dispatcher accepts. That way you could bind + and - to scale a window.

src/managers/KeybindManager.cpp
```diff
@@ -2148,6 +2149,32 @@ SDispatchResult CKeybindManager::resizeActive(std::string args) {
...
+
+    try {
+        scaleResult = stof(args);
+    } catch (...) {
+        Debug::log(ERR, "Invalid arg \"{}\" in scaleactive!", args);
+        return {.success = false, .error = "Invalid scale in scaleactive!"};
+    }
+
+    float scale = scaleResult.value();
+    if (scale > 0.0f) {
```
vaxerski, line 2168: no {}

vaxerski (commented):
as someone else pointed out, relative scaling for the dispatcher would be a nice addition

dawsers:
OK, done.

I have one question. I want to try to fix the bug I mentioned about the application thinking the window doesn't fit in the viewport and locating some tooltips in the "wrong" position. Where is the code that sends monitor information to the applications? I found how to send window sizes (sendWindowSize() sends an Ack request through the protocol), but I haven't found how monitor sizes are communicated.

vaxerski:
they aren't, it's xdg_positioner. Good luck with getting that to work. Check `CXDGPositionerRules::getPosition`

dawsers:
Everything seems to be working now for Wayland windows. I am going to see if XWayland can be adapted too. Scaling is OK, popups are not, but I haven't tried anything yet.

dawsers:
I see XWayland popups are "regular" floating windows, and don't keep any reference to a parent. I see there is a `CWindow::m_pX11Parent` field, but it is never used or updated. This means I cannot know if the window needs its coordinates  tansformed, so I don't think there is anything else to do for now until this changes or anyone has an idea on how to make this work. Maybe some other structure keeps parenting information or are these popups handled specifically by the application?

So, current status: It seems everything is working for Wayland windows: popups and not. For X windows, scaling works, but popups render at the wrong location. Anything else?

vaxerski:
for X11, I'd say:
who gaf? :P

I'll test this MR a bit tomorrow, if I don't forget.

dawsers:
Cool.

I think this may close #3861

src/managers/KeybindManager.cpp
```diff
@@ -2148,6 +2149,40 @@ SDispatchResult CKeybindManager::resizeActive(std::string args) {
...
+    if (!scaleResult.has_value()) {
+        Debug::log(ERR, "Invalid arg \"{}\" in scaleactive!", args);
+        return {.success = false, .error = "Invalid scale argument in scaleactive!"};
+    }
+
+    float scale = scaleResult.value();
+    if (!exact)
+        scale += 1.0 / PLASTWINDOW->m_sWindowData.contentScale.valueOr(1.0f);
+
+    if (scale > 0.0f)
```
vaxerski, line 2176: do we want to allow scale to be 0.001? I'd say no, because that might crash with the size of the buffer. I'd cap all scales at 0.2 or 0.5

vaxerski (commented):
res ok

src/managers/KeybindManager.cpp
dawsers: OK, I will cap it to 0.25 for example to create a clean x4 multiplier.

dawsers:
Do we want this or should I close it?

vaxerski:
sorry, just busy and I get tons of notifs. We do, just fix the conflict pls

src/desktop/WindowRule.cpp
```diff
@@ -10,7 +10,7 @@ static const auto RULES = std::unordered_set<std::string>{
 static const auto RULES_PREFIX = std::unordered_set<std::string>{
     "animation", "bordercolor",   "bordersize", "center",    "content", "fullscreenstate", "group",    "idleinhibit",   "maxsize",     "minsize",
     "monitor",   "move",          "opacity",    "plugin:",   "prop",    "pseudo",          "rounding", "roundingpower", "scrollmouse", "scrolltouchpad",
-    "size",      "suppressevent", "tag",        "workspace", "xray",
+    "size",      "suppressevent", "tag",        "workspace", "xray",    "contentscale",
```
MightyPlaza, line 13: this shouldn't be needed, props are already being searched
(and same for `rounding` `roundingpower` `scrollmouse` `scrolltouchpad` it seems)

src/desktop/WindowRule.cpp
```diff
@@ -55,6 +55,8 @@ CWindowRule::CWindowRule(const std::string& rule, const std::string& value, bool
         ruleType = RULE_MAXSIZE;
     else if (rule.starts_with("minsize"))
         ruleType = RULE_MINSIZE;
+    else if (rule.starts_with("contentscale"))
```
MightyPlaza, line 58: same thing here

src/managers/KeybindManager.cpp
```diff
@@ -3135,6 +3171,9 @@ SDispatchResult CKeybindManager::setProp(std::string args) {
             PWINDOW->m_sWindowData.minSize = CWindowOverridableVar(configStringToVector2D(VAL), PRIORITY_SET_PROP);
             PWINDOW->clampWindowSize(PWINDOW->m_sWindowData.minSize.value(), std::nullopt);
             PWINDOW->setHidden(false);
+        } else if (PROP == "contentscale") {
+            PWINDOW->m_sWindowData.contentScale = CWindowOverridableVar(1.0f / std::stof(VAL), PRIORITY_SET_PROP);
```
MightyPlaza, line 3175: this should also use the already existing window prop code to support unset and avoid duplication
you won't be able to store the inverted value doe
as for the `PWINDOW->sendWindowSize();` maybe add a separate if check at the end just for it

src/managers/KeybindManager.cpp
```diff
@@ -2148,6 +2149,41 @@ SDispatchResult CKeybindManager::resizeActive(std::string args) {
     return {};
 }
 
+SDispatchResult CKeybindManager::scaleActive(std::string args) {
```
MightyPlaza, line 2152: can't the `setprop` dispatcher be used instead of making a new one?

dawsers: The dispatcher allows absolute and delta increases, as requested above.

src/desktop/WindowRule.cpp
dawsers: I added this because like you saw..."everybody is doing it, why don't we". I thought maybe there is some need or plan to have every property listed there, even if they don't get accessed this way. Do I remove it?

src/desktop/WindowRule.cpp
dawsers: I believe the property needs to be here, because it needs special treatment. Like you mention below, it needs to invert the value and also call `sendWindowSize()` to update the client's resolution. Adding an `if` below is in my opinion, less clean than stating the property is "special" in a way. In the special code (in `src/desktop/Window.cpp`), aside from inversion and updating the client, we could also add checks for the range of the value if wanted (as there are in the dispatcher, suggested by @vaxerski

src/managers/KeybindManager.cpp
dawsers: So we need to agree on whether this property is "special" like the others in the `if` statements or not.

When I implemented it at the beginning, I tried to add it to the list, like you suggest. But then I saw I couldn't do the things I needed if I added it there. I saw other properties had the same problem, and I followed their process.

src/desktop/Window.cpp
```diff
@@ -769,6 +769,14 @@ void CWindow::applyDynamicRule(const SP<CWindowRule>& r) {
             } catch (std::exception& e) { Debug::log(ERR, "minsize rule \"{}\" failed with: {}", r->szRule, e.what()); }
             break;
         }
+        case CWindowRule::RULE_CONTENTSCALE: {
```
MightyPlaza, line 772: this shouldn't be needed

src/desktop/WindowRule.hpp
```diff
@@ -26,6 +26,7 @@ class CWindowRule {
         RULE_IDLEINHIBIT,
         RULE_MAXSIZE,
         RULE_MINSIZE,
+        RULE_CONTENTSCALE,
```
MightyPlaza, line 29: nor this

src/managers/KeybindManager.cpp
MightyPlaza: https://github.com/hyprwm/Hyprland/pull/9566 would allow that for all props

src/desktop/WindowRule.cpp
MightyPlaza: if you're using a prop (`NWindowProperties`) it's already checked for you

src/desktop/WindowRule.cpp
MightyPlaza: yeah it does if you're using your custom rule type, but like I said bellow this should use the default treatment if you're using a prop (`NWindowProperties`)

src/managers/KeybindManager.cpp
MightyPlaza: I don't see a point in inverting the value here
just invert the value when reading/writing or just make the actual internal value what the user will input

as for adding an `if`, I don't know how heavy that function is
but the best is probably just adding it at the end where all the recalculation is being done, if not inside `recalculateWindow();`, which should be called at the end by `g_pLayoutManager->getCurrentLayout()->recalculateMonitor(m->ID);`

src/desktop/WindowRule.hpp
dawsers: So what is your suggestion? It is the only way I found to be able to update the client's resolution, invert the scale etc. It seems to me a generic floating point property doesn't allow me to do those things.

src/managers/KeybindManager.cpp
dawsers: OK, so we will wait for that to happen.

src/managers/KeybindManager.cpp
dawsers: The input value is chosen to be similar to fractional scale so the user has a better intuition on how to use the feature, but then internally it is easier to manage it inverted. That is the only reason, and inverting it just here is easier and more efficient than having to do it every time the value is read or written, or so I thought.

MightyPlaza:
here's an example
had to swap `contentscale` to `scalecontent` due hyprland mathing the `content` window rule instead (which is a bug)
```diff
diff --git a/src/desktop/Window.cpp b/src/desktop/Window.cpp
index e5924b81..487302c7 100644
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -769,14 +769,6 @@ void CWindow::applyDynamicRule(const SP<CWindowRule>& r) {
             } catch (std::exception& e) { Debug::log(ERR, "minsize rule \"{}\" failed with: {}", r->szRule, e.what()); }
             break;
         }
-        case CWindowRule::RULE_CONTENTSCALE: {
-            try {
-                const auto SCALE           = std::stof(r->szRule.substr(13));
-                m_sWindowData.contentScale = CWindowOverridableVar(1.0f / SCALE, priority);
-                sendWindowSize();
-            } catch (std::exception& e) { Debug::log(ERR, "contentscale rule \"{}\" failed with: {}", r->szRule, e.what()); }
-            break;
-        }
         case CWindowRule::RULE_RENDERUNFOCUSED: {
             m_sWindowData.renderUnfocused = CWindowOverridableVar(true, priority);
             g_pHyprRenderer->addWindowToRenderUnfocused(m_pSelf.lock());
@@ -835,6 +827,7 @@ void CWindow::updateDynamicRules() {
     EMIT_HOOK_EVENT("windowUpdateRules", m_pSelf.lock());
 
     g_pLayoutManager->getCurrentLayout()->recalculateMonitor(monitorID());
+    sendWindowSize();
 }
 
 // check if the point is "hidden" under a rounded corner of the window
@@ -1238,7 +1231,7 @@ float CWindow::getScrollTouchpad() {
 }
 
 float CWindow::getContentScale() {
-    return m_sWindowData.contentScale.valueOr(1.0f);
+    return 1.f / m_sWindowData.contentScale.valueOr(1.0f);
 }
 
 bool CWindow::canBeTorn() {
diff --git a/src/desktop/Window.hpp b/src/desktop/Window.hpp
index 9e2a55d3..318b16f0 100644
--- a/src/desktop/Window.hpp
+++ b/src/desktop/Window.hpp
@@ -191,6 +191,7 @@ struct SWindowData {
 
     CWindowOverridableVar<float>              scrollMouse;
     CWindowOverridableVar<float>              scrollTouchpad;
+    CWindowOverridableVar<float>              contentScale;
 
     CWindowOverridableVar<std::string>        animationStyle;
     CWindowOverridableVar<Vector2D>           maxSize;
@@ -199,8 +200,6 @@ struct SWindowData {
     CWindowOverridableVar<CGradientValueData> activeBorderColor;
     CWindowOverridableVar<CGradientValueData> inactiveBorderColor;
 
-    CWindowOverridableVar<float>              contentScale;
-
     CWindowOverridableVar<bool>               persistentSize;
 };
 
@@ -578,7 +577,7 @@ namespace NWindowProperties {
         {"roundingpower", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.roundingPower; }},
         {"scrollmouse", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.scrollMouse; }},
         {"scrolltouchpad", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.scrollTouchpad; }},
-        {"contentscale", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.contentScale; }},
+        {"scalecontent", [](const PHLWINDOW& pWindow) { return &pWindow->m_sWindowData.contentScale; }},
     };
 };
 
diff --git a/src/desktop/WindowRule.cpp b/src/desktop/WindowRule.cpp
index 598c29e7..23269085 100644
--- a/src/desktop/WindowRule.cpp
+++ b/src/desktop/WindowRule.cpp
@@ -10,7 +10,7 @@ static const auto RULES = std::unordered_set<std::string>{
 static const auto RULES_PREFIX = std::unordered_set<std::string>{
     "animation", "bordercolor",   "bordersize", "center",    "content", "fullscreenstate", "group",    "idleinhibit",   "maxsize",     "minsize",
     "monitor",   "move",          "opacity",    "plugin:",   "prop",    "pseudo",          "rounding", "roundingpower", "scrollmouse", "scrolltouchpad",
-    "size",      "suppressevent", "tag",        "workspace", "xray",    "contentscale",
+    "size",      "suppressevent", "tag",        "workspace", "xray",
 };
 
 CWindowRule::CWindowRule(const std::string& rule, const std::string& value, bool isV2, bool isExecRule) : szValue(value), szRule(rule), v2(isV2), execRule(isExecRule) {
@@ -57,8 +57,6 @@ CWindowRule::CWindowRule(const std::string& rule, const std::string& value, bool
         ruleType = RULE_MAXSIZE;
     else if (rule.starts_with("minsize"))
         ruleType = RULE_MINSIZE;
-    else if (rule.starts_with("contentscale"))
-        ruleType = RULE_CONTENTSCALE;
     else if (rule.starts_with("monitor"))
         ruleType = RULE_MONITOR;
     else if (rule.starts_with("move"))
diff --git a/src/desktop/WindowRule.hpp b/src/desktop/WindowRule.hpp
index 98a680ef..f0722362 100644
--- a/src/desktop/WindowRule.hpp
+++ b/src/desktop/WindowRule.hpp
@@ -26,7 +26,6 @@ class CWindowRule {
         RULE_IDLEINHIBIT,
         RULE_MAXSIZE,
         RULE_MINSIZE,
-        RULE_CONTENTSCALE,
         RULE_MONITOR,
         RULE_MOVE,
         RULE_OPACITY,
diff --git a/src/managers/KeybindManager.cpp b/src/managers/KeybindManager.cpp
index 6cb19e6a..f2ccef2b 100644
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -3171,9 +3171,6 @@ SDispatchResult CKeybindManager::setProp(std::string args) {
             PWINDOW->m_sWindowData.minSize = CWindowOverridableVar(configStringToVector2D(VAL), PRIORITY_SET_PROP);
             PWINDOW->clampWindowSize(PWINDOW->m_sWindowData.minSize.value(), std::nullopt);
             PWINDOW->setHidden(false);
-        } else if (PROP == "contentscale") {
-            PWINDOW->m_sWindowData.contentScale = CWindowOverridableVar(1.0f / std::stof(VAL), PRIORITY_SET_PROP);
-            PWINDOW->sendWindowSize();
         } else if (PROP == "alpha") {
             PWINDOW->m_sWindowData.alpha = CWindowOverridableVar(SAlphaValue{std::stof(VAL), PWINDOW->m_sWindowData.alpha.valueOrDefault().m_bOverride}, PRIORITY_SET_PROP);
         } else if (PROP == "alphainactive") {
@@ -3240,6 +3237,7 @@ SDispatchResult CKeybindManager::setProp(std::string args) {
             return {.success = false, .error = "Prop not found"};
     } catch (std::exception& e) { return {.success = false, .error = std::format("Error parsing prop value: {}", std::string(e.what()))}; }
 
+    PWINDOW->sendWindowSize();
     g_pCompositor->updateAllWindowsAnimatedDecorationValues();
 
     if (!(PWINDOW->m_sWindowData.noFocus.valueOrDefault() == noFocus)) {
```

MightyPlaza:
on another note, the scaling doesn't seem to be working during animations that resize the window

dawsers:
> here's an example had to swap `contentscale` to `scalecontent` due hyprland mathing the `content` window rule instead (which is a bug)
> 
> ```diff
> +    PWINDOW->sendWindowSize();
>      g_pCompositor->updateAllWindowsAnimatedDecorationValues();
>  
>      if (!(PWINDOW->m_sWindowData.noFocus.valueOrDefault() == noFocus)) {
> ```


So, to be able to remove the special rule, you'd' rather call `sendWindowSize()` every time a property changes, no matter which one it is. `sendWindowSize()` seems to be an expensive operation, it tells the client to change the resolution at which it is rendering.

And that is also why during animations, this "feature" doesn't work until the animation is finished, because you shouldn't tell the client once very tick to update the resolution at which it has to render the application window.

MightyPlaza:
> > here's an example had to swap `contentscale` to `scalecontent` due hyprland mathing the `content` window rule instead (which is a bug)
> > ```diff
> > +    PWINDOW->sendWindowSize();
> >      g_pCompositor->updateAllWindowsAnimatedDecorationValues();
> >  
> >      if (!(PWINDOW->m_sWindowData.noFocus.valueOrDefault() == noFocus)) {
> > ```
> 
> So, to be able to remove the special rule, you'd' rather call `sendWindowSize()` every time a property changes, no matter which one it is. `sendWindowSize()` seems to be an expensive operation, it tells the client to change the resolution at which it is rendering.
> 
> And that is also why during animations, this "feature" doesn't work until the animation is finished, because you shouldn't tell the client once very tick to update the resolution at which it has to render the application window.

again, you can add a separate `if` inside the loop if that is a problem
the main concern is having a separate parsing of the variable itself, so features across floating `NWindowProperties` don't get inconsistent (for example toggle for bools, unset, or the relative value from the other PR)

dawsers:
OK, we will wait until those PRs are in and I will try to make those changes then, because right now the system doesn't support what you are asking while covering the needs of this feature. I have been waiting for three weeks, and merged upstream changes quite a few times already, so this is starting to take much more time that I was expecting. We can also forget about this PR, and you take the pieces of this you want and use them. I don't care either way. I am not asking or care for any attribution, I only added this PR because I wanted to use the feature in my plugin. Now I don't even know if I will be able to after these changes anyway.

src/desktop/WindowRule.cpp
vaxerski: right, since you are using a prop it's handled for you

MightyPlaza:
https://github.com/hyprwm/Hyprland/pull/9566 has been merged

opensourcecheemsburgers:
I installed your fork with this PR and it works great.

I have a 4k display and I am running Overwatch in native Wayland without gamescope. If I change to a borderless window, this PR allows me to scale my resolution to 1080p so it doesn't cook my poor 5700xt reference card.

Issue #9376, opened 2025-02-09
explicit_sync - off limit frames are entirely black


Skidamek:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

Probably not a regression / I don't remember it happening before

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.47.2 built from branch  at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: Sun Feb 2 00:47:17 2025
Tag: v0.47.2, commits: 5767
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set


System Information:
System name: Linux
Node name: Cachy
Release: 6.13.1-2-cachyos
Version: #1 SMP PREEMPT_DYNAMIC Tue, 04 Feb 2025 03:36:12 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA104 [GeForce RTX 3070 Ti] [10de:2482] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  570.86.16  Release Build  (notroot@898e65f6bf83)  Tue Feb  4 03:49:39 UTC 2025


os-release: NAME="CachyOS Linux"
PRETTY_NAME="CachyOS"
ID=cachyos
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://cachyos.org/"
DOCUMENTATION_URL="https://wiki.cachyos.org/"
SUPPORT_URL="https://discuss.cachyos.org/"
BUG_REPORT_URL="https://github.com/cachyos"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=cachyos


plugins:
```

</details>

### Description

After resolving #9361 The only thing left that frustrates me is that Hyprland renders entire black frames when Hyprland uses explicit sync, game is v-synced and frame goes off limit.

Looks like some implementations of frame limit / v-sync are not perfect like in Minecraft. I can notice it rarely in cyberpunk as well.

[...]

Issue #9340, opened 2025-02-05
Discontinuous picture at high FPS in Games


FaDaQ:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

I don't know, I started using Hyprland only recently

### System Info and Hyprland Version

<details> 
<summary>
My PC
</summary>

```
Kernel: Linux 6.13.1-zen1-1-zen
CPU: i3 10105F
GPU: RTX 2060S
Monitors: 1920x1080@75 and 1920x1080@144
RAM: 32Gb RAM

The configuration from JaKooLit is installed
```

```
Hyprland 0.47.2 built from branch  at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: Sun Feb 2 00:47:17 2025
Tag: v0.47.2, commits: 5767
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set
```
</details>

### Description

Image lags when the FPS is unlocked or too high

The situation is as follows and very strange:
Geometry Dash is a game capable of delivering up to 6k FPS on my PC in the menu, at the BUSSIN level (no joke) in this case, I get ~ 2000 fps at the beginning until the moment I move to the main part of the level - there I literally see a slideshow, despite the FPS counter in the game shows me over 800 frames ...
At the same time, if I limit the FPS to 540 (or another certain value that is below 800) - this will not happen.

A similar problem is present in The Witcher 3.
I get ~75 fps in a certain scene and they look VERY smooth - as long as vertical sync is on...
As soon as I turn it off, the FPS doesn't change, but the picture become lagging.

I see the same effects in DOOM Eternal, although it's much more difficult to achieve there, and my average FPS in DOOM Eternal is generally higher than on Windows(Which is amazing)

### How to reproduce

I'm sorry that I'm doing this through video, but still...

https://youtu.be/_qaxbsxV2Aw

[hyprctlinfo.txt](https://github.com/user-attachments/files/18677316/hyprctlinfo.txt)

### Attach not paste

- [x] I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam

### Checklist of files to include below

- [x] Hyprland config - `hyprctl systeminfo -c` (always include)
- [ ] Crash report (always include in case of crash)
[...]

Issue #8588, opened 2024-11-26
Fullscreen games lag the desktop super heavy while really high fps


rabaimorp:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version
```
Hyprland 0.45.2 built from branch  at commit 12f9a0d0b93f691d4d9923716557154d74777b0a  ([gha] Nix: update inputs).
Date: Tue Nov 19 21:47:18 2024
Tag: v0.45.2, commits: 5451
built against aquamarine 0.5.0


flags set:
debug


System Information:
System name: Linux
Node name: legion
Release: 6.12.1-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 22 Nov 2024 16:04:23 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD107M [GeForce RTX 4050 Max-Q / Mobile] [10de:28e1] (rev a1) (prog-if 00 [VGA controller])
06:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Raphael [1002:164e] (rev db) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  565.57.01  Thu Oct 10 12:29:05 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
  hyprexpo by Vaxry ver 1.0

======Config-Start======


Config File: /home/rabaimor/.config/hypr/hyprland.conf: Read Succeeded
# This file contains general customization options 
# For Environment variables see env.conf
# For Keybinds variables see keybinds.conf 
# For Auto-run stuff see execs.conf
# For Window/layer rules see rules.conf

source=~/.config/hypr/hyprland/env.conf

# MONITOR CONFIG
#monitor=,preferred,auto,1
[...]

Issue #7857, opened 2024-09-19
Choppy game(like 5 fps) if gpu reaches 100% utilization.


JunaidQrysh:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 9856378384539e35cd943604e6a4d696a9d25447  (Nix: use mold linker).
Date: Thu Sep 19 21:23:34 2024
Tag: v0.43.0-44-g98563783, commits: 5240
built against aquamarine 0.4.1


flags: (if any)


System Information:
System name: Linux
Node name: asus
Release: 6.10.10-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Thu, 12 Sep 2024 17:21:02 +0000


GPU information: 
0000:00:02.0 VGA compatible controller [0300]: Intel Corporation TigerLake-H GT1 [UHD Graphics] [8086:9a60] (rev 01) (prog-if 00 [VGA controller])
0000:01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA107M [GeForce RTX 3050 Mobile] [10de:25a2] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  560.35.03  Fri Aug 16 21:39:15 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/qrysh/.config/hypr/hyprland.conf: Read Succeeded
source = ~/.config/hypr/source/*


Config File: /home/qrysh/.config/hypr/source/01-startup.conf: Read Succeeded
$terminal = kitty
$fileManager = thunar
$scrPath = $HOME/.local/share/bin

env = KWIN_FORCE_SW_CURSOR,1
env = GBM_BACKEND,nvidia-drm
[...]

PR #9437, opened 2025-02-18
syncobj: use eventfd instead of stalling fd checks

gulafaran:
use eventfd and add it to the event loop and when it recieves a signal release the the queued surfacestate, this means we dont stall entire compositor when waiting for materilisation of the fd. and change its related usage.

should solve a lot of "compositor 5 fps while ingame overlay shows 300fps" and other similiar weird stalls.

on nvidia this shuffles compositor stalls to sometimes a noticeable rendering stutter, technically the acquire point isnt signaled fast enough so the buffer is stuck and nothing gets rendered until it is. pretty much the same stall as seen before only now its per surface rendering. but that i contribute to nvidia driver being simply shit, and is visible on other compositors.

https://github.com/NVIDIA/open-gpu-kernel-modules/issues/777
https://github.com/NVIDIA/open-gpu-kernel-modules/issues/743

fixes https://github.com/hyprwm/Hyprland/issues/7857
fixes https://github.com/hyprwm/Hyprland/issues/9340
fixes https://github.com/hyprwm/Hyprland/issues/9340
fixes https://github.com/hyprwm/Hyprland/issues/8588
fixes https://github.com/hyprwm/Hyprland/issues/7643
fixes https://github.com/hyprwm/Hyprland/issues/7317
fixes https://github.com/hyprwm/Hyprland/issues/9376
fixes https://github.com/hyprwm/Hyprland/issues/6844
fixes https://github.com/hyprwm/Hyprland/issues/6912

potentially helps:
https://github.com/hyprwm/Hyprland/issues/6617
https://github.com/hyprwm/Hyprland/issues/9011

### remaining quirks to be fixed
- [ x ] resizing rendering window like vkcube is delayed hard
-  its because we are getting a gazzillion xdg shell configure requests that fill the wayland event loop, and eventfd signal gets behind in the queue.
- [ x ] walker crashes https://paste.pika-os.com/upload/bat-tiger-jaguar
-  https://github.com/wmww/gtk4-layer-shell/issues/50 requires gtk4-layer-shell 1.0.4
- [ ] test and see if implicit/explicit sync didnt break hard or some other card/vendors AMD/Intel/etc
- [ ] test and see if gpu/cpu usage didnt regress into other areas. not resetting damage or similiar
- [ x ] syncobj points for async buffers

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1391,16 +1391,17 @@ bool CMonitor::attemptDirectScanout() {
 
     auto explicitOptions = g_pHyprRenderer->getExplicitSyncSettings(output);
 
-    bool DOEXPLICIT = PSURFACE->syncobj && PSURFACE->syncobj->current.acquireTimeline && PSURFACE->syncobj->current.acquireTimeline->timeline && explicitOptions.explicitKMSEnabled;
+    bool DOEXPLICIT = PSURFACE->syncobj && PSURFACE->current.buffer && PSURFACE->current.buffer->acquire && explicitOptions.explicitKMSEnabled;
     if (DOEXPLICIT) {
         // wait for surface's explicit fence if present
-        CFileDescriptor fd = PSURFACE->syncobj->current.acquireTimeline->timeline->exportAsSyncFileFD(PSURFACE->syncobj->current.acquirePoint);
+        CFileDescriptor fd = PSURFACE->current.buffer->acquire->exportAsFD();
         if (fd.isValid()) {
             Debug::log(TRACE, "attemptDirectScanout: setting IN_FENCE for aq to {}", fd.get());
             output->state->setExplicitInFence(fd.get());
-        } else
+        } else {
             Debug::log(TRACE, "attemptDirectScanout: failed to acquire an sync file fd for aq IN_FENCE");
-        DOEXPLICIT = fd.isValid();
+            DOEXPLICIT = false;
+        }
     }
 
     commitSeq++;
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1426,7 +1426,7 @@ bool CMonitor::attemptDirectScanout() {
         Debug::log(LOG, "Entered a direct scanout to {:x}: \"{}\"", (uintptr_t)PCANDIDATE.get(), PCANDIDATE->m_szTitle);
     }
 
-    if (!PBUFFER->lockedByBackend)
+    if (!PBUFFER->lockedByBackend || PBUFFER->hlEvents.backendRelease)
         return true;
 
     // lock buffer while DRM/KMS is using it, then release it when page flip happens since DRM/KMS should be done by then
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1394,10 +1394,10 @@ bool CMonitor::attemptDirectScanout() {
     bool DOEXPLICIT = PSURFACE->syncobj && PSURFACE->current.buffer && PSURFACE->current.buffer->acquire && explicitOptions.explicitKMSEnabled;
     if (DOEXPLICIT) {
         // wait for surface's explicit fence if present
-        CFileDescriptor fd = PSURFACE->current.buffer->acquire->exportAsFD();
-        if (fd.isValid()) {
-            Debug::log(TRACE, "attemptDirectScanout: setting IN_FENCE for aq to {}", fd.get());
-            output->state->setExplicitInFence(fd.get());
+        inFence = PSURFACE->current.buffer->acquire->exportAsFD();
+        if (inFence.isValid()) {
+            Debug::log(TRACE, "attemptDirectScanout: setting IN_FENCE for aq to {}", inFence.get());
+            output->state->setExplicitInFence(inFence.get());
         } else {
             Debug::log(TRACE, "attemptDirectScanout: failed to acquire an sync file fd for aq IN_FENCE");
             DOEXPLICIT = false;
--- a/src/helpers/sync/SyncReleaser.cpp
+++ b/src/helpers/sync/SyncReleaser.cpp
@@ -3,7 +3,7 @@
 #include "../../render/OpenGL.hpp"
 #include <sys/ioctl.h>
 
-#if defined(Q_OS_LINUX)
+#if defined(__linux__)
 #include <linux/sync_file.h>
 #else
 struct sync_merge_data {
--- a/src/helpers/sync/SyncReleaser.cpp
+++ b/src/helpers/sync/SyncReleaser.cpp
@@ -2,13 +2,15 @@
 #include "SyncTimeline.hpp"
 #include "../../render/OpenGL.hpp"
 
-CSyncReleaser::CSyncReleaser(WP<CSyncTimeline> timeline_, uint64_t point_) : timeline(timeline_), point(point_) {
+CSyncReleaser::CSyncReleaser(SP<CSyncTimeline> timeline_, uint64_t point_) : timeline(timeline_), point(point_) {
     ;
 }
 
 CSyncReleaser::~CSyncReleaser() {
-    if (timeline.expired())
+    if (!timeline) {
+        Debug::log(ERR, "CSyncReleaser destructing without a timeline");
         return;
+    }
 
     if (sync)
         timeline->importFromSyncFileFD(point, sync->fd());
--- a/src/helpers/sync/SyncReleaser.cpp
+++ b/src/helpers/sync/SyncReleaser.cpp
@@ -45,11 +45,10 @@ CFileDescriptor CSyncReleaser::mergeSyncFds(const CFileDescriptor& fd1, const CF
     do {
         err = ioctl(fd1.get(), SYNC_IOC_MERGE, &data);
     } while (err == -1 && (errno == EINTR || errno == EAGAIN));
-    if (err < 0) {
+    if (err < 0)
         return CFileDescriptor{};
-    } else {
+    else
         return CFileDescriptor(data.fence);
-    }
 }
 
 void CSyncReleaser::addReleaseSync(SP<CEGLSync> sync) {
--- a/src/helpers/sync/SyncReleaser.cpp
+++ b/src/helpers/sync/SyncReleaser.cpp
@@ -53,7 +53,7 @@ CFileDescriptor CSyncReleaser::mergeSyncFds(const CFileDescriptor& fd1, const CF
 
 void CSyncReleaser::addReleaseSync(SP<CEGLSync> sync) {
     if (m_fd.isValid())
-        mergeSyncFds(m_fd, sync->takeFD());
+        m_fd = mergeSyncFds(m_fd, sync->takeFD());
     else
         m_fd = sync->fd().duplicate();
 
--- a/src/helpers/sync/SyncReleaser.cpp
+++ b/src/helpers/sync/SyncReleaser.cpp
@@ -19,20 +19,20 @@ struct sync_merge_data {
 
 using namespace Hyprutils::OS;
 
-CSyncReleaser::CSyncReleaser(SP<CSyncTimeline> timeline_, uint64_t point_) : timeline(timeline_), point(point_) {
+CSyncReleaser::CSyncReleaser(SP<CSyncTimeline> timeline, uint64_t point) : m_timeline(timeline), m_point(point) {
     ;
 }
 
 CSyncReleaser::~CSyncReleaser() {
-    if (!timeline) {
+    if (!m_timeline) {
         Debug::log(ERR, "CSyncReleaser destructing without a timeline");
         return;
     }
 
     if (m_fd.isValid())
-        timeline->importFromSyncFileFD(point, m_fd);
+        m_timeline->importFromSyncFileFD(m_point, m_fd);
     else
-        timeline->signal(point);
+        m_timeline->signal(m_point);
 }
 
 CFileDescriptor CSyncReleaser::mergeSyncFds(const CFileDescriptor& fd1, const CFileDescriptor& fd2) {
@@ -61,5 +61,5 @@ void CSyncReleaser::addReleaseSync(SP<CEGLSync> sync) {
 }
 
 void CSyncReleaser::drop() {
-    timeline.reset();
+    m_timeline.reset();
 }
--- a/src/helpers/sync/SyncReleaser.cpp
+++ b/src/helpers/sync/SyncReleaser.cpp
@@ -1,6 +1,23 @@
 #include "SyncReleaser.hpp"
 #include "SyncTimeline.hpp"
 #include "../../render/OpenGL.hpp"
+#include <sys/ioctl.h>
+
+#if defined(Q_OS_LINUX)
+#include <linux/sync_file.h>
+#else
+struct sync_merge_data {
+    char  name[32];
+    __s32 fd2;
+    __s32 fence;
+    __u32 flags;
+    __u32 pad;
+};
+#define SYNC_IOC_MAGIC '>'
+#define SYNC_IOC_MERGE _IOWR(SYNC_IOC_MAGIC, 3, struct sync_merge_data)
+#endif
+
+using namespace Hyprutils::OS;
 
 CSyncReleaser::CSyncReleaser(SP<CSyncTimeline> timeline_, uint64_t point_) : timeline(timeline_), point(point_) {
     ;
@@ -12,14 +29,36 @@ CSyncReleaser::~CSyncReleaser() {
         return;
     }
 
-    if (sync)
-        timeline->importFromSyncFileFD(point, sync->fd());
+    if (m_fd.isValid())
+        timeline->importFromSyncFileFD(point, m_fd);
     else
         timeline->signal(point);
 }
 
-void CSyncReleaser::addReleaseSync(SP<CEGLSync> sync_) {
-    sync = sync_;
+CFileDescriptor CSyncReleaser::mergeSyncFds(const CFileDescriptor& fd1, const CFileDescriptor& fd2) {
+    struct sync_merge_data data{
+        .name  = "merged release fence",
+        .fd2   = fd2.get(),
+        .fence = -1,
+    };
+    int err = -1;
+    do {
+        err = ioctl(fd1.get(), SYNC_IOC_MERGE, &data);
+    } while (err == -1 && (errno == EINTR || errno == EAGAIN));
+    if (err < 0) {
+        return CFileDescriptor{};
+    } else {
+        return CFileDescriptor(data.fence);
+    }
+}
+
+void CSyncReleaser::addReleaseSync(SP<CEGLSync> sync) {
+    if (m_fd.isValid())
+        mergeSyncFds(m_fd, sync->takeFD());
+    else
+        m_fd = sync->fd().duplicate();
+
+    m_sync = sync;
 }
 
 void CSyncReleaser::drop() {
--- a/src/helpers/sync/SyncReleaser.hpp
+++ b/src/helpers/sync/SyncReleaser.hpp
@@ -15,7 +15,7 @@ class CEGLSync;
 
 class CSyncReleaser {
   public:
-    CSyncReleaser(WP<CSyncTimeline> timeline_, uint64_t point_);
+    CSyncReleaser(SP<CSyncTimeline> timeline_, uint64_t point_);
     ~CSyncReleaser();
 
     // drops the releaser, will never signal anymore
@@ -25,7 +25,7 @@ class CSyncReleaser {
     void addReleaseSync(SP<CEGLSync> sync);
 
   private:
-    WP<CSyncTimeline> timeline;
+    SP<CSyncTimeline> timeline;
     uint64_t          point = 0;
     SP<CEGLSync>      sync;
 };
--- a/src/helpers/sync/SyncReleaser.hpp
+++ b/src/helpers/sync/SyncReleaser.hpp
@@ -4,6 +4,7 @@
 #include <optional>
 #include <vector>
 #include <functional>
+#include <hyprutils/os/FileDescriptor.hpp>
 #include "../memory/Memory.hpp"
 
 /*
@@ -22,10 +23,12 @@ class CSyncReleaser {
     void drop();
 
     // wait for this gpu job to finish before releasing
-    void addReleaseSync(SP<CEGLSync> sync);
+    Hyprutils::OS::CFileDescriptor mergeSyncFds(const Hyprutils::OS::CFileDescriptor& fd1, const Hyprutils::OS::CFileDescriptor& fd2);
+    void                           addReleaseSync(SP<CEGLSync> sync);
 
   private:
-    SP<CSyncTimeline> timeline;
-    uint64_t          point = 0;
-    SP<CEGLSync>      sync;
+    SP<CSyncTimeline>              timeline;
+    uint64_t                       point = 0;
+    Hyprutils::OS::CFileDescriptor m_fd;
+    SP<CEGLSync>                   m_sync;
 };
--- a/src/helpers/sync/SyncReleaser.hpp
+++ b/src/helpers/sync/SyncReleaser.hpp
@@ -16,7 +16,7 @@ class CEGLSync;
 
 class CSyncReleaser {
   public:
-    CSyncReleaser(SP<CSyncTimeline> timeline_, uint64_t point_);
+    CSyncReleaser(SP<CSyncTimeline> timeline, uint64_t point);
     ~CSyncReleaser();
 
     // drops the releaser, will never signal anymore
@@ -27,8 +27,8 @@ class CSyncReleaser {
     void                           addReleaseSync(SP<CEGLSync> sync);
 
   private:
-    SP<CSyncTimeline>              timeline;
-    uint64_t                       point = 0;
+    SP<CSyncTimeline>              m_timeline;
+    uint64_t                       m_point = 0;
     Hyprutils::OS::CFileDescriptor m_fd;
     SP<CEGLSync>                   m_sync;
 };
--- a/src/protocols/DRMSyncobj.cpp
+++ b/src/protocols/DRMSyncobj.cpp
@@ -168,7 +168,6 @@ void CDRMSyncobjSurfaceResource::removeAllWaiters() {
 
 CDRMSyncobjSurfaceResource::~CDRMSyncobjSurfaceResource() {
     removeAllWaiters();
-    PROTO::sync->destroyResource(this);
 }
 
 bool CDRMSyncobjSurfaceResource::protocolError() {
@@ -215,10 +214,6 @@ CDRMSyncobjTimelineResource::CDRMSyncobjTimelineResource(UP<CWpLinuxDrmSyncobjTi
         resource->error(WP_LINUX_DRM_SYNCOBJ_MANAGER_V1_ERROR_INVALID_TIMELINE, "Timeline failed importing");
         return;
     }
-}
-
-CDRMSyncobjTimelineResource::~CDRMSyncobjTimelineResource() {
-    PROTO::sync->destroyResource(this);
 }
 
 WP<CDRMSyncobjTimelineResource> CDRMSyncobjTimelineResource::fromResource(wl_resource* res) {
@@ -282,10 +277,6 @@ CDRMSyncobjManagerResource::CDRMSyncobjManagerResource(UP<CWpLinuxDrmSyncobjMana
 
         LOGM(LOG, "New linux_drm_timeline at {:x}", (uintptr_t)RESOURCE.get());
     });
-}
-
-CDRMSyncobjManagerResource::~CDRMSyncobjManagerResource() {
-    PROTO::sync->destroyResource(this);
 }
 
 bool CDRMSyncobjManagerResource::good() {
--- a/src/protocols/DRMSyncobj.cpp
+++ b/src/protocols/DRMSyncobj.cpp
@@ -8,7 +8,8 @@
 #include <fcntl.h>
 using namespace Hyprutils::OS;
 
-CDRMSyncobjSurfaceResource::CDRMSyncobjSurfaceResource(SP<CWpLinuxDrmSyncobjSurfaceV1> resource_, SP<CWLSurfaceResource> surface_) : surface(surface_), resource(resource_) {
+CDRMSyncobjSurfaceResource::CDRMSyncobjSurfaceResource(UP<CWpLinuxDrmSyncobjSurfaceV1>&& resource_, SP<CWLSurfaceResource> surface_) :
+    surface(surface_), resource(std::move(resource_)) {
     if UNLIKELY (!good())
         return;
 
@@ -104,7 +105,7 @@ bool CDRMSyncobjSurfaceResource::good() {
     return resource->resource();
 }
 
-CDRMSyncobjTimelineResource::CDRMSyncobjTimelineResource(SP<CWpLinuxDrmSyncobjTimelineV1> resource_, CFileDescriptor&& fd_) : fd(std::move(fd_)), resource(resource_) {
+CDRMSyncobjTimelineResource::CDRMSyncobjTimelineResource(UP<CWpLinuxDrmSyncobjTimelineV1>&& resource_, CFileDescriptor&& fd_) : fd(std::move(fd_)), resource(std::move(resource_)) {
     if UNLIKELY (!good())
         return;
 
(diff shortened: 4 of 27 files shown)


src/protocols/DRMSyncobj.hpp
```diff
@@ -13,21 +14,26 @@ class CSyncTimeline;
...
+    ~CDRMSyncobjSurfaceResource();
 
     bool                   good();
 
     WP<CWLSurfaceResource> surface;
-    struct {
-        WP<CDRMSyncobjTimelineResource> acquireTimeline, releaseTimeline;
-        uint64_t                        acquirePoint = 0, releasePoint = 0;
-    } current, pending;
+    struct TimeLineState {
```
vaxerski, line 22: STimelineState

src/protocols/core/Compositor.cpp
```diff
@@ -412,16 +411,17 @@ CRegion CWLSurfaceResource::accumulateCurrentBufferDamage() {
 }
 
 void CWLSurfaceResource::lockPendingState() {
-    stateLocks++;
+    stateLocked = true;
```
vaxerski, line 414: why remove the proper lock system?

src/protocols/core/Compositor.cpp
```diff
@@ -433,8 +433,9 @@ void CWLSurfaceResource::commitPendingState() {
 
     events.roleCommit.emit();
 
-    if (syncobj && syncobj->current.releaseTimeline && syncobj->current.releaseTimeline->timeline && current.buffer && current.buffer->buffer)
-        current.buffer->releaser = makeShared<CSyncReleaser>(syncobj->current.releaseTimeline->timeline, syncobj->current.releasePoint);
+    if (syncobj && syncobj->release.resource && syncobj->release.resource->timeline && current.buffer && current.buffer->buffer) {
```
vaxerski, line 436: no {}

vaxerski (commented):
huge if works

src/protocols/core/Compositor.cpp
gulafaran: as i saw it the only place where the locking/unlocking was used is from drmsyncobj, and it locks until an acquire point is signaled, but with the "ref" counting it became an rather annoyance to keep track of unrefing and refing when either a new buffer appears because a move/resize of window or similiar so a new acquire point arrives without the first signaled, and if you suddenly do some other shenanigans making the same race condition. when all we want to know is, is it locked or not, so the ref count suddenly isnt +2 but the waiter is gone and new arrives and wont unlock and render and poof trouble. its doable but just felt easier to change it to a locked or not, true/false

src/protocols/core/Compositor.cpp
gulafaran: changed

src/protocols/DRMSyncobj.hpp
gulafaran: changed

src/protocols/core/Compositor.cpp
vaxerski: mhm, ok

ikalco:
wait... for ds, shouldn't we release the wl_buffer whenever DRM signals the out fence??

gulafaran:
> wait... for ds, shouldn't we release the wl_buffer whenever DRM signals the out fence? so the client can reuse that buffer, and the output won't artifact cause drm said scanout is done
> 
> edit: uuuh it seems HL just isn't using the out fence lol, am I missing something? for reference (https://docs.kernel.org/gpu/drm-kms.html)
> 
> > “OUT_FENCE_PTR”:
> > Use this property to pass a file descriptor pointer to DRM. Once the Atomic Commit request call returns OUT_FENCE_PTR will be filled with the file descriptor number of a Sync File. This Sync File contains the CRTC fence that will be signaled when all framebuffers present on the Atomic Commit * request for that given CRTC are scanned out on the screen.

There is a lot wrong at the moment, explicit sync currently relies on new buffer attached, the client can commit multiple times on the same buffer if it wants to, and does sometimes, and this PR probably exposed that quirk even more, and yeah direct scanout i havent begun reading into but removed that buffer manual buffer shenanigans because it was conflicting with the changes i made, there is also a few other things im checking that has popped up while testing

gulafaran:
> wait... for ds, shouldn't we release the wl_buffer whenever DRM signals the out fence? so the client can reuse that buffer, and the output won't artifact cause drm said scanout is done
> 
> edit: uuuh it seems HL just isn't using the out fence lol, am I missing something? for reference (https://docs.kernel.org/gpu/drm-kms.html)
> 
> > “OUT_FENCE_PTR”:
> > Use this property to pass a file descriptor pointer to DRM. Once the Atomic Commit request call returns OUT_FENCE_PTR will be filled with the file descriptor number of a Sync File. This Sync File contains the CRTC fence that will be signaled when all framebuffers present on the Atomic Commit * request for that given CRTC are scanned out on the screen.

And "releasing" when it comes to explicit sync is sending the release point not the buffer .release

UjinT34:
`output->state->enableExplicitOutFenceForNextCommit()` should be called before the commit if out fence is needed. It fills `output->state->state().explicitOutFence`.
Who should handle that fd?

Note that `OUT_FENCE_PTR` shouldn't be used when tearing is allowed.

JunaidQrysh:
I tested this pr & it fixed glmark2 lag but minecraft lag is still the same and also it introduced slight lag in other areas like workspace switching and cursor lag when wlogout is opened.

ikalco:
> `output->state->enableExplicitOutFenceForNextCommit()` should be called before the commit if out fence is needed. It fills `output->state->state().explicitOutFence`. Who should handle that fd?
> 
> Note that `OUT_FENCE_PTR` shouldn't be used when tearing is allowed.

we use that fd, i think we import it as a sync_file, then wait for it to signal, meaning aq plane is done with the primary buffer

gulafaran:
> I tested this pr & it fixed glmark2 lag but minecraft lag is still the same and also it introduced slight lag in other areas like workspace switching and cursor lag when wlogout is opened.

yeah got a few very broken things to figure out, im trying tho.

ikalco:
for this part
```
// #TODO does this apply to explicit sync?
    if (!syncobj && previousBuffer && previousBuffer->buffer && !previousBuffer->buffer->isSynchronous()) {
```

the lockedByBackend is unlocked when page flip buffer swap happens, so technically that's when we should send release point since we're no longer using the buffer
but also, im pretty sure this is what the DRM out fence does, so you could use that instead

czM1K3:
I can also say that this fixed my issue with my RX 580 so it might fix something like #8119 and others.

gulafaran:
> I can also say that this fixed my issue with my RX 580 so it might fix something like #8119 and others.

Probably will, and also some minecraft flickering and other things, im pushing soon a fix for a gtk issue and hyprlock issue so its not quite flawless yet 👍

gulafaran:
> I tested this pr & it fixed glmark2 lag but minecraft lag is still the same and also it introduced slight lag in other areas like workspace switching and cursor lag when wlogout is opened.

should be in a better state now, but not done yet.

src/helpers/sync/SyncReleaser.cpp
```diff
@@ -1,25 +1,66 @@
...
+    };
+    int err = -1;
+    do {
+        err = ioctl(fd1.get(), SYNC_IOC_MERGE, &data);
+    } while (err == -1 && (errno == EINTR || errno == EAGAIN));
+    if (err < 0) {
+        return CFileDescriptor{};
+    } else {
+        return CFileDescriptor(data.fence);
+    }
```
nnyyxxxx, line 52:
```suggestion
    if (err < 0)
        return CFileDescriptor{};
    else
        return CFileDescriptor(data.fence);
```

src/protocols/DRMSyncobj.cpp
```diff
@@ -34,77 +104,109 @@ CDRMSyncobjSurfaceResource::CDRMSyncobjSurfaceResource(SP<CWpLinuxDrmSyncobjSurf
...
+        if (!surface->pending.buffer && !surface->pending.newBuffer && !surface->current.buffer) {
+            surface->commitPendingState(surface->pending); // no pending buffer, no current buffer. probably first commit
             return;
         }
 
-        if (materialized.value())
-            return;
+        if (protocolError()) {
+            return; // bad client.. return.
+        }
```
nnyyxxxx, line 148:
```suggestion
        if (protocolError())
            return; // bad client.. return.
```

src/helpers/Monitor.cpp
```diff
@@ -1315,10 +1315,11 @@ bool CMonitor::attemptDirectScanout() {
     auto explicitOptions = g_pHyprRenderer->getExplicitSyncSettings(output);
 
     // wait for the explicit fence if present, and if kms explicit is allowed
-    bool DOEXPLICIT = PSURFACE->syncobj && PSURFACE->syncobj->current.acquireTimeline && PSURFACE->syncobj->current.acquireTimeline->timeline && explicitOptions.explicitKMSEnabled;
+    bool DOEXPLICIT =
+        PSURFACE->syncobj && PSURFACE->current.buffer && PSURFACE->current.buffer->acquire && !PSURFACE->current.buffer->acquire->expired() && explicitOptions.explicitKMSEnabled;
```
nnyyxxxx, line 1319: current.buffer could be null

gulafaran: it already checks for that? `PSURFACE->syncobj && PSURFACE->current.buffer`

src/protocols/DRMSyncobj.cpp
gulafaran: thanks, changed.

src/helpers/sync/SyncReleaser.cpp
gulafaran: thanks, changed.

src/protocols/LinuxDMABUF.cpp
```diff
@@ -113,6 +113,7 @@ CLinuxDMABuffer::CLinuxDMABuffer(uint32_t id, wl_client* client, Aquamarine::SDM
 CLinuxDMABuffer::~CLinuxDMABuffer() {
     buffer.reset();
     listeners.bufferResourceDestroy.reset();
+    PROTO::linuxDma->destroyResource(this);
```
vaxerski, line 116: we shouldnt do this. The resource should only be destroyed once the client asks for it.

src/protocols/LinuxDMABUF.cpp
```diff
@@ -191,6 +192,10 @@ CLinuxDMABUFParamsResource::CLinuxDMABUFParamsResource(SP<CZwpLinuxBufferParamsV
     });
 }
 
+CLinuxDMABUFParamsResource::~CLinuxDMABUFParamsResource() {
+    PROTO::linuxDma->destroyResource(this);
```
vaxerski, line 196: same here and in other places

src/protocols/core/Compositor.cpp
```diff
@@ -428,30 +428,16 @@ CRegion CWLSurfaceResource::accumulateCurrentBufferDamage() {
...
-void CWLSurfaceResource::lockPendingState() {
-    stateLocks++;
-}
-
-void CWLSurfaceResource::unlockPendingState() {
-    stateLocks--;
-    if (stateLocks <= 0)
-        commitPendingState();
-}
+void CWLSurfaceResource::commitPendingState(SSurfaceState& state) {
```
vaxerski, line 431: is this ever called with anything that isnt pending? Why the param

src/protocols/core/Compositor.hpp
```diff
@@ -53,6 +53,34 @@ class CWLRegionResource {
     SP<CWlRegion> resource;
 };
 
+struct SSurfaceState {
```
vaxerski, line 56: why extract this struct from the surface resource?

src/helpers/sync/SyncReleaser.cpp
```diff
@@ -1,25 +1,65 @@
 #include "SyncReleaser.hpp"
 #include "SyncTimeline.hpp"
 #include "../../render/OpenGL.hpp"
+#include <sys/ioctl.h>
 
-CSyncReleaser::CSyncReleaser(WP<CSyncTimeline> timeline_, uint64_t point_) : timeline(timeline_), point(point_) {
+#if defined(Q_OS_LINUX)
```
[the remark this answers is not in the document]
gulafaran: true that, ;)

src/protocols/core/Compositor.hpp
gulafaran: i wanted the states in drmsyncobj protocol to store ready/waiting states in a container/queue

src/protocols/core/Compositor.cpp
vaxerski: gotcha

src/protocols/core/Compositor.hpp
vaxerski: move it to types/

src/helpers/sync/SyncReleaser.cpp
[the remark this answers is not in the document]
gulafaran: changed.

src/protocols/LinuxDMABUF.cpp
gulafaran: yeah fair enough, changed.

src/protocols/core/Compositor.hpp
gulafaran: moved it to types/SurfaceState.hpp , i assume this is inline of how you visioned it.

vaxerski (approved):
fine by me. Runs alright on my end, but I don't know about others.

fufexan:
WFM as well, fixed desync problems.

fxzzi:
Also running great for me :)

tested on my nvidia desktop, haven't tested any other configurations.

gulafaran:
yeah there has been mostly limited desktop nvidia testing from what i can tell, but so far it improves the nvidia situation drasticly, and exposing a configure xdg shell spam issue. but wouldnt hurt if some intel/amd weird combos test things but i suspect its in a rather workable state right now atleast, always room for improvements.

vaxerski (approved):
me pullin up to merge a banger mr where I did not write a single line

https://github.com/user-attachments/assets/b9fc3bb0-52a6-4e67-8cff-9ed49eaaeb67

Skidamek:
It breaks vrr for me

vaxerski:
ah there it is. Man I did not even get to press the merge button yet

vaxerski:
what even breaks?

fxzzi:
> what even breaks?

currently trying to dbg in #hyprland-dev, vrr just doesn't activate for them and OSD fps counter is stuck at max (aka not active)

JunaidQrysh:
https://github.com/user-attachments/assets/70271d9e-fb21-46ef-83f0-3e1299f0fb85


https://github.com/user-attachments/assets/ee57e350-bc4e-46c5-a205-32749aa8ed86

Wallpaper glitch on startup.
The latest commit fixed lag slightly. The whole hyprland lagged before the latest commit.
Also wlogout lag is not visible enough due to 60 fps video but it looks horrible on 120 hz monitor.

gulafaran:
> ThisMR.mp4
> LatestHyprland.mp4
> 
> Wallpaper glitch on startup. The latest commit fixed lag slightly. The whole hyprland lagged before the latest commit. Also wlogout lag is not visible enough due to 60 fps video but it looks horrible on 120 hz monitor.

is that on intel?

gulafaran:
> intel + nvidia

yeah but only internal connected or an external? so its only using intel or both?

JunaidQrysh:
only internal connected ( intel as primary gpu )

gulafaran:
> only internal connected ( intel as primary gpu )

does something like this make it any better?

[sync.patch](https://github.com/user-attachments/files/19127739/sync.txt)

doesnt work on nvidia tho but.

JunaidQrysh:
> > only internal connected ( intel as primary gpu )
> 
> does something like this make it any better?
> 
> [sync.patch](https://github.com/user-attachments/files/19127739/sync.txt)
> 
> doesnt work on nvidia tho but.

still the same

gulafaran:
> > > only internal connected ( intel as primary gpu )
> > 
> > 
> > does something like this make it any better?
> > [sync.patch](https://github.com/user-attachments/files/19127739/sync.txt)
> > doesnt work on nvidia tho but.
> 
> still the same

i figured out i need to double buffer this, not single buffer it like im doing right now. il try and put something together this weekend. most likely why it works on our nuclear gaming rigs while it falls short on something a bit less powerful

JunaidQrysh:
@gulafaran The MRs in this will be helpful.
https://invent.kde.org/plasma/kwin/-/issues/124.
Kwin uses triple buffering when on Intel gpu otherwise always double buffering. I think hyprland should also do the same( though I might be wrong ).

vaxerski:
> Kwin uses triple buffering when on Intel gpu otherwise always double buffering

different kind of triple buffering, unrelated to the problem at hand here

gulafaran:
> > > only internal connected ( intel as primary gpu )
> > 
> > 
> > does something like this make it any better?
> > [sync.patch](https://github.com/user-attachments/files/19127739/sync.txt)
> > doesnt work on nvidia tho but.
> 
> still the same

seems we were supposed to lock the current async buffer and not the previous. also made a bit cleanup in eglsyncing. is it to much to hope that fixed your lag? :)

JunaidQrysh:
It only improved just slightly but comparing to latest main branch the difference is night and day. Especially on battery.
The main branch also introduces lag(but very little) on battery, But this MR just makes the core i9 11 gen look like a celeron. I tuned on debug overlay and it showed constant 60 fps but they look like 20 on battery and 40 and on ac.

gulafaran:
> It only improved just slightly but comparing to latest main branch the difference is night and day. Especially on battery. The main branch also introduces lag(but very little) on battery, But this MR just makes the core i9 11 gen look like a celeron. I tuned on debug overlay and it showed constant 60 fps but they look like 20 on battery and 40 and on ac.

yeah i believe you. i just find this so mind boggling. works like chefs kiss on my 4080 and my onboard intel, but apparently last change makes one nvidia card vram leak, and one other guys VRR stop working. but not for others. and yours lag hard. but not for others. and i cant reproduce either -_-

gulafaran:
> > > It only improved just slightly but comparing to latest main branch the difference is night and day. Especially on battery. The main branch also introduces lag(but very little) on battery, But this MR just makes the core i9 11 gen look like a celeron. I tuned on debug overlay and it showed constant 60 fps but they look like 20 on battery and 40 and on ac.
> > 
> > 
> > yeah i believe you. i just find this so mind boggling. works like chefs kiss on my 4080 and my onboard intel, but apparently last change makes one nvidia card vram leak, and one other guys VRR stop working. but not for others. and yours lag hard. but not for others. and i cant reproduce either -_-
> 
> I think it should be merged so that hopefully it can be debugged easily if more people experience it. Otherwise atleast games run flawlessly and i barely use my laptop on battery.

feels kinda wrong merging something that regresses, i rather just have it here pending and rebase towards main until its figured out better.

JunaidQrysh:
@gulafaran I wish I knew cpp, i want to make this works so bad. Btw i know enough C to write some basic programs. So how much time do you think it will take me to to learn cpp enough to contribute to hyprland. I can spend HOURS daily.

PlasmaPower:
I made a PR based on this one, https://github.com/gulafaran/Hyprland/pull/1

It gets rid of the buffer reference system in favor of a new "attached buffer" abstraction, which makes it a lot easier to reason about the lifetimes of buffer attachments and when they get released imo. It should also fix an issue in the current codebase of syncReleaser not respecting lockedByBackend. Let me know what you think.

gulafaran:
big thanks to @PlasmaPower @ikalco @UjinT34 and others, we now have working directscanout, async buffers seemingly dropped correctly and things should behave saner. so everyone rebuild retest see if we kind find any other quirks or if we actually solved them with doing it more right. if in luck VRR regression might be solved otherwise there is work pending for such things too https://github.com/hyprwm/Hyprland/pull/9334 , so unless anything major pops up besides VRR i would say this PR is getting close to done. and VRR can be worked on followup PR

calebaden:
I've been testing this PR for the past couple of days with VRR and it works fine for me

src/helpers/sync/SyncReleaser.hpp
```diff
@@ -15,17 +16,19 @@ class CEGLSync;
...
     // wait for this gpu job to finish before releasing
-    void addReleaseSync(SP<CEGLSync> sync);
+    Hyprutils::OS::CFileDescriptor mergeSyncFds(const Hyprutils::OS::CFileDescriptor& fd1, const Hyprutils::OS::CFileDescriptor& fd2);
+    void                           addReleaseSync(SP<CEGLSync> sync);
 
   private:
-    WP<CSyncTimeline> timeline;
-    uint64_t          point = 0;
-    SP<CEGLSync>      sync;
+    SP<CSyncTimeline>              timeline;
```
vaxerski, line 30: why do we mix m_ and no m_?

Just add m_ to all as we now do

src/protocols/DRMSyncobj.cpp
```diff
@@ -34,77 +104,102 @@ CDRMSyncobjSurfaceResource::CDRMSyncobjSurfaceResource(SP<CWpLinuxDrmSyncobjSurf
...
-            return;
+        if (!pendingRelease.expired()) {
+            surface->pending.buffer->release = makeUnique<CDRMSyncPointState>(std::move(pendingRelease));
+            pendingRelease                   = {};
         }
 
-        if (materialized.value())
-            return;
+        if (protocolError())
+            return; // bad client.. return.
```
vaxerski, line 141:
```
// this quacks
quack();
```

src/protocols/DRMSyncobj.hpp
```diff
@@ -2,64 +2,95 @@
...
 
   private:
-    SP<CWpLinuxDrmSyncobjSurfaceV1> resource;
+    void                            removeAllWaiters();
+    WP<CWLSurfaceResource>          surface;
+    UP<CWpLinuxDrmSyncobjSurfaceV1> resource;
+
+    CDRMSyncPointState              pendingAcquire;
+    CDRMSyncPointState              pendingRelease;
+    std::list<SSurfaceState>        pendingStates;
```
vaxerski, line 57: _please_ don't use a list.

src/protocols/core/Compositor.cpp
```diff
@@ -494,11 +471,25 @@ void CWLSurfaceResource::commitPendingState() {
             nullptr);
     }
 
-    lastBuffer = current.buffer ? current.buffer->buffer : WP<IHLBuffer>{};
+    // release the buffer if it's synchronous as update() has done everything thats needed
+    // so we can let the app know we're done.
+    if (!syncobj && current.buffer && current.buffer->buffer && current.buffer->buffer->isSynchronous()) {
```
vaxerski, line 476: if can be commented out too

src/protocols/DRMSyncobj.hpp
gulafaran: its used for iterator validity after removal in the waiter

vaxerski: `std::vector<UP<SSurfaceState>>`

PlasmaPower: Is your concern with the list cache locality / performance?

vaxerski: make it a sp vec then and
```cpp
state.buffer->acquire->addWaiter([this, surf = surface, wp = CWeakPointer<SSurfaceState>(*pendingStates.end())] {
            if (!surf)
                return;

            surf->commitPendingState(wp.lock());
            std::erase(pendingStates, wp);
        });
```

src/protocols/DRMSyncobj.cpp
gulafaran: removed.

src/protocols/DRMSyncobj.hpp
gulafaran: ok changed it its now a vector with shared pointers.

```
state->buffer->acquire->addWaiter([this, surf = surface, wp = CWeakPointer<SSurfaceState>(*std::prev(pendingStates.end()))] {
            if (!surf)
                return;

            surf->commitPendingState(*wp.lock());
            std::erase(pendingStates, wp);
        });
```

src/protocols/core/Compositor.cpp
```diff
@@ -494,11 +471,25 @@ void CWLSurfaceResource::commitPendingState() {
...
+    // }
+
+    // for async buffers, we can only release the buffer once we are unrefing it from current.
+    // if the backend took it, ref it with the lambda. Otherwise, the end of this scope will release it.
+    if (current.buffer && current.buffer->buffer && !current.buffer->buffer->isSynchronous()) {
+        if (current.buffer->buffer->lockedByBackend && !current.buffer->buffer->hlEvents.backendRelease) {
+            current.buffer->buffer->lock();
+            current.buffer->buffer->onBackendRelease([this]() { current.buffer->buffer->unlock(); });
+        }
+    }
```
ikalco, line 488: this should be removed, the lockedByBackend only needs to be used in attemptDirectScanout

src/protocols/DRMSyncobj.hpp
PlasmaPower: Not that this should block merging, but isn't an unordered_set strictly better than a vector here?

src/protocols/core/Compositor.cpp
gulafaran: no we want it here, otherwise sync releaser wont be reset until the new buffer arrives which causes it to become a bit delayish. especially noticeable in minecraft vsync on. like we are always a buffer behind. thats atleast my interpretation and testing of it.

gulafaran: in short we want async buffers to signal as early as possible to let clients know they are free to reuse the same shared memory. and this does it after its been drawn and on next pageflip, not when a new buffer arrives on the next commit.

UjinT34: https://github.com/hyprwm/Hyprland/pull/8966#issuecomment-2575817415 might be relevant

gulafaran: might even be worth testing dropping synchronous buffers in same manner too and not wait until next buffer attach+commit.

ikalco: >no we want it here, otherwise sync releaser wont be reset until the new buffer arrives which causes it to become a bit delayish.

<details>
<summary>The buffer release is supposed to be delayed so that there is always a valid buffer during scanout</summary>
Basically, once a buffer is committed to KMS (using `output->commit();`), it is locked until the next frame.
Then in the new frame's attemptDirectScanout, if we have a new buffer to commit, we will commit it to KMS and then the AQ backend will unlock the original buffer once the new one is being scanned out
</details>

Also `current.buffer->buffer->lockedByBackend` will never be set here because it is only set after a KMS `output->commit()` of the surface buffer, which only happens in attemptDirectScanout in the _next_ onMonitorFrame after a surface commit provides a new/different buffer

ikalco: >might even be worth testing dropping synchronous buffers in same manner too and not wait until next buffer attach+commit.

it's definitely possible to release synchronous buffers earlier since technically we are done with them inside commitPendingState after we copy their data with `buffer->update();`

for asynchronous buffers we are only done with them once OpenGL finishes rendering to the final buffer (when rendering, not direct scanout),
which means async buffers can be released (both wl_buffer.release and release point) with the eglSync mechanism inside `endRender`, since when the eglSync signals that means OpenGL has finished all its queued commands

ikalco: >im not entirerly read up on lockedByBackend but try it yourself, ommit that buffer lock/unlock enter minecraft. and this is just a windowed game without directscanout, set it to vsync and notice the input delay for actual movement/turning/rotation

page didn't refresh lol so I just saw this
Anyway, I'm not at my computer right now but try putting a Debug::log in there to see if the `current.buffer->buffer->lock();` is actually running, if it is then it's a bug somewhere else because lockedByBackend shouldn't be true here

gulafaran: if we skip directscanout entirerly, heres what im seeing and thinking. applications are either using sync or async buffers, sometimes both. they mostly use double buffering. can even be triple buffering. up to the client.

client attaches A, commits A, we draw A. but we never release it or signal it, sometimes client even commits more times on A, then client sends B, commits B, we draw B. only now client is stuck until A is released/signaled. and we get a delay, because we are done with B but dont recieve a new buffer until client has both seen release of A and had its time to do stuff to A

with that in mind, the reason for not dropping current buffer is both allowing same buffer commits but also because it causes issues with non syncobj applications like gtk2, it seems there is a way around that too by actually checking if the buffer fd[0] is readable with https://github.com/hyprwm/hyprutils/blob/dd1f720cbc2dbb3c71167c9598045dd3261d27b3/src/os/FileDescriptor.cpp#L82, and keep them also in a queue/waiter. thats how kwin does it.

UjinT34: A double buffered client will want to switch buffers immediately after it commits one of them. There is a small gap between scanout finish an a new buffer commit. We want to unlock the buffer within this gap. Unlocking it after the commit disrupts the client's pacing and doesn't leave enough time to prepare a new frame in time. This causes stutters and missed frames. Noticeable on lower-end hw or with very high framerates or with strict vsync timing requirements.

How do you tell a sync buffer from async? `buffer->isSynchronous()` is always a predefined constant. Do we get different buffer classes inside a commit? Which ones are used by clients with issues?
Might need different logic for every possible drm/shm - sync/async - explicit/implicit combo.

ikalco: >client attaches A, commits A, we draw A. but we never release it or signal it, sometimes client even commits more times on A, then client sends B, commits B, we draw B. only now client is stuck until A is released/signaled. and we get a delay, because we are done with B but dont recieve a new buffer until client has both seen release of A and had its time to do stuff to A

so where is A released currently? because it shouldn't be through lockedByBackend, if it is then that's unintended
So, the minecraft stuttering being dependent on lockedByBackend when rendering (not ds) would 100% be unintended

AFAIK currently, with explicit sync, we release A in the explicit presented thing at the end of `commitPendingAndDoExplicit`, but without explicit sync A is released on next surface/buffer commit by client, which is fine but potentially a bit performance wasteful but we can optimize later in another pr

My point is, this code should be removed because it _shouldn't_ be doing anything, if it is, then that's a bug

>with that in mind, the reason for not dropping current buffer is both allowing same buffer commits but also because it causes issues with non syncobj applications like gtk2, it seems there is a way around that too by actually checking if the buffer fd[0] is readable with https://github.com/hyprwm/hyprutils/blob/dd1f720cbc2dbb3c71167c9598045dd3261d27b3/src/os/FileDescriptor.cpp#L82, and keep them also in a queue/waiter. thats how kwin does it.

yeah I agree we should do this whenever we optimize the buffer releasing, but I would say out of the scope of this pr

gulafaran: i could live with something like this. but yeah `commitPendingAndDoExplicit` merely adds the fence, not until the destruction of the timeline is it merged/signaled. there is certainly some work to be done about buffer releases :p

```
diff --git a/src/protocols/core/Compositor.cpp b/src/protocols/core/Compositor.cpp
index 6ccb398e..bf3da781 100644
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -81,6 +81,10 @@ CWLSurfaceResource::CWLSurfaceResource(SP<CWlSurface> resource_) : resource(reso
             pending.size       = res && res->buffer ? res->buffer->size : Vector2D{};
             pending.texture    = res && res->buffer ? res->buffer->texture : nullptr;
             pending.bufferSize = res && res->buffer ? res->buffer->size : Vector2D{};
+
+            if (pending.buffer && current.buffer && current.buffer->buffer) {
+                dropCurrentBuffer();
+            }
         }
 
         Vector2D oldBufSize = current.buffer ? current.bufferSize : Vector2D{};
@@ -470,22 +474,6 @@ void CWLSurfaceResource::commitPendingState(SSurfaceState& state) {
             },
             nullptr);
     }
-
-    // release the buffer if it's synchronous as update() has done everything thats needed
-    // so we can let the app know we're done.
-    // if (!syncobj && current.buffer && current.buffer->buffer && current.buffer->buffer->isSynchronous()) {
-    // dropCurrentBuffer(); // lets not drop it at all, it will get dropped on next commit if a new buffer arrives.
-    // solves flickering on nonsyncobj apps on explicit sync.
-    // }
-
-    // for async buffers, we can only release the buffer once we are unrefing it from current.
-    // if the backend took it, ref it with the lambda. Otherwise, the end of this scope will release it.
-    if (current.buffer && current.buffer->buffer && !current.buffer->buffer->isSynchronous()) {
-        if (current.buffer->buffer->lockedByBackend && !current.buffer->buffer->hlEvents.backendRelease) {
-            current.buffer->buffer->lock();
-            current.buffer->buffer->onBackendRelease([this]() { current.buffer->buffer->unlock(); });
-        }
-    }
 }
 
 void CWLSurfaceResource::updateCursorShm(CRegion damage) {

```

vaxerski: `current.buffer->buffer->lockedByBackend` is never set to true here. Or, well, shouldn't be.

vaxerski: it still shouldnt have to be handled here tho

gulafaran: https://github.com/hyprwm/Hyprland/pull/9437/commits/a14d37c92faadceabc5cd3d9bda2d0b8368d7a10

but yeah i feel like buffer drops needs a lookover perhaps. even as mentioned above making a queue/waiter for non syncobj buffers and check if fd is readable etc. seems the delay i noticed came from having asan turned on the test build -_- , but yeah adding an Debug::log its not triggering.

vaxerski (dismissed):
lgtm, are there any issues left? Works fine for me

Issue #9443, opened 2025-02-19
Switch to the upstream color-management-v1 protocol


helixoid:
### Description

Maybe it's time to switch to the staging color-management-v1 protocol for HDR and color management as mpv,sdl and mesa are adopting that protocol.

nnyyxxxx:
doesnt hyprland use its own protocols

nnyyxxxx:
actually i think its a mix of the 2

helixoid:
> doesnt hyprland use its own protocols

I think it's KDE's protocol modified to fit Hyprland
And KDE's was just the upstream protocol but in early stage without some bits which made KDE to switch to upstream a breeze!

helixoid:
But I may be wrong as I am not a contributor.
It's just a wild guess!

UjinT34:
I am on it but it won't be useful without https://github.com/hyprwm/Hyprland/issues/9064
The proto itself doesn't mean much. All CM features are optional and a compositor without any CM can claim proto support.
Current implementation only works with a limited subset of features for fullscreen apps. There is no way to reveal those feature for fullscreen apps and if a windowed one tries to use them than it'll end up with wrong colors. It's not an issue with xx-cm and frog-cm since they are experimental and have limited client support which is usually locked behind some flag. Upstream CM will have more issues with the current approach.

nnyyxxxx:
makes sense

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

Issue #9064, opened 2025-01-14
Color management support


UjinT34:
### Description

An app has some content in format A and says to compositor that it is in format B. The compositor stores it in format C and tells video drivers that it is in D. Monitor display it as E which leaves the user with what the F is going on?
To reduce this frustration and get some better visuals a color management system is needed for proper content processing on the way to a display device.

1. Compositor needs to decide which format better suits current clients and outputs:
 - Simple: single monitor - use the max supported format
 - Advanced: take into account all the clients and use a lesser format if it is enough to reduce resource consumption
 - Tricky: multi-monitor setup - needs some common format
 Caveat: monitors lie about their capabilities, clients lie about their content

   Some relevant info is in `output->parsedEDID` for monitor capabilities and `surface->colorManagement->imageDescription()` for a client (assume sRGB if missing)

2. Compositor should convert all clients' buffers to the selected format and prepare a buffer for displaying
Needs some shaders. Direct buffer manipulation might be easier but highly ineffective. This step should be avoided if possible.

3. Compositor should convert prepared buffer and/or set some drm properties to match output capabilities
  - https://drmdb.emersion.fr/properties/3435973836/DEGAMMA_LUT `output->state->setDeGammaLut`
  - https://drmdb.emersion.fr/properties/3435973836/CTM `output->state->setCTM`
  - https://drmdb.emersion.fr/properties/3435973836/GAMMA_LUT `output->state->setGammaLut`
  - https://drmdb.emersion.fr/properties/3233857728/HDR_OUTPUT_METADATA `output->state->setHDRMetadata`
  Caveat: not all properties might be supported by the given video driver, some of this properties might accept different values on different drivers, properties listed in a "natural" color pipeline order but it is not guaranteed to be so

HDR10 for a fullscreen app is the easiest thing to do because it can got straight to step 3 with little to none modifications. Might still need some drm properties adjusting. Current implementation uses only `HDR_OUTPUT_METADATA`.
On paper color management should be relatively simple and straightforward. In reality some bits should be ignored in favor of some sane defaults. Some double back and forth conversions might be needed just to slap a transformation in between. What those bits and defaults are might depend on hw setup.

Lots of docs about color and HDR https://gitlab.freedesktop.org/pq/color-and-hdr

Somewhat a duplicate of https://github.com/hyprwm/Hyprland/issues/4377

https://github.com/hyprwm/Hyprland/pull/9506
https://github.com/hyprwm/Hyprland/pull/9600
[...]

PR #9444, opened 2025-02-19
Support wp color management proto

UjinT34:
Support wp color-management. https://github.com/hyprwm/Hyprland/issues/9443
Added `debug:full_cm_proto` to switch proto into debug mode (requires restart).

In regular mode this implementation does almost nothing. It handles the minimum required subset of features.
In debug mode all features and values are marked as supported. Features related to fullscreen HDR will work the same as with xx-cm and frog-cm. Other features will only store and pass values and output some log messages.

Needs https://github.com/hyprwm/Hyprland/issues/9064 to be useful. Without it xx-cm and frog-cm should be used instead because those protos are less likely to break things for newer clients with correct wp-cm support.
Needs nix wp version bump.
Needs a proper way of handling images description ids to avoid duplication (not a priority, very unlikely conditions to be useful)

Ready.

- [x] use xml from wayland-protocols (requires wp 1.41)
- [x] move bits from xx-cm and frog-cm to wp
- [x] use internal enums
- [x] add some common CM constants
- [x] correctly state supported features
- [x] add debug mode with all the features
- [x] handle cm attached to subsurfaces


vaxerski:
thanks as always! tag me for review when this is ready.

UjinT34:
@vaxerski ready for review

src/protocols/XXColorManagement.cpp
```diff
@@ -0,0 +1,639 @@
+#include "XXColorManagement.hpp"
+#include "Compositor.hpp"
+#include "ColorManagement.hpp"
+#include "color-management-v1.hpp"
+#include "protocols/types/ColorManagement.hpp"
+
+using namespace NColorManagement;
+
+static wpColorManagerV1TransferFunction getWPTransferFunction(xxColorManagerV4TransferFunction tf) {
```
vaxerski, line 9: just like with content type, instead of converting between wayland types, can't we have a hyprland type?

src/protocols/core/Compositor.cpp
```diff
@@ -306,6 +306,23 @@ void CWLSurfaceResource::breadthfirst(std::function<void(SP<CWLSurfaceResource>,
...
+            continue;
+        const auto found = findFirstHelper(sub->surface.lock(), fn);
+        if (found)
+            return found;
+    }
+    return nullptr;
+}
+
+SP<CWLSurfaceResource> CWLSurfaceResource::findFirst(std::function<bool(SP<CWLSurfaceResource>)> fn) {
+    return findFirstHelper(self.lock(), fn);
```
vaxerski, line 323: why is this preorder and not bf? If it's supposed to be preorder, the fn should mention this (e.g. findFirstPreorder)

src/protocols/types/ColorManagement.hpp
```diff
@@ -1,47 +1,157 @@
...
+        CM_TRANSFER_FUNCTION_LOG_316    = 7,
+        CM_TRANSFER_FUNCTION_XVYCC      = 8,
+        CM_TRANSFER_FUNCTION_SRGB       = 9,
+        CM_TRANSFER_FUNCTION_EXT_SRGB   = 10,
+        CM_TRANSFER_FUNCTION_ST2084_PQ  = 11,
+        CM_TRANSFER_FUNCTION_ST428      = 12,
+        CM_TRANSFER_FUNCTION_HLG        = 13,
+    };
+
+    // FIXME should be ok this way. unsupported primaries/tfs must be rejected earlier. might need a proper switch-case with exception as default.
```
vaxerski, line 35: each enum should have a `0` for INVALID and return invalid if needed

src/protocols/XXColorManagement.cpp
UjinT34: this was made earlier. xx and frog are temporary protos and should be removed as soon as clients move to the wp version. no point in spending more time on this

src/protocols/core/Compositor.cpp
UjinT34: no reason, can be any order. just an easy solution which assumes that there are few total subsurfaces and only one of them has CM attached. it's probably a wine-wayland quirk and most clients will have cm on the main surface. if it's not the case then we can't just passthrough cm properties to hdr metadata.
leave it as is, `findAny`, `findFirstPreorder`?

src/protocols/XXColorManagement.cpp
vaxerski: ok

src/protocols/core/Compositor.cpp
vaxerski: `findFirstPreorder`

src/protocols/types/ColorManagement.hpp
vaxerski: then it shouldn't be a FIXME but a warning at best

vaxerski (approved):

vaxerski:
stellar work as always

myamusashi:
Truly Fast and active development

src/helpers/Monitor.hpp
```diff
@@ -12,6 +12,7 @@
 #include "Timer.hpp"
 #include "math/Math.hpp"
 #include <optional>
+#include "protocols/types/ColorManagement.hpp"
```
zjeffer, line 15: this include is wrong, should be "../protocols/types/ColorManagement.hpp"

causes build failure for plugins: https://github.com/Duckonaut/split-monitor-workspaces/issues/166

zjeffer: @UjinT34 2nd time this has to be fixed after your PR btw :laughing:   (previous time: https://github.com/hyprwm/Hyprland/commit/31431a92714ab7f53c25d2ececb26c5b7264b7e3, fix: https://github.com/hyprwm/Hyprland/commit/44004abc0189b9ba49444431e2a39d2df7e930bb)

UjinT34: auto-import set me up %)

nnyyxxxx: > auto-import set me up %)

what auto import are you using :sob: i didn't even know that was a thing for c++

PR #9436, opened 2025-02-18
feat(anr): add xwayland support

nnyyxxxx:
adds xwayland support to anr

fixes https://github.com/hyprwm/Hyprland/issues/7297

PR #9456, opened 2025-02-21
anr: add xwayland support

vaxerski:
supersedes #9436

for some reason xwayland apps dont wanna reply, dk why for now, but 9436 doesnt work on my end anyways

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1801,8 +1801,5 @@ void CWindow::deactivateGroupMembers() {
 }
 
 bool CWindow::isNotResponding() {
-    if (!m_pXDGSurface)
-        return false;
-
-    return g_pANRManager->isNotResponding(m_pXDGSurface->owner.lock());
+    return g_pANRManager->isNotResponding(m_pSelf.lock());
 }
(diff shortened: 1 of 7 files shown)


vaxerski:
fixed, I guess that's what I get for blindly trusting your X usage and not reading the spec.

nnyyxxxx:
> fixed, I guess that's what I get for blindly trusting your X usage and not reading the spec.

:skull: idk? #9436 works perfectly fine for me

vaxerski:
it's completely off spec

nnyyxxxx:
wrong

vaxerski:
another mistake is that _NET_WM_PING is not required to be supported, the app reports whether it does. xeyes doesnt and under your MR it would always be marked as ANR

vaxerski:
read up on X11 WM_PROTOCOLS spec here https://specifications.freedesktop.org/wm-spec/1.4/ar01s06.html

nnyyxxxx:
not an issue with the implementation though, we are just pinging the application. if it doesnt respond it doesnt respond

vaxerski:
if the app doesn't report supporting pings, we cannot mark it as always ANR because it might be responding, just not to your pings.

nnyyxxxx:
vaxry finally doing something for once

nnyyxxxx:
my motivation is unreal

PR #9460, opened 2025-02-22
Less crash-prone monitor config parser

cyanbun96:
monitor=HDMI-A-2, disabled 1680x1050, 1920x0, 1

The above config line crashes the entire app. stoi gets handed "disabled 1680" which is definitely not an integer. The check for "x" isn't enough to catch it. I'm unsure how (or if) this should be fixed so the band-aid try-catch hack is just a placeholder. Consider this a bug report.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1946,11 +1946,17 @@ std::optional<std::string> CConfigManager::handleMonitor(const std::string& comm
             error += "invalid resolution ";
             newrule.resolution = Vector2D();
         } else {
-            newrule.resolution.x = stoi(ARGS[1].substr(0, ARGS[1].find_first_of('x')));
-            newrule.resolution.y = stoi(ARGS[1].substr(ARGS[1].find_first_of('x') + 1, ARGS[1].find_first_of('@')));
+            try {
+                newrule.resolution.x = stoi(ARGS[1].substr(0, ARGS[1].find_first_of('x')));
+                newrule.resolution.y = stoi(ARGS[1].substr(ARGS[1].find_first_of('x') + 1, ARGS[1].find_first_of('@')));
 
-            if (ARGS[1].contains("@"))
-                newrule.refreshRate = stof(ARGS[1].substr(ARGS[1].find_first_of('@') + 1));
+                if (ARGS[1].contains("@"))
+                    newrule.refreshRate = stof(ARGS[1].substr(ARGS[1].find_first_of('@') + 1));
+            }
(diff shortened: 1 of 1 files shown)


nnyyxxxx:
i can repro this

nnyyxxxx:
> I'm unsure how (or if) this should be fixed so the band-aid try-catch hack is just a placeholder. Consider this a bug report.

if thats the case then if i were you i would draft this

cyanbun96:
> or this could just be moved to an issue lol

Weeeell... _technically_ is **does** fix the issue it describes and doesn't _seem_ to introduce any. It _could_ be merged as-is. It just feels wrong to submit an earnest pull request that's a single try-catch statement where a more elegant solution could be better. Anyway, the acknowledgment and discussion of the issue are what I care about.

nnyyxxxx:
oh okay

gulafaran:
you could use something like

```
auto is_numeric = [](const std::string& str) {
        return !str.empty() && std::all_of(str.begin(), str.end(), ::isdigit);
    };
```

to just make sure the string being passed to stoi is valid

cyanbun96:
> you could use something like
> 
> ```
> auto is_numeric = [](const std::string& str) {
>         return !str.empty() && std::all_of(str.begin(), str.end(), ::isdigit);
>     };
> ```
> 
> to just make sure the string being passed to stoi is valid

That's good for this particular edge case and definitely better than a try-catch, though I wonder if there are other similar issues that it wouldn't catch. I mean, the "check for x" condition just above the problematic line was good enough too, until it wasn't.

cyanbun96:
> image
> 
> we could do something like this

I struggle to come up with a good error message for this. I mean, the "disable/disabled" is supposed to be handled at the top of the function, but then it tries to interpret "disable 1680" as a single parameter in the resolution parser... Someone who understands that whole function and is good at error messages should chime in.

vaxerski (approved):
right. Every stoi should be in a try catch.

vaxerski:
@gulafaran brother we have isNumber in hyprutils

anyways since we can't have a nothrow stoi from the stl just wrap it in try catch

gulafaran:
> @gulafaran brother we have isNumber in hyprutils
> 
> anyways since we can't have a nothrow stoi from the stl just wrapping it in try catch is fine

oh so it seems, good to know :)

vaxerski:
missing clang-format

cyanbun96:
> missing clang-format

sorry, fixed

vaxerski (approved):

PR #9476, opened 2025-02-23
chore: add CXX_STANDARD_REQUIRED flag to ensure c++ version must be c++26

CollieIsCute:
As the [cmake documentation](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html) said, when we use the `CMAKE_CXX_STANDARD` without `CXX_STANDARD_REQUIRED`, then the standard version we set will be an optional value, and may decay back to older c++ version. So I add this configuration into cmakelist to ensure c++ version is `c++26`.

Nope.

Yes, its done.


vaxerski (approved):
thanks!

PR #9500, opened 2025-02-26
Fix ColorManagement.hpp include path in Monitor.hpp

zjeffer:
https://github.com/hyprwm/Hyprland/pull/9444/files#r1972357664

Ideally CI/CD would pick up these kinds of issues

ready

### DIFF
--- a/src/helpers/Monitor.hpp
+++ b/src/helpers/Monitor.hpp
@@ -12,7 +12,7 @@
 #include "Timer.hpp"
 #include "math/Math.hpp"
 #include <optional>
-#include "protocols/types/ColorManagement.hpp"
+#include "../protocols/types/ColorManagement.hpp"
 #include "signal/Signal.hpp"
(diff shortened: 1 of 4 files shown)


nnyyxxxx:
is that it or are there more

zjeffer:
Good point, looks like there are more, will fix

src/protocols/ColorManagement.hpp
```diff
@@ -4,8 +4,8 @@
 #include <vector>
 #include <cstdint>
 #include "WaylandProtocol.hpp"
-#include "helpers/Monitor.hpp"
-#include "protocols/core/Compositor.hpp"
```
nnyyxxxx, line 8: lol what is this

vaxerski (approved):

vaxerski:
generally should be solved with #9297 as that should fail the test

PR #9506, opened 2025-02-27
Simple color management

UjinT34:
Added a simple CM shader and relevant settings.
New monitor rule `cm`:
- auto: subject to change. srgb for 8bpc, wide for 10bpc if supported
-  srgb: default, sRGB primaries
- wide:  wide color gamut, BT2020 primaries
- edid: primaries from edid (known to be inaccurate)
- hdr: wide color gamut and HDR PQ transfer function
- hdredid: same as hdr with edid primaries

New monitor rule `sdrbrightness`: brightness multiplier for SDR -> HDR mapping (sane value 1.0 - 2.0)
New monitor rule `sdrsaturation`: saturation modifier for SDR -> HDR mapping (sane value 0.9 - 1.1)

Removed `experimental:wide_color_gamut`
Removed `experimental:hdr`
Added `render:cm_fs_passthrough` to disable hdr metadata changes for fullscreen apps to avoid modesetting and black screen during the switch

Partially implements https://github.com/hyprwm/Hyprland/issues/9064

Desktop HDR has some weird interactions with blur and other effects.
SDR might also have incorrect interactions with blur and other effects. In general should be unnoticeable.
Old shaders are kept in place but no longer used.
New CM shader might be slow.
Some color transformations might be incorrect.
Interaction with screen capture or mirroring was not tested.
`TEXTURE_EXTERNAL` isn't supported, probably unused.

Ready for merging.
- [x] Convert primaries
- [x] Decode to linear
  - [x] SRGB
  - [x] GAMMA22
  - [x] GAMMA28
  - [x] ST2084_PQ
  - [x] HLG
- [x] Encode from linear
  - [x] SRGB
  - [x] GAMMA22
  - [x] GAMMA28
  - [x] ST2084_PQ
  - [x] HLG
- [x] Handle luminances
- [x] Mark relevant proto features as supported
- [x] Add a way to change output image description
- [x] Add a setting for fullscreen passthrough


UjinT34:
Ready for testing. SDR -> HDR with `experimental:hdr = 1` should have correct colors but has some weird flickers. HDR -> SDR should work fine. Fullscreen apps should work the same.
Blur, fade animations, dimaround, custom screen shaders and external CM plugins might be broken.

abihf:
Hi @UjinT34 , I tried to build from this branch but it crash on start up.
```
[ERR] Failed to link shader: Screen shader parser: Error compiling shader: 0:208(23): error: no matching function for call to `sqrt(int)'; candidates are:
0:208(23): error:    float sqrt(float)
0:208(23): error:    vec2 sqrt(vec2)
0:208(23): error:    vec3 sqrt(vec3)
0:208(23): error:    vec4 sqrt(vec4)
0:208(11): error: operands to arithmetic operators must be numeric
0:208(7): error: no matching function for call to `pow(error, vec3)'; candidates are:
0:208(7): error:    float pow(float, float)
0:208(7): error:    vec2 pow(vec2, vec2)
0:208(7): error:    vec3 pow(vec3, vec3)
0:208(7): error:    vec4 pow(vec4, vec4)
```
[hyprlandCrashReport180489.txt](https://github.com/user-attachments/files/19042355/hyprlandCrashReport180489.txt)

src/render/shaders/CM.frag
```diff
@@ -0,0 +1,392 @@
...
+                vec3(PQ_M2)
+            );
+            break;
+        case CM_TRANSFER_FUNCTION_GAMMA22:
+            return pow(max(color.rgb, vec3(0.0)), vec3(1.0 / 2.2));
+        case CM_TRANSFER_FUNCTION_GAMMA28:
+            return pow(max(color.rgb, vec3(0.0)), vec3(1.0 / 2.8));
+        case CM_TRANSFER_FUNCTION_HLG:
+            isLow = lessThanEqual(color.rgb, vec3(HLG_E_CUT));
+            lo = pow(color.rgb / sqrt(3), vec3(2));
```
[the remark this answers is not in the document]
UjinT34: strange it didn't crash for me

abihf:
just curious question: why are you not using [lcms2](http://www.littlecms.com/) like gamescope or weston?

UjinT34:
> just curious question: why are you not using [lcms2](http://www.littlecms.com/) like gamescope or weston?

No idea how to use it in our case. Might be useful for some matrix calculations but most of the work is inside shaders

abihf:
seems like this lib support icc, might be useful for next iteration.

lkcv:
Is it expected for non-HDR content to be really dim? The colors are still correct and look like they do when cm is set to wide.

abihf:
I replace 80 to 300 and it looks better. But I agree, configuration like `sdrnit` would be nice.

9oOzv:
Hi,

I've installed hyprland the other day for the first time and I've been testing this out for a bit.

Seems to be a lot better than the previous `experimental:hdr = 1`. Setting SDR luminance to 300-400 makes the image bright as it should be.

Should this map SDR programs correctly to BT2020? My browser (and possibly other SDR programs (?)) still seem too saturated. At least oranges seem too red. I don't know color management and color spaces well enough to figure out what exactly the problem is. What I've gathered:

* `sourcePrimaries` seem to be sRGB. `targetPrimaries` are BT2020. 
    * Tried logging these before render.
* The shader code converting primaries seem to be running
    * Played around with the shader code and changing `convertPrimaries` does affect the image
* On chromium, forcing color profile to `ITU-R BT.2020` seems to fix the colors (Quite sure `sourcePrimaries` still evaluates to sRGB).
* Alternatively, setting gamut on my TV manually to "Adobe RGB" (only other option the settings menu has), seems to fix the color (or at least it is much closer to what I think it should).

Is this just how it is supposed to work, or are the primaries not being mapped correctly? Maybe it is just the browser doing weird things (?). Though I do not really understand, how would a client outputting sRGB end up looking too saturated on a wider gamut monitor. If the colors are correctly mapped.

UjinT34:
I've tried several simple SDR brightness modifications. The image becomes brighter but I don't like the resulting quality. Probably needs a more sophisticated approach than simple multiplication.

A client outputs an integer value. Maximum red in sRGB and maximum red in BT2020 will have the same value. But should look different. `convertPrimaries` should change an sRGB maximum red to closest similar looking value in BT2020 which is lower than maximum. Current implementation does that but the effect is very subtle for some reason. Might be some issues with drm properties which aren't touched by this PR.
Chromium with forced BT2020 probably does its own conversion reducing the values and then they are reduced again by HL.

UjinT34:
@9oOzv primaries conversion should work better now

9oOzv:
> @9oOzv primaries conversion should work better now

Great. I think it looks closer now.

Though I do still notice some weird tint here and there compared to all my SDR monitor. Some pinks tint towards orange. Some skin tones towards green (I think... quite difficult to say actually other than it looks a bit off still). Different enough for me to believe it is not just calibration issues.

Nonetheless, I think with the addition of 'sdr brightness' setting, I think this would make Hyprland much more usable on HDR monitors.

9oOzv:
I think Having color intensity config for SDR would be nice at some point, as sRGB colors, even if correctly mapped, can look disappointingly unsaturated compared to wide gamut content on the same screen,. Or compared to the regular SDR monitors, that we are used to and often are much more saturated.

Getting the primaries conversion correct should probably be the first step. I do not think it is correct quite  yet, but I cannot tell for sure just by looking at the picture.

Soliprem:
general/total brightness is very low, even if colors are improved since the last commit (at least on my monitor)

UjinT34:
At this point I am pretty sure that conversion between primaries with the same white point is correct. No white point compensation for now. Other stuff which happens before and/or after the conversion might throw it off.
10bpc is not enough to map all 8bpc sRGB values to 10bpc BT2020. Some precision will be lost in favor of wider range.

Added two new monitor rules to control SDR -> HDR brightness and saturation. `bitdepth,10,cm,hdr,sdrbrightness,1.2,sdrsaturation,1.0` works for me.

lkcv:
With the latest changes, the saturation when using auto or wide seems to have been greatly reduced. The colors are very vivid if I roll back to (https://github.com/hyprwm/Hyprland/pull/9506/commits/58cb7e7d2bacc43080563e85e3af139f27311e46). I'll try to bisect in a bit.

UjinT34:
> With the latest changes, the saturation when using auto or wide seems to have been greatly reduced. The colors are very vivid if I roll back to ([58cb7e7](https://github.com/hyprwm/Hyprland/commit/58cb7e7d2bacc43080563e85e3af139f27311e46)). I'll try to bisect in a bit.

Vivid colors are oversaturated. You might have some monitor settings that compensated that oversaturation and now they cause undersaturation.

lkcv:
I probably should have described the issue differently. Colors appear to have a yellow tint to them after that commit. For example, shades of pink tends to look like a peach color now.

Btw, I check my TV's color settings, and changed the color gamut and gamma curve values and they didn't have an effect on this.

9oOzv:
Looks good to me. `sdrbrightness` and `sdrsaturation` seems to work. For me, something like `sdrbrightness, 1.35, sdrsaturation, 1.04` works quite good.

I am still not sure if the saturated pinks/magentas are missing or if my other monitors just are not showing true `sRGB`. But I guess that can be a separate issue if needed. Ill dig a bit deeper when/if I find time for it.

9oOzv:
> I probably should have described the issue differently. Colors appear to have a yellow tint to them after that commit. For example, shades of pink tends to look like a peach color now.
> 
> Btw, I check my TV's color settings, and changed the color gamut and gamma curve values and they didn't have an effect on this.

This sounds exactly like what I was referring to in my previous comment

Soliprem:
no notes from the last commit. LGTM

Soliprem:
upon further checking/testing, there's something weird about greens specifically. Other colors look fine though. 
Greens in particular look undersaturated (to me). This did not reflect on the monitor test I linked above, so I can't properly quantify it. It's also not a dealbreaker as `hdr` is marked as experimental, but worth checking out (maybe someone here has better testing tools to corroborate what I'm experiencing)

lkcv:
> > I probably should have described the issue differently. Colors appear to have a yellow tint to them after that commit. For example, shades of pink tends to look like a peach color now.
> > Btw, I check my TV's color settings, and changed the color gamut and gamma curve values and they didn't have an effect on this.
> 
> This sounds exactly like what I was referring to in my previous comment

Ah, I didn't see that at first. Is this happening for you when using auto/wide or only in hdr? For me, it happens in the former.

9oOzv:
> > > I probably should have described the issue differently. Colors appear to have a yellow tint to them after that commit. For example, shades of pink tends to look like a peach color now.
> > > Btw, I check my TV's color settings, and changed the color gamut and gamma curve values and they didn't have an effect on this.
> > 
> > 
> > This sounds exactly like what I was referring to in my previous comment
> 
> Ah, I didn't see that at first. Is this happening for you when using auto/wide or only in hdr? For me, it happens in the former.

I have only tried in HDR. I took a picture of my 3 monitors (top one is an HDR OLED TV (Hyprland), bottom left is an SDR one (Hyprland), Bottom right is SDR (attached to a Windows computer). The problem is quite apparent at some specific colors.

[

](url)

9oOzv:
> Shouldn't the primary conversion on happen for SDR->HDR and not in SDR-only scenarios, such as auto/wide?

I think it should happen for `wide` as well. SDR programs output `sRGB`. Monitor configured as `wide` is BT2020. So you need to convert from the programs `sRGB` primaries to the monitors `BT2020` primaries.

lkcv:
Thanks for the explanation. I guess I was wondering why the recent primary conversion changes weren't specific to hdr cm.

Before commit:


After commit:

Soliprem:
what's unclear to me is how to have fullscreen hdr content display correctly. HDR content rn looks like HDR mapped to SDR, when using `cm,hdr` with `render:cm_fs_passthrough=false` (which I've set because otherwise fullscreen applications that aren't hdr content would display in hdr, making colors oversaturated, akin to the old `experimental:hdr`. Am I missing something, @UjinT34 ?

UjinT34:
Modified primaries conversion to avoid transpose shenanigans and added white point adaptation. 
I am at the point where I can't distinguish "not correct" and "don't like".

`render:cm_fs_passthrough = true` disables CM for fullscreen content and tries to rely on drm properties. HDR content should look fine with it.
`render:cm_fs_passthrough = false` maps all content to the monitor settings. This includes HDR -> HDR mapping which in theory should adapt source content to the monitor capabilities better than the passtrough method.
In general `cm` should default to `auto` and `render:cm_fs_passthrough` to `false`. They are set to `srgb` and `true` so that users should get the same experience as before this PR.

Soliprem:
can't see the yellow tint anymore, personally. Will test some more, but for now I'd say LGTM

UjinT34:
Rebased on latest main with some cm proto fixes

UjinT34:
@vaxerski I guess it's time to review this, it's getting too big and not that simple

9oOzv:
sRGB colors look correct to me now. I don't see any other problems either.

src/render/OpenGL.cpp
```diff
@@ -866,6 +868,37 @@ void CHyprOpenGLImpl::initShaders() {
     m_RenderData.pCurrentMonData->m_shQUAD.radius        = glGetUniformLocation(prog, "radius");
     m_RenderData.pCurrentMonData->m_shQUAD.roundingPower = glGetUniformLocation(prog, "roundingPower");
 
+    prog                                                   = createProgram(TEXVERTSRC320, TEXFRAGSRCCM);
```
vaxerski, line 871: you need to omit this if LEGACY_RENDERER is enabled. That forces GLES2

src/render/OpenGL.cpp
```diff
@@ -1316,16 +1352,18 @@ void CHyprOpenGLImpl::renderTextureInternalWithDamage(SP<CTexture> tex, const CB
             usingFinalShader = true;
         } else {
             switch (tex->m_iType) {
-                case TEXTURE_RGBA: shader = &m_RenderData.pCurrentMonData->m_shRGBA; break;
-                case TEXTURE_RGBX: shader = &m_RenderData.pCurrentMonData->m_shRGBX; break;
-                case TEXTURE_EXTERNAL: shader = &m_RenderData.pCurrentMonData->m_shEXT; break;
+                case TEXTURE_RGBA:
```
vaxerski, line 1355: same here and below in CM. Has to have a fallback for GLES2. You might want to extract the logic to a separate fn.

src/render/shaders/Textures.hpp
```diff
@@ -94,6 +94,10 @@ void main() {
     v_texcoord = texcoord;
 })#";
 
+inline const std::string TEXFRAGSRCCM =
+#include "CM.frag"
```
vaxerski, line 98: not that I am complaining, but if you're doing this, might as well extract all shaders to be uniform.

src/render/shaders/CM.frag
```diff
@@ -0,0 +1,439 @@
...
+    E = pow(clamp(ICtCp[0], 0.0, 1.0), PQ_M1);
+    ICtCp[0] = pow(
+        (PQ_C1 + PQ_C2 * E) / (1.0 + PQ_C3 * E),
+        PQ_M2
+    ) / HDR_MAX_LUMINANCE;
+    return vec4(fromLMS * toLinear(vec4(ICtCpPQInv * ICtCp, 1.0), CM_TRANSFER_FUNCTION_ST2084_PQ).rgb * HDR_MAX_LUMINANCE, color[3]);
+}
+
+layout(location = 0) out vec4 fragColor;
+void main() {
```
vaxerski, line 396: note to self: one more thing added to texture shaders and I am writing a CShaderBuilder.

src/render/OpenGL.cpp
UjinT34: might need to disable CM for GLES2... some shader functions work differently...

src/render/shaders/Textures.hpp
UjinT34: https://github.com/hyprwm/Hyprland/pull/9600 WIP

src/render/OpenGL.cpp
vaxerski: that's fine

src/render/shaders/CM.frag
UjinT34: https://github.com/hyprwm/Hyprland/pull/9600 `processShaderIncludes` does a simple substitution for `#include` directives, `includes` substitution map can be used to build a customised shader

9oOzv:
I think `sdrbrightness` is affecting alpha as well.  At least transparent windows get darker with higher brightness.

UjinT34:
> I think `sdrbrightness` is affecting alpha as well. At least transparent windows seem to get less transparent with higher brightness.

Should be fixed in #9600

vaxerski (approved):

vaxerski:
thank you!

kRHYME7:
Hi @UjinT34 I got some massive issues starting from these commits https://github.com/hyprwm/Hyprland/issues/9641. I hope I can get some insights. Thank you.

aruhier:
About the sdrbrightness, would it be possible to have it in nits instead? The factor makes it confusing and dependent of the default value.

gusjengis:
Great work, I've really been looking forward to this feature. I just installed the new release and configured my monitor, almost everything looks great, but I'm noticing an issue with the black levels.

https://github.com/hyprwm/Hyprland/issues/9716

joshuakb2:
Maybe I'm missing something simple, but I'm seeing window borders and cursors rendered much more brightly than other SDR content. For instance, with this rule:
```
general.col.active_border = rgba(33ccffee)
```
And a blank webpage with this style:
```css
body {
  background-color: #33ccffee;
}
```
Here's a photo where you can see the difference between the border color and the webpage color.
PXL_20250326_175237108

PR #9513, opened 2025-03-01
Provide a non-zero time for button release triggered during window unmap

ScatteredRay:

It doesn't seem that this is actually required, but it seems more consistant with other areas, such as in CVirtualKeyboardV1Resource::releasePressed.

I discovered this due to this assert in blender, which may itself be erroneous, but it doesn't seem like it hurts to provide better event time data in this case.

https://projects.blender.org/blender/blender/src/commit/97a1ec806355d7ffc2839706b7deaaac62a84f9b/intern/ghost/intern/GHOST_SystemWayland.cc#L782

Seems unlikely to break compatibility, been testing it locally with no issues.

Ready for merging.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1707,8 +1707,11 @@ void CInputManager::releaseAllMouseButtons() {
     if (PROTO::data->dndActive())
         return;
 
+    timespec now;
+    clock_gettime(CLOCK_MONOTONIC, &now);
+
     for (auto const& mb : buttonsCopy) {
-        g_pSeatManager->sendPointerButton(0, mb, WL_POINTER_BUTTON_STATE_RELEASED);
+        g_pSeatManager->sendPointerButton(now.tv_sec * 1000 + now.tv_nsec / 1000000, mb, WL_POINTER_BUTTON_STATE_RELEASED);
     }
 
     m_lCurrentlyHeldButtons.clear();


vaxerski (approved):
fair enough

Vladimir-csp:
FYI, I still got a scroll event without timestamp on `v0.48.1` in Blender while messing around with swiping workspaces with three fingers gesture (scrolling works fine though).

vaxerski:
well, that's odd. Is this reproducible?

Vladimir-csp:
Launch blender (native Wayland), close the welcome dialog, focus on main work area, start swiping workspaces. I get this on incomplete swipe that ends with cursor inside blender.

vaxerski:
swipe as in touchpad swipe?

Vladimir-csp:
yes, 3 finger gesture.

vaxerski:
[patch.txt](https://github.com/user-attachments/files/19622039/patch.txt)
try this patch

Vladimir-csp:
Seems to have no effect.

vaxerski:
that's odd. Are you sure you applied it correctly? I don't see any other option that could send 0 as time

Vladimir-csp:
It's applied. I'm building debs via [these](https://github.com/cpiber/hyprland-ppa/) src packages. Tried applying both manually and proper way via quilt, so two tests.

vaxerski:
I'll treat this as a bug. Can you open a discussion, I'll test it on my laptop tomorrow?

ScatteredRay:
Hi,

Regarding reproducing this... This shouldn't reproduce with the standard release build of Blender, as that assert should not be enabled in release builds.

To aid in reproduction, @Vladimir-csp As you write up the bug, can you be specific about which build of blender you are using, and how you got it?

To reproduce this issue, you would have to build with a version of blender with the flag `-DWITH_ASSERT_ABORT=ON` explicitly set ON (Or otherwise a config that sets it ON)

ScatteredRay:
@Vladimir-csp I'd also recommend filing a bug with the package maintainer, as Blender explicitly doesn't support running with that flag (And others not specifically set in their Release config)

vaxerski:
generally yes, but still sending 0ms is suboptimal

Vladimir-csp:
> I'd also recommend filing a bug with the package maintainer

Yes, already did that. It's a Debian sid package with a set of problems.

Vladimir-csp:
https://github.com/hyprwm/Hyprland/discussions/9957

PR #9519, opened 2025-03-02
recent commit fixed underlying issue of double free of globals

ikalco:
This part was originally added by me to prevent a crash,
but it was caused by a double free so this is no longer needed due to 445337d

yes


vaxerski:
I'd still do this to avoid the situation where protocol is destroyed after we called wl_display_destroy

ikalco:
> I'd still do this to avoid the situation where protocol is destroyed after we called wl_display_destroy

yeah makes sense,
then we need to make sure all protocols get .reset()
currently some newly added protos weren't added to dtor, maybe I'll add a banner at the top of the page

PR #9553, opened 2025-03-06
Fix touch calibration matrix

khalid151:
The previous implementation overrides any calibration matrix that is set (like in hwdb), even when `transform` is not set in the config. This PR checks if a calibration matrix is set, and would ignore transforms as they'd already be applied. Otherwise, matrices set by transform are used.

This could also be applied to tablets in `setTabletConfigs` but I haven't tested that.

I believe it is ready.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -605,7 +605,7 @@ CConfigManager::CConfigManager() {
     registerConfigVar("input:touchpad:scroll_factor", {1.f});
     registerConfigVar("input:touchpad:flip_x", Hyprlang::INT{0});
     registerConfigVar("input:touchpad:flip_y", Hyprlang::INT{0});
-    registerConfigVar("input:touchdevice:transform", Hyprlang::INT{0});
+    registerConfigVar("input:touchdevice:transform", Hyprlang::INT{-1});
     registerConfigVar("input:touchdevice:output", {"[[Auto]]"});
     registerConfigVar("input:touchdevice:enabled", Hyprlang::INT{1});
     registerConfigVar("input:tablet:transform", Hyprlang::INT{0});
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -724,7 +724,7 @@ CConfigManager::CConfigManager() {
     m_pConfig->addSpecialConfigValue("device", "scroll_button", Hyprlang::INT{0});
     m_pConfig->addSpecialConfigValue("device", "scroll_button_lock", Hyprlang::INT{0});
     m_pConfig->addSpecialConfigValue("device", "scroll_points", {STRVAL_EMPTY});
-    m_pConfig->addSpecialConfigValue("device", "transform", Hyprlang::INT{0});
+    m_pConfig->addSpecialConfigValue("device", "transform", Hyprlang::INT{-1});
     m_pConfig->addSpecialConfigValue("device", "output", {STRVAL_EMPTY});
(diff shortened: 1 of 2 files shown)


src/managers/input/InputManager.cpp
```diff
@@ -1578,8 +1578,14 @@ void CInputManager::setTouchDeviceConfigs(SP<ITouch> dev) {
...
             Debug::log(LOG, "Setting calibration matrix for device {}", PTOUCHDEV->hlName);
-            if (libinput_device_config_calibration_has_matrix(LIBINPUTDEV))
-                libinput_device_config_calibration_set_matrix(LIBINPUTDEV, MATRICES[ROTATION]);
+            if (libinput_device_config_calibration_has_matrix(LIBINPUTDEV)) {
+                float current_matrix[6] = {.0, .0, .0, .0, .0, .0};
+                if (libinput_device_config_calibration_get_default_matrix(LIBINPUTDEV, current_matrix)) {
+                    Debug::log(LOG, "Touch device {} has a non-default calibration. Ignoring transform", PTOUCHDEV->hlName);
+                } else {
+                    libinput_device_config_calibration_set_matrix(LIBINPUTDEV, MATRICES[ROTATION]);
+                }
```
nnyyxxxx, line 1587:
```suggestion
                if (libinput_device_config_calibration_get_default_matrix(LIBINPUTDEV, current_matrix))
                    Debug::log(LOG, "Touch device {} has a non-default calibration. Ignoring transform", PTOUCHDEV->hlName);
                else
                    libinput_device_config_calibration_set_matrix(LIBINPUTDEV, MATRICES[ROTATION]);
```

vaxerski (commented):
what if the user wants to forcefully override this?

khalid151:
Then it would be better to check if the transform is set in the config, and this means getting rid of the default transform value, only applying it when it's set.

Or having a `calibration_matrix` config variable that takes precedence over transform.

But I personally think that it's redundant, because the only way a device would have a non-default calibration matrix is when it is set by the user.

khalid151:
Now transform works as expected, but only when it's set.

src/config/ConfigManager.cpp
```diff
@@ -605,7 +605,7 @@ CConfigManager::CConfigManager() {
     registerConfigVar("input:touchpad:scroll_factor", {1.f});
     registerConfigVar("input:touchpad:flip_x", Hyprlang::INT{0});
     registerConfigVar("input:touchpad:flip_y", Hyprlang::INT{0});
-    registerConfigVar("input:touchdevice:transform", Hyprlang::INT{0});
+    registerConfigVar("input:touchdevice:transform", Hyprlang::INT{-1});
```
vaxerski, line 608: forgot to update in device specific config

vaxerski (commented):
wiki mr needed too

src/config/ConfigManager.cpp
khalid151: This?
https://github.com/hyprwm/Hyprland/blob/7753e8ea686ba0aeaa825502f27e5b3f813faade/src/config/ConfigManager.cpp#L728

If this looks like a good solution, then what do you think about doing the same for tablets as they can also have a calibration matrix set?

vaxerski: sure

khalid151: I think it should be good to go now. Updated the transform value and applied the same check for tablets.

vaxerski:
ah I see you made one. Thanks!

PR #9556, opened 2025-03-07
fix direct scanout glitches

ikalco:
this fixes direct scanout glitches by ensuring that attemptDirectScanout doesn't try to recommit the same buffer to AQ
which would cause a pageflip event and the backendRelease to release the same buffer too early

consider enabling ds by default?
I haven't really looked for other issues
but i remember that the glitches was the blocker for default enabled ds when AQ was merged

yes


vaxerski (approved):
rip scanout glitches, you will be missed

vaxerski:
> consider enabling ds by default?

maybe after #9437

PR #9566, opened 2025-03-08
allow incrementing window props

MightyPlaza:
allows incrementing window properties with `hyprctl dispatch setprop class:foot rounding +10` for example
seems to work fine, but might need some clean up and a wiki MR
current syntax also breaks setting negative values to props

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1807,3 +1807,13 @@ void CWindow::deactivateGroupMembers() {
 bool CWindow::isNotResponding() {
     return g_pANRManager->isNotResponding(m_pSelf.lock());
 }
+
+template <>
+int CWindowOverridableVar<int>::valueOrDefault() {
+    return defaultValue.has_value() ? valueOr(defaultValue.value()) : valueOr(*CConfigValue<Hyprlang::INT>(configValueString.value()));
+}
+
+template <>
+float CWindowOverridableVar<float>::valueOrDefault() {
+    return defaultValue.has_value() ? valueOr(defaultValue.value()) : valueOr(*CConfigValue<Hyprlang::FLOAT>(configValueString.value()));
+}
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -11,6 +11,7 @@
 #include "../render/decorations/CHyprGroupBarDecoration.hpp"
 #include "../render/decorations/CHyprBorderDecoration.hpp"
 #include "../config/ConfigValue.hpp"
+#include "../config/ConfigManager.hpp"
 #include "../managers/TokenManager.hpp"
 #include "../managers/AnimationManager.hpp"
 #include "../managers/ANRManager.hpp"
@@ -780,7 +781,7 @@ void CWindow::applyDynamicRule(const SP<CWindowRule>& r) {
             const CVarList VARS(r->szRule, 0, ' ');
             if (auto search = NWindowProperties::intWindowProperties.find(VARS[1]); search != NWindowProperties::intWindowProperties.end()) {
                 try {
-                    *(search->second(m_pSelf.lock())) = CWindowOverridableVar(std::stoi(VARS[2]), priority);
+                    *(search->second(m_pSelf.lock())) = CWindowOverridableVar(Hyprlang::INT(std::stoi(VARS[2])), priority);
                 } catch (std::exception& e) { Debug::log(ERR, "Rule \"{}\" failed with: {}", r->szRule, e.what()); }
             } else if (auto search = NWindowProperties::floatWindowProperties.find(VARS[1]); search != NWindowProperties::floatWindowProperties.end()) {
                 try {
@@ -1806,14 +1807,4 @@ void CWindow::deactivateGroupMembers() {
 
 bool CWindow::isNotResponding() {
     return g_pANRManager->isNotResponding(m_pSelf.lock());
-}
-
-template <>
-int CWindowOverridableVar<int>::valueOrDefault() {
-    return defaultValue.has_value() ? valueOr(defaultValue.value()) : valueOr(*CConfigValue<Hyprlang::INT>(configValueString.value()));
-}
-
-template <>
-float CWindowOverridableVar<float>::valueOrDefault() {
-    return defaultValue.has_value() ? valueOr(defaultValue.value()) : valueOr(*CConfigValue<Hyprlang::FLOAT>(configValueString.value()));
 }
--- a/src/desktop/Window.hpp
+++ b/src/desktop/Window.hpp
@@ -83,13 +83,17 @@ enum eOverridePriority : uint8_t {
     PRIORITY_SET_PROP,
 };
 
-template <typename T, bool Addable = std::is_same<T, bool>::value || std::is_same<T, int>::value || std::is_same<T, float>::value>
+template <typename T, bool Extended = std::is_same<T, bool>::value || std::is_same<T, int>::value || std::is_same<T, float>::value>
 class CWindowOverridableVar {
   public:
-    CWindowOverridableVar(T const& value, eOverridePriority priority, bool relative = false) {
-        values[priority] = std::make_pair(value, relative);
+    CWindowOverridableVar(T const& value, eOverridePriority priority) {
+        values[priority] = value;
     }
+
     CWindowOverridableVar(T const& value) : defaultValue{value} {}
+    CWindowOverridableVar(std::string const& value)
+        requires(std::is_same_v<T, int> || std::is_same_v<T, float>)
+        : configValueString{value} {}
 
     CWindowOverridableVar()  = default;
     ~CWindowOverridableVar() = default;
@@ -116,7 +120,7 @@ class CWindowOverridableVar {
 
     T value() {
         if (!values.empty())
-            return std::get<0>(std::prev(values.end())->second);
+            return std::prev(values.end())->second;
         else
             throw std::bad_optional_access();
     }
@@ -129,20 +133,14 @@ class CWindowOverridableVar {
     }
 
     T valueOr(T const& other) {
-        if (hasValue()) {
-            if constexpr (Addable) {
-                if constexpr (std::is_same<T, bool>::value)
-                    return (isValueRelative() ? other : 0) ^ value();
-                else
-                    return (isValueRelative() ? other : 0) + value();
-            } else
-                return value();
-        } else
+        if (hasValue())
+            return value();
+        else
             return other;
     }
 
     T valueOrDefault() {
-        return valueOr(defaultValue);
+        return valueOr(defaultValue.value());
     }
 
     eOverridePriority getPriority() {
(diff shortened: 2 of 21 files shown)


MightyPlaza:
don't see any way to get around having a different code for booleans since `true + true = true` instead of wrapping around
aside from that just needs syntax suggestions to separate `-10` absolute from `-10` relative so I can make a wiki MR

vaxerski (commented):
might want to add min/max value support while you're at it.

MightyPlaza:
I ended with the diff bellow
however if, for example, the relative border value, the value passed to `valueOr` and the max value are all 10
when trying to lower the value it will go to 15 and stay a 10, then to 10 which will stay at 10 again and only will be changed at the 3rd reduction
not sure how to get around that in a reasonable way

```diff
diff --git a/src/desktop/Window.hpp b/src/desktop/Window.hpp
index 7bccb3f5..3ebfe018 100644
--- a/src/desktop/Window.hpp
+++ b/src/desktop/Window.hpp
@@ -83,12 +83,18 @@ enum eOverridePriority : uint8_t {
     PRIORITY_SET_PROP,
 };
 
-template <typename T, bool Addable = std::is_same<T, bool>::value || std::is_same<T, int>::value || std::is_same<T, float>::value>
+template <typename T>
+T clampOptional(T const& value, std::optional<T> const& min, std::optional<T> const& max) {
+    return std::clamp(value, min.value_or(std::numeric_limits<T>::min()), max.value_or(std::numeric_limits<T>::max()));
+}
+
+template <typename T, bool Extended = std::is_same<T, bool>::value || std::is_same<T, int>::value || std::is_same<T, float>::value>
 class CWindowOverridableVar {
   public:
     CWindowOverridableVar(T const& value, eOverridePriority priority, bool relative = false) {
         values[priority] = std::make_pair(value, relative);
     }
+    CWindowOverridableVar(std::optional<T> const& min, std::optional<T> const& max) : min{min}, max{max} {}
     CWindowOverridableVar(T const& value) : defaultValue{value} {}
 
     CWindowOverridableVar()  = default;
@@ -130,11 +136,11 @@ class CWindowOverridableVar {
 
     T valueOr(T const& other) {
         if (hasValue()) {
-            if constexpr (Addable) {
+            if constexpr (Extended) {
                 if constexpr (std::is_same<T, bool>::value)
                     return (isValueRelative() ? other : 0) ^ value();
                 else
-                    return (isValueRelative() ? other : 0) + value();
+                    return clampOptional((isValueRelative() ? other : 0) + value(), min, max);
             } else
                 return value();
         } else
@@ -153,7 +159,7 @@ class CWindowOverridableVar {
     }
 
     void increment(T const& other, eOverridePriority priority) {
-        if constexpr (!Addable)
+        if constexpr (!Extended)
             return;
 
         if (hasValue() && getPriority() <= priority) {
@@ -182,6 +188,8 @@ class CWindowOverridableVar {
   private:
     std::map<eOverridePriority, std::pair<T, bool>> values;
     T                                               defaultValue; // used for toggling, so required for bool
+    std::optional<T>                                min;
+    std::optional<T>                                max;
 };
 
 struct SWindowData {
@@ -211,10 +219,10 @@ struct SWindowData {
     CWindowOverridableVar<bool>               xray               = false;
     CWindowOverridableVar<bool>               renderUnfocused    = false;
 
-    CWindowOverridableVar<int>                borderSize;
-    CWindowOverridableVar<int>                rounding;
+    CWindowOverridableVar<int>                borderSize = CWindowOverridableVar<int>(0, std::nullopt);
+    CWindowOverridableVar<int>                rounding   = CWindowOverridableVar<int>(0, std::nullopt);
 
-    CWindowOverridableVar<float>              roundingPower;
+    CWindowOverridableVar<float>              roundingPower = CWindowOverridableVar<float>(0, std::nullopt);
     CWindowOverridableVar<float>              scrollMouse;
     CWindowOverridableVar<float>              scrollTouchpad;

```

vaxerski:
what?

MightyPlaza:
> what?

I can't clamp relative values when they're written to, since they're only assigned a real value on `valueOr()` (can be clamped here)
this means you can increment them as much as you'd like without having any real value change
when you'd try to decrement them you'd have to undo all the increments above which would have no real change either
this could be perceived as a bug to someone who has no idea how it works in the background

vaxerski:
> I can't clamp relative values when they're written to

I don't see why not?

MightyPlaza:
> > I can't clamp relative values when they're written to
> 
> I don't see why not?

that would mean clamping their relative value
so for example between `-5` and `+5` added to the current value, instead of between `0` and `20`
which I don't think it's very useful

vaxerski:
can't you just... `value = clamp(value + relative, 0, 20);`?

MightyPlaza:
> can't you just... `value = clamp(value + relative, 0, 20);`?

yes but that can only be done when you read the value, only then do we have access to the `value` to add to
and that's what my patch is doing, but it results in the issue where changing `relative` might not have any impact (explained in more detail above)

vaxerski:
what the hell is a "relative" value? I am dead lost here. What is going on?

Window props hold a fucking int or float. No relatives here. No "when you read"

```
increment(T by) {
  value += by;
}
```

????

MightyPlaza:
> what the hell is a "relative" value? I am dead lost here. What is going on?
> 
> Window props hold a fucking int or float. No relatives here. No "when you read"
> 
> ```
> increment(T by) {
>   value += by;
> }
> ```
> 
> ????

it's what I've implemented in this PR
I've had to it instead of using the specified default value because

MightyPlaza:
> what the hell is a "relative" value? I am dead lost here. What is going on?
> 
> Window props hold a fucking int or float. No relatives here. No "when you read"
> 
> ```
> increment(T by) {
>   value += by;
> }
> ```
> 
> ????

it's what I've implemented in this PR
I've added them instead of using the specified default value because they're not available at compile time / startup since they're from the config
and could be overwritten by something elsewhere, hence just storing some relative value and only evaluate it when needed

if this was not the case clamping would be very easy, but I'm not sure how to get around that

vaxerski:
no, the config should just set them. the value itself can be constructed with the min and max.

MightyPlaza:
what about now?
also min and max are intended to be hardcoded right?
on another note, maybe Vector2D can be added to extended later

src/desktop/Window.hpp
```diff
@@ -185,12 +227,12 @@ struct SWindowData {
...
 
-    CWindowOverridableVar<int>                rounding;
-    CWindowOverridableVar<float>              roundingPower;
-    CWindowOverridableVar<int>                borderSize;
+    CWindowOverridableVar<int>                borderSize = {std::string("general:border_size"), 0, std::nullopt};
+    CWindowOverridableVar<int>                rounding   = {std::string("decoration:rounding"), 0, std::nullopt};
 
-    CWindowOverridableVar<float>              scrollMouse;
-    CWindowOverridableVar<float>              scrollTouchpad;
+    CWindowOverridableVar<float>              roundingPower  = std::string("decoration:rounding_power");
```
vaxerski, line 233: why did these 3 suddenly get initialized

vaxerski (commented):
rest looks reasonable

src/desktop/Window.hpp
MightyPlaza: if no value is passed `std::optional` will throw when `value()` is called
which increment will cal if no value is set previously

vaxerski: but some aren't initialized still so?

MightyPlaza: I'm talking about the default value, the min and max are optional
Just added it to some as an example and because it made sense to not have border and rounding negative

Additionally I just realized that `scrollMouse` and `scrollTouchpad` also have config options so will be adding them if that was what you meant

src/desktop/Window.cpp
```diff
@@ -1806,3 +1806,13 @@ void CWindow::deactivateGroupMembers() {
 bool CWindow::isNotResponding() {
     return g_pANRManager->isNotResponding(m_pSelf.lock());
 }
+
+template <>
+int CWindowOverridableVar<int>::valueOrDefault() {
+    return defaultValue.has_value() ? valueOr(defaultValue.value()) : valueOr(*CConfigValue<Hyprlang::INT>(configValueString.value()));
```
vaxerski, line 1812: ugh, please cache CConfigValue<> objects. These perform an expensive lookup in the ctor.

vaxerski: forward declare?

MightyPlaza:
```
diff --git a/src/desktop/Window.hpp b/src/desktop/Window.hpp
index a2e9ca12..7e3d85e0 100644
--- a/src/desktop/Window.hpp
+++ b/src/desktop/Window.hpp
@@ -20,6 +20,7 @@
 #include "Workspace.hpp"
 #include "WindowRule.hpp"
 #include "../protocols/types/ContentType.hpp"
+#include <hyprlang.hpp>
 
 class CXDGSurfaceResource;
 class CXWaylandSurface;
@@ -88,6 +89,9 @@ T clampOptional(T const& value, std::optional<T> const& min, std::optional<T> co
     return std::clamp(value, min.value_or(std::numeric_limits<T>::min()), max.value_or(std::numeric_limits<T>::max()));
 }
 
+template <typename T>
+class CConfigValue;
+
 template <typename T, bool Extended = std::is_same<T, bool>::value || std::is_same<T, int>::value || std::is_same<T, float>::value>
 class CWindowOverridableVar {
   public:
@@ -181,10 +185,11 @@ class CWindowOverridableVar {
 
   private:
     std::map<eOverridePriority, T> values;
     std::optional<T>               defaultValue; // used for toggling, so required for bool
     std::optional<std::string>     configValueString;
     std::optional<T>               minValue;
     std::optional<T>               maxValue;
+    CConfigValue<Hyprlang::FLOAT>  configValue;
 };
 
 template <>
```
smth like above doesn't compile
not sure if I'm missing something

vaxerski: ah, ya cant forward declare a templated type. hm. I'll revisit this tomorrow, ping me if I dont

vaxerski:
I can't push to this branch asshole

vaxerski:
[patch.txt](https://github.com/user-attachments/files/19652231/patch.txt)
patch

MightyPlaza:
> I can't push to this branch asshole

image
no idea why

> [patch.txt](https://github.com/user-attachments/files/19652231/patch.txt) patch

seems to be missing the `WindowOverridableVar.hpp` file

vaxerski:
[0001-e.patch.txt](https://github.com/user-attachments/files/19656729/0001-e.patch.txt)

MightyPlaza:
patch commited

vaxerski (approved):
wiki mr neeedde

MightyPlaza:
anything missing in https://github.com/hyprwm/hyprland-wiki/pull/998 ?

vaxerski:
yes my attention was missing

PR #9569, opened 2025-03-08
fix(window): store floating size on close instead of resize

nnyyxxxx:
minor patch towards b80b64cd6c913f8c8ac820a1e4ca615a62ff958f

storing window size on resize is incorrect because the size can be lost (probably overwritten by something) if a resize event isn't sent


nnyyxxxx:
# Before this patch ( no resize event sent )

https://github.com/user-attachments/assets/236d0b78-16ef-46ed-9bf3-b6ed2f6b6614

# After this patch ( stored on close )

https://github.com/user-attachments/assets/27a852e7-5303-4b5d-8463-046c661d5a7c

nnyyxxxx:
@vaxerski this also fixes the issue we were talking about in the original mr

vaxerski:
nice wallpaper

vaxerski (approved):

PR #9578, opened 2025-03-08
Allow adjusting gap size for groupbar

phonetic112:
Adds two new variables: `gaps_out`, for changing the gap between the gradients and the window, and `gaps_in` for the gap between the gradients themselves. Basically just replaces `BAR_PADDING_OUTER_VERT` and `BAR_HORIZONTAL_PADDING`.


vaxerski (approved):
wiki mr needed

RichAyotte:
Nice! All that's left now is `BAR_TEXT_PAD` and I'll be able to drop my local stash.

PR #9599, opened 2025-03-11
keybinds: add sendkeystate

### LINKED ISSUE #9488: Support sending individual key events to non-focused windows
### Description

Hello, I'd like to ask if you could consider supporting sending keys with state to background windows.

Sending key click events is already possible using:

`$ hyprctl dispatch sendshortcut ,j ,class:firefox` (for sending "j" click to firefox without focusing it)

What I'm asking for is sending individual events `j down`, `j repeat`, `j up`, which I don't think is possible right now.

Some common use-cases:
- automating background tasks that need more than just clicking keys
- getting key events from a secondary keyboard, internet or other source and forwarding them to a window without focusing it
- various accessibility scripting

Not sure what the best way to implement this would be, but I assume adding a dispatcher that takes the key state as an additional parameter would work (`up`, `down`, `repeat`). Having an efficient way to do it like writing to a socket would be nice.

nnyyxxxx:
Adds a new dispatcher "sendkeystate" that allows sending key events with specific states (down,repeat,up) to windows 

Requires the window that's receiving the events to keep the focus or else the events will stop

fixes #9488 

https://github.com/hyprwm/hyprland-wiki/pull/1002


Kommynct:
I haven't tested it but I did read through and didn't see anywhere where it releases any held modifiers

If I bound something to ctrl+super+alt+whatever with this it'd just start pushing ctrl+super+alt+all of the inputs, I think. You should have to specify held modifiers in the list of buttons pressed.

Unless I missed something, I am an idiot with code.

nnyyxxxx:
it's not really possible to send input to windows without focusing them, anyways this is just a draft

> If I bound something to ctrl+super+alt+whatever with this it'd just start pushing ctrl+super+alt+all of the inputs, I think. You should have to specify held modifiers in the list of buttons pressed.

keydown holds the key down until keyup is received

nnyyxxxx:
this is essentially just a wrapper around sendshortcut with state specification

nnyyxxxx:
> it's not really possible to send input to windows without focusing them, anyways this is just a draft

unless the window is xwayland of course

Kommynct:
> it's not really possible to send input to windows without focusing them, anyways this is just a draft
> 
> > If I bound something to ctrl+super+alt+whatever with this it'd just start pushing ctrl+super+alt+all of the inputs, I think. You should have to specify held modifiers in the list of buttons pressed.
> 
> keydown holds the key down until keyup is received

You misunderstand me, 

bindr  = SUPER, i, exec, sleep 0.2 ; ( copyq clipboard | ydotool type -d 3 -H 2 -f - )

right now, this doesn't work... because super stays held down because I hit super i.

with your current setup, if i bound copyq clipboard to super i, it'd start randomly hitting all my super-keybinds.

I hope this makes more sense.

nnyyxxxx:
uhm, this is unrelated to ydotool? not sure i follow with that bind there

> bindr = SUPER, i, exec, sleep 0.2 ; ( copyq clipboard | ydotool type -d 3 -H 2 -f - )

nnyyxxxx:
> with your current setup, if i bound $(copyq clipboard) with sendkeystate to super i, it'd start randomly hitting all my super-keybinds.

you have to send a keyup, keydown stays down until keyup is received

nnyyxxxx:
> it would accidentally activate my lockscreen because super is being held as its sending the keystate.

should it not do that? seems intentional to me instead of accidental

Kommynct:
> > it would accidentally activate my lockscreen because super is being held as its sending the keystate.
> 
> should it not do that? seems intentional to me instead of accidental

No, because this makes it extremely inconvenient to use it to type things in another window.

If I wanted it to push super, I should have to add that as a keyup and keydown, it should ignore the modifiers that are already being held down to activate the shortcut, right?

nnyyxxxx:
i mean you're giving it keys to hold down, by design it should just hold down all of them until keyup is received

Kommynct:
> i mean you're giving it keys to hold down, by design it should just hold down all of them until keyup is received

You're not understanding me.

bindr = super,x,sendkeystate keydown v keyup v windowname
bindr = super,v,startmylockscreen

should send v to windowname, ignoring any syntax issues

instead, it will hit super+v because super is being held down to activate the shortcut and pull up my lockscreen.

If I want it to just send v to a window, there's no way to do that because it will instead activate my lockscreen.

nnyyxxxx:
just change the bind? should we really ignore the users existing binds when using sendkeystate, kinda takes control away from the user

what if the user actually wants to activate their lockscreen with sendkeystate

Kommynct:
> just change the bind? should we really ignore the users existing binds when using sendkeystate, kinda takes control away from the user
> 
> what if the user actually wants to activate their lockscreen with sendkeystate

If the user wants to activate their lockscreen they'd do

bindr = super,x,sendkeystate keydown super keydown v keyup v keyup super windowname

nnyyxxxx:
uhm can you give me an explanation on what it should do regarding modifier keys in one sentence?

ive been hardly reading these paragraphs you've been giving me, too many words

Kommynct:
it shouldn't pass modifiers that are in the bind, rather than in the sendkeystate

bindr = super,x,sendkeystate blah

instead of this sending blah to the window i want, it'll actually send super+blah because i'm holding down super to push the bind.

shiro:
He's saying the modifiers on the trigger side of the bind shouldn't be forwarded to the target window.
To do that you'd have to explicitly release the trigger-side mods, then keep track of that and ignore the `up` event for them when the mod is released.

I think that's a bit outside of the scope of what we want to do here though, the current behavior seems pretty good to me. (I mostly dispatch from scripts rather than keybinds, so take my opinion with a bunch of salt)

nnyyxxxx:
ig this can be undrafted, since sending continuous events to non focused windows is not really plausible ( focus would have to stay on that window )

nnyyxxxx:
@vaxerski bump

vaxerski (approved):
wiki mr needed

nnyyxxxx:
> wiki mr needed

https://github.com/hyprwm/hyprland-wiki/pull/1002

PR #9600, opened 2025-03-11
Extract shaders from source

UjinT34:
Extracts shaders from source code to reloadable and overridable assets. Shader assets can be overridden in `~/.config/hypr/shaders`.
Adds a simple shader `#include` processor.
Adds `hyprctl reloadshaders`  to reload shaders.
Fixes CM for some effects.
~~Adds `render:use_color_management` to disable proto and use old shaders. Requires restart.~~
Lowers es version requirement https://github.com/hyprwm/Hyprland/pull/9655
~~Fixes `render:cm_fs_passthrough` and changes the default to 2 - passthrough only HDR content. #9680~~

`GL_ARB_shading_language_include` is not used, it's enabled for ide shader linting.
`#include` supports only predefined set of filenames.
Some other shaders might need a CM support.
~~When testing a local build without installation shaders should be copied/linked to  `~/.config/hypr/shaders` (any installed shaders will have higher priority than builtin ones)~~
~~Might need meson install fix. Plugins relying on in-source shader constants will be broken #9619~~
`preBlurForCurrentMonitor` isn't called when CM settings change.

Ready

- [x] CM for borders
- [x] CM for shadows
- [x] CM for blur
- [x] Extract all shaders
- [x] Fix assets subfolders


vaxerski:
wat da hail is "cm for blur" and shadow

UjinT34:
Blur uses a main fb which can be in HDR and modifies it like it is sRGB. So main fb -> sRGB. Without it you get a gray smudge in HDR mode.
Shadows and borders are defined in sRGB. So sRGB -> main. Without it you get oversaturated shadows and borders.
Probably need to have 3 image descriptions: source, main fb and output. And keep main at something simple and precise like linear fp16. But main fb -> output will require a second draw pass if `decoration:screen_shader` is used.

abihf:
error at https://github.com/hyprwm/Hyprland/pull/9600/files#diff-90a04c13f5ee4e0258fd226a879957a956751d2906c1e5b13ccf47f7dc8a7dc5R1577 

should be
```diff
- if (shader == &m_RenderData.pCurrentMonData->m_shCM && !usingFinalShader && (texType == TEXTURE_RGBA || texType == TEXTURE_RGBX)) {
+ if (shader == &m_RenderData.pCurrentMonData->m_shaders->m_shCM && !usingFinalShader && (texType == TEXTURE_RGBA || texType == TEXTURE_RGBX)) {
```

vaxerski:
this is not really what I envisioned for extracting shaders from source. My idea was more along the lines of:
```cpp
CShader({
   .vertex = false,
   .type = TEXTURE,
   .cm = true,
   .rounding = true
 });
 ```
 which would build the string. But this is fine too, however:
 - I'd move the `version` to the source. When you create a shader, pass the string and version, and the shader class should append it.
 - the 320 ES vertex is still needed. We allow screen shaders to be 320 ES and some people use it. (Another reason for the above)
 
 > When testing a local build without installation shaders should be copied/linked to ~/.config/hypr/shaders
 
 Oi oi, does that mean hyprland will not launch without shaders installed to the system? That is not acceptable.

**Comment by UjinT34** [comment]
>     * I'd move the `version` to the source. When you create a shader, pass the string and version, and the shader class should append it.

One of the reasons to extract shaders was syntax highlighting and static checking. 

>     * the 320 ES vertex is still needed. We allow screen shaders to be 320 ES and some people use it. (Another reason for the above)

It's there and supported (both 300 and 320).

> > When testing a local build without installation shaders should be copied/linked to ~/.config/hypr/shaders
> 
> Oi oi, does that mean hyprland will not launch without shaders installed to the system? That is not acceptable.

I'll think of something. Probably won't be as clean as I'd like it to be.

**Comment by vaxerski** [comment]
> One of the reasons to extract shaders was syntax highlighting and static checking.

hm, fair, tho syntax highlighting for shaders aint that important

**Comment by UjinT34** [comment]
> hm, fair, tho syntax highlighting for shaders aint that important

Kind of important if that's your first shader %)

Added a script to generate includes for source code. Hyprland now starts without installed shaders. Still need to manually override installed shaders when testing a local build with some shader modifications.

**Comment by vaxerski** [comment]
> Added a script to generate includes for source code. Hyprland now starts without installed shaders. Still need to manually override installed shaders when testing a local build with some shader modifications.

Just don't install shaders.

**Comment by vaxerski** [comment]
`hyprctl reloadshaders` could be relegated to a separate MR. Ideally, I'd want this entire thing relegated as much as possible, because I want #9680 fixed for 0.48 but I don't want another heap of changes that could break that has to be tested for a few days

**Comment by UjinT34** [comment]
> `hyprctl reloadshaders` could be relegated to a separate MR. Ideally, I'd want this entire thing relegated as much as possible, because I want #9680 fixed for 0.48 but I don't want another heap of changes that could break that has to be tested for a few days

#9698 should be enough to fix foot

**Comment by 9oOzv** [comment]
Blur CM seems seems somewhat broken? Transparent areas on tiled windows whitish/grayish/unsaturated. Is this expected / known issue?

Floating windows seem fine.

**Comment by UjinT34** [comment]
> Blur CM seems seems somewhat broken? Transparent areas on tiled windows whitish/grayish/unsaturated. Is this expected / known issue?
> 
> Floating windows seem fine. Turning blur off seems fine.

Known. Tiled windows use an optimized codepath with a preblurred buffer which lacks some CM bits.

**Comment by UjinT34** [comment]
> `hyprctl reloadshaders` could be relegated to a separate MR. Ideally, I'd want this entire thing relegated as much as possible, 

Don't think it can be split in a sensible way. `hyprctl reloadshaders` is a small addition to everything else. Less than 10 lines of extra code which doesn't break anything. If this code itself has issues they can be easily solved by "don't use hyprctl reloadshaders". It's more a debugging/development feature

File: src/render/OpenGL.hpp [hunk 1]
```diff
@@ -92,21 +113,8 @@ struct SMonitorRenderData {
...
-    CShader m_shBLUR1;
-    CShader m_shBLUR2;
-    CShader m_shBLURPREPARE;
-    CShader m_shBLURFINISH;
-    CShader m_shSHADOW;
-    CShader m_shBORDER1;
-    CShader m_shGLITCH;
-    CShader m_shCM;
+    bool                 m_bShadersInitialized = false;
+    SP<SPreparedShaders> m_shaders;
```
vaxerski, line 117: while we're at it, `m_shaders` should probably be in CHyprOpenGL. I don't think there is a reason to split shaders between monitors.

src/render/OpenGL.cpp
```diff
@@ -1305,6 +1432,45 @@ void CHyprOpenGLImpl::renderTextureWithDamage(SP<CTexture> tex, const CBox& box,
     scissor(nullptr);
 }
 
+void CHyprOpenGLImpl::passCMUniforms(const CShader* shader, const NColorManagement::SImageDescription& imageDescription,
```
vaxerski, line 1435: const CShader&

assets/shaders/CM.frag
```diff
@@ -0,0 +1,55 @@
+#version 300 es
```
vaxerski, line 1: all .frag files: I'd prefer them in `src/render/shaders/glsl`, then compiled to `src/render/shaders`

src/render/OpenGL.cpp
```diff
@@ -856,215 +865,325 @@ void CHyprOpenGLImpl::setDamage(const CRegion& damage_, std::optional<CRegion> f
...
+        return SHADERS.at(filename);
+    throw std::runtime_error(std::format("Couldn't load shader {}", filename));
+}
+
+static void loadShaderInclude(const std::string& filename, std::map<std::string, std::string>& includes) {
+    includes.insert({filename, loadShader(filename)});
+}
+
+static void processShaderIncludes(std::string& source, const std::map<std::string, std::string>& includes) {
+    for (auto it = includes.begin(); it != includes.end(); ++it)
```
vaxerski, line 891: this for is missing {}

src/render/OpenGL.cpp
```diff
@@ -856,215 +865,325 @@ void CHyprOpenGLImpl::setDamage(const CRegion& damage_, std::optional<CRegion> f
...
+        RE2::Replace(&source, "#include \"" + it->first + "\"", it->second);
```
vaxerski, line 892: is there a reason for this to be a regex and not just `Hyprutils::String::replaceInString`?

src/render/OpenGL.cpp
```diff
@@ -856,215 +865,325 @@ void CHyprOpenGLImpl::setDamage(const CRegion& damage_, std::optional<CRegion> f
...
+        const auto TEXFRAGSRCEXT          = processShader("ext.frag", includes);
+        const auto FRAGBLUR1              = processShader("blur1.frag", includes);
+        const auto FRAGBLUR2              = processShader("blur2.frag", includes);
+
+        prog = createProgram(shaders->TEXVERTSRC, QUADFRAGSRC, isDynamic);
+        if (!prog)
+            return false;
+        shaders->m_shQUAD.program = prog;
+        getRoundingShaderUniforms(shaders->m_shQUAD);
+        shaders->m_shQUAD.proj      = glGetUniformLocation(prog, "proj");
```
vaxerski, line 995: you always have the shaders as a string here. While we're at this, why not make a map of (name, offset) and just scan for `^(uniform .{3,5} )(.*)(;)$`, get `$2` in the matches, and call `glGetUniformLocation`? :)

Don't store them all in a map, as those map lookups while rendering would be slow, but here, for the sake of cleanliness (and to avoid accidental "oops forgor") it would be nice.

(`CShader::getUniformLocation` and `m_muUniforms` can be nuked)

src/render/OpenGL.cpp
```diff
@@ -856,215 +865,325 @@ void CHyprOpenGLImpl::setDamage(const CRegion& damage_, std::optional<CRegion> f
...
+
+        Debug::log(ERR, "Shaders update failed: {}", e.what());
+        return false;
+    }
 
+    m_RenderData.pCurrentMonData->m_shaders             = shaders;
     m_RenderData.pCurrentMonData->m_bShadersInitialized = true;
 
     Debug::log(LOG, "Shaders initialized successfully.");
+    g_pHyprError->destroy();
```
vaxerski, line 1185: huh, why?

UjinT34: If `reloadshaders` fails it shows a notification. It removes this notification on subsequent successful reload. It might remove config parsing errors notification... but why anyone would want to call `reloadshaders` while there are config errors and be unhappy when this notifications disappears?
Shouldn't be a big deal. Other code uses `g_pHyprError->destroy()` also.
Needs some api modifications to remove only specific notification.

src/render/OpenGL.cpp
UjinT34: Should use `layout(location = ...)` with some enum and get rid of `getUniformLocation`  https://www.khronos.org/opengl/wiki/Layout_Qualifier_(GLSL)
Don't want to do it rn.
Those are called only on startup or reload. Performance issues shouldn't be a concern.

vaxerski: it's not about performance, just that we're reworking this code, might as well make it cleaner and more robust

src/render/OpenGL.cpp
vaxerski: ok

src/render/OpenGL.cpp
vaxerski: in general I dont mind if you dont wanna do it here tho, fair enough

vaxerski (approved):
thank you!

PlasmaPower:
This seems to cause a regression in #9795 for me according to a bisect. Not a huge deal but hopefully fixable

PR #9612, opened 2025-03-13
nix: mesa -> libgbm

andresilva:


nakoo:
I tested using github:hyprwm/Hyprland?ref=pull/9612/head, and it seems to be working well. 
Without this PR, Hyprland experiences significant performance degradation, such as frame drops.

PR #9615, opened 2025-03-14
XWayland: restore the abstract socket, make it optional, extend the default permissions for the regular X Unix socket

### LINKED ISSUE #9574: After the update, i cannot connect to XWayland session from another user
### Already reported ? *

### Regression?

Definitely a regression - something broke after update (requires bisect)

```
Hyprland 0.47.2 built from branch  at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: Sun Feb 2 00:47:17 2025
Tag: v0.47.2, commits: 5767
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.2


no flags were set


System Information:
System name: Linux
Node name: hamster
Release: 6.11.6-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 01 Nov 2024 03:30:35 +0000


GPU information:
00:02.0 VGA compatible controller [0300]: Intel Corporation RocketLake-S GT1 [UHD Graphics 750] [8086:4c8a] (rev 04) (prog-if 00 [VGA controller])
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 31 [Radeon RX 7900 XT/7900 XTX/7900 GRE/7900M] [1002:744c] (rev ce) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
LOGO=archlinux-logo


plugins:
```

</details>

### Description

I'm used to running some app from separate users. However, after the last updates X11 applications doesn't start with message like "Can't open display :0". The same applications, when started from original user, work well.
After rolling back to the last installed version `hyprland-0.46.2-8` X11 apps are working as usual.

* Arch Linux, all (except `linux-zen`) packages updated, tested Hyprland version is ~`0.47.2-2`~ the last main branch commit
* `DISPLAY=:0` is set
* `xhost +` does not affect
* I found that `chmod a+w /tmp/.X11-unix/X0` fix the issue, but it's not the solution. (upd however `xhost` ACL still works, maybe it is)

`git bisect` brought me to the next PR: #8874

### How to reproduce

```bash
> sudo useradd -m testuser
> xhost + # or `xhost SI:localuser:testuser`
> sudo -iu testuser bash
> xrandr
Can't open display :0

```

### Attach not paste

- [x] Hyprland config - `hyprctl systeminfo -c` (always include)
- [ ] Logs (might contain useful info such as errors)

[hyprland_config_dump.txt](https://github.com/user-attachments/files/19145395/hyprland_config_dump.txt)

PtaxLaine:
Resolve #9574.

##### XWayland
Roughly removing support for the abstract Unix domain sockets [broke](https://github.com/hyprwm/Hyprland/issues/9574) the `xhost` utility. Moreover, this way potentially has a few [other](https://gitlab.gnome.org/GNOME/mutter/-/issues/1613) [issues](https://gitlab.gnome.org/GNOME/mutter/-/issues/1454) repored in the GNOME tracker.

Keeping that in mind, I made a three changes
* The abstract X Unix domain socket has been restored.
* The abstract X Unix domain socket can now be turned off using the config variable `xwayland:create_abstract_socket`.
* The regular X Unix domain socket's permission mode is set to `0666`. This change ensures `xhost` continues to function when the abstract socket is disabled. And it's safe because XWayland must take care of actual socket ACL.

* On Linux (with disabled abstract sockets) we create two regular Unix sockets, e.g., `X0` and `X0_`. This corresponds to the current Hyprland behavior, but the second one is unnecessary. It's not a real bug, and fixing it would require rewriting bulk of code, so I've left it as-is.
* I'm not sure sould we set `xwayland:create_abstract_socket` to `false` by default or not.
   * #8874 says yes
   * GNOME [probably](https://gitlab.gnome.org/GNOME/mutter/-/issues/1613) says no
   * atm ~~`true`~~ **`false`** is the default value

It's ready. All changes has been tested.
Only the default value for `xwayland:create_abstract_socket` make me worried a little.

### DIFF
--- a/src/xwayland/Server.cpp
+++ b/src/xwayland/Server.cpp
@@ -20,6 +20,7 @@
 
 #include "Server.hpp"
 #include "XWayland.hpp"
+#include "config/ConfigValue.hpp"
 #include "debug/Log.hpp"
 #include "../defines.hpp"
 #include "../Compositor.hpp"
@@ -113,6 +114,8 @@ static std::string getSocketPath(int display, bool isLinux) {
 }
 
 static bool openSockets(std::array<CFileDescriptor, 2>& sockets, int display) {
+    static auto CREATEABSTRACTSOCKET = CConfigValue<Hyprlang::INT>("xwayland:create_abstract_socket");
+
     if (!ensureSocketDirExists())
         return false;
 
@@ -120,11 +123,20 @@ static bool openSockets(std::array<CFileDescriptor, 2>& sockets, int display) {
     std::string path;
 
 #ifdef __linux__
-    // cursed...
-    addr.sun_path[0] = 0;
-    path             = getSocketPath(display, true);
-    strncpy(addr.sun_path + 1, path.c_str(), path.length() + 1);
+    if (*CREATEABSTRACTSOCKET) {
+        // cursed...
+        // but is kept as an option for better compatibility
+        addr.sun_path[0] = 0;
+        path             = getSocketPath(display, true);
+        strncpy(addr.sun_path + 1, path.c_str(), path.length() + 1);
+    } else {
+        path = getSocketPath(display, false);
+        strncpy(addr.sun_path, path.c_str(), path.length() + 1);
+    }
 #else
+    if (*CREATEABSTRACTSOCKET) {
+        Debug::log(WARN, "The abstract XWayland Unix domain socket might be used only on Linux systems. A regular one'll be created insted.");
+    }
     path = getSocketPath(display, false);
     strncpy(addr.sun_path, path.c_str(), path.length() + 1);
 #endif
@@ -227,12 +239,10 @@ CXWaylandServer::~CXWaylandServer() {
     safeRemove(lockPath);
 
     std::string path;
-#ifdef __linux__
-    path = getSocketPath(display, true);
-#else
-    path = getSocketPath(display, false);
-#endif
-    safeRemove(path);
+    for (bool isLinux : {true, false}) {
+        path = getSocketPath(display, isLinux);
+        safeRemove(path);
+    }
 }
 
 void CXWaylandServer::die() {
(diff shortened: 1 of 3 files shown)


vaxerski:
I think false is a better default

PtaxLaine:
@vaxerski done

src/xwayland/Server.cpp
```diff
@@ -220,12 +253,10 @@ CXWaylandServer::~CXWaylandServer() {
     safeRemove(lockPath);
 
     std::string path;
-#ifdef __linux__
-    path = getSocketPath(display, true);
-#else
-    path = getSocketPath(display, false);
-#endif
-    safeRemove(path);
+    for (bool isLinux : {true, false}) {
```
vaxerski, line 256: unrelated change

PtaxLaine: @vaxerski 
The original code doesn't do its job properly on non-Linux systems. And since we disable the abstract socket, it doesn't work on Linux either.
Please check the `openSockets` function. When the abstract socket is disabled we always create two regular sockets: `getSocketPath(_, true)` & `getSocketPath(_, false)`. This corresponds to the [current](https://github.com/hyprwm/Hyprland/blob/main/src/xwayland/Server.cpp#L115) Hyprland behavior.
Without this change, we'll never properly remove one of them. I think that this change is not ideal but must exist.

PtaxLaine: > I think that this change is not ideal but must exist.

At least, until we fix the next:

> On Linux (with disabled abstract sockets) we create two regular Unix sockets, e.g., X0 and X0_. This corresponds to the current Hyprland behavior, but the second one is unnecessary. It's not a real bug, and fixing it would require rewriting bulk of code, so I've left it as-is.

vaxerski: ah, my bad

vaxerski (approved):

PR #9617, opened 2025-03-14
Groupbar: Add options for setting group bar title font weight (and indicator gap)

mhydock:
Adds config options for setting the font weight of active and inactive group bar title text. The option is set to be a string, but can also be an integer. Valid values are based on Pango weight enums (`thin`, `ultralight`, `light`, `semilight`, `book`, `normal`, `medium`, `semibold`, `bold`, `ultrabold`, `heavy`, `ultraheavy`, defaults to `normal` for both).

Also adds a config option to set the space between the group bar indicator and the title/gradient (defaults to 0).

No

Ready to merge

### DIFF
--- a/src/config/ConfigDataValues.hpp
+++ b/src/config/ConfigDataValues.hpp
@@ -4,10 +4,10 @@
 #include <vector>
 
 enum eConfigValueDataTypes : int8_t {
-    CVD_TYPE_INVALID    = -1,
-    CVD_TYPE_GRADIENT   = 0,
-    CVD_TYPE_CSS_VALUE  = 1,
-    CVD_TYPE_STR_OR_INT = 2,
+    CVD_TYPE_INVALID     = -1,
+    CVD_TYPE_GRADIENT    = 0,
+    CVD_TYPE_CSS_VALUE   = 1,
+    CVD_TYPE_FONT_WEIGHT = 2,
 };
 
 class ICustomConfigValueData {
@@ -143,13 +143,13 @@ class CFontWeightConfigValueData : public ICustomConfigValueData {
     CFontWeightConfigValueData() = default;
     CFontWeightConfigValueData(const char* weight) {
         std::string strWeight = weight;
-        value = parseWeight(strWeight);
+        value                 = parseWeight(strWeight);
     }
 
     int64_t                       value;
 
     virtual eConfigValueDataTypes getDataType() {
-        return CVD_TYPE_STR_OR_INT;
+        return CVD_TYPE_FONT_WEIGHT;
     }
 
     virtual std::string toString() {
--- a/src/config/ConfigDataValues.hpp
+++ b/src/config/ConfigDataValues.hpp
@@ -4,9 +4,10 @@
 #include <vector>
 
 enum eConfigValueDataTypes : int8_t {
-    CVD_TYPE_INVALID   = -1,
-    CVD_TYPE_GRADIENT  = 0,
-    CVD_TYPE_CSS_VALUE = 1
+    CVD_TYPE_INVALID    = -1,
+    CVD_TYPE_GRADIENT   = 0,
+    CVD_TYPE_CSS_VALUE  = 1,
+    CVD_TYPE_STR_OR_INT = 2,
 };
 
 class ICustomConfigValueData {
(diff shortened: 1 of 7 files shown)


src/config/ConfigManager.cpp
```diff
@@ -145,6 +145,67 @@ static void configHandleGapDestroy(void** data) {
...
+    return w_i;
+}
+
+static Hyprlang::CParseResult configHandleFontWeightSet(const char* VALUE, void** data) {
+    std::string V = VALUE;
+
+    if (!*data)
+        *data = new int;
+
+    const auto             DATA = reinterpret_cast<Hyprlang::INT*>(*data);
```
vaxerski, line 191: this will not pass tests. custom types in hyprland must be a `ICustomConfigValueData` (see `SConfigOptionDescription::jsonify()`)

mhydock: before i go and make a CustomConfigValue class to represent "string or int", would it perhaps be more valuable to modify `CConfig::addConfigValue` in hyprlang to accept an optional handler function? this would allow more flexibility in how values are parsed, and/or allow `configStringToInt` to be decomposed into smaller, more precise parsing functions. on a side note, is there a reason why `Hyprlang::BOOL` doesn't exist?

vaxerski: what? just make a new class inheriting the iface

mhydock: ok done. can you address my previous questions now? also, you mentioned that my change wouldn't "pass tests". is there an automated testing suite attached to this project? and if so, where is it and how do i run it? is this information in the wiki?

vaxerski: it's in the works in #9297

src/config/ConfigDataValues.hpp
```diff
@@ -136,3 +137,18 @@ class CCssGapData : public ICustomConfigValueData {
         return std::format("{} {} {} {}", top, right, bottom, left);
     }
 };
+
+class CStringOrInt : public ICustomConfigValueData {
```
vaxerski, line 141: I don't like this. Why not make this class a `CFontWeightConfigValueData` and have it take a `const char*` in the ctor and set its value on that?

mhydock: Class renamed as requested. The reason I named it like I did was because I'd hoped to reuse it for other config values that might benefit from having a human-readable value in the config file, but an integer representation in the code. Admittedly, this feels like a sloppy solution, and I still think there's _some_ value to allowing `CConfig::addConfigValue` to take an optional parse function. For now though, this is good enough.

src/config/ConfigDataValues.hpp
```diff
@@ -151,4 +155,41 @@ class CStringOrInt : public ICustomConfigValueData {
     virtual std::string toString() {
         return std::format("{}", value);
     }
+
+    static int parseWeight(const std::string& weight) {
```
vaxerski, line 159: any reason for the static? I'd omit it. Also, why is it public?

src/config/ConfigDataValues.hpp
```diff
@@ -136,3 +137,59 @@ class CCssGapData : public ICustomConfigValueData {
         return std::format("{} {} {} {}", top, right, bottom, left);
     }
 };
+
+class CFontWeightConfigValueData : public ICustomConfigValueData {
+  public:
+    CFontWeightConfigValueData() = default;
+    CFontWeightConfigValueData(const char* weight) {
+        std::string strWeight = weight;
+        value                 = parseWeight(strWeight);
```
vaxerski, line 146: also this local is redundant, you can just pass weight

src/config/ConfigDataValues.hpp
```diff
@@ -136,3 +137,41 @@ class CCssGapData : public ICustomConfigValueData {
...
 };
+
+class CFontWeightConfigValueData : public ICustomConfigValueData {
+  public:
+    CFontWeightConfigValueData() = default;
+    CFontWeightConfigValueData(const char* weight) {
+        parseWeight(weight);
+    }
+
+    int64_t                       value;
```
vaxerski, line 148: please initialize this primitive to something reasonable.

vaxerski (commented):
rest lgtm

vaxerski (dismissed):
thanks! Wiki mr needed

vaxerski:
ah already there

vaxerski (approved):

PR #9620, opened 2025-03-14
dmabuf: pop buffer on failure

gulafaran:
ensure it doesnt permanently gets stuck in the container on failure, pop it from the container.

### DIFF
--- a/src/protocols/LinuxDMABUF.cpp
+++ b/src/protocols/LinuxDMABUF.cpp
@@ -221,6 +221,7 @@ void CLinuxDMABUFParamsResource::create(uint32_t id) {
 
     if UNLIKELY (!buf->good() || !buf->buffer->success) {
         resource->sendFailed();
+        PROTO::linuxDma->m_vBuffers.pop_back();
         return;
     }
 


vaxerski (commented):
will this not cause the resource to be destroyed which is wrong?

gulafaran:
> will this not cause the resource to be destroyed which is wrong?

hm `CLinuxDMABUFParamsResource` , create or createImmed calls this. it creates a `CLinuxDMABuffer` which in turn creates a `CDMABuffer` and it creates a `CWLBufferResource` from a `CWlBuffer` , so depends on which resource you are concerned about here? cant say i have hit this failure merely noticed it not being removed on failure which seems like it should be leaking/remaining forever in the container if such case happend. the `CLinuxDMABUFParamsResource` shouldnt be destroyed, but the cwlbuffer and the client might be but if the failure happends perhaps thats what should happend?. or does all of these trips down the buffering lane mean it eventually calls the `CLinuxDMABUFParamsResource` setondestroy?

vaxerski:
generally if we create an object, we should wait for the client to ask us to destroy it

gulafaran:
> generally if we create an object, we should wait for the client to ask us to destroy it

true that, oh well lets leave this as a maybe situation. il see if i can make some failure of a client that triggers this and see what actually happends. :P

PlasmaPower:
The other similar protocols all remove it on failure though. I don't think the client is expected to clean it up if the request to create it fails, though I could be wrong. If I am wrong I think we need to fix a lot of other protocol impls.

PlasmaPower:
In particular, unless create_immed is used, on failure we never even give the client the handle to call destroy on https://github.com/hyprwm/Hyprland/blob/67e1f17a5e652d36dec64d0c1310e8a8477eb5a8/src/protocols/LinuxDMABUF.cpp#L229

In the create_immed case failure is basically undefined behavior per the spec

vaxerski:
> In the create_immed case failure is basically undefined behavior per the spec

feel free to report it to the fdo guys, I can't, I am banned

PlasmaPower:
No I mean the spec itself basically says it's undefined behavior:

> Upon import failure, either of the following may happen, as seen fit by the implementation: - the client is terminated with one of the following fatal protocol errors: - INCOMPLETE, INVALID_FORMAT, INVALID_DIMENSIONS, OUT_OF_BOUNDS, in case of argument errors such as mismatch between the number of planes and the format, bad format, non-positive width or height, or bad offset or stride. - INVALID_WL_BUFFER, in case the cause for failure is unknown or platform specific. - the server creates an invalid wl_buffer, marks it as failed and sends a 'failed' event to the client. The result of using this invalid wl_buffer as an argument in any request by the client is defined by the compositor implementation.

The last sentence in particular means it'd be fine to delete the buffer here I think. And as I mentioned earlier for the non-immediate create case, I think we need to delete the buffer for sure.

https://wayland.app/protocols/linux-dmabuf-v1#zwp_linux_buffer_params_v1:request:create_immed

vaxerski:
I see. Alright.

vaxerski (approved):

PR #9645, opened 2025-03-16
managers: Use primary backend for cursor swapchain

PlasmaPower:

This PR completes the AQ PR https://github.com/hyprwm/aquamarine/pull/159 by using the correct backend for the cursor swapchain. Without this one, the monitor renders fine, but hardware cursors don't work. This PR also uses the new getPrimary method of AQ's IBackendImplementation from the aforementioned AQ PR.

Nope, I've confirmed this works perfectly locally

Ready for merging


PlasmaPower:
I updated the aq requirement in meson.build and CMakeLists.txt to 0.8.0, but I'm not on nix so I don't think I can update the flake.lock

PR #9646, opened 2025-03-17
misc: Fix compiler warnings

PlasmaPower:
Fixes all compiler warnings I encounter building locally

There's a couple of changes I wasn't sure about that I'll comment on via github comments.

Ready for merging

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1240,7 +1240,7 @@ void CMonitor::setSpecialWorkspace(const PHLWORKSPACE& pWorkspace) {
     }
 
     g_pEventManager->postEvent(SHyprIPCEvent{"activespecial", pWorkspace->m_szName + "," + szName});
-    g_pEventManager->postEvent(SHyprIPCEvent{"activespecialv2", pWorkspace->m_iID + "," + pWorkspace->m_szName + "," + szName});
+    g_pEventManager->postEvent(SHyprIPCEvent{"activespecialv2", std::to_string(pWorkspace->m_iID) + "," + pWorkspace->m_szName + "," + szName});
 
     g_pHyprRenderer->damageMonitor(self.lock());
 
--- a/src/plugins/PluginAPI.hpp
+++ b/src/plugins/PluginAPI.hpp
@@ -53,11 +53,14 @@ struct SVersionInfo {
     std::string commits;
 };
 
-#define APICALL extern "C++"
+#define APICALL extern "C"
 #define EXPORT  __attribute__((visibility("default")))
 #define REQUIRED
 #define OPTIONAL
 #define HANDLE void*
+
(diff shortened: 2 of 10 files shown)


src/helpers/Monitor.cpp
```diff
@@ -1240,7 +1240,7 @@ void CMonitor::setSpecialWorkspace(const PHLWORKSPACE& pWorkspace) {
     }
 
     g_pEventManager->postEvent(SHyprIPCEvent{"activespecial", pWorkspace->m_szName + "," + szName});
-    g_pEventManager->postEvent(SHyprIPCEvent{"activespecialv2", pWorkspace->m_iID + "," + pWorkspace->m_szName + "," + szName});
+    g_pEventManager->postEvent(SHyprIPCEvent{"activespecialv2", std::to_string(pWorkspace->m_iID) + "," + pWorkspace->m_szName + "," + szName});
```
PlasmaPower, line 1243: Previously, the C++ compiler was interpreting this as `","[pWorkspace->m_iID]` . I think this would either result in garbage or the event being identical to the v1 activespecial event. I don't think this should be left as undefined behavior, but we could either solidify it as identical to the v1 event or merge this code which fixes it to what I assume was intended.

src/plugins/PluginAPI.hpp
```diff
@@ -53,7 +53,7 @@ struct SVersionInfo {
     std::string commits;
 };
 
-#define APICALL extern "C"
+#define APICALL extern "C++"
```
PlasmaPower, line 56: I'm also not confident on this change, but it seems right to me. These functions definitely aren't C ABI compatible because they have all sorts of C++ types like `std::string`, so I think this is the correct change, but I'm not 100% sure what these extern function modifiers are doing.

src/helpers/Monitor.cpp
vaxerski: ye

src/plugins/PluginAPI.hpp
vaxerski: this is wrong. The extern C is to disable name mangling

src/plugins/PluginAPI.hpp
```diff
@@ -59,6 +59,9 @@ struct SVersionInfo {
 #define OPTIONAL
 #define HANDLE void*
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wreturn-type-c-linkage"
```
PlasmaPower, line 63: This actually works for clang too despite its name https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

vaxerski: even if it didn't, it'd likely be ignored

vaxerski (approved):

PR #9648, opened 2025-03-17
fix hw cursor not being committed cause of same main buffer in DS

ikalco:
fixes some stuttering during direct scanout

yes


PlasmaPower:
Can confirm this fixes direct scanout to alacritty for me. Without this PR cursor movements get dropped and the cursor only updates when I type something into alacritty causing it to render new frames. Otherwise the cursor just stays in the same position on-screen even as I move the mouse.

vaxerski (approved):

PR #9655, opened 2025-03-17
Autodetect GLES version

System64fumo:
This should automatically use the highest available version of GLES.
Should be fine for GLES 3.x es since nothing major changed between the versions, IE it should be compatible.

Haven't tested screen shaders or cm since i don't have one available on hand but should be all good according to logs.
I feel like this could be done better, LMK if there's any changes needed.

Ready to be merged.

Final words:
Vaxry if you're going to do Embedded System stuff at least support it properly.
As a man who plays with this stuff daily i feel like i have a say on this.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -154,7 +154,7 @@ void CHyprOpenGLImpl::initEGL(bool gbm) {
     attrs.push_back(EGL_CONTEXT_MAJOR_VERSION);
     attrs.push_back(3);
     attrs.push_back(EGL_CONTEXT_MINOR_VERSION);
-    attrs.push_back(2);
+    attrs.push_back(0);
 #else
     attrs.push_back(EGL_CONTEXT_CLIENT_VERSION);
     attrs.push_back(2);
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
this is a horrible idea, don't do this.

System64fumo:
I know that this is a terrible idea.. But at least this would increase compatibility.
There aren't any major changes between gl 3.x es versions afaik so this should be safe to do, Plus should anything go wrong it would just get disabled and complain in the logs.

Any particular reason you're against this? (Other than unpredictability)
gl 2.0 es is unaffected by this change btw.

vaxerski:
this is just wrong, and there is no reason to put version to 320 if the shader does well at 300. Instead, just make it 300 if we don't use any 310 / 320 features.

System64fumo:
You could do that, But PR guidelines state "no hacks" imo that's not a solution but a hack
It would also negate the minor performance improvements that newer version of gles offer, Not that it makes a huge difference but if it's possible to use whatever your card is compatible with then why not?

System64fumo:
That being said i've just tested these changes with some shaders on a few of my hardware and it works on all of them so i'm gonna say this works afaik.

Tested on:
Samsung Galaxy E5
OpenGL ES: 3.0

Orange Pi 5 Plus
OpenGL ES: 3.1

OnePlus 6T
OpenGL ES: 3.2

So as far as i'm aware shaders and color management should work fine everywhere.

UjinT34:
Does it solve any issue caused by "incorrect" version? "Works fine" or "we hope it'll work fine" doesn't mean it'll work better.
I am not saying that we shouldn't change this stuff. We should have something measurable or comparable to make a decision.

System64fumo:
> Does it solve any issue caused by "incorrect" version?

Yes it allows the use of shaders on non 3.2 capable machines

> "Works fine" or "we hope it'll work fine" doesn't mean it'll work better.

I mean so far this got rid of that annoying warning, Prevents compilation errors, And allows for shader usage which i guess is better than not having that.

> We should have something measurable or comparable to make a decision.

As stated above, I've tested this on a few devices that have different versions of supported GLES.
Everything worked fine there.
And should continue working in the future as there really aren't that many differences between GLES versions,
It mostly comes down to hardware support, Optimizations, And extensions.

If you were talking about the performance hit then yeah sure idrc about forcing 3.0 if 3.2 is supported.
But this is the better approach to this problem, Up to vaxry to decide.

leiserfg:
Isn't this the same as setting `#version 300` to all 3** shaders? I don't think the version of the glsl will change performance, the driver is the same.

System64fumo:
Fine, Screw it.
Lazy hack it is!

System64fumo:
I went to buy milk but now i'm back with the commit.
This now uses GLES 3.0 instead of auto detecting.

Dear @vaxerski have a look will ya?

vaxerski:
this is still wrong. You can't change the gles version we run, as we offer 320 compat for screen shaders.

UjinT34:
Seems to work fine with 300 es #9600. 320 was picked because it was already used for custom screen shaders.

PR #8143, opened 2024-10-17
feat: drag the window when the cursor starts moving

myamusashi:
#### Describe your PR, what does it fix/add?
When cursor is clicked on a window but the cursor is not moving, the window will not float until cursor starts moving
#8099

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Nah.

#### Is it ready for merging, or does it need work?
Not ready

PR #9658, opened 2025-03-18
input: Add drag_threshold and related bind flags to distinguish clicks from bindm drags with the same button

### LINKED ISSUE #9631: mutually exclusive bindm and bind on the same mouse button
### Description

It would be useful to have a way to prevent bindm and bind on the same mouse button from firing together.

I have some extra mouse buttons I want use for window management. I want to have mutually exclusive actions for clicks and drags because while it's the same button it's different input paradigms.  In a bindm the mouse button is more like a modifier for the motion than it is a key press.  An implementation would mean bind's against a mouse button would not fire if there was also motion while the button is down and bindm's would only trigger once motion is initiated during the button being held.  This indicates that flag 'release' should be a requirement for the bind to have this functionality.

One possible way to handle this is to implement the longPress flag on bindm, though it's not the best UX imho.

Dregu:
#### New features

- Added `binds:drag_threshold` to yoink windows with mouse binds only after moving a few pixels
- Added `bindc` / `bindg` (**c**lick / dra**g**) flags for release binds that check if inside/outside drag_threshold on release

Threshold handling is slightly different whether there is a bindm on the same button or not:

- If there is no bindm on the same button, any amount of movement that ends inside the threshold will trigger a click
- If bindm exists and has been triggered, returning inside the threshold will **not** trigger a click (triggers drag)

c and g are mutually exclusive. I don't really have a use for g, but you could probably make a script that checks if a window was dragged to the top of the screen to fullscreen it for example.

Simple example that adds more window handling to the same 2 buttons:

```
binds {
    drag_threshold = 20
}

bindm = SUPER, mouse:272, movewindow
bindc = SUPER, mouse:272, togglefloating
bindm = SUPER, mouse:273, resizewindow
bindc = SUPER, mouse:273, fullscreen
```

#### Thoughts

The default threshold = 0 should work like current default behavior, i.e. windows are grabbed immediately on mousedown. With threshold > 0 yoinking happens later on mousemove.

This probably obsoletes #8143 and fixes #9631.

#### Is it ready for merging

- Can't think of anything else to add
- I have been daily driving this for exactly one day with the example binds, seems fine to me
- Definitely made this with horse blinders on and this is my first PR, could've messed up something completely unrelated to my goals
- Wiki PR (needs updating still): https://github.com/hyprwm/hyprland-wiki/pull/1009

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1302,6 +1302,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_BOOL,
         .data        = SConfigOptionDescription::SBoolData{false},
     },
+    SConfigOptionDescription{
+        .value       = "binds:drag_threshold",
+        .description = "Movement threshold in pixels for window dragging and c/g bind flags. 0 to disable and grab on mousedown.",
+        .type        = CONFIG_OPTION_INT,
+        .data        = SConfigOptionDescription::SRangeData{0, 0, INT_MAX},
+    },
 
     /*
      * xwayland:
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -228,7 +228,8 @@ bool IHyprLayout::onWindowCreatedAutoGroup(PHLWINDOW pWindow) {
 }
 
 void IHyprLayout::onBeginDragWindow() {
-    const auto DRAGGINGWINDOW = g_pInputManager->currentlyDraggedWindow.lock();
+    const auto DRAGGINGWINDOW  = g_pInputManager->currentlyDraggedWindow.lock();
+    const auto DRAGTHRESHOLDSQ = std::pow(*CConfigValue<Hyprlang::INT>("binds:drag_threshold"), 2);
 
     m_iMouseMoveEventCount = 1;
     m_vBeginDragSizeXY     = Vector2D();
@@ -240,38 +241,11 @@ void IHyprLayout::onBeginDragWindow() {
         return;
     }
 
-    if (DRAGGINGWINDOW->isFullscreen()) {
-        Debug::log(LOG, "Dragging a fullscreen window");
-        g_pCompositor->setWindowFullscreenInternal(DRAGGINGWINDOW, FSMODE_NONE);
-    }
-
-    const auto PWORKSPACE = DRAGGINGWINDOW->m_pWorkspace;
-
-    if (PWORKSPACE->m_bHasFullscreenWindow && (!DRAGGINGWINDOW->m_bCreatedOverFullscreen || !DRAGGINGWINDOW->m_bIsFloating)) {
-        Debug::log(LOG, "Rejecting drag on a fullscreen workspace. (window under fullscreen)");
-        g_pKeybindManager->changeMouseBindMode(MBIND_INVALID);
+    // Try to pick up dragged window now if drag_threshold is disabled
+    // or at least update dragging related variables for the cursors
+    g_pInputManager->m_bDragThresholdReached = DRAGTHRESHOLDSQ <= 0;
+    if (updateDragWindow())
         return;
-    }
-
-    DRAGGINGWINDOW->m_bDraggingTiled = false;
-
-    m_vDraggingWindowOriginalFloatSize = DRAGGINGWINDOW->m_vLastFloatingSize;
-
-    if (!DRAGGINGWINDOW->m_bIsFloating) {
-        if (g_pInputManager->dragMode == MBIND_MOVE) {
-            DRAGGINGWINDOW->m_vLastFloatingSize = (DRAGGINGWINDOW->m_vRealSize->goal() * 0.8489).clamp(Vector2D{5, 5}, Vector2D{}).floor();
-            changeWindowFloatingMode(DRAGGINGWINDOW);
-            DRAGGINGWINDOW->m_bIsFloating    = true;
-            DRAGGINGWINDOW->m_bDraggingTiled = true;
-
-            *DRAGGINGWINDOW->m_vRealPosition = g_pInputManager->getMouseCoordsInternal() - DRAGGINGWINDOW->m_vRealSize->goal() / 2.f;
-        }
-    }
-
-    m_vBeginDragXY         = g_pInputManager->getMouseCoordsInternal();
-    m_vBeginDragPositionXY = DRAGGINGWINDOW->m_vRealPosition->goal();
-    m_vBeginDragSizeXY     = DRAGGINGWINDOW->m_vRealSize->goal();
-    m_vLastDragXY          = m_vBeginDragXY;
 
     // get the grab corner
     static auto RESIZECORNER = CConfigValue<Hyprlang::INT>("general:resize_corner");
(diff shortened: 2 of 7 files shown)


alaricljs:
For what it's worth, lgtm.  Yes on the wiki PR.
I'll test this asap and throw in my 2p since I want it pretty badly ;)

alaricljs:
It does the thing!  The only recommendation I have is a code change outside the specific scope of this, so maybe a separate PR?  I'd want to see window dragging side effects removed, per @vaxerski's usual preferences this would be behind another config option.  In particular I'm talking about the focusWindow() and changeWindowZOrder() calls.  With those disabled, the user has total control of how a window drag impacts that window via stacked keybinds.

vaxerski:
if you don't have a use for g, and can't even think of one, why add it?

Dregu:
But I did think if one and I wrote it up there as an example too?

I also thought of another one I'm actually using though, to maximize the amount of things I can put on mouse buttons, `bindig = , Super_R, exec ...` launches a program when slighty dragged by itself, but can also be used as a modifier for the other buttons in the simple example. (I have a Super_R key on my mouse.)

vaxerski:
I understand c and no c/g. What does g do?

vaxerski:
will g trigger instantly once the threshold is passed or when the mouse is released?

Dregu:
bindm will trigger (i.e. starts actually moving the dragged window) immediately when the threshold is passed. bindg will trigger when button is released.

Dregu:
I made a video to demonstrate some of the features with dummy notifications. It starts out with drag_threshold disabled (not an optimal way to use the new bind flags, but supported) and then drag_threshold turned to 100 to exaggerate the inputs required.

https://github.com/user-attachments/assets/f6e9fa35-1150-4445-b414-85e829a8e718

vaxerski:
now I getchu, thanks.

src/config/ConfigDescriptions.hpp
```diff
@@ -547,6 +547,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_INT,
         .data        = SConfigOptionDescription::SRangeData{1, 0, 2},
     },
+    SConfigOptionDescription{
+        .value       = "input:drag_threshold",
```
vaxerski, line 551: better place would be `binds:`

src/layout/IHyprLayout.cpp
```diff
@@ -548,14 +555,43 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
...
+    if (*PDRAGTHRESHOLD > 0 && !g_pInputManager->m_bDragThresholdReached) {
+        if (m_vBeginDragXY.distance(mousePos) <= *PDRAGTHRESHOLD && m_vBeginDragXY == m_vLastDragXY)
+            return;
+
+        if (DRAGGINGWINDOW->isFullscreen()) {
+            Debug::log(LOG, "Dragging a fullscreen window");
+            g_pCompositor->setWindowFullscreenInternal(DRAGGINGWINDOW, FSMODE_NONE);
+        }
+
+        const auto PWORKSPACE = DRAGGINGWINDOW->m_pWorkspace;
```
vaxerski, line 577: looks convoluted and duped. Why not extract the logic for "picking up" the window to a method and call that?

src/managers/KeybindManager.cpp
```diff
@@ -453,13 +453,12 @@ bool CKeybindManager::onKeyEvent(std::any event, SP<IKeyboard> pKeyboard) {
...
-        .modmaskAtPressTime = MODS,
-        .sent               = true,
-        .submapAtPress      = m_szCurrentSelectedSubmap,
-    };
+    const auto KEY = SPressedKeyWithMods{.keysym             = keysym,
+                                         .keycode            = KEYCODE,
+                                         .modmaskAtPressTime = MODS,
+                                         .sent               = true,
+                                         .submapAtPress      = m_szCurrentSelectedSubmap,
+                                         .posAtPress         = g_pInputManager->getMouseCoordsInternal()};
```
vaxerski, line 461: put a , after the last arg

src/managers/KeybindManager.cpp
```diff
@@ -735,6 +736,14 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
                 found = true; // suppress the event
                 continue;
             }
+
+            // Require mouse to stay inside drag_threshold for clicks, outside for drags
+            // Check if either a mouse bind has triggered or currently over the threshold (maybe there is no mouse bind on the same key)
+            const auto THRESHOLDREACHED = key.posAtPress.distance(g_pInputManager->getMouseCoordsInternal()) > *PDRAGTHRESHOLD;
```
vaxerski, line 742: prefer distanceSq

src/layout/IHyprLayout.cpp
```diff
@@ -548,14 +555,43 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
...
 
     // Window invalid or drag begin size 0,0 meaning we rejected it.
     if ((!validMapped(DRAGGINGWINDOW) || m_vBeginDragSizeXY == Vector2D())) {
         g_pKeybindManager->changeMouseBindMode(MBIND_INVALID);
         return;
     }
 
+    // Yoink window here instead if using drag_threshold
+    if (*PDRAGTHRESHOLD > 0 && !g_pInputManager->m_bDragThresholdReached) {
+        if (m_vBeginDragXY.distance(mousePos) <= *PDRAGTHRESHOLD && m_vBeginDragXY == m_vLastDragXY)
```
vaxerski, line 569: prefer distanceSq

src/layout/IHyprLayout.cpp
Dregu: I added a new method that both updates the drag origin related values and tries to yoink the window if the threshold is reached/disabled. Pretty tilted right now, don't know if anything makes sense or if it's in the right place.

At least using move and resize with a big threshold feels very smooth now that the DragXY values are updated again when the threshold is reached...

src/layout/IHyprLayout.cpp
```diff
@@ -228,7 +228,8 @@ bool IHyprLayout::onWindowCreatedAutoGroup(PHLWINDOW pWindow) {
 }
 
 void IHyprLayout::onBeginDragWindow() {
-    const auto DRAGGINGWINDOW = g_pInputManager->currentlyDraggedWindow.lock();
+    const auto DRAGGINGWINDOW  = g_pInputManager->currentlyDraggedWindow.lock();
+    const auto DRAGTHRESHOLDSQ = std::pow(*CConfigValue<Hyprlang::INT>("binds:drag_threshold"), 2);
```
vaxerski, line 232: illegal. Make the config variable static and deref that.

src/managers/KeybindManager.hpp
```diff
@@ -62,6 +64,7 @@ struct SPressedKeyWithMods {
     uint32_t     modmaskAtPressTime = 0;
     bool         sent               = false;
     std::string  submapAtPress      = "";
+    Vector2D     posAtPress         = {};
```
vaxerski, line 67: mousePosAtPress is clearer

Dregu:
I included the center floating fullscreen fix from #9780 cause it's a good fix and poking the same code.

vaxerski:
> I included the center floating fullscreen fix from https://github.com/hyprwm/Hyprland/pull/9780 cause it's a good fix and poking the same code.

uhhh, don't, please.

vaxerski:
don't meant don't include another MR, I don't know why you closed it.

Dregu:
I tried to open the editor but I just didn't feel like it any more 🤷🏻‍♂️, but you're free to make it like you want it yourself. The thing I included was a two line fix that would get mowed by this PR and it was bugging me.

PR #9668, opened 2025-03-18
ci: Fail on warnings

PlasmaPower:
Changes CI to require that no warnings are emitted by the compiler. This will help prevent issues like the `activespecialv2` bug fixed in https://github.com/hyprwm/Hyprland/pull/9646

This PR also fixes a couple of color protocol warnings which were introduced since #9646, and also fixes the ignoring of `-Wreturn-type-c-linkage` which it turns out is a warning only Clang has.

The main disadvantage of this is that future compiler versions introducing new warnings will break CI. However, I think this is worth it. This change can be temporarily reverted if a lot of warnings are introduced while they're being fixed, and I think it's worth fixing warnings in general. Most of the time they're fine as-is, but the bugs they find make up for that.

Ready for merging


vaxerski (approved):
fair enough

Issue #9636, opened 2025-03-16
Stuttering in animations after commit 6ffde36


goggi:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

Definitely a regression - something broke after update (requires bisect)

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.47.0 built from branch  at commit 79859186391381a226df908ff6cd55d73d6fc82d  ().
Date: 2025-03-16
Tag: v0.47.0, commits: 
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.2
 hyprcursor 0.1.11
 hyprgraphics 0.1.2
no flags were set
System Information:
System name: Linux
Node name: gza
Release: 6.12.18
Version: #1-NixOS SMP PREEMPT_DYNAMIC Fri Mar  7 17:25:47 UTC 2025
GPU information: 
2f:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6600/6600 XT/6600M] [1002:73ff] (rev c1) (prog-if 00 [VGA controller])
os-release: ANSI_COLOR="0;38;2;126;186;228"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="25.05.20250313.6607cf7"
CPE_NAME="cpe:/o:nixos:nixos:25.05"
DEFAULT_HOSTNAME=nixos
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
ID_LIKE=""
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 25.05 (Warbler)"
SUPPORT_URL="https://nixos.org/community.html"
VARIANT=""
VARIANT_ID=""
VENDOR_NAME=NixOS
VENDOR_URL="https://nixos.org/"
VERSION="25.05 (Warbler)"
VERSION_CODENAME=warbler
VERSION_ID="25.05"
```

</details>

### Description

The following commit: https://github.com/hyprwm/Hyprland/commit/6ffde3646688cee448683f95a858fc1754c0c35a
reopened the same issue I had previously: https://github.com/hyprwm/Hyprland/issues/8983

And the solution is exactly the same, remove the following comments removes the stuttering:

[...]

Issue #6951, opened 2024-07-21
DBeaver flickring on hyprland


kareemAbd0:
### Regression?

No

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 918d8340afd652b011b937d29d5eea0be08467f5  (flake.lock: update).
Date: Tue Jun 25 12:06:02 2024
Tag: v0.41.2, commits: 4886

flags: (if any)


System Information:
System name: Linux
Node name: KareemsPC
Release: 6.9.9-1-MANJARO
Version: #1 SMP PREEMPT_DYNAMIC Fri Jul 12 10:32:52 UTC 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation UHD Graphics 620 [8086:5917] (rev 07) (prog-if 00 [VGA controller])


os-release: NAME="Manjaro Linux"
PRETTY_NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://forum.manjaro.org/"
BUG_REPORT_URL="https://docs.manjaro.org/reporting-bugs/"
PRIVACY_POLICY_URL="https://manjaro.org/privacy-policy/"
LOGO=manjarolinux


plugins:

======Config-Start======


Config File: /home/kareem/.config/hypr/hyprland.conf: Read Succeeded
########################################################################################
#AUTOGENERATED HYPR CONFIG.
#PLEASE USE THE CONFIG PROVIDED IN THE GIT REPO /examples/hypr.conf AND EDIT IT,
#OR EDIT THIS ONE ACCORDING TO THE WIKI INSTRUCTIONS.
########################################################################################

#
# Please note not all available settings / options are set here.
# For a full list, see the wiki
#

#autogenerated = 1 # remove this line to remove the warning

# See https://wiki.hyprland.org/Configuring/Monitors/
monitor=HDMI-A-1,1920x1080@60,0x0,1
monitor=eDP-1,1920x1080@60,1920x0,1


workspace =1,monitor:HDMI-A-1
workspace =2, monitor:eDP-1
workspace =3,monitor:HDMI-A-1
workspace =5,monitor:HDMI-A-1
workspace =6,monitor:HDMI-A-1
[...]

PR #9675, opened 2025-03-19
syncobj: restore SHM buffer reset

gulafaran:
reset shm buffers early to mitigate stuttering animations, also reuse the monitors eglSync and store the eglsync per monitor. this however reintroduces flickering in dbeaver nonsyncobj application.

reintroduces #6951 and potentially #6912.

until buffer releases is properly figured out, reset shm buffers early. and store eglsync per monitor. and add it as releasesync.
might require some testing so this doesnt break something else that wasbeing tested when eventfd went through its rigorous testing on VRR/DS. 

see https://github.com/hyprwm/Hyprland/issues/9636

### DIFF
--- a/src/helpers/Monitor.hpp
+++ b/src/helpers/Monitor.hpp
@@ -57,6 +57,7 @@ struct SMonitorRule {
 
 class CMonitor;
 class CSyncTimeline;
+class CEGLSync;
 
 class CMonitorState {
   public:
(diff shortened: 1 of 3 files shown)


Skidamek:
> might require some testing so this doesnt break something else that wasbeing tested when eventfd went through its rigorous testing on VRR/DS.

vrr with ds seem to work fine

PR #9689, opened 2025-03-21
syncobj: store CDRMSyncobjTimelineResource as SP

gulafaran:
store the CDRMSyncobjTimelineResource as a shared pointer to ensure the syncobj fd doesnt destruct before we actually are done with it.

### DIFF
--- a/src/protocols/DRMSyncobj.cpp
+++ b/src/protocols/DRMSyncobj.cpp
@@ -9,50 +9,27 @@
 #include <fcntl.h>
 using namespace Hyprutils::OS;
 
-CDRMSyncPointState::CDRMSyncPointState(WP<CDRMSyncobjTimelineResource> resource_, uint64_t point_) : m_resource(resource_), m_point(point_) {}
+CDRMSyncPointState::CDRMSyncPointState(SP<CSyncTimeline> timeline_, uint64_t point_) : m_timeline(timeline_), m_point(point_) {}
 
 const uint64_t& CDRMSyncPointState::point() {
     return m_point;
 }
 
-WP<CDRMSyncobjTimelineResource> CDRMSyncPointState::resource() {
-    return m_resource;
-}
-
 WP<CSyncTimeline> CDRMSyncPointState::timeline() {
-    if (expired()) {
-        Debug::log(ERR, "CDRMSyncPointState: getting a timeline on a expired point");
-        return {};
-    }
-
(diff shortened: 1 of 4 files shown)


src/protocols/DRMSyncobj.hpp
```diff
@@ -31,9 +31,8 @@ class CDRMSyncPointState {
     void                                             signal();
 
   private:
-    WP<CDRMSyncobjTimelineResource> m_resource         = {};
+    SP<CDRMSyncobjTimelineResource> m_resource         = {};
```
PlasmaPower, line 34: I think we should just store an `SP<CSyncTimeline>` here instead of storing the resource. The only times we use the resource are to get the timeline.

gulafaran: problem is CDRMSyncobjTimelineResource owns the syncobj file that the CSyncTimeline gets, and if CDRMSyncobjTimelineResource destructs it closes it before the CSyncTimeline is done with it.

PlasmaPower: Perhaps the timeline itself should own the fd? But yeah this'd be a larger refactor and this change is fine on its own too

gulafaran: ok how about now?

src/protocols/DRMSyncobj.cpp
```diff
@@ -126,12 +93,12 @@ CDRMSyncobjSurfaceResource::CDRMSyncobjSurfaceResource(UP<CWpLinuxDrmSyncobjSurf
...
         }
 
-        if (!pendingAcquire.expired()) {
+        if (!pendingAcquire.timeline().expired()) {
             surface->pending.buffer->acquire = makeUnique<CDRMSyncPointState>(std::move(pendingAcquire));
             pendingAcquire                   = {};
         }
 
-        if (!pendingRelease.expired()) {
+        if (!pendingRelease.timeline().expired()) {
```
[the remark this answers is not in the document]
gulafaran: changed

PlasmaPower:
This looks great to me

vaxerski:
> This looks great to me

just like me hehe im so handsome my grandma told me that

https://github.com/user-attachments/assets/6563e69e-4f6d-431e-a385-01e65d8c6687

vaxerski (approved):

PR #9696, opened 2025-03-22
screencast: wait for fence if explicit sync

gulafaran:
wait for fence if explicit sync is enabled, to reduce jankiness on screencasting.

before
https://github.com/user-attachments/assets/b6a5279c-9481-4b97-87d3-f89eb61d9ae0

after, shitty quality because file got huge, so dropped bitrate etc.

https://github.com/user-attachments/assets/12ae2e33-5b24-4d8f-a57d-b44ff6055c11

### DIFF
--- a/src/protocols/Screencopy.cpp
+++ b/src/protocols/Screencopy.cpp
@@ -221,6 +221,12 @@ bool CScreencopyFrame::copyDmabuf() {
     g_pHyprOpenGL->m_RenderData.blockScreenShader = true;
     g_pHyprRenderer->endRender();
 
+    auto explicitOptions = g_pHyprRenderer->getExplicitSyncSettings(pMonitor->output);
+    if (explicitOptions.explicitEnabled) {
+        auto sync = g_pHyprOpenGL->createEGLSync();
+        sync->wait();
+    }
+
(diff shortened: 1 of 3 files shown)


PlasmaPower:
Ideally we'd do this async instead of blocking on the sync, e.g. with CSyncTimeline::addWaiter

PlasmaPower:
In fact, I think we might be able to use the existing CMonitor timeline and commitSeq for this, though I haven't looked at the screencopy codepath super closely so there might be some issues with that approach

gulafaran:
> In fact, I think we might be able to use the existing CMonitor timeline and commitSeq for this, though I haven't looked at the screencopy codepath super closely so there might be some issues with that approach

if i understand this, correctly. this isnt a blocking -wait() it merely makes the gl pipeline sync to the fence on the current context. on the gpu side.

PlasmaPower:
Good point. I'm not sure why this PR does anything right now, because no gl operations are performed after that pipeline sync, so it shouldn't matter. That said, apparently it does work, so idk might be some driver quirk.

Here's how I think it should screencopy should use explicit sync: https://github.com/hyprwm/Hyprland/pull/9697

gulafaran:
dropping in favour of #9697

PR #9706, opened 2025-03-23
Simplify and fix hdr metadata setting

UjinT34:
Fixes unwanted switch to SDR for fullscreen apps with CM support when `cm_fs_passthrough=2` (foot should no longer cause a modeset)
Fixes metadata creation. Primaries weren't sent correctly. Doesn't seem to have any effect on nvidia, other drivers might have some color improvements (need some info on how those values should be handled by the driver)
Sends primaries and luminances info for SDR content in case it'll be useful for fixing SDR passthrough on AQ side. Might make sense to set a hdr metadata blob with SDR eotf if those primaries and luminances are handled by the driver.
Simplifies hdr metadata switching logic

CM skip for rendering without a surface data should be removed. E verything that doesn't explicitly set an image description is considered sRGB and should go through the CM if monitor isn't in sRGB. If there are cases when `renderTextureInternalWithDamage` is called without a surface and the provided texture is encoded the same way then there should be a way to provide this info explicitly. Image description should be set on a `CTexture` level probably. Keeping it as is till #9600

Ready

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1382,8 +1382,9 @@ void CHyprOpenGLImpl::renderTextureInternalWithDamage(SP<CTexture> tex, const CB
     const auto imageDescription =
         m_RenderData.surface.valid() && m_RenderData.surface->colorManagement.valid() ? m_RenderData.surface->colorManagement->imageDescription() : SImageDescription{};
 
-    const bool skipCM = !*PENABLECM                                      /* CM disabled by the user */
-        || !m_bCMSupported                                               /* CM unsupported - hw failed to compile the shader probably */
+    const bool skipCM = !*PENABLECM /* CM disabled by the user */
+        || !m_RenderData.surface    /* FIXME unknown texture settings should be treated as sRGB and go through CM if monitor isn't in sRGB mode */
+        || !m_bCMSupported          /* CM unsupported - hw failed to compile the shader probably */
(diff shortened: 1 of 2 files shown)


vaxerski (approved):

aruhier:
@UjinT34, sorry I use this PR to ask this question, as it's related to `cm_fs_passthrough`. Can you add more details in the [wiki](https://wiki.hyprland.org/Configuring/Variables/#render) about this variable?

It's written that it's a boolean set by default to `true`, but if I understand correctly, it's more a int 0/1/2 option. For example, even with this patch, foot in fullscreen applies a wrong color mapping with `cm_fs_passthrough = 1` (or `true`), but is correct with `cm_fs_passthrough = 2`. What is the difference between 1 and 2?

UjinT34:
Looks like site is outdated https://github.com/hyprwm/hyprland-wiki/blob/main/pages/Configuring/Variables.md
1 is bugged for fullscreen sdr apps. If your monitor is in hdr mode you should use `cm_fs_passthrough = 0`. If it's in sdr mode `cm_fs_passthrough = 2` will auto-switch to hdr when there is a fullscreen hdr app.
Fullscreen hdr apps might look different with `cm_fs_passthrough = 0` and `cm_fs_passthrough = 1 or 2`. The former will apply the CM shader, the latter will send the content as is.

aruhier:
Perfect, thanks so much for the explanations!

Sorry I missed it, I didn't notice that the wiki set on git was outdated. I'm creating an issue for hyprland-wiki.

PR #9715, opened 2025-03-24
xwl: dont close the fd to early

gulafaran:
dont close the fd until the wl_event_source is removed, so we dont get another event triggered with an already closed fd.

seems to be the reason some people hit Xwayland not starting randomly. but a bit hard to pinpoint since it only happends rarely and only for some.

### DIFF
--- a/src/xwayland/Server.cpp
+++ b/src/xwayland/Server.cpp
@@ -177,8 +177,7 @@ static void startServer(void* data) {
 }
 
 static int xwaylandReady(int fd, uint32_t mask, void* data) {
-    CFileDescriptor xwlFd{fd};
-    return g_pXWayland->pServer->ready(std::move(xwlFd), mask);
+    return g_pXWayland->pServer->ready(fd, mask);
 }
 
 static bool safeRemove(const std::string& path) {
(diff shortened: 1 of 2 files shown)


src/xwayland/Server.cpp
```diff
@@ -300,7 +299,7 @@ void CXWaylandServer::runXWayland(CFileDescriptor& notifyFD) {
     }
 
     auto cmd =
-        std::format("Xwayland {} -rootless -core -listenfd {} -listenfd {} -displayfd {} -wm {}", displayName, xFDs[0].get(), xFDs[1].get(), notifyFD.get(), xwmFDs[1].get());
+        std::format("Xwayland {} -rootless -core -listenfd {} -listenfd {} -displayfd {} -wm {}", displayName, xFDs[0].get(), xFDs[1].get(), notifyFD.take(), xwmFDs[1].get());
```
PlasmaPower, line 302: Doesn't this leak notifyFD? Not that it's a huge deal as we only startup Xwayland once

src/xwayland/Server.cpp
```diff
@@ -421,6 +420,7 @@ int CXWaylandServer::ready(CFileDescriptor fd, uint32_t mask) {
 
     Debug::log(LOG, "XWayland is ready");
 
+    close(fd);
```
PlasmaPower, line 423: Doesn't this close the fd earlier, not later? I think in the previous code fd would be destructed at the end of the function and closed then.

src/xwayland/Server.cpp
PlasmaPower: I guess this is fine actually as this fork only executes Xwayland. Though I think this change also shouldn't matter as notifyFD's destructor should never be called in this function unless we fail to exec into Xwayland

src/xwayland/Server.cpp
PlasmaPower: Ah I see the difference, if this function returns early due to not reading enough bytes it'd previously close the file descriptor. This fix makes sense then.

gulafaran: Yep 🙂

src/xwayland/Server.cpp
gulafaran: Couldnt find who should close it so just reasoned as such, its only opened once, even if it leaks it doesnt matter much

PR #9722, opened 2025-03-25
monitor: prevent a crash when exiting where monitor eglsync is destroyed after renderer

ikalco:
self explanatory

yes


ikalco:
#9743 fixes this

PR #9726, opened 2025-03-25
config: completely remove windowrulev2

Zebra2711:
remove the config option `windowrulev2` which doesn't need anymore in version 0.48.0

I think we should replace all `windowrulev2` with `windowrule` in the config file instead of keeping both options that do the same thing

Ready


vaxerski:
the reason is that we deprecate it first, giving people time to switch

Zebra2711:
> the reason is that we deprecate it first, giving people time to switch

Ok

PR #9761, opened 2025-03-27
Monitor v2 config

UjinT34:
Adds a `monitorv2` config special category similar to `device`.

CM needs more monitor settings. Placing them into the current `monitor` config makes it too complicated.
Minimum `monitorv2` config should look like this:
```
monitorv2 {
  output = DP-1
  mode = 1920x1080@144
  position = 0x0
  scale = 1
}
```

Other settings have the same names as in `monitor` config.

`handleMonitorv2` is  called for every `monitorv2` at once. `addSpecialCategory` or `SSpecialCategoryOptions` need an additional handler parameter similar to `registerHandler` to handle each `monitorv2` individually.

Ready

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -740,6 +740,20 @@ CConfigManager::CConfigManager() {
     m_pConfig->addSpecialConfigValue("device", "flip_x", Hyprlang::INT{0});                   // only for touchpads
     m_pConfig->addSpecialConfigValue("device", "flip_y", Hyprlang::INT{0});                   // only for touchpads
 
+    m_pConfig->addSpecialCategory("monitorv2", {.key = "output"});
+    m_pConfig->addSpecialConfigValue("monitorv2", "disabled", Hyprlang::INT{0});
+    m_pConfig->addSpecialConfigValue("monitorv2", "mode", {"preferred"});
+    m_pConfig->addSpecialConfigValue("monitorv2", "position", {"auto"});
+    m_pConfig->addSpecialConfigValue("monitorv2", "scale", Hyprlang::FLOAT{1.0});
+    m_pConfig->addSpecialConfigValue("monitorv2", "addreserved", {STRVAL_EMPTY});
+    m_pConfig->addSpecialConfigValue("monitorv2", "mirror", {STRVAL_EMPTY});
+    m_pConfig->addSpecialConfigValue("monitorv2", "bitdepth", Hyprlang::INT{8});
+    m_pConfig->addSpecialConfigValue("monitorv2", "cm", {"auto"});
+    m_pConfig->addSpecialConfigValue("monitorv2", "sdrbrightness", Hyprlang::FLOAT{1.0});
+    m_pConfig->addSpecialConfigValue("monitorv2", "sdrsaturation", Hyprlang::FLOAT{1.0});
+    m_pConfig->addSpecialConfigValue("monitorv2", "vrr", Hyprlang::INT{0});
+    m_pConfig->addSpecialConfigValue("monitorv2", "transform", Hyprlang::INT{0});
+
     // keywords
     m_pConfig->registerHandler(&::handleExec, "exec", {false});
     m_pConfig->registerHandler(&::handleRawExec, "execr", {false});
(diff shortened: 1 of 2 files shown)


vaxerski:
what about dynamic calls?

UjinT34:
Do they work with `device`? Do we need those with `monitorv2`? Some config includes can be used to achieve same results. Existing `monitor` dynamic calls should work the same.

vaxerski:
yes they do work

UjinT34:
Looks like `hyprctl keyword monitorv2[desc:something]...` isn't supported by parser. `hyprctl keyword monitorv2[output]...` should work fine.
I don't think there is a point in adding anything else with this PR. Need some hyprlang and special categories changes first.

vaxerski:
`hyprctl keyword monitorv2[DP-2]:blahblah 2` works though, right? `desc:` might break due to the colon, that might be something to-fix in hyprlang.

src/debug/HyprCtl.cpp
```diff
@@ -1087,6 +1087,19 @@ static std::string dispatchKeyword(eHyprCtlOutputFormat format, std::string in)
     if (COMMAND == "monitor" || COMMAND == "source")
         g_pConfigManager->m_bWantsMonitorReload = true; // for monitor keywords
 
+    if (COMMAND.contains("monitorv2")) {
+        if (COMMAND.contains('[') && COMMAND.contains(']')) {
```
vaxerski, line 1091: why are we handling this? This is handled by hyprlang, no?!

UjinT34: Special categories are parsed without any handlers attached. Need a way to trigger a handler after the parsing for a given output.

UjinT34:
> `hyprctl keyword monitorv2[DP-2]:blahblah 2` works though, right? `desc:` might break due to the colon, that might be something to-fix in hyprlang.

`DP-2` works. Probably should leave ``hyprctl keyword monitorv2` as unsupported/experimental for now.

src/debug/HyprCtl.cpp
vaxerski: I think it's just because this MR has a weird approach. Why not use the approach like for device configs? After a dyncall, or after parsing, recheck monitorv2 stuff. If anything changed, update and reload monitor mode

src/config/ConfigManager.cpp
```diff
@@ -1952,210 +2029,261 @@ static bool parseModeLine(const std::string& modeline, drmModeModeInfo& mode) {
...
+}
+
+std::optional<std::string> CMonitorRuleParser::getError() {
+    if (m_error.empty())
+        return {};
+    return m_error;
+}
+
+bool CMonitorRuleParser::parseMode(const std::string& value) {
+    if (value.starts_with("pref")) {
```
vaxerski: no {} here

src/debug/HyprCtl.cpp
```diff
@@ -1109,6 +1109,23 @@ static std::string dispatchKeyword(eHyprCtlOutputFormat format, std::string in)
...
+        // if (COMMAND.contains('[') && COMMAND.contains(']')) {
+        //     const auto L = COMMAND.find_first_of('[');
+        //     const auto R = COMMAND.find_last_of(']');
+
+        //     if (L < R) {
+        //         const auto output = COMMAND.substr(L + 1, R - L - 1);
+        //         g_pConfigManager->handleMonitorv2(output);
+        //         g_pConfigManager->m_wantsMonitorReload = true;
+        //     }
+        // }
```
vaxerski: remove the commented out code

src/debug/HyprCtl.cpp
```diff
@@ -1109,6 +1109,23 @@ static std::string dispatchKeyword(eHyprCtlOutputFormat format, std::string in)
     if (COMMAND == "monitor" || COMMAND == "source")
         g_pConfigManager->m_wantsMonitorReload = true; // for monitor keywords
 
+    if (COMMAND.contains("monitorv2")) {
+        // Updates configs for every monitorv2 rule
+        g_pConfigManager->handleMonitorv2();
+        g_pConfigManager->m_wantsMonitorReload = true;
```
vaxerski: don't do this. Instead, do `handleMonitorV2` on reload in the reload func. Here, just do:
```cpp
g_pEventLoopManager->doLater([this] {
g_pConfigManager->m_wantsMonitorReload = true;
});
```

vaxerski (commented):
in general lgtm, wiki mr needed ofc

alba4k:
I like this, but wouldn't it make sense to have resolution and refresh rate as two different entries? I know they're together even in the current syntax, but is there any reason why this wouldn't be doable?

vaxerski:
because we can put a modeline there? I am not giga opposed but I don't think it makes sense. Up to you @UjinT34, I'd leave it as is

alba4k:
I see, fair point.

I did that when hyprland stopped supporting 1920x1080@72 on my 60 Hz monitors before aquamarine.

However couldn't a new `modeline = ` be an optional syntax that replaces both resolution and refresh-rate? would be a bit cleaner as a modeline is neither one or the other

UjinT34:
Maybe later. No need to change this atm. This one is meant to be a direct replacement of a `monitor` rule without any new stuff.

vaxerski (approved):
wiki mr needed

UjinT34:
https://github.com/hyprwm/hyprland-wiki/pull/1110

vaxerski:
there is no MR tho?

UjinT34:
Looks like I forgot to actually create a pull request %) https://github.com/hyprwm/hyprland-wiki/pull/1110

fxzzi:
I think this PR might've just nuked the relatively new `auto-center-{left,right,etc}` options

image

vaxerski:
:( it did, I can see it now. @UjinT34 care to fix that?

PR #9772, opened 2025-03-28
Fix the non-working tearing #9429

### LINKED ISSUE #9429: Allow tearing will not enable since the update to 0.47.2
### Already reported ? *

### Regression?

Probably not a regression / I don't remember it happening before

Hyprland 0.47.2 built from branch  at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: Sun Feb 2 00:47:17 2025
Tag: v0.47.2, commits: 5767
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.2

no flags were set

System Information:
System name: Linux
Node name: zachs-pc
Release: 6.13.2-2-cachyos
Version: #1 SMP PREEMPT_DYNAMIC Sat, 08 Feb 2025 11:26:09 +0000

GPU information:
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M/6850M XT] [1002:73df] (rev c1) (prog-if 00 [VGA controller])

os-release: NAME="CachyOS Linux"
PRETTY_NAME="CachyOS"
ID=cachyos
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://cachyos.org/"
DOCUMENTATION_URL="https://wiki.cachyos.org/"
SUPPORT_URL="https://discuss.cachyos.org/"
BUG_REPORT_URL="https://github.com/cachyos"
LOGO=cachyos

plugins:

</details>

### Description

I've tried numerous configurations to no avail allow tearing will not work anymore. I used it just fine on 0.46, but since I moved to 0.47.2 it will no longer enable anymore. Monitoring hyprctl monitors activelyTearing stays false & no tearing is seen on my preferred test game where it would be obvious.

### How to reproduce

1. set allow_tearing to true in general section of config
2. set immediate windowrulev2 on window class of game
3. while game is fullscreen go to another monitor to run command 'watch hyprctl monitors' to view activelyTearing
4. See if you notice any tearing in the game

### Attach not paste

- [x] Hyprland config - `hyprctl systeminfo -c` (always include)
- [x] Logs (might contain useful info such as errors)

[hyprland_config_dump.txt](https://github.com/user-attachments/files/18839433/hyprland_config_dump.txt)

iddm:
Fixes the non-working tearing by skipping a few checks for the solitary windows and making exceptions for the X11 windows when checking for the solitary window for a monitor.

The corresponding changes touch the nature of the X11 scaling in Wayland (Hyprland in particular), where there is a difference between the reported window size and real size, all depending on the scaling and whether the `xwayland:force_zero_scaling` is set to `true` or `false`. By avoiding the checks for the X11 windows, which require calculations based on the real window size, or comparing the two (the real and the reported) sizes, the solitary window check ends up finding a solitary window on the screen, and so, the tearing works again.

Fixes #9429

Ready for merge and review.

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1148,9 +1148,6 @@ bool CWindow::opaque() {
     if (m_fAlpha->value() != 1.f || m_fActiveInactiveAlpha->value() != 1.f)
         return false;
 
-    if (m_vRealSize->goal().floor() != m_vReportedSize)
-        return false;
-
     const auto PWORKSPACE = m_pWorkspace;
 
     if (m_pWLSurface->small() && !m_pWLSurface->m_bFillIgnoreSmall)


iddm:
CC @PaideiaDilemma

iddm:
I guess this fails https://github.com/iddm/Hyprland/actions/runs/14136858572/job/39610560281#logs due to my commit being done in my fork, rather than this repository.

src/desktop/Window.cpp
```diff
@@ -1148,7 +1148,7 @@ bool CWindow::opaque() {
     if (m_fAlpha->value() != 1.f || m_fActiveInactiveAlpha->value() != 1.f)
         return false;
 
-    if (m_vRealSize->goal().floor() != m_vReportedSize)
+    if (!m_bIsX11 && m_vRealSize->goal().floor() != m_vReportedSize)
```
vaxerski, line 1151: dafoq is this check even for come to think of it

src/render/Renderer.cpp
```diff
@@ -2099,8 +2099,9 @@ void CHyprRenderer::recheckSolitaryForMonitor(PHLMONITOR pMonitor) {
     if (!PCANDIDATE->opaque())
         return;
 
-    if (PCANDIDATE->m_vRealSize->value() != pMonitor->vecSize || PCANDIDATE->m_vRealPosition->value() != pMonitor->vecPosition || PCANDIDATE->m_vRealPosition->isBeingAnimated() ||
-        PCANDIDATE->m_vRealSize->isBeingAnimated())
+    if (!PCANDIDATE->m_bIsX11 &&
```
vaxerski, line 2102: this check is correct and should work on x11 too

iddm: It doesn't account for the disabled scaling for the X11 windows this way, if the scaling for the monitor is different than `1`.

src/desktop/Window.cpp
iddm: This check fails when we want to tear an X11 window on a scaled monitor with `xwayland:force_zero_scaling` set to `true`. Making here an exception for the X11 windows makes it work here. We can also potentially add here the check of the  `xwayland:force_zero_scaling`  flag to only avoid it here in case the zero scaling is specified, cause this size check will always fail anyway otherwise.

src/render/Renderer.cpp
vaxerski: real* are logical.

iddm: You are right, should work on x11 too, I'll remove this.

src/desktop/Window.cpp
vaxerski: > If you ask me what this check does here in general, I have no idea. 

yes that was the question. I'd remove it altogether

iddm: Thank you for confirming! Removing it also fixes the issue, just as intended. So I'll rewrite the commit to remove this line and this should be the fix then.

vaxerski (approved):
clean and small patch :)

PR #9775, opened 2025-03-28
Hyprland toplevel mapping protocol

### LINKED ISSUE #9381: Unable to get window for toplevel handle
I'm currently creating a custom share picker with window thumbnails. In the `XDPH_WINDOW_SHARING_LIST` provided by the desktop portal to the share picker only the toplevel id, window class and window title are provided. However, the window class and window title don't uniquely identify a window. I need the window address as handle for the frame capturing using the `hyprland-toplevel-export-v1` protocol.

As far as I can tell it's currently not possible to get the window address for a toplevel by it's id.

The easiest way to solve this problem would probably be to add the window address as an additional field to the windows in the `XDPH_WINDOW_SHARING_LIST`. However, this is only solving the problem for this specific use case. 

To solve the problem in the general case I would probably change the  `app_id` field in the ExtForeignToplevelHandleV1 and ZwlrForeignToplevelHandleV1 to the stringified window address rather than the window class. I think it would make sense for those objects to contain the window address since using the address one can get the class and all other properties of a window using the socket.

https://github.com/hyprwm/Hyprland/blob/f2d43e5f2180adb81b3b7b51a7d0f39dace70b05/src/protocols/ForeignToplevel.cpp#L61

https://github.com/hyprwm/Hyprland/blob/f2d43e5f2180adb81b3b7b51a7d0f39dace70b05/src/protocols/ForeignToplevelWlr.cpp#L209

Another option would be to add the toplevel id to the windows in the window list returned by `hyprctl clients`

WhySoBad:

This pull requests adds the server implementation of the hyprland-toplevel-mapping-v1 protocol introduced in https://github.com/hyprwm/hyprland-protocols/pull/16

**Related**:
- desktop portal implementation: https://github.com/hyprwm/xdg-desktop-portal-hyprland/pull/322
- protocol: https://github.com/hyprwm/hyprland-protocols/pull/16

close #9381

Ready for merge

### DIFF
--- a/src/protocols/ForeignToplevel.cpp
+++ b/src/protocols/ForeignToplevel.cpp
@@ -171,10 +171,19 @@ bool CForeignToplevelProtocol::windowValidForForeign(PHLWINDOW pWindow) {
 
 PHLWINDOW CForeignToplevelProtocol::windowFromHandleResource(wl_resource* res) {
     for (auto const& h : m_vHandles) {
-        if (h->res() != res)
+        if (h->resource->resource() != res)
             continue;
 
         return h->window();
+    }
+
+    return nullptr;
+}
+
+SP<CForeignToplevelHandle> CForeignToplevelProtocol::handleFromWindow(uint64_t addr) {
+    for (auto const& h : m_vHandles) {
+        if ((uint64_t)h->window().get() == addr)
+            return h;
     }
 
     return nullptr;
--- a/src/protocols/ForeignToplevel.cpp
+++ b/src/protocols/ForeignToplevel.cpp
@@ -179,12 +179,3 @@ PHLWINDOW CForeignToplevelProtocol::windowFromHandleResource(wl_resource* res) {
 
     return nullptr;
 }
-
-SP<CForeignToplevelHandle> CForeignToplevelProtocol::handleFromWindow(uint64_t addr) {
-    for (auto const& h : m_vHandles) {
-        if ((uint64_t)h->window().get() == addr)
-            return h;
-    }
-
-    return nullptr;
-}
--- a/src/protocols/ForeignToplevel.cpp
+++ b/src/protocols/ForeignToplevel.cpp
@@ -168,3 +168,14 @@ void CForeignToplevelProtocol::destroyHandle(CForeignToplevelHandle* handle) {
 bool CForeignToplevelProtocol::windowValidForForeign(PHLWINDOW pWindow) {
     return validMapped(pWindow) && !pWindow->isX11OverrideRedirect();
 }
+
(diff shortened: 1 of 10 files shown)


WhySoBad:
Hi,

Sorry for adding all labels, I accidentally rebased and pushed which triggered the bot.

Anyways, I don't know how to fix the meson setup and Nix CI errors. I'd appreciate if someone could help me in this matter :)

fufexan:
Can you rebase on main? Should fix the Nix CI. Don't know what's up with the Arch Meson CI.

WhySoBad:
I think the problem with the Nix build is that the hyprland-protocols revision is locked? However, I don't know anything about Nix so this is only a guess

src/protocols/ForeignToplevel.cpp
```diff
@@ -168,3 +168,14 @@ void CForeignToplevelProtocol::destroyHandle(CForeignToplevelHandle* handle) {
 bool CForeignToplevelProtocol::windowValidForForeign(PHLWINDOW pWindow) {
     return validMapped(pWindow) && !pWindow->isX11OverrideRedirect();
 }
+
+PHLWINDOW CForeignToplevelProtocol::windowFromHandleResource(wl_resource* res) {
```
vaxerski, line 172: this sucks, as it's slow and unintuitive. See e.g. `CWLSurface::fromResource` for how to do it right.

src/protocols/ToplevelMapping.hpp
```diff
@@ -0,0 +1,35 @@
...
+    CToplevelMappingProtocol(const wl_interface* iface, const int& ver, const std::string& name);
+
+    virtual void bindManager(wl_client* client, void* data, uint32_t ver, uint32_t id);
+
+  private:
+    void                                                    onManagerResourceDestroy(CToplevelMappingManager* mgr);
+    void                                                    destroyHandle(CHyprlandToplevelWindowMappingHandleV1* handle);
+
+    std::vector<UP<CToplevelMappingManager>>                m_vManagers;
+    std::vector<SP<CHyprlandToplevelWindowMappingHandleV1>> m_vHandles;
```
vaxerski, line 28: don't rawdog resource types, please. Wrap it like you did for the manager

WhySoBad: done

src/protocols/ForeignToplevel.cpp
WhySoBad: alright, changed it. I also changed it for the wlr toplevel which already had this method

WhySoBad:
Hi,

I don't want to pressure but it's been some time since your review. Is anything preventing this pull request from being merged?

vaxerski:
oops, got lost in my notifications

vaxerski (approved):

PR #9776, opened 2025-03-28
protocols: Don't update hdr metadata if image description is unchanged

PlasmaPower:
Gamescope sends a new frog color management protocol `set_render_intent`, `set_known_container_color_volume`, and `set_known_transfer_function` every commit. Previously, this would trigger a modeset every commit, despite the new image description being equal to the old image description.

I'm not familiar with color management so it might be good to get @UjinT34's review, but this PR does fix the issue for the user who [reported it on Discord](https://discord.com/channels/961691461554950145/1070436481912549497/1355298799756771389).

Ready for merging

### DIFF
--- a/src/protocols/ColorManagement.cpp
+++ b/src/protocols/ColorManagement.cpp
@@ -331,11 +331,16 @@ const hdr_output_metadata& CColorManagementSurface::hdrMetadata() {
 }
 
 void CColorManagementSurface::setHDRMetadata(const hdr_output_metadata& metadata) {
-    m_hdrMetadata      = metadata;
-    m_needsNewMetadata = false;
+    m_hdrMetadata          = metadata;
+    m_lastImageDescription = m_imageDescription;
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
looks reasonable to me, but I'll wait for ujin's input

UjinT34:
lgtm

Won't help if client keeps spamming different image descriptions. Might happen to gamescope with mangohud. Looks like it alternates between the game and mangohud layer but always ends up with the same description. If something throws it off then it may result in non-working hdr or constant sdr - hdr switching.

PlasmaPower:
I think the different settings are on different subsurfaces so it's fine. It at least fixed the issue for the user who reported it and whose logs we've been looking at. There might be a second issue of how to deal with multiple subsurfaces but this at least fixes the performance part.

vaxerski (approved):

PR #9785, opened 2025-03-29
Auto hdr

UjinT34:
Adds `render:cm_auto_hdr` for hdr automation. Should be used with `cm_fs_passthrough = 0`, won't do anything with other values.
0 - off
1 - sets `cm, hdr`
2 - sets `cm, hdredid`

Can be used to fix the same issue as in https://github.com/hyprwm/Hyprland/pull/9776 in a more reliable way (both PRs are needed to cover all possible configurations)

Needs more testing with different `cm`, `cm_fs_passthrough` and `cm_auto_hdr` values. Might not work as expected with some combo. I've tested most sensible ones.

Ready

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1398,9 +1398,9 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     },
     SConfigOptionDescription{
         .value       = "render:cm_auto_hdr",
-        .description = "Auto-switch to hdr mode when fullscreen app is in hdr (cm_fs_passthrough can switch to hdr even when this setting is off)",
-        .type        = CONFIG_OPTION_BOOL,
-        .data        = SConfigOptionDescription::SBoolData{true},
+        .description = "Auto-switch to hdr mode when fullscreen app is in hdr, 0 - off, 1 - hdr, 2 - hdredid (cm_fs_passthrough can switch to hdr even when this setting is off)",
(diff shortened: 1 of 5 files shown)


LionHeartP:
I tested this PR just now with these settings.
`monitor=DP-1,2560x1440@144,auto,1, vrr, 0, bitdepth, 10`
`render:cm_auto_hdr=1` and `2`
`render:cm_fs_passthrough = 0`

While it solved the issue that #9776 also solves, it made the colors look weird in gamescope. I could immediately tell by looking at Mangohud that red looked kinda orange-ish. 

I tested the other game I'm currently playing on Steam (RE4R) with wine-wayland, and the colors looked exactly like before, inluding Mangohud.

UjinT34:
It's expected to look different.
With passthrough mangohud image description is ignored and it becomes oversaturated. Without passthrough it would be correct... but frog-cm assumes that everything is hdr and gamescope has some builtin cm. This results in double or triple conversion. With xx-cm or wp-cm the conversion should be correct.
Game colors might be different with and without passthrough. Probably less saturated without passthrough. It might be the intended look but I don't like it. Hard to tell what is wrong and why.
`render:cm_auto_hdr=2` might be better or worse depending on EDID correctness.

LionHeartP:
Well there's an open PR for Gamescope to switch over to upstream CM protocol, so I'm expecting this to be a non issue sooner rather than later.

vaxerski (dismissed):

PR #9793, opened 2025-03-29
layerrules: add abovelock to render above lockscreen

VirtCode:
This PR adds a new layerrule `abovelock`. If this rule is set on a layer, it will be drawn above the lockscreen if the session is locked. Layers which are drawn above the lockscreen will _not_ be drawn twice (i.e. only above the lock and not also below it) if a lockscreen is active, so semi-transparent layers won't look shit when the lockscreen is just partially opaque.

The layerrule takes a _optional_ boolean as an argument to specify whether the surface should be interactable on the lockscreen. Default is false, meaning the user would not be able to interact with it. Currently, only pointer focus is implemented, so the **keyboard focus will stay on the lockscreen**, as that is probably the expected behaviour anyways.

Configuration example:
```
layerrule = abovelock true, notifications
```

I mainly want this for my own setup but I can imagine plenty of use cases where this would be very cool:
- Be able to see incoming notifications on the lockscreen
- One might want to have their bar above the lockscreen so you can still see your system info (e.g. battery level) on the lockscreen
- An OSD would also be nice if it would be shown on the lockscreen
- You'd be able to use OSKs for touch devices to be able to enter a password
- etc.

- I am passing a new bool `aboveLockscreen` to `vectorToLayerSurface` such that I can filter the surfaces when checking them. Just checking the result for abovelock would lead to edge-cases e.g. when there are multiple layers above each other, not all being abovelock. Lmk if you have a better idea of handling this.

Yes

### DIFF
--- a/src/desktop/LayerRule.cpp
+++ b/src/desktop/LayerRule.cpp
@@ -4,8 +4,8 @@
 #include <algorithm>
 #include "../debug/Log.hpp"
 
-static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur", "blurpopups", "dimaround", "abovelock"};
-static const auto RULES_PREFIX = std::unordered_set<std::string>{"ignorealpha", "ignorezero", "xray", "animation", "order"};
+static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur", "blurpopups", "dimaround"};
+static const auto RULES_PREFIX = std::unordered_set<std::string>{"ignorealpha", "ignorezero", "xray", "animation", "order", "abovelock"};
 
 CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_targetNamespace(ns_), m_rule(rule_) {
     const bool VALID = RULES.contains(m_rule) || std::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
@@ -33,6 +33,8 @@ CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_tar
         m_ruleType = RULE_ANIMATION;
     else if (m_rule.starts_with("order"))
         m_ruleType = RULE_ORDER;
+    else if (m_rule.starts_with("abovelock"))
+        m_ruleType = RULE_ABOVELOCK;
     else {
         Debug::log(ERR, "CLayerRule: didn't match a rule that was found valid?!");
         m_ruleType = RULE_INVALID;
--- a/src/desktop/LayerRule.cpp
+++ b/src/desktop/LayerRule.cpp
@@ -4,7 +4,7 @@
 #include <algorithm>
 #include "../debug/Log.hpp"
 
-static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur", "blurpopups", "dimaround"};
+static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur", "blurpopups", "dimaround", "abovelock"};
 static const auto RULES_PREFIX = std::unordered_set<std::string>{"ignorealpha", "ignorezero", "xray", "animation", "order"};
 
 CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_targetNamespace(ns_), m_rule(rule_) {
@@ -21,6 +21,8 @@ CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_tar
         m_ruleType = RULE_BLURPOPUPS;
     else if (m_rule == "dimaround")
         m_ruleType = RULE_DIMAROUND;
+    else if (m_rule == "abovelock")
+        m_ruleType = RULE_ABOVELOCK;
     else if (m_rule.starts_with("ignorealpha"))
         m_ruleType = RULE_IGNOREALPHA;
     else if (m_rule.starts_with("ignorezero"))
--- a/src/desktop/LayerRule.cpp
+++ b/src/desktop/LayerRule.cpp
@@ -21,8 +21,6 @@ CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_tar
         m_ruleType = RULE_BLURPOPUPS;
     else if (m_rule == "dimaround")
         m_ruleType = RULE_DIMAROUND;
-    else if (m_rule == "abovelock")
-        m_ruleType = RULE_ABOVELOCK;
     else if (m_rule.starts_with("ignorealpha"))
         m_ruleType = RULE_IGNOREALPHA;
     else if (m_rule.starts_with("ignorezero"))
--- a/src/desktop/LayerSurface.cpp
+++ b/src/desktop/LayerSurface.cpp
@@ -440,6 +440,11 @@ void CLayerSurface::applyRules() {
             }
             case CLayerRule::RULE_ABOVELOCK: {
                 m_aboveLockscreen = true;
+
+                CVarList vars{rule->m_rule, 0, ' '};
+                try {
+                    m_aboveLockscreenInteractable = configStringToInt(vars[1]).value_or(false);
+                } catch (...) {}
                 break;
             }
             default: break;
--- a/src/desktop/LayerSurface.cpp
+++ b/src/desktop/LayerSurface.cpp
@@ -421,9 +421,8 @@ void CLayerSurface::applyRules() {
             }
             case CLayerRule::RULE_XRAY: {
                 CVarList vars{rule->m_rule, 0, ' '};
-                try {
-                    m_xray = configStringToInt(vars[1]).value_or(false);
-                } catch (...) {}
+                m_xray = configStringToInt(vars[1]).value_or(false);
+
                 break;
             }
             case CLayerRule::RULE_ANIMATION: {
@@ -442,9 +441,8 @@ void CLayerSurface::applyRules() {
                 m_aboveLockscreen = true;
 
                 CVarList vars{rule->m_rule, 0, ' '};
-                try {
-                    m_aboveLockscreenInteractable = configStringToInt(vars[1]).value_or(false);
-                } catch (...) {}
+                m_aboveLockscreenInteractable = configStringToInt(vars[1]).value_or(false);
+
                 break;
             }
             default: break;
--- a/src/desktop/LayerSurface.cpp
+++ b/src/desktop/LayerSurface.cpp
@@ -438,6 +438,10 @@ void CLayerSurface::applyRules() {
                 } catch (...) { Debug::log(ERR, "Invalid value passed to order"); }
                 break;
             }
+            case CLayerRule::RULE_ABOVELOCK: {
+                m_aboveLockscreen = true;
+                break;
+            }
             default: break;
         }
     }
--- a/src/desktop/LayerSurface.hpp
+++ b/src/desktop/LayerSurface.hpp
@@ -50,6 +50,7 @@ class CLayerSurface {
     float                      m_ignoreAlphaValue = 0.f;
     bool                       m_dimAround        = false;
     int64_t                    m_order            = 0;
+    bool                       m_aboveLockscreen  = false;
 
     std::optional<std::string> m_animationStyle;
 
--- a/src/desktop/LayerSurface.hpp
+++ b/src/desktop/LayerSurface.hpp
@@ -43,14 +43,14 @@ class CLayerSurface {
     bool                       m_noProcess     = false;
     bool                       m_noAnimations  = false;
 
-    bool                       m_forceBlur        = false;
-    bool                       m_forceBlurPopups  = false;
-    int64_t                    m_xray             = -1;
-    bool                       m_ignoreAlpha      = false;
-    float                      m_ignoreAlphaValue = 0.f;
-    bool                       m_dimAround        = false;
-    int64_t                    m_order            = 0;
-    bool                       m_aboveLockscreen  = false;
+    bool                       m_forceBlur                   = false;
+    bool                       m_forceBlurPopups             = false;
+    int64_t                    m_xray                        = -1;
+    bool                       m_ignoreAlpha                 = false;
+    float                      m_ignoreAlphaValue            = 0.f;
+    bool                       m_dimAround                   = false;
+    int64_t                    m_order                       = 0;
+    bool                       m_aboveLockscreen             = false;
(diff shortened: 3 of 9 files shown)


earboxer:
> you can't interact with it

> Be able to see incoming notifications on the lockscreen


Wouldn't you want to be able to interact with your notifications?

I think it would be cool to have this for Virtual Keyboards... so I can type my password into my lockscreen on my phone

PaideiaDilemma:
Kde does that as well. They have a protocol to allow/disallow that for their internal surfaces. https://wayland.app/protocols/kde-lockscreen-overlay-v1

I think it would be cool and useful for stuff like you described. But I do agree with earboxer that we probably want to have it be interactive as well. At least users gonna ask for it and someone probably will make a PR for it at some point. 

The implementation does conflict a bit with my plan to stop rendering workspaces below sessionlock clients. This [hyprlock PR](https://github.com/hyprwm/hyprlock/pull/726) preps for that on the client side, cause we still rely on it in hyprlock. It would be nice to have it decoupled from `renderAllClientsForWorkspace` somehow, to make the code path for session-lock more obvious and allow us to remove rendering of workspaces below session-lock

VirtCode:
> I think it would be cool to have this for Virtual Keyboards... so I can type my password into my lockscreen on my phone

Interactivity would indeed be very useful for that, didn't think of OSKs before. Thought it wasn't worth having to dig into that convoluted focus logic for my use-cases, but you are right, interactivity would sooner or later be asked for anyways. So I'm gonna have a look at it this evening probably.

> The implementation does conflict a bit with my plan to stop rendering workspaces below sessionlock clients. This https://github.com/hyprwm/hyprlock/pull/726 preps for that on the client side, cause we still rely on it in hyprlock. It would be nice to have it decoupled from renderAllClientsForWorkspace somehow, to make the code path for session-lock more obvious and allow us to remove rendering of workspaces below session-lock

Sure, if I understand correctly we'd have to refactor out the lockscreen rendering anyways, so we can also just render the abovelock surfaces again there. Depending on the implementation there we maybe would specifically have to handle the damage for these surfaces but that shouldn't be too difficult. I think this is best done in another mr, as currently the rendering changes are quite simple.

PaideiaDilemma:
> Sure, if I understand correctly we'd have to refactor out the lockscreen rendering anyways, so we can also just render the abovelock surfaces again there. Depending on the implementation there we maybe would specifically have to handle the damage for these surfaces but that shouldn't be too difficult. I think this is best done in another mr, as currently the rendering changes are quite simple.

No the rendering for session-lock is already self-contained. This PR introduces the requirement to do normal rendering when locked. Currently the change for not rendering workspaces below session-lock is < 10 LOC changes. With this I will have to rethink how this feature is implemented. We could already think about that now, or I will have to think about it later.
I am fine with that though if we decide to merge this as is.

VirtCode:
I have now implemented pointer focus for abovelock surfaces and seems to work as expected on my end. I did not yet look at keyboard focus because I think that always being on the lockscreen feels more intuitive and foolproof. But if anyone has a use-case where this would be needed, I might have a look at it.

> With this I will have to rethink how this feature is implemented. We could already think about that now, or I will have to think about it later.

I don't really understand what you mean here. As far as I understand is the current rendering split between `renderWorkspace` and `renderLockscreen`. This PR renders the layer surfaces inside the `renderLockscreen` method again, so nothing regarding that really changed? I mean we can just not call `renderWorkspace` and the abovelock surfaces will still be drawn. Or am I missing something here or what do you mean by "normal rendering"?

PaideiaDilemma:
> I don't really understand what you mean here. As far as I understand is the current rendering split between renderWorkspace and renderLockscreen. This PR renders the layer surfaces inside the renderLockscreen method again, so nothing regarding that really changed? I mean we can just not call renderWorkspace and the abovelock surfaces will still be drawn. Or am I missing something here or what do you mean by "normal rendering"?

Right, sorry I didn't see that you render the above lockscreen layers in renderLockscreen and thought that it would rely on rendering the rest of the workspace. My bad.

> I did not yet look at keyboard focus because I think that always being on the lockscreen feels more intuitive and foolproof. But if anyone has a use-case where this would be needed, I might have a look at it.

Yeah keeb focus staying on the lockscreen surfaces seems like a good idea!

src/desktop/LayerSurface.cpp
```diff
@@ -433,6 +433,15 @@ void CLayerSurface::applyRules() {
                 } catch (...) { Debug::log(ERR, "Invalid value passed to order"); }
                 break;
             }
+            case CLayerRule::RULE_ABOVELOCK: {
+                aboveLockscreen = true;
+
+                CVarList vars{rule->rule, 0, ' '};
+                try {
```
vaxerski, line 440: I don't think configStringToInt can throw?

src/render/Renderer.cpp
```diff
@@ -679,10 +679,14 @@ void CHyprRenderer::renderWindow(PHLWINDOW pWindow, PHLMONITOR pMonitor, timespe
     g_pHyprOpenGL->m_RenderData.currentWindow.reset();
 }
 
-void CHyprRenderer::renderLayer(PHLLS pLayer, PHLMONITOR pMonitor, timespec* time, bool popups) {
+void CHyprRenderer::renderLayer(PHLLS pLayer, PHLMONITOR pMonitor, timespec* time, bool popups, bool lockscreen) {
```
vaxerski, line 682: ew. Just filter them before you get here. Don't add another param

src/desktop/LayerSurface.hpp
```diff
@@ -44,13 +44,15 @@ class CLayerSurface {
...
-    int64_t                    order            = 0;
+    bool                       forceBlur                   = false;
+    bool                       forceBlurPopups             = false;
+    int64_t                    xray                        = -1;
+    bool                       ignoreAlpha                 = false;
+    float                      ignoreAlphaValue            = 0.f;
+    bool                       dimAround                   = false;
+    int64_t                    order                       = 0;
+    bool                       aboveLockscreen             = false;
+    int64_t                    aboveLockscreenInteractable = false;
```
vaxerski, line 55: int = false?

src/render/Renderer.cpp
VirtCode: The thing is that we don't want to render layers that are "abovelock" below the lockscreen (i.e. when they are normally rendered). This means we'd need to filter the layers every time we use `renderLayer`, which is at least four times for each layer-surface layer. So I thought it would be cleanest to move the check inside the method.

src/desktop/LayerSurface.cpp
VirtCode: right, copied that from the `xray` layer rule but it looks like it indeed doesn't throw. changed it there too

src/render/Renderer.cpp
vaxerski: why not do it in renderWorkspace? Like we do with windows (see renderallclients...)

VirtCode: I don't really know what you mean. The only other option I can currently think of would be adding a check before almost every call to render layer, which is pretty ugly. Like here's what I could do:

<details>
<summary>ugly patch</summary>

```diff
diff --git a/src/render/Renderer.cpp b/src/render/Renderer.cpp
index 2079df2b..5f76182a 100644
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -862,16 +862,20 @@ void CHyprRenderer::renderAllClientsForWorkspace(PHLMONITOR pMonitor, PHLWORKSPA
             g_pHyprOpenGL->clearWithTex(); // will apply the hypr "wallpaper"
 
         for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]) {
-            renderLayer(ls.lock(), pMonitor, time);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time);
         }
         for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]) {
-            renderLayer(ls.lock(), pMonitor, time);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time);
         }
         for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]) {
-            renderLayer(ls.lock(), pMonitor, time);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time);
         }
         for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]) {
-            renderLayer(ls.lock(), pMonitor, time);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time);
         }
 
         return;
@@ -884,10 +888,12 @@ void CHyprRenderer::renderAllClientsForWorkspace(PHLMONITOR pMonitor, PHLWORKSPA
             g_pHyprOpenGL->clearWithTex(); // will apply the hypr "wallpaper"
 
         for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]) {
-            renderLayer(ls.lock(), pMonitor, time);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time);
         }
         for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]) {
-            renderLayer(ls.lock(), pMonitor, time);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time);
         }
     }
 
@@ -956,7 +962,8 @@ void CHyprRenderer::renderAllClientsForWorkspace(PHLMONITOR pMonitor, PHLWORKSPA
 
     // Render surfaces above windows for monitor
     for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]) {
-        renderLayer(ls.lock(), pMonitor, time);
+        if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+            renderLayer(ls.lock(), pMonitor, time);
     }
 
     // Render IME popups
@@ -965,12 +972,14 @@ void CHyprRenderer::renderAllClientsForWorkspace(PHLMONITOR pMonitor, PHLWORKSPA
     }
 
     for (auto const& ls : pMonitor->m_aLayerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]) {
-        renderLayer(ls.lock(), pMonitor, time);
+        if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+            renderLayer(ls.lock(), pMonitor, time);
     }
 
     for (auto const& lsl : pMonitor->m_aLayerSurfaceLayers) {
         for (auto const& ls : lsl) {
-            renderLayer(ls.lock(), pMonitor, time, true);
+            if (ls && (!ls->aboveLockscreen || !g_pSessionLockManager->isSessionLocked()))
+                renderLayer(ls.lock(), pMonitor, time, true);
         }
     }
 
@@ -995,7 +1004,8 @@ void CHyprRenderer::renderLockscreen(PHLMONITOR pMonitor, timespec* now, const C
             // render layers for abovelock rule
             for (auto const& lsl : pMonitor->m_aLayerSurfaceLayers) {
                 for (auto const& ls : lsl) {
-                    renderLayer(ls.lock(), pMonitor, now, false, true);
+                    if (ls && ls->aboveLockscreen)
+                        renderLayer(ls.lock(), pMonitor, now, false, true);
                 }
             }
```

</details>

Other than that, I just noticed that we also have to render popups for layer surfaces. Should I also be rendering them for abovelock surfaces above the lockscreen?

vaxerski: hm, right. fair enough I guess.

> Other than that, I just noticed that we also have to render popups for layer surfaces. Should I also be rendering them for abovelock surfaces above the lockscreen?

of course

VirtCode: done, I'll mark this as resolved then

earboxer:
I tested using wvkbd and swaylock-effects on a non-touch device, and noticed
1. if my cursor was above the keyboard before the lockscreen started, the cursor is visible, but as soon as I move the mouse off of it and back on, the cursor is invisible
2. during the time in which the cursor is visible, having not moved it off of wvkbd, the characters typed don't get inputted into swaylock-effects (so the lockscreen doesn't yet have keyboard focus)

it looks like swaylock-effects sets exclusive keyboard interactivity, so I think this is a bug [with code in Hyprland]

haven't yet tested on a touchscreen device, but I suspect 2 will also apply there

src/desktop/LayerRule.cpp
```diff
@@ -21,6 +21,8 @@ CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_tar
         m_ruleType = RULE_BLURPOPUPS;
     else if (m_rule == "dimaround")
         m_ruleType = RULE_DIMAROUND;
+    else if (m_rule == "abovelock")
```
vaxerski, line 24: this is redundant, your starts with covers this.

vaxerski (commented):
rest lbgmt

src/desktop/LayerRule.cpp
VirtCode: done, somehow reintroduced that in the rebase

vaxerski (approved):

JuneStepp:
Unfortunately, this doesn't seem to work with touch input. The events just go straight through to the lock screen.

VirtCode:
right, didn't test that as I don't have a touch device

can you try this [touch-patch.txt](https://github.com/user-attachments/files/20008629/touch-patch.txt)

VirtCode:
Okay turns out my 6am brain made a pretty stupid mistake lol. Anyways, dug out an old laptop I had lying around and the touch input seems to be working there. So I opened a pr with the changed patch here #10253

Would be nice if you could test if this also works fine on your end - only did some pretty rudimentary testing.

PR #9804, opened 2025-03-30
types: properly use buffer refs

ikalco:
#### Describe your PR, what does it fix/add?
part 1 of #9774

#### Is it ready for merging, or does it need work?
yes

Issue #9640, opened 2025-03-16
Flickering when using NVIDIA GPU


BlueskyFR:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

I don't know, I started using Hyprland only recently

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.47.2 built from branch v0.47.2-b at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: 2025-02-02
Tag: v0.47.2, commits: 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.1
 hyprcursor 0.1.11
 hyprgraphics 0.1.2


no flags were set


System Information:
System name: Linux
Node name: nzxt
Release: 6.12.19
Version: #1-NixOS SMP PREEMPT_DYNAMIC Thu Mar 13 12:02:20 UTC 2025


GPU information:
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU102GL [Quadro RTX 6000/8000] [10de:1e30] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  570.124.04  Release Build  (nixbld@)


os-release: ANSI_COLOR="0;38;2;126;186;228"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="25.05pre-git"
CPE_NAME="cpe:/o:nixos:nixos:25.05"
DEFAULT_HOSTNAME=nixos
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
ID_LIKE=""
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 25.05 (Warbler)"
SUPPORT_URL="https://nixos.org/community.html"
VARIANT=""
VARIANT_ID=""
VENDOR_NAME=NixOS
VENDOR_URL="https://nixos.org/"
VERSION="25.05 (Warbler)"
VERSION_CODENAME=warbler
VERSION_ID="25.05"


plugins:
```

</details>

### Description

[...]

PR #9805, opened 2025-03-30
core: refactor and improve surface commit

ikalco:
part 2 of #9774

relevant commits:
[cleanup SSurfaceState and surface pending commit tracking](https://github.com/hyprwm/Hyprland/pull/9805/commits/dfcfb1d68306ceb73215c14536390a63070a553d)
[move surface code from DRMSyncobj, and move acquire to SSurfaceState](https://github.com/hyprwm/Hyprland/pull/9805/commits/9fbd34d0c75d4b3f76954300520164cb4999b9e0)
[use queue for comitted pending surface states like proto says](https://github.com/hyprwm/Hyprland/pull/9805/commits/3cb43e04b1b83fb81ebc2b78f1572a3c54f07c68)
[drop, not release, prev buffer if 2nd buffer wl_surface.attach is sent](https://github.com/hyprwm/Hyprland/pull/9805/commits/3002845e0e384f89fe872b9325142ee1d43a6230)

based on part 1

<s>should be gtg now</s>
<s>nvm lol nvidia borkde again (╥_╥)</s>
yes finally

### DIFF
--- a/src/protocols/types/SurfaceState.cpp
+++ b/src/protocols/types/SurfaceState.cpp
@@ -35,7 +35,7 @@ CRegion SSurfaceState::accumulateBufferDamage() {
 }
 
 void SSurfaceState::updateSynchronousTexture(SP<CTexture> lastTexture) {
-    auto [dataPtr, fmt, size] = buffer->buffer->beginDataPtr(0);
+    auto [dataPtr, fmt, size] = buffer->beginDataPtr(0);
     if (dataPtr) {
         auto drmFmt = NFormatUtils::shmToDRM(fmt);
         auto stride = bufferSize.y ? size / bufferSize.y : 0;
@@ -45,7 +45,7 @@ void SSurfaceState::updateSynchronousTexture(SP<CTexture> lastTexture) {
         } else
             texture = makeShared<CTexture>(drmFmt, dataPtr, stride, bufferSize);
     }
-    buffer->buffer->endDataPtr();
+    buffer->endDataPtr();
 }
 
 void SSurfaceState::reset() {
@@ -65,7 +65,7 @@ void SSurfaceState::updateFrom(SSurfaceState& ref) {
         *this = ref;
         ref.damage.clear();
         ref.bufferDamage.clear();
-        ref.buffer.reset();
+        ref.buffer = {};
     } else {
         if (ref.updated & SURFACE_UPDATED_DAMAGE) {
             damage       = ref.damage;
--- a/src/protocols/types/SurfaceState.cpp
+++ b/src/protocols/types/SurfaceState.cpp
@@ -49,45 +49,46 @@ void SSurfaceState::updateSynchronousTexture(SP<CTexture> lastTexture) {
 }
 
 void SSurfaceState::reset() {
+    updated.all = false;
+
+    // After commit, there is no pending buffer until the next attach.
+    buffer = {};
+
+    // wl_surface.commit assings pending ... and clears pending damage.
     damage.clear();
     bufferDamage.clear();
-    transform = WL_OUTPUT_TRANSFORM_NORMAL;
-    scale     = 1;
-    offset    = {};
-    size      = {};
 }
 
 void SSurfaceState::updateFrom(SSurfaceState& ref) {
     updated = ref.updated;
 
-    if (ref.updated & SURFACE_UPDATED_BUFFER) {
-        ref.updated &= ~SURFACE_UPDATED_BUFFER;
-        *this = ref;
-        ref.damage.clear();
-        ref.bufferDamage.clear();
-        ref.buffer = {};
-    } else {
-        if (ref.updated & SURFACE_UPDATED_DAMAGE) {
-            damage       = ref.damage;
-            bufferDamage = ref.bufferDamage;
-        }
-
-        if (ref.updated & SURFACE_UPDATED_INPUT)
-            input = ref.input;
-
-        if (ref.updated & SURFACE_UPDATED_OPAQUE)
-            opaque = ref.opaque;
-
-        if (ref.updated & SURFACE_UPDATED_OFFSET)
-            offset = ref.offset;
-
-        if (ref.updated & SURFACE_UPDATED_SCALE)
-            scale = ref.scale;
-
-        if (ref.updated & SURFACE_UPDATED_VIEWPORT)
-            viewport = ref.viewport;
-
-        if (ref.updated & SURFACE_UPDATED_TRANSFORM)
-            transform = ref.transform;
+    if (ref.updated.buffer) {
+        buffer     = ref.buffer;
+        texture    = ref.texture;
+        size       = ref.size;
+        bufferSize = ref.bufferSize;
     }
+
+    if (ref.updated.damage) {
+        damage       = ref.damage;
+        bufferDamage = ref.bufferDamage;
+    }
+
+    if (ref.updated.input)
+        input = ref.input;
+
+    if (ref.updated.opaque)
+        opaque = ref.opaque;
+
+    if (ref.updated.offset)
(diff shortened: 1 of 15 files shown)


gulafaran:
currently discord typing is occasionally delayed, not updating. minecraft flickers, cs2 flickers.

gulafaran:
and obs -> right click preview -> open windowed projector should die with 

```
[ 653985.111] {Display Queue} wl_display#1.error(wp_linux_drm_syncobj_surface_v1#77, 4, "Missing acquire timeline")
wp_linux_drm_syncobj_surface_v1#77: error 4: Missing acquire timeline
```

as it does on -git, currently doesnt.

ikalco:
> and obs -> right click preview -> open windowed projector should die with
> 
> ```
> [ 653985.111] {Display Queue} wl_display#1.error(wp_linux_drm_syncobj_surface_v1#77, 4, "Missing acquire timeline")
> wp_linux_drm_syncobj_surface_v1#77: error 4: Missing acquire timeline
> ```
> 
> as it does on -git, currently doesnt.

fixed

ikalco:
works on nvidia now, no flicker or delayed frames :)
I had to dig up my old nvidia card, and thank god I could repro lol

PlasmaPower:
Did you test this with the vkcube FIFO mode Tom mentioned was broken? That might've been from another PR but still worth testing here

ikalco:
tested with explicit_sync both on and off, all work on nvidia
* minecraft, vsync and unlimited, windowed and ds
* csgo, windowed and ds
* obs recording
* `vkcube --wsi wayland --present_mode {number here}` with 0, 1, and 2

ikalco:
@vaxerski do you mind if we just merge this as one pr lol, cause I don't really wanna go through all the combinations of testing all over again for #9804

maybe we can do something like
git squash the buffer ref stuff into one commit
git squash the surface stuff into another
and merge those 2 kernel style like you said before

p.s. the other two prs aren't required for this, still drafts until I get the time to clean/fix them up and test

PlasmaPower:
I've been running this PR for a couple days and it seems to be working well for my usecases on NVIDIA.

ikalco:
rebased to main

vaxerski:
my question is what is this supposed to fix?

ikalco:
> my question is what is this supposed to fix?

there isn't a specific bug this pr is fixing, but it is building towards more well defined surface commit code
so that in something like my proper buffer release pr (#9807) I know what gets called where,
and can hopefully fix the last of those pesky nvidia flickering bugs (#9640)

this pr also follows compositor protocol closer, like in the last 2 commits
and while I was in there changing all of that, I also decided to clean up SSurfaceState by
* making `updated` easier to use
* setting the right initial values based on the protocol text
* grouping functions and adding some clarifying comments

----

p.s.
<details><summary>here are some exmaples of previously not well defined behavior</summary>

* previously, buffer refs were tracked using a `SP<CHLBufferReference>` which has 2 ref counting methods, the SP and buffer->nLocks... that's confusing and unneeded, especially for having proper buffer releases
* release and acquire points were held in the CHLBufferReference, [which is wrong](https://github.com/hyprwm/Hyprland/pull/9774#discussion_r2019681644)
  this pr moves the acquire point to SurfaceState,
  and the release will be done properly in the later release stuff pr
* some commitPendingState calls were obscured through a events.precommit.emit() and done in DRMSyncobj protocol
  which again adds some confusion as to how and where state is committed
  and sometimes the DRMSyncobj protocol committed state incorrectly
  ```cpp
  if (!surface->pending.buffer && !PENDING_HAS_NEW_BUFFER && surface->current.buffer) {
          surface->current.bufferDamage.clear();
          surface->current.damage.clear();
          surface->commitPendingState(surface->current);
          return; // no new buffer, but we still have current around and a commit happend, commit current again.
    }
  ```

* previously, the pending surface state after a commit was being reset in multiple places in different ways, now we just have SSurfaceState::reset() to clearly define that new pending state
* simlarly to right above, the SSurfaceState::updateFrom just copied everything when a buffer commit happened, which is wrong. Now we only copy committed things
  ```cpp
  if (ref.updated & SURFACE_UPDATED_BUFFER) {
          ref.updated &= ~SURFACE_UPDATED_BUFFER;
          *this = ref;
          ref.damage.clear();
          ref.bufferDamage.clear();
          ref.buffer = {};
  }
  ```

</details>

src/protocols/types/SurfaceState.hpp
```diff
@@ -2,44 +2,59 @@
...
-    wl_output_transform    transform = WL_OUTPUT_TRANSFORM_NORMAL;
-    int                    scale     = 1;
-    SP<CHLBufferReference> buffer; // buffer ref will be released once the buffer is no longer locked. For checking if a buffer is attached to this state, check texture.
-    SP<CTexture>           texture;
-    Vector2D               offset;
-    Vector2D               size, bufferSize;
+    union {
+        uint16_t all = 0;
+        struct {
+            bool buffer : 1;
```
vaxerski, line 14: first time I see this syntax what the fuck? xD

src/protocols/types/SurfaceState.cpp
```diff
@@ -45,49 +45,56 @@ void SSurfaceState::updateSynchronousTexture(SP<CTexture> lastTexture) {
...
     }
-    buffer->buffer->endDataPtr();
+    buffer->endDataPtr();
 }
 
 void SSurfaceState::reset() {
+    updated.all = false;
+
+    // After commit, there is no pending buffer until the next attach.
+    buffer = {};
```
vaxerski, line 55: who said so? I've seen apps that do:
```
attach buffer
commit
set scale 
set something
commit
attach buffer 2 
commit
...
```

ikalco: https://wayland.app/protocols/wayland#wl_surface:request:attach
>The initial surface contents are void; there is no content. wl_surface.attach assigns the given wl_buffer as the pending wl_buffer. wl_surface.commit makes the pending wl_buffer the new surface contents, and the size of the surface becomes the size calculated from the wl_buffer, as described above. **After commit, there is no pending buffer until the next attach.**

src/protocols/types/SurfaceState.hpp
ikalco: each bool takes up one bit in the 2byte union and compiler does all the bit shifting stuff for us so its nice to use
I can make it just a struct of bools if you want, but I bet the compiler will just optimize it like above

PlasmaPower: If you made a struct of bools normally I think the compiler would have to put each in its own byte for alignment and the ability to take a pointer to each of them

ikalco: maybe  ¯\_(ツ)_/¯, who knows its the cpp compiler

src/protocols/types/SurfaceState.cpp
vaxerski: oke, I cant read, am a zoomer

src/protocols/types/SurfaceState.hpp
ikalco: https://www.programiz.com/online-compiler/65xnymqumVjNH

`dothething()` basically ANDs `a` and `b`, so only true if both `a` and `b` are true
and it only works in that scenario :+1: 

also, heres a reference for the syntax, its called a [bit-field](https://en.cppreference.com/w/cpp/language/bit_field)

vaxerski: huh. I know what a bitfield is, just never seen this syntax. Great, guess we all learn every day. Thanks!

vaxerski:
cs works fine on my end so yea

vaxerski (approved):

PR #9815, opened 2025-03-31
Don't force default cursor on config reload/monitor reconfigure

X2E4VXpZKv:

When you change monitor rotation/reload config it forces the cursor to be visible even when you were using a touchscreen before, and it also gets stuck like that until you move it with a mouse.
The really annoying part is the getting stuck bug but its also silly to make the cursor visible when you might not even have a touchpad/mouse connected. (also i dont know how i would specifically fix that)

Idk it doesnt seem to break things?

### DIFF
--- a/src/managers/CursorManager.cpp
+++ b/src/managers/CursorManager.cpp
@@ -302,8 +302,6 @@ void CCursorManager::updateTheme() {
             m_pHyprcursor->loadThemeStyle(m_sCurrentStyleInfo);
     }
 
-    setCursorFromName("left_ptr");
-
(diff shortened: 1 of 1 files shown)


vaxerski (approved):

user111111111111111111111111111111111:
after this commit, nested sessions have a large chance to crash (with default config) (`cursor:use_cpu_buffer = 0` makes it smaller but makes the cursor theme overlap multiple cursors)
[hyprlandCrashReport21703.txt](https://github.com/user-attachments/files/19622226/hyprlandCrashReport21703.txt)

vaxerski:
can you open a discussion? In general this is probably a problem elsewhere

Issue #9771, opened 2025-03-28
Did not get update news popup for 0.48.0


winkelnp:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

Probably not a regression / I don't remember it happening before

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.48.0 built from branch  at commit 5ee35f914f921e5696030698e74fb5566a804768  (version: bump to 0.48.0).
Date: Sun Mar 23 14:56:13 2025
Tag: v0.48.0, commits: 5922
built against:
 aquamarine 0.8.0
 hyprlang 0.6.0
 hyprutils 0.5.2
 hyprcursor 0.1.12
 hyprgraphics 0.1.2


no flags were set


System Information:
System name: Linux
Node name: archlinux
Release: 6.13.8-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Sun, 23 Mar 2025 17:17:30 +0000


GPU information: 
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Phoenix4 [1002:1901] (rev c2) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:


```

</details>

### Description

The update news popup did not show on upgrade to 0.48.0 although `no_update_news = false`

### How to reproduce

1. Upgrade from 0.47.2-4 on archlinux to 0.48.0-1
2. restart Hyprland

### Attach not paste

- [x] I understand that all text files must be *attached*, and not pasted directly. If not respected, this issue will likely get closed as spam

### Checklist of files to include below

- [x] Hyprland config - `hyprctl systeminfo -c` (always include)
- [ ] Crash report (always include in case of crash)
- [ ] Video (always include in case of a visual bug)
- [x] Logs (might contain useful info such as errors)

### Additional info & File uploads

[...]

PR #9829, opened 2025-04-01
versionkeeper: create version file if not present

arminius-smh:

I think this should fix https://github.com/hyprwm/Hyprland/issues/9771 by creating the version file if it does not exist?
I guess initially this was done here when it was introduced: https://github.com/hyprwm/Hyprland/blob/bba2d9a1971b51a25a5ae3fc30a453e106e6b8a9/src/managers/VersionKeeperManager.cpp#L105C1-L110C6

( https://github.com/hyprwm/Hyprland/commit/bba2d9a1971b51a25a5ae3fc30a453e106e6b8a9 set it to 0.0.0 so I did that the same way here as well again )

but it seems like it got accidentally removed in https://github.com/hyprwm/Hyprland/commit/b5fb6110ab1f75440b4b90d09f95d304d8c2080b with the refactor of the function and the addition to the donation nag

so right now after installing Hyprland and starting it the first time, the update popup with the version gets shown, since I suppose that was the intention by setting it to 0.0.0

i guess it is ready

### DIFF
--- a/src/managers/VersionKeeperManager.cpp
+++ b/src/managers/VersionKeeperManager.cpp
@@ -26,10 +26,12 @@ CVersionKeeperManager::CVersionKeeperManager() {
     if (!DATAROOT)
         return;
 
-    const auto LASTVER = NFsUtils::readFileAsString(*DATAROOT + "/" + VERSION_FILE_NAME);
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
right, I'm dumb, I missed that

PR #9838, opened 2025-04-03
fix(dwindle): ignore fullscreen window for positioning when use_active_for_splits=false

rszyma:
#### Repro Steps: 
1. Use dwindle + `use_active_for_splits=0`
2. have 3 windows open next to each other (layed out horizontally and spaced evenly)
3. maximize the leftmost one
4. move mouse pointer near the right edge of the screen
5. open a new window (e.g. a terminal)

#### Actual:

The new window opens next to leftmost window.

#### Expected:

The new window opens next to the rightmost window - nearby where the mouse cursor is. That's what `use_active_for_splits=0` should do (from wiki: "prefer mouse position for splits")

Note, that this behavor is exactly what would happen if you inserted an extra step of minimizing the fullscreen window between the steps 4 and 5. And this PR attempts to do exacly this, trying to ignore the current fullscreen window when creating a new one.

#### Before vs after videos

In the following videos I have also set `smart_split=1`, which makes the issue more pronounced. Btw on videos I'm not doing the exact steps above, sowwy about that (they are old vids), but it should be clear enough I hope.

Before: 

<details>
<summary>Details</summary>

https://github.com/user-attachments/assets/722bfe01-0de8-44b5-a57b-75285f5c79a0

</details>

After:

<details>
<summary>Details</summary>

https://github.com/user-attachments/assets/80b62702-182f-41a8-a851-2231d4a60889

</details>

It actually used to work like this before a year ago or so, and it's been annoying me for a long time. I've had enough.

should be ready to go, unless we want to make this configurable.

### DIFF
--- a/src/desktop/Window.hpp
+++ b/src/desktop/Window.hpp
@@ -44,14 +44,14 @@ enum eGroupRules : uint8_t {
 };
 
 enum eGetWindowProperties : uint8_t {
-    WINDOW_ONLY      = 0,
-    RESERVED_EXTENTS = 1 << 0,
-    INPUT_EXTENTS    = 1 << 1,
-    FULL_EXTENTS     = 1 << 2,
-    FLOATING_ONLY    = 1 << 3,
-    ALLOW_FLOATING   = 1 << 4,
-    USE_PROP_TILED   = 1 << 5,
-    NON_FULLSCREEN   = 1 << 6,
+    WINDOW_ONLY              = 0,
+    RESERVED_EXTENTS         = 1 << 0,
+    INPUT_EXTENTS            = 1 << 1,
+    FULL_EXTENTS             = 1 << 2,
+    FLOATING_ONLY            = 1 << 3,
+    ALLOW_FLOATING           = 1 << 4,
+    USE_PROP_TILED           = 1 << 5,
+    SKIP_FULLSCREEN_PRIORITY = 1 << 6,
 };
 
 enum eSuppressEvents : uint8_t {
--- a/src/desktop/Window.hpp
+++ b/src/desktop/Window.hpp
@@ -51,6 +51,7 @@ enum eGetWindowProperties : uint8_t {
     FLOATING_ONLY    = 1 << 3,
     ALLOW_FLOATING   = 1 << 4,
     USE_PROP_TILED   = 1 << 5,
+    NON_FULLSCREEN   = 1 << 6,
 };
 
 enum eSuppressEvents : uint8_t {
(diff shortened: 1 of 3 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -246,7 +246,7 @@ void CHyprDwindleLayout::onWindowCreatedTiling(PHLWINDOW pWindow, eDirection dir
 
     if (PMONITOR->ID == MONFROMCURSOR->ID &&
         (PNODE->workspaceID == PMONITOR->activeWorkspaceID() || (g_pCompositor->isWorkspaceSpecial(PNODE->workspaceID) && PMONITOR->activeSpecialWorkspace)) && !*PUSEACTIVE) {
-        OPENINGON = getNodeFromWindow(g_pCompositor->vectorToWindowUnified(MOUSECOORDS, RESERVED_EXTENTS | INPUT_EXTENTS));
+        OPENINGON = getNodeFromWindow(g_pCompositor->vectorToWindowUnified(MOUSECOORDS, RESERVED_EXTENTS | INPUT_EXTENTS | NON_FULLSCREEN));
```
vaxerski, line 249: what about the case when we really do open above the fullscreen window in the layout? Let's say the top left window is fullscreened, and our cursor is above the top left quadrant.

rszyma: you mean sth like this? Seems to work as expected:

https://github.com/user-attachments/assets/387d9bab-3c91-4682-92ec-2d0009106c50

rszyma: no, it's just how `smart_split=1` works, which I have enabled, but btw it's not necessary to trigger the issue I'm tryna fix. I think u misunderstood what I mean, which is totally my bad for unclear writing. I've just updated the OP, so take a look again.

vaxerski: no, smart split is only for deciding which side, not which orientation?

rszyma: from wiki on smart_split:
> if enabled, allows a more precise control over the window split direction based on the cursor’s position. The window is conceptually divided into four triangles, and cursor’s triangle determines the split direction. This feature also turns on preserve_split.	

it says about 4 triangles so I'd assume it's to enable splitting into all 4 directions. Otherwise what this option would do compared to it being disabled? Idk, I don't see any issue with it.

vaxerski: oops, mistook for something else. My bad.

src/desktop/Window.hpp
```diff
@@ -51,6 +51,7 @@ enum eGetWindowProperties : uint8_t {
     FLOATING_ONLY    = 1 << 3,
     ALLOW_FLOATING   = 1 << 4,
     USE_PROP_TILED   = 1 << 5,
+    NON_FULLSCREEN   = 1 << 6,
```
vaxerski, line 54: misleading for what it does. It doesn't "return a non-fullscreen window" but rather "skip fullscreen window's priority"

I suggest `SKIP_FULLSCREEN_PRIORITY`

vaxerski: fwiw that's what made me think it'd break with the fs window

rszyma: yeah bad name. renamed in 3a18d46

vaxerski (approved):
:+1:

PR #9780, opened 2025-03-29
layout: center floating window at cursor when picked up from fullscreen

nnyyxxxx:
when picking up a floating window after it had been fullscreened before it would return to its previous position which looked ugly because the cursor could be no where near the windows original position, this patch makes it so that the window is returned to the users current cursor position

video of the previous behavior:

https://github.com/hyprwm/Hyprland/issues/9779#issuecomment-2763122913

it's wayyy better now

fixes #9779

PR #9839, opened 2025-04-03
binds: drag_threshold for click/drag isolation

### LINKED ISSUE #9631: mutually exclusive bindm and bind on the same mouse button

alaricljs:
This code was abandoned in #9658 and I'm picking it back up.

New features

    Added binds:drag_threshold to yoink windows with mouse binds only after moving a few pixels
    Added bindc / bindg (click / drag) flags for release binds that check if inside/outside drag_threshold on release

Threshold handling is slightly different whether there is a bindm on the same button or not:

    If there is no bindm on the same button, any amount of movement that ends inside the threshold will trigger a click
    If bindm exists and has been triggered, returning inside the threshold will not trigger a click (triggers drag)

These are mutually exclusive.

Simple example that adds more window handling to the same 2 buttons:

binds {
    drag_threshold = 20
}

bindm = SUPER, mouse:272, movewindow
bindc = SUPER, mouse:272, togglefloating
bindm = SUPER, mouse:273, resizewindow
bindc = SUPER, mouse:273, fullscreen
                                                                                                                       
Thoughts

The default threshold = 0 should work like current default behavior, i.e. windows are grabbed immediately on mousedown. With threshold > 0 yoinking happens later on mousemove.

This probably obsoletes #8143 and fixes #9631.

Is it ready for merging:
- It was nearly at the finish line.  I removed the duplicate modifactions of #9780
- Wiki PR https://github.com/hyprwm/hyprland-wiki/pull/1023

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1326,6 +1326,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_BOOL,
         .data        = SConfigOptionDescription::SBoolData{false},
     },
+    SConfigOptionDescription{
+        .value       = "binds:drag_threshold",
+        .description = "Movement threshold in pixels for window dragging and c/g bind flags. 0 to disable and grab on mousedown.",
+        .type        = CONFIG_OPTION_INT,
+        .data        = SConfigOptionDescription::SRangeData{0, 0, INT_MAX},
+    },
 
     /*
      * xwayland:
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -639,6 +639,7 @@ CConfigManager::CConfigManager() {
     registerConfigVar("binds:disable_keybind_grabbing", Hyprlang::INT{0});
     registerConfigVar("binds:window_direction_monitor_fallback", Hyprlang::INT{1});
     registerConfigVar("binds:allow_pin_fullscreen", Hyprlang::INT{0});
+    registerConfigVar("binds:drag_threshold", Hyprlang::INT{0});
 
     registerConfigVar("gestures:workspace_swipe", Hyprlang::INT{0});
     registerConfigVar("gestures:workspace_swipe_fingers", Hyprlang::INT{3});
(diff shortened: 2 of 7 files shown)


vaxerski:
ping me tomorrow pls

alaricljs:
ping @vaxerski, a day late but I'm not at my desk much on weekends.

vaxerski (approved):
looks pretty ok to me. Needs a wiki mr

vaxerski:
hm. Now that I look at it, wouldn't c and g be better as _extensions_ to m?

`bindmc` `bindmg`

and error out if it's not `m`? Otherwise it's less intuitive

Dregu:
I originally though about including the `m` too, but decided it would also be confusing, because `m` implies you can *only use mouse binds*, which is not true at all with `c/g`. I guess it's just about how you document it. In the end I did it this way simply because I thought the other way would've made the flag parsing code and documentation more complicated *for me*, and obviously I'm lazy.

Anyway glad you picked this up again.

alaricljs:
As Dregu points out, the drag_threshold is measured for c, g, and m bind flags.  c and g are only applicable to non-m binds since drag_threshold is fundamentally part of bindm.  Are you looking to have a bindm option to toggle off drag_threshold?

alaricljs:
I've pushed a wiki change with hopefully a clearer explanation/description.

vaxerski:
yes, that's clearer. Thanks

nnyyxxxx:
ugh. this reverts a few commits

nnyyxxxx:
> It was nearly at the finish line.  I removed the duplicate modifactions of [layout: center floating window at cursor when picked up from fullscreen #9780](https://github.com/hyprwm/Hyprland/pull/9780)

that's wrong since you are moving the same affected codepaths around, https://github.com/hyprwm/Hyprland/pull/9658#issuecomment-2764752746 vaxry was wrong here as well #9780 should have been supersceded by #9658

Dregu:
Oh no exactly the thing I said would happen happened /s

PR #9874, opened 2025-04-05
fix: disconnected monitors are reconnected to an empty workspace

matjam:

Fixes the issue around https://github.com/hyprwm/Hyprland/issues/3248

The issue is that when a monitor is disconnected, its removed from the list of real monitors; which is fine but that list isn't really a list of actual real monitors. The state which tracks which workspace the monitor is on is lost, however.

When that monitor is reconnected, then it gets assigned a new workspace.

So for monitors that don't really "sleep" but rather just turn off, your idle daemon will sleep the display, it will disconnect, and when you wake up you end up on an empty workspace.

This patch adds some state to track the last workspace a monitor was seen using, and reassigns that workspace on a reconnect. This should degrade cleanly, as the usual process to reorder workspaces runs and sorts out any gaps, etc.

AFAIK no.

Its tested by me, but other testing would be good.


Mabi19:
big if true, that issue has been so annoying for me lately :P

If I have time I'll try to test this, but no guarantees

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,21 @@ void CMonitor::onConnect(bool noRule) {
...
+        Debug::log(LOG, "Monitor {} was on workspace {}", szName, it);
+
+        auto ws = g_pCompositor->getWorkspaceByID(it);
+        if (ws) {
+            activeWorkspace             = ws;
+            activeWorkspace->m_bVisible = true;
+        }
+    } else {
+        Debug::log(LOG, "Monitor {} was not on any workspace", szName);
+    }
```
nnyyxxxx, line 251:
```suggestion
    } else
        Debug::log(LOG, "Monitor {} was not on any workspace", szName);
```

src/Compositor.hpp
```diff
@@ -39,39 +39,41 @@ class CCompositor {
     std::string                                m_szCurrentSplash     = "error";
 
     std::vector<PHLMONITOR>                    m_vMonitors;
-    std::vector<PHLMONITOR>                    m_vRealMonitors; // for all monitors, even those turned off
+    std::vector<PHLMONITOR>                    m_vRealMonitors; // for all monitors, even those turned off <- lies, turned off monitors are removed
```
matjam, line 42: this is a little snarky, sorry

src/Compositor.hpp
```diff
@@ -39,39 +39,41 @@ class CCompositor {
...
-    PHLMONITORREF                              m_pUnsafeOutput; // fallback output for the unsafe state
-    bool                                       m_bIsShuttingDown         = false;
-    bool                                       m_bFinalRequests          = false;
-    bool                                       m_bDesktopEnvSet          = false;
-    bool                                       m_bWantsXwayland          = true;
-    bool                                       m_bOnlyConfigVerification = false;
+    std::unordered_map<std::string, WORKSPACEID>
+         m_mDisconnectedMonitorWorkspaceMap; // a map of monitor names to workspace IDs, used for when a monitor is disconnected and we want to preserve what workspace it had
+
+    void initServer(std::string socketName, int socketFd);
```
matjam, line 53: unavoidable reformat, clang-format insists

src/helpers/Monitor.cpp
nnyyxxxx: https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/#other

src/Compositor.hpp
```diff
@@ -39,39 +39,41 @@ class CCompositor {
     std::string                                m_szCurrentSplash     = "error";
 
     std::vector<PHLMONITOR>                    m_vMonitors;
-    std::vector<PHLMONITOR>                    m_vRealMonitors; // for all monitors, even those turned off
+    std::vector<PHLMONITOR>                    m_vRealMonitors; // for all monitors.
```
matjam, line 42: the "even those turned off" comment led me quite astray for a while. But its not accurate, so I removed it.

I'm actually not sure there's any difference between these two vectors at this point.

vaxerski: how is it not accurate? If you have a monitor that is `disable`d in the config, it will be in real but not in monitors.

src/Compositor.hpp
vaxerski: because your comment is long. Put it on the line above, not the same line.

src/Compositor.hpp
matjam: turned off monitors seem to get removed from that list. At least, with my specific hardware they do.

vaxerski: that might be a weird controller on your display, but in general, they shouldnt

src/Compositor.hpp
```diff
@@ -25,53 +25,54 @@ class CCompositor {
...
+    std::vector<PHLMONITOR>                      m_vMonitors;
+    std::vector<PHLMONITOR>                      m_vRealMonitors; // for all monitors, even those turned off
+    std::vector<PHLWINDOW>                       m_vWindows;
+    std::vector<PHLLS>                           m_vLayers;
+    std::vector<PHLWORKSPACE>                    m_vWorkspaces;
+    std::vector<PHLWINDOWREF>                    m_vWindowsFadingOut;
+    std::vector<PHLLSREF>                        m_vSurfacesFadingOut;
+
+    std::unordered_map<std::string, MONITORID>   m_mMonitorIDMap;
+    std::unordered_map<std::string, WORKSPACEID> m_mSeenMonitorWorkspaceMap; // map of seen monitor names to workspace IDs
```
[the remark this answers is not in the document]
vaxerski: no, that's ok

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,20 @@ void CMonitor::onConnect(bool noRule) {
         }
     }
 
+    Debug::log(LOG, "checking if we have seen this monitor before: {}", szName);
+    // if we saw this monitor before, set it to the workspace it was on
+    auto it = g_pCompositor->m_mSeenMonitorWorkspaceMap[szName];
+    if (it != 0) {
```
vaxerski, line 241: g_pCompositor->m_mseenmonitorworkspacemap.contains()

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,20 @@ void CMonitor::onConnect(bool noRule) {
...
+        Debug::log(LOG, "Monitor {} was on workspace {}, setting it to that", szName, it);
+
+        auto ws = g_pCompositor->getWorkspaceByID(it);
+        if (ws) {
+            activeWorkspace             = ws;
```
vaxerski, line 246: wrong wrong wrong, use moveworkspacetomonitor

matjam: :heart:

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,20 @@ void CMonitor::onConnect(bool noRule) {
         }
     }
 
+    Debug::log(LOG, "checking if we have seen this monitor before: {}", szName);
+    // if we saw this monitor before, set it to the workspace it was on
+    if (g_pCompositor->m_mseenmonitorworkspacemap.contains(szName)) {
```
vaxerski, line 240: eeee did you even compile and test this? This won't compile as you forgot the capitalization

matjam: I'll click the re-request button when ready

no I didn't - I made the change and pushed so I could test on my personal machine when I get a minute - I can't build Hyprland on macos.

matjam: also you can highlight code with the suggest change button in the conversation and suggest a change

image

might save some back and forth

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,28 @@ void CMonitor::onConnect(bool noRule) {
...
+            if (m.get() == this) {
+                THISMONITOR = m;
+                break;
+            }
+        }
+
+        auto ws = g_pCompositor->getWorkspaceByID(workspaceID);
+        if (ws) {
+            g_pCompositor->moveWorkspaceToMonitor(ws, THISMONITOR);
+            this->changeWorkspace(ws, true, true, true);
```
matjam, line 255: this appears to be the right way to move the workspace and make it active on this monitor?

re the additional parameters to `changeWorkspace()`; the first two seem reasonable, based on what I see, I'm not sure what `noFocus` is doing.

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,28 @@ void CMonitor::onConnect(bool noRule) {
...
+            if (m.get() == this) {
+                THISMONITOR = m;
+                break;
+            }
+        }
+
+        auto ws = g_pCompositor->getWorkspaceByID(workspaceID);
+        if (ws) {
+            g_pCompositor->moveWorkspaceToMonitor(ws, THISMONITOR);
+            this->changeWorkspace(ws, true, true, false);
```
matjam, line 255: this appears to be the right way to move the workspace and make it active on this monitor?

re the call to `changeWorkspace()`, `internal = true` avoids animation however `noMouseMove` and `noFocus` appear unused if `internal == true`, which seems appropriate in this context.

matjam:
Built and tested locally, appreciate additional testing from those _not_ experiencing this issue to ensure I've not introduced any unwanted side-effects.

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,28 @@ void CMonitor::onConnect(bool noRule) {
...
+        auto workspaceID = g_pCompositor->m_mSeenMonitorWorkspaceMap[szName];
+        Debug::log(LOG, "Monitor {} was on workspace {}, setting it to that", szName, workspaceID);
+
+        PHLMONITOR THISMONITOR = nullptr;
+        for (auto const& m : g_pCompositor->m_vMonitors) {
+            if (m.get() == this) {
+                THISMONITOR = m;
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
vaxerski: monitor->self.lock() would be the best

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,28 @@ void CMonitor::onConnect(bool noRule) {
...
+
+        auto ws = g_pCompositor->getWorkspaceByID(workspaceID);
+        if (ws) {
+            g_pCompositor->moveWorkspaceToMonitor(ws, THISMONITOR);
+            this->changeWorkspace(ws, true, false, false);
```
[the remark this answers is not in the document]
vaxerski: yeah, this is fine

src/helpers/Monitor.cpp
```diff
@@ -235,6 +235,19 @@ void CMonitor::onConnect(bool noRule) {
...
     }
 
+    Debug::log(LOG, "checking if we have seen this monitor before: {}", szName);
+    // if we saw this monitor before, set it to the workspace it was on
+    if (g_pCompositor->m_mSeenMonitorWorkspaceMap.contains(szName)) {
+        auto workspaceID = g_pCompositor->m_mSeenMonitorWorkspaceMap[szName];
+        Debug::log(LOG, "Monitor {} was on workspace {}, setting it to that", szName, workspaceID);
+        auto ws = g_pCompositor->getWorkspaceByID(workspaceID);
+        if (ws) {
+            g_pCompositor->moveWorkspaceToMonitor(ws, this->self.lock());
```
vaxerski, line 245: we don't use `this->`

vaxerski (commented):
rest lgtm

src/helpers/Monitor.cpp
matjam: old habits

vaxerski (approved):
thanks!

Issue #9915, opened 2025-04-06
Implement permission controls for non-portal screencopy sessions


vaxerski:
### Discussion Link
#9896

## Description
screencopy sessions that do not go through the xdg-desktop-portal are not monitored and can be done at any point by any client. That's how it's always worked, and that's how it works on any compositor.

Instead, add a popup asking for permission before sharing. While the user did not act on it, we can render black. If the user denied it, we can either render black or a "denied" texture, or something.

This has to be done before ext-image-copy-capture.

PR #9919, opened 2025-04-06
helpers: Add an async dialog box impl

vaxerski:
This depends on hyprutils 0.6.0, and adds a much more robust, easy to use and safe dialog box implementation, also makes it contained instead of exclusive in ANR.

Also makes the ANR dialog use it.

First part of #9915


vaxerski:
@fufexan can I get a nix bump

vaxerski:
bazinga

PR #9939, opened 2025-04-06
Nix/module: add defaultSettings

fufexan:

TO BE MERGED AFTER #9930

Add `defaultSettings` option in the NixOS module. Can be disabled by setting it to `{}`.

Default settings currently includes allowing screencopy for XDPH and grim, and optionally adding `exec-once = uwsm finalize` in case `withUWSM = true`.

Brainstorming is welcome. This is what made sense to me at the time of opening this PR.

Untested, no.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -9,6 +9,25 @@ let
   inherit (pkgs.stdenv.hostPlatform) system;
   selflib = import ./lib.nix lib;
   cfg = config.programs.hyprland;
+
+  hyprlangType =
+    with lib.types;
+    let
+      valueType =
+        nullOr (oneOf [
+          bool
+          int
+          float
+          str
+          path
+          (attrsOf valueType)
+          (listOf valueType)
+        ])
+        // {
+          description = "Hyprland configuration value";
+        };
+    in
+    valueType;
 in
 {
   options = {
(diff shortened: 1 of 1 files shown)


fxzzi:
Are you sure we need `uwsm finalize`? According to the uwsm docs it states:

> TLDR; if your compositor puts WAYLAND_DISPLAY (and along with it DISPLAY, or other important or useful variables) into systemd activation environment, uwsm will make everything work automagically, proceed to section 3.

> Otherwise configure compositor to run uwsm finalize command at the end of its startup. It will deal with putting WAYLAND_DISPLAY and DISPLAY (if set) variables into activation environments in the best possible ways and signal unit readiness to systemd.

I am assuming hyprland does set these as uwsm wants and therefore the manual finalize shouldn't be needed

I am assuming this since I don't use this finalize command and I haven't faced any issues regarding unset `WAYLAND_DISPLAY` or `DISPLAY`, let me know if I'm incorrect here though.

fufexan:
It may be a remnant from the early days of uwsm support, as it was needed sometimes to properly finish setting up the session.

If it's no longer needed, I'll remove it.

fufexan:
Another possible approach would be to have a toggle `enableDefaultConfig` explaining what settings are being added, since it's something you either want or don't want.

But I believe this would hurt visibility, as having the option an attrset directly shows its value as `default`.

PR #9945, opened 2025-04-06
input: add warp_on_toggle_special

VirtCode:
Currently the `cursor:warp_on_change_workspace` option does only apply to changing normal workspaces and not to toggling special ones. This has been bothering me for quite some time because it seems unintuitive and I would often accidentally switch windows on my special workspace because of it.

This PR also warps the cursor on special workspace toggles when `cursor:warp_on_change_workspace` is enabled. I think this should be the expected behavior because to me opening a special workspace is also a "change of workspace".

I can imagine that some users might prefer the old behaviour (i.e. only warping cursor on normal workspace change). So maybe it would be best to gate this behind an additional config var? I also thought about using a different number for `warp_on_change_workspace` to enable it for specials too, but that wouldn't really make too much sense because `2` is already used for forcing.

Yes

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1476,7 +1476,8 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     },
     SConfigOptionDescription{
         .value       = "cursor:warp_on_toggle_special",
-        .description = "Move the cursor to the last focused window when toggling a special workspace. Options: 0 (Disabled), 1 (Enabled), 2 (Force - ignores cursor:no_warps option)",
+        .description = "Move the cursor to the last focused window when toggling a special workspace. Options: 0 (Disabled), 1 (Enabled), "
+                       "2 (Force - ignores cursor:no_warps option)",
         .type        = CONFIG_OPTION_CHOICE,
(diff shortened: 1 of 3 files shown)


vaxerski:
does this mean the cursor will warp to the center if you toggle special workspace on the current monitor? if so, I'd hate that, so definitely a config var

VirtCode:
If there are windows open on the special workspace, then it would warp to the focused window there (so to its center if not using `persistent_warps`). If there is no window open on it, no warping happens. When closing the special workspace, it would warp back to the window that is focused on the current normal workspace. So yes, if you open and close it, without `persistent_warps`, the cursor would be warped to the focused window's center.

As for the config var, should I do an entirely separate one (e.g. `warp_on_toggle_special`) or gate it behind an additional number for `warp_on_change_workspace` (and if so, which number)?

vaxerski:
up 2 u

VirtCode:
Okay, have now decided to gate it behind `warp_on_toggle_special`. It has the same options as `toggle_on_change_workspace`, i.e. allows being set to 2 to force over no warps.

Also made a wiki mr here https://github.com/hyprwm/hyprland-wiki/pull/1029

vaxerski (approved):

Issue #8891, opened 2024-12-30
commit deb077c causes screen freezes on multi monitor nvidia setups


Fxzzi:
### Already reported ? *

- [X] I have searched the existing open and closed issues.

### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.46.0 built from branch main at commit deb077c34653eab251bfe71970bb5987913e7c6d  (ctm: add an internal fade animation to ctm transitions).
Date: Sun Dec 29 18:21:20 2024
Tag: v0.46.0-45-gdeb077c3, commits: 5600
built against:
 aquamarine 0.5.1
 hyprlang 0.6.0
 hyprutils 0.3.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set


System Information:
System name: Linux
Node name: Kappa
Release: 6.12.7-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Fri, 27 Dec 2024 14:24:32 +0000


GPU information: 
07:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA104 [GeForce RTX 3070] [10de:2484] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  565.77  Release Build  (archlinux-builder@)  


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

======Config-Start======


Config File: /home/faaris/.config/hypr/hyprland.conf: Read Succeeded
# █▀▀ ▀▄▀ █▀▀ █▀▀
# ██▄ █░█ ██▄ █▄▄
exec-once		= uwsm app -- random-wall.sh
exec-once		= uwsm app -- gsettings.sh # Applies gtk themes, cursor themes, etc
exec-once		= uwsm app -- /etc/xdg/autostart/polkit-mate-authentication-agent-1.desktop
[...]

PR #9976, opened 2025-04-07
CTMControl: allow CTM animation to work on Nvidia IF there is only one monitor enabled

fxzzi:

This PR adds an extra check for `ctm_animation = 2` so that if Nvidia is in use BUT there is only one monitor enabled / attached, still allow ctm animations to work. The bug from #8891 only applies to multi monitor setups, so we should still allow them on single monitor setups.

Idk, might need testing from other nvidia users but works fine on my end. I temporarily disabled my secondary monitor with `hyprctl -i 0 keyword monitor desc:Philips,disable`, and triggered hyprsunset, which correctly displayed the ctm animations without freezes. Upon enabling my secondary monitor again, the animation was correctly skipped as it would cause freezes otherwise.

Should be

### DIFF
--- a/src/protocols/CTMControl.cpp
+++ b/src/protocols/CTMControl.cpp
@@ -109,7 +109,7 @@ bool CHyprlandCTMControlProtocol::isCTMAnimationEnabled() {
     static auto PENABLEANIM = CConfigValue<Hyprlang::INT>("render:ctm_animation");
 
     if (*PENABLEANIM == 2)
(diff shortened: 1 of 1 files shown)


vaxerski:
I'm generally against. I'd rather wait for nvidia to fix the bug rather than "lets check 20 things and pray to god nvidia doesn't shit itself this time on this setup at this hour of this day because it doesnt on tuesdays"

PR #9984, opened 2025-04-07
input/layers: Fix exclusive LS focus / refocus after unmap

### LINKED ISSUE #9980: After recovering focus from a layer, keyboard focus can get lost with follow_mouse = 2
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/9929

## Description

Basically:
- open 2 apps on 2 monitors
- focus one and execute slurp
- select a thing on the other monitor
- keyboard focus does not return, despite other focus. 

This only happens one way, not the other. The first app (buggy) has to be on a monitor with a lower id

vaxerski:
Fixes #9980

There are two parts at play here:
- firstly, refocusLastWindow should not attempt that when we have an active exclusive LS 
- secondly, mouseMoveUnified should force focus to exclusive LSes in general 

@user111111111111111111111111111111111 as op for testing, works on my end

### DIFF
--- a/src/desktop/LayerSurface.cpp
+++ b/src/desktop/LayerSurface.cpp
@@ -234,9 +234,10 @@ void CLayerSurface::onUnmap() {
 
     // refocus if needed
     //                                vvvvvvvvvvvvv if there is a last focus and the last focus is not keyboard focusable, fallback to window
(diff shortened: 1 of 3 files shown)


user111111111111111111111111111111111:
fixes it for me too, thanks

PR #9986, opened 2025-04-07
CursorManager: Store cursor pixel data retrieved from X/HC as a copy.

### LINKED ISSUE #9972: CursorManager's buffers pass around unsafe pointers to cairo / pixel data and fall off a bike
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/9965

## Description
As in the discussion, those raw pointers there are a dangerous thing. They were written way back in the day with wlroots in mind, but now that we don't use wlr, we can make them memory safe for once.

vaxerski:
Instead of storing pointers as refs (which could randomly get invalid very easily) copy the data. They aren't big anyways (4Bpp * ~32^2 = 4kB) and we gain memory safety.

Should fix #9972 but I haven't hit that myself so I'm pinging @zjeffer @onlycs and @nnyyxxxx for testing (which afaiu can repro this)

### DIFF
--- a/src/managers/CursorManager.cpp
+++ b/src/managers/CursorManager.cpp
@@ -18,12 +18,16 @@ static void hcLogger(enum eHyprcursorLogLevel level, char* message) {
     Debug::log(NONE, "[hc] {}", message);
 }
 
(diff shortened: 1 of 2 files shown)


onlycs:
I'll build and test this as soon as I get home

zjeffer (approved):
Fixed, thanks!

vaxerski:
https://github.com/user-attachments/assets/fcd48df3-b92d-4e83-859c-5aab9f37474b

onlycs:
can confirm, this fix works for me. thanks.

PR #9996, opened 2025-04-09
remove renderer finalDamage since it's unused

ikalco:
realized that finalDamage is just a copy of damage in renderMonitor,
then saw that nothing ever actually uses it

yes, did a quick test with blur on and off and it seemed fine


vaxerski (approved):

vaxerski:
reverting cuz #10036 you troller

PR #10031, opened 2025-04-10
core: recheck floating windows after monitor layout changes

### LINKED ISSUE #10030: Floating windows go offscreen on monitor layout changes
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/9956

## Description

when the monitor layout changes, no checks are done for floating windows. They should be moved to a nearby monitor.

vaxerski:
Rechecks the position of floating windows after monitor layout changes.

Fixes #10030

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2901,6 +2901,38 @@ void CCompositor::arrangeMonitors() {
     PROTO::xdgOutput->updateAllOutputs();
 }
 
+void CCompositor::recheckFloatingWindowsOnScreen() {
+    for (auto& w : m_vWindows) {
+        if (!w->m_bIsMapped || !w->m_bIsFloating)
+            continue;
+
+        const CBox WBOX = w->getWindowMainSurfaceBox();
+
+        const bool IN = std::ranges::any_of(m_vMonitors, [w, WBOX](const auto& m) {
+            if (!m->m_bEnabled)
+                return false;
+            CBox MONBOX = m->logicalBox();
+            return !MONBOX.intersection(WBOX).empty();
+        });
+
+        if (IN)
+            continue;
+
+        auto monitor = w->m_pMonitor;
+
+        if (!monitor || !monitor->m_bEnabled)
+            monitor = getMonitorFromVector(WBOX.middle());
+
+        if (!monitor)
+            continue; // ?!?!?!
+
+        const auto NEW_POS = monitor->middle() - WBOX.size() / 2.F;
+
+        *w->m_vRealPosition = NEW_POS;
+        w->m_vPosition      = NEW_POS;
+    }
+}
+
 void CCompositor::enterUnsafeState() {
     if (m_bUnsafeState)
         return;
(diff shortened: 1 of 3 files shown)


Kyuunex:
It no longer goes offscreen. Still buggy tho. 

When I revert back to dual screen, it goes to the left screen. After that, When I switch to a different workstation on the right screen, it disappears on the left screen because it technically still belongs to the workstation that was on the right screen and that isn't being displayed anymore.

viccooper142:
Still reproduces on **0.56.2** (`efb5099`), so this PR would still be valuable.

Worth noting that the two paths behave differently, which may narrow the fix.

Setup:

```
eDP-1     2560x1600 scale 1.6 @ 0x0      -> logical x[0..1600]
HDMI-A-1  1920x1080 scale 1   @ -1920x0  -> logical x[-1920..0]
```

**Config-driven monitor moves are already handled.** Changing a monitor's
position at runtime translates floating windows along with it:

```bash
hyprctl repl 'hl.monitor({output="HDMI-A-1", mode="preferred", position="1600x0", scale=1})'
# floating window follows: x[-923..-47] -> x[2597..3473]   OK, still visible
```

**Disconnect + reconnect is not.**

```bash
hyprctl repl 'hl.monitor({output="HDMI-A-1", disabled=true})'
hyprctl reload   # reapplies the configured -1920x0 position
```

After the monitor comes back, floating windows that lived on it are reassigned
to it but keep coordinates in **eDP-1's** coordinate space:

```
pid=5263  ws=2  monitor=HDMI-A-1  at=[996,252]  size=[876,600]
pid=15995 ws=2  monitor=HDMI-A-1  at=[319,142]  size=[1280,820]
```

`monitor` is `HDMI-A-1` (x[-1920..0]) while `at.x` is 996 and 319 — inside
x[0..1600], which is eDP-1. Workspace 2 only displays on HDMI-A-1, so these
windows render nowhere. They are alive and mapped (`hidden: false`), just
unreachable by mouse or by looking at the workspace.

This is the shape that bites in the wild: a laptop suspend/resume or a dock
unplug leaves floating windows holding coordinates for a monitor they are no
longer on. I lost a terminal and an app window this way and only found them by
diffing `hyprctl clients -j` rects against `hyprctl monitors -j` regions.

**One suggestion for the fix:** the useful invariant is not "the window
overlaps *some* monitor" but "the window overlaps **the monitor it is assigned
to**". Both windows above pass the first test and fail the second, so a
recheck that clamps only against the global layout bounds would leave them
stranded. (I hit exactly this writing my own recovery script — the naive
any-monitor check reported everything fine while two windows were invisible.)

Recovery, for anyone landing here from a search:

```bash
# find windows whose `at` falls outside their own `monitor` region
hyprctl clients -j
hyprctl repl 'for _,w in ipairs(hl.get_windows()) do if w.pid==PID then
  hl.dispatch(hl.dsp.focus({window=w})); hl.dispatch(hl.dsp.window.center()) end end'
```

diogochaves:
Another path that still leaves floating windows offscreen on 0.56.2 (`efb5099`): the monitor's **logical size shrinking**, through a mode change or a scale increase. No second monitor involved.

**Repro**
1. Single output at 1920x1080, scale 1.
2. Open a floating window, 875x600, centred (it lands at 523,259).
3. `hl.monitor({ output = "<name>", mode = "1280x720@60", position = "0x0", scale = 1 })`, or raise the scale (Omarchy binds this to SUPER+/ via `omarchy-hyprland-monitor-scaling up`).
4. The window stays at 523,259 with its size unchanged, so it runs past the right and bottom edges of the 1280x720 output.

Tiled windows are rearranged. Every floating window I tried behaves the same, so it isn't app-specific.

PR #10035, opened 2025-04-10
cmake: do not install version.h.in

alba4k:
fixes https://github.com/hyprwm/Hyprland/discussions/10018

Maybe there is a cleaner way to check for `.h` and `.hpp` on a single line? not regex so idk. This way is alright

ready


vaxerski (approved):
this is fine, we don't use `.hh`

PR #10045, opened 2025-04-11
Groupbar auto

Eleocraft:
In stock hyprland when using groups you need to set one keybind to toggle whether or not the current window is a group and a second one to lock the group if you want to add new windows without adding them to the group. When a group with more than one element is toggled to inactive all windows in the group are dispersed into different tiles. I've found that (at least for me) this is never the desired behavior. similarly locking a group with one element is basically useless as you can simply disable the group to archive the same result. Therefore I've made a dispatcher to combine the two actions, essentially running togglegroup if the active window is a) not a group or b) a group with one element. Otherwise if the active window is part of a group with more than one window it locks/unlocks the group.
If I am simply missing something with my workflow feel free to let me know.

This is a farely simple change, all the code basically already exists in either lockActiveGroup or toggleGroup, so I don't expect it to break anything

I haven't touched the hyprctl stuff, as far as i can tell I might need to add some things for the dispatcher to work with the hyprctl command.


vaxerski:
> Therefore I've made a dispatcher to combine the two actions, essentially running togglegroup if the active window is a) not a group or b) a group with one element. Otherwise if the active window is part of a group with more than one window it locks/unlocks the group.

`hyprctl activewindow | grep 'grouped:' | grep -v 'Window' tr -dc ',' | wc -c `
if returns 1 or 0 it's not a group or a group with one element.

Your logic as a one-liner:
```
if [ $(hyprctl activewindow | grep 'grouped:' | grep -v 'Window' | tr -dc ',' | wc -c) -gt "1" ]; then echo "a"; else echo "b"; fi
```

I think this MR is unnecessary.

Eleocraft:
While I still think this is a dispatcher that many people might find useful to exist natively, I won't pursue this further.

small correction for those who want to use the one-liner, you actually want to check if there are 0 commas (and therefore zero or one item in the group list):
```
if [ $(hyprctl activewindow | grep 'grouped:' | grep -v 'Window' | tr -dc ',' | wc -c) -gt "0" ]; then echo "a"; else echo "b"; fi
```
in the config (and with ripgrep) it looks like this:
```
bind=SUPER, T, exec, if [ $(hyprctl activewindow | rg 'grouped:' | rg -v 'Window' | tr -dc ',' | wc -c) -gt '0' ]; then hyprctl dispatch lockactivegroup toggle; else hyprctl dispatch togglegroup; fi
```

PR #10050, opened 2025-04-12
screencopy: Handle explicit sync failure

PlasmaPower:
If addWaiter fails, we should fall back on just triggering the callback immediately. It means explicit sync isn't working, but that's better than screenshare not working entirely.

This should fix the explicit sync related issue in https://github.com/hyprwm/Hyprland/discussions/9953 , though it seems like other people in that discussion have a different nix packaging screenshare issue not related to explicit sync.

I also kept the EGLSync alive in the waiter closure for good measure. I'm not sure that's necessary, but it's a shared pointer anyways so better safe than sorry.

Nope

Ready for merging


vaxerski (approved):
holy if statement lmao

PR #10089, opened 2025-04-16
Popup: damage window on unmap

nnyyxxxx:
https://github.com/user-attachments/assets/ac8f300a-0801-4412-8fb4-cb52ea3fc350

addresses this issue


vaxerski:
@PlasmaPower

nnyyxxxx:
@vaxerski this should just be merged, https://github.com/hyprwm/Hyprland/commit/10035a85cc05f4ee572efb13e62aa1a1cfec6c8b is very problematic as seem in #10079 and in the video linked above

src/protocols/core/Compositor.cpp
```diff
@@ -91,14 +91,14 @@ CWLSurfaceResource::CWLSurfaceResource(SP<CWlSurface> resource_) : resource(reso
...
         }
 
-        if (pending.bufferSize != current.bufferSize) {
-            pending.updated.damage = true;
-            pending.bufferDamage   = CBox{{}, {INT32_MAX, INT32_MAX}};
-        }
+        Vector2D oldBufSize = current.buffer ? current.bufferSize : Vector2D{};
+        Vector2D newBufSize = pending.buffer ? pending.bufferSize : Vector2D{};
+
+        if (oldBufSize != newBufSize || current.buffer != pending.buffer)
```
PlasmaPower, line 100: Why should the surface be damaged if the buffer is changed? That's nowhere in the wayland docs that I've seen, and it means that for any dmabuf application you're pretty much guaranteed to damage the entire surface every frame, because dmabuf requires rotating through buffers used for rendering.

PlasmaPower (changes requested):

src/protocols/core/Compositor.cpp
nnyyxxxx: this shrimply just reverts that commit linked above

PlasmaPower: Ok sure but if you just revert the commit you break dbeaver which that commit fixed

nnyyxxxx: well, its a sacrifice we'll have to make, 2>1

if there is a better way to fix #6844 along with the artifacts then be my guest

PlasmaPower: I don't even have that issue lol, I fixed it for the numerous people in that report. It won't make my hyprland experience any worse if we revert this, but I think the current behavior aligns better with the spec. There might be an issue somewhere else causing the symptoms you see.

nnyyxxxx: no, both issues were bisected to your commit, multiple other people are experiencing artifacts with foot as well

PlasmaPower: Here's what I think the actual issue is: https://github.com/hyprwm/Hyprland/pull/10094

PlasmaPower: Yeah reverting my commit also fixes it by damaging the entire surface every commit, but we should try to fix damage tracking instead of basically disabling it

nnyyxxxx: ill take a look in a sec

PlasmaPower: wdym "when newlining"? The scrollbar just disappears right? Are you seeing artifacting there?

nnyyxxxx: no the scrollbar keeps its position

nnyyxxxx: https://github.com/user-attachments/assets/d55e8155-d435-4966-be63-c197058abe1f

yeah actually, its very clearly still artifacting in this video

nnyyxxxx: like i said, the original commit is problematic; needs to just be reverted

PlasmaPower: I think my PR fixed the position changing issue with the scrollbar but there seems to be some remaining issue with subsurfaces being unmapped that you're experiencing and I haven't replicated yet. I'll try some more to see what it is

nnyyxxxx: well i don't think that mr really changed or fixed anything

artifacting still occurs in both places, other than being more subtle inside of foot

PlasmaPower: Well I'll leave it up to vaxry if he wants to try to make damage work or not then

nnyyxxxx: well i'm not trying to change the maintainers opinion in anyway, i'm just stating the obvious

1: the commit that caused this is very problematic
2: reverting that commit would be best as of right now
3: the additional commit #10094 will probably add additional problems to the table

there was originally 1 issue:

flickering inside of dbeaver

now there are 2:

artifacts with popups, and foots scrollbar

the new commit #10094 now has foots scrollbar artifacting in random places which persists "artifacts with foots scrollbar", and additionally popups are still artifacting

with this commit, both issues with artifacts are gone, with a slight downside of dbeaver flickering again

nnyyxxxx: 1: it fixed 1 issue that was linked against it #6844 
2: right
3: i think that because you are trying to work around the 2 problems, which will inevitably cause more issues

and additionally ill pull down #10094 and see if any issues were resolved

PlasmaPower: I'd consider this PR much more of a workaround by ignoring damage entirely, whereas #10094 actually tries to correctly compute damage

nnyyxxxx: since both issues are fixed, ill just close this then

i still think that #10094 will cause additional problems, anyway

PR #10094, opened 2025-04-16
desktop: Damage subsurface when position changes

PlasmaPower:
Previously, when a subsurface was committed at a new position, its old position wouldn't be damaged. Fixes https://github.com/hyprwm/Hyprland/discussions/10079 for me

Right now this PR just damages the entire parent surface, which is what was done for size as well. This works though isn't optimal for performance, but I think that's fine for now.

Ready for merging

### DIFF
--- a/src/desktop/Subsurface.cpp
+++ b/src/desktop/Subsurface.cpp
@@ -115,21 +115,17 @@ void CSubsurface::onCommit() {
     // I do not think this is correct, but it solves a lot of issues with some apps (e.g. firefox)
     checkSiblingDamage();
 
-    if (m_vLastSize != m_pWLSurface->resource()->current.size) {
-        // TODO: fix this
+    if (m_vLastSize != m_pWLSurface->resource()->current.size || m_vLastPosition != m_pSubsurface->position) {
+        // TODO: only damage the parts of surface that this subsurface was or is covering
         // CBox box{COORDS, m_vLastSize};
         // g_pHyprRenderer->damageBox(box);
         // m_vLastSize = m_pWLSurface->resource()->current.size;
         // box         = {COORDS, m_vLastSize};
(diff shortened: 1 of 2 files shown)


nnyyxxxx:
re: https://github.com/hyprwm/Hyprland/pull/10089#discussion_r2047121412

src/protocols/core/Subcompositor.cpp
```diff
@@ -99,8 +99,10 @@ CWLSubsurfaceResource::~CWLSubsurfaceResource() {
 }
 
 void CWLSubsurfaceResource::destroy() {
-    if (surface && surface->mapped)
+    if (surface && surface->mapped) {
+        surface->events.unmap.emit();
```
vaxerski, line 103: pretty sure unmap() will already do it if the surface is mapped.

PlasmaPower: It won't, the callsites need to. We could change that

PlasmaPower:
@fxzzi I pushed up one additional commit to fix a TODO and only damage where the subsurface was, would you mind testing that as well? If it doesn't work we can revert that commit

src/desktop/Subsurface.cpp
```diff
@@ -115,21 +115,17 @@ void CSubsurface::onCommit() {
     // I do not think this is correct, but it solves a lot of issues with some apps (e.g. firefox)
     checkSiblingDamage();
```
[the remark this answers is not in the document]
vaxerski: damage was broken

vaxerski (approved):

nnyyxxxx:
https://github.com/user-attachments/assets/1662e10c-aad6-4f7c-97f7-8efda639604d

interesting bug

PR #10099, opened 2025-04-17
drm-lease: Add Multi-GPU Support

Joel-Valenciano:
Adds multi-GPU support for the `wp_drm_lease_v1` protocol, and should fix the issue in #10053.

It seems to work fine on my system; I can start and stop monado several times without issues, but I'm not sure if it's fully ready to merge.

Turns `PROTO::lease` into an array of `CDRMLeaseProtocol` pointers, adds a `backend` parameter to the constructor for `CDRMLeaseProtocol`, and adds `WP<CDRMLeaseDeviceResource>` parameters to the constructors for a few resource types.

Resource freeing could be an issue, I had a few crashes with the `onDestroy` callbacks for at least the `CDRMLeaseResource`.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -173,8 +173,13 @@ void CMonitor::onConnect(bool noRule) {
 
     if (m_output->nonDesktop) {
         Debug::log(LOG, "Not configuring non-desktop output");
-        if (PROTO::lease)
-            PROTO::lease->offer(m_self.lock());
+
+        for (auto& [name, lease] : PROTO::lease) {
+            if (!lease || m_output->getBackend() != lease->getBackend())
+                continue;
+
+            lease->offer(m_self.lock());
+        }
 
         return;
     }
--- a/src/managers/ProtocolManager.cpp
+++ b/src/managers/ProtocolManager.cpp
@@ -203,10 +203,14 @@ CProtocolManager::CProtocolManager() {
         if (b->type() != Aquamarine::AQ_BACKEND_DRM)
             continue;
 
-        PROTO::lease = makeUnique<CDRMLeaseProtocol>(&wp_drm_lease_device_v1_interface, 1, "DRMLease");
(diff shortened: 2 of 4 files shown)


vaxerski:
check how wl_output globals are done. You should do something similar to that for each drm node (instead of a std;:string)

Honkazel:
Rebase and use ranges for find

Joel-Valenciano:
My bad, I completely forgot about it. I combined the commits and made a few more changes, like merging the fields of `CDRMLeaseDevice` into `CDRMLeaseProtocol`.

Regarding `std::find`, should I do the same for `std::erase`? (i.e., turning `std::erase_if(...)` into `std::erase(std::ranges::remove_if(...))`, etc.)

Honkazel:
About erase
Clang will choke

src/helpers/Monitor.cpp
```diff
@@ -173,8 +173,12 @@ void CMonitor::onConnect(bool noRule) {
 
     if (m_output->nonDesktop) {
         Debug::log(LOG, "Not configuring non-desktop output");
-        if (PROTO::lease)
-            PROTO::lease->offer(m_self.lock());
+
+        for (auto& [name, lease] : PROTO::lease) {
+            if (lease && m_output->getBackend() == lease->getBackend()) {
```
vaxerski: prefer guards and no {} around short ifs

if (bad)
   continue;
   
do()

vaxerski (commented):
in general looks fine but I cannot test if it works, I dont own any vr equipment or multi gpu systems

src/helpers/Monitor.cpp
Joel-Valenciano: Fixed, I think.

Joel-Valenciano:
You can use normal displays with drm-lease, but it's kind of complicated since you have to override the EDID in the kernel's debugfs, or use a boot option.

Although I guess that doesn't help with the multi-gpu part. My bad.

PR #10102, opened 2025-04-17
managers: allow executing binds not bound to a key

nnyyxxxx:
https://github.com/user-attachments/assets/ba90f9d3-1f4a-47f6-9daa-1cb48b193a0d

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -150,6 +150,10 @@ CKeybindManager::CKeybindManager() {
             if (!m_pLastLongPressKeybind || g_pSeatManager->keyboard.expired())
                 return;
 
+            const auto PACTIVEKEEB = g_pSeatManager->keyboard.lock();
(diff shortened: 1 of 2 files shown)


src/managers/KeybindManager.cpp
```diff
@@ -163,14 +167,17 @@ CKeybindManager::CKeybindManager() {
             if (m_vActiveKeybinds.size() == 0 || g_pSeatManager->keyboard.expired())
                 return;
 
+            const auto PACTIVEKEEB = g_pSeatManager->keyboard.lock();
+            if (!PACTIVEKEEB->allowBinds)
```
vaxerski, line 171: active doesn't really mean much to be honest. What if someone has two keyboards? Some weird PCBs act like that.

nnyyxxxx: i dont think it matters

vaxerski: meh fair

vaxerski (approved):

PR #10111, opened 2025-04-18
render, helpers: Call OpenGL destroyMonitorResources on disconnect

### LINKED ISSUE #10098: Unplugging / replugging a DRM-driven display leaks VRAM
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/9893

## Description

Unplugging/replugging a _real_ DRM display (not wl / headless) makes Hyprland eat more VRAM, which is never freed. Looks about 2x buf each replug which suggests to me a leak in aq, but since it's only drm, it could be the refcounted gbm/gem handles (possibly KMS attachments)

PlasmaPower:
Fixes #10098

I think the monitor resources might be better implemented using e.g. the attachment system from hyprutils, but this fixes it for now.

To prevent leaks we only need to call `destroyMonitorResources` in `~CMonitor`, but also calling it in `onDisconnect` is helpful to free up resources early and it's imo worth deallocating the VRAM temporarily even if the user reenables the monitor later.

Ready for merging

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -51,6 +51,7 @@ CMonitor::CMonitor(SP<Aquamarine::IOutput> output_) : state(this), output(output
 
 CMonitor::~CMonitor() {
     events.destroy.emit();
+    g_pHyprOpenGL->destroyMonitorResources(self);
 }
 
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
damn, what a fail.

PR #10163, opened 2025-04-24
Fix order of urgency flag and event emmission

jmoggr:
When the urgent flag is set, `m_bIsUrgent` was previously updated after emitting the `urgent` event. This meant plugin listeners had to handle an inconsistent state where `m_bIsUrgent` was still false.

This change updates `m_bIsUrgent` before emitting the event, so listeners always observe the correct state.

I was unable to find any other examples where the internal state was modified after event emission. So there is precedent for ordering the update this way.


vaxerski:
clang-format needed

PR #10190, opened 2025-04-27
Update window fullscreen descriptor in hyprctl clients command

Promytheos:
Spent more time than I'm proud to admit trying to figure out what "fullscreenClient" referred to especially cause the documentation only referred to "fullscreenState" for "internal" and "client". Figured it's worth updating the output of the `clients` command to more accurately reflect what is being shown

This: 
20250427_18h19m02s_grim
20250427_18h20m38s_grim

Becomes this:
20250427_18h20m14s_grim
20250427_18h20m23s_grim

There is a possibility that some people are making use of this incorrect output for their automation - so maybe some sort of deprecation message would be useful.

Ready for merge.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -252,8 +252,10 @@ std::string CHyprCtl::getWindowData(PHLWINDOW w, eHyprCtlOutputFormat format) {
     "pid": {},
     "xwayland": {},
     "pinned": {},
-    "fullscreen": {},
-    "fullscreenClient": {},
+    "fullscreenState": {{
+        "internal": {},
+        "client": {},
+    }},
     "grouped": [{}],
     "tags": [{}],
     "swallowing": "0x{:x}",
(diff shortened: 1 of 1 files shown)


vaxerski:
yeah this breaks configs for no reason, really. Why not just improve the docs?

Promytheos:
Fair, but the description of the output is wrong - this scrutiny should have been done when the output of: `(uint8_t)w->m_sFullscreenState.internal, (uint8_t)w->m_sFullscreenState.client` was shown as `fullscreen` and `fullscreenClient`. I don't see how changing docs to say "when `hyprctl clients` says x, read it as y but when you configure, use y". This doesn't make logical sense to me

Promytheos:
Not saying it needs to be done in a patch release. Always grinds me when api changes are done in a patch release, but in a minor release or something this makes sense. I don't mind updating the docs in the meantime to make a note of this. Hell worst case (as an interim solution) have both in the `hyprctl clients` output but show a message warning of this deprecation - this feels like overkill for this situation though. I'm assuming the configs which make use of this are grepping yeah? Maybe to keep that in mind the output could rather be changed to read similar to the window rules configuration for fullscreenstate `fullscreenstate [internal] [client]` e.g. `fullscreenstate 1 2`

Promytheos:
Closing this, tbh it's not worth it, went through the other PRs and looks like you got bigger fish to fry. I'll update docs and _maybe_ if I manage to do some other meaningful contribution in the future I'll see about slipping this in, but eh if it's just preference and docs include the translation from hyprctl clients to window rules - then all good

PR #10213, opened 2025-04-29
DonationNag: ask after each major update

vaxerski:
This changes how the donation nag timing works.

The donation nag will now appear:
- after a major update (e.g. 48 -> 49)\*
- once in late july
- once in december

_however_, a donation nag will never pop up more than once a month. So, if there is an update on the 26th of November, and you get a popup on the 28th, you will _not get one in december_.

This is of course still disableable in your config.

*\*Not directly after. Directly after comes the update screen. The next time you launch Hyprland, the donation popup will appear.*

**Why?**

Living ain't free. Furthermore, the results of the fifth census so far show that users support this approach.


vaxerski:
heell naw

kRHYME7:
I'm on git and can't repro the reported issue.

someone reported in our discord server that the donation nag always pops up across reboots.

vaxerski:
shouldn't be possible? Also, nag or update screen?

PR #10211, opened 2025-04-29
hyprpm: move to system directories for storing plugins

vaxerski:
Right now, we store plugins in `~/.local/share/hyprpm`. This is bad, as anyone with user access can replace the `.so` files, or load a new plugin.

Instead, now, plugins and state are all stored in `/var/cache/hyprpm`. Still per-user, ofc.

While building, we still use a local user directory (just like yay) to avoid having the build scripts ran as sudo.

These commands will now ask you for your sudo password:
 - `add`
 - `remove`
 - `enable`
 - `disable`
 - `update`
 
 New command that also requires sudo:
- `purge-cache` -> removes all the built plugins, repos, and headers

Needs:
- [ ] Testing
- [x] detect and use doas (are there any others people use?)

PR #10234, opened 2025-05-01
hyprpm: unbreak build with libc++

jbeich:
Probably regressed by #10211


jbeich:
Hmm, `<fstream>` is probably unused.

vaxerski (approved):

PR #10246, opened 2025-05-02
hyprpm: Minor optimizations and refactor of helpers and progress bar

nots1dd:
### PR Contents

File 1: Sys.cpp (in `hyprpm/helpers`)

1. Made `SUPERUSER_BINARIES` into `std::array` of `std::string_view` to avoid useless allocs with subsequent changes within the same file.
2. In the function `executableExistsInPath()`, instead of repeatedly concatenating strings, used `std::filesystem::path(p)` operators instead along with some minor changes in the function.
3. Converted `execAndGet` to an `optional` for clear failure indications that accepts `std::string_view` as parameter instead of `std::string` (can convert to `std::string` whenever needed, string view makes sense for just read only input here)
4. New function `shellEscape` that takes in a command to be executed as param:
    a. Escapes shell commands safely to handle special characters (', ;, &, etc.).
    b. Prevents shell injection when wrapping commands in sh -c.

5. In `dropSudo` function, just compare `bin` (previously SB) with the string counterpart without converting to `std::string_view` explicitly
6. New function `fetchSuperuserBins` to print the `Debug::die` to show list of supported super user binaries dynamically instead of hardcoding

File 2: CProgressBar.cpp (in `hyprpm/progress`)

1. Use stringstream instead of instead repeated string concat
2. Added ANSI escape sequence to clear entire line without unnecessary logic previously in `printMessageAbove`
3. Moved logic to find terminal size to new function `getTerminalSize`
5. Have a divide by zero logical check in `print` while checking for percentage done
7. Clamp as a sanity check

This PR is be ready for merge

### DIFF
--- a/hyprpm/src/helpers/Sys.cpp
+++ b/hyprpm/src/helpers/Sys.cpp
@@ -33,15 +33,15 @@ static std::string fetchSuperuserBins() {
 }
 
 static bool executableExistsInPath(const std::string& exe) {
-    const char* pathEnv = std::getenv("PATH");
-    if (!pathEnv)
+    const char* PATHENV = std::getenv("PATH");
+    if (!PATHENV)
         return false;
 
-    CVarList        paths(pathEnv, 0, ':', true);
+    CVarList        paths(PATHENV, 0, ':', true);
     std::error_code ec;
 
-    for (const auto& p : paths) {
-        std::filesystem::path candidate = std::filesystem::path(p) / exe;
+    for (const auto& PATH : paths) {
+        std::filesystem::path candidate = std::filesystem::path(PATH) / exe;
         if (!std::filesystem::exists(candidate, ec) || ec)
             continue;
         if (!std::filesystem::is_regular_file(candidate, ec) || ec)
@@ -100,11 +100,11 @@ bool NSys::isSuperuser() {
 std::string NSys::runAsSuperuser(const std::string& cmd) {
     const std::string escapedCmd = shellEscape(cmd);
 
-    for (const auto& bin : SUPERUSER_BINARIES) {
-        if (!executableExistsInPath(std::string{bin}))
+    for (const auto& BIN : SUPERUSER_BINARIES) {
+        if (!executableExistsInPath(std::string{BIN}))
             continue;
 
-        const auto result = execAndGet(std::string{bin} + " /bin/sh -c " + escapedCmd, true);
+        const auto result = execAndGet(std::string{BIN} + " /bin/sh -c " + escapedCmd, true);
         if (!result.has_value() || result->second != 0)
             Debug::die("Failed to run a command as sudo. This could be due to an invalid password, or a hyprpm bug.");
 
@@ -122,15 +122,15 @@ void NSys::cacheSudo() {
 }
 
 void NSys::dropSudo() {
-    for (const auto& bin : SUPERUSER_BINARIES) {
-        if (!executableExistsInPath(std::string{bin}))
+    for (const auto& BIN : SUPERUSER_BINARIES) {
+        if (!executableExistsInPath(std::string{BIN}))
             continue;
 
-        if (bin == "sudo")
+        if (BIN == "sudo")
             execAndGet("sudo -k");
         else {
             // note the superuser binary that is being dropped
-            std::println("{}", infoString("Don't know how to drop timestamp for '{}', ignoring.", bin));
+            std::println("{}", infoString("Don't know how to drop timestamp for '{}', ignoring.", BIN));
         }
     }
 }
--- a/hyprpm/src/helpers/Sys.cpp
+++ b/hyprpm/src/helpers/Sys.cpp
@@ -128,8 +128,9 @@ void NSys::dropSudo() {
 
         if (bin == "sudo")
             execAndGet("sudo -k");
-        else
+        else {
             // note the superuser binary that is being dropped
             std::println("{}", infoString("Don't know how to drop timestamp for '{}', ignoring.", bin));
+        }
     }
 }
--- a/hyprpm/src/helpers/Sys.cpp
+++ b/hyprpm/src/helpers/Sys.cpp
@@ -69,18 +69,6 @@ static std::optional<std::pair<std::string, int>> execAndGet(std::string_view cm
     return {{proc.stdOut(), proc.exitCode()}};
 }
 
-static std::string shellEscape(const std::string& input) {
-    std::string escaped = "'";
-    for (char c : input) {
-        if (c == '\'')
-            escaped += "'\\''";
-        else
-            escaped += c;
-    }
-    escaped += "'";
-    return escaped;
-}
-
 int NSys::getUID() {
     const auto UID   = getuid();
     const auto PWUID = getpwuid(UID);
@@ -98,13 +86,11 @@ bool NSys::isSuperuser() {
 }
 
 std::string NSys::runAsSuperuser(const std::string& cmd) {
-    const std::string escapedCmd = shellEscape(cmd);
-
     for (const auto& BIN : SUPERUSER_BINARIES) {
         if (!executableExistsInPath(std::string{BIN}))
             continue;
 
-        const auto result = execAndGet(std::string{BIN} + " /bin/sh -c " + escapedCmd, true);
+        const auto result = execAndGet(std::string{BIN} + " /bin/sh -c " + cmd, true);
         if (!result.has_value() || result->second != 0)
             Debug::die("Failed to run a command as sudo. This could be due to an invalid password, or a hyprpm bug.");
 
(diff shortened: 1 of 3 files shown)


nots1dd:
@vaxerski This PR should be ready for merge

hyprpm/src/helpers/Sys.cpp
```diff
@@ -96,15 +122,15 @@ void NSys::cacheSudo() {
...
+        if (bin == "sudo") {
             execAndGet("sudo -k");
-        else
-            std::println("{}", infoString("Don't know how to drop timestamp for {}, ignoring.", SB));
-
-        return;
+        } else {
+            // note the superuser binary that is being dropped
+            std::println("{}", infoString("Don't know how to drop timestamp for '{}', ignoring.", bin));
+        }
```
nnyyxxxx, line 134: no {} around single line if/elses

there are more down below

nots1dd: > no {} around single line if/elses
> 
> there are more down below

Latest commit should fix this

nnyyxxxx: not exactly, you removed braces around elses that had more than 1 line?

hyprpm/src/helpers/Sys.cpp
```diff
@@ -96,15 +122,14 @@ void NSys::cacheSudo() {
...
 
-        if (SB == std::string_view{"sudo"})
+        if (bin == "sudo")
             execAndGet("sudo -k");
         else
-            std::println("{}", infoString("Don't know how to drop timestamp for {}, ignoring.", SB));
-
-        return;
+            // note the superuser binary that is being dropped
+            std::println("{}", infoString("Don't know how to drop timestamp for '{}', ignoring.", bin));
```
nnyyxxxx, line 133: missing brace for this else, one down below too

nots1dd: > missing brace for this else, one down below too

Oh i assumed comments are not really counted, ill change that

hyprpm/src/progress/CProgressBar.cpp
```diff
@@ -1,82 +1,78 @@
...
-    } else
-        message += Colors::RESET;
+    if (i < barWidth) {
+        oss << "╍" << Colors::RESET;
+        ++i;
+        for (; i < barWidth; ++i)
+            oss << "━";
+    } else {
+        oss << Colors::RESET;
+    }
```
nnyyxxxx, line 69: {} here too

hyprpm/src/helpers/Sys.cpp
nnyyxxxx: https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/#other

nots1dd: > https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/#other

My bad, did not check that out just read CoC

nots1dd: This should be resolved now, multi line statements are wrapped in {} and single line statements are not

hyprpm/src/progress/CProgressBar.cpp
nots1dd: This should be resolved now

hyprpm/src/helpers/Sys.cpp
nots1dd: @nnyyxxxx I hope everything is good to go now?

nnyyxxxx: wait for a review from the maintainer, personally i don't really see the need in these changes, the previous codepaths just werked 

but yes this specific review is good since you fixed it in [0683051](https://github.com/hyprwm/Hyprland/pull/10246/commits/06830516645c4de45acd7f8c9b3bba79470adebf) and [30bd8b3](https://github.com/hyprwm/Hyprland/pull/10246/commits/30bd8b3a1a802e40f25ad80bcd24885488d0519c)

nots1dd: > wait for a review from the maintainer, personally i don't really see the need in the changes in the mr, the previous codepaths just werked
> 
> but yes this specific review is good since you fixed it in [0683051](https://github.com/hyprwm/Hyprland/pull/10246/commits/06830516645c4de45acd7f8c9b3bba79470adebf) and [30bd8b3](https://github.com/hyprwm/Hyprland/pull/10246/commits/30bd8b3a1a802e40f25ad80bcd24885488d0519c)

The shell escape function is the main addition really, other than that remaining is just minor code improvements (like the clear line logic was pretty weird, etc.)

hyprpm/src/helpers/Sys.cpp
```diff
@@ -1,59 +1,84 @@
...
-        return {"error", 1};
+        // optional handles nullopt gracefully
+        return std::nullopt;
+
+    return {{proc.stdOut(), proc.exitCode()}};
+}
 
-    return {proc.stdOut(), proc.exitCode()};
+static std::string shellEscape(const std::string& input) {
+    std::string escaped = "'";
```
vaxerski, line 73: use `Hyprutils::String::replaceAllInString`

hyprpm/src/helpers/Sys.cpp
```diff
@@ -69,23 +94,24 @@ int NSys::getEUID() {
...
 bool NSys::isSuperuser() {
-    return getuid() != geteuid() || !geteuid();
+    return getuid() != geteuid() || geteuid() == 0;
 }
 
 std::string NSys::runAsSuperuser(const std::string& cmd) {
-    for (const auto& SB : SUPERUSER_BINARIES) {
-        if (!executableExistsInPath(SB))
-            continue;
+    const std::string escapedCmd = shellEscape(cmd);
```
vaxerski, line 101: this is bad. We should escape _before_. Reason:
```
runAsSuperuser("cd '/somedir/with spaces'")
```
will be incorrectly "sanitized" to `\'/somedir...` which will return a bad path

vaxerski (approved):

hyprpm/src/helpers/Sys.cpp
nots1dd: > this is bad. We should escape _before_. Reason:
> 
> ```
> runAsSuperuser("cd '/somedir/with spaces'")
> ```
> 
> will be incorrectly "sanitized" to `\'/somedir...` which will return a bad path

Am assuming that path is always encapsulated in `'` right?

nots1dd: If so then I can prolly have a fix for this

nots1dd: > this is bad. We should escape _before_. Reason:
> 
> ```
> runAsSuperuser("cd '/somedir/with spaces'")
> ```
> 
> will be incorrectly "sanitized" to `\'/somedir...` which will return a bad path

Is it possible to do something like this:

```cpp
// instead of this
            NSys::runAsSuperuser("mkdir -p -m 755 '/var/cache/hyprpm/'");
 // go with this...
           std::string path = shellEscape("/var/cache/hyprpm/");
           NSys::runAsSuperuser("mkdir -p -m 755 " + path);
  ```
  
  In this case, I believe the issue should be resolved and ensures that malformed paths are handled properly

Comment (vaxerski) [reply hyprpm/src/helpers/Sys.cpp:101]: the quotes are still necessary for if the path has spaces.
You should sanitize the full path, and _not_ sanitize inside runAsSuperuser, then pass the path with singlequotes.

Comment (nots1dd) [reply hyprpm/src/helpers/Sys.cpp:101]: The above example is doing that right? Or am i missing something?

File: hyprpm/src/helpers/Sys.cpp (hunk 4 above)
Comment (nots1dd) [reply hyprpm/src/helpers/Sys.cpp:73]: Ok cool will do

File: hyprpm/src/helpers/Sys.cpp (hunk 5 above)
Comment (nots1dd) [reply hyprpm/src/helpers/Sys.cpp:101]: @vaxerski If I understand this right, this shell escape can look something like this:

```cpp
static std::string shellEscape(const std::string& input) {
    std::string escaped = input;
    Hyprutils::String::replaceInString(escaped, "'", "'\\''");
    return "'" + escaped + "'";
}
```

then before invoking `runAsSuperuser`, we sanitize the path and pass it along with the required command. Anything am missing here?

vaxerski: I'd say just don't sanitize things that are not user input. So sanitization is not done in Sys.cpp, but wherever its needed in pluginmanager.cpp

nots1dd: > I'd say just don't sanitize things that are not user input. So sanitization is not done in Sys.cpp, but wherever its needed in pluginmanager.cpp

Ok I will have to rethink my approach for this, Ill just revert the shell escape changes as it doesn't really pertain to this PR, hope you don't mind

I'll add a new PR regarding the sanitization for user input paths in plugin manager in the future

nots1dd:
@vaxerski I have reverted the changes for shell escape logic, will open a separate PR for it as it doesn't really pertain to the aim of this PR

vaxerski (approved):

PR #10289, opened 2025-05-05
desktop: cleanup code and use std::ranges

nots1dd:

Changes are only made in `src/desktop`:

1. Used and replaced older methods with std::ranges (C++20 >) in `src/desktop`

2. Removed const casting in WindowRule.cpp and instead created a new function resolveRule to call within the constructor initialization to modify m_rule if it is a prop (m_rule stays const)

3. Added designated initializers for SAlphaValue, SBoxExtends, and SHyprIPCEvent

4. In Workspace.cpp, replaced calling spawnWithRules with g_pKeybindManager with CKeybindManager class as it is a static method

5. Used std::max and std::min instead of if stmts in Window.cpp

Not sure if Hyprland or Hyprutils has any custom impls of `std::max` and `std::min`, so just called `std::max` and `std::min` directly. 

Also I noticed the codebase is using a custom macro for PI, not sure why that is needed since `std::numbers::pi` exists (can create a commit for this if needed)

Ready for merge

### DIFF
--- a/src/desktop/LayerRule.cpp
+++ b/src/desktop/LayerRule.cpp
@@ -8,7 +8,7 @@ static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur
 static const auto RULES_PREFIX = std::unordered_set<std::string>{"ignorealpha", "ignorezero", "xray", "animation", "order", "abovelock"};
 
 CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_targetNamespace(ns_), m_rule(rule_) {
-    const bool VALID = RULES.contains(m_rule) || std::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
+    const bool VALID = RULES.contains(m_rule) || std::ranges::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
 
     if (!VALID)
         return;
--- a/src/desktop/LayerRule.cpp
+++ b/src/desktop/LayerRule.cpp
@@ -8,7 +8,7 @@ static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur
 static const auto RULES_PREFIX = std::unordered_set<std::string>{"ignorealpha", "ignorezero", "xray", "animation", "order", "abovelock"};
 
 CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_targetNamespace(ns_), m_rule(rule_) {
-    const bool VALID = RULES.contains(m_rule) || std::ranges::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
+    const bool VALID = RULES.contains(m_rule) || std::ranges::any_of(RULES_PREFIX, [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
 
     if (!VALID)
         return;
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -5,7 +5,6 @@
 #include <any>
 #include <bit>
 #include <string_view>
-#include <algorithm>
 #include "Window.hpp"
 #include "../Compositor.hpp"
 #include "../render/decorations/CHyprDropShadowDecoration.hpp"
@@ -291,7 +290,7 @@ void CWindow::updateWindowDecos() {
     }
 
     for (auto const& wd : decos) {
-        if (std::ranges::find_if(m_windowDecorations.begin(), m_windowDecorations.end(), [wd](const auto& other) { return other.get() == wd; }) == m_windowDecorations.end())
+        if (std::ranges::find_if(m_windowDecorations, [wd](const auto& other) { return other.get() == wd; }) == m_windowDecorations.end())
             continue;
         wd->updateWindow(m_self.lock());
     }
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <hyprutils/animation/AnimatedVariable.hpp>
 #include <re2/re2.h>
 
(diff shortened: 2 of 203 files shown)


nots1dd:
@vaxerski this PR should be good to merge, I had a few other changes in mind, but wanted to clarify first before doing so, I have mentioned it above.

src/desktop/LayerRule.cpp
```diff
@@ -8,7 +8,7 @@ static const auto RULES        = std::unordered_set<std::string>{"noanim", "blur
 static const auto RULES_PREFIX = std::unordered_set<std::string>{"ignorealpha", "ignorezero", "xray", "animation", "order", "abovelock"};
 
 CLayerRule::CLayerRule(const std::string& rule_, const std::string& ns_) : m_targetNamespace(ns_), m_rule(rule_) {
-    const bool VALID = RULES.contains(m_rule) || std::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
+    const bool VALID = RULES.contains(m_rule) || std::ranges::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule_](const auto& prefix) { return rule_.starts_with(prefix); });
```
vaxerski, line 11: if you are using ranges, drop the begin and end, just do `RULES_PREFIX`. Those stl containers are ranges themselves so it's fine

src/desktop/Window.cpp
```diff
@@ -297,7 +291,7 @@ void CWindow::updateWindowDecos() {
     }
 
     for (auto const& wd : decos) {
-        if (std::find_if(m_windowDecorations.begin(), m_windowDecorations.end(), [wd](const auto& other) { return other.get() == wd; }) == m_windowDecorations.end())
+        if (std::ranges::find_if(m_windowDecorations.begin(), m_windowDecorations.end(), [wd](const auto& other) { return other.get() == wd; }) == m_windowDecorations.end())
```
vaxerski, line 294: same ere

src/desktop/Workspace.cpp
```diff
@@ -47,10 +47,11 @@ void CWorkspace::init(PHLWORKSPACE self) {
 
     if (self->m_wasCreatedEmpty)
         if (auto cmd = WORKSPACERULE.onCreatedEmptyRunCmd)
-            g_pKeybindManager->spawnWithRules(*cmd, self);
+            // static member called with class rather than using g_pKeybindManager
```
vaxerski, line 50: useless comment?

src/desktop/WindowRule.cpp
```diff
@@ -13,12 +13,26 @@ static const auto RULES_PREFIX = std::unordered_set<std::string>{
...
+}
+
+CWindowRule::CWindowRule(const std::string& rule, const std::string& value, bool isV2, bool isExecRule) :
+    m_value(value), m_rule(resolveRule(rule)), m_v2(isV2), m_execRule(isExecRule) {
     const auto VALS  = CVarList(rule, 2, ' ');
-    const bool VALID = RULES.contains(rule) || std::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule](auto prefix) { return rule.starts_with(prefix); }) ||
-        (NWindowProperties::boolWindowProperties.find(VALS[0]) != NWindowProperties::boolWindowProperties.end()) ||
-        (NWindowProperties::intWindowProperties.find(VALS[0]) != NWindowProperties::intWindowProperties.end()) ||
-        (NWindowProperties::floatWindowProperties.find(VALS[0]) != NWindowProperties::floatWindowProperties.end());
+    const bool VALID = RULES.contains(rule) || std::ranges::any_of(RULES_PREFIX.begin(), RULES_PREFIX.end(), [&rule](auto prefix) { return rule.starts_with(prefix); }) ||
```
vaxerski, line 33: same ranges comment

src/desktop/WindowRule.cpp
```diff
@@ -82,17 +96,7 @@ CWindowRule::CWindowRule(const std::string& rule, const std::string& value, bool
...
-            NWindowProperties::boolWindowProperties.find(VARS[0]) != NWindowProperties::boolWindowProperties.end() ||
-            NWindowProperties::floatWindowProperties.find(VARS[0]) != NWindowProperties::floatWindowProperties.end()) {
-            *const_cast<std::string*>(&m_rule) = "prop " + rule;
-            m_ruleType                         = RULE_PROP;
-            Debug::log(LOG, "CWindowRule: direct prop rule found, rewritten {} -> {}", rule, m_rule);
-        } else {
-            Debug::log(ERR, "CWindowRule: didn't match a rule that was found valid?!");
-            m_ruleType = RULE_INVALID;
-        }
+        Debug::log(ERR, "CWindowRule: didn't match a rule that was found valid?!");
```
vaxerski, line 99: pretty sure this will break prop rules. Why the extraction to resolveRule? I don't think there is a point.

src/desktop/LayerRule.cpp
nots1dd: Yep got it

src/desktop/Workspace.cpp
nots1dd: Yea i put it there coz previously it was using g_pKbman so thought it would make it clear, ill remove it

src/desktop/WindowRule.cpp
nots1dd: The logic is the same, the only difference is that instead of using const cast for m_rule, i am calling resolve rule first to handle m_rule to ensure that it's promise of being const is kept, since the checks for `m_ruleType` is done with `rule` itself and not `m_rule`, the checks are not affected, and if the rule type is unresolved it just fallbacks to `RULE_INVALID` just like before

src/desktop/LayerRule.cpp
nots1dd: > if you are using ranges, drop the begin and end, just do `RULES_PREFIX`. Those stl containers are ranges themselves so it's fine

This is resolved in the latest commit along with the useless comment

src/desktop/WindowRule.cpp
nots1dd: @vaxerski this one is still pending

vaxerski: yeah because I still dont like the extraction. It's only used once.

nots1dd: > yeah because I still dont like the extraction. It's only used once.

But there is a valid reason for it to be there, const cast breaks the promise of m_rule being const, and the resolveRule function is static and only present in WindowRule.cpp

The logic also is intact, but if it really an issue, am willing to revert this change.

vaxerski: doesn't matter. The const is only for what happens after the constructor.

nots1dd: alright then will revert the changes

nots1dd: @vaxerski changes reverted, this should be resolved now

vaxerski (approved):

PR #10299, opened 2025-05-06
hyprpm: add missing return

Zebra2711:

Add a missing return statement after handling the first superuser binary in the `dropSudo` function

Fixes: 1c530cb

I think @nots1dd accidentally removed the return statement in MR #10246

 yes


nots1dd (approved):
Ah yes, the return is needed coz we only want to deal with the first valid binary found I believe.

Now that I double check, there is some log for `run0` (since I have both sudo and run0 in my system) and it iterates through the loop even after it found sudo. 

The log will print 

```text
Don't know how to drop timestamp for 'run0', ignoring.
```

So yes we do need to return after we find the first valid bin that's on me.

I guess you have to wait for some other maintainer to approve the changes.

vaxerski (approved):

PR #10303, opened 2025-05-06
set fullscreenstate on movetoworkspace

MightyPlaza:
sets fullscreenstate on movetoworkspace

current behavior is keeping the fullscreen state of the workspace and focusing the moved window, which leads to the focused window not being the fullscreen window

a config option to choose which state to restore could be added if wanted


vaxerski:
I think we should just not focus the window

MightyPlaza:
> I think we should just not focus the window

that's what `movetoworkspacesilent` does
`movetoworkspace` should probably focus the window to stay consistent

vaxerski (approved):

PR #10304, opened 2025-05-06
hyprpm: clean up root access and properly check input

vaxerski:
This MR restructures how hyprpm accesses root privileges, limiting how we use the shell (minimizing root shell overall to one instance where it's not really possible) and sanitizes the input. 

Also restricts plugin names to reasonable names only (`A-Za-z0-9\-_`)

### DIFF
--- a/hyprpm/src/core/Manifest.cpp
+++ b/hyprpm/src/core/Manifest.cpp
@@ -1,6 +1,12 @@
 #include "Manifest.hpp"
 #include <toml++/toml.hpp>
-#include <iostream>
+#include <algorithm>
+
+// Alphanumerics and -_ allowed for plugin names. No magic names.
+// [A-Za-z0-9\-_]*
+static bool validManifestName(const std::string_view& n) {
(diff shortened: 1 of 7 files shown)


hyprpm/src/core/Manifest.cpp
```diff
@@ -1,6 +1,12 @@
 #include "Manifest.hpp"
 #include <toml++/toml.hpp>
-#include <iostream>
+#include <algorithm>
+
+// Alphanumerics and -_ allowed for plugin names. No magic names.
+// [A-Za-z0-9\-_]*
+static bool validManifestName(const std::string_view& n) {
+    return std::ranges::all_of(n, [](const char& c) { return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '-' || c == '_' || c == '=' || (c >= '0' && c <= '9'); });
```
SoSeDiK, line 8: Is `c == '='` intended? Not mentioned as allowed char in comment / MR description
@vaxerski

vaxerski: yes, documentation bug I guess. `=` is completely fine.

PR #10319, opened 2025-05-07
renderer: precompute fullalpha

gulafaran:
precompute it once instead of calculating it every single call, was showing up as a lot of time wasted in profiling.


vaxerski (approved):
huge

PR #10356, opened 2025-05-09
renderer: render blur on fade out

vaxerski:
This MR makes it so that the blur of a window / ls is still there when fading out, instead of disappearing.

Needs testing.

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1156,7 +1156,7 @@ bool CWindow::opaque() {
     if (m_wlSurface->small() && !m_wlSurface->m_fillIgnoreSmall)
         return false;
 
-    if (PWORKSPACE->m_alpha->value() != 1.f)
+    if (PWORKSPACE && PWORKSPACE->m_alpha->value() != 1.f)
         return false;
 
     if (m_isX11 && m_xwaylandSurface && m_xwaylandSurface->m_surface && m_xwaylandSurface->m_surface->m_current.texture)
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -751,7 +751,8 @@ void Events::listener_unmapWindow(void* owner, void* data) {
         g_pCompositor->setWindowFullscreenInternal(PWINDOW, FSMODE_NONE);
 
     // Allow the renderer to catch the last frame.
-    g_pHyprRenderer->makeWindowSnapshot(PWINDOW);
+    if (g_pHyprRenderer->shouldRenderWindow(PWINDOW))
+        g_pHyprRenderer->makeWindowSnapshot(PWINDOW);
 
     // swallowing
     if (valid(PWINDOW->m_swallowed)) {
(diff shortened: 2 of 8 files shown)


DashieTM:
Seems to work as intended, cool stuff :)

aphelei:
it works it looks weird. The blur is "shrink" and doesn't matches the wallpaper behind

It's clearly visible with very slow animations. 

https://github.com/user-attachments/assets/83cf95c7-f8e8-4215-b93b-9662931f31f9

It's still a little noticeable with quick animations 

https://github.com/user-attachments/assets/895e833a-b565-4f5b-9232-872c60444c89


will retry without the patch. But frankly it's ok I think

aphelei:
It looks better without the patch, though that's just what I'm accustomed to.

It would be better if the window blurs dynamically the background and doesn't "freeze" the last state.

vaxerski:
> It would be better if the window blurs dynamically the background and doesn't "freeze" the last state.

the problem is that this would be way more resource intensive and way less straightforward to code in

vaxerski:
check now @aphelei this should solve it. Code is a bit messy but it's just for testing atm

aphelei:
> check now @aphelei this should solve it. Code is a bit messy but it's just for testing atm

it works flawlessly

> > It would be better if the window blurs dynamically the background and doesn't "freeze" the last state.
> 
> the problem is that this would be way more resource intensive and way less straightforward to code in

Were you able to work around the performance impact ?

vaxerski:
kinda? it will now have the same impact as regular blur instead of no additional impact

vaxerski:
cleaned up the code, @aphelei care to verify it's still good?

aphelei:
all good

vaxerski:
thanks!

VirtCode:
Just found that this pr has regressed the fade-out of layer surfaces on transformed outputs. While the blur seems to work and be transformed correctly, the surface texture seems to be rendered with an incorrect transform. Here's a quick video (~ignore the cursor, seems to happen on screencap with software cursors apparently~):

https://github.com/user-attachments/assets/672e0ede-027e-4d5b-9057-c933eebe1d08

Sorry I didn't get to test this earlier, but here we are.

vaxerski:
as usual weirdos who rotate monitors...

VirtCode:
sorry it's been a bit longer now but: @vaxerski 
(but no need to hurry, it's purely cosmetic and doesn't bother me too much)

VirtCode:
pretty sure that it was this pr that regressed it, think I have tested the commit before it and it was fine last week. but I'll try to test/bisect again when I get home.

VirtCode:
well can confirm that this pr is indeed the regressor. 

Yes, I also don't really see any obvious changes to the transforms but if you replace the `renderTextureWithBlur` in the tex pass with the normal `renderTextureInternalWithDamage` again (so pretty much like it was before), the transforms are right again. So probably the `renderTextureWithBlur` method struggles with textures from layers somehow? I tried a couple of things out and it looks like if I do this, [hl-blur-transforms-something-idk.txt](https://github.com/user-attachments/files/20266367/hl-blur-transforms-something-idk.txt), the fade is transformed correctly again, but the blur texture is no longer transformed when the animation is done???? Don't really know what I am talking about here, so you might aswell just ignore this, but maybe it gives you some clues

PR #10364, opened 2025-05-10
opengl: optimize shaders and reduce unneeded drawcalls

gulafaran:
reduce the amount of glFlush that shouldnt technically be needed at all in these places.

ensure glUseProgram isnt calling same program twice or more "small overhead but everything counts" can be seen happening in apitrace.
 
use a more modern vertex array object approach with the shaders, makes it a onetime setup on shader creation instead of once per drawcall, also should make the driver not have to revalidate the vertex format on each render call.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -977,6 +977,7 @@ bool CHyprOpenGLImpl::initShaders() {
                 shaders->m_shCM.applyTint         = glGetUniformLocation(prog, "applyTint");
                 shaders->m_shCM.tint              = glGetUniformLocation(prog, "tint");
                 shaders->m_shCM.useAlphaMatte     = glGetUniformLocation(prog, "useAlphaMatte");
+                shaders->m_shCM.createVao();
             } else
                 Debug::log(ERR,
                            "WARNING: CM Shader failed compiling, color management will not work. It's likely because your GPU is an old piece of garbage, don't file bug reports "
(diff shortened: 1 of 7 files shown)


vaxerski:
on my laptop seems to work nicely and reduce power usage

fxzzi:
Currently testing on my PC, everything looking great :) nothing out of the ordinary

Would test on my laptop too but there's some stupid hydra build failures (NixOS shenanigans) and its making me build a lot of stuff locally which my laptop would take hours doing

Arisa-Snowbell:
Tested on my computer, seems to be running fine, didn't notice anything weird.
But also didn't notice big difference in performance, seems about same but clearly using VAO is nicer. I am sure my laptop would welcome these changes way more and would show bigger difference.

src/render/OpenGL.cpp
```diff
@@ -2859,6 +2824,39 @@ void CHyprOpenGLImpl::initMissingAssetTexture() {
...
 
+void CHyprOpenGLImpl::useProgram(GLuint prog) {
+    if (m_currentProgram == prog)
+        return;
+
+    glUseProgram(prog);
+    m_currentProgram = prog;
+}
+
+void CHyprOpenGLImpl::createShaderVao(SShader& shader) {
```
vaxerski, line 2835: this should be moved to SShader as a member

gulafaran: moved to SShader.

PR #10402, opened 2025-05-13
input: disallow virtual keyboards from changing LED state

### LINKED ISSUE #10183: Keyboard LED state is reset on window focus change
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10162

## Description
When window focus moves to another `wl_client`, keyboard's LED state is reset to 0 until a keeb input occurs.

vaxerski:
Virtual keyboards should not be allowed to update LED state.

Thanks @Kam1k4dze for finding, can you verify this works?

Fixes #10183


Kam1k4dze:
It seems to be fixed for me. Thanks for the fix! ❤️

PR #10408, opened 2025-05-14
core: drop the legacy renderer

gulafaran:
the legacy renderer is broken and barely used, drop it. even asahi has 3.x support nowadays.
then we can also focus more on the new fancy gles3 features. that actually improves performance


gulafaran:
probably needs wiki PR, nix PR too. and breaking changes listing. if breaking broken things is considered as such.

PR #10416, opened 2025-05-14
input: do not send mouse events when outside of a surface

outfoxxed:
Stops mouse events from being sent to surfaces that aren't under the mouse (except when dragging).

Tested normal mouse movement, dragging, and multi client d&d, all of which work.

Ready to merge

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -201,6 +201,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus, bool mouse) {
     Vector2D               surfacePos = Vector2D(-1337, -1337);
     PHLWINDOW              pFoundWindow;
     PHLLS                  pFoundLayerSurface;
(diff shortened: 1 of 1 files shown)


vaxerski:
why? This will have unwanted side effects in many apps.

outfoxxed:
Because I've been seeing side effects with events sent outside the window, namely buttons that shouldnt be clickable being clickable from outside of the window since the toolkit expects the compositor to clip.

What side effects are there from not sending the events? I tested all the easy cases so I assume you have something specific in mind.

vaxerski:
a bunch of apps (notably x11) will trigger popups on hover when moving the mouse out if you dont send the events

outfoxxed:
Would always sending if x11 be reasonable, or do you know of any wayland clients that break?

outfoxxed:
Ignored for x11, though i wasn't able to find any cases where x11 wasn't already ignored by the mouse not being in the window.

vaxerski (approved):

Issue #9970, opened 2025-04-07
Screencopy / Toplevel Export with shm does not transform image data properly


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/9910

## Description
When copying via shm, the window / screen's contents are not transformed properly to be upright.

vaxerski:
@WhySoBad can you still repro this? I can't

WhySoBad:
Yes, I can still reproduce it on hyprland-git `v0.48.0-110-gf5c5cfa96` (I was unable to start `v0.48.0-118` as it just hung up permanently)

vaxerski:
how? grim works for me

WhySoBad:
Ah nevermind, this protocol is not even implemented by hyprland. They use the transform supplied by the wl_output here https://gitlab.freedesktop.org/emersion/grim/-/blob/master/main.c#L237

vaxerski:
[patch.txt](https://github.com/user-attachments/files/20006594/patch.txt)
can  you try this

WhySoBad:
> [patch.txt](https://github.com/user-attachments/files/20006594/patch.txt) can you try this

With this patch it's exactly the same as without it. 

![Image](https://github.com/user-attachments/assets/2ff9c854-2ccf-4d68-ae04-01eea61f76ab)

vaxerski:
can you give some repro steps?

vaxerski:
I think this might be correct, since everything else works?

WhySoBad:
What do you mean? That the transforms are not applied on the server?

vaxerski:
yes, essentially

vaxerski:
feel free to gather that list then. With toplevel export I agree.

WhySoBad:
After a bit more thinking about it, I think we can save collecting the list. 

The output export also needs to be transformed since (sandboxed) applications which use the pipewire interface for screen capture, only have limited access to the wloutput object and can therefore not know about the transformation of the output.

PR #10482, opened 2025-05-20
Add noscreenshare windowrule

Kam1k4dze:
I don't think we can do any better without fully rerendering.

### DIFF
--- a/src/protocols/Screencopy.cpp
+++ b/src/protocols/Screencopy.cpp
@@ -200,6 +200,17 @@ void CScreencopyFrame::renderMon() {
     g_pHyprOpenGL->renderTexture(TEXTURE, monbox, 1);
     g_pHyprOpenGL->setRenderModifEnabled(true);
     g_pHyprOpenGL->setMonitorTransformEnabled(false);
+    for (auto const& w : g_pCompositor->m_windows) {
+        if (w->m_windowData.noScreenShare.valueOrDefault()) {
+            const auto PWORKSPACE = w->m_workspace;
+            if (!PWORKSPACE) {
+                continue;
+            }
+            const auto REALPOS          = w->m_realPosition->value() + (w->m_pinned ? Vector2D{} : PWORKSPACE->m_renderOffset->value());
+            const auto noScreenShareBox = CBox{REALPOS.x, REALPOS.y, std::max(w->m_realSize->value().x, 5.0), std::max(w->m_realSize->value().y, 5.0)}.scale(m_monitor->m_scale);
+            g_pHyprOpenGL->renderRect(noScreenShareBox, {0, 0, 0, 255});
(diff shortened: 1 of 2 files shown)


src/protocols/Screencopy.cpp
```diff
@@ -200,6 +200,20 @@ void CScreencopyFrame::renderMon() {
     g_pHyprOpenGL->renderTexture(TEXTURE, monbox, 1);
     g_pHyprOpenGL->setRenderModifEnabled(true);
     g_pHyprOpenGL->setMonitorTransformEnabled(false);
+    for (auto const& w : g_pCompositor->m_windows) {
```
vaxerski, line 203: needs some work:
Firstly, this will draw over other windows if there are any above the "private" window.
Secondly, it doesnt check the current workspace... will render even if the window is on a different workspace.
Lastly, doesn't take rounding into account.

Find windows above the window being "private'd" (rudimentary is okay)

An idea like this isn't even that bad:
- Every time a frame is rendered, send weak window pointers to the screencopy protocol with: Window, geometry.
- When you find one that is private when sharing, take the geometries that come _later_ and use them as a mask if they occlude our private window.

vaxerski: :thinking: you're right. At least fix the workspace and rounding parts then.

Kam1k4dze: Done, also added rendering of rects over popups

Kam1k4dze: Also, Screencopy of the scaled and transformed monitor is currently broken, not sure if this is related to https://github.com/hyprwm/Hyprland/issues/9970

Kam1k4dze: Also, Screencopy of the scaled and transformed monitor is currently broken, not sure if this is related to #9970

Kam1k4dze:
Waiting for review...

nnyyxxxx:
u gotta ping the maintainer he forgot

Kam1k4dze:
@vaxerski 😭

Kam1k4dze:
wiki mr https://github.com/hyprwm/hyprland-wiki/pull/1086

Kam1k4dze:
No, it is not possible https://github.com/hyprwm/Hyprland/pull/10482#discussion_r2098019508

vaxerski:
I've added some breathing room, code was very claustrophobic.

This looks good. We just need to also reject toplevel export. Probably just render black. Screencopy part lgtm.

Kam1k4dze:
I think we can just not call renderWindow.

Kam1k4dze:
> it is, but it's less trivial. We can check for `w->opaque()`

I don't know how to handle rounding correctly in that case.

PR #10502, opened 2025-05-21
hyprpm: ignore pins when adding a package with a git rev

dumbasPL:
fixes #10436

makes `hyprpm add` do what it claims in the help section. This is already handled correctly in `hyprpm update`, but seems to have been missed here.

It will always be Hyp**E**rland in my heart.

ready


vaxerski (approved):
thanks!

PR #10594, opened 2025-05-30
refactor: replace all `typedef` with `using`

Kam1k4dze:


ReStranger:
First, correct Ci in your past PR

vaxerski:
> First, correct Ci in your past PR

No need to be aggressive

vaxerski:
what's the benefit of `using` there?

Kam1k4dze:
> what's the benefit of `using` there?

Consistency across the codebase. It's a more C++-ish way of doing things. `using` is just a superset of `typedef` with template support. Also, some people say that it has better syntax.

PR #10591, opened 2025-05-30
Revert "refactor: use std::ranges whenever possible"

ReStranger:
Reverts hyprwm/Hyprland#10584

What was it merged?  It didn't even pass CI

Either roll back  [this](https://github.com/hyprwm/Hyprland/commit/9190443d951c21ab24201d033accf90b4d172f46#diff-4e1d4134b02a69e00c2481018960bca9e1f5c71bac62fe285b4f30fce402530aR49) change, or correct its work

PR #10595, opened 2025-05-30
nix: use gcc15

mitsuruu:
As an alternative to the revert PR (#10591), instead we can replace the use of `gcc14Stdenv` with `gcc15Stdenv` to use the features required to compile 9190443

The `nixpkgs` input needed to be updated as gcc15 was not available.

Yes


fufexan (approved):
Great, thanks!

PR #10606, opened 2025-06-01
Fix HyprError not displaying at startup

Kam1k4dze:
Fixes `HyprError` not showing at startup - it's currently destroyed right after we load the config and queue the errors
Idk why it was even there in the first place.
Was added in https://github.com/hyprwm/Hyprland/pull/9600


vaxerski (approved):
lgtm

PR #10628, opened 2025-06-02
screencopy: fix applying `noscreenshare` to invisible special workspaces

XPhyro:

The `noscreenshare` window rule is applied even if a window resides in a special workspace, and is hidden from view. This PR adds a simple check to ensure the screen is not blocked if the window is hidden, accounting for animation transparency.

This is just one scenario I encountered myself. Apart from this, and what's already mentioned in #10482, there might be other missing cases.

I'm also not well-versed in the Hyprland codebase, or otherwise a Hyprland expert, so I might be missing some catch with this approach.

Should be ready.

### DIFF
--- a/src/protocols/Screencopy.cpp
+++ b/src/protocols/Screencopy.cpp
@@ -214,9 +214,6 @@ void CScreencopyFrame::renderMon() {
         if UNLIKELY (!PWORKSPACE)
             continue;
 
-        if UNLIKELY (PWORKSPACE->m_isSpecialWorkspace && !PWORKSPACE->m_visible && !PWORKSPACE->m_alpha->isBeingAnimated())
-            continue;
-
         const auto REALPOS          = w->m_realPosition->value() + (w->m_pinned ? Vector2D{} : PWORKSPACE->m_renderOffset->value());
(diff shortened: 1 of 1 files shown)


Kam1k4dze:
Interesting. We are already checking `shouldRenderWindow`, which should be sufficient (though it may not be), and I don't see any difference with or without this MR. Could you please demonstrate when this occurs?

XPhyro:
Perhaps you're testing regular workspaces, and not special ones? The issue only occurs when prohibited windows reside in special workspaces. I attached two videos below to demonstrate it.

1. This one shows the behaviour without the provided patch. It correctly catches when the window is in a different workspace, but fails to catch when it is in a special workspace, and continues to render the black overlay even when the window is not visible.

https://github.com/user-attachments/assets/3e0949bf-1ba0-44f2-a3a4-fb9484e214a6

2. This one shows the behaviour with the provided patch. It stops rendering the black overlay when the special workspace is no longer in view, unlike the original behaviour

https://github.com/user-attachments/assets/5ae001ff-6b6b-4dd8-8507-59f8f3d1bc55

Kam1k4dze:
Yeh, now I understand. I think this fix would be better:
[patch.txt](https://github.com/user-attachments/files/20582450/patch.txt)

XPhyro:
Yup. Definitely seems better.

Still tested it for completeness, and this too yields the fixed results shown above.

Thanks for the hand :)

PR #10631, opened 2025-06-03
input: support configuring drag_3fg from libinput

UfukUstali:
closes #10586

i did consider using a Choice instead of Range
```cpp
    SConfigOptionDescription{
        .value       = "input:touchpad:drag_3fg",
        .description = "Three Finger Drag",
        .data        = SConfigOptionDescription::SChoiceData{0, "disabled,3,4"},
    },
```

but it ends up picking the 3 or 4 as the return value and not the index of 3 or 4 when used with `g_pConfigManager->getDeviceInt(devname, "drag_3fg", "input:touchpad:drag_3fg")`.
i guess, not sure about this, `CConfigValue` does that but i didn't see a usage of it for per device options.

it is most likely ready


src/managers/input/InputManager.cpp
```diff
@@ -1252,6 +1252,11 @@ void CInputManager::setPointerConfigs() {
                     libinput_device_config_scroll_set_natural_scroll_enabled(LIBINPUTDEV, g_pConfigManager->getDeviceInt(devname, "natural_scroll", "input:natural_scroll"));
             }
 
+            if (libinput_device_config_3fg_drag_get_finger_count(LIBINPUTDEV) >= 3) {
+                const auto DRAG_3FG_STATE = static_cast<enum libinput_config_3fg_drag_state>(g_pConfigManager->getDeviceInt(devname, "drag_3fg", "input:touchpad:drag_3fg"));
```
vaxerski: enum keyword can be dropped, this is c++

vaxerski (commented):
wiki mr needde

src/config/ConfigManager.cpp
```diff
@@ -654,6 +654,7 @@ CConfigManager::CConfigManager() {
     registerConfigVar("input:touchpad:scroll_factor", {1.f});
     registerConfigVar("input:touchpad:flip_x", Hyprlang::INT{0});
     registerConfigVar("input:touchpad:flip_y", Hyprlang::INT{0});
+    registerConfigVar("input:touchpad:drag_3fg", Hyprlang::INT{0});
```
vaxerski, line 657: I think you missed adding this var to per device cfgs as well

cpiber:
The functions used in this PR are only available from libinput 1.28.0, CMakeLists.txt and meson.build should be updated to reflect that

PR #10569, opened 2025-05-28
Refactored functions in XWM.cpp

JassonCordones:
#### Describe your PR, what does it fix/add?
*SXSelection::onSelection*
- Made the function more readable and less redundant
- Extracted repeated conditions into booleans.
- Reduced nested conditionals
- Reused (conn) pointer

*CXWM::readProp*
- Put properties into named lambda
- Reduced repetition

*CXWM::initSelection*
- Put masks in constants 
- Put window creation into lambda
- Cleaned duplicated xcb_create_window logic
- Better variable names

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
No

#### Is it ready for merging, or does it need work?

Ready, I might refactor a lot of nasty functions in this file when I find some spare time.

PR #10646, opened 2025-06-04
fix xwayland -> wayland clipboard

JassonCordones:
Fix regression introduced by #10569

probably needs more testing

yes

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -1214,6 +1214,7 @@ void CXWM::initSelection() {
     };
 
     createSelectionWindow(m_clipboard.window, "CLIPBOARD_MANAGER");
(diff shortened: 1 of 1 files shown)


Zebra2711:
> Fix regression introduced by #10569

Damn, so that is because u

> probably needs more testing

~That still not fixed, though~

~edit: Forgot to reboot. It works now~

Sometimes it worked, sometimes it didn't, also making it crash and hang when pasting

Zebra2711 (approved):

JassonCordones:
The crash and hang when pasting was documented before I made that commit in #10332 for example.

MohammedGoder:
focus-editor seems to be crashing when copying text to the clipboard since this update.

It was working before hyprland got his update.

I'm assuming that this commit has something to do with it.

PR #10650, opened 2025-06-04
Improve ultrawide support for dwindle layout

vlofgren:

The pull request adds the ability to specify an aspect ratio for the dwindle layout, to be used when only a single window exists.  

This is necessary on ultrawide monitors where you basically never want to display an application in fullscreen (except some games and movies).  Master layout works well on ultrawide, thanks to #1642, but dwindle works to poorly it's nearly unusuable in its default functionality unless you pad the main window with idle terminals or something.

To illustrate the problem, here is the ultrawide fullscreen experience, note 10% of the UI buttons are half a kilometer to the right from the rest of the UI, and you physically need to turn your head to see them.

image

The code adds two new configuration options

    dwindle:single_window_aspect_ratio

Defaults to [0,0]; specifies the aspect ratio to coerce single windows into when a workspace only shows one window.  If the y-axis is zero, then the code doesn't run.

    dwindle:single_window_aspect_ratio_tolerance

Default 0.1.  If the added padding on one side is smaller than this value *  the width or height of the screen, no padding is added.  This is used to avoid introducing undesired padding when using waybar and similar widgets that push the windows in a bit.  

Demo of the changed logic below, coaxing a 16:9 screen into showing a 4:3 window:

43_screenshot

I've added the logic in two places, both onWindowCreatedTiling and calculateWorkspace.  I'm not 100% sure the first pass of logic is necessary, but from reading the code it looks like it might be even if I haven't manage to trigger any bugs when it was removed. 

It may also be desirable to be able to configure this on a per-monitor basis, as the current fix only helps if you have one regular monitor and one ultrawide, not if you have a horizontal monitor and a vertical ultrawide monitor.  Though I don't know if this is a stopper for merging the change.

It's also unclear if single_window_aspect_ratio_tolerance needs to be configurable.  Setting it to something like 0.1 does seem good enough in most use cases.  

I'm daily driving this code on my computer and not seeing any issues.  The design decisions above should probably be discussed though.

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1,8 +1,6 @@
 #pragma once
 
 #include "ConfigManager.hpp"
-#include "ConfigManager.hpp"
-#include "ConfigManager.hpp"
 
 inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
 
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1,6 +1,8 @@
 #pragma once
 
 #include "ConfigManager.hpp"
+#include "ConfigManager.hpp"
+#include "ConfigManager.hpp"
 
 inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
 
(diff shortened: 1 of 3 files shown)


src/config/ConfigDescriptions.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "ConfigManager.hpp"
+#include "ConfigManager.hpp"
 #include "ConfigManager.hpp"
```
Honkazel: Don't do that 3 times

vlofgren: Ah wtf.  Fixed.

Honkazel:
@vaxerski

vaxerski:
hello I am vaxry

src/layout/DwindleLayout.cpp
```diff
@@ -295,7 +295,32 @@ void CHyprDwindleLayout::onWindowCreatedTiling(PHLWINDOW pWindow, eDirection dir
 
     // if it's the first, it's easy. Make it fullscreen.
     if (!OPENINGON || OPENINGON->pWindow.lock() == pWindow) {
-        PNODE->box = CBox{PMONITOR->m_position + PMONITOR->m_reservedTopLeft, PMONITOR->m_size - PMONITOR->m_reservedTopLeft - PMONITOR->m_reservedBottomRight};
+        const auto REQUESTED_RATIO           = *CConfigValue<Hyprlang::VEC2>("dwindle:single_window_aspect_ratio");
```
vaxerski: I think this logic belongs to `applyNodeDataToWindow`.

vlofgren: That makes sense.  I've moved the logic to be next to the gaps calculation in that function and re-tested it to make sure it works.

vaxerski (approved):
very nice, wiki mr needed

PR #10667, opened 2025-06-07
Add config options to color inactive and locked groupbar titles

jirassimok:

This adds a set of new config options (`group:groupbar:text_color_inactive`, `text_color_locked_active`, and `text_color_locked_inactive`), that set the text color for inactive and/or locked windows' groupbar titles, as requested in #7764, #9747, and #10647 (well, the requests are for inactive windows; it was my idea to add separate options for locked windows).

The new variables default back to the old one, so nobody should notice a difference.

[The first two commits](https://github.com/jirassimok/Hyprland/compare/2794f485...d3114c3a) just add the inactive-window option, and I think they shouldn't require any changes. The [other three commits](https://github.com/jirassimok/Hyprland/compare/d3114c3a...groupbar-text-color) add the options for locked bars, and 6bc448fa might need to be removed or replaced if preprocessor macros are frowned upon here.

The first two commits are ready to merge, and if the rest are suitable well-styled, the whole thing is ready.


src/config/ConfigDescriptions.hpp
```diff
@@ -1014,10 +1014,28 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
...
-        .description = "controls the group bar text color",
+        .description = "color for window titles in the groupbar",
         .type        = CONFIG_OPTION_COLOR,
         .data        = SConfigOptionDescription::SColorData{0xffffffff},
     },
+    SConfigOptionDescription{
+        .value       = "group:groupbar:text_color_inactive",
+        .description = "color for inactive windows' titles in the groupbar (if unset, defaults to text_color)",
+        .type        = CONFIG_OPTION_COLOR,
+        .data        = SConfigOptionDescription::SColorData{}, //TODO: UNSET?
```
vaxerski, line 1025: set to int64_max I guess

jirassimok: I copied this from `shadow:color_inactive`, which was the only other color that falls back to another value.

Actually, since this is a signed integer, maybe it would make more sense for the "unset" value to be -1, which might be more obvious in the output of `hyprctl descriptions`, which is the only place this actually seems to be used (internally, at least).

vaxerski: ah right, yeah, -1 makes sense.

jirassimok: I've added a commit that makes the fallback value for color variables -1 (both the new ones in this PR and `shadow:color_inactive`).

vaxerski (approved):

vaxerski:
wiki mrn eeded

jirassimok:
Wiki PR made at hyprwm/hyprland-wiki#1117, though the more I think about it, the less sure I am that the locked-inactive color should default to the inactive color rather than the locked color. But I guess it doesn't really matter when it's just one more config variable either way.

PR #10702, opened 2025-06-10
add sticky option for drag_lock

m4rch3n1ng:

libinput 1.27 (released nov 2024) comes with a new option for the `drag_lock` config: `ENABLED_STICKY`. this pr makes it so you can actually set that option in hyprland.

the official release docs (https://lists.freedesktop.org/archives/wayland-devel/2024-November/043860.html) recommend changing the default to `sticky`, so i did just that, but in a separate commit, so i can very easily revert it if you do not want to change the default.

i tested it locally and it worked fine for me, this should be ready.

hyprland-wiki pr ~~will be up in a second~~ up at https://github.com/hyprwm/hyprland-wiki/pull/1114.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -652,7 +652,7 @@ CConfigManager::CConfigManager() {
     registerConfigVar("input:touchpad:middle_button_emulation", Hyprlang::INT{0});
     registerConfigVar("input:touchpad:tap-to-click", Hyprlang::INT{1});
     registerConfigVar("input:touchpad:tap-and-drag", Hyprlang::INT{1});
-    registerConfigVar("input:touchpad:drag_lock", Hyprlang::INT{0});
+    registerConfigVar("input:touchpad:drag_lock", Hyprlang::INT{2});
     registerConfigVar("input:touchpad:scroll_factor", {1.f});
     registerConfigVar("input:touchpad:flip_x", Hyprlang::INT{0});
     registerConfigVar("input:touchpad:flip_y", Hyprlang::INT{0});
@@ -775,7 +775,7 @@ CConfigManager::CConfigManager() {
     m_config->addSpecialConfigValue("device", "middle_button_emulation", Hyprlang::INT{0});
     m_config->addSpecialConfigValue("device", "tap-to-click", Hyprlang::INT{1});
     m_config->addSpecialConfigValue("device", "tap-and-drag", Hyprlang::INT{1});
-    m_config->addSpecialConfigValue("device", "drag_lock", Hyprlang::INT{0});
+    m_config->addSpecialConfigValue("device", "drag_lock", Hyprlang::INT{2});
     m_config->addSpecialConfigValue("device", "left_handed", Hyprlang::INT{0});
     m_config->addSpecialConfigValue("device", "scroll_method", {STRVAL_EMPTY});
     m_config->addSpecialConfigValue("device", "scroll_button", Hyprlang::INT{0});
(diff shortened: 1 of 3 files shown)


src/config/ConfigManager.cpp
```diff
@@ -652,7 +652,7 @@ CConfigManager::CConfigManager() {
     registerConfigVar("input:touchpad:middle_button_emulation", Hyprlang::INT{0});
     registerConfigVar("input:touchpad:tap-to-click", Hyprlang::INT{1});
     registerConfigVar("input:touchpad:tap-and-drag", Hyprlang::INT{1});
-    registerConfigVar("input:touchpad:drag_lock", Hyprlang::INT{0});
+    registerConfigVar("input:touchpad:drag_lock", Hyprlang::INT{2});
```
vaxerski, line 655: why is it enabled by defaul

src/managers/input/InputManager.cpp
```diff
@@ -1233,10 +1233,8 @@ void CInputManager::setPointerConfigs() {
             else
                 libinput_device_config_tap_set_drag_enabled(LIBINPUTDEV, LIBINPUT_CONFIG_DRAG_ENABLED);
 
-            if (g_pConfigManager->getDeviceInt(devname, "drag_lock", "input:touchpad:drag_lock") == 0)
-                libinput_device_config_tap_set_drag_lock_enabled(LIBINPUTDEV, LIBINPUT_CONFIG_DRAG_LOCK_DISABLED);
-            else
-                libinput_device_config_tap_set_drag_lock_enabled(LIBINPUTDEV, LIBINPUT_CONFIG_DRAG_LOCK_ENABLED);
+            const auto TAP_DRAG_LOCK = static_cast<libinput_config_drag_lock_state>(g_pConfigManager->getDeviceInt(devname, "drag_lock", "input:touchpad:drag_lock"));
+            libinput_device_config_tap_set_drag_lock_enabled(LIBINPUTDEV, TAP_DRAG_LOCK);
```
vaxerski: does libinput verify the passed val? I'd clamp it.

vaxerski (commented):
also needs a bump of minimum libinput ver requirement in cmake/meson

m4rch3n1ng:
> also needs a bump of minimum libinput ver requirement in cmake/meson

just fyi this should have happened in #10631, this depends on libinput version 1.27, while that pr uses a feature from 1.28.

not sure how to do that, but i'll figure it out just give me a moment.

src/config/ConfigManager.cpp
m4rch3n1ng: it's what libinput recommends (from https://lists.freedesktop.org/archives/wayland-devel/2024-November/043860.html):

> Desktop environments are encouraged to
> use this as the default as it provides a better experience for anyone with
> less-than-perfect dexterity. For backwards-compatibility reasons libinput cannot
> easily change its defaults without risking bugs in the callers.

but like i said, i put that in a separate commit so i can revert it if you prefer.

src/managers/input/InputManager.cpp
m4rch3n1ng: it does, if you give it an invalid parameter it returns a LIBINPUT_CONFIG_STATUS_INVALID (which hyprland ignores), see:

https://gitlab.freedesktop.org/libinput/libinput/-/blob/ef15d7ed08b7b7af082103f8c2acc2224da59488/src/libinput.c?page=5#L4104

i can change it to only set the option if it is valid, but i wanted to point out that i just copied the code from like 5 lines under it: https://github.com/hyprwm/Hyprland/blob/d14f81e6ac3debd57f8e3e70ead5bb52606a8c90/src/managers/input/InputManager.cpp#L1256-L1257

m4rch3n1ng:
i bumped the required libinput version, hopefully correctly

src/managers/input/InputManager.cpp
m4rch3n1ng: changed it so i only set the config if the value is >= 0 and <= 2, is that okay?

vaxerski: yep

src/config/ConfigManager.cpp
vaxerski: fair enough

PR #10733, opened 2025-06-13
monitor: retry after monitor modesetting fails

### LINKED ISSUE #10678: If a driver rejects all modes, the size remains 0 and crashes the renderer afterwards.
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10649

## Description
Sometimes a display's driver rejects everything and says "fuck you" for magical reasons. In these cases, vecPixelSize etc will not be set, leading to crashes on empty FBs etc.

We should firstly just not render displays that are 0x0, and secondly, potentially try a few times to bring them back, because it could be temporary ("oh the driver doesnt feel like it for the first 50ms after a suspend")

vaxerski:
When a modesetting fails, we can retry up to 4 times. This sometimes happens due to some wonky driver timings.

If an invalid mode is passed and all tries fail, we still have https://github.com/hyprwm/Hyprland/commit/f7526d6be00d75f3abde72883903f9a7116de097 so it should not crash.

Fixes https://github.com/hyprwm/Hyprland/issues/10678

Needs testing from affected users @Lederstrumpf @ryzendew

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -452,6 +452,9 @@ bool CMonitor::applyMonitorRule(SMonitorRule* pMonitorRule, bool force) {
         force = true;
     }
 
+    if (m_failedModeRetries > 0 /* 0 = success */ && m_failedModeRetries < 5 /* Arbitrary number I chose as reasonable */)
+        force = true;
+
     // Check if the rule isn't already applied
     // TODO: clean this up lol
     if (!force && DELTALESSTHAN(m_pixelSize.x, RULE->resolution.x, 1) /* ↓ */
@@ -696,8 +699,12 @@ bool CMonitor::applyMonitorRule(SMonitorRule* pMonitorRule, bool force) {
 
     if (!success) {
         Debug::log(ERR, "Monitor {} has NO FALLBACK MODES, and an INVALID one was requested: {:X0}@{:.2f}Hz", m_name, RULE->resolution, RULE->refreshRate);
+        m_failedModeRetries = 1;
+        g_pEventLoopManager->doLater([] { g_pConfigManager->m_wantsMonitorReload = true; });
         return true;
     }
+
+    m_failedModeRetries = 0;
 
     m_vrrActive = m_output->state->state().adaptiveSync // disabled here, will be tested in CConfigManager::ensureVRR()
         || m_createdByUser;                             // wayland backend doesn't allow for disabling adaptive_sync
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -699,7 +699,7 @@ bool CMonitor::applyMonitorRule(SMonitorRule* pMonitorRule, bool force) {
 
     if (!success) {
         Debug::log(ERR, "Monitor {} has NO FALLBACK MODES, and an INVALID one was requested: {:X0}@{:.2f}Hz", m_name, RULE->resolution, RULE->refreshRate);
-        m_failedModeRetries = 1;
+        m_failedModeRetries++;
         g_pEventLoopManager->doLater([] { g_pConfigManager->m_wantsMonitorReload = true; });
         return true;
     }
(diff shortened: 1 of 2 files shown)


Lederstrumpf (commented):
Thanks for this!

https://github.com/hyprwm/Hyprland/commit/f7526d6be00d75f3abde72883903f9a7116de097 does indeed improves the situation since hyprland now no longer crashes after plugging the external monitor back in, but only fails to recognize it. Upon replugging the monitor, I get a bunch of errors like:
``` bash
error in client communication (pid 2126)
error in client communication (pid 26820)
error in client communication (pid 26905)
error in client communication (pid 26990)
error in client communication (pid 27197)
error in client communication (pid 27281)
```

So in my case (tested with a couple external monitors), the 4 reattempts from this PR didn't help.
Let me know if there are some more detailed logs I should provide.

vaxerski:
yeah that looks like clients dying, my primary concern is whether the monitor comes back up

Lederstrumpf:
> yeah that looks like clients dying, my primary concern is whether the monitor comes back up

The client culling is the only logged crud when the monitor reconnected.

The monitor does come up briefly in `hyprctl monitors all` during the 5 attempts to reconnect, albeit with none of the available modes etc. recognized (Monitor `HDMI-A-1`):
``` bash
❯ hyprctl monitors all
Monitor eDP-1 (ID 0):
	3456x2234@60.00000 at 0x0
	description:
	make:
	model:
	serial:
	active workspace: 1 (1)
	special workspace: 0 ()
	reserved: 0 39 0 0
	scale: 2.00
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: false
	solitary: 0
	activelyTearing: false
	directScanoutTo: 0
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 3456x2234@60.00Hz 3456x2234@59.94Hz 3456x2234@50.00Hz 3456x2234@48.00Hz 3456x2234@47.95Hz

Monitor HDMI-A-1 (ID 1):
	0x0@60.00000 at 1728x0
	description:
	make:
	model:
	serial:
	active workspace: 3 (3)
	special workspace: 0 ()
	reserved: 0 0 0 0
	scale: 2.00
	transform: 0
	focused: no
	dpmsStatus: 1
	vrr: false
	solitary: 0
	activelyTearing: false
	directScanoutTo: 0
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes:
```

After the 5th attempt, the monitor is no longer listed:
```
❯ hyprctl monitors all
Monitor eDP-1 (ID 0):
	3456x2234@60.00000 at 0x0
	description:
	make:
	model:
	serial:
	active workspace: 1 (1)
	special workspace: 0 ()
	reserved: 0 46 0 0
	scale: 2.00
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: false
	solitary: 0
	activelyTearing: false
	directScanoutTo: 0
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 3456x2234@60.00Hz 3456x2234@59.94Hz 3456x2234@50.00Hz 3456x2234@48.00Hz 3456x2234@47.95Hz
```

It's a strange issue since the monitor is recognized right away if I plug in after a reboot. My amateur hunch is that either some cached connection information is cleared during the reboot, or another related process dies when the system enters suspend.

vaxerski:
my guess is your driver being wonky.

src/helpers/Monitor.cpp
```diff
@@ -696,9 +699,13 @@ bool CMonitor::applyMonitorRule(SMonitorRule* pMonitorRule, bool force) {
 
     if (!success) {
         Debug::log(ERR, "Monitor {} has NO FALLBACK MODES, and an INVALID one was requested: {:X0}@{:.2f}Hz", m_name, RULE->resolution, RULE->refreshRate);
+        m_failedModeRetries = 1;
```
staticssleever668: Doesn't this only set it to non-zero instead of incrementing it?

vaxerski: yes, that's correct.

vaxerski: oh, right. It should increment...

PR #10757, opened 2025-06-16
reduce a lot of glcalls and cache various states

gulafaran:
reading nvidia docs they say this.

```
All relevant GL state should be initialized during application initialization and not in the main render loop. For instance, 
occasionally glClearDepthf, glColorMask, or glViewport finds its way into the application render loop even though the 
values passed to these functions are always constant. Other times they are set unconditionally in the loop, just in case 
their values have changed per frame. Only call these functions when the values actually do need to change.
```

gluniform calls are also cached per program, and unless the value has changed it does not need to be called yet again.

this reduced gl* calls in apitrace by a large amount. moving a lot of driver churn to small cpu churn. but needs some testing from others that this doesnt make it now use a lot more cpu then needed. locally it doesnt.

the perfect way would be to really spend weeks into the code and appropiatly only call these uniforms/viewports/scissor when actually needed. but the easy way out is wrapping and caching the values.

i also havent tried HDR, since i dont have a HDR monitor. needs to be tested so it didnt break.


gulafaran:
@UjinT34 i need hdr testing! :heart:

UjinT34:
Might be tricky to test because these changes don't prevent HDR activation and some incorrect/outdated shader values might not be obvious. I guess it's time to choose/create some tool to use as a reference for CM testing.
`SImageDescription` supports faster `==` comparison when `id` is set.  `CHyprOpenGLImpl::passCMUniforms` uses it for color conversion matrix caching. Can be used in a similar way to return early and avoid all the new `setUniform` stuff and checks.

vaxerski:
I think this is ready no?

gulafaran:
> I think this is ready no?

sure ready for review and testing.

src/render/Texture.hpp
```diff
@@ -14,6 +14,17 @@ enum eTextureType : int8_t {
...
 
+enum eTextureParam : int {
+    TEXTURE_WRAP_S     = GL_TEXTURE_WRAP_S,
+    TEXTURE_WRAP_T     = GL_TEXTURE_WRAP_T,
+    TEXTURE_MAG_FILTER = GL_TEXTURE_MAG_FILTER,
+    TEXTURE_MIN_FILTER = GL_TEXTURE_MIN_FILTER,
+    TEXTURE_SWIZZLE_R  = GL_TEXTURE_SWIZZLE_R,
+    TEXTURE_SWIZZLE_B  = GL_TEXTURE_SWIZZLE_B,
+
+    TEXTURE_LAST,
```
vaxerski: this will cause you to allocate like kilobytes extra per texture (where you do std::array). Don't do this. Make a translation table.

gulafaran: how about now?

vaxerski: linear search is way slower compared to an index lookup. Why not do a translation table like I did?

Zebra2711:
I rebased this on the main. Without HDR enabled, HDR displays

```ini
monitorv2 {
    output = eDP-1
    mode = 1920x1080@144
    position = 0x0
    scale = 1
    # bitdepth = 10
    cm = srgb
    sdrbrightness = 2.20
    sdrsaturation = 1.07
    supports_wide_color = 1
    supports_hdr = 0
    sdr_min_luminance = 0.005
    sdr_max_luminance = 200
    min_luminance = 0
    max_luminance = 1000
    max_avg_luminance = 200
}
```

20250621_22h55m55s_grim

```ini
monitorv2 {
    output = eDP-1
    mode = 1920x1080@144
    position = 0x0
    scale = 1
    # bitdepth = 10
    cm = srgb
    sdrbrightness = 2.20
    sdrsaturation = 1.07
    supports_wide_color = 1
    supports_hdr = 0
    sdr_min_luminance = 0.005
    sdr_max_luminance = 200
    # min_luminance = 0
    # max_luminance = 1000
    # max_avg_luminance = 200
}
```
20250621_22h56m25s_grim

Black screen with:
```ini
monitorv2 {
    output = eDP-1
    mode = 1920x1080@144
    position = 0x0
    scale = 1
    # bitdepth = 10
    cm = srgb
    sdrbrightness = 2.20
    sdrsaturation = 1.07
    supports_wide_color = 1
    supports_hdr = 0
    sdr_min_luminance = 0.005
    sdr_max_luminance = 200
    # min_luminance = 0
    # max_luminance = 1000
    max_avg_luminance = 200
}
```
I just want to test this. Not sure if my config is wrong or not, but hope this helps

src/render/Texture.hpp
vaxerski: that's still linear search, lol

gulafaran: I never changed, went out on the island 😅. midsummer vacation

vaxerski: oh right my bad I noticed the commit below

UjinT34:
CM shader values aren't passed correctly. With `cm = srgb` the CM is skipped in most cases. Anything that triggers CM will result in a red tint. `cm = edid` should be enough for testing (might need to force wide color gamut and use `cm = wide` if EDID primaries are not set or equal to sRGB).

src/render/Shader.cpp
```diff
@@ -1,44 +1,204 @@
...
+    auto& cached = uniformStatus.at(location);
+
+    if (cached.index() != 0) {
+        auto val = std::get<SUniformMatrix4Data>(cached);
+        if (val.count == count && val.transpose == transpose && compareFloat(val.value, value))
+            return;
+    }
+
+    cached = SUniformMatrix4Data{.count = count, .transpose = transpose, .value = value};
+    glUniformMatrix3fv(uniformLocations[location], count, transpose, value.data());
```
UjinT34: copy-paste typo %)

src/render/OpenGL.cpp
```diff
@@ -1435,51 +1443,51 @@ void CHyprOpenGLImpl::renderTextureWithDamage(SP<CTexture> tex, const CBox& box,
...
+        const auto                   mat = imageDescription.getPrimaries().convertMatrix(targetImageDescription.getPrimaries()).mat();
         const std::array<GLfloat, 9> glConvertMatrix = {
             mat[0][0], mat[1][0], mat[2][0], //
             mat[0][1], mat[1][1], mat[2][1], //
             mat[0][2], mat[1][2], mat[2][2], //
         };
         primariesConversionCache.insert(std::make_pair(cacheKey, glConvertMatrix));
     }
-    glUniformMatrix3fv(shader.convertMatrix, 1, false, &primariesConversionCache[cacheKey][0]);
+    shader.setUniformMatrix3fv(SHADER_CONVERT_MATRIX, 1, false, {primariesConversionCache[cacheKey][0]});
```
UjinT34: Should be just `primariesConversionCache[cacheKey]` (the array itself instead of the first element copy)

Zebra2711: Changing to `primariesConversionCache[cacheKey]` fixes the red tint

src/render/Shader.cpp
gulafaran: fixed

src/render/OpenGL.cpp
gulafaran: yeah fixed

UjinT34:
@Zebra2711 should work now

Zebra2711:
> @Zebra2711 should work now

Yep, no more red tint when `cm=edid` (or `wide`,...) is set

vaxerski (dismissed):
lgtm

Skidamek:
https://github.com/user-attachments/assets/8e52bd37-28c2-4e65-bdfb-5b92ba5189fb

[systeminfo.txt](https://github.com/user-attachments/files/20865824/systeminfo.txt)

there's something messed up, with animations/borders? best visible on bright wallpaper on special workspace
this is with this patch applied on top of cf7e3aa (main), without it, i don't see these artifacts

gulafaran:
> 2025-06-23.15-09-48.mp4
> 
> [systeminfo.txt](https://github.com/user-attachments/files/20865824/systeminfo.txt)
> 
> there's something messed up, with animations/borders? best visible on bright wallpaper on special workspace this is with this patch applied on top of [cf7e3aa](https://github.com/hyprwm/Hyprland/commit/cf7e3aa448f8c9e0d9e8f407e6ed730da55acc69) (main), without it, i don't see these artifacts

that looks bad indeed, but whats up with all your env vars in your config. LIBGL_ALWAYS_SOFTWARE, __NV_PRIME_RENDER_OFFLOAD , __VK_LAYER_NV_optimus etc

gulafaran:
> > 2025-06-23.15-09-48.mp4
> > [systeminfo.txt](https://github.com/user-attachments/files/20865824/systeminfo.txt)
> > there's something messed up, with animations/borders? best visible on bright wallpaper on special workspace this is with this patch applied on top of [cf7e3aa](https://github.com/hyprwm/Hyprland/commit/cf7e3aa448f8c9e0d9e8f407e6ed730da55acc69) (main), without it, i don't see these artifacts
> 
> that looks bad indeed, but whats up with all your env vars in your config. LIBGL_ALWAYS_SOFTWARE, __NV_PRIME_RENDER_OFFLOAD , __VK_LAYER_NV_optimus etc

there is like 3 different places you set decoration blur/shadow too xD

Skidamek:
> that looks bad indeed, but whats up with all your env vars in your config. LIBGL_ALWAYS_SOFTWARE, __NV_PRIME_RENDER_OFFLOAD , __VK_LAYER_NV_optimus etc

idk, i had them long time lol - they dont seem to affect that

> there is like 3 different places you set decoration blur/shadow too xD

true, i need to clean it up sometime, axshell replaces them tho, i just checked that its reproducible with just [axshell](https://github.com/Axenide/Ax-Shell)

Zebra2711:
> true, i need to clean it up sometime, axshell replaces them tho, i just checked that its reproducible with just [axshell](https://github.com/Axenide/Ax-Shell)

That's because some functions changed and axshell used old functions like `GL.glTexParameteri`
https://github.com/Axenide/Ax-Shell/blob/2ac60548ea449041ba4819ef26197161484486a4/modules/shader.py#L324

Skidamek:
wait, reproduced it without ax shell, just with the hyprland config it changed:

```
general {
    gaps_in = 2
    gaps_out = 4
    border_size = 2
    layout = dwindle
}

cursor {
  no_warps=true
}

decoration {
    blur {
        enabled = yes
        size = 5
        passes = 3
        new_optimizations = yes
        contrast = 1
        brightness = 1
    }
    rounding = 14
    shadow {
      enabled = true
      range = 10
      render_power = 2
      color = rgba(0, 0, 0, 0.25)
    }
}

animations {
    enabled = yes
    bezier = myBezier, 0.4, 0.0, 0.2, 1.0
    animation = windows, 1, 2.5, myBezier, popin 80%
    animation = border, 1, 2.5, myBezier
    animation = fade, 1, 2.5, myBezier
    animation = workspaces, 1, 2.5, myBezier, slidefade 20%
}
```

gulafaran:
> > true, i need to clean it up sometime, axshell replaces them tho, i just checked that its reproducible with just [axshell](https://github.com/Axenide/Ax-Shell)
> 
> That's because some functions changed and axshell used old functions like `GL.glTexParameteri` https://github.com/Axenide/Ax-Shell/blob/2ac60548ea449041ba4819ef26197161484486a4/modules/shader.py#L324

technically i dont think that should break, it merely will call it uncached

UjinT34:
> > > true, i need to clean it up sometime, axshell replaces them tho, i just checked that its reproducible with just [axshell](https://github.com/Axenide/Ax-Shell)
> > 
> > 
> > That's because some functions changed and axshell used old functions like `GL.glTexParameteri` https://github.com/Axenide/Ax-Shell/blob/2ac60548ea449041ba4819ef26197161484486a4/modules/shader.py#L324
> 
> technically i dont think that should break, it merely will call it uncached

Uncached calls won't update the cache and later cached calls might get skipped instead of updating values and cache.

vaxerski:
isn't that a separate process though? I'd assume it's not hyprland as it's written in python.

gulafaran:
> > > > true, i need to clean it up sometime, axshell replaces them tho, i just checked that its reproducible with just [axshell](https://github.com/Axenide/Ax-Shell)
> > > 
> > > 
> > > That's because some functions changed and axshell used old functions like `GL.glTexParameteri` https://github.com/Axenide/Ax-Shell/blob/2ac60548ea449041ba4819ef26197161484486a4/modules/shader.py#L324
> > 
> > 
> > technically i dont think that should break, it merely will call it uncached
> 
> Uncached calls won't update the cache and later cached calls might get skipped instead of updating values and cache.

oOh yes -_-. we set whatever to 2. module sets it to 0, later we try to set it to 2 again, only now its cached as 2 but the module changed it, mismatch and 0xb00b

vaxerski:
> oOh yes -_-. we set whatever to 2. module sets it to 0, later we try to set it to 2 again, only now its cached as 2 but the module changed it, mismatch and 0xb00b

if that python code is a plugin (runs as part of hl) then it's the plugin's job to update itself as hyprland changes. If it's just a client process, it has nothing to do with us.

gulafaran:
@Skidamek and @Zebra2711 reverted a change that broke contrast/blur and stuff. so it should be a lot happier now.

and as always, big thanks for testing things so they dont end up merged and causing havoc on -git :heart:

Skidamek:
Seems to work perfectly fine now, thanks

gulafaran:
@vaxerski yeah think we ruled out the issues, its good to go from here on, from my end.

vaxerski (approved):

PR #10763, opened 2025-06-16
Fix: Errors in previous refactor

JassonCordones:
Removes namespace from CUniquePointer signature name 
Removes get  from reply 

I noticed I didn't push after making changes in last PR

yes

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -20,6 +20,8 @@
 #include "../managers/ANRManager.hpp"
 #include "../protocols/XWaylandShell.hpp"
 #include "../protocols/core/Compositor.hpp"
+using Hyprutils::Memory::CUniquePointer;
+
 using namespace Hyprutils::OS;
 
 #define XCB_EVENT_RESPONSE_TYPE_MASK 0x7f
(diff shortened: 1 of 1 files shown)


src/xwayland/XWM.cpp
```diff
@@ -1289,7 +1291,7 @@ void CXWM::getTransferData(SXSelection& sel) {
     if (transferIndex >= sel.transfers.size())
         return;
 
-    Hyprutils::Memory::CUniquePointer<SXTransfer>& updatedTransfer = sel.transfers[transferIndex];
+    CUniquePointer<SXTransfer>& updatedTransfer = sel.transfers[transferIndex];
```
vaxerski: UP macro already exists

src/xwayland/XWM.cpp
```diff
@@ -578,7 +580,7 @@ xcb_atom_t CXWM::mimeToAtom(const std::string& mime) {
     ReplyPtr reply(xcb_intern_atom_reply(m_connection, cookie, nullptr), &free);
```
staticssleever668: By the way, you don't have to store pointer to 'free' in the unique_ptr if you use own deleter type.

Like this
```cpp
#include <cstdlib>
#include <iostream>
#include <memory>

struct CFree {
  void operator()(void *ptr) { std::free(ptr); }
};

int main(int argc, char *argv[]) {
  std::unique_ptr<int, CFree> reply(static_cast<int *>(malloc(sizeof(int))));

  std::cout << sizeof(std::unique_ptr<int, CFree>) << '\n';                // 8
  std::cout << sizeof(std::unique_ptr<int, decltype(&std::free)>) << '\n'; // 16

  return 0;
}
```

src/xwayland/XWM.cpp
JassonCordones: added it NS back then

src/xwayland/XWM.cpp
JassonCordones: You are right, thanks

PR #10794, opened 2025-06-19
CI/Nix: add cache-nix-action

fufexan:

After https://github.com/DeterminateSystems/magic-nix-cache-action has been deprecated, we've been running Nix Actions without a cache for `/nix/store`. This meant that all dependencies had to be downloaded on each action run.

With this PR, caching will work again.

Use nixbuild/nix-quick-install-action which pairs well with nix-community/cache-nix-action.

Parameters are taken from https://github.com/nix-community/cache-nix-action and may be tweaked later.

Should be replicated for all hyprwm repos with Nix actions.

No.

Ready, but I'd prefer to merge this last, adding the same functionality to all repos down the dependency tree.

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -18,7 +18,35 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - uses: DeterminateSystems/nix-installer-action@main
+      - name: Install Nix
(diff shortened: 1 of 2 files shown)


fufexan:
Looks good, shaved off almost a minute of build time.

PR #10799, opened 2025-06-20
feat: implement comprehensive render batching system

carabistouflette:
## Summary

  - Implement core render batching system with BatchManager
  - Add instanced rendering support for GPU optimization
  - Integrate batched rendering for drop shadows and group bars
  - Add comprehensive performance tests and documentation
  - Include configuration options for render batching control

  ## Key Features

  - **Core Batching System**: New BatchManager that combines similar render operations
  - **Instanced Rendering**: GPU-based instancing for maximum performance on compatible hardware
  - **Shadow Batching**: Optimized batch rendering for drop shadows
  - **Group Bar Integration**: Batched rendering support for group bar decorations
  - **Performance Monitoring**: Built-in metrics and testing framework
  - **Configuration Control**: Runtime toggle for render batching features

  ## Performance Impact

  - Reduces OpenGL state changes by batching similar operations
  - Improves frame rates especially with many windows/decorations
  - Provides fallback paths for compatibility
  - Includes comprehensive test suite for validation

  ## Test Plan

  - [x] Core batching functionality tested
  - [x] Shadow and group bar batching confirmed
  - [x] Configuration options functional
  - [x] Code formatting compliance verified


vaxerski:
There are a few problems with this MR:
- needs a rebase on main
- doc mds are to be removed
- we don't use gtest, and those tests dont look very complex (redundant)
- is there an actual performance benefit? We don't do many drawcalls, I'd say its unlikely this matters?

gulafaran:
> There are a few problems with this MR:
> 
> * needs a rebase on main
> * doc mds are to be removed
> * we don't use gtest, and those tests dont look very complex (redundant)
> * is there an actual performance benefit? We don't do many drawcalls, I'd say its unlikely this matters?

On a multi gpu scenario every bit counts, less synchronisation, less latency if unnecessary work has to be done, but requires benchmarks to properly say how much it matters. im already seeing benefits when caching those uniforms in the hypropti pr, but mostly when dual gpuing.

vaxerski:
the difference is that hypropti is much less code maintenance

vaxerski:
or actually... no. This smells AI to me - half the code is not code but "examples" and "docs". wtf?

carabistouflette:
The examples are part of what I made for you (in order that you understand a bit more easily what i've done). I refined and tested them, but it was mainly so that you understand how batching is implemented. It's done with AI because let's not lie, doing this kind of work is redundant and often tedious, but it can be easily removed. It's better to have too much than not enough, especially when this kind of thing can be removed in just 2 clicks. It's not made for production, it's more for people to understand how to use batching than anything else so when the merge will be done (if it's done), it will be removed of course. The rest isn't made with AI that's why it works lol.

carabistouflette:
Also i'm working on cleaning the fork, the MDs are removed as you asked, if you don't need the examples anymore i'll remove them too

carabistouflette:
I actually came up with the idea for batch processing because I had to implement it on a school project. If it requires too much maintenance, I would totally understand. As @gulafaran said, the solution has advantages in multi GPU scenarios. Anyway, it's more of a gimmick than a core feature. I thought it would be worth looking into some more optimization.

vaxerski:
there is 0 point in examples / docs. We are not making a library. We will have to read the code anyways.

Examples are for users, and users don't care at all about the code.

I'm in general not opposed to the idea, but we need some actual benchmarking done in order to see whether it _actually_ helps. With our amount of drawcalls, I have doubts, but I'm always up to be proven wrong.

carabistouflette:
I'll make some benchmark in this case

PR #10802, opened 2025-06-20
xwayland: Don't leave shell process

Vladimir-csp:
Just adds `exec` to shell command that launches Xwayland.
Before:
image

After:
image

I suppose it would be more elegant to compose argument array and feed it directly to `execl` without any shell, but here is just a minimal change that does the job.

Ready.

### DIFF
--- a/src/xwayland/Server.cpp
+++ b/src/xwayland/Server.cpp
@@ -299,7 +299,7 @@ void CXWaylandServer::runXWayland(CFileDescriptor& notifyFD) {
         _exit(EXIT_FAILURE);
     }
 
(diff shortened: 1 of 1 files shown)


Vladimir-csp:
Potential side effect is that journal will retain `sh` as source process name for messages from Xwayland. This will not be directly visible with uwsm's unit since it has SyslogIdentifier set, but it will still be in message metadata.

fufexan:
Wouldn't `exec -a Xwayland Xwayland...` change the process name?

Vladimir-csp:
`-a` is not in POSIX, although I myself guilty of relying on `env -0` in a key mechanism.

Vladimir-csp:
`dash` does not support `exec -a`, so relying on this would be extra unstable.

Vladimir-csp:
I tried to make an array for execvp(), but I cannot into C :(

PR #10811, opened 2025-06-22
Fix/eventloopmanger

JassonCordones:
Throttle timer loop to avoid high CPU usage and track idle callbacks to avoid flooding

Might need very careful review

Not Ready


Zebra2711:
This makes a huge delay and kinda buggy

JassonCordones:
True, my bad didn't test the correct build

PR #10827, opened 2025-06-24
Added a new config option `master:focus_master_on_close` to improve focus behavior in the master layout.Add config to focus master on window close in master layout

TEJASJONDHALE:
### Description
This PR adds a new configuration option `master:focus_master_on_close` that improves workflow continuity in the master layout by automatically focusing the next master window when a master window is closed, instead of potentially focusing a stack window.

### Problem
Currently, when a window in the master area is closed, focus can end up in the stack area instead of staying in the master area. This disrupts workflow continuity for users who prefer to work primarily in the master area.

### Solution
- Added a new config option `master:focus_master_on_close` (default: 0)
- When enabled, closing a master window will automatically focus the next available master window in the same workspace
- Falls back to default focus behavior if no master window is available or the option is disabled

### Changes

**Files Modified:**
- `src/config/ConfigManager.cpp` - Added config variable registration
- `src/config/ConfigDescriptions.hpp` - Added config description and improved existing descriptions
- `src/layout/MasterLayout.hpp` - Added override for `getNextWindowCandidate`
- `src/layout/MasterLayout.cpp` - Implemented master window focus logic

**Implementation Details:**
- Overrides `getNextWindowCandidate` in `CHyprMasterLayout`
- Checks if the closed window was a master window
- Searches for the next available master window in the same workspace
- Maintains backward compatibility by falling back to default behavior

### Usage
Add to your Hyprland config:
```ini
master:focus_master_on_close = 1

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1122,7 +1122,7 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     },
     SConfigOptionDescription{
         .value       = "misc:force_default_wallpaper",
-        .description = "Enforce any of the 3 default wallpapers. Setting this to 0 or 1 disables the anime background. -1 means “random”. [-1/0/1/2]",
+        .description = "Enforce any of the 3 default wallpapers. Setting this to 0 or 1 disables the anime background. -1 means "random". [-1/0/1/2]",
         .type        = CONFIG_OPTION_INT,
         .data        = SConfigOptionDescription::SRangeData{-1, -1, 2},
     },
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by JassonCordones** [comment]
Why was this closed? wasn't it trying to fix #10813?

**Comment by vaxerski** [comment]
rebase on main please and add a test for your feature

**Comment by vaxerski** [comment]
nonono tests are in hyprtester/, see https://wiki.hypr.land/Contributing-and-Debugging/PR-Guidelines/#how-to-add-tests

Add a new file you can name it `tests/main/master.cpp` because we don't have tests for master layout yet and do it there :)

Check the neighboring files to see how tests are written

File: scripts/test_master_focus_on_close.sh [hunk 1]
```diff
@@ -0,0 +1,227 @@
...
+set -e
+
+echo "=== Testing master:focus_master_on_close feature ==="
+echo
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
```
C0Florent: You might want to use `$'string'`s which have escape sequence parsing built-in

So you could do
```bash
RED=$'\033[0;31m'
GREEN=$'\033[0;32m'
YELLOW=$'\033[1;33m'
NC=$'\033[0m' # No Color
```
It also supports `\e` instead of `\033`, may or may not be more readable.

The main change is that you then wouldn't have to `echo -e` in the rest of your script because the escape sequences would already be interpreted, and stored in the variables like that.

vaxerski: the script is to be completely removed so it doesn't matter

TEJASJONDHALE (commented):
Use $'...' syntax for color escape sequences to remove need for echo -e

vaxerski:
can you rebase on main? I think you included the entire test framework MR in this by accident.

vaxerski:
well now you don't have any tests for this anymore

PR #10865, opened 2025-06-27
session-lock: don't render workspaces when locked

PaideiaDilemma:

Includes a bunch of session lock improvements I wanted to do for a while.
Let me know if you want me to split the PR. The changes are somewhat depend on each other, so I did it like this.
I managed to keep the commit history pretty clean, so it does a ok job of explaining what this PR changes.

##### Details

The major change is that this disables rendering of workspaces when the session is locked by default.
We stop rendering workspaces as soon as the client recieved `locked` or `finished` or the `misc:lockdead_screen_delay` has passed. This works slightly different than other session-lock implementations, that don't wait for `locked` or `finished` to stop rendering workspaces. This behaviour is well within what the protocol describes and makes sure there are no bad frames between the lockscreen client fully covering the screens and the request to lock the session in the first place.

Users can bypass this with the `misc:session_lock_xray` option.

This change sort of made `getRedScreenAlphaForMonitor` obsolete, so I removed it.

To ensure that we can't get into a situation, where neither `locked` or `finished` is sent, I added an event loop timer of 5 seconds that sends finished in case the client  wasn't able to create and commit lock surfaces within that time.

- With hyprlock 0.8.2, this breaks the optics of fade-in (as long as `misc:session_lock_xray` is disabled). It is intended to be used with hyprlock >= https://github.com/hyprwm/hyprlock/commit/a9638986c31ab74fd95e83057f7cdd5187747ec7

- hyprlock's `shape:xray` now requires `misc:session_lock_xray`.

- Not sure, but workspaces stop being rendered may introduce some unexpected side effects. So far I haven't encountered any though 

Theoretically ready. Testing appreciated.

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -94,7 +94,8 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
     g_pCompositor->focusSurface(nullptr);
     g_pSeatManager->setGrab(nullptr);
 
-    m_sessionLock->sendDeniedTimer = SP<CEventLoopTimer>(new CEventLoopTimer(
+    m_sessionLock->sendDeniedTimer = makeShared<CEventLoopTimer>(
+        // Within this arbitrary amount of time, a session-lock client is expected to create and commit a lock surface for each output. If the client fails to do that, it will be denied.
         std::chrono::seconds(5),
         [](auto, auto) {
             if (!g_pSessionLockManager || g_pSessionLockManager->clientLocked() || g_pSessionLockManager->clientDenied())
@@ -113,7 +114,7 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
                 g_pSessionLockManager->m_sessionLock->hasSentDenied = true;
             }
         },
-        nullptr));
+        nullptr);
 
     if (m_sessionLock->sendDeniedTimer)
         g_pEventLoopManager->addTimer(m_sessionLock->sendDeniedTimer);
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -100,6 +100,14 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
             if (!g_pSessionLockManager || g_pSessionLockManager->clientLocked() || g_pSessionLockManager->clientDenied())
                 return;
 
+            if (g_pCompositor->m_unsafeState || !g_pCompositor->m_aqBackend->hasSession() || !g_pCompositor->m_aqBackend->session->active) {
+                // Because the session is inactive, there is a good reason for why the client did't recieve locked or denied.
+                // We send locked, although this could lead to imperfect frames when we start to render again.
+                g_pSessionLockManager->m_sessionLock->lock->sendLocked();
+                g_pSessionLockManager->m_sessionLock->hasSentLocked = true;
+                return;
+            }
+
             if (g_pSessionLockManager->m_sessionLock && g_pSessionLockManager->m_sessionLock->lock) {
                 g_pSessionLockManager->m_sessionLock->lock->sendDenied();
                 g_pSessionLockManager->m_sessionLock->hasSentDenied = true;
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -51,12 +51,15 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
     static auto PALLOWRELOCK = CConfigValue<Hyprlang::INT>("misc:allow_session_lock_restore");
 
     if (PROTO::sessionLock->isLocked() && !*PALLOWRELOCK) {
-        Debug::log(LOG, "Cannot re-lock, misc:allow_session_lock_restore is disabled");
+        LOGM(LOG, "Cannot re-lock, misc:allow_session_lock_restore is disabled");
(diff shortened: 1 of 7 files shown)


src/render/Renderer.cpp
```diff
@@ -990,66 +993,71 @@ void CHyprRenderer::renderLockscreen(PHLMONITOR pMonitor, const Time::steady_tp&
...
-    CBox       monbox = {{}, pMonitor->m_pixelSize};
+    // render image, with instructions. Lock is gone.
+    CBox                         monbox = {{}, pMonitor->m_pixelSize};
+    CTexPassElement::SRenderData data;
+    data.tex = g_pHyprOpenGL->m_lockDeadTexture;
+    data.box = monbox;
+    data.a   = 1;
 
-    const bool ANY_PRESENT = g_pSessionLockManager->anySessionLockSurfacesPresent();
+    m_renderPass.add(makeShared<CTexPassElement>(data));
```
PointerDilemma, line 1054: Was there a particular reason why renderSessionLockMissing didn't use render passes?

vaxerski: a mistake?

PointerDilemma: Also can't find a reason why. Works just fine.

PointerDilemma:
A plugin test fails. Not sure why.

vaxerski:
says right there
20250629_00h52m11s

you either removed an option and didnt remove the description, or added one and forgot the description.

src/managers/SessionLockManager.cpp
```diff
@@ -90,14 +94,47 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
     g_pCompositor->focusSurface(nullptr);
     g_pSeatManager->setGrab(nullptr);
 
+    m_sessionLock->sendDeniedTimer = SP<CEventLoopTimer>(new CEventLoopTimer(
```
vaxerski: why not makeShared

src/managers/SessionLockManager.cpp
```diff
@@ -90,14 +94,47 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
...
+        std::chrono::seconds(5),
```
vaxerski, line 99: ?

VirtCode:
Did some testing of this today and things seem to work on my end.

What's not yet changed is how damage is propagated. So if I let's say am watching a video and then lock my screen, the lockscreen is still constantly damaged in the area where the video was rendered (and is still playing). Maybe we can discard the damage from things that are not rendered? A notable exception for the damage would be `abovelock` layer surfaces though.

src/managers/SessionLockManager.cpp
PointerDilemma: I yoinked it from the animation manager and didn't think about it. I replaced both occurrences with makeShared.

src/managers/SessionLockManager.cpp
PointerDilemma: PR desc:

> To ensure that we can't get into a situation, where neither locked or finished is sent, I added an event loop timer of 5 seconds that sends finished in case the client wasn't able to create and commit lock surfaces within that time.

See also the comment I added. The timer has two purposes:
It is there so that a client can't just create a new lock and then do nothing (thus essentially inhibiting a new session-lock). And with this PR, in addition, the rendering for lockscreen and workspaces is tied to `locked` and `finished`. Thus it also serves as a fallback if for example the session-lock client fails to create a surface for one of the outputs.

This is the relevant protocol section:
> The compositor may wait for the client to create and render session lock surfaces before sending the locked event to avoid displaying intermediate blank frames. However, it must impose a reasonable time limit if waiting and send the locked event as soon as the hard requirements described above can be met if the time limit expires. Clients should immediately create lock surfaces for all outputs on creation of this object to make this possible.

PointerDilemma:
Yeah I was aware that some windows would still continue to do stuff.
In terms of damage, i am not sure how relevant that is, because I think now the damage is essentially ignored, because we don't actually render it.

Maybe it would make sense to suspend all toplevels when locking and unsuspending them on an unlock???

vaxerski:
I think we could employ a similar tactic as we do with occlusion in pass to damage in general. But that's long-term

src/render/Renderer.cpp
```diff
@@ -990,66 +993,71 @@ void CHyprRenderer::renderLockscreen(PHLMONITOR pMonitor, const Time::steady_tp&
...
 
-    if (ANY_PRESENT) {
-        // render image2, without instructions. Lock still "alive", unless texture dead
-        g_pHyprOpenGL->renderTexture(g_pHyprOpenGL->m_lockDead2Texture, monbox, ALPHA);
-    } else {
-        // render image, with instructions. Lock is gone.
-        g_pHyprOpenGL->renderTexture(g_pHyprOpenGL->m_lockDeadTexture, monbox, ALPHA);
```
PointerDilemma, line 1040: Oh no, I sort of forgot about lockdead2, cause it rarely shows.
I will add it back with the ANY_PRESENT. At least in theory, it should handle cases like that: https://github.com/basecamp/omarchy/issues/45#issuecomment-3034496321

src/managers/SessionLockManager.cpp
vaxerski: ok

src/render/Renderer.cpp
vaxerski: ok

vaxerski:
wiki mr pls

PointerDilemma:
linked

Issue #9888, opened 2025-04-05
Border rendering does not work for the first few seconds on launch


vaxerski:
### Discussion Link
#9875

## Description
Borders don't render for the first few seconds on start.

They start working after opening the first window and then a few seconds later.

Bisected to https://github.com/hyprwm/Hyprland/pull/9600

nnyyxxxx:
> Borders don't render for the first few seconds on start. 
actually borders dont get rendered UNTIL you open a window and even after spawning that window it takes a few seconds to actually render

nnyyxxxx:
so if you never end up spawning a window, wait 10 minutes, open a window you'll have to wait like 2 seconds until you see the border

vaxerski:
edited for clarity, thanks

UjinT34:
Might be fixed or reduced by #9814

vaxerski:
do we know why it's happening at all?

UjinT34:
If it doesn't happen with `render:cm_enabled = 0` than it is some shader compilation performance issue. Should be the same with shadows.

alba4k:
> If it doesn't happen with `render:cm_enabled = 0`

it doesn't indeed, but does that mean your pr fix the issue?

vaxerski:
doesn't change it on my end :P

PR #10888, opened 2025-06-30
opengl: add missing skipcm if case

gulafaran:
missing skipcm if case so the CM uniforms where never added on the gradient2 renderBorder case, until the non gradient2 one had run atleast once. causing it to not render on first launch/delayed.

fixes: https://github.com/hyprwm/Hyprland/issues/9888

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -2367,6 +2367,12 @@ void CHyprOpenGLImpl::renderBorder(const CBox& box, const CGradientValueData& gr
     blend(true);
 
     useProgram(m_shaders->m_shBORDER1.program);
+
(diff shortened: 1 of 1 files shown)


gulafaran:
@UjinT34 i guess this makes sense? was mostly trial and error here

gulafaran:
> @UjinT34 i guess this makes sense? was mostly trial and error here

yeah it certainly makes sense, borders where glitching over shadows before with cm_enabled = true this solves that too.

UjinT34:
Yeah, missed this one. Those 2 `renderBorder` functions have a lot of common code. Might be reasonable to extract the common part to avoid similar mistakes in the future

vaxerski (approved):

Issue #10471, opened 2025-05-18
Cursor is not re-shown after hyprlock exits with hide_cursor true


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10449

## Description
If hyprlock has hide cursor enabled, and there are no windows on the workspace, cursor will not be shown on unlock.

nnyyxxxx:
> there are no windows on the workspace, cursor will not be shown on unlock.

it happens when the cursor is over the background, windows can be on the workspace

PR #10889, opened 2025-06-30
sessionlock: restore cursor if hidden on unlock

gulafaran:
if session locks have hidden the cursor its gonna be missing unless a new cursor shape is set, hovering windows makes us get one, moving the wallpaper/desktop does not. set it again to left_ptr as is default on compositor start.

fixes: https://github.com/hyprwm/Hyprland/issues/10471

### DIFF
--- a/src/protocols/SessionLock.cpp
+++ b/src/protocols/SessionLock.cpp
@@ -6,6 +6,7 @@
 #include "core/Compositor.hpp"
 #include "core/Output.hpp"
 #include "../helpers/Monitor.hpp"
+#include "../render/Renderer.hpp"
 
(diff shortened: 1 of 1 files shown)


src/protocols/SessionLock.cpp
```diff
@@ -120,6 +121,9 @@ CSessionLock::CSessionLock(SP<CExtSessionLockV1> resource_) : m_resource(resourc
 
         m_events.unlockAndDestroy.emit();
 
+        if (g_pHyprRenderer) // if lock tools have hidden it and destroys.
```
vaxerski: eeee wat?

gulafaran: was thinking of some edge case on comp destruction incase hyprlock was running, and locked. But i can remove the if check it might not be possible 😄

gulafaran: removed.

PR #10907, opened 2025-07-02
hyprpm: print all dependencies that are missing

bruhman325:
It prints the first dependency that is missing.
(I was very confused for like 30 minutes because I was missing one dependency and I just didn't realize it)

No.

I think so.

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -983,7 +983,7 @@ bool CPluginManager::hasDeps() {
 
     for (auto const& d : deps) {
         if (!execAndGet("command -v " + d).contains("/")) {
-            std::println(stderr, "\n{}", failureString("Missing dependency: {}", d));
+            std::println(stderr, "{}", failureString("Missing dependency: {}", d));
             hasAllDeps = false;
         }
     }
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
I'd print all that are missing, for the sake of people's sanity when they miss 3 or 4

bobrat1:
> I'd print all that are missing, for the sake of people's sanity when they miss 3 or 4

alright this should be better

hyprpm/src/core/PluginManager.cpp
```diff
@@ -978,11 +978,15 @@ std::string CPluginManager::headerErrorShort(const eHeadersErrors err) {
...
 
 bool CPluginManager::hasDeps() {
+    bool hasAllDeps = true;
     std::vector<std::string> deps = {"meson", "cpio", "cmake", "pkg-config", "g++", "gcc", "git"};
+
     for (auto const& d : deps) {
-        if (!execAndGet("command -v " + d).contains("/"))
-            return false;
+        if (!execAndGet("command -v " + d).contains("/")) {
+            std::println(stderr, "\n{}", failureString("Missing dependency: {}", d));
```
vaxerski: why the newline...? println already puts a newline.

bobrat1: sorry I copied another line that printed a failure string and thats why thats there

vaxerski (approved):

PR #10920, opened 2025-07-03
handle setGroupCurrent properly on fulscreen groups

MightyPlaza:
always warps window size and pos on `setGroupCurrent`
replaces hacky solution for restoring last floating size and pos by just overriding the values set when changing in/out of fullscreen

seems to work fine

example of what this fixes:
create a workspace with a window and a group with 2 windows
fulscreen the group
change the active group window
notice it always animates

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1041,29 +1041,32 @@ void CWindow::setGroupCurrent(PHLWINDOW pWindow) {
 
     const auto CURRENTISFOCUS = PCURRENT == g_pCompositor->m_lastWindow.lock();
 
+    const auto PWINDOWSIZE                 = PCURRENT->m_realSize->goal();
+    const auto PWINDOWPOS                  = PCURRENT->m_realPosition->goal();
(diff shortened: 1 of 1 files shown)


src/desktop/Window.cpp
```diff
@@ -1041,30 +1041,33 @@ void CWindow::setGroupCurrent(PHLWINDOW pWindow) {
...
+
+    pWindow->m_lastFloatingSize     = PWINDOWLASTFLOATINGSIZE;
+    pWindow->m_lastFloatingPosition = PWINDOWLASTFLOATINGPOSITION;
 
     g_pCompositor->updateAllWindowsAnimatedDecorationValues();
 
     if (CURRENTISFOCUS)
         g_pCompositor->focusWindow(pWindow);
 
-    if (FULLSCREEN)
```
vaxerski, line 1065: why remove this? Wouldn't that unfullscreen the window?

MightyPlaza: wasn't removed, just shifted above
made more sense to set fullscreen before updating the decorations, but can be reverted

vaxerski: ah. right. Yeah.

vaxerski (approved):

PR #10924, opened 2025-07-04
a few optimisations

gulafaran:
in the hunt for glmark2 lost points i gone and optimized a few different things ive found. the glmark2 hunt still continues tho.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -1201,10 +1201,10 @@ void CHyprRenderer::renderMonitor(PHLMONITOR pMonitor) {
         firstLaunchAnimActive = false;
     }
 
-    renderStart = std::chrono::high_resolution_clock::now();
-
-    if (*PDEBUGOVERLAY == 1)
+    if (*PDEBUGOVERLAY == 1) {
+        renderStart = std::chrono::high_resolution_clock::now();
         g_pDebugOverlay->frameData(pMonitor);
+    }
 
     if (!g_pCompositor->m_sessionActive)
         return;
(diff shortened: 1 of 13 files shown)


src/desktop/Window.hpp
```diff
@@ -282,6 +282,9 @@ class CWindow {
     // focus stuff
     bool m_stayFocused = false;
 
+    // only render it once per frame
+    bool m_rendered = false;
```
vaxerski: I really don't think this is the right place to keep such props.

src/render/Renderer.cpp
```diff
@@ -169,6 +169,9 @@ CHyprRenderer::~CHyprRenderer() {
 }
 
 bool CHyprRenderer::shouldRenderWindow(PHLWINDOW pWindow, PHLMONITOR pMonitor) {
+    if (pWindow->m_rendered)
```
vaxerski: why? We already track that in render client fns

gulafaran: Because if we have multiple windows and passes going in one begin -> end , a lot of loops is looping over same windows this makes it early return and continue instead of doing tons of various checks before going nope, only for the next loop to also. thats the idea but feels like there might be time for a 1k loc line overhaul and reduce all the enormous amounts of looping on same windows everywhere

vaxerski: we dont do that though. See `CHyprRenderer::renderWorkspaceWindows`. We don't loop over all of them many times.

gulafaran: renderWorkspaceWindowsFullscreen , loops all of them multiple times. and then functions like renderAllClientsForWorkspace as you mentioned and sendFrameEventsToWorkspace and damageWindow if forcefull damage, all loops over m_windows yet again even if it has been rendered once. meaning they all do the various calcs for visible/etc etc of shouldRender without an early return that it has actually already been processed. there is also various places where its used inside certain checks for windows ("the part im worried might regress")

gulafaran: so what im meaning is m_windows contains idk 50 windows, its all looped and iterated all over and gets worse the more passes/windows/monitors you have going on. and all the checks and calcs multiply down exponentially. even tho its already rendered. 

how to solve it cleanly idk, feels like i shortcuted it xD

gulafaran: got an idea, il drop this commit and prototype it in another PR :p

src/desktop/Window.hpp
gulafaran: got an idea, il drop this commit and prototype it in another PR :p

src/render/Renderer.cpp
vaxerski: ok

PR #10952, opened 2025-07-06
config: properly handle symlinks

fufexan:

Should fix #10887.
@joo-was-already-taken @cebem1nt please test this.

Possibly not idiomatic handling of vectors/maps.

Needs testing.

### DIFF
--- a/src/config/ConfigWatcher.cpp
+++ b/src/config/ConfigWatcher.cpp
@@ -1,6 +1,7 @@
 #include "ConfigWatcher.hpp"
 #include <sys/inotify.h>
 #include "../debug/Log.hpp"
+#include <algorithm>
 #include <ranges>
 #include <fcntl.h>
 #include <unistd.h>
@@ -40,6 +41,7 @@ void CConfigWatcher::setWatchList(const std::vector<std::string>& paths) {
     }
 
     m_watches.clear();
+    m_symlinks.clear();
 
     // add new paths
     for (const auto& path : paths) {
@@ -52,9 +54,11 @@ void CConfigWatcher::setWatchList(const std::vector<std::string>& paths) {
         const auto      CANONICAL  = std::filesystem::canonical(path, ec);
         const auto      IS_SYMLINK = std::filesystem::is_symlink(path, ec2);
         if (!ec && !ec2 && IS_SYMLINK) {
+            Debug::log(INFO, "Found symlink {} -> {}", path, CANONICAL.c_str());
+            m_symlinks.insert_or_assign(path, CANONICAL.c_str());
             m_watches.emplace_back(SInotifyWatch{
                 .wd   = inotify_add_watch(m_inotifyFd.get(), CANONICAL.c_str(), IN_MODIFY),
-                .file = path,
+                .file = CANONICAL.c_str(),
             });
         }
     }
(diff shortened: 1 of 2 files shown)


src/config/ConfigWatcher.hpp
```diff
@@ -27,6 +28,7 @@ class CConfigWatcher {
 
     std::function<void(const SConfigWatchEvent&)> m_watchCallback;
     std::vector<SInotifyWatch>                    m_watches;
+    std::map<std::string, std::string>            m_symlinks;
```
C0Florent, line 31: From what I've just quickly read, there is no specific need for this map to be ordered, so we might prefer `std::unordered_map` here?

src/config/ConfigWatcher.cpp
```diff
@@ -74,6 +78,33 @@ void CConfigWatcher::onInotifyEvent() {
...
+                m_watches.erase(begin, end);
+
+                Debug::log(INFO, "Removing {} inotify watch", prev_symlink);
+                inotify_rm_watch(m_inotifyFd.get(), prev_symlink_watch->wd);
+
+                Debug::log(INFO, "Adding {} inotify watch", CANONICAL.c_str());
+                m_watches.emplace_back(SInotifyWatch{
+                    .wd   = inotify_add_watch(m_inotifyFd.get(), CANONICAL.c_str(), IN_MODIFY),
+                    .file = CANONICAL.c_str(),
+                });
```
C0Florent, line 104: I've been reading this bit of logic again and again to clearly figure out what's happening, and I think we could simply edit in place the vector element which is being updated instead of removing then adding again.

I also feel like this will be useful because the removal logic seems invalid to me: `std::ranges::remove_if` is already removing elements, so calling `std::erase` right after would probably erase out-of-range elements.

By the way, I think we already have access to the previous watch in `prev_symlink_watch`, so there's probably no need to loop (or `remove_if` / `find_id`) over the `m_watches`

Something along the lines of the following:
```suggestion
                Debug::log(INFO, "Removing {} inotify watch", prev_symlink);
                inotify_rm_watch(m_inotifyFd.get(), prev_symlink_watch->wd);

                Debug::log(INFO, "Adding {} inotify watch", CANONICAL.c_str());
                *prev_symlink_watch = SInotifyWatch{
                    .wd   = inotify_add_watch(m_inotifyFd.get(), CANONICAL.c_str(), IN_MODIFY),
                    .file = CANONICAL.c_str(),
                };
```

No idea if this works, untested, but at least it makes sense to me, hopefully makes sense to you too

fufexan: `remove/remove_if` doesn't actually remove any elements, it just puts them at the end of the range (or after the last valid iterator). I took https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom into account when writing this. Seems we can use `erase_if` directly though.

And yeah we can probably directly replace `prev_symlink_watch` instead of remove-emplace.

joo-was-already-taken: This hasn't fixed it for me, still the same behavior as described in #10887. The errors in logs also look the same. If there is something specific you want me to test or share, I'm happy to help.
[Logs capturing `home-manager switch`](https://github.com/user-attachments/files/21107669/hypr-logs.txt)  
[Full logs ending after `home-manager switch`](https://github.com/user-attachments/files/21107681/hyprland.log)

joo-was-already-taken:
This hasn't fixed it for me, still the same behavior as described in #10887. The errors in logs also look the same. If there is something specific you want me to test or share, I'm happy to help.
[Logs capturing `home-manager switch`](https://github.com/user-attachments/files/21107669/hypr-logs.txt)  
[Full logs ending after `home-manager switch`](https://github.com/user-attachments/files/21107681/hyprland.log)
[My home-manager hyprland module](https://github.com/user-attachments/files/21108009/hyprland-default.txt)

fufexan:
@joo-was-already-taken are you sure it's not the kb variant interfering, or maybe your layout changing script?

joo-was-already-taken:
Ok, I know why my config didn't work, the last line of Hyprland config was `submap = group` instead of `submap = reset` closing a submap (changing this fixed the issue). I am very sorry for not noticing this earlier. Also checked if changing the above line works for me on the main branch, it does. As far as I'm concerned the issue was resolved even before this fix.

On the side note, it would be nice to treat an unclosed submap definition as error.

cebem1nt:
In my case there is no not closed submap, though issue appears really rarely and the last time it happened, i did not take any logs, my bad. I can test this pull, though I really can not represent an issue on purpose, that means it can take long time on saying: does this pull fix the issue or no.

fufexan:
I'll close this then, seeing it's not actually needed. Vaxry wasn't happy with the code quality anyway.

C0Florent:
> Ok, I know why my config didn't work, the last line of Hyprland config was `submap = group` instead of `submap = reset` closing a submap (changing this fixed the issue).

You'll be able to easily avoid this kind of mistake when my PR for hyprland submaps in home-manager will be merged (I still need to provide tests for it)

PR #10974, opened 2025-07-09
Prevent double-free in attemptDirectScanout 

Kam1k4dze:
Fixes https://github.com/hyprwm/Hyprland/discussions/10940

This is what came to my mind after looking at the asan https://github.com/hyprwm/Hyprland/discussions/10940#discussioncomment-13702123
Sorry, if this is not right way to fix that.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1554,10 +1554,9 @@ bool CMonitor::attemptDirectScanout() {
     // lock buffer while DRM/KMS is using it, then release it when page flip happens since DRM/KMS should be done by then
     // btw buffer's syncReleaser will take care of signaling release point, so we don't do that here
     PBUFFER->lock();
-    auto weakBuffer = WP<IHLBuffer>(PBUFFER);
-    PBUFFER->onBackendRelease([weakBuffer]() {
-        if (auto buffer = weakBuffer.lock())
-            buffer->unlock();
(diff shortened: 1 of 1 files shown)


gulafaran:
just to keep it somewhat like the other lambda captures does i would do 

```
PBUFFER->onBackendRelease([weakBuffer = WP<IHLBuffer>(PBUFFER)]() {
    if (!weakBuffer)
         return;
         
    weakBuffer->unlock();
});

```

but should work the same practically. :+1:

src/helpers/Monitor.cpp
```diff
@@ -1554,7 +1554,11 @@ bool CMonitor::attemptDirectScanout() {
     // lock buffer while DRM/KMS is using it, then release it when page flip happens since DRM/KMS should be done by then
     // btw buffer's syncReleaser will take care of signaling release point, so we don't do that here
     PBUFFER->lock();
-    PBUFFER->onBackendRelease([PBUFFER]() { PBUFFER->unlock(); });
+    auto weakBuffer = WP<IHLBuffer>(PBUFFER);
+    PBUFFER->onBackendRelease([weakBuffer]() {
+        if (auto buffer = weakBuffer.lock())
+            buffer->unlock();
+    });
```
vaxerski, line 1560:
```cpp
PBUFFER->onBackendRelease([wb = WP<IHLBuffer>{PBUFFER}] {
    if (wb)
        wb->unlock();
});
```

vaxerski (approved):

PR #10976, opened 2025-07-09
Do not allow a null `CEGLSync` to reach `doOnReadable`

myQwil:

The function `CEventLoopManager::doOnReadable` is expecting its first parameter to be a non-null value, but `CEGLSync::create` can potentially return a `nullptr`.

If `m_sync` is null, and it reaches `doOnReadable`, Hyprland will crash.

This is one way of addressing the problem, but there may be a better one. For example, maybe the null check should happen directly in `doOnReadable`. Here, I'm checking only where a null value might potentially reach the function.

It's ready for merging.

### DIFF
--- a/src/helpers/MonitorFrameScheduler.cpp
+++ b/src/helpers/MonitorFrameScheduler.cpp
@@ -102,6 +102,10 @@ void CMonitorFrameScheduler::onFrame() {
 
 void CMonitorFrameScheduler::onFinishRender() {
     m_sync = CEGLSync::create(); // this destroys the old sync
+    if (!m_sync) {
+        Debug::log(ERR, "onFinishRender: failed to create CEGLSync, skipping.");
+        return;
+    }
     g_pEventLoopManager->doOnReadable(m_sync->fd().duplicate(), [this, mon = m_monitor] {
         if (!m_monitor) // might've gotten destroyed
             return;


gulafaran:
you could use the isValid to catch more cases.

```
    if (!m_sync->isValid()) {
        Debug::log(ERR, "onFinishRender: failed to create CEGLSync, skipping.");
        return;
    }
```

question is if it should call onSyncFired directly then or return?

myQwil:
I'm not sure that'll work. `CFileDescriptor::isValid()` is also expecting a non-null value, plus `doOnReadable` already calls the `isValid()` function

gulafaran:
> I'm not sure that'll work. `CFileDescriptor::isValid()` is also expecting a non-null value, plus `doOnReadable` already calls the `isValid()` function

oh lol the CEGLSync::isValid() is rather broken indeed, if the creation fails nullptr is returned. and isValid wont work.

myQwil:
Actually, never mind. I think the problem is `m_sync->fd().duplicate()`, so it never even enters the function `doOnReadable`. Instead, it crashes while trying to perform `fd().duplicate()`

gulafaran:
> Actually, never mind. I think the problem is `m_sync->fd().duplicate()`, so it never even enters the function `doOnReadable`. Instead, it crashes while trying to perform `fd().duplicate()`

when is this crashing?

gulafaran:
ah renderer does this, https://github.com/hyprwm/Hyprland/blob/9517d0eaa4ef93de67dc80fecca7a826f7ad556d/src/render/Renderer.cpp#L2316 so do something similiar i guess?

myQwil:
> when is this crashing?

When trying to run a debug build of Hyprland, it was crashing while starting up. It has something to do with `eglCreateSyncKHR` failing.

gulafaran:
> > when is this crashing?
> 
> I'm trying to run a debug build of Hyprland and it crashes while starting up. It has to do with `eglCreateSyncKHR` failing.

```
void CMonitorFrameScheduler::onFinishRender() {
    m_sync = CEGLSync::create(); // this destroys the old sync

    if(m_sync && m_sync->isValid()) {
        g_pEventLoopManager->doOnReadable(m_sync->fd().duplicate(), [this, mon = m_monitor] {
            if (!m_monitor) // might've gotten destroyed
                return;
            onSyncFired();
        });
    }
    else
        Debug::log(ERR, "onFinishRender: failed to create CEGLSync, skipping.");
}
```

now the question still remains if at the end it should just call onSyncFired() anyways.
@vaxerski

vaxerski:
if we can't sync, we should not try to triple buffer, as it won't work. Entire pipeline should be skipped and fallback to standard frame sync should be used

PR #10980, opened 2025-07-09
renderer: use CRegion foreach over getRects

gulafaran:
instead of allocating and returning a vector, use forEach to directly call a function on the rects.

depends on: https://github.com/hyprwm/hyprutils/pull/65


gulafaran:
with the release of hyprutils 0.8.2 this should be good to go @vaxerski

PR #10991, opened 2025-07-11
remove rathandler and dont spuriously trigger onx11events

gulafaran:
rathandler doesnt seem used, remove it.

wl_event_source_check makes the event trigger until onx11event returns non
zero. but we arent removing the event source from the event loop so lets
not mark it at all and recieve spurious constant calls. meaning if we have no x11 events its gonna endlessly trigger and poll and return 0.


gulafaran:
https://github.com/hyprwm/hyprland-wiki/pull/1144 wiki PR too

vaxerski (approved):

PR #11018, opened 2025-07-14
config: disable hw on mgpu nvidia by default

vaxerski:
cc @gulafaran can you test this works?

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -73,7 +73,7 @@ CHyprRenderer::CHyprRenderer() {
 
             drmFreeVersion(DRMV);
         }
-        m_mgpu = drmDevices > 0;
+        m_mgpu = drmDevices > 1;
     } else {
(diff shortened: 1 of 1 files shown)


gulafaran:
Wont this only disable it when the main gpu is nvidia? The problem arises when intel is driving and blitting to nvidia too

vaxerski:
No, m_nvidia is set if any gpu is nvidia.

src/render/Renderer.cpp
```diff
@@ -2371,6 +2374,10 @@ bool CHyprRenderer::isNvidia() {
     return m_nvidia;
 }
 
+bool CHyprRenderer::isMgpu() {
+    return m_nvidia;
```
staticssleever668:
```suggestion
    return m_mgpu;
```
typo?

vaxerski: typo

PR #11026, opened 2025-07-15
Fix preferred image description

UjinT34:
Additional fixes for https://github.com/mpv-player/mpv/issues/16533
- [x] Incorrect icc fd check
- [x] Return output's image description as preferred global if it's the only one
- [x] Send surface preferred description after it's mapped (might be requested before that)

mpv checks the preferred image description before HL finishes initializing the window. The proto requires an immediate response to such a request. This response will use a global image description which will be a default SDR for multi-monitor setups. Also mpv won't support `render:cm_auto_hdr` by default since this is triggered by the client request and mpv will request the same as the current monitor mode. ~For HDR with mpv greater than v0.40.0 monitor should be switched to HDR mode manually. Can be done both before and during video playback.~ Use `mpv --target-colorspace-hint-mode=source ...` to fix that.

Ready

### DIFF
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -541,18 +541,18 @@ void CWLSurfaceResource::commitState(SSurfaceState& state) {
 }
 
 SImageDescription CWLSurfaceResource::getPreferredImageDescription() {
-    auto PARENT = m_self;
-    if (PARENT->m_role->role() == SURFACE_ROLE_SUBSURFACE) {
-        auto subsurface = ((CSubsurfaceRole*)PARENT->m_role.get())->m_subsurface.lock();
-        PARENT          = subsurface->t1Parent();
+    auto parent = m_self;
+    if (parent->m_role->role() == SURFACE_ROLE_SUBSURFACE) {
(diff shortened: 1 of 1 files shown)


src/protocols/core/Compositor.cpp
```diff
@@ -539,7 +541,18 @@ void CWLSurfaceResource::commitState(SSurfaceState& state) {
 }
 
 SImageDescription CWLSurfaceResource::getPreferredImageDescription() {
-    return m_enteredOutputs.size() == 1 ? m_enteredOutputs[0]->m_imageDescription : g_pCompositor->getPreferredImageDescription();
+    auto PARENT = m_self;
+    if (PARENT->m_role->role() == SURFACE_ROLE_SUBSURFACE) {
+        auto subsurface = ((CSubsurfaceRole*)PARENT->m_role.get())->m_subsurface.lock();
+        PARENT          = subsurface->t1Parent();
+    }
+    WP<CMonitor> MONITOR;
```
vaxerski: the caps for parent and monitor are a bit confusing (caps measn const) so this should bedropped

vaxerski (approved):

PR #11051, opened 2025-07-17
seat: set keyboardFocus when m_lastFocus doesn't match the keyboard focus…

PaideiaDilemma:

Attempts to fix issues like this: https://github.com/hyprwm/hyprlock/issues/797

I thought it would also fix the cursor not being hidden when switching ttys without mouse movement. For some reason that is not the case though. (Also not when instead of using `setKeyboardFocus()`, calling the entire `g_pInputManager->refocus()` function.)

The current problem is easily reproducible.

1. `sleep 5 && hyprlock`
2. switch tty and wait for the sleep to finish
3. switch back
4. SLS has no keyboard focus

This fixes that.
Although I am not sure if there would be a better place to handle this stuff.
Alternative solution ideas welcome.

Theoretically ready.

### DIFF
--- a/src/managers/SeatManager.cpp
+++ b/src/managers/SeatManager.cpp
@@ -90,8 +90,13 @@ void CSeatManager::setKeyboard(SP<IKeyboard> KEEB) {
         m_keyboard->m_active = false;
     m_keyboard = KEEB;
 
-    if (KEEB)
+    if (KEEB) {
         KEEB->m_active = true;
+
+        // m_lastFocus was set without a keyboard
+        if (KEEB->m_enabled && g_pCompositor->m_lastFocus && g_pCompositor->m_lastFocus != m_state.keyboardFocus)
+            setKeyboardFocus(g_pCompositor->m_lastFocus.lock());
+    }
 
     updateActiveKeyboardData();
 }


vaxerski (commented):
looks like a band-aid that doesnt fix the root cause. I don't like this. It'd be better to just refocus on coming back from vt. IIRC there is an event we listen to, .activate?

PointerDilemma:
Yeah it is a band aid solution.
The problem is the order of events when activating.
When it comes to focus, we don't have a keyboard yet.

vaxerski:
Refocus on the first keyboard added?

PointerDilemma:
@vaxerski 

You like this better? I feel like the fix this PR here is a bit better.

```diff
diff --git a/src/managers/input/InputManager.cpp b/src/managers/input/InputManager.cpp
index dff97166f9..f8265d88dc 100644
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -937,6 +937,9 @@
 
     setupKeyboard(PNEWKEYBOARD);
 
+    if (m_keyboards.size() == 1 && g_pCompositor->m_lastFocus)
+        g_pSeatManager->setKeyboardFocus(g_pCompositor->m_lastFocus.lock());
+
     Debug::log(LOG, "New keyboard created, pointers Hypr: {:x}", rc<uintptr_t>(PNEWKEYBOARD
.get()));
 }
```

vaxerski:
that looks more reasonable yes

PR #11058, opened 2025-07-17
Add cursor:enabled to allow to disable it 

Mozzarella32:
It allows users to disable the cursor as it now is possible to draw you own cursor using the screen shader

There is an issue when using the hardware cursor with disabled cursor as the monitor(s) don't get damaged(with demage tracking 0 as otherwise the pointer can not be used by the shader) and the shader is not rerun

Advice is needed on the hadware cursor issue

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -2030,6 +2030,7 @@ void CHyprRenderer::setCursorFromName(const std::string& name, bool force) {
 }
 
 void CHyprRenderer::ensureCursorRenderingMode() {
+    static auto PENABLED       = CConfigValue<Hyprlang::INT>("cursor:enabled");
     static auto PCURSORTIMEOUT = CConfigValue<Hyprlang::FLOAT>("cursor:inactive_timeout");
     static auto PHIDEONTOUCH   = CConfigValue<Hyprlang::INT>("cursor:hide_on_touch");
     static auto PHIDEONKEY     = CConfigValue<Hyprlang::INT>("cursor:hide_on_key_press");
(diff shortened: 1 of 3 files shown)


UjinT34:
`rendered` or `invisible` might be a better name.
HW cursor won't work with the shader because it is overlayed on screen without damaging the underlying image (so no need for damage tracking). And shaders draw on the image itself. This new setting should also act as `cursor:no_hardware_cursors`. You might also need to set a transparent rectangle instead of the regular cursor instead of hiding it.

Mozzarella32:
I renamed cursor:enabled to cursor:invisible and made it enforce software rendering. All combies of no_hadware_cursors and invisible worked as expected and the shader got invoked when the mouse moved

src/render/Renderer.cpp
```diff
@@ -2044,7 +2045,10 @@ void CHyprRenderer::ensureCursorRenderingMode() {
     if (*PCURSORTIMEOUT > 0)
         m_cursorHiddenConditions.hiddenOnTimeout = *PCURSORTIMEOUT < g_pInputManager->m_lastCursorMovement.getSeconds();
 
-    const bool HIDE = m_cursorHiddenConditions.hiddenOnTimeout || m_cursorHiddenConditions.hiddenOnTouch || m_cursorHiddenConditions.hiddenOnKeyboard;
+    bool HIDE = m_cursorHiddenConditions.hiddenOnTimeout || m_cursorHiddenConditions.hiddenOnTouch || m_cursorHiddenConditions.hiddenOnKeyboard;
```
vaxerski: can be constified

vaxerski (approved):
lgtm wiki mr needed

PR #11076, opened 2025-07-18
fix: support pipes for wlr-gamma-protocol

MrFantOlas:
This merge request adds support for pipes and potentially other valid file descriptors. It also adds check for more data on the socket than the required amount as per protocol.

This has been tested using my own client for pipes and `wlsunset` with different parameters including a small transition of 30 seconds.

### DIFF
--- a/src/protocols/GammaControl.cpp
+++ b/src/protocols/GammaControl.cpp
@@ -73,14 +73,14 @@ CGammaControl::CGammaControl(SP<CZwlrGammaControlV1> resource_, wl_resource* out
 
         ssize_t readBytes = read(gammaFd.get(), m_gammaTable.data(), m_gammaTable.size() * sizeof(uint16_t));
 
-        ssize_t moreBytes;
+        ssize_t moreBytes = 0;
         {
-            const size_t BUF_SIZE = 1;
-            char         buf[BUF_SIZE];
-            moreBytes = read(gammaFd.get(), buf, BUF_SIZE);
+            const size_t BUF_SIZE      = 1;
+            char         buf[BUF_SIZE] = {};
+            moreBytes                  = read(gammaFd.get(), buf, BUF_SIZE);
         }
 
-        if (readBytes < 0 || (size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t), moreBytes != 0) {
+        if (readBytes < 0 || (size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t) || moreBytes != 0) {
             LOGM(ERR, "Failed to read bytes");
 
             if ((size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t), moreBytes > 0) {
--- a/src/protocols/GammaControl.cpp
+++ b/src/protocols/GammaControl.cpp
@@ -71,11 +71,19 @@ CGammaControl::CGammaControl(SP<CZwlrGammaControlV1> resource_, wl_resource* out
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/protocols/GammaControl.cpp
```diff
@@ -71,11 +71,19 @@ CGammaControl::CGammaControl(SP<CZwlrGammaControlV1> resource_, wl_resource* out
             return;
         }
 
-        ssize_t readBytes = pread(gammaFd.get(), m_gammaTable.data(), m_gammaTable.size() * sizeof(uint16_t), 0);
-        if (readBytes < 0 || (size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t)) {
+        ssize_t readBytes = read(gammaFd.get(), m_gammaTable.data(), m_gammaTable.size() * sizeof(uint16_t));
+
+        ssize_t moreBytes;
```
vaxerski: don't leave uninitialized primitives

src/protocols/GammaControl.cpp
```diff
@@ -71,11 +71,19 @@ CGammaControl::CGammaControl(SP<CZwlrGammaControlV1> resource_, wl_resource* out
...
+        {
+            const size_t BUF_SIZE = 1;
+            char         buf[BUF_SIZE];
+            moreBytes = read(gammaFd.get(), buf, BUF_SIZE);
+        }
+
+        if (readBytes < 0 || (size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t), moreBytes != 0) {
```
vaxerski: what the heck is this syntax? What's `, moreBytes != 0`? I've never seen this comma syntax.

MrFantOlas: Sorry I have been writing in other languages, I am just surprised it did not error out on compilation. Like wtf.

MrFantOlas: Okay so it is just interpreted as, from cppreference:

```cpp
// comma operators can be chained; the result of the last
    // (rightmost) expression is the result of the whole chain:
    int n = 1;
    int m = (++n, std::cout << "n = " << n << '\n', ++n, 2 * n);
```

Thus only `moreBytes != 0` is returned for the *whole* condition.

PS: I had no idea this is actually a valid statement.

vaxerski: neither did I, but it's kinda neat.

src/protocols/GammaControl.cpp
```diff
@@ -71,11 +71,19 @@ CGammaControl::CGammaControl(SP<CZwlrGammaControlV1> resource_, wl_resource* out
...
+            const size_t BUF_SIZE      = 1;
+            char         buf[BUF_SIZE] = {};
+            moreBytes                  = read(gammaFd.get(), buf, BUF_SIZE);
+        }
+
+        if (readBytes < 0 || (size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t) || moreBytes != 0) {
             LOGM(ERR, "Failed to read bytes");
 
-            if ((size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t)) {
+            if ((size_t)readBytes != m_gammaTable.size() * sizeof(uint16_t), moreBytes > 0) {
```
vaxerski: it remains here btw

MrFantOlas:
I am not the most familiar with GitHub. Should I squash the changes or is that done automatically ?

vaxerski:
I'll squash them

vaxerski (approved):

Fervv268 (approved):

PR #11084, opened 2025-07-18
renderer: implement wp-color-management-v1 transfer functions

Tom94:

Implements all remaining transfer functions from wp-color-management-v1 and their inverses.

(Also cleans up trailing whitespaces, mixed tabs and spaces, and a redundant alpha mult+div in `CM.glsl`.)

The implementation is based on the primary source (the [ITU H.273 standard](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.1361-0-199802-W!!PDF-E.pdf)) as well as my own mathematical derivations of some of the constants and inverses which the standard does not provide.

I'm the author of an [image viewer](https://github.com/Tom94/tev) that encounters most of these transfer functions in the wild on a daily basis, so I'm reasonably confident in the correctness.

I also wrote a debug shader (not included in this PR) that verified that the GLSL implementation has correct rountrip behavior (i.e. `fromLinear(toLinear(c)) == c`) for all transfer functions.

Ready for merging.

### DIFF
--- a/src/protocols/ColorManagement.cpp
+++ b/src/protocols/ColorManagement.cpp
@@ -546,6 +546,13 @@ CColorManagementParametricCreator::CColorManagementParametricCreator(SP<CWpImage
             case WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_GAMMA22: break;
             case WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_GAMMA28: break;
             case WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_HLG: break;
+            case WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_BT1886: break;
+            case WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_ST240: break;
+            case WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_LOG_100: break;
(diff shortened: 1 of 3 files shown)


vaxerski:
@UjinT34 for hdr, if you could take a look

UjinT34:
`setSetTfNamed` needs an update with a list of new tfs. Keep all of them listed explicitly because proto enum can be extended as a backward compatible change without creating a new major version.

Tom94:
Done! I also flagged all xx-color-manager-v4 tfs as supported while I was at it. Looks like they are all already mapped to the respectively equivalent color-management-v1 tfs.

Should I also have a look at frog tfs? Their mapping to color-management-v1 seems a bit less trivial than the xx- protocol, so I held off on that for now.

UjinT34:
xx and frog are for removal, shouldn't be used by any recent client.

lgtm

vaxerski (approved):

Tom94:
Please wait with merging. I just noticed that Hyprland's existing implementation of the HLG transfer function is incorrect -- inverse and forward appear to be mistakenly swapped. Going to push a fix in a moment.

Tom94:
Should be all good now. Sorry for not catching this earlier.

vaxerski (approved):

PR #11101, opened 2025-07-19
Add example config for home-manager

auipga:
This PR adds the Nix/Home-Manager equivalent to the [example/hyprland.conf](https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.conf) file. The result of this nix file is identical.
___

The only changes I made are 3 more URLs to the Nix section in the Wiki
https://github.com/hyprwm/Hyprland/blob/0bbb2fc1387444db344c9ccb93d287942ae007b3/example/hyprland-hm.nix#L7-L9
and a todo
https://github.com/hyprwm/Hyprland/blob/0bbb2fc1387444db344c9ccb93d287942ae007b3/example/hyprland-hm.nix#L14-L17

### DIFF
--- a/example/hyprland-hm.nix
+++ b/example/hyprland-hm.nix
@@ -96,10 +96,8 @@
 
       border_size = 2;
 
-      col = {
-        active_border = "rgba(33ccffee) rgba(00ff99ee) 45deg";
-        inactive_border = "rgba(595959aa)";
-      };
+      "col.active_border" = "rgba(33ccffee) rgba(00ff99ee) 45deg";
+      "col.inactive_border" = "rgba(595959aa)";
 
       # Set to true enable resizing windows by clicking and dragging on borders and gaps
       resize_on_border = false;
--- a/example/hyprland-hm.nix
+++ b/example/hyprland-hm.nix
@@ -1,0 +1,339 @@
+{
+  wayland.windowManager.hyprland.settings = {
+
+    # This is an example Hyprland config file for Nix.
+    # Refer to the wiki for more information.
+    # https://wiki.hypr.land/Configuring/
+    # https://wiki.hypr.land/Nix/
+    # https://wiki.hypr.land/Nix/Hyprland-on-NixOS/
+    # https://wiki.hypr.land/Nix/Hyprland-on-Home-Manager/
+
+    # Please note not all available settings / options are set here.
+    # For a full list, see the wiki
+
+    # You can split this configuration into multiple files
+    # Create your files separately and then link them to this file like this:
+    # source = ~/.config/hypr/myColors.conf
+    # todo: make the line above nix-ish
+
+
+    ################
+    ### MONITORS ###
+    ################
+
+    # See https://wiki.hypr.land/Configuring/Monitors/
+    monitor = ",preferred,auto,auto";
+
+
+    ###################
+    ### MY PROGRAMS ###
+    ###################
+
+    # See https://wiki.hypr.land/Configuring/Keywords/
+
+    # Set programs that you use
+
+    "$terminal" = "kitty";
+    "$fileManager" = "dolphin";
+    "$menu" = "wofi --show drun";
+
+
+    #################
+    ### AUTOSTART ###
+    #################
+
+    # Autostart necessary processes (like notifications daemons, status bars, etc.)
+    # Or execute your favorite apps at launch like this:
+
+    # "exec-once" = [
+    #   "$terminal"
+    #   "nm-applet &"
+    #   "waybar & hyprpaper & firefox"
+    # ];
+
+
+    #############################
+    ### ENVIRONMENT VARIABLES ###
+    #############################
+
+    # See https://wiki.hypr.land/Configuring/Environment-variables/
+
+    env = [
+      "XCURSOR_SIZE,24"
+      "HYPRCURSOR_SIZE,24"
+    ];
+
+
+    ###################
+    ### PERMISSIONS ###
+    ###################
+
+    # See https://wiki.hypr.land/Configuring/Permissions/
+    # Please note permission changes here require a Hyprland restart and are not applied on-the-fly
+    # for security reasons
+
+    # ecosystem = {
+    #   enforce_permissions = 1;
+    # };
+
+    # permission = [
+    #   "/usr/(bin|local/bin)/grim, screencopy, allow"
+    #   "/usr/(lib|libexec|lib64)/xdg-desktop-portal-hyprland, screencopy, allow"
+    #   "/usr/(bin|local/bin)/hyprpm, plugin, allow"
+    # ];
+
+
+    #####################
+    ### LOOK AND FEEL ###
+    #####################
+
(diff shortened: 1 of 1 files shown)


example/hyprland-hm.nix
```diff
@@ -0,0 +1,337 @@
...
+        "layers, 1, 3.81, easeOutQuint"
+        "layersIn, 1, 4, easeOutQuint, fade"
+        "layersOut, 1, 1.5, linear, fade"
+        "fadeLayersIn, 1, 1.79, almostLinear"
+        "fadeLayersOut, 1, 1.39, almostLinear"
+        "workspaces, 1, 1.94, almostLinear, fade"
+        "workspacesIn, 1, 1.21, almostLinear, fade"
+        "workspacesOut, 1, 1.94, almostLinear, fade"
+        "zoomFactor, 1, 7, quick"
+      ];
```
fxzzi, line 169: These should be moved out of the animations section. the current hm generator toHyprconf probably allows for this but if they ever switch to toHyprlang (like the hyprland flake module does) it won't work in its current state.

example/hyprland-hm.nix
```diff
@@ -0,0 +1,337 @@
...
+    gestures = {
+      workspace_swipe = false;
+    };
+
+    # Example per-device config
+    # See https://wiki.hypr.land/Configuring/Keywords/#per-device-input-configs for more
+    device = {
+      name = "epic-mouse-v1";
+      sensitivity = -0.5;
+    };
```
fxzzi, line 235: should be `"device[epic-mouse-v1]"` and then remove the name value

example/hyprland-hm.nix
```diff
@@ -0,0 +1,337 @@
...
+
+    # ecosystem = {
+    #   enforce_permissions = 1;
+    # };
+
+    # permission = [
+    #   "/usr/(bin|local/bin)/grim, screencopy, allow"
+    #   "/usr/(lib|libexec|lib64)/xdg-desktop-portal-hyprland, screencopy, allow"
+    #   "/usr/(bin|local/bin)/hyprpm, plugin, allow"
+    # ];
```
fxzzi, line 83: maybe update these to follow the nix paths instead? Like was shown here: https://github.com/hyprwm/Hyprland/pull/9939/files#diff-e0355b04e7febb871fc670d6ca39f7041e2f0cb6ff92afa5c48028ccfc77f153R79-R82

example/hyprland-hm.nix
```diff
@@ -0,0 +1,337 @@
...
+    #################
+
+    # Autostart necessary processes (like notifications daemons, status bars, etc.)
+    # Or execute your favorite apps at launch like this:
+
+    # "exec-once" = [
+    #   "$terminal"
+    #   "nm-applet &"
+    #   "waybar & hyprpaper & firefox"
+    # ];
```
fxzzi, line 52: `programs.nm-applet.enable = true`?

Also, does the `&` on that line do anything anyway? 

and hyprpaper too is enabled through hm

example/hyprland-hm.nix
```diff
@@ -0,0 +1,337 @@
...
+    # https://wiki.hypr.land/Nix/
+    # https://wiki.hypr.land/Nix/Hyprland-on-NixOS/
+    # https://wiki.hypr.land/Nix/Hyprland-on-Home-Manager/
+
+    # Please note not all available settings / options are set here.
+    # For a full list, see the wiki
+
+    # You can split this configuration into multiple files
+    # Create your files separately and then link them to this file like this:
+    # source = ~/.config/hypr/myColors.conf
```
fxzzi, line 16:
```nix
source = [
   "~/.config/hypr/myColors.conf"
];
```

example/hyprland-hm.nix
```diff
@@ -0,0 +1,337 @@
...
+    ### MY PROGRAMS ###
+    ###################
+
+    # See https://wiki.hypr.land/Configuring/Keywords/
+
+    # Set programs that you use
+
+    "$terminal" = "kitty";
+    "$fileManager" = "dolphin";
+    "$menu" = "wofi --show drun";
```
fxzzi, line 38: to ensure these are installed with the config, maybe do

```nix
"$terminal" = "${lib.getExe pkgs.kitty}"
```

and so on and so forth?

fxzzi (commented):
my general feedback but you should wait for confirmation from someone more worthy (xd) to comment too

example/hyprland-hm.nix
auipga: Great, that explains how to define more than one device in Nix 👍

example/hyprland-hm.nix
auipga: I didn't ponder it. I just blindly copied it from the official conf file. Even though I think the ambersand shouldn't be necessary I don't know the internals so maybe it is required.

example/hyprland-hm.nix
fufexan: Can be a string as well, as long as it's a single path. Though if we're already using Nix and the output file doesn't matter that much, we can just do `imports` with multiple Nix files modifying a different part of the config.

example/hyprland-hm.nix
fufexan: I wouldn't add nm-applet, waybar or hyprpaper there. They're started by systemd. So no need for the `programs..` either.

PointerDilemma:
One reason for not adding this would be that the converter is not perfect and probably won't ever be. This is one of the things I don't like as a nix user. Insisting on nix-code configuration everywhere.

For example there is `binds {}` category and `binds =` keyword in hyprland.

example/hyprland-hm.nix
fxzzi: I usually think of source for nix as being a way to import non-declaritive files into the configuration. Like pywal / wallust generated colours etc.

Yeah it can just be a string too but imo makes more sense to show it as a list so users know that they can just list more

fxzzi:
binds section can be configured with

```nix
binds = {
    foo = true;
};
OR
binds.foo = true;
```

binds keyword can be used via a list

```nix
binds = [
    "foo,bar,idc"
];
```

To me this issue stems more from Hyprlang as a language than Nix

PointerDilemma:
Yeah but you can't use both, because nix doesn't allow duplicate keys.
So one wants to set `binds.scroll_event_delay`, but already has a `binds = []` section, it won't work in nix.

fxzzi:
The keyword is actually `bind` right? And then the section is `binds`

C0Florent:
> The keyword is actually `bind` right? And then the section is `binds`

`binds` could be a keyword, actually all keywords that match `/bind[lrgcoenmtisdp]*/` will be recognised as bind keywords, which includes `binds`

fxzzi:
Oh yeah you're right, I get it, forgot about those extras

example/hyprland-hm.nix
fufexan: Right, that makes sense. I guess we could have both an `imports` and a `source` example with respective comments.

PR #11087, opened 2025-07-18
syncobj: use rendernode for timelines

gulafaran:
use rendernode for timelines instead of the drmfd, some devices dont support to use the drmfd for this.

requires: https://github.com/hyprwm/aquamarine/pull/193

fixes: #10401

PR #11117, opened 2025-07-19
drm: check syncobj timeline support before advertising protocol

mavonarx:

Fixes crashes in Vulkan applications (Zed, WezTerm, etc.) on systems where the DRM driver doesn't support syncobj timelines, particularly Apple Silicon devices running Asahi Linux with the Honeykrisp driver.

The issue was that Hyprland advertised the `linux-drm-syncobj-v1` protocol regardless of hardware support. When applications attempted to use explicit sync, they would crash with "Timeline failed importing" errors.

This PR adds `DRM_CAP_SYNCOBJ_TIMELINE` capability checking at two levels:
- Protocol creation: Only advertise the protocol when hardware supports it
- Timeline creation: Additional safety check when creating sync timelines

No breaking compatibility - this only prevents advertising a protocol when the underlying hardware doesn't support it. Applications gracefully fall back to implicit sync when explicit sync isn't available.

Ready for merging. Tested on Fedora Asahi Remix with Apple M1 Pro. Applications that previously crashed now start successfully and run normally.


vaxerski:
this MR is based on some old rev, needs to be rebased on top of main, that's first.

Second - Asahi and Apple Silicon support explicit sync - just not on the display node, and Tom is working on that in https://github.com/hyprwm/Hyprland/pull/11087

mavonarx:
Rebased and retested

src/helpers/sync/SyncTimeline.cpp
```diff
@@ -6,7 +6,16 @@
 #include <sys/eventfd.h>
 using namespace Hyprutils::OS;
 
+static bool checkDrmSyncobjTimelineSupport(int drmFD) {
+    uint64_t cap = 0;
+    int ret = drmGetCap(drmFD, DRM_CAP_SYNCOBJ_TIMELINE, &cap);
+    return (ret == 0 && cap != 0);
```
vaxerski: this should probably be static. It won't change, and I don't know how expensive drmGetCap is. Furthermore, you're duplicating this here and in protocolmanager. Maybe add a bool to CCompositor that is set in `::initServer`?

mavonarx:
Fixed - moved the capability check to  init and removed the duplicate functions. Retested again - works.

src/managers/ProtocolManager.cpp
```diff
@@ -210,8 +211,14 @@ CProtocolManager::CProtocolManager() {
         else
             lease.reset();
 
-        if (g_pHyprOpenGL->m_exts.EGL_ANDROID_native_fence_sync_ext && !PROTO::sync)
-            PROTO::sync = makeUnique<CDRMSyncobjProtocol>(&wp_linux_drm_syncobj_manager_v1_interface, 1, "DRMSyncobj");
+        if (g_pHyprOpenGL->m_exts.EGL_ANDROID_native_fence_sync_ext && !PROTO::sync) {
+            if (g_pCompositor->supportsDrmSyncobjTimeline()) {
+                PROTO::sync = makeUnique<CDRMSyncobjProtocol>(&wp_linux_drm_syncobj_manager_v1_interface, 1, "DRMSyncobj");
+                Debug::log(LOG, "DRM Syncobj Timeline support detected, enabling explicit sync protocol");
+            } else {
```
vaxerski: style nit: no {}

vaxerski:
c-f failed, run c-f pls

src/Compositor.hpp
```diff
@@ -152,7 +152,11 @@ class CCompositor {
     NColorManagement::SImageDescription getPreferredImageDescription();
     bool                                shouldChangePreferredImageDescription();
 
-    std::string                         m_explicitConfigPath;
+    bool                                supportsDrmSyncobjTimeline() const {
+        return m_bDrmSyncobjTimelineSupported;
```
vaxerski: this was unnecessary, please keep function bodies outside of headers

vaxerski (approved):

PR #11128, opened 2025-07-20
Fix: Replace unsafe strcpy with snprintf

Mirza-Samad-Ahmed-Baig:
Replaced unsafe strcpy calls with snprintf to prevent potential buffer overflows and improve code safety.

Breaking nothing.

Ready to merge.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -107,8 +107,7 @@ static int openRenderNode(int drmFd) {
             Debug::log(LOG, "DRM dev versionName", render_version->name);
             if (strcmp(render_version->name, "evdi") == 0) {
                 free(renderName);
-                renderName = (char*)malloc(sizeof(char) * 15);
-                snprintf(renderName, 15, "%s", "/dev/dri/card0");
+                renderName = strdup("/dev/dri/card0");
             }
             drmFreeVersion(render_version);
         }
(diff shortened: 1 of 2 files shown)


abihf:
why not `strncpy`?

Mirza-Samad-Ahmed-Baig:
I chose snprintf over strncpy because it's safer and more predictable — it always null-terminates, avoids unnecessary padding, and handles formatting cleanly. In contrast, strncpy can leave the string unterminated if the source length equals or exceeds the buffer size, which is dangerous especially when working with socket paths or system interfaces.

src/render/OpenGL.cpp
```diff
@@ -108,7 +108,7 @@ static int openRenderNode(int drmFd) {
             if (strcmp(render_version->name, "evdi") == 0) {
                 free(renderName);
                 renderName = (char*)malloc(sizeof(char) * 15);
-                strcpy(renderName, "/dev/dri/card0");
+                snprintf(renderName, 15, "%s", "/dev/dri/card0");
```
C0Florent: Use `strdup` instead of `malloc`+`snprintf`?

As to not duplicate the size of the string, and also more readable; properly allocates, null-terminates, and the inner strlen call will certainly inlined, so there's no overhead for it

Mirza-Samad-Ahmed-Baig: Good catch, thank you. Done with it.

vaxerski (approved):

PR #11132, opened 2025-07-20
example: make screen shader example compatible with glsl 300 (#10846)

00-KAMIDUKI:
Original example causes an error.

ready for merging


gulafaran:
while in there you can change to

```
vec4 pixColor = texture(tex, v_texcoord);
```

since texture2D is a deprecated function in gles3 and above

PR #11161, opened 2025-07-22
Named persistent workspaces: ensures unique identifiers

arnaud-ma:

In this calll: https://github.com/hyprwm/Hyprland/blob/50758505d5c784052437a371a707fc2dc60bb34a/src/Compositor.cpp#L3101-L3107

All named persistent rules had the exact same ID, so only the first one was created. See #11146. This PR fixes this by checking ID uniqueness not only across existing workspaces, but also across persistent ones (which might not yet exist).

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1719,6 +1719,13 @@ WORKSPACEID CCompositor::getNextAvailableNamedWorkspace() {
             lowest = w->m_id;
     }
 
+    // Give priority to persistent workspaces to avoid any conflicts between them.
+    for (auto const& rule : g_pConfigManager->getAllWorkspaceRules()) {
+        if (rule.isPersistent && rule.workspaceId < -1 && rule.workspaceId < lowest) {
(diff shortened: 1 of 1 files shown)


src/Compositor.cpp
```diff
@@ -1719,6 +1719,13 @@ WORKSPACEID CCompositor::getNextAvailableNamedWorkspace() {
             lowest = w->m_id;
     }
 
+    // Give priority to persistent workspaces to avoid any conflicts between them.
+    for (auto const& rule : g_pConfigManager->getAllWorkspaceRules()) {
+        if (rule.isPersistent && rule.workspaceId < -1 && rule.workspaceId < lowest) {
```
vaxerski: no {} and can you use a guard instead

vaxerski (dismissed):
lgtm, just misses tests for this to not happen again.

arnaud-ma:
Oooops i did a mistake with the rebase.. sorry :(

PR #11169, opened 2025-07-22
fix wrong animations in setGroupCurrent

MightyPlaza:
on `setGroupCurrent` for fullscreen / maximized:
- matches the windows alpha values
- don't animate from the inactive border color
- make the first frame have the proper window texture size (don't recalculate the monitor before turning fullscreen back on)

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1862,11 +1862,13 @@ void CCompositor::updateWindowAnimatedDecorationValues(PHLWINDOW pWindow) {
     auto* const GROUPINACTIVELOCKEDCOL = (CGradientValueData*)(PGROUPINACTIVELOCKEDCOL.ptr())->getData();
 
     auto        setBorderColor = [&](CGradientValueData grad) -> void {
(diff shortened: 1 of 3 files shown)


vaxerski:
c-f fails

MightyPlaza:
fixed

vaxerski (approved):

vaxerski:
just misses some tests. I think all of the bugs could be tested via having a dispatcher in the plugin to query props like the texture size color etc

MightyPlaza:
for the first 2 commits I don't think it's worth it, they have been issues for as long as I can remember, and are not really prone to accidental breakage
for the 3rd one it is probably a good idea, specially since it still happens for floating groups and [other cases](https://github.com/hyprwm/Hyprland/issues/9057), however I've only patched the problem for a specific case here
I still have no idea how does calling `recalculateMonitor()` with a wrong size and then the right size cause the first frame to be drawn with the wrong size (it doesn't help I'm not familiar with the render loop)

PR #11204, opened 2025-07-24
add support for focus_on_close

### LINKED ISSUE #10655: Add a history mode to focus_on_close
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10247

## Description
Add an option to `focus_on_close` to focus the last window from the history

tgys:
this adds a new config option master:focus_on_close and sets a parameter in CCompositor::focusWindow which sets a member variable that gets checked again in CCompositor::closeWindow.
fixes #10655


Vladimir-csp:
Is this only for master, not dwindle? Same name as `input:focus_on_close`.

PR #11205, opened 2025-07-24
virtualkeyboard: do not sent release event for pressed key during destroy

xzn:

The current behavior breaks when fcitx5 (which uses input method v2 and virtual keyboard) is running, leading to issues in third-party applications like this one:

https://github.com/H3rmt/hyprshell/issues/294

More discussion here:

https://github.com/fcitx/fcitx5/pull/1391

(I previously reported the issue for hyprland here: https://github.com/hyprwm/Hyprland/discussions/11186 ; after discussion with fcitx5 maintainer it is believed that hyprland shouldn't send release events on virtual keyboard destroy)

### DIFF
--- a/src/protocols/VirtualKeyboard.cpp
+++ b/src/protocols/VirtualKeyboard.cpp
@@ -9,12 +9,10 @@ CVirtualKeyboardV1Resource::CVirtualKeyboardV1Resource(SP<CZwpVirtualKeyboardV1>
         return;
 
     m_resource->setDestroy([this](CZwpVirtualKeyboardV1* r) {
-        releasePressed();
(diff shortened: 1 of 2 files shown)


xzn:
I did the test on sway with `wev -f wl_keyboard` to see key release event after moving cursor into the window while having a key pressed (with window focus follow mouse), and it seems to behave the same as on hyprland. Now I'm wondering if this really should be fixed from fcitx5 after all.

If there's a way to have fcitx5's virtual keyboard's modifier state separated from physical, then enabling persistent virtual keyboard in fcitx5 may be a workaround.

Edit:

Maybe this should be configurable with default the old behavior for compat?

PR #11206, opened 2025-07-24
input: do not accumulate modifiers from all keyboards for binds and seats

xzn:

The current behavior breaks when fcitx5 (which uses input method v2 and virtual keyboard) is running, specifically when persistent virtual keyboard is enabled in fcitx5.

Manifested issues include modifier keys get stuck when modifier keys are held during window focus switching, (e.g. from fcitx5 active window to fcitx5 non-active window)

Possibly related:

https://github.com/hyprwm/Hyprland/discussions/10360
https://github.com/hyprwm/Hyprland/discussions/10967
https://github.com/hyprwm/Hyprland/discussions/11034

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -457,7 +457,7 @@ bool CKeybindManager::onKeyEvent(std::any event, SP<IKeyboard> pKeyboard) {
     if (handleInternalKeybinds(internalKeysym))
         return false;
 
-    const auto MODS = g_pInputManager->accumulateModsFromAllKBs();
+    const auto MODS = g_pInputManager->getLastKeyboardMods();
 
     m_timeLastMs    = e.timeMs;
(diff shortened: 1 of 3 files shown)


vaxerski:
no good. Some keyboards are split and show up as separate devices, sometimes you have split numpads, etc. Some PCBs split the keyboards for n-key rollover. You'll break a lot of keybinds.

xzn:
Would it make sense to introduce something like wlroots' keyboard group? May need change to aquamarine as well (for libinput keyboard group detection).

Edit:

Alternatively would it make sense to check only physical keyboards for the merged modifier state? So each virtual send their state unmerged, while physical ones send merged state of all physicals.

Edit 2:

Feel free to correct me: hyprland currently put all input devices in a single wl_seat, and wl_keyboard wants no repeated presses/released events in succession. I'll try to implement something to follow that effect.

PR #11214, opened 2025-07-25
virtualkeyboard: Add options to skip releasing pressed keys on close and to skip sharing key states

xzn:

Supersedes https://github.com/hyprwm/Hyprland/pull/11206 and https://github.com/hyprwm/Hyprland/pull/11205

Added new config options:

- `input:virtualkeyboard:share_states`
- - Whether virtual keyboards will share their key down and mod states. (Default to **false**, whereas current behavior is the same as **true**)
- `input:virtualkeyboard:release_pressed_on_close`
- - Whether virtual keyboards will release pressed keys automatically on close. (Default to **false**, whereas current behavior is the same as **true**)

Per device option for virtual keyboards available. Default chosen to hopefully make compat with fcitx5 better ootb.

- `misc:name_vk_after_proc`
- - Whether to name virtual keyboards after the processes that create them, so each virtual keyboard created by different process will have their own name. E.g. `/usr/bin/fcitx5` created virtual keyboards will have name `hl-virtual-keyboard-fcitx5`. (Default to **true**, whereas current behavior is the same as **false**)

Default chosen to make per device option usable by default.

`accumulateModsFromAllKBs` removed, may affect plugins.

Key down state for all seats (there's really only one) cached in InputManager. Virtual keyboard with `input:virtualkeyboard:share_states` set to false won't have their key down and mod states shared with other keyboards. Other keyboards will still share theirs with the virtual keyboard however, to make behavior sensible..

InputManager now suppresses repeated pressed/released events from multiple devices (including virtual ones) when sending events to seats, to match the spec.

I only tested this in `wev` with multiple separate keyboards. It's be great to have people with actual keyboard that uses multiple logical devices for n-key rollover to test this as well. Though the logic I changed is quite simple and should have no problem.

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -97,7 +97,7 @@ static SDispatchResult vkb(std::string in) {
     g_pInputManager->newKeyboard(tkb1);
     g_pInputManager->newKeyboard(vkb0);
 
-    CScopeGuard    x([] {
+    CScopeGuard    x([&] {
         tkb0->destroy();
         tkb1->destroy();
         vkb0->destroy();


src/protocols/VirtualKeyboard.cpp
```diff
@@ -89,6 +85,24 @@ CVirtualKeyboardV1Resource::CVirtualKeyboardV1Resource(SP<CZwpVirtualKeyboardV1>
...
+    if (CLIENTNAME.has_value()) {
+        const auto PATH = std::filesystem::path(CLIENTNAME.value());
+        if (PATH.has_filename()) {
+            const auto FILENAME = PATH.filename();
+            const auto NAME     = deviceNameToInternalString(FILENAME);
+            m_name += NAME;
+            return;
+        }
+    }
+    m_name += "unknown";
```
vaxerski: style-wise, I'd put the name getting into a separate function, for clarity's sake

src/protocols/VirtualKeyboard.cpp
```diff
@@ -89,6 +85,24 @@ CVirtualKeyboardV1Resource::CVirtualKeyboardV1Resource(SP<CZwpVirtualKeyboardV1>
     });
 
     m_name = "hl-virtual-keyboard";
+
+    static auto PVKNAMEPROC = CConfigValue<Hyprlang::INT>("misc:name_vk_after_proc");
+    if (!*PVKNAMEPROC) {
```
vaxerski: no {}

src/devices/IKeyboard.cpp
```diff
@@ -405,3 +392,27 @@ void IKeyboard::updateXkbStateWithKey(uint32_t xkbKey, bool pressed) {
...
+    if (contains)
+        std::erase(m_pressed, key);
+    else
+        m_pressed.emplace_back(key);
+
+    return true;
+}
+
+bool IKeyboard::getPressed(uint32_t key) {
+    return std::ranges::find(m_pressed, key) != m_pressed.end();
```
vaxerski: std::ranges::contains :)

src/managers/input/InputManager.cpp
```diff
@@ -1398,18 +1399,37 @@ void CInputManager::onKeyboardKey(const IKeyboard::SKeyEvent& event, SP<IKeyboar
...
+
     const auto EMAP = std::unordered_map<std::string, std::any>{{"keyboard", pKeyboard}, {"event", event}};
     EMIT_HOOK_EVENT_CANCELLABLE("keyPress", EMAP);
 
-    bool passEvent = DISALLOWACTION || g_pKeybindManager->onKeyEvent(event, pKeyboard);
+    bool passEvent = DISALLOWACTION;
 
-    if (passEvent) {
-        const auto IME = m_relay.m_inputMethod.lock();
+    if (!DISALLOWACTION) {
```
vaxerski: no {}

xzn:
Updated SeatManager to send wl_array of pressed keys (now that we have them) in wl_keyboard::enter.

For style update I've changed a few other instances of `std::ranges::find` to `std::ranges::contains` also. It's a newish addition in C++ and previous code were using `find`. Using `contains` causes no issue on my machine (up-to-date Arch Linux) though so it should be good.

I'll write the tests in 2 days maybe when I have more time :>

vaxerski:
alright, thanks!

xzn:
Added test, it's quite barebone. Lmk if you have any suggestions.

src/devices/IKeyboard.hpp
```diff
@@ -118,5 +123,9 @@ class IKeyboard : public IHID {
   private:
     void                  clearManuallyAllocd();
 
-    std::vector<uint32_t> m_pressedXKB;
+    std::vector<uint32_t> m_pressed;
+
+  protected:
+    bool updatePressed(uint32_t key, bool pressed);
+    bool m_share_states = true;
```
vaxerski: bad naming: `m_shareStates`

hyprtester/plugin/src/main.cpp
```diff
@@ -48,11 +49,85 @@ static SDispatchResult snapMove(std::string in) {
...
+    vkb0->sendKey(TESTKEY, true);
+    tkb1->sendKey(TESTKEY, false);
+    if (std::ranges::contains(PRESSED, TESTKEY)) {
+        return {
+            .success = false,
+            .error   = "Expected released key found in pressed (vkb no share state)",
+        };
+    }
+
+    return {};
```
vaxerski, line 135: if you could remove the keyboards here after the tests are done to end on a clean state

vaxerski (commented):
generally lgtm

PR #11216, opened 2025-07-25
texture: avoid issuing many glTexSubImage2D calls

gulafaran:
applications that use QtWidgets like dolphin adds each folder item or individual file to the damage region and gives us a lot of different regions.

issuing multiple glTexSubImage2D calls is slow but on the other hand smaller regions can still be faster by simply calling glTexSubImage2D on each of them.

the tradeoff is uploading the extent region a few pixels that perhaps wasnt damaged yet again. but should be fine doing so.

so if case the rects size and pack the buffer extents and upload it in one call, and if few rects do it as before.


gulafaran:
breaks dbeaver, il close this for now until its actually profiled as a slow thing.

PR #11220, opened 2025-07-25
monitor: makes visibleOnMonitor calculation include decorations

jmanc3:
Fixes: https://github.com/hyprwm/Hyprland/discussions/11203

The window turned invisible when just outside the monitor bounds, even though it should have stayed visible given its decorations.

The fix was to include the decorations when determining if a window is on a monitor.


vaxerski (commented):
I think we should _only_ do that for floating windows. If someone sets shadows to be large, I don't think them overflowing to another monitor (when tiled) would be pleasant.

jmanc3:
I hate github so much

PR #11228, opened 2025-07-26
Solitary clients with single subsurface & verbose solitary/tearing/DS checks

### LINKED ISSUE #10452: Allow tearing / scanout with subsurface-based renderers
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10396

## Description
Some apps, notably browsers, but apparently also winewayland, use a single, full-frame subsurface to render their content.

This will not allow tearing / scanout to happen as surface number is checked to be 1.

We should probably check for this case, and allow both if we have a single, opaque, full-frame subsurface.

UjinT34:
Moved solitary/tearing/DS checks to `CMonitor`.
Added enums to describe possible blocking reasons for those features.
Added `solitaryBlockedBy`, `tearingBlockedBy` and `directScanoutBlockedBy` fields to `hyprctl monitors`.
Allowed clients with single opaque full-sized subsurface to be recognized as solitary clients.

Should help with debugging DS and tearing activation issues and fix some issues with wine-wayland.

Partially fixes #10452. DS activation on my system goes up to this point:
```
[TRACE] [AQ] drm: Using drmModeAddFB2WithModifiers to import buffer into KMS: Size [Vector2D: x: 5120, y: 1440] with format XB24 and mod 216172782128496660
[ERR] [AQ] drm: Failed to submit a buffer with drmModeAddFB2WithModifiers
[ERR] [AQ] drm: Failed to submit a buffer to KMS
```
Maybe some DS step isn't updated with this new solitary logic or it can be an nvidia bug or some [format negotiation](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_EXT_image_drm_format_modifier.html#_usage_patterns) ussue. Needs more testing with different apps and hw.

~~Tearing may report as blocked by next frame not being torn while tearing is active and working because the relevant flag is constantly flipped between frames. Probably should remove this value from the output.~~

DS and tearing checks should keep the current logic. Some of this logic is known to be incorrect, duplicated or might miss some edge cases. This PR will mess with fixes from #10020

Doesn't seem to break anything that was working with DS/tearing. Known to break some apps that couldn't activate DS/tearing without this PR. These features should be turned off manually for such apps and it should be investigated in a separate issue.

Ready


fxzzi:
[hyprlandCrashReport1261.txt](https://github.com/user-attachments/files/21453157/hyprlandCrashReport1261.txt)
[hyprlandCrashReport1681.txt](https://github.com/user-attachments/files/21453158/hyprlandCrashReport1681.txt)

randomly crashing when opening and closing foot windows


tested it anyway, seems tearing is unable to trigger citing "next frame is not torn" in all instances I've seen. Ghost of Tsushima running in wine wayland shown below:

Screenshot from 25 07 27 10:58:01
Screenshot from 25 07 27 10:58:08


Then I tested with CEMU, playing The Legend of Zelda: the Windwaker HD (pretty sure this is fullscreen subsurface)
Tearing was unable to trigger here when I was focused on the game, moving focus to my secondary monitor froze the game, and then tearing went to true (still blocked by "next frame is not torn")

Moving focus back to the game, the game wouldn't recover from being frozen, switching workspace on that monitor to another and back to the game made it continue, with tearing now disabled:

Screenshot from 25 07 27 10:59:19
Screenshot from 25 07 27 11:00:08

UjinT34:
`activelyTearing: true` means that it's active. "next frame is not torn" flag is flipped between frames and `hyprctl` doesn't catch it in active state, ignore this "blocker" for now. Probably will remove it from output later.

Both crash reports point to `CPopup::breadthfirst` called by `CWindow::popupsCount` which wasn't modified. The only difference is that `CWindow::popupsCount` can be called by `hyprctl monitors` now and probably hits some popups during destruction. `n` is a `WP` inside `CPopup::bfHelper` and it isn't checked before accessing `n->m_children`

fxzzi:
hmm, the foot clients I opened were empty actually, I didn't have `hyprctl monitors` running in them, they were just in shell.

Tearing is working on wine-wayland now afaict, disabled VRR on my monitor and could notice tearing in some instances.

However not working on CEMU still with the aforementioned issue of it only triggering when not focussed on it for some reason

UjinT34:
Does it randomly crash or always on close?
Always crashes on close for me. And in a strange way:
```
        # | ./build/Hyprland(_Z12getBacktracev+0x62) [0x563245d1d785]
                getBacktrace()
                /home/ujin/projects/Hyprland/src/helpers/MiscFunctions.cpp:746 (discriminator 2)
        #1 | ./build/Hyprland(_ZN14NCrashReporter18createAndSaveCrashEi+0x869) [0x563245ba3cef]
                NCrashReporter::createAndSaveCrash(int)
                /home/ujin/projects/Hyprland/src/debug/CrashReporter.cpp:200
        #2 | ./build/Hyprland(+0x902fed) [0x563245a78fed]
                handleUnrecoverableSignal(int)
                /home/ujin/projects/Hyprland/src/Compositor.cpp:113
        #3 | /usr/lib64/libc.so.6(+0x43b30) [0x7f35dca43b30]
                ??
                ??:0
        #4 | ./build/Hyprland(_ZNK9Hyprutils6Memory12CWeakPointerI6CPopupE7expiredEv+0x14) [0x563245c2c9ba]
                Hyprutils::Memory::CWeakPointer<CPopup>::expired() const
                /usr/include/hyprutils/memory/WeakPtr.hpp:116
        #5 | ./build/Hyprland(_ZN6CPopup12breadthfirstESt8functionIFvN9Hyprutils6Memory12CWeakPointerIS_EEPvEES5_+0x38) [0x563245c1fcaa]
                CPopup::breadthfirst(std::function<void (Hyprutils::Memory::CWeakPointer<CPopup>, void*)>, void*)
                /home/ujin/projects/Hyprland/src/desktop/Popup.cpp:348 (discriminator 1)
        #6 | ./build/Hyprland(_ZN7CWindow11popupsCountEv+0x78) [0x563245c5f67c]
                CWindow::popupsCount()
                /home/ujin/projects/Hyprland/src/desktop/Window.cpp:1325 (discriminator 6)

```
screenshot-1753614952

fxzzi:
only quitting yea

UjinT34:
Should be fixed now. Happened when `opaque` was called on an unmapped window.

src/debug/HyprCtl.cpp
```diff
@@ -109,6 +110,392 @@ static std::string availableModesForOutput(PHLMONITOR pMonitor, eHyprCtlOutputFo
     return result;
 }
 
+std::string CHyprCtl::getSolitaryBlockedReason(Hyprutils::Memory::CSharedPointer<CMonitor> m, eHyprCtlOutputFormat format) {
+    const auto reasons = m->isSolitaryBlocked(true);
+    if (!reasons)
+        return "null";
+    std::string reasonStr = "";
+    if (format == eHyprCtlOutputFormat::FORMAT_JSON) {
+        if (reasons & CMonitor::SC_UNKNOWN) {
```
vaxerski: I think it would take way less lines and would be much more readable if you made an array of explanations.
```cpp
const std::array<const char*, 10> SOLITARY_NAMES = {
"BLAH",
"BLAH2",
...
};
```

then you can loop over them and use the `i` to get the name, and `& (1 << i)` to check.

src/desktop/Popup.cpp
```diff
@@ -333,8 +333,10 @@ void CPopup::bfHelper(std::vector<WP<CPopup>> const& nodes, std::function<void(W
     nodes2.reserve(nodes.size() * 2);
 
     for (auto const& n : nodes) {
-        for (auto const& c : n->m_children) {
-            nodes2.push_back(c->m_self);
+        if (!n.expired()) {
```
vaxerski: use operator bool here
```cpp
if (!n)
   continue;
```

src/desktop/Window.cpp
```diff
@@ -1860,3 +1862,24 @@ PHLWINDOW CWindow::parent() {
 bool CWindow::priorityFocus() {
     return !m_isX11 && CAsyncDialogBox::isPriorityDialogBox(getPID());
 }
+
+SP<CWLSurfaceResource> CWindow::getSolitaryResource() {
+    if (!m_wlSurface || !m_wlSurface->resource())
```
vaxerski, line 1870: code is suffocating add some newlines please

src/helpers/Monitor.cpp
```diff
@@ -1443,30 +1444,259 @@ void CMonitor::setCTM(const Mat3x3& ctm_) {
...
+    }
 
-    if (!PCANDIDATE)
-        return false;
+    const auto PWORKSPACE = m_activeWorkspace;
+    if (!PWORKSPACE) {
+        reasons |= SC_WORKSPACE;
+        if (!full)
+            return reasons;
+    } else {
```
vaxerski: else can be dropped if you do `if (!full) return reasons;` -> `return reasons;` (readability, this is hell)

src/helpers/Monitor.cpp
```diff
@@ -1443,30 +1444,259 @@ void CMonitor::setCTM(const Mat3x3& ctm_) {
...
+                return reasons;
+        }
+
+        const auto PCANDIDATE = PWORKSPACE->getFullscreenWindow();
+
+        if (!PCANDIDATE) {
+            reasons |= SC_CANDIDATE;
+            if (!full)
+                return reasons;
+        } else {
```
vaxerski: same as above

src/helpers/Monitor.cpp
```diff
@@ -1443,30 +1444,259 @@ void CMonitor::setCTM(const Mat3x3& ctm_) {
...
+    uint8_t     reasons = 0;
+
+    static auto PTEARINGENABLED = CConfigValue<Hyprlang::INT>("general:allow_tearing");
+
+    if (!m_tearingState.nextRenderTorn) {
+        reasons |= TC_NOT_TORN;
+        if (!full)
+            return reasons;
+    }
+    if (!*PTEARINGENABLED) {
```
vaxerski, line 1625: breathing room

vaxerski (commented):
also I'd love a few tests for these.

fxzzi:
> Should be fixed now. Happened when `opaque` was called on an unmapped window.

yep fixed.

more info on the cemu issue, it seems like the tearingBlockedBy is null when the window is frozen if that helps

Screenshot from 25 07 27 17:53:24

if you want to test cemu for free i've provided a basic homebrew pong app which you can run in the emulator without needing any firmware or official games. this can be used to reproduce the issue on your side (run the .elf file in cemu)
[pong.zip](https://github.com/user-attachments/files/21455122/pong.zip)

the fps is unlocked here but still repro'd the issue for me

EDIT: whilst the window is frozen if i move my mouse it unfreezes. freezes again when cursor is still

src/debug/HyprCtl.cpp
UjinT34: yeah. but such arrays should be kept in sync manually. and a proper map won't help much.
those checks will definitely change in near future. iterating up to some magic constant doesn't sound reliable either. I don't like this copy-paste stuff but I don't know how to do it better so it doesn't break with the next commit

src/helpers/Monitor.cpp
UjinT34: that's intentional in case something will be added after the else branch. it's easy to miss this stuff

UjinT34:
This PR doesn't fix DS/tearing in general. Only makes single subsurface case acceptable for the DS/tearing code. And makes it easier to debug. So it'll be obvious when DS/tearing doesn't work due to config settings ("user settings", "window settings"), due to some other intentional internal check or due to something being broken (blocked by "null"). Internal checks and broken stuff are being fixed in #10020
If cemu doesn't freeze with this PR when tearing is disabled then I'd leave it as is since tearing doesn't work for cemu either way. Try `cursor:no_hardware_cursors = 1`, #10020 disables them when tearing is active.

ikalco:
hey if you want to you can cherry pick [this commit](https://github.com/hyprwm/Hyprland/pull/10020/commits/3cf7a3cc7b47c80638418548f7d2297438d6f540) from my old pr #10020,
It seems relevant to this pr and I don't really feel like rebasing mine for the millionth time

UjinT34:
> hey if you want to you can cherry pick [this commit](https://github.com/hyprwm/Hyprland/pull/10020/commits/3cf7a3cc7b47c80638418548f7d2297438d6f540) from my old pr #10020, It seems relevant to this pr and I don't really feel like rebasing mine for the millionth time

that single commit doesn't seem to be enough. and there a lot of changes overall. better to cherry pick/merge/rebase in a separate PR after this one is merged

src/helpers/Monitor.cpp
vaxerski: it's ugly.

LionHeartP:
I tried this PR and Hyprland crashes for me 1-2 seconds in.
[hyprlandCrashReport512.txt](https://github.com/user-attachments/files/21492961/hyprlandCrashReport512.txt)

UjinT34:
> I tried this PR and Hyprland crashes for me 1-2 seconds in. [hyprlandCrashReport512.txt](https://github.com/user-attachments/files/21492961/hyprlandCrashReport512.txt)

fixed

vaxerski (commented):
generally lgtm just tests would be nice

LionHeartP:
I have been testing games with wine-wayland + HDR. 
Tearing works now!
Direct Scanout works now! (There's a noticeable decrease in latency when moving the mouse)

Issue is if you have both enabled.

`hyprctl monitors` wil list Direct Scanout as blocked by tearing which is correct, BUT, the game will be frozen on one random frame. Sometimes it will render a few every few seconds and then stop, sometimes it'll just be a black screen.

Turning either one off, makes the other one work flawlessly.

Bonus fix (I'm not 100% sure it worked before), VRR works flawlessly with the wine-wayland + HDR combo. In the past, it'll work for a bit (like until the main menu of the game I can see the refresh rate adjusting), but when in-game, it would actually be stuck at max refresh rate. VRR worked fine on Xwayland btw, it was only on wine-wayland where this issue existed.

UjinT34:
> generally lgtm just tests would be nice

any docs on how to write and run tests?

vaxerski:
yeah https://wiki.hypr.land/Contributing-and-Debugging/PR-Guidelines/#testing-and-ci

UjinT34:
Some tests might be broken. Couldn't verify them because for some reason I get
```
Monitor HEADLESS-2 (ID 1):
        0x0@60.00000 at 0x0
        ...
        active workspace: -1337 (snap)
        special workspace: 0 ()
        reserved: 0 0 0 0
        scale: 1.00
        transform: 0
        ...
        availableModes: 1920x1080@0.06Hz 
```
Should be ok with the correct monitor setup.
Is it possible to add tearing support to a headless output?

vaxerski:
maybe its the snap test not coming back to workspace 1?

UjinT34:
Fixed most of the issues. Hard to make a sensible test without some hw features for headless outputs.

LionHeartP:
The only issue I have with this PR still is that tearing + DS makes it stop rendering, even though DS is deactivated according to `hyprctl monitors` due to tearing. If I disable DS, tearing works and vice versa.

Also is there a reason tearing should be preferred over DS? Shouldn't DS take precedence? 

edit: Also, with just tearing enabled and not DS, it's bit wonky. It will freeze on a frame until I toggle fullscreen a few times.

ALL THESE TESTS ARE DONE WITH `wine-wayland` and HDR.

UjinT34:
The activation logic should be the same as on main. So the same issues should be present on main for apps that don't need this subsurface fix.
No particular reason to prefer tearing, it's the same on main. DS is usually a global setting and tearing is per-app through `immediate` rule, makes sense to prefer tearing when it's explicitly requested for an app.
Those issues and interaction between tearing and DS will be fixed after this PR.

HDR should produce more reliable results. CM for SDR DS doesn't exist atm and some SDR formats can't be used with DS at all.

This PR was meant to fix DS in wine-wayland on my nvidia setup and I am not interested in tearing. Ironically tearing seems to be working on every setup and DS works on mesa drivers. Might also work with some other videocard and nvidia driver version combos. No luck for me

hyprtester/src/shared.hpp
```diff
@@ -88,3 +88,23 @@ namespace Colors {
     }
 
 #define OK(x) EXPECT(x, "ok")
+
+#define FIXME(code)                                                                                                                                                                \
```
vaxerski: what?

UjinT34: that's for tests that don't work in testing environment yet. it inverts the test result so it doesn't fail until fixed and the `FIXME` doesn't stay hanging around after the fix. DS/tearing activation depends on hw support and there is no point in testing those without some sort of fake hw support. if there are no plans to add such support for testing environment then those tests should be removed.

vaxerski: they should be removed, we wont add tests on rootful hw

Arisa-Snowbell:
VRR on wine-wayland started working with this patch but after the force-push I think? not sure, It stopped working again, a lot of ghosting again, @LionHeartP could you check if it's true for you?

Arisa-Snowbell:
Also with wine-wayland with DS, i have to move the fullscreen game from one workspace to another to make it use DS, not sure if related

njdom24:
> VRR on wine-wayland started working with this patch but after the force-push I think? not sure, It stopped working again, a lot of ghosting again, @LionHeartP could you check if it's true for you?

That's around when I checked out this branch to test my VRR issues at https://github.com/hyprwm/Hyprland/discussions/11406. Wonder if it's related.

LionHeartP:
> VRR on wine-wayland started working with this patch but after the force-push I think? not sure, It stopped working again, a lot of ghosting again, @LionHeartP could you check if it's true for you?

Just tested right now on latest git on Returnal with wine-wayland, HDR and DS enabled and it works fine here.

https://github.com/user-attachments/assets/3d37d409-2264-4e71-9345-270b7a0ff71e

Arisa-Snowbell:
Welp, then I have no idea why it stopped working for me from one day to another as I didn't change anything else :/

vaxerski:
is this ready or...?

UjinT34:
Ready

vaxerski (approved):

PR #11270, opened 2025-07-29
dnd: drop drag-and-drop on tablet pen tip up

DarkCreepy59:
Currently it seems like dropping dnd does not check for tablet events like is done with touch and mouse. This PR makes it possible to end dnd with a stylus.

https://github.com/hyprwm/Hyprland/discussions/9878#discussioncomment-12904712 mentions dnd not being finished only "in some cases", which I find odd because I don't recall ever being able to drop dnd with a tablet.
Is there something that I am missing?

Ready but before merging I would like feedback as see above.

### DIFF
--- a/src/managers/input/Tablets.cpp
+++ b/src/managers/input/Tablets.cpp
@@ -2,6 +2,7 @@
 #include "../../desktop/Window.hpp"
 #include "../../protocols/Tablet.hpp"
 #include "../../devices/Tablet.hpp"
+#include "../../managers/HookSystemManager.hpp"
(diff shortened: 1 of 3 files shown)


vaxerski:
this generally looks good to me, I don't use tablets so I can't give input on your question. Is there anything else?

DarkCreepy59:
My main concern was whether there was some way for tablets to be handled by dnd in a way that I'm not aware of.
If you say this looks good to you, then I guess it can be merged.

And I don't think testing is necessary, I don't see any way this could cause a regression considering it was already broken before.

PR #11272, opened 2025-07-29
[hyprpm] Remove -nn flag and make notification behaviour more consist…

KAGEYAM4:

Previously` -nn` was introduced in https://github.com/hyprwm/Hyprland/pull/8167.  `-nn` turns on `-n` --> `notifyFail = notify = true;` and many `notify()` are ran without checking the flag, Creating 3 cases where notification will be generated ->

1. Some notification always gets generated.
https://github.com/hyprwm/Hyprland/blob/43966cc787c4a8844ac1e7affaadeedde8f4cc60/hyprpm/src/main.cpp#L60-L61
https://github.com/hyprwm/Hyprland/blob/43966cc787c4a8844ac1e7affaadeedde8f4cc60/hyprpm/src/main.cpp#L147-L148
https://github.com/hyprwm/Hyprland/blob/43966cc787c4a8844ac1e7affaadeedde8f4cc60/hyprpm/src/main.cpp#L170-L171
2. `-n` only generated one notification showing everything loaded successfully.
https://github.com/hyprwm/Hyprland/blob/43966cc787c4a8844ac1e7affaadeedde8f4cc60/hyprpm/src/main.cpp#L209-L210
3. `-nn`, turns on `-n` and also generates more warning/errors
https://github.com/hyprwm/Hyprland/blob/43966cc787c4a8844ac1e7affaadeedde8f4cc60/hyprpm/src/main.cpp#L152-L153
https://github.com/hyprwm/Hyprland/blob/43966cc787c4a8844ac1e7affaadeedde8f4cc60/hyprpm/src/main.cpp#L197-L205

Under this PR,  warning and error notification will always work, -n will only give visual confirmation that plugin loaded successfully, eye candy.

Compatibility may break, for that a solution is needed.

Breaking Changes for users using this flag.

### DIFF
--- a/hyprpm/src/main.cpp
+++ b/hyprpm/src/main.cpp
@@ -25,8 +25,8 @@ constexpr std::string_view HELP = R"#(┏ hyprpm, a Hyprland Plugin Manager
 ┃
 ┣ Flags:
 ┃
-┣ --notify       | -n    → Send a hyprland notification for important events (including both successes and fail events).
-┣ --notify-fail  | -nn   → Send a hyprland notification for fail events only.
+┣ --notify       | -n    → Send a hyprland notification confirming successful plugin load.
+┃                          Warnings/Errors trigger notifications regardless of this flag.
 ┣ --help         | -h    → Show this menu.
 ┣ --verbose      | -v    → Enable too much logging.
 ┣ --force        | -f    → Force an operation ignoring checks (e.g. update -f).
@@ -47,7 +47,7 @@ int                        main(int argc, char** argv, char** envp) {
     }
 
     std::vector<std::string> command;
-    bool                     notify = false, notifyFail = false, verbose = false, force = false, noShallow = false;
+    bool                     notify = false, verbose = false, force = false, noShallow = false;
     std::string              customHlUrl;
 
     for (int i = 1; i < argc; ++i) {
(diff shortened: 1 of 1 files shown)


KAGEYAM4:
@vaxerski  any thougts?

KAGEYAM4:
Done https://github.com/hyprwm/hyprland-wiki/pull/1281

hyprpm/src/main.cpp
```diff
@@ -57,8 +57,6 @@ int                        main(int argc, char** argv, char** envp) {
                 return 0;
             } else if (ARGS[i] == "--notify" || ARGS[i] == "-n") {
                 notify = true;
-            } else if (ARGS[i] == "--notify-fail" || ARGS[i] == "-nn") {
```
vaxerski: maybe keep it but mention its deprecated to not break?

KAGEYAM4:
```cpp
else if (ARGS[i] == "--notify-fail" || ARGS[i] == "-nn") {
  std::println(stderr, "{}", failureString("Deprececated flag."));
  g_pPluginManager->notify(ICON_INFO, 0, 10000, "[hyprpm] -n flag is deprecated, see hyprpm --help.");
}
```

is this good ?

vaxerski: yeah

KAGEYAM4: Done

PR #11276, opened 2025-07-29
Include physical monitor size in IPC monitor info

WhySoBad:

This pull request adds the physical size information about a monitor to the `monitors` IPC request

It's ready for merging


leiserfg:
Yes please! This will make possible to implement https://github.com/hyprwm/Hyprland/discussions/11261 in "userland" without the need of reading edid.

WhySoBad:
You're right, I haven't thought about this. 
It now prints doubles

leiserfg:
@vaxerski  sorry to borrow your time, but how does wayland/hyprland handle the case when there is no physical size (for instance a projector?) which is something considered in `edid`?

WhySoBad:
I've tested it with a headless output, in this case the size is just 0x0

leiserfg:
After checking wayland protcols I think that the original impl (with rounding) was fine, as in [wl_output::geometry](https://wayland.app/protocols/wayland#wl_output:event:geometry) they are integers. And yes, when it does not make sense they are set to `0`. 

I'm being picky here 'cause I will make the pr to add the fields in the rust client as soon as the PR is merged :smile:

WhySoBad:
> After checking wayland protcols I think that the original impl (with rounting) was fine, as in [wl_output::geometry](https://wayland.app/protocols/wayland#wl_output:event:geometry) they are integers. And yes, when it does not make sense they are set to `0`.
> 
> I'm being picky here 'cause I will make the pr to add the fields in the rust client as soon as the PR is merged 😄

True, I've just had a look at the EDID spec and (at least on wikipedia) it states that the physical size is a 1 byte centimeter value, so there is no need to allow sub-millimeter values: 
image

PR #11278, opened 2025-07-30
session-lock: explicitly consider dpms states for sending locked or denied

PointerDilemma:

Fixes https://github.com/hyprwm/Hyprland/discussions/11268

I thought we always enter unsafe state when dpms is off, but that changed at some point? Or do I remember that wrong?

Ready

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -91,38 +91,41 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
     g_pCompositor->focusSurface(nullptr);
     g_pSeatManager->setGrab(nullptr);
 
-    if (!g_pCompositor->m_unsafeState) {
-        m_sessionLock->sendDeniedTimer = makeShared<CEventLoopTimer>(
-            // Within this arbitrary amount of time, a session-lock client is expected to create and commit a lock surface for each output. If the client fails to do that, it will be denied.
-            std::chrono::seconds(5),
-            [](auto, auto) {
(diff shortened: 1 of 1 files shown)


vaxerski:
we pretty much never enter unsafe with dpms, it's only some garbage displays that physically "unplug" themselves on dpms off

src/managers/SessionLockManager.cpp
```diff
@@ -99,18 +99,20 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
                 if (!g_pSessionLockManager || g_pSessionLockManager->clientLocked() || g_pSessionLockManager->clientDenied())
                     return;
 
-                if (g_pCompositor->m_unsafeState || !g_pCompositor->m_aqBackend->hasSession() || !g_pCompositor->m_aqBackend->session->active) {
-                    // Because the session is inactive, there is a good reason for why the client did't receive locked or denied.
+                if (!g_pSessionLockManager->m_sessionLock || !g_pSessionLockManager->m_sessionLock->lock)
+                    return;
+
+                if (!g_pCompositor->m_dpmsStateOn || g_pCompositor->m_unsafeState || !g_pCompositor->m_aqBackend->hasSession() || !g_pCompositor->m_aqBackend->session->active) {
```
vaxerski: m_dpmsStateOn should be nuked, it's dumb as dpms is per monitor, so a global doesnt make sense.

PointerDilemma: Ok, then we need to ignore the FALLBACK mon explicitly. I will do that when i get back home.

PointerDilemma:
@vaxerski this should be better (and what I should have done from the start tbh)

Issue #11284, opened 2025-07-30
CM causes screenshots to be washed out


vaxerski:
### Discussion link
N/A

## Description

When apps that don't understand CM capture the screen, the colors are washed out due to the CM buffer colors being in the CM space and not sRGB. We need to transform them back.

giga ad hoc patch:
[patch.txt](https://github.com/user-attachments/files/21515252/patch.txt)

needs cleanups and actual good code

PR #11286, opened 2025-07-30
opengl: improve render fn arg clarity

vaxerski:
Improves the readability of opengl fns by using structs for params

First step before #11284 because I am not adding another param to a raw ass fn call

needs testing if it didnt break anything, this is very typo prone

probably breaks plugins


fxzzi:
https://github.com/user-attachments/assets/f8834541-ea8a-4861-ad1e-a49ff48b5e1f

hi, opening discord (native wayland, not vesktop not webcord, just discord) looks like this on this branch

edit: actually, it happens any time the window is resized

VirtCode:
breaks the rounding of window borders on my end. the framebuffer of the window itself is rounded, just not the border:
image

vaxerski:
fixed both. Anything else?

VirtCode:
I think when resizing a window to be larger, the blur is not applied for some frames, so it becomes transparent. Here's a video of what I mean. Doesn't happen on main as far as I can tell.

https://github.com/user-attachments/assets/3b52c31e-c465-4de2-8a21-bc6db2321483

other than that I didn't notice anything off

vaxerski:
fixed

VirtCode:
can confirm, everything looks good now on my end!

vaxerski:
thanks merging then

Issue #10558, opened 2025-05-27
FP16 support


UjinT34:
scRGB requires FP16 to work properly #9064
FP16 allows color values outside 0.0-1.0 range. Some shaders aren't ready to handle such values.
It makes sense to keep main buffers in linear rgb with FP16 format to simplify the CM and keep the maximum current precision. That will require to do CM twice: from client to main buffer and from main buffer to device. The second CM step will clash with `decoration:screen_shader`.

- [ ] Add an option to change internal format to FP16
- [ ] Use linear sRGB (instead of monitor format) to render to the main buffer
- [ ] Add a second CM step to convert linear sRGB to monitor format instead of `decoration:screen_shader`
- [ ] Add back `decoration:screen_shader` (second draw call? some way to integrate into the CM shader?)
- [ ] Verify that all shaders work as intended with FP16
  - [ ] blur
  - [ ] borders
  - [ ] shadows
- [ ] Check various image exports
   - [ ] Screenshots
   - [ ] Screen record
   - [ ] Mirroring

PR #11294, opened 2025-07-31
screencopy: un-hdr screencopy buffers for cm-unaware clients

### LINKED ISSUE #11284: CM causes screenshots to be washed out

vaxerski:
Fixes #11284 

Transform HDR monitor buffers back to SDR if the screencopy client is cm-unaware.

Doesn't do a _perfect_ job, but it's close enough IMO. 1:1 translation might be hard to achieve.

cc @UjinT34

### DIFF
--- a/src/protocols/ColorManagement.cpp
+++ b/src/protocols/ColorManagement.cpp
@@ -204,6 +204,10 @@ bool CColorManager::good() {
     return m_resource->resource();
 }
 
+wl_client* CColorManager::client() {
+    return m_resource->client();
+}
+
 CColorManagementOutput::CColorManagementOutput(SP<CWpColorManagementOutputV1> resource, WP<CMonitor> monitor) : m_resource(resource), m_monitor(monitor) {
     if UNLIKELY (!good())
         return;
@@ -816,6 +820,10 @@ void CColorManagementProtocol::onMonitorImageDescriptionChanged(WP<CMonitor> mon
     // recheck feedbacks
     for (auto const& feedback : m_feedbackSurfaces)
         feedback->onPreferredChanged();
+}
+
+bool CColorManagementProtocol::isClientCMAware(wl_client* client) {
+    return std::ranges::any_of(m_managers, [client](const auto& m) { return m->client() == client; });
 }
 
 void CColorManagementProtocol::destroyResource(CColorManager* resource) {
--- a/src/protocols/ColorManagement.hpp
+++ b/src/protocols/ColorManagement.hpp
@@ -18,7 +18,8 @@ class CColorManager {
   public:
     CColorManager(SP<CWpColorManagerV1> resource);
 
-    bool good();
+    bool       good();
+    wl_client* client();
 
   private:
     SP<CWpColorManagerV1> m_resource;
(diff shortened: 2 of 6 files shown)


UjinT34:
sRGB -> wide -> sRGB requires at least 12-bit precision to be considered lossless (as in indistinguishable by human eye). HDR requires more and in general loses more data depending on the source and target luminance ranges.
With #10558 it should be much easier to combine everything together and achieve 1:1 conversions. It's just a lot of internal breaking changes. It's a good time to merge #11000. Has a bunch of fixes that should work well with #11084 and `experimental:use_fp16` can be used as a starting point for FP16 transition.

vaxerski:
does this hurt though? The colors arent lossless, but they are _way_ better.

UjinT34:
No. It's probably the best solution atm.
Just a bit clunky that to get a screenshot of an SDR client with `cm = hdr` the data goes: sRGB -> linear -> HDR fb -> linear -> sRGB. Lossy and two extra math-heavy conversions. Also might have some weird interactions with monitor sdr -> hdr settings. They aren't rolled back correctly. If someone complains about those then it's probably not worth fixing without FP16.

vaxerski:
yeah, but we can't do much here, and these ops are still cheaper than re-rendering everything

LionHeartP:
Using `cm=edid` here and trying to screenshot a full screen window of a game running in HDR, looks like this
Screenshot_2025-07-31_19-16-46

It *should* look like this
Screenshot_2025-07-31_19-17-52

This is exactly how it looks if you try to view HDR content when the screen is not modeset to HDR.

UjinT34:
> yeah, but we can't do much here, and these ops are still cheaper than re-rendering everything

With FP16 it'll be "any source -> linear fb -> any target". Without the need for extra conditions and image description swapping. And monitor settings won't mess with the buffer and recording tools.

@LionHeartP monitor sdr and luminance settings and `cm_fs_passthrough` will modify the data in some way. The displayed image might not change because some of those setting are passed to the video driver and videocard/monitor should adjust the image. Screenshotting tools won't get this info and reverse transition isn't entirely correct. Should produce a better result with the default monitor settings. With `cm_fs_passthrough = 1 or 2` most of the hdr stuff is done by hw and the reverse transition will be applied to unmodified source data. Source data description might not match monitor's one in that case.

doomperfect:
Broke screenshots (grim) and screenshare (vesktop) for me :( Both only display a completely black screen now. Tried with both bitdepth 8/10, cm=srgb and wide/edid. Reverting this commit fixes the issue

vaxerski:
...how? Both work perfectly fine for me... what if you just comment out lines 208 - 209 in `src/protocols/screencopy.cpp`?
20250802_14h37m51s

doomperfect:
@vaxerski Yea that also fixed it (tested on top of main)! But I guess that's basically also a revert? 

Also I checked and hyprpicker is also broken for me without reverting this MR, just like in the linked issue

vaxerski:
can you open a discussion for this?

doomperfect:
#11315

PR #11303, opened 2025-08-01
renderer: create fence after each surface pass

gulafaran:
instead of adding the buffers to a vector and late release it on the fence in endrender that signals on ALL prior opengl work.

now its making a synckhr after the surface pass. and in endrender duping the native fd , and moving the dup nativefd to a seperate function in ceglsync that takes a param if we want to flush or not. in this case the endrender fence flushes, the buffer fence just gets the fd. and we are releasing buffers individually as early as possible with as few glflush as we can. however still warrants for some benchmarking and testing. 

because it now means creating a eglsynckhr and fd per surface. gain some, loose some.

### DIFF
--- a/src/helpers/MonitorFrameScheduler.cpp
+++ b/src/helpers/MonitorFrameScheduler.cpp
@@ -51,20 +51,13 @@ void CMonitorFrameScheduler::onPresented() {
 
     m_pendingThird = false;
 
-    Debug::log(TRACE, "CMonitorFrameScheduler: {} -> onPresented, missed, committing pending at the earliest convenience.", m_monitor->m_name);
+    auto mon = m_monitor.lock();
+    g_pHyprRenderer->commitPendingAndDoExplicitSync(mon); // commit the pending frame. If it didn't fire yet (is not rendered) it doesn't matter. Syncs will wait.
 
-    m_pendingThird = false;
-
-    g_pEventLoopManager->doLater([m = m_monitor.lock()] {
-        if (!m)
-            return;
-        g_pHyprRenderer->commitPendingAndDoExplicitSync(m); // commit the pending frame. If it didn't fire yet (is not rendered) it doesn't matter. Syncs will wait.
-
-        // schedule a frame: we might have some missed damage, which got cleared due to the above commit.
-        // TODO: this is not always necessary, but doesn't hurt in general. We likely won't hit this if nothing's happening anyways.
-        if (m->m_damage.hasChanged())
-            g_pCompositor->scheduleFrameForMonitor(m);
-    });
+    // schedule a frame: we might have some missed damage, which got cleared due to the above commit.
+    // TODO: this is not always necessary, but doesn't hurt in general. We likely won't hit this if nothing's happening anyways.
+    if (mon->m_damage.hasChanged())
+        g_pCompositor->scheduleFrameForMonitor(mon);
 }
 
 void CMonitorFrameScheduler::onFrame() {
@@ -104,12 +97,15 @@ void CMonitorFrameScheduler::onFrame() {
 }
 
 void CMonitorFrameScheduler::onFinishRender() {
-    m_sync = CEGLSync::create(); // this destroys the old sync
-    m_sync->dupNativeFence(true);
+    if (m_pendingSync || !m_monitor->m_inFence.isValid())
+        return;
 
-    g_pEventLoopManager->doOnReadable(m_sync->fd().duplicate(), [this, mon = m_monitor] {
+    m_pendingSync = true;
+    g_pEventLoopManager->doOnReadable(m_monitor->m_inFence.duplicate(), [this, mon = m_monitor] {
         if (!mon) // might've gotten destroyed
             return;
+        m_pendingSync = false;
+
         onSyncFired();
     });
 }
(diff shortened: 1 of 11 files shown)


vaxerski:
what about toplevel export?

gulafaran:
> what about toplevel export?

changed it around, now its making a synckhr after the surface pass. and in endrender duping the native fd , and moving the dup nativefd to a seperate function in ceglsync that takes a param if we want to flush or not. in this case the endrender fence flushes, the buffer fence just gets the fd. and as such toplevelexport should be working, and we are releasing buffers individually as early as possible with as few glflush as we can. however still warrants for some benchmarking and testing.

Zebra2711:
@gulafaran what happened with the old version [87f7a3a](https://github.com/hyprwm/Hyprland/commit/87f7a3a208d4496914bea7184e641d10dbbbef0b)?
adding `m_sync->dupNativeFence(true)` in func `void CMonitorFrameScheduler::onFinishRender` makes my obs record with `render:new_render_scheduling = true` have freeze frames and some screen artifacts

gulafaran:
> @gulafaran what happened with the old version [87f7a3a](https://github.com/hyprwm/Hyprland/commit/87f7a3a208d4496914bea7184e641d10dbbbef0b)? adding `m_sync->dupNativeFence(true)` in func `void CMonitorFrameScheduler::onFinishRender` makes my obs record with `render:new_render_scheduling = true` have freeze frames and some screen artifacts

well it merely restored old behaviour, since i moved the duping of the fd into that function, without it on readable would be called on an invalid fd, odd it affected it at all.

gulafaran:
technically i think the new_render_sheduling is a bit wrong, i have it disabled myself because it really messes up dual gpu / external monitor. its also by default off if im not mistaken. but i might have missed something obvious with my changes aswell heh. im gonna be offgrid until wednesday tho. so its gonna have to wait for figuring out what or why is happening here

Zebra2711:
> technically i think the new_render_sheduling is a bit wrong, i have it disabled myself because it really messes up dual gpu / external monitor. its also by default off if im not mistaken. but i might have missed something obvious with my changes aswell heh. im gonna be offgrid until wednesday tho. so its gonna have to wait for figuring out what or why is happening here

Yeah, i ran a test with three versions using `render:new_render_scheduling = true`: without this mr, with the old version [87f7a3a](https://github.com/hyprwm/Hyprland/commit/87f7a3a208d4496914bea7184e641d10dbbbef0b) and the latest. It seems the issue happens on all versions. So, I think that weird stuff is mainly caused by `new_render_scheduling`, but it's strange that it happens randomly

gulafaran:
> > technically i think the new_render_sheduling is a bit wrong, i have it disabled myself because it really messes up dual gpu / external monitor. its also by default off if im not mistaken. but i might have missed something obvious with my changes aswell heh. im gonna be offgrid until wednesday tho. so its gonna have to wait for figuring out what or why is happening here
> 
> Yeah, i ran a test with three versions using `render:new_render_scheduling = true`: without this mr, with the old version [87f7a3a](https://github.com/hyprwm/Hyprland/commit/87f7a3a208d4496914bea7184e641d10dbbbef0b) and the latest. It seems the issue happens on all versions. So, I think that weird stuff is mainly caused by `new_render_scheduling`, but it's strange that it happens randomly

reverted a commitpending in dolater, fixed my obs. but im still seeing a weird artifacting on launch of hyprland for a frame or two. otherwise the new render sheduling seems happy here.

gulafaran:
> > technically i think the new_render_sheduling is a bit wrong, i have it disabled myself because it really messes up dual gpu / external monitor. its also by default off if im not mistaken. but i might have missed something obvious with my changes aswell heh. im gonna be offgrid until wednesday tho. so its gonna have to wait for figuring out what or why is happening here
> 
> Yeah, i ran a test with three versions using `render:new_render_scheduling = true`: without this mr, with the old version [87f7a3a](https://github.com/hyprwm/Hyprland/commit/87f7a3a208d4496914bea7184e641d10dbbbef0b) and the latest. It seems the issue happens on all versions. So, I think that weird stuff is mainly caused by `new_render_scheduling`, but it's strange that it happens randomly

reverted a commitpending in dolater, fixed my obs. but im still seeing a weird artifacting on launch of hyprland for a frame or two. otherwise the new render sheduling seems happy here.

> > > technically i think the new_render_sheduling is a bit wrong, i have it disabled myself because it really messes up dual gpu / external monitor. its also by default off if im not mistaken. but i might have missed something obvious with my changes aswell heh. im gonna be offgrid until wednesday tho. so its gonna have to wait for figuring out what or why is happening here
> > 
> > 
> > Yeah, i ran a test with three versions using `render:new_render_scheduling = true`: without this mr, with the old version [87f7a3a](https://github.com/hyprwm/Hyprland/commit/87f7a3a208d4496914bea7184e641d10dbbbef0b) and the latest. It seems the issue happens on all versions. So, I think that weird stuff is mainly caused by `new_render_scheduling`, but it's strange that it happens randomly
> 
> reverted a commitpending in dolater, fixed my obs. but im still seeing a weird artifacting on launch of hyprland for a frame or two. otherwise the new render sheduling seems happy here.

ok nvm so it only worked fine on dual monitors, going single it brought back the reason for vaxry adding dolater -_-. time do dig.

Zebra2711:
It seems to be working fine now with `new_render_scheduling=true`. But, when open obs and change `new_render_scheduling` from `false` to `true` the screen is freeze (key bind still work), so it seems that only the screen rendering stops. Undoing the change (undo key in vim and saved hyprland.conf) makes it normal again.

gulafaran:
yeah im sending a PR only fixing the new_render soon, before i fiddle with this per surface fence. seems there is a bunch of buffer release assumptions made here that needs to be incorporated to the actual pageflip now that the new_render can "prerender" a frame.

Issue #11088, opened 2025-07-18
single_window_aspect_ratio should not be used when using maximize


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11057

## Description
When using maximize (fullscreen, 1) single_window_aspect_ratio should be ignored.

vlofgren:
I implemented the original PR (and have been daily driving the setting since it was merged, just never used that particular window rule).  I can have a look at fixing it if you'd like.

felipemdutra:
Hello. I'm looking to make my first contribution. Do you think this bug is a good starter issue?

gnusenpai:
I actually liked this behavior, maybe it should be configurable?

abenz1267:
Over the last few days i noticed that i personally would want both behaviours. So maybe another options for the fullscreen dispatcher would be better? One that doesn't ignore (current behaviour) + one that ignores.

vaxerski:
> So maybe another options for the fullscreen dispatcher would be better?

sounds like clutter. I think we can just make it configurable.

foggerty:
When playing with this feature, I also noticed that while changes to `single_window_aspect_ratio_tolerance` will be picked up when the config is saved/reloaded, changes to `single_window_aspect_ratio` will be ignored until Hyprland is restarted.

vlofgren:
> When playing with this feature, I also noticed that while changes to `single_window_aspect_ratio_tolerance` will be picked up when the config is saved/reloaded, changes to `single_window_aspect_ratio` will be ignored until Hyprland is restarted.

When I implemented the change it was reloading just fine, and I've been daily driving on that build until 0.50 dropped, but now  reloading of the property it appears to have broken, so it's possibly a regression elsewhere or an interaction with some other change to the layout engine.

I'm certain it worked because I was testing `single_window_aspect_ratio_tolerance` specifically by reloading various changes to `single_window_aspect_ratio`.

NyxErinys:
> > So maybe another options for the fullscreen dispatcher would be better?
> 
> sounds like clutter. I think we can just make it configurable.

yeah I would love for this to be a config option, I have `single_window_aspect_ratio` set to 16 9 because of my 32 9 screen but every once and a while i want the single window to be fully maximized on it, currently having figured out a way to do that (without using `fullscreen 0` instead which isn't ideal

Dregu:
[...]

PR #11305, opened 2025-08-02
dwindle: fix single_window_aspect_ratio not updating with config reload

cryeprecision:

The config value `dwindle:single_window_aspect_ratio` doesn't update at runtime because the `VEC2` is stored as a value in a static variable. This PR fixes this by storing the `CConfigValue` in a static variable and accessing the `VEC2` value through the `ptr` method.

This fixes the problem mentioned in #11088

- https://github.com/hyprwm/Hyprland/issues/11088#issuecomment-3094729535
- https://github.com/hyprwm/Hyprland/issues/11088#issuecomment-3094733502

I tested this in a debug build and it works, i.e., the config value is now applied immediately when the config is saved.
I also ran `clang-format`, checked that the project still builds, and ran `hyprtester`.

This should be ready to merge.

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -160,10 +160,10 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
 
     Vector2D          ratioPadding;
 
-    if (REQUESTEDRATIO.ptr()->y != 0 && !pNode->pParent) {
+    if ((*REQUESTEDRATIO).y != 0 && !pNode->pParent) {
         const Vector2D originalSize = PMONITOR->m_size - PMONITOR->m_reservedTopLeft - PMONITOR->m_reservedBottomRight;
 
-        const double   requestedRatio = REQUESTEDRATIO.ptr()->x / REQUESTEDRATIO.ptr()->y;
+        const double   requestedRatio = (*REQUESTEDRATIO).x / (*REQUESTEDRATIO).y;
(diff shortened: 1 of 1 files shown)


src/layout/DwindleLayout.cpp
```diff
@@ -155,15 +155,15 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
...
     auto              calcSize = PWINDOW->m_size;
 
-    const static auto REQUESTEDRATIO          = *CConfigValue<Hyprlang::VEC2>("dwindle:single_window_aspect_ratio");
+    const static auto REQUESTEDRATIO          = CConfigValue<Hyprlang::VEC2>("dwindle:single_window_aspect_ratio");
     const static auto REQUESTEDRATIOTOLERANCE = CConfigValue<Hyprlang::FLOAT>("dwindle:single_window_aspect_ratio_tolerance");
 
     Vector2D          ratioPadding;
 
-    if (REQUESTEDRATIO.y != 0 && !pNode->pParent) {
+    if (REQUESTEDRATIO.ptr()->y != 0 && !pNode->pParent) {
```
vaxerski: `(*REQUESTEDRATIO).y` here and below

vaxerski (commented):
pfff that's an easy one to miss indeed

src/layout/DwindleLayout.cpp
cryeprecision: Done 🫡

vaxerski (approved):

PR #11323, opened 2025-08-03
protocols: implement ext-data-control

noocsharp:
This protocol has superseded wlr-data-control

This implements the ext-data-control protocol which has superceded wlr-data-control. Because the protocols are essentially identical, the actual protocol code has been copied verbatim, only renaming things from the wlr-data-control implementation.

I have tested it with my clipboard utility and it works fine, hasn't been tested with wl-clipboard, although the similarity between the protocols means that there shouldn't be any issues.

#### Is it ready for merging, or does it need work

This is ready for merging.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -382,6 +382,7 @@ protocolnew("staging/color-management" "color-management-v1" false)
 protocolnew("staging/xdg-toplevel-tag" "xdg-toplevel-tag-v1" false)
 protocolnew("staging/xdg-system-bell" "xdg-system-bell-v1" false)
 protocolnew("staging/ext-workspace" "ext-workspace-v1" false)
+protocolnew("staging/ext-data-control" "ext-data-control-v1" false)
 
 protocolwayland()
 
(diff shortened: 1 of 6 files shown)


vaxerski (commented):
I can't see how this integrates with the clipboard. You can send to yourself, but you can't become a clipboard manager as you can't override the clipboard. You implemented the protocol - but didn't connect it to the rest of the codebase...?

noocsharp:
Apologies, this was an oversight on my part. Just for clarity, do you mean that sources that come through the old protocol should result in offers in the new protocol, and vice-versa? Or is there some additional integration that needs to be done as well?

vaxerski:
yes, I mean, how else would you want a clipboard manager to work?

noocsharp:
I did some more research on this, and it looks like the burden is not on the compositor to copy the offers across protocols, it is on the creator of the data source. That is, if a program wants to have some data available via both protocols, it has to create the data sources and offers on both protocols. And if a program wants to receive from both protocols, it has to listen for offers on both protocols.

In wlroots, the implementation of ext-data-control is line-by-line copy of the wlr-data-control-unstable. Given this, I think this pull request is fine as-is and is ready to merge.

vaxerski:
this is not a copy of wlr data control though.

See for example
image

xref'd dataWlr

noocsharp:
OK, should be fixed now. Let me know if I missed anything else.

PR #11335, opened 2025-08-04
fix clang + change term fg to white after it fails

enessmr:
Fix clang and fix term color, so its white after fail. (It might produce a new line.)

Unchecked code. I did not built hyprtester

Ready to merge. (Can need work, but it feels like that it's READY)


hyprtester/src/main.cpp
```diff
@@ -184,6 +184,7 @@ int main(int argc, char** argv, char** envp) {
     NLog::log("{}launching hl", Colors::YELLOW);
     if (!launchHyprland(configPath, binaryPath)) {
         NLog::log("{}well it failed", Colors::RED);
+        std::cout << "\033[37m";
```
C0Florent: 37 is ANSI white, which may be be different from the default text color (depending on terminal colour theme)
`Colors::RESET` seems more appropriate here, if the intention is to reset text formatting to default

enessmr: OK I got it 👍

umbrageodotus:
Needed?

enessmr:
Gosh this thing is old af

umbrageodotus:
@enessmr use color::reset

ItsOhen:
Doesn't hyprland throw on pipefailure and dip before this is run?

enessmr (commented):
> /build/source/hyprtester/src/main.cpp:189:13: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
       >   189 |             NLog::log("", Colors::RESET);
       >       |             ^~~~

This is why i hate C++

PR #11337, opened 2025-08-05
configWatcher: fix inotify event reading buffer size

ImanSeyed:
Read full variable-length inotify_event structure instead of just the fixed-size header. The previous code only read `sizeof(inotify_event)` bytes, missing the trailing name field, which could cause truncated events and undefined behavior.

`inotify_event` has a variable-length structure:
```c
struct inotify_event {
    int wd;
    uint32_t mask;  
    uint32_t cookie;
    uint32_t len;
    char name[];
};
```
The kernel writes complete events including the `name` field, so we need to read `sizeof(inotify_event) + NAME_MAX + 1` bytes to avoid truncation. Also added handling for multiple events that may be packed in a single `read()` call.

Yes, it is ready for merging.

### DIFF
--- a/src/config/ConfigWatcher.cpp
+++ b/src/config/ConfigWatcher.cpp
@@ -65,17 +65,34 @@ void CConfigWatcher::setOnChange(const std::function<void(const SConfigWatchEven
 }
 
 void CConfigWatcher::onInotifyEvent() {
-    inotify_event ev;
-    while (read(m_inotifyFd.get(), &ev, sizeof(ev)) > 0) {
-        const auto WD = std::ranges::find_if(m_watches.begin(), m_watches.end(), [wd = ev.wd](const auto& e) { return e.wd == wd; });
+    constexpr size_t                                     BUFFER_SIZE = sizeof(inotify_event) + NAME_MAX + 1;
+    alignas(inotify_event) std::array<char, BUFFER_SIZE> buffer      = {};
+    const ssize_t                                        bytesRead   = read(m_inotifyFd.get(), buffer.data(), buffer.size());
+    if (bytesRead <= 0)
+        return;
 
-        if (WD == m_watches.end()) {
-            Debug::log(ERR, "CConfigWatcher: got an event for wd {} which we don't have?!", ev.wd);
-            return;
+    for (size_t offset = 0; offset < (size_t)bytesRead;) {
+        const auto* ev = (const inotify_event*)(buffer.data() + offset);
+
+        if (offset + sizeof(inotify_event) > (size_t)bytesRead) {
+            Debug::log(ERR, "CConfigWatcher: malformed inotify event, truncated header");
+            break;
         }
 
-        m_watchCallback(SConfigWatchEvent{
-            .file = WD->file,
-        });
+        if (offset + sizeof(inotify_event) + ev->len > (size_t)(bytesRead)) {
+            Debug::log(ERR, "CConfigWatcher: malformed inotify event, truncated name field");
+            break;
+        }
+
+        const auto WD = std::ranges::find_if(m_watches, [wd = ev->wd](const auto& e) { return e.wd == wd; });
+
+        if (WD == m_watches.end())
+            Debug::log(ERR, "CConfigWatcher: got an event for wd {} which we don't have?!", ev->wd);
+        else
+            m_watchCallback(SConfigWatchEvent{
+                .file = WD->file,
+            });
+
+        offset += sizeof(inotify_event) + ev->len;
     }
 }


src/config/ConfigWatcher.cpp
```diff
@@ -65,17 +65,37 @@ void CConfigWatcher::setOnChange(const std::function<void(const SConfigWatchEven
...
 
 void CConfigWatcher::onInotifyEvent() {
-    inotify_event ev;
-    while (read(m_inotifyFd.get(), &ev, sizeof(ev)) > 0) {
-        const auto WD = std::ranges::find_if(m_watches.begin(), m_watches.end(), [wd = ev.wd](const auto& e) { return e.wd == wd; });
+    constexpr size_t BUFFER_SIZE = sizeof(inotify_event) + NAME_MAX + 1;
+    alignas(inotify_event) std::array<char, BUFFER_SIZE> buffer{};
+
+    const ssize_t bytesRead = read(m_inotifyFd.get(), buffer.data(), buffer.size());
+    if (bytesRead <= 0) {
```
vaxerski: no {}

src/config/ConfigWatcher.cpp
```diff
@@ -65,17 +65,37 @@ void CConfigWatcher::setOnChange(const std::function<void(const SConfigWatchEven
...
+        return;
+    }
+
+    for (size_t offset = 0; offset < static_cast<size_t>(bytesRead);) {
+        const auto* ev = reinterpret_cast<const inotify_event*>(buffer.data() + offset);
```
vaxerski: we generally prefer C-style casts, `(inotify_event*)(buffer.data() ...)`

ImanSeyed: Even for `static_cast`? I changed both `static_cast` and `reinterpret_cast` to C-style casts. Please let me know if you only meant to change `reinterpret_cast`.

gulafaran: I wouldnt use C casts at all, the c++ ones has stricter rules that makes it safer, reinterpret cast is a bit nuclear so in this case with this struct it wont matter, but casting classes up/down with c style cast it can easily become undefined behaviour.

vaxerski: thats correct but you see `static_cast` is many more characters than `()`

maybe we could get some using directives and shorten it to `sc<>` `rc<>` etc, but thats for down the line

src/config/ConfigWatcher.cpp
vaxerski: you didnt remove the braces

src/config/ConfigWatcher.cpp
```diff
@@ -65,17 +65,37 @@ void CConfigWatcher::setOnChange(const std::function<void(const SConfigWatchEven
...
+        }
+
+        if (offset + sizeof(inotify_event) + ev->len > (size_t)(bytesRead)) {
+            Debug::log(ERR, "CConfigWatcher: malformed inotify event, truncated name field");
+            break;
+        }
+
+        const auto WD = std::ranges::find_if(m_watches, [wd = ev->wd](const auto& e) { return e.wd == wd; });
 
         if (WD == m_watches.end()) {
```
vaxerski: here no braces too

vaxerski:
clang-format needed

ImanSeyed:
> clang-format needed

`clang-format` is trying to add 40 spaces (literally) behind the `bytesRead` and `BUFFER_SIZE`.
```c++
    constexpr size_t                                     BUFFER_SIZE = sizeof(inotify_event) + NAME_MAX + 1;
    alignas(inotify_event) std::array<char, BUFFER_SIZE> buffer{};

    const ssize_t                                        bytesRead = read(m_inotifyFd.get(), buffer.data(), buffer.size());
```
Is this normal?

EDIT: `AlignConsecutiveDeclarations: AcrossEmptyLines` was adding excessive whitespace to align with longer variable names. Switching from `{}` to `= {}` initialization fixed the formatting (both work identically for `std::array`, so it should be okay).
```c++
    constexpr size_t                                     BUFFER_SIZE = sizeof(inotify_event) + NAME_MAX + 1;
    alignas(inotify_event) std::array<char, BUFFER_SIZE> buffer      = {};
    const ssize_t                                        bytesRead   = read(m_inotifyFd.get(), buffer.data(), buffer.size());

```

vaxerski:
yes thats normal

ImanSeyed:
> yes thats normal

Done. Let me know if you need anything else from me on this.

vaxerski (approved):

marwuint:
Looks like we’re missing a header :)  
`../src/config/ConfigWatcher.cpp:68:96: error: ‘NAME_MAX’ was not declared in this scope`

I assume `#include <linux/limits.h>` would fix it? I can open a PR tonight if no one else gets to it first.

PR #11357, opened 2025-08-06
nix/lib: add bezier to topCommandsPrefixes

fxzzi:

if any custom beziers are defined for animations, hyprland will complain that the beziers haven't been defined. I think this change makes sense as its likely most configurations are defining custom beziers anyway.

all looks fine. Does this file need to be formatted? running alejandra on the file also reformatted your comments so leaving it for now.

should be ready

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -82,7 +82,7 @@ lib: let
     :::
   */
   toHyprlang = {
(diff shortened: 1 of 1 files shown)


fufexan (approved):
LGTM. Don't know what's up with formatting but we should soon move to nixfmt.

PR #11361, opened 2025-08-06
masterlayout: add previous mode for focusmaster command

VoltaireNoir:
## Summary
This PR adds a new `previous` mode to the existing `focusmaster` Master layout command / layout message.

Behavior: When the master is focused from a non-master window, the previously focused non-master window is remembered. If the command is issued and the master is already in focus, the focus is switched to the previously remembered window. If no saved window exists, ~focus stays on master~ fallback to `auto` mode.

This feature makes it trivial to switch back and forth between a master and a slave window with a single command, especially when there are several windows in the workspace.

## Notes

1. If a new mode is not desirable, the existing auto mode can be enhanced with the same behavior.
2. The PR doesn't add any validity checks for the remembered/saved window, since it is already handled in the `switchToWindow` closure.
3. ~~I am aware that the use of `goto` is generally frowned upon, but I feel the usage is sensible here. I'd be happy to remove it if the maintainers disagree.~~

### DIFF
--- a/hyprtester/src/tests/main/persistent.cpp
+++ b/hyprtester/src/tests/main/persistent.cpp
@@ -25,7 +25,7 @@ static bool test() {
 
     // test on workspace "window"
     NLog::log("{}Switching to workspace 1", Colors::YELLOW);
-    OK(getFromSocket("/dispatch workspace 1"));
+    getFromSocket("/dispatch workspace 1"); // no OK: we might be on 1 already
 
     OK(getFromSocket("/keyword workspace 5, monitor:HEADLESS-2, persistent:1"));
     OK(getFromSocket("/keyword workspace 6, monitor:HEADLESS-PERSISTENT-TEST, persistent:1"));


src/layout/MasterLayout.cpp
```diff
@@ -1138,12 +1139,30 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
+            if (ARG == "master") {
+                return 0;
+                // previous: switch to previously saved window
+            } else if (ARG == "previous") {
+                const auto WS = getMasterWorkspaceData(PWINDOW->workspaceID());
+                switchToWindow(WS->focusMasterPrev.lock());
+            } else {
+                Debug::log(ERR, "Invalid argument '{}' for 'focusmaster' layoutmsg", ARG);
+                // fallback to auto
+                goto focusauto;
```
vaxerski: goto is strictly banned. Please use sane programming choices like ifs

src/layout/MasterLayout.cpp
```diff
@@ -1138,12 +1139,30 @@ std::any CHyprMasterLayout::layoutMessage(SLayoutMessageHeader header, std::stri
...
             return 0;
 
+        const bool ARGS = vars.size() >= 2;
+
         if (PMASTER->pWindow.lock() != PWINDOW) {
             switchToWindow(PMASTER->pWindow.lock());
-        } else if (vars.size() >= 2 && vars[1] == "master") {
-            return 0;
+            // save previously focused window (only for `previous` mode)
+            if (ARGS && vars[1] == "previous") {
```
vaxerski: no {} around short ifs

VoltaireNoir:
@vaxerski I've applied suggested changes and rebased on main.

**Update**: I just spotted a bug where if the previously saved window is moved to another workspace, issuing the command while on master switches to that workspace. I don't think this behavior is desirable. I'll add a workspace ID check to fix it.

Also, after daily driving the feature, I think using the `auto` mode as a fallback makes more sense when the previously saved window is not available or is no longer valid, instead of keeping the focus on master.

vaxerski:
oh also you forgot adding tests woops so yeah please add tests

VoltaireNoir:
> oh also you forgot adding tests woops so yeah please add tests

Got it. I'll add the tests. I don't see any master layout related tests in `hyprtester` so I presume I'll have to add a new file at `hyprtester/src/tests/main/master.cpp`? or somewhere else? @vaxerski

vaxerski:
you can append new tests inside the master tests

VoltaireNoir:
@vaxerski Added tests.

I also added [another condition](https://github.com/hyprwm/Hyprland/pull/11361/commits/5deed21ad1902ea2ab1d085e077cdcdb0ca616d2#diff-c0920369e3b4867c61cc18b8be73db48eed3eed4b35886ac1cf03150f7d204b8R1167) to fallback to `auto` (if prev win is same as the current one).

vaxerski (dismissed):
lgtm, thanks for the tests. Wiki mr needed for the new opt

VoltaireNoir:
> lgtm, thanks for the tests. Wiki mr needed for the new opt

Wiki PR created: https://github.com/hyprwm/hyprland-wiki/pull/1200

vaxerski (approved):

PR #11363, opened 2025-08-06
fix: Use proper Wayland protocol error codes in color management

xander1421:
Previously, the color management implementation used an invalid error code -1, which is not defined in the Wayland protocol specification. This caused protocol violations and crashes in Firefox-based browsers when monitors were temporarily unavailable during mode switches (e.g., entering fullscreen).

The issue affected Firefox and Firefox-based browsers but NOT Chromium:
- Firefox would crash with: 'wp_color_manager_v1#48: error -1: Invalid output (2)'
- Monitor would show 'mode not supported' requiring cable replug
- Issue was specific to Firefox's Wayland color management implementation

The issue was particularly noticeable with multi-monitor setups using non-standard resolutions (like 4096x2160 DCI 4K).

Changes:
- Replace error code -1 with WP_COLOR_MANAGER_V1_ERROR_UNSUPPORTED_FEATURE (0) as defined in the wp_color_manager_v1 protocol
- Add proper LOGM(WARN) logging for debugging temporary resource unavailability
- Improve error messages to be more descriptive
- Apply same fix to ContentType.cpp for consistency

This prevents protocol violations and allows applications to gracefully handle temporary monitor unavailability during mode switches.

Fixes crashes when:
- Switching videos to fullscreen in Firefox/Zen Browser
- Changing display modes in Firefox-based browsers
- Using multi-monitor setups with non-standard resolutions

Environment:
- Arch Linux (6.15.8-arch1-1 x86_64)
- NVIDIA RTX 4070 Ti SUPER (driver 575.64.05)
- egl-wayland 4:1.1.19-1
- Multi-monitor: 4096x2160@60Hz, 2560x1440@144Hz, 5120x1440@240Hz
- Hyprland 0.50.1

Tested browsers:
- Firefox Developer Edition 142.0b7 (crashed before, works after)
- Zen Browser 1.14.10b (crashed before, works after)
- Chromium 139.0.7258.66-1 (was not affected)


UjinT34:
It's not the correct code for this case. If a client queries non-existent output than a fake wp-cm-output should be created and this fake output should act as an inert object:
> If this protocol object is inert, the resulting image description object shall immediately deliver the wp_image_description_v1.failed event with the no_output cause.

https://wayland.app/protocols/color-management-v1#wp_color_management_output_v1:request:get_image_description

And the same for surfaces.

Too much fuss to make it the right way. And this shouldn't happen unless monitor switches off in the middle of the request. Probably the issue is somewhere else. E.g. a client receives invalid output list or mapping wl output to hl output isn't correct. Looks like this PR just hides the underlying issue.

xander1421:
should i also consider issues being from the monitor or cable?

UjinT34:
I don't think it's cable. If some conditions trigger a mode change in fullscreen then monitor might do this by disconnecting and reconnecting. This can change the output list.
To fully debug the issue you should track which wl_outputs were sent to the browser, verify that it requests cm for one of those outputs, the requested wl_output exists during the request and that it's correctly mapped to a HL output.
The `r->error(-1, "Invalid output (2)")` should be closer to an unreachable code. If you can hit it reliably then there is something else going wrong.

Flat:
> Too much fuss to make it the right way. And this shouldn't happen unless monitor switches off in the middle of the request. Probably the issue is somewhere else. E.g. a client receives invalid output list or mapping wl output to hl output isn't correct. Looks like this PR just hides the underlying issue.

Do we currently have a way to log the output list sent to clients, or would that need to be added?

> The r->error(-1, "Invalid output (2)") should be closer to an unreachable code. If you can hit it reliably then there is something else going wrong.

I can trigger this 100% on Firefox/Librefox 142 by just unplugging both monitors, no full screen or media playback required. 

see https://github.com/hyprwm/Hyprland/discussions/11529

Doesn't happen on Gnome, did happen on KDE, which has a MR to fix (null ptr).

mhwasil:
is there any updates on this issue and PR?
discussion #11529

PR #11368, opened 2025-08-07
Hardened config logic against Time-Of-Check race conditions

maxolasersquad:

This PR provides some hardening around [Time-Of-Check to Time-Of-Use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race conditions in the configuration manager.

This makes two improvements. The first one is when no configuration file is found, so a new one is created. I believe this is a straightforward fix.

The second is really only a minor improvement. Instead of first checking if it exists, and then later checking if it is a "regular" file, it only asks the filesystem once, and then branches logic off of that. This is safer. However, it still acts on the file after the check, which still allows for TOCTOU. Fixing that requires a bigger refactor.

I believe so. I did a fresh compilation from these changes on a new Arch install. On first launch it created the config file. On subsequent launches it read the existing config file.


src/config/ConfigManager.cpp
```diff
@@ -873,11 +873,14 @@ CConfigManager::CConfigManager() {
...
     std::string parentPath = std::filesystem::path(configPath).parent_path();
 
-    if (!std::filesystem::is_directory(parentPath)) {
-        Debug::log(WARN, "Creating config home directory");
-        try {
-            std::filesystem::create_directories(parentPath);
-        } catch (std::exception& e) { throw e; }
+    try {
+        if (!parentPath.empty()) {
+            if (std::filesystem::create_directories(parentPath))
```
vaxerski: why not use a `std::error_code` instead of a try catch?

src/config/ConfigManager.cpp
```diff
@@ -3041,27 +3044,28 @@ std::optional<std::string> CConfigManager::handleSource(const std::string& comma
...
     for (size_t i = 0; i < glob_buf->gl_pathc; i++) {
         auto value = absolutePath(glob_buf->gl_pathv[i], m_configCurrentPath);
-
-        if (!std::filesystem::is_regular_file(value)) {
-            if (std::filesystem::exists(value)) {
-                Debug::log(WARN, "source= skipping non-file {}", value);
+        try {
+            auto file_status = std::filesystem::status(value);
+
+            if (std::filesystem::is_regular_file(file_status)) {
```
vaxerski: here as well

src/config/ConfigManager.cpp
maxolasersquad: Yeah, that makes sense. I'll make the change as soon as I have the time.

maxolasersquad: Let me know if this refactor works.

src/config/ConfigManager.cpp
```diff
@@ -873,11 +873,15 @@ CConfigManager::CConfigManager() {
...
 
-    if (!std::filesystem::is_directory(parentPath)) {
-        Debug::log(WARN, "Creating config home directory");
-        try {
-            std::filesystem::create_directories(parentPath);
-        } catch (std::exception& e) { throw e; }
+    if (!parentPath.empty()) {
+        std::error_code ec;
+        bool            created = std::filesystem::create_directories(parentPath, ec);
+        if (ec) {
```
vaxerski, line 879: can this set ec if already exists? Already existing is not fatal

maxolasersquad: If the config file already exists then `created` will be `false`. `if (ec)` will evaluate to false. I tested this behavior by compiling this on to a fresh install (in a vm) of Arch. I was able to run Hyprland, and it created the example config file. I then exited out of Hyprland and then ran it again. I did not receive any errors.

vaxerski: alright, thanks.

vaxerski (approved):

PR #11380, opened 2025-08-08
Resolve potential null pointer crash and two Time of Check conditions

maxolasersquad:
Resolves a potential null pointer crash and two [Time-of-Check to Time-of-Use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) conditions.

While Hyprland does some sanity checking when it starts, a future refactor could potentially allow XDG_RUNTIME_DIR to not be set in the CCompositor constructor. If that happened a null pointer exception would be thrown. I added a check to make sure this cannot happen.

A very minor change is that `m_hyprTempDataRoot` is now constructed with `std::filesystem` so that the platform will build the directory separators itself.

Two possible TOCTOU conditions existed when created hyprTempDataRoot directory and when creating the instancePath directories. These are now patched up with ask-for-forgiveness.

I compiled these changes on an Arch system and Hyprland still runs fine.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -206,22 +206,20 @@ CCompositor::CCompositor(bool onlyConfig) : m_onlyConfigVerification(onlyConfig)
 
     setenv("HYPRLAND_INSTANCE_SIGNATURE", m_instanceSignature.c_str(), true);
 
-    if (!std::filesystem::exists(m_hyprTempDataRoot))
-        mkdir(m_hyprTempDataRoot.c_str(), S_IRWXU);
-    else if (!std::filesystem::is_directory(m_hyprTempDataRoot)) {
-        std::println("Bailing out, {} is not a directory", m_hyprTempDataRoot);
+    std::error_code ec;
+    std::filesystem::create_directories(m_hyprTempDataRoot, ec);
+    if (ec) {
+        std::println("Bailing out, could not create or access {} : {}", m_hyprTempDataRoot, ec.message());
         throw std::runtime_error("CCompositor() failed");
     }
 
-    m_instancePath = m_hyprTempDataRoot + "/" + m_instanceSignature;
+    m_instancePath = (std::filesystem::path(m_hyprTempDataRoot) / m_instanceSignature).string();
+    if (!std::filesystem::create_directory(m_instancePath, ec)) {
+        if (ec) {
+            std::println("Bailing out, couldn't create {} : {}", m_instancePath, ec.message());
+        } else
+            std::println("Bailing out, {} exists??", m_instancePath);
 
-    if (std::filesystem::exists(m_instancePath)) {
-        std::println("Bailing out, {} exists??", m_instancePath);
-        throw std::runtime_error("CCompositor() failed");
-    }
-
-    if (mkdir(m_instancePath.c_str(), S_IRWXU) < 0) {
-        std::println("Bailing out, couldn't create {}", m_instancePath);
         throw std::runtime_error("CCompositor() failed");
     }
 


src/Compositor.cpp
```diff
@@ -188,12 +188,12 @@ CCompositor::CCompositor(bool onlyConfig) : m_onlyConfigVerification(onlyConfig)
 
     setMallocThreshold();
 
-    m_hyprTempDataRoot = std::string{getenv("XDG_RUNTIME_DIR")} + "/hypr";
-
-    if (m_hyprTempDataRoot.starts_with("/hypr")) {
-        std::println("Bailing out, $XDG_RUNTIME_DIR is invalid");
-        throw std::runtime_error("CCompositor() failed");
+    const char* xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
```
vaxerski: XDG_RUNTIME_DIR

src/Compositor.cpp
```diff
@@ -188,12 +188,12 @@ CCompositor::CCompositor(bool onlyConfig) : m_onlyConfigVerification(onlyConfig)
...
+    if (xdg_runtime_dir == nullptr || *xdg_runtime_dir == '\0') {
```
vaxerski: `!XDG_RUNTIME_DIR`
and `XDG_RUNTIME_DIR[0]`

src/Compositor.cpp
```diff
@@ -206,22 +206,20 @@ CCompositor::CCompositor(bool onlyConfig) : m_onlyConfigVerification(onlyConfig)
...
+    if (ec) {
+        std::println("Bailing out, could not create or access {} : {}", m_hyprTempDataRoot, ec.message());
         throw std::runtime_error("CCompositor() failed");
     }
 
-    if (mkdir(m_instancePath.c_str(), S_IRWXU) < 0) {
-        std::println("Bailing out, couldn't create {}", m_instancePath);
+    m_instancePath = (std::filesystem::path(m_hyprTempDataRoot) / m_instanceSignature).string();
+    if (!std::filesystem::create_directory(m_instancePath, ec)) {
+        if (ec) {
```
vaxerski, line 218: no {} or around else {}

src/Compositor.cpp
vaxerski: also the validity check is redundant as main.cpp will not launch without xdg runtime dir set

maxolasersquad: Thanks for the feedback! I've made the change to the `XDG_RUNTIME_DIR` variable name.

Regarding the validity check, I'd like to share my reasoning for adding it, but I'm happy to remove it if you still feel it's unnecessary.

My main concern is making the code robust through defensive programming. While `main.cpp` currently ensures `$XDG_RUNTIME_DIR` is set, that dependency is external to this class. If the startup logic in `main.cpp` ever changes, or if this class is used in a different context in the future, this check prevents a potential null pointer bug from surfacing here. It makes the class more self-contained and reliable. I find it's a small price to pay for future stability.

If you agree to keep the check, let's decide on the style. I think you suggested `!XDG_RUNTIME_DIR and XDG_RUNTIME_DIR[0]`. This would actually cause a crash if the variable is null due to the `&&`. Perhaps you meant `if (!XDG_RUNTIME_DIR || !XDG_RUNTIME_DIR[0])`?

That version is functionally identical to my original `if (XDG_RUNTIME_DIR == nullptr || *XDG_RUNTIME_DIR == '\0')`. I personally find the original more explicit, but I'm happy to use the more idiomatic C-style version if you prefer it.

Just let me know how you'd like to proceed!

vaxerski: we fundamentally cannot proceed without xdg runtime dir and many things wont work, so I think removing this check is ok

src/Compositor.cpp
```diff
@@ -206,22 +206,20 @@ CCompositor::CCompositor(bool onlyConfig) : m_onlyConfigVerification(onlyConfig)
 
     setenv("HYPRLAND_INSTANCE_SIGNATURE", m_instanceSignature.c_str(), true);
 
-    if (!std::filesystem::exists(m_hyprTempDataRoot))
-        mkdir(m_hyprTempDataRoot.c_str(), S_IRWXU);
-    else if (!std::filesystem::is_directory(m_hyprTempDataRoot)) {
-        std::println("Bailing out, {} is not a directory", m_hyprTempDataRoot);
+    std::error_code ec;
+    std::filesystem::create_directories(m_hyprTempDataRoot, ec);
```
vaxerski, line 210: no, we can't use STL here because it doesn't offer permissions for the directory at creation

Issue #11191, opened 2025-07-23
Allow getting props (akin to setprop)


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11162

## Description
`setprop` exists. `getprop` could be useful.

PR #11394, opened 2025-08-09
hyprctl: added getprop

Brumus14:
It adds `getprop` to `hyprctl` to complement the existing `setprop`.
It allows getting properties about windows.
Fixes: hyprwm/Hyprland#11191 

There were some properties that have no obvious default value:
- `animationstyle` I have returned `(unset)` and an empty string for JSON
- `minsize` has an internal default of `MIN_WINDOW_SIZE` that I returned
- `maxsize` has an internal default of `MAX_WINDOW_SIZE` that I returned
- `bordercolor` get their default values from the config

It should be ready for reviewing, would appreciate if someone could check that the default return values I chose make sense.
Haven't added getprop to completion files (like hyprctl.bash / hyprctl.usage) as I don't have experience with them.

### DIFF
--- a/hyprctl/Strings.hpp
+++ b/hyprctl/Strings.hpp
@@ -49,6 +49,7 @@ commands:
                           the same format as in colors in config. Will reset
                           when Hyprland's config is reloaded
     setprop ...         → Sets a window property
+    getprop ...         → Gets a window property
     splash              → Get the current splash
     switchxkblayout ... → Sets the xkb layout index for a keyboard
     systeminfo          → Get system info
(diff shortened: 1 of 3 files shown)


Brumus14:
Ok will do

hyprtester/src/tests/main/hyprctl.cpp
```diff
@@ -29,6 +154,10 @@ static bool test() {
         EXPECT(jqProc.exitCode(), 0);
     }
 
+    if (!testGetprop()) {
```
vaxerski: no {}

Brumus14:
Great and I've just fixed that now

vaxerski (approved):
lgtm, just needs a wiki mr now

Brumus14:
I'm having a look at doing that now

PR #11422, opened 2025-08-13
Reference command-line arguments instead of copying them

Hato1125:
Instead of copying command-line arguments into a `std::vector`, this change uses `std::span` and `std::string_view` to reference them directly, improving performance and reducing memory usage.

No known issues. This change is fully backward-compatible.

Ready for merging.


src/main.cpp
```diff
@@ -53,87 +55,91 @@ int main(int argc, char** argv) {
     int                      socketFd   = -1;
     bool                     ignoreSudo = false, verifyConfig = false;
 
-    std::vector<std::string> args{argv + 1, argv + argc};
+    if (argc > 1) {
+        std::span<char*> args{ argv + 1, static_cast<std::size_t>(argc - 1) };
```
vaxerski: rebase on main and use `Hyprutils::Memory::sc<T>`, `using Hyprutils::Memory`

Hato1125: Thank you for the review. I have rebased on `main`, replaced `static_cast` with `Hyprutils::Memory::sc`, and added `using Hyprutils::Memory`.

vaxerski (dismissed):
thanks!

Hato1125:
I formatted the code because the PR failed the clang-format check.
As stated in the PR guidelines, I ran `clang-format -i src/**/*{cpp,hpp}` from the project’s root directory. However, it did not format the files as expected.
I then ran clang-format directly on the files that failed the check.

vaxerski:
it only formats hyprland not hyprtester :P

PR #11427, opened 2025-08-13
fix: prevent infinite event loop in XWM clipboard transfers

adityalohuni:
Error: [ERR] [xwm] No transfer with property data found

Fix: Prevent infinite event loop in XWM clipboard transfers

The writeDataSource callback was being recreated even when onWrite() 
failed (returned 0), causing an infinite loop and high CPU usage.

Only recreate the event source when onWrite() returns 1 (needs continuation).


adityalohuni:
Done.

PR #11448, opened 2025-08-16
config: `split_bias`: simplify logic and set of possible values.

myQwil:

The 3rd option for `split_bias` is redundant because the split ratio can simply be inverted (`2 - ratio`) to make the opposite window or the opposite directions receive the larger half.

n/a

Ready for merging.

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -1,22 +1,24 @@
-#include "tests.hpp"
-#include "../../shared.hpp"
-#include "../../hyprctlCompat.hpp"
-#include <print>
+#include <cmath>
 #include <thread>
 #include <chrono>
 #include <hyprutils/os/Process.hpp>
 #include <hyprutils/memory/WeakPtr.hpp>
-#include <csignal>
-#include <cerrno>
+
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
Needs a wiki mr and some tests while we're at it

hyprtester/src/tests/main/window.cpp
```diff
@@ -1,22 +1,26 @@
...
-#define UP CUniquePointer
-#define SP CSharedPointer
+static bool spawnKitty(const std::string& class_, int count) {
+    NLog::log("{}Spawning {}", Colors::YELLOW, class_);
+    if (!Tests::spawnKitty(class_)) {
+        NLog::log("{}Error: {} did not spawn", Colors::RED, class_);
+        return false;
+    }
+    NLog::log("{}Expecting {} window(s)", Colors::YELLOW, count);
+    EXPECT(Tests::windowCount(), count);
```
vaxerski: Tests::spawnKitty already does this. The window count expect is redundant. You could just do EXPECT(Tests::spawnKitty(), true) iirc

the reason its there is because legacy. Window count tests can be dropped after kitty spawns

myQwil: I wasn't able to find an example of `EXPECT` and `Tests::spawnKitty()` being used together in that way, but I did get rid of the window count.

PR #11453, opened 2025-08-16
renderer: improve zoom in anims

vaxerski:
Improves the monitor zoom in animations when a display is connected.

Fixes:
- zoom in anims now are per-display
- zoom in anims should now wait for the display to modeset before starting
- zoom in anims now follow the animation tree with `monitorAdded`

Removed `animations:first_launch_animation` as it's useless now

Part of #11440, @System64fumo can you test this


System64fumo:
Works, Display takes too long to show anything so animation is only seen if configured to be really slow.

Issue #10979, opened 2025-07-09
New render scheduling causes lags and stutters on select setups


vaxerski:
### Moved from
https://github.com/hyprwm/aquamarine/issues/190

## Description

On some select setups, `render:new_render_scheduling` causes lagspikes and low fps.

vaxerski:
ref https://github.com/hyprwm/Hyprland/pull/10973

didnt fix but thats where the code tests will occur (I can't repro)

kRHYME7:
@vaxerski  is `render:new_render_scheduling` option available in `extra/hyprland` ?  

I should probably inform them if that's the case.

vaxerski:
Why would it be if the new render scheduling was merged 2 days ago?

kRHYME7:
Because arch updates it yesterday https://archlinux.org/packages/extra/x86_64/hyprland/

So this was just an aquamarine version bump? and ` render:new_render_scheduling` should be for `0.50` release?


---

edit: So that issue is different. Thanks! Indeed it's only an AQ version bump.

vaxerski:
well you said the lag was introduced with https://github.com/hyprwm/Hyprland/commit/8f948827a69499c0b112043debad35f82a312b6b which is NOT in 0.49, so decide then what breaks things?

kRHYME7:
My bad, this issue https://github.com/HyDE-Project/HyDE/issues/960 is the reason why I realized the regression, although might be unrelated. We will try to fix this downstream first. 

Moving forward, let me know if you want me to test this too https://github.com/hyprwm/Hyprland/pull/10973 or any patch.

vaxerski:
if you don't have any lag on new render scheduling, no need to. I know @outfoxxed does though, that's why the MR exists

kRHYME7:
> if you don't have any lag on new render scheduling, no need to.

I do.  `render:new_render_scheduling = false` fixes it for me.  We got same issues as described by @outfoxxed  in the MR. 0


<img width="421" height="117" alt="Image" src="https://github.com/user-attachments/assets/5c9398af-1042-48c6-b019-79a040419441" />

fxzzi:
im in a country with like 0.4mbps internet and 99.99% of the web is blocked so cant test.

maybe this is related to my dropped frames in mpv after the merge? Is this a scenario in which this PR would help?

kRHYME7:
[...]

PR #11454, opened 2025-08-16
renderer: store used buffers per monitor

gulafaran:
instead of storing used async bufs in the renderer and dropping them on the first monitor pageflip, store them per monitor and drop on the monitors pageflip.

problem before was the first monitor could be dropping the second monitors surfaces before it had even rendered them causing glitches. especially noticeable on dual gpu laptop and external monitor.

move surface frames to CSurfaceManager aswell and schedule them for scanout, and make third pending frame call it aswell in new_render_scheduling.

#TODO
- [ ] ensure the new_render_scheduler actually works as intended
- [x] test toplevel export

fixes: #12181
maybe: #10979


vaxerski (commented):
yes, but we need a cross-monitor solution too...

I propose to make a manager like `CBufferReleaseManager` that has:
 - `::add(surf, mon)`
 - `::drop(mon)`

Each surface stored gets a `state` with a `std::vector<PHLMONITORREF>` and it gets dropped once that vec has either 0 elements or all elements are null or disabled (in case of a monitor disconnect)

gulafaran:
> yes, but we need a cross-monitor solution too...
> 
> I propose to make a manager like `CBufferReleaseManager` that has:
> 
>     * `::add(surf, mon)`
> 
>     * `::drop(mon)`
> 
> 
> Each surface stored gets a `state` with a `std::vector<PHLMONITORREF>` and it gets dropped once that vec has either 0 elements or all elements are null or disabled (in case of a monitor disconnect)

can be a way to deal with surfaces showing on more then one monitor yes, should buffers even be dropped if the rendermode isnt RENDER_MODE_NORMAL?

vaxerski:
> can be a way to deal with surfaces showing on more then one monitor yes, should buffers even be dropped if the rendermode isnt RENDER_MODE_NORMAL?

uhh... in general no, but there is toplevel export, which doesn't use NORMAL, and should drop to avoid the window going stale... It already uses a flag there blockSurfaceFeedback iirc, you could use that. If the feedback isnt blocked, add it to be dropped

gulafaran:
@vaxerski did some pseudo thing here. it breaks DS here but rest seems fine. and im a bit unsure if on that visible check and also the CHLBufferReference implentation, is it like a shared pointer or do the monitors just end up with 2 different local ones so the syncreleasers are seperate and doesnt actually merge the fences of both, but just keeps them seperate one per monitor.

but this is in general is the idea that should probably be done. to get buffer releases somewhat right. new_render_shed i got no clue if it broke even further. it already was here heh

gulafaran:
@vaxerski ok got buffer releases working without breaking DS or OBS now. i still dont know how to proceed with new_render_shed tho since its all sorts of wonky even before

gulafaran:
@Zebra2711 wanna test this now? i still have to figure out if there can be something done on laptops tho, seems now its blitting a lot more making it actually slower on new_render_shed when having dualgpu/external monitor tho

gulafaran:
@outfoxxed seeing you had various stall/lag issues would be neat if you could test aswell.

what i do know is DS might require some finicking now that commitpendinganddoexplicit is called on new_render_scheduling while it isnt on setting it false.

Zebra2711:
> @Zebra2711 wanna test this now? i still have to figure out if there can be something done on laptops tho, seems now its blitting a lot more making it actually slower on new_render_shed when having dualgpu/external monitor tho

It hit my performance so hard, my hyprland on the iGPU couldn't get more than 60 fps with or without `new_render_scheduling` enabled

gulafaran:
> > @Zebra2711 wanna test this now? i still have to figure out if there can be something done on laptops tho, seems now its blitting a lot more making it actually slower on new_render_shed when having dualgpu/external monitor tho
> 
> It hit my performance so hard, my hyprland on the iGPU couldn't get more than 60 fps with or without `new_render_scheduling` enabled

Lol wat in what? I finally got overlay showing 240 + 160 here and cs2 200+ on external

gulafaran:
> > @Zebra2711 wanna test this now? i still have to figure out if there can be something done on laptops tho, seems now its blitting a lot more making it actually slower on new_render_shed when having dualgpu/external monitor tho
> 
> It hit my performance so hard, my hyprland on the iGPU couldn't get more than 60 fps with or without `new_render_scheduling` enabled

unless ofc you just measured something that is vsynced, glxgears is normally. i dont think it has been correct before. hyprland pushes like 100fps if i crunch hard on the gpu before glxgears would still report 240., well if we are pageflipping at 100fps we cant exactly render glxgears at 240.

Zebra2711:
> Lol wat in what? I finally got overlay showing 240 + 160 here and cs2 200+ on external

idk, my hyprland build with #11000, #11087, #11228, and #11216, and your old closed PR #11303, with `new_render_scheduling=true` when started. Everything is extremely smooth 🙃

gulafaran:
> > Lol wat in what? I finally got overlay showing 240 + 160 here and cs2 200+ on external
> 
> idk, my hyprland build with #11000, #11087, #11228, and #11216, and your old closed PR #11303, with `new_render_scheduling=true` when started. Everything is extremely smooth 🙃

yeah because its kinda cheating, two monitors? if so if any monitor pageflips its dropping the buffer no matter if the second monitor is still rendering it. only reason it didnt artifact or break is because clients mostly double or more buffer. so they begin working on another and once they reuse the dropped one if they do. its work is done luckily enough.

i mean turn on debug:overlay = true, and see the fps. the overlay comes with its own cost so it drops a bit too but. and then run glxgears, if it isnt matching something is very off in my mind

gulafaran:
new_render_sched also hits the blitting harder in AQ if this is a dual gpu thing. and its very slow sadly, needs some work.

gulafaran:
> started. Everything is extremely

what setup is it, dual gpu? dual monitors?

Zebra2711:
ok, I built hyprland with #11000, #11087, #11228, and #11216. This MR seems to be performant, close to #11303, and I tested it with `glxgears`.
186xx fps (#11303) vs 172xx fps this MR. `debug:overlay = false`

> what setup is it, dual gpu? dual monitors?

intel igpu + nvidia gpu and no external display

gulafaran:
> ok, I built hyprland with #11000, #11087, #11228, and #11216. This MR seems to be performant, close to #11303, and I tested it with `glxgears`. 186xx fps (#11303) vs 172xx fps this MR. `debug:overlay = false`
> 
> > what setup is it, dual gpu? dual monitors?
> 
> intel igpu + nvidia gpu and no external display

did a thing, hope its moving in the right direction. but havent tested all cases since im on my laptop atm, or well with my laptop away from my desk.

Zebra2711:
i think the main cause of the lag here is that hyprland is using more gpu than it normally would. Lag is reduced compared to the old version, but not much. I still feel laggy with this. Could be that my config is messing with this. I will test it out with the default config tomorrow

Out of the scope of this, when I resize a window or move my mouse, the igpu usage just jumps up

outfoxxed:
> @outfoxxed seeing you had various stall/lag issues would be neat if you could test aswell.
> 
> what i do know is DS might require some finicking now that commitpendinganddoexplicit is called on new_render_scheduling while it isnt on setting it false.

I can try to test it soon

gulafaran:
> > @outfoxxed seeing you had various stall/lag issues would be neat if you could test aswell.
> > what i do know is DS might require some finicking now that commitpendinganddoexplicit is called on new_render_scheduling while it isnt on setting it false.
> 
> I can try to test it soon

still think new_render_sched is borked tho still not satisfied with that

src/render/pass/SurfacePassElement.cpp
```diff
@@ -153,8 +155,24 @@ void CSurfacePassElement::draw(const CRegion& damage) {
...
+
+            if (m_data.pMonitor == m)
+                g_pBufferReleaseManager->add(m, m_data.surface->m_current.buffer);
+            else if (m_data.pWindow && m_data.pWindow->visibleOnMonitor(m) && g_pBufferReleaseManager->add(m, m_data.surface->m_current.buffer)) {
+                CBox wbox = {m_data.pWindow->m_realPosition->value(), m_data.pWindow->m_realSize->value()};
+
+                if (m_data.pWindow->m_isFloating)
+                    wbox = m_data.pWindow->getFullWindowBoundingBox();
+
+                m->addDamage(wbox.intersection({m->m_position, m->m_size}));
```
vaxerski: ya cant damage while rendring this doesnt do anythig

src/render/Renderer.hpp
```diff
@@ -90,22 +90,21 @@ class CHyprRenderer {
...
-    float                           m_crashingDistort    = 0.5f;
-    wl_event_source*                m_crashingLoop       = nullptr;
-    wl_event_source*                m_cursorTicker       = nullptr;
-
-    std::vector<CHLBufferReference> m_usedAsyncBuffers;
+    bool             m_crashingInProgress = false;
+    float            m_crashingDistort    = 0.5f;
+    wl_event_source* m_crashingLoop       = nullptr;
+    wl_event_source* m_cursorTicker       = nullptr;
+    CTimer           m_renderTimer;
```
vaxerski: rebase fail, removed variable brought bac

src/helpers/MonitorFrameScheduler.cpp
```diff
@@ -43,37 +43,24 @@ void CMonitorFrameScheduler::onSyncFired() {
...
-void CMonitorFrameScheduler::onPresented() {
-    if (!newSchedulingEnabled())
-        return;
-
-    if (!m_pendingThird)
-        return;
-
-    Debug::log(TRACE, "CMonitorFrameScheduler: {} -> onPresented, missed, committing pending.", m_monitor->m_name);
-
-    m_pendingThird = false;
```
vaxerski, line 64: I think we need to split these changes into two MRs - just do the buffer dropping here

kRHYME7:
Hi! How should I test this one, any logs that can help?

gulafaran:
> Hi! How should I test this one, any logs that can help?

nah thats fine i sort of know whats going wrong, there is two issues one is frame callbacks is being sent wrong. and explicit sync breaks, i just havent managed to cleanly implent it without breaking other things. so i have just put this on hold for now

PR #11464, opened 2025-08-17
renderer: add overscan support

### LINKED ISSUE #11462: Add overscan support
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10092

## Description

Some garbage-ass TVs crop the input image. We could add overscan via 2 ways:
- hard: actually render at a smaller resolution
- easy: squish the rendered image to fit

IMO easy is enough (nobody will game on that shit, it's just to see things), and hard would require a bunch of renderer hacks probably...

vaxerski:
adds basic overscanning by squishing the image. Configurable only via v2, per-side with px: `overscan_` + `l/r/t/b`.

Issues:
 - [ ] some damage issues (open 1 window and close it, trails behind)

Fixes https://github.com/hyprwm/Hyprland/issues/11462


Mohataseem89:
The “squish” approach feels like the right trade-off here minimal renderer complexity, predictable behavior and good enough for the class of displays that actually need overscan. the per side `overscan_l/r/t/b` config also gives users fine control without overengineering it.

I can reproduce the damage artifacts mentioned which suggests damage tracking might not be fully accounting for the postscale transform. possibly the damage region needs to be expanded or recalculated after the overscan transform is applied

PR #11480, opened 2025-08-19
Fix DPMS animation playing when not changing state

aaronjamt:
When running `hyprctl dispatch dpms on`, the animation plays every time, even if the display is currently on. This PR prevents that and only plays the animation if the state has actually changed. Fixes #11479

There's probably a better way to do this, feel free to close it and do it a different way if you want.

Should be ready for merging unless someone knows of a better method.


vaxerski (approved):
lgtm

PR #11501, opened 2025-08-22
plugins: add eRenderStage::RENDER_POST_WALLPAPER

jmanc3:
I was trying to draw on the wallpaper with my plugin. The problem was that `RENDER_PRE_WINDOWS` comes after docks and panels are drawn, and so I would, unexpectedly end up drawing over docks.

This PR adds an `eRenderStage` that comes after the wallpaper texture and background layers are drawn, but before `RENDER_PRE_WINDOWS`, and I've tested it as working.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -884,6 +884,8 @@ void CHyprRenderer::renderAllClientsForWorkspace(PHLMONITOR pMonitor, PHLWORKSPA
             renderLayer(ls.lock(), pMonitor, time);
         }
 
+        EMIT_HOOK_EVENT("render", RENDER_POST_WALLPAPER);
+
         for (auto const& ls : pMonitor->m_layerSurfaceLayers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]) {
             renderLayer(ls.lock(), pMonitor, time);
         }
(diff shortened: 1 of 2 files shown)


src/render/Renderer.cpp
```diff
@@ -909,6 +911,8 @@ void CHyprRenderer::renderAllClientsForWorkspace(PHLMONITOR pMonitor, PHLWORKSPA
             renderLayer(ls.lock(), pMonitor, time);
         }
 
+        EMIT_HOOK_EVENT("render", RENDER_POST_WALLPAPER);
```
vaxerski, line 914: I think you forgot about the fullscreen case, renderallclientsforfullscreen or something forgot the fn name

jmanc3: I wasn't able to find a function like what you are describing. 

What I did to find where to insert these places was to grep where `clearWithTex` was called from, and these were the only two places. `clearWithTex` is the function that draws the default wallpapers when none is set and so it seemed that if I emitted after wherever in the code that function was called (and after the background layers are drawn), it should be good.

What is the 'fullscreen case' you thought needed attending to though? Like avoiding emitting if something is fullscreen or...?

vaxerski: ah my bad, durnk

PR #11503, opened 2025-08-22
Fix CM for DS and SDR passthrough

UjinT34:
Fixes SDR primaries conversion for DS and `cm_fs_passthrough = 1` . Blocks DS when CM without shader isn't supported.

Replaces CTM provided by ctm proto in fullscreen when necessary.

Ready

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -3,7 +3,6 @@
 #include "../helpers/math/Math.hpp"
 #include <algorithm>
 #include <aquamarine/output/Output.hpp>
-#include <cstdint>
 #include <filesystem>
 #include "../config/ConfigValue.hpp"
 #include "../config/ConfigManager.hpp"
--- a/hyprtester/src/shared.hpp
+++ b/hyprtester/src/shared.hpp
@@ -88,23 +88,3 @@ namespace Colors {
     }
 
 #define OK(x) EXPECT(x, "ok")
-
-#define FIXME(code)                                                                                                                                                                \
-    do {                                                                                                                                                                           \
-        const int OLD_FAILED = TESTS_FAILED;                                                                                                                                       \
-        const int OLD_RET    = ret;                                                                                                                                                \
-                                                                                                                                                                                   \
-        { code }                                                                                                                                                                   \
-                                                                                                                                                                                   \
-        if (TESTS_FAILED > OLD_FAILED) {                                                                                                                                           \
-            NLog::log("{}FIXME Broken test has failed, counting as passed", Colors::YELLOW);                                                                                       \
-            TESTS_FAILED--;                                                                                                                                                        \
-            TESTS_PASSED++;                                                                                                                                                        \
-            ret = OLD_RET;                                                                                                                                                         \
-        } else {                                                                                                                                                                   \
-            NLog::log("{}FIXME Broken test has passed, change it to EXPECT", Colors::YELLOW);                                                                                      \
-            TESTS_FAILED++;                                                                                                                                                        \
-            TESTS_PASSED--;                                                                                                                                                        \
-            ret = 1;                                                                                                                                                               \
-        }                                                                                                                                                                          \
-    } while (0)


LionHeartP:
- #11228 + DS, PQ HDR games work fine
- #11228 with DS disabled, scRGB games work fine.

With this PR, PQ games work without DS (blocked by color management) and scRGB games don't work at all (completely wrong colors not just oversaturated)

AMD GPU.

LionHeartP:
Retested after new set of commits
- HDR PQ works with DS now
- scRGB successfully blocks DS, but colors are oversaturated now

Tested with all `cm_fs_passthrough` settings.

UjinT34:
Something is wrong during initial start. Switching workspaces should help with oversaturated scRGB. I am looking into it.

UjinT34:
Fixed

LionHeartP:
Tested Control. It works fine.

Cyberpunk is a bit more iffy. Switching between HDR modes and/or SDR, sometimes doesn't trigger AutoHDR properly and you need a workspace toggle to fix it. It's the only game I know/own that has multiple HDR options so it's not easy to test it in a different game.

Without this PR, Control is very hit or miss and Cyberpunk is even worse.

UjinT34:
> Cyberpunk is a bit more iffy. Switching between HDR modes and/or SDR, sometimes doesn't trigger AutoHDR properly and you need a workspace toggle to fix it. It's the only game I know/own that has multiple HDR options so it's not easy to test it in a different game.

I've tested it with Cyberpunk and mangohud. The game itself shouldn't show any noticeable difference between PQ and scRGB. Mangohud has washed out (correct?) colors with scRGB and oversaturated with PQ. Works the same with different DS and passthrough combos.
Any reliable way to reproduce? And some logs would be nice.

LionHeartP:
> I've tested it with Cyberpunk and mangohud. The game itself shouldn't show any noticeable difference between PQ and scRGB. Mangohud has washed out (correct?) colors with scRGB and oversaturated with PQ. Works the same with different DS and passthrough combos. Any reliable way to reproduce? And some logs would be nice.

Can't reproduce anymore, everything works fine

src/render/Renderer.cpp
```diff
@@ -1545,19 +1556,39 @@ bool CHyprRenderer::commitPendingAndDoExplicitSync(PHLMONITOR pMonitor) {
...
+            if (FS_DESC.has_value()) {
+                Debug::log(INFO, "[CM] Updating fullscreen CTM");
+                pMonitor->m_noShaderCTM        = true;
+                const auto                 mat = FS_DESC->getPrimaries().convertMatrix(pMonitor->m_imageDescription.getPrimaries()).mat();
+                const std::array<float, 9> CTM = {
+                    mat[0][0], mat[0][1], mat[0][2], //
+                    mat[1][0], mat[1][1], mat[1][2], //
+                    mat[2][0], mat[2][1], mat[2][2], //
+                };
+                pMonitor->m_output->state->setCTM(CTM);
```
vaxerski, line 1588: flaky: a ctm manager can override this

UjinT34: that's handled by `if (pMonitor->m_ctmUpdated && !pMonitor->m_noShaderCTM) {`

vaxerski: no, I mean you set ctm, and then a ctm manager sets its own, you dont guard ag that

`CHyprlandCTMControlProtocol::setCTM`

UjinT34: `CHyprlandCTMControlProtocol::setCTM` calls `CMonitor::setCTM` which updates internal `CMonitor::m_ctm` and raises `CMonitor::m_ctmUpdated` to send this internal ctm to AQ. This sending is blocked by the mentioned condition. Fullscreen logic doesn't touch `CMonitor::m_ctm` and sends ctm directly to AQ. Upon exiting fullscreen the `CMonitor::m_ctmUpdated` is raised and ctm is set to the latest value from the proto (or the default identity matrix if proto wasn't used)

vaxerski: mb

PR #11505, opened 2025-08-23
protocols: xdg_activation send invalid token when serial is not valid

ross96D:
### Add serial validation to the protocol xdg_activation 

The current xdg activation protocol implementation allows clients to activate toplevel without any validation of user input. Some popular apps abuse this behavior (Telegram with the Draw attention setting on), also other apps could abuse this.

This makes having `misc:focus_on_activate true` a bit annoying.

If we validate the serial from the user input then we solve the problem of apps activating them self without user interaction.

#### Ruff edges
The current way of invalidating serials is not the best IMO. The gtk 3 implementation of gtk_window_present will use the last serial received. This means that even if the serial was created two hours ago the serial will be valid.

### DIFF
--- a/src/protocols/XDGActivation.cpp
+++ b/src/protocols/XDGActivation.cpp
@@ -1,6 +1,7 @@
 #include "XDGActivation.hpp"
 #include "../managers/TokenManager.hpp"
 #include "../Compositor.hpp"
+#include "../managers/SeatManager.hpp"
 #include "core/Compositor.hpp"
 #include <algorithm>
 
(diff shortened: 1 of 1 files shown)


ross96D:
> good for now

After a bit more testing and researching I realize this commit is trash and maybe should be reverted, for several reasons, mainly because no one seems to inject the XDG_ACTIVATION_TOKEN (ej: alacritty/ghostty) and specs like the status item notifier (tray system) dont offer a way to pass them.

This is just worse then having `focus_on_activate` on or off

vaxerski:
what does this have to do with the env? The serial is sent over the wl wire and makes sense to reject activations without one

ross96D:
> makes sense to reject activations without one

Yes, it does makes sense.

> what does this have to do with the env

The issue is that there could be a user experience regression because apps are not respecting the xdg_activation protocol. At least, maybe it should be behind an option. 

In the terminal example.. When clicking an url link the terminal needs to run the browser / xdg_open with the XDG_ACTIVATION_TOKEN env set, otherwise the browser will call xdg_activation_v1::get_activation_token, set a serial that they receive in some previous time that is not associated to the action received in the terminal.

This means that sometimes the token will work.. and some other time the token will not work.

vaxerski:
hm, that is true. We can revert this I guess.

PR #11508, opened 2025-08-23
config: update environment if cfg changes live

vaxerski:
Discussed in https://github.com/basecamp/omarchy/pull/808

Needs a wiki mr

Haven't really tested but should work


dhh:
Oh fantastic! This would be a huge help for folks setting up the screen ENVs in particular 🤘

PR #11536, opened 2025-08-27
xwayland: handle minimize and maximize requests

jmanc3:
I was missing requests from x clients to maximize or to minimize for my plugin.

_NET_STATE atoms are described [here](https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html#id-1.6.8).

If you're wondering, WM_CHANGE_STATE is generated by XIconifyWindow (or equivalent xcb call), and is described [here](https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/XIconifyWindow.html). 

I've tested this PR as working.

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -433,6 +433,14 @@ void CXWM::handleClientMessage(xcb_client_message_event_t* e) {
 
             XSURF->m_events.stateChanged.emit();
         }
+    } else if (e->type == HYPRATOMS["WM_CHANGE_STATE"]) {
+		int state = e->data.data32[0];
+        if (state == XCB_ICCCM_WM_STATE_ICONIC || state == XCB_ICCCM_WM_STATE_WITHDRAWN) {
+            XSURF->m_state.requestsMinimize = true;
+        } else if (state == XCB_ICCCM_WM_STATE_NORMAL) {
+            XSURF->m_state.requestsMinimize = false;
+        }
+        XSURF->m_events.stateChanged.emit();
     } else if (e->type == HYPRATOMS["_NET_ACTIVE_WINDOW"]) {
         XSURF->m_events.activate.emit();
     } else if (e->type == HYPRATOMS["XdndStatus"]) {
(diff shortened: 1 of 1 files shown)


src/xwayland/XWM.cpp
```diff
@@ -425,10 +425,22 @@ void CXWM::handleClientMessage(xcb_client_message_event_t* e) {
...
+                    XSURF->m_state.requestsMinimize = updateState(action, XSURF->m_minimized);
+                if (prop == HYPRATOMS["_NET_WM_STATE_MAXIMIZED_VERT"] || prop == HYPRATOMS["_NET_WM_STATE_MAXIMIZED_HORZ"])
+                    XSURF->m_state.requestsMaximize = updateState(action, XSURF->m_maximized);
             }
 
             XSURF->m_events.stateChanged.emit();
         }
+    } else if (e->type == HYPRATOMS["WM_CHANGE_STATE"]) {
+        int state = e->data.data32[0];
+        if (state == XCB_ICCCM_WM_STATE_ICONIC || state == XCB_ICCCM_WM_STATE_WITHDRAWN) {
```
vaxerski: no {}

UjinT34:
Needs some testing with games #6375. #6966 probably misses some bits. This PR might add them.

jmanc3:
> Needs some testing with games #6375. #6966 probably misses some bits. This PR might add them.

I'll check it out and see if it those bugs are re-introduced, or if those are fixed.

jmanc3:
Well, at a minimum I can confirm that pico park classic edition still does the blackscreen when you switch away from the workspace (if I had to guess it's something like the x client is watching map notify and going on low power when it knows it's unmapped and then hyprland doesn't ever send a map notify so that it can switch out of that state), both with this PR and without it.

I think that fullscreen bug is a seperate issue to this PR.

What I did discover was that you can rescue a game from the blackscreen of death if you have a bind like
```bash
bind = $mainMod, F, fullscreen
```

And then just toggle the window to fullscreen until it fixes itself.

jmanc3:
I've isolated the problem to
```c++
void CXWaylandSurface::configure(const CBox& box) {
```
If in
```c++
void CHyprXWaylandManager::activateWindow(PHLWINDOW pWindow, bool activate)
```
you write 
```c++
        pWindow->m_xwaylandSurface->activate(activate);
        pWindow->m_xwaylandSurface->configure(pWindow->m_xwaylandSurface->m_geometry);
```
it'll reproduce the fullscreen bug every time the window loses focus. What I can't figure out is why that's the case.

jmanc3:
Nvm, fixed the blackscreen bug in: https://github.com/hyprwm/Hyprland/pull/11542

PR #11551, opened 2025-08-29
plugins: expose csd functionality

jmanc3:
This PR allows plugins who'd like to handle CSD requests sent by clients, the ability to do so, by centralizing that functionality into functions which can be hooked and intercepted.

I've tested it as working.

### DIFF
--- a/src/protocols/ServerDecorationKDE.cpp
+++ b/src/protocols/ServerDecorationKDE.cpp
@@ -13,12 +13,6 @@ CServerDecorationKDE::CServerDecorationKDE(SP<COrgKdeKwinServerDecoration> resou
         mostRecentlySent      = sendMode;
         mostRecentlyRequested = mode;
     });
-    m_resource->setRelease([this](COrgKdeKwinServerDecoration* pMgr) {
-        auto sendMode = kdeModeOnReleaseCSD();
-        m_resource->sendMode(sendMode);
-        mostRecentlySent      = sendMode;
-        mostRecentlyRequested = 0;
-    });
 
     // we send this and ignore request_mode.
     auto sendMode = kdeDefaultModeCSD();


vaxerski (commented):
why can't you hook the two constructors?

jmanc3:
It's a bit tricky to explain but basically it has to do with what happens when you unload the plugin. Technically it's possible to call `m_resource->setRequestMode` with your own function and you'd be able to handle csd requests. The problem is what happens when your plugin is unloaded. It isn't possible to fix the state of this protocol to original behaviour because those anonymous functions which this functions creates are too hidden.

```c++
  private:
    struct {
        F<void(COrgKdeKwinServerDecoration*)> release;
        F<void(COrgKdeKwinServerDecoration*, uint32_t)> requestMode;
    } requests;
```

`private` things are possible to poke with templates, but these functions eventually store anonymous functions in anonymous private structs. no template blackmagic can make those accessible to be able to restore them on unloading.

You also can't just try and create a new function when your plugin gets unloaded that does the original behaviour because your memory space is about to get deleted, and therefore the function is corrupted.

The best you could do as a plugin is to set those functions to nullptr (on unload).

But then that leaves the protocol completely broken as wayland clients will never receive CSD replies anymore.

---

As far as I can tell the two solutions are this PR. Or what I was going to ask you eventually anyways which is, can `.hpp` files in `/protocols` be modified, or are they generated and to be left untouched? Because if it's possible to edit them, then (using `class COrgKdeKwinServerDecoration` as an example), 

```c++
  private:
    struct {
        F<void(COrgKdeKwinServerDecoration*)> release;
        F<void(COrgKdeKwinServerDecoration*, uint32_t)> requestMode;
    } requests;
```
could be changed to something like

```c++
  public:
    struct Requests {
        F<void(COrgKdeKwinServerDecoration*)> release;
        F<void(COrgKdeKwinServerDecoration*, uint32_t)> requestMode;
    }
  private:
    Requests requests;
```
That way, it would be possible for original behavior to be restored on unload. (I'm a big fan of solution 2 btw, it would be such short modifications and would allow for plugins to be fully transparent when it comes to protocols. it would be fire, but of course it depends on if these files are meant to be modified or not).

vaxerski:
I meant hooking `CServerDecorationKDE::CServerDecorationKDE` and executing your own logic. You can just mimic the original code but change the mode

jmanc3:
Yes that works, but when the plugin is unloaded, you'll crash Hyprland when it tries interacting with the lamda the plugin made (dlclose invalidates all memory created by the plugin)..
crash
And if you try hooking and keeping a copy of the anonymous lamda that Hyprland creates in the first place and try restoring THAT valid memory lambda on plugin exit, that will still crash Hyprland because std::functions can be copied by the compiler (therefore that memory is about to become just as invalid as soon as the plugin exit finishes).

And in terms of functionality, the state tracking this PR introduces makes it possible to set the correct decorations even if you weren't there to witness and hook the creation of the decorations.

vaxerski:
this makes sense, alright.

I'd rather that this was just a member of the protocol instead of some global functions, though.

jmanc3:
should be good

src/protocols/ServerDecorationKDE.cpp
```diff
@@ -1,21 +1,47 @@
...
 
     m_resource->setRelease([this](COrgKdeKwinServerDecoration* pMgr) { PROTO::serverDecorationKDE->destroyResource(this); });
     m_resource->setOnDestroy([this](COrgKdeKwinServerDecoration* pMgr) { PROTO::serverDecorationKDE->destroyResource(this); });
+    m_resource->setRequestMode([this](COrgKdeKwinServerDecoration*, uint32_t mode) {
+        auto sendMode = kdeModeOnRequestCSD(mode);
+        m_resource->sendMode(sendMode);
+        mostRecentlySent      = sendMode;
+        mostRecentlyRequested = mode;
+    });
+    m_resource->setRelease([this](COrgKdeKwinServerDecoration* pMgr) {
```
vaxerski: ??? release is a dtor

gulafaran:
yeah this PR made 1 core pinned to 100% when firefox runs doesnt matter what you do. im not entirerly sure how this protocol works but it seems to be causing some sendmode/request mode loop. profiled it into CServerDecorationKDE, and simply this makes it stop.

```
diff --git a/src/protocols/ServerDecorationKDE.cpp b/src/protocols/ServerDecorationKDE.cpp
index 8f625168..3307f841 100644
--- a/src/protocols/ServerDecorationKDE.cpp
+++ b/src/protocols/ServerDecorationKDE.cpp
@@ -8,6 +8,7 @@ CServerDecorationKDE::CServerDecorationKDE(SP<COrgKdeKwinServerDecoration> resou
     m_resource->setRelease([this](COrgKdeKwinServerDecoration* pMgr) { PROTO::serverDecorationKDE->destroyResource(this); });
     m_resource->setOnDestroy([this](COrgKdeKwinServerDecoration* pMgr) { PROTO::serverDecorationKDE->destroyResource(this); });
     m_resource->setRequestMode([this](COrgKdeKwinServerDecoration*, uint32_t mode) {
+        return; // stop eat my CPU
         auto sendMode = kdeModeOnRequestCSD(mode);
         m_resource->sendMode(sendMode);
         mostRecentlySent      = sendMode;
```

tannerellen:
> yeah this PR made 1 core pinned to 100% when firefox runs doesnt matter what you do...> 

The high cpu usage is also affecting Thunderbird too after this merge.

vaxerski:
https://github.com/hyprwm/Hyprland/issues/11591

PR #11556, opened 2025-08-29
allow workspace swipe wrap around

CyrenArkade:
When `workspace_swipe_create_new` and `workspace_swipe_use_r` are false, allow the workspace swipe gesture to wrap around from the first to last and last to first workspaces  as would be expected of `m-1`/`m+1`. This could be changed into its own option, but the previous behavior feels more like a bug since `m-1` and `m+1` wrap around but telling Hyprland to use these resulted in no wrapping.

Should this be documented on the wiki?


vaxerski (commented):
it should be documented and added a test and a setting for. I for example dont want this

pointer-scroll.txt
vaxerski, line 1: oops?

src/managers/input/UnifiedWorkspaceSwipeGesture.cpp
```diff
@@ -34,7 +34,7 @@ void CUnifiedWorkspaceSwipeGesture::update(double delta) {
         return;
 
     static auto  PSWIPEDIST             = CConfigValue<Hyprlang::INT>("gestures:workspace_swipe_distance");
-    static auto  PSWIPENEW              = CConfigValue<Hyprlang::INT>("gestures:workspace_swipe_create_new");
+    static auto  PSWIPEBOUND            = CConfigValue<Hyprlang::INT>("gestures:workspace_swipe_bound");
```
vaxerski, line 37: reason for breaking?

pointer-scroll.txt
CyrenArkade: That is indeed an oops, I swear I removed that.

src/managers/input/UnifiedWorkspaceSwipeGesture.cpp
CyrenArkade: There needs to be three states: One for doing nothing, one for creating a new workspace, and one for wrapping. If I were to add a `workspace_swipe_wrap` option, what would the behavior be when both are true?

vaxerski: overriding _new?

PR #11564, opened 2025-08-31
window: treat maximize as toggle request

jmanc3:
This PR changes the meaning of maximize requests from clients to be "toggle the current maximize state", which is something that we (as the compositor) know, but the application does not.

This fixes the following behavior where it's possible to be stuck with a maximized window, if the user doesn't have a binding to the fullscreen toggle, or doesn't even realize why the window won't demaximize.

https://github.com/user-attachments/assets/bf4bfec5-e5c8-4092-bb20-048eb307e62f

And it throttles requests at 100ms because some apps I've tried out like firefox can misbehave and send two instant request rather than one.

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1466,13 +1466,11 @@ void CWindow::onUpdateState() {
     }
 
     if (requestsMX.has_value() && !(m_suppressedEvents & SUPPRESS_MAXIMIZE)) {
-        if (requestsMX.has_value()) {
-            if (m_isMapped) {
-                auto window    = m_self.lock();
-                auto state     = sc<int8_t>(window->m_fullscreenState.client);
-                bool maximized = (state & sc<uint8_t>(FSMODE_MAXIMIZED)) != 0;
-                g_pCompositor->changeWindowFullscreenModeClient(window, FSMODE_MAXIMIZED, !maximized);
-            }
+        if (m_isMapped) {
+            auto window    = m_self.lock();
+            auto state     = sc<int8_t>(window->m_fullscreenState.client);
+            bool maximized = (state & sc<uint8_t>(FSMODE_MAXIMIZED)) != 0;
+            g_pCompositor->changeWindowFullscreenModeClient(window, FSMODE_MAXIMIZED, !maximized);
         }
     }
 }


src/Compositor.cpp
```diff
@@ -2286,9 +2286,24 @@ bool CCompositor::workspaceIDOutOfBounds(const WORKSPACEID& id) {
 }
 
 void CCompositor::changeWindowFullscreenModeClient(const PHLWINDOW PWINDOW, const eFullscreenMode MODE, const bool ON) {
-    setWindowFullscreenClient(
-        PWINDOW,
-        sc<eFullscreenMode>(ON ? sc<uint8_t>(PWINDOW->m_fullscreenState.client) | sc<uint8_t>(MODE) : (sc<uint8_t>(PWINDOW->m_fullscreenState.client) & sc<uint8_t>(~MODE))));
+    auto       state = sc<int8_t>(PWINDOW->m_fullscreenState.client);
```
vaxerski: this is _absolutely_ the wrong place for this. If the app uses xdg_shell for it, that should be used.

jmanc3: moving it to 

```c++
void CWindow::onUpdateState() {
    std::optional<bool>      requestsFS = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsFullscreen : m_xwaylandSurface->m_state.requestsFullscreen;
    std::optional<MONITORID> requestsID = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsFullscreenMonitor : MONITOR_INVALID;
    std::optional<bool>      requestsMX = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsMaximize : m_xwaylandSurface->m_state.requestsMaximize;
```
should be good i think  (as you were right that foreigntopwlr was using this function as well)

jmanc3:
And btw if you were wondering if this change is sound, here are the four majors consumers of the API treating maximize request as a toggle

WAYLAND via firefox
QT WAYLAND via pavucontrol/kcalc
GTK WAYLAND via lutris
X11 via chromium


https://github.com/user-attachments/assets/04ad8680-c3c0-4ebb-8a2a-ad69053a089c

vaxerski:
Yeeeahhh, we're breaking the spec here, but it's due to us always sending maximized to apps I guess, this is fine

Vladimir-csp:
> Yeeeahhh, we're breaking the spec here, but it's due to us always sending maximized to apps I guess, this is fine

Offtopic: is there an option to disable this behavior (lying to apps about maximized state)?

I have this in my config, is this enough?

```
# always sync fullscreen states
windowrule = fullscreenstate 0 0, fullscreenstate:0
windowrule = fullscreenstate 2 2, fullscreenstate:2
windowrule = fullscreenstate 3 3, fullscreenstate:3
```

vaxerski:
> Offtopic: is there an option to disable this behavior (lying to apps about maximized state)?

No

Vladimir-csp:
Oh,  there is `syncfullscreen` rule. Is this an equivalent, or something is still missing?

```
windowrule = syncfullscreen on, class:.*
```

vaxerski:
no, you literally cant

kolayne:
> no, you literally cant

@vaxerski, Is it a design choice that it's impossible? Why?

vaxerski:
because we tell apps they are always maximized

kolayne:
Yeah, my question is: why is it not configurable? Is there a good reason why Hyprland user should never want to tell windows the truth?

This might be what causes the bug with recent chromium: the browser window, apparently, attempts to unmaximize itself when exiting fullscreen, but instead ends up maximized...

vaxerski:
yes, apps will render non-window csd otherwise and it will mess up rendering + positioning + input

PR #11572, opened 2025-09-02
screencopy: implement transparency for non-shareable windows

arrowpc:

Makes `noscreenshare` windows transparent in screencopy instead of drawing black boxes.

This was previously rejected in #10390 because “making the window disappear” required a full rerender (too slow/ugly tradeoff), so black boxes were chosen instead. This PR avoids full rerenders by using [MRT](https://en.wikipedia.org/wiki/Multiple_Render_Targets) so we don’t pay that cost.

- color attachment 0 = regular scanout
- color attachment 1 = capture buffer

When a window has `noscreenshare` and is visible on the target monitor, we mask its writes only on the capture attachment. 

- If MRT isn’t supported by the GPU/driver, it falls back to the existing black box impl
- MRT path only kicks in when screencopy is pending and a `noscreenshare` window is visible on that monitor, otherwise nothing changes 

- Basic testing done:
  - dragging/sending across monitor/workspaces, overlapping windows
  - on a NVIDIA GTX 1080 and AMD Radeon 780M iGPU
  - ~32 MiB VRAM overhead observed when  `noscreenshare` windows were on a captured 4K monitor

I hope so >.<


Zebra2711:
> #### Is it ready for merging, or does it need work?
> I hope so >.<

well, this breaks a lot of rendering stuff

[output.webm](https://github.com/user-attachments/assets/f6c22337-80e7-476b-bbeb-a62ec9d3b4a6)

record with obs

arrowpc:
yeah I just retested on another machine and it's extremely buggy. Definitely not ready for merging at all I'll have to keep working on it

vaxerski:
you know that MRT is exactly the same as rendering twice, right?

arrowpc:
Maybe I misunderstood your original issue then? I thought you meant a whole second render pass was too much. With this approach we're running the frag shader once per draw and writing the result to multiple color attachments in the same pass. The extra cost is mainly additional color writes/blend to the second attachment, and not doubling draw submissions, state changes, resampling, etc. Yes it's not as cheap as drawing a black box over the `noscreenshare` windows, but it's still much cheaper than a second render pass no? I could've worded the PR better that's my bad.

vaxerski:
I mean yeah its cheaper but it's still additional rendering, and what's the benefit, exactly? I kinda think I'm missing the point.

I'd even say skipping the rendering is actually more confusing - maybe replace surfaces with black, so that we still _see_ the windows, they're just black.

Anyways, I can't really see because you forgot to commit `capture.glsl` it seems?

arrowpc:
The benefit to me (and the motivation to work on this) is niche to be fully honest. I was building a cross-platform windowing app that relied on the main window being fully transparent in capture which worked fine when I implemented it for macOS & Windows as their window exclusion API makes the windows disappear. Hyprland seems like the most progressive Wayland compositor and what I daily-drive on my NixOS machine so I thought why not write a PR. In hindsight you're right I agree that it is more confusing this way. Embarrassingly my thought process did not extend beyond, "well macOS and Windows do it, and it would selfishly help with my project, that must mean it's the better approach."

arrowpc:
> Anyways, I can't really see because you forgot to commit capture.glsl it seems?

I planned too far ahead with that one, didn't end up needing it so oops! (it was injected at runtime anyway so that shouldn't have been the cause of any issues)

arrowpc:
I fixed the issues I was seeing on my end and think it works fine now. @Zebra2711 if you could test again that would be greatly appreciated. 

And @vaxerski thoughts on letting people choose the way they wanna hide the windows as a compromise? Or if this feature is untenable that's okay too.

Zebra2711:
> I fixed the issues I was seeing on my end and think it works fine now. @Zebra2711 if you could test again that would be greatly appreciated.
> 

It worked with the default hyprland config, but still have some bugs. The hidden window shows up when closed, so I think it could be caused by animation. also till have render bug.

[2025-09-03 12-35-51.webm](https://github.com/user-attachments/assets/9b24b9b0-269a-4a90-a67c-8a92fe2bcf49)

arrowpc:
> also still have render bug.

I can't reproduce that. What wallpaper daemon are you using?

Zebra2711:
> I can't reproduce that. What wallpaper daemon are you using?

I disabled the default wallpaper(hyprland girl) in my hyprland config. It's an empty screen without wallpaper daemon

arrowpc:
> I disabled the default wallpaper(hyprland girl) in my hyprland config. It's an empty screen without wallpaper daemon

I did not consider this case. That should be working now

Zebra2711:
Nice, seem like worked smoothly now. Only have big problem is this will break hdr setup, in my case is option I commented
```ini
monitorv2 {
    output = eDP-1
    mode = 1920x1080@144
    position = 0x0
    scale = 1
    cm = srgb
    # sdrbrightness = 2.20
    # sdrsaturation = 1.07
    # supports_wide_color = 1
    # sdr_min_luminance = 0.005
    # sdr_max_luminance = 200
    # min_luminance = 0
    # max_luminance = 1000
    # max_avg_luminance = 200
}
```
And also impact to unstable config option `render:new_render_scheduling = true` which is disabled on default. But not sure thought, cuz i apply a bunch of patches in my hyprland build so it could be conflict thought.
Edit: only need to fix hdr issue, `new_render_scheduling` is early broken. And till have window when close, but it is like transparent 95% so hard to see

arrowpc:
@Zebra2711 the HDR issue should be fixed now. Been busy so couldn't work on this soz

> And till have window when close, but it is like transparent 95% so hard to see

I do not believe any toplevel/window content leaks.  Just the fadeout animation should be visible. Issue with my approach is that I wasn't able to figure out how to make it so on fadeout the screen content stays on the scanout texture. I could make it so no fadeout happens at all for `noscreenshare` windows but that would only make sense if I also figured out a way to make it appear on the scanout texture.

arrowpc:
@vaxerski is there hope for this PR with maybe an option for transparent/disappearance vs blackout or is it untenable?

vaxerski:
I am unsure. I'd need to see some benchmarks of MRT to justify its worth, as it's going to be a lot of added complexity to the renderer codewise.

In general if its performant and you make it turn the window's surfaces black, I'm ok

Zebra2711:
Performance is bad on my intel igpu, so I think this should be configurable or auto disabled by default if hyprland is running on igpu

arrowpc:
@Zebra2711 can you pls check if perf is still bad? On my AMD iGPU (M780) the latency actually improves with MRT compared to the current approach for me but GPU util goes up as more work is done in parallel but balances out if I'm doing actual work. I'll try to post benchmark results soon

Zebra2711:
now, it is all just a black box, regardless of the value I assign to `render:capture_mrt_mode` :)

arrowpc:
@Zebra2711 test with the noscreenshare rule. i just do smth like `hyprctl keyword windowrulev2 'noscreenshare opactiy:0.5 decorations:on, class:^(kitty)$'`for now. I made it default to 1.0 opacity with decorations on. `render:capture_mrt_mode` is there to test with/without MRT by setting either 1 or 2

arrowpc:
https://github.com/user-attachments/assets/ae063ba5-728f-4e60-ab20-6234759c2d32

@vaxerski v crude benchmarking

First half of the video is benchmarking the current implementation and the latter is for MRT. 

The mean time per call for `renderMon` seems to actually go down with MRT but GPU utililsation is higher. My guess it's because the GPU is doing more work in parallel rather than waiting for the final frame then drawing over it.

When I run something more graphically intensive like a https://testufo.com window on Firefox the GPU utililisation becomes quite equal(ly really high, ~69% vs ~71%).

Test was done on an AMD M780 iGPU 2048x1280@120.

I tried it on my other old GTX 1080 machine and the story is pretty much the same but with much much lower GPU util as it's a more powerful dedicated GPU.

Zebra2711:
> First half of the video is benchmarking the current implementation and the latter is for MRT.
> 
> The mean time per call for `renderMon` seems to actually go down with MRT but GPU utililsation is higher. My guess it's because the GPU is doing more work in parallel rather than waiting for the final frame then drawing over it.
> 
> When I run something more graphically intensive like a https://testufo.com window on Firefox the GPU utililisation becomes quite equal(ly really high, ~69% vs ~71%).
> 
> Test was done on an AMD M780 iGPU 2048x1280@120.
> 
> I tried it on my other old GTX 1080 machine and the story is pretty much the same but with much much lower GPU util as it's a more powerful dedicated GPU.

I have a quick test here.

[2025-09-26 16-35-21.webm](https://github.com/user-attachments/assets/8735038d-7a44-4f3a-a836-2ded0ee33948)

Seem like no transparency anymore :(.

arrowpc:
> Seem like no transparency anymore :(.

@Zebra2711 `windowrule = noscreenshare opacity:0.0 decorations:off, class:^(octopi)$`

Zebra2711:
> > Seem like no transparency anymore :(.
> 
> @Zebra2711 `windowrule = noscreenshare opacity:0.0 decorations:off, class:^(octopi)$`

i don't have any performance issues when running hyprland on the igpu with this MR anymore, but this window rule no longer works from MR `11689`

arrowpc:
yeah makes sense, it's a hacked together proof of concept instead of a proper config option/rule. 

I think if this PR goes through the ability to set the opacity and toggling decorations is deffo the way to go. But the actual way to define and go about it is not really worth putting time into till I get the ok from @vaxerski / more direction on how to approach it

vaxerski:
tbh to me everything past "replace this window's surfaces with white / transparent" is overkill.

arrowpc:
fair enough. It's 100% overkill but doesn't really complicate the code much further than a discrete "blackout vs disappear" option. What does add complexity is the ability to individually set that blackout/disappear distinction between layers & windows. It doesn't make sense to remove the latter and keep the former so inadvertently it does end up adding complexity. 

Making it universal is much simpler and could just be a `registerConfigVar("misc:screencopy_noscreenshare_visibility", Hyprlang::INT{1});`

I can work on that if that's what you think is best @vaxerski (if tagging you directly is not the way to go pls lmk)

vaxerski:
tagging doesnt change anything cuz I ignore tags and just go through my inbox

Universal sounds ok to me

arrowpc:
ok i cleaned mostly everything up and removed the fallback/prev impl for blacking out as after doing research literally no one will hit the fallback unless they're on a low-end <2007 laptop that can't run Hyprland anyway

Zebra2711:
@arrowpc need rebase :)

arrowpc:
@Zebra2711 done ty. idk if I should routinely check for conflicts and update or just wait for @vaxerski's review

Zebra2711:
@arrowpc i found a new bug with walker :)

```ini
layerrule = noscreenshare,^(walker)$
```
https://github.com/user-attachments/assets/4682b420-bb7c-4f6f-bf02-e2fe90dab9e5

arrowpc:
@Zebra2711 fixed ty

Zebra2711:
@arrowpc need rebase against (conflicts with [`c330d433`](https://github.com/hyprwm/Hyprland/commit/c330d4334f0f14ea46fdcafc164faeec09ea82e9)), also tag varxy too :)

arrowpc:
well he said tagging is useless but whatevs, @vaxerski PR is ready if you weren't already aware

Zebra2711:
@arrowpc need rebase against :)

arrowpc:
i've been way too busy I will have free time this weekend

arrowpc:
I completely lied I underestimated life. Think I'll wait till important screencopy PRs are merged before I resolve conflicts

Zebra2711:
ping @arrowpc

arrowpc:
I did see the merge went through I'll work on this most likely this week ty for the ping

arrowpc:
probably a sloppy rebase, I didn’t do much thinking about how to make it fit better with the new code and mostly frankensteined it to get rid of conflicts.

I know that `misc:screencopy_noscreenshare_visibility` is a horrible name that needs to be changed

and that preserving alpha for blacked out `no_screen_share` content is a horrible idea because some layers/windows make the background translucent with text fully opaque which can make sensitive text readable in capture (also will show the silhouette of icons)

I wanted to preserve the alpha because it made rounded layers (launchers & bars) look much better but o well.

Zebra2711:
Flickering on non-shared windows/layers. Worked fine without this PR

https://github.com/user-attachments/assets/65e8d3f7-6eff-479e-9fd5-b403874aa743

https://github.com/user-attachments/assets/dbee4357-0ff5-4157-9797-707057a3b727

arrowpc:
@Zebra2711 try again 🙏

Zebra2711:
> @Zebra2711 try again 🙏

This still isn’t fixed

> https://github.com/user-attachments/assets/dbee4357-0ff5-4157-9797-707057a3b727

misc:screencopy_noscreenshare_visibility = 1

Zebra2711:
Okay, it’s good now :+1:

Zebra2711:
@arrowpc need rebase against. It’s hard to keep the patch updated when the code changes so fast :(

Dhruvpatel-10:
Hey, just wanted to chime in as someone who really needs this feature. I do technical interviews over Google Meet with full-screen share, and I need my interview terminal hidden from the capture. The current black rectangle approach is immediately visible and defeats the purpose.

I ended up writing my own local patch that does a full re-render excluding `no_screen_share` windows — it works great but I understand why the MRT approach here is better for performance. 

This is a real gap compared to macOS (`NSWindow.sharingType = .none`) and Windows (`SetWindowDisplayAffinity`), both of which make windows fully invisible in capture. Hyprland is the most progressive Wayland compositor and this would bring it to parity.

Would love to see this merged and released. Happy to help test if needed.

arrowpc:
> Hey, just wanted to chime in as someone who really needs this feature. I do technical interviews over Google Meet with full-screen share, and I need my interview terminal hidden from the capture. The current black rectangle approach is immediately visible and defeats the purpose.
> 
> I ended up writing my own local patch that does a full re-render excluding `no_screen_share` windows — it works great but I understand why the MRT approach here is better for performance.
> 
> This is a real gap compared to macOS (`NSWindow.sharingType = .none`) and Windows (`SetWindowDisplayAffinity`), both of which make windows fully invisible in capture. Hyprland is the most progressive Wayland compositor and this would bring it to parity.
> 
> Would love to see this merged and released. Happy to help test if needed.

Yes sorry been way too busy I resonate 100%. @vaxerski's a busy duder ;p

salah-chafai:
Awesome work, @arrowpc thanks!

I hit one issue: I have a window with the no_screen_share rule that I toggle via a keybind. When the window unmaps, the screen's background goes black and stays that way
until I switch workspace or do anything else that triggers a redraw.

One caveat for you: I see the same black flash even with the rule removed and misc:screencopy_noscreenshare_visibility = 1 (default), so this might actually live in the main
this branch sits on (the merge-base is ~350 commits behind current main), not in the PR itself. A rebase would tell us, happy to retest after.

arrowpc:
I have more free time to work on this now and be responsive as personal issues are more or less dealt with

I think, at the very least temporarily, I will refactor the PR to minimise conflicts such that it can be less prone to conflicts and should hopefully be easier to review

I have not read the changes done to the renderer since `ext_image_copy_capture_v1` was merged so bear with me

@salah-chafai thank you, I'm not too sure what would be causing that at the moment cause I forgot lots of the details but it's probably from a rebase I didn't put much thought into months ago :D

(Also want to add a disclaimer that I will be finally trying out LLMs to see if they're actually useful now for things such as creating a summary of the relevant commit history from this PR to main)

john-griffin:
Good work, looking forward to it.

PR #11608, opened 2025-09-06
presentation: handle vrr for v1 clients

njdom24:
The potential for an issue was mentioned in https://github.com/hyprwm/Hyprland/pull/11306, but it turns out that Gamescope is a notable `wp_presentation_v1`-only client that expects `refresh` to be 0 in the VRR case ([ref](https://github.com/ValveSoftware/gamescope/blob/cf288b95fa376a15f30fe8d1a9f750cad54742df/src/Backends/WaylandBackend.cpp#L1720)).

Resolves https://github.com/hyprwm/Hyprland/discussions/11406

I opened an uglier/less-likely-to-be-accepted PR to support V2 on gamescope [here](https://github.com/ValveSoftware/gamescope/pull/1963), but V1 is more reliable for communicating VRR status anyway.

I'm new to Wayland protocols and contributing to Hyprland, so I'm open to criticism.

Works on my machine with `gamescope --backend wayland -f -H 1440 -h 1440 -W 2560 -w 2560 --adaptive-sync -r 180 -- env MANGOHUD_CONFIG="no_display,fps_limit=133" vkgears`

### DIFF
--- a/src/protocols/PresentationTime.cpp
+++ b/src/protocols/PresentationTime.cpp
@@ -64,8 +64,10 @@ void CPresentationFeedback::sendQueued(WP<CQueuedPresentationData> data, const T
     if (sizeof(time_t) > 4)
         tv_sec = TIMESPEC.tv_sec >> 32;
 
+    uint32_t refreshNs = m_resource->version() == 1 && data->m_monitor->m_vrrActive ? 0 : untilRefreshNs;
+
     if (data->m_wasPresented)
-        m_resource->sendPresented(sc<uint32_t>(tv_sec), sc<uint32_t>(TIMESPEC.tv_sec & 0xFFFFFFFF), sc<uint32_t>(TIMESPEC.tv_nsec), untilRefreshNs, sc<uint32_t>(seq >> 32),
+        m_resource->sendPresented(sc<uint32_t>(tv_sec), sc<uint32_t>(TIMESPEC.tv_sec & 0xFFFFFFFF), sc<uint32_t>(TIMESPEC.tv_nsec), refreshNs, sc<uint32_t>(seq >> 32),
                                   sc<uint32_t>(seq & 0xFFFFFFFF), sc<wpPresentationFeedbackKind>(flags));
     else
         m_resource->sendDiscarded();


Justsnoopy30:
This PR appears to cause a regression. I tracked it down to this presentation commit, and on my machine, it causes chrome and chromium-based software to tank their framerate and hover around 60hz for the rest of the session when vrr is enabled (my monitors are 240hz and 165hz).

Also, it happens on my non-vrr monitor too if the default for vrr is set to 1 and you:
1. Open or close a special workspaces
2. Make a change to hyprland config that causes a reload

Then it seems to think that the non-VRR monitor is VRR and the bug occurs for the rest of the session unless VRR is explicitly set to 0 for the monitor.

I'm running chrome in native wayland mode and I tested on both my Intel igpu and the nvidia dgpu. I also checked with just one monitor enabled connected to the igpu - same result.

No idea why.

vaxerski:
can you post `hyprctl systeminfo`

Justsnoopy30:
```Hyprland 0.51.0 built from branch main at commit 559024c3314e4b1180b10b80fce4e9f20bad14c8  (gestures/float: fix typo).
Date: Sat Sep 13 19:52:41 2025
Tag: v0.51.0-8-g559024c3, commits: 6426
built against:
 aquamarine 0.9.4
 hyprlang 0.6.3
 hyprutils 0.8.4
 hyprcursor 0.1.13
 hyprgraphics 0.1.5


no flags were set


System Information:
System name: Linux
Node name: gaming-laptop
Release: 6.16.7-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Thu, 11 Sep 2025 17:42:36 +0000


GPU information: 
0000:00:02.0 VGA compatible controller [0300]: Intel Corporation Raptor Lake-S UHD Graphics [8086:a788] (rev 04) (prog-if 00 [VGA controller])
0000:01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD104M [GeForce RTX 4080 Max-Q / Mobile] [10de:27e0] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  580.82.09  Fri Aug 29 17:44:50 UTC 2025


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
LOGO=archlinux-logo


plugins:
  Hyprtasking by raybbian ver 0.1

Explicit sync: supported
GL ver: 3.2
Backend: drm

Monitor info:
	Panel HDMI-A-1: 1920x1080, HDMI-A-1 Microstep MSI MAG273R2 CB6H191400474 -> backend drm
		explicit ✔️
		edid:
			hdr ✔️
			chroma ✔️
			bt2020 ✔️
		vrr capable ❌
		non-desktop ❌
		
	Panel eDP-1: 2560x1600, eDP-1 BOE 0x0B17  -> backend drm
		explicit ✔️
		edid:
			hdr ❌
			chroma ✔️
			bt2020 ❌
		vrr capable ✔️
		non-desktop ❌
```
(no plugins were enabled during testing to be clear)

vaxerski:
20250915_22h40m33s

comment was right

try this 
[patch.txt](https://github.com/user-attachments/files/22351714/patch.txt)

Justsnoopy30:
Yep, this fixes the 2nd issue with it applying to the non-VRR screen.

vaxerski:
I commited that patch then, what issue remains?

Justsnoopy30:
On the monitor *with* VRR support, the same issue remains where the framerate is much lower and hovers around 60fps.

tannerellen:
> I commited that patch then, what issue remains?

Not OP but I can reproduce the issue that remains after the patch you pushed. I'm on the latest git and before and after your patch if VRR is enabled chrome / chromium based apps get capped at 60fps. If you disable VRR the cap is not present.

It's easy to reproduce. Just enabled VRR and go to a site like https://frameratetest.com/refresh-rate-tester/ On a chromium based browser it won't go higher than 60fps. On Firefox it will go as high as your refresh rate allows.

Edit: here is a screenshot with chrome on the left and firefox on the right with VRR enabled:
image

vaxerski:
I think this would be a chromium bug?

tannerellen:
> I think this would be a chromium bug?

I can bisect the issue to this commit. Before this commit no problem chrome is able to max out my refresh rate. After this commit then it doesn't go higher than 60hz (give or take). Looks like this issue is also reported on this discussion: https://github.com/hyprwm/Hyprland/discussions/11729#discussioncomment-14420358

So if it's a chromium bug something about this commit is making it happen.

vaxerski:
well this commit fixes a behavior thats out of spec.

20250916_21h16m15s

to me VRR does mean there is no constant refresh rate, and since chromium binds v1, it gets 0.

tannerellen:
Yep, that definitely seems like what is happening. I guess this is on chromium then.

tannerellen:
Chromium bug confirmed: https://issues.chromium.org/issues/410071538

Justsnoopy30:
This would probably be a good reason to make the novrr window rule work when vrr = 1, so that users can work around application bugs like this without making VRR only usable in fullscreen. It doesn't look like the chromium bug is going anywhere anytime soon.

Issue #11604, opened 2025-09-05
Exclude reserved area from popup unconstrain


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11601

## Description
it can lead to wonkiness, we should exclude reserved

PR #11615, opened 2025-09-07
xdg: exclude reserved area from popup unconstrain (use work area)

ritesh006:
Context
-------
XDG popups (e.g., Chromium context menu) were unconstrained against the *full*
monitor bounds, so they could overlap bars/panels that reserve space via
exclusive zones. See #11604.

What this changes
-----------------
- In `CPopup::reposition()`, the unconstrain region passed to
  `applyPositioning(...)` is now the monitor **work area**:
  `monitor.position + reservedTopLeft` ..
  `monitor.size - reservedTopLeft - reservedBottomRight`.
- In `CPopup::initAllSignals()`, listen for `monitor.modeChanged` and call
  `reposition()` so active popups re-clamp after resolution/scale changes.

Why it’s safe / scope
---------------------
- IME popups are a separate path (unchanged).
- Change is localized to `desktop/Popup.*`.
- No behavior change to regular windows; only xdg-popups’ unconstrain bounds.

User-visible effect
-------------------
Popups no longer overlap exclusive zones (top/bottom bars). If the bar size or
monitor mode changes while a popup is open, the popup repositions to remain
inside the usable area.

Testing
-------
Manual on Wayland:
1) `hyprctl keyword monitor eDP-1,addreserved,40,0,0,0`
2) Right-click near the top edge in Chromium → popup stays below the bar.
3) Change scale/resolution (`hyprctl keyword monitor eDP-1,preferred,auto,1.25`)
   → popup re-snaps within work area.
4) Repeat with reserved on other edges.

Notes
-----
- If the tree exposes a dedicated workarea/reserved-changed signal on the
  monitor, we can additionally hook it to call `reposition()`. This patch keeps
  it minimal by using `modeChanged`, which already covers common changes.

### DIFF
--- a/src/desktop/Popup.cpp
+++ b/src/desktop/Popup.cpp
@@ -85,6 +85,13 @@ void CPopup::initAllSignals() {
     m_listeners.destroy    = m_resource->m_surface->m_events.destroy.listen([this] { this->onDestroy(); });
     m_listeners.commit     = m_resource->m_surface->m_events.commit.listen([this] { this->onCommit(); });
     m_listeners.newPopup   = m_resource->m_surface->m_events.newPopup.listen([this](const auto& resource) { this->onNewPopup(resource); });
+
+    // Keep the popup within the work area when the monitor mode/scale changes
+    if (auto mon = g_pCompositor->getMonitorFromVector(t1ParentCoords())) {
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
we do not accept AI submissions

PR #11629, opened 2025-09-08
layout/dwindle: add presets

vaxerski:
Adds layout presets for dwindle. Available as both a layoutmsg to snap, and as a workspace rule.

syntax:
```ini
preset = dwindle, [name], [rules...]
```

For example:
```ini
preset = dwindle, myPreset, horizontal 33% right, horizontal 50% right, vertical 50% right, up up up left vertical 50% right
```
For creating an H shaped layout.

you can also specify windows for slots and add a root slot:
```ini
preset = dwindle, myPreset, root class:amogus, horizontal 33% right, horizontal 50% right class:amogus2, vertical 50% right, up up up left vertical 50% right
```

Workspace rule:
```ini
workspace = 1, layoutopt:preset one
```

layoutmsg (dwindle):
```sh
hyprctl dispatch layoutmsg preset one
```

https://github.com/user-attachments/assets/8db5a597-ad57-4c72-8231-24fc21203ab1

https://github.com/user-attachments/assets/2646ba55-f902-4010-a3e9-b6a94cd70ce3

TODO: 
- [ ] testing
- [x] tests

### DIFF
--- a/hyprtester/src/tests/main/dwindle.cpp
+++ b/hyprtester/src/tests/main/dwindle.cpp
@@ -1,0 +1,96 @@
+#include "tests.hpp"
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
+#include <thread>
+#include <chrono>
+#include <hyprutils/os/Process.hpp>
+#include <hyprutils/memory/WeakPtr.hpp>
+#include "../shared.hpp"
+
+static int ret = 0;
+
+using namespace Hyprutils::OS;
+using namespace Hyprutils::Memory;
+
+#define UP CUniquePointer
+#define SP CSharedPointer
+
+static bool test() {
+    NLog::log("{}Testing dwindle", Colors::GREEN);
+
+    OK(getFromSocket("/dispatch workspace 101"));
+
+    // test the layout preset
+
+    Tests::spawnKitty();
+
+    {
+        auto str = getFromSocket("/clients");
+        EXPECT_COUNT_STRING(str, "at: 22,22", 1);
+        EXPECT_COUNT_STRING(str, "size: 1876,1036", 1);
+    }
+
+    Tests::spawnKitty();
+
+    {
+        auto str = getFromSocket("/clients");
+        EXPECT_COUNT_STRING(str, "at: 22,22", 1);
+        EXPECT_COUNT_STRING(str, "size: 605,1036", 1);
+
+        EXPECT_COUNT_STRING(str, "at: 641,22", 1);
+        EXPECT_COUNT_STRING(str, "size: 1257,1036", 1);
+    }
+
+    Tests::spawnKitty();
+
+    {
+        auto str = getFromSocket("/clients");
+        EXPECT_COUNT_STRING(str, "at: 22,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 641,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 1284,22", 1);
+
+        EXPECT_COUNT_STRING(str, "size: 605,1036", 1);
+        EXPECT_COUNT_STRING(str, "size: 629,1036", 1);
+        EXPECT_COUNT_STRING(str, "size: 614,1036", 1);
+    }
+
+    Tests::spawnKitty();
+
+    {
+        auto str = getFromSocket("/clients");
+        EXPECT_COUNT_STRING(str, "at: 22,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 641,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 1284,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 1284,547", 1);
+
+        EXPECT_COUNT_STRING(str, "size: 605,1036", 1);
+        EXPECT_COUNT_STRING(str, "size: 629,1036", 1);
+        EXPECT_COUNT_STRING(str, "size: 614,511", 2);
+    }
+
+    Tests::spawnKitty();
+
+    {
+        auto str = getFromSocket("/clients");
+        EXPECT_COUNT_STRING(str, "at: 22,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 641,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 1284,22", 1);
+        EXPECT_COUNT_STRING(str, "at: 1284,547", 1);
+        EXPECT_COUNT_STRING(str, "at: 22,547", 1);
+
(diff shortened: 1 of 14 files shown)


Dregu:
I've been testing this and trying to replicate features of my nstack fork. This really gives a lot of options for ultrawide and mixing different aspect ratios. I have a 16:9, 21:9 and 21:9 portrait setup, and all of those really need a different amount of columns. I can now easily have a default per monitor and temporarily swap to a different layout with binds. Really like how this autotiles and then undwindles when closing windows. Setting specific classes to root automatically is also neat.

Ran into few issues:

- single_window_aspect_ratio needs to be monitor or actually workspace specific, my portrait monitor looks very silly with only one window (gapsout workspace rule could be used for this, or any amount of windows for that matter, **if it supported css style gaps**...)
- the dynamic presets (see portrait config) are not updating when closing windows (going from 4 to 3 windows still uses preset p4, should use p3 cause there are only 3 windows now)
- using layoutmsg preset requires me to focus another window to apply when 2 windows on workspace (I dunno, it's weird, but works with 3 or more windows just fine)
- still very confused by the konami code, kinda just tried random stuff until it worked...

Here's a test video of my setup and the relevant config:

https://github.com/user-attachments/assets/88a0e32b-475d-401f-8578-f0914bc5049f

```
dwindle:single_window_aspect_ratio = 14 9
$root = class:codium|firefox

bindd = SUPER, V, Dwindle: Default, layoutmsg, preset uw
bindd = SUPER CTRL, V, Dwindle: 50/50, layoutmsg, preset uw50
bindd = SUPER ALT, V, Dwindle: 3 columns, layoutmsg, preset uw3col
bindd = SUPER SHIFT, V, Dwindle: Flip, layoutmsg, preset uwflip

# wide, 2 stacks
preset = dwindle, w, root $root, horizontal 50% right, vertical 50% right, up up left vertical 50% right
workspace = m[DVI-D-1], layoutopt:preset w

# ultrawide, 3 stacks
preset = dwindle, uw, root $root, horizontal 65% right, vertical 50% right, up up left horizontal 50% right, vertical 50% right, up up up up left left vertical 50% right
preset = dwindle, uw50, root $root, horizontal 50% right, vertical 50% right, up up left horizontal 50% right, vertical 50% right, up up up up left left vertical 50% right
preset = dwindle, uw3col, root $root, horizontal 33% right, horizontal 50% right
preset = dwindle, uwflip, root $root, horizontal 33% left, horizontal 50% right, up up right horizontal 50% left, vertical 50% left, up up up up right right vertical 50% left
workspace = m[DP-1], layoutopt:preset uw

# portrait, 1 stack auto sizing
preset = dwindle, p2, vertical 50% right
preset = dwindle, p3, vertical 33.3333% right, vertical 50% right
preset = dwindle, p4, vertical 25% right, vertical 33.3333% right, vertical 50% right
workspace = m[HDMI-A-1]w[tv2], layoutopt:preset p2
workspace = m[HDMI-A-1]w[tv3], layoutopt:preset p3
workspace = m[HDMI-A-1]w[tv4], layoutopt:preset p4
```

Dregu:
More issues:

- using swapwindow and then opening more windows resets the whole layout to the original order
- editing running presets seems to not work, I have to restart hyprland or keep renaming my preset to test stuff
- I still have to mess with dwindle:split_width_multiplier to get 5 rows on the portrait monitor, and then that breaks the non-ultrawide monitor. the preset should override all automated guessing of what I wanted

vaxerski:
fixed check now

Dregu:
Yeah I believe you fixed some of the issues, took a while to guess which ones.

- dynamic `w[tv#]` presets seem to apply correctly when closing windows
- editing presets at runtime seems to work

However:

- the existing windows are still reordering and flying all over the place when opening more windows after using swapwindow and/or changing preset. even if browser is specifically forced root, it's somehow ending up the smallest "slave" very often when I close a terminal
- still being weird with exactly two windows, i.e. the correct preset is not applied after closing the third window and ending up with exactly two windows. like it resets to 50/50 until I focus the root window with mouse, which snaps it back to the correct 65/35 preset that I never even changed

Here's an example of the two window problem. I'm simply opening and closing windows, not changing the preset.

https://github.com/user-attachments/assets/8260607c-da86-4329-908c-b8ddc8cf0bc8

d0m1n1q0:
Just adding some insult to injury, although I am not sure if the following occurs because of "layout".

Applying the "desktop" solution as suggested [](https://wiki.hypr.land/Configuring/Uncommon-tips--tricks/#show-desktop)
works fine. But restoring from "desktop" the windows are reshuffled and with no focus.

vaxerski:
what do you mean "restoring"?

vaxerski:
the 2 window issue should be fixed.

> the existing windows are still reordering and flying all over the place when opening more windows after using swapwindow and/or changing preset. even if browser is specifically forced root, it's somehow ending up the smallest "slave" very often when I close a terminal

can you give an example?

Dregu:
> the 2 window issue should be fixed.

Indeed very good now. You also fixed the reordering issue accidentally. I also recently learned that the gapsout rule does support 4-way, just don't put commas there like the docs told you to do. Some issues just fix themselves...

> can you give an example?

Since I already recorded it, might as well show what I mean. Browser kept jumping back to its original place when opening/closing terminals. But like I said, this actually works fine now, windows are staying where they should after swapped.

https://github.com/user-attachments/assets/5fe8c056-f7e5-4aaf-8e86-886b61ce4788

### Issues still persisting:

- dynamic presets issue is back and the sizes once again don't update correctly when closing windows on the portrait monitor
- still need to carefully set `split_width_multiplier = 1.3` to achieve what I want on all monitors, and it's not a good solution
- layoutmsg still has the two window issue, i.e. all of the binds initially just do 50/50 until I wiggle the mouse to focus another window, which updates to the correct preset. works fine with three windows
- when moving a window to another workspace (on another monitor), the destination preset gets applied to the source workspace (also the preset doesn't update on the destination workspace)
- when a window opens on an inactive monitor (due to silent window rules), the preset from the inactive monitor is applied to the active monitor instead

d0m1n1q0:
"Restoring" means toggle to reveal the hidden clients.
The "desktop" works fine as I said.  As long as you stay on the workspace. Togging "desktop" hides all clients.
Change to another workspace and come back again to the "desktop-ed" workspace and then toggle again the "desktop" to un-hide the clients. They are not anymore in the layout that were "desktop-ed".

Best I can explain.
The "desktop" script is this
https://wiki.hypr.land/Configuring/Uncommon-tips--tricks/#show-desktop

wh1le:
Thanks for working on this feature. Looking forward to it. Missing flexible tiling setups for complex workflows.

PR #11657, opened 2025-09-10
Permission display binary path and line break.

xzn:
Actually display binary path for permission popup.

Also fixed a typo what looked like should be a line-break.

(Ready for merge)

### DIFF
--- a/src/managers/permissions/DynamicPermissionManager.cpp
+++ b/src/managers/permissions/DynamicPermissionManager.cpp
@@ -59,10 +59,10 @@ static const char* permissionToString(eDynamicPermissionType type) {
 
 static const char* permissionToHumanString(eDynamicPermissionType type) {
     switch (type) {
-        case PERMISSION_TYPE_UNKNOWN: return "An application <b>{}</b> is requesting an unknown permission.";
-        case PERMISSION_TYPE_SCREENCOPY: return "An application <b>{}</b> is trying to capture your screen.<br/><br/>Do you want to allow it to do so?";
-        case PERMISSION_TYPE_PLUGIN: return "An application <b>{}</b> is trying to load a plugin: <b>{}</b>.<br/><br/>Do you want to load it?";
-        case PERMISSION_TYPE_KEYBOARD: return "A new keyboard has been plugged in: {}.<br/><br/>Do you want to allow it to operate?";
+        case PERMISSION_TYPE_UNKNOWN: return "An application <b>{:?}</b> is requesting an unknown permission.";
+        case PERMISSION_TYPE_SCREENCOPY: return "An application <b>{:?}</b> is trying to capture your screen.<br/><br/>Do you want to allow it to do so?";
+        case PERMISSION_TYPE_PLUGIN: return "An application <b>{:?}</b> is trying to load a plugin: <b>{:?}</b>.<br/><br/>Do you want to load it?";
+        case PERMISSION_TYPE_KEYBOARD: return "A new keyboard has been plugged in: <b>{:?}</b>.<br/><br/>Do you want to allow it to operate?";
     }
 
     return "error";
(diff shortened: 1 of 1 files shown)


src/managers/permissions/DynamicPermissionManager.cpp
```diff
@@ -261,10 +261,10 @@ void CDynamicPermissionManager::askForPermission(wl_client* client, const std::s
 
         if (type == PERMISSION_TYPE_PLUGIN) {
             const auto LOOKUP = binaryNameForPid(pid);
-            description       = std::format(std::runtime_format(permissionToHumanString(type)), lookup, binaryPath);
+            description       = std::format(std::runtime_format(permissionToHumanString(type)), std::format("{}<br/> ({})", lookup, binaryPath));
         } else {
             const auto LOOKUP = binaryNameForPid(pid);
-            description       = std::format(std::runtime_format(permissionToHumanString(type)), lookup, binaryPath);
+            description       = std::format(std::runtime_format(permissionToHumanString(type)), std::format("{}<br/> ({})", lookup, binaryPath));
```
C0Florent: Just out of curiosity, how are non-printing characters displayed here?

Since the `binaryPath` shown is supposed to allow the user to actually check the file that is going to be to be executed, the displayed name should reflect precisely the actual path, displaying non-printing characters in some way

xzn: Good point. Would using `{:?}` debug formatting work?

src/managers/permissions/DynamicPermissionManager.cpp
```diff
@@ -248,8 +248,8 @@ void CDynamicPermissionManager::askForPermission(wl_client* client, const std::s
     if (binaryPath.empty())
         description = std::format(std::runtime_format(permissionToHumanString(type)), std::format("unknown application (wayland client ID 0x{:x})", rc<uintptr_t>(client)));
     else if (client) {
-        std::string binaryName = binaryPath.contains("/") ? binaryPath.substr(binaryPath.find_last_of('/') + 1) : binaryPath;
-        description            = std::format(std::runtime_format(permissionToHumanString(type)), std::format("{}</b> ({})", binaryName, binaryPath));
+        std::string binaryName = std::filesystem::path(binaryPath).filename().string();
+        description            = std::format(std::runtime_format(permissionToHumanString(type)), std::format("{:?}<br/>({:?})", binaryName, binaryPath));
```
vaxerski, line 252: I dont think this is a perfectly sound idea.
```
An application myApp
/usr/bin/myApp is trying to capture your screen
```

that looks... cursed?

C0Florent: I mean for such security-related stuff it's a valuable info to know for the user. We don't want to let arbitrary apps fool the user by setting "OBS" as their window title, right?

Perhaps better phrasing in the dialog window could help make it look a bit nicer

xzn: It's actually not window title, just the filename of the full binary path so it does look cursed, but idk good UI design :/

C0Florent: > It's actually not window title, just the filename of the full binary path so it does look cursed, but idk good UI design :/

I didn't see that, anyways I feel like having the full path somewhere should be better. Either a full path or a PID or something else that uniquely identifies an app would work.

It might a bit of a stretch, but imagine with the current setup some malicious app creates an executable somewhere (and thus chooses the executable file name) and runs it, and this executable makes the request, then forwarding the display capture to anything it wants.

This might be a stupid argument because when it reaches the point where an app can create and run an arbitrary executable you might be screwed already for anything security-related, idk.

vaxerski: what's wrong with the current design though? It shows both the path and the name...

PR #11658, opened 2025-09-10
keep top and overlay layer windows in presence of fullscreen window

xzn:
Currently layer shell windows with top/overlay layer are not shown in the presence of fullscreen windows under certain conditions. This fixes that.

(Ready for merge.)

P.S.: Similar issue for IME popup where it's not drawn when the text input is in a fullscreen window (so far I've had it happen in fullscreen games running under proton wayland) (**EDIT**: also in plain fullscreen SDL windows, with or without software rendering). Unfortunately I wasn't able to find the cause for this one. If anyone have any idea, helps are appreciated.

### DIFF
--- a/src/events/Windows.cpp
+++ b/src/events/Windows.cpp
@@ -697,7 +697,8 @@ void Events::listener_mapWindow(void* owner, void* data) {
     // recalc the values for this window
     g_pCompositor->updateWindowAnimatedDecorationValues(PWINDOW);
     // avoid this window being visible
-    if (PWORKSPACE->m_hasFullscreenWindow && !PWINDOW->isFullscreen() && !PWINDOW->m_isFloating)
+    if (PWORKSPACE->m_hasFullscreenWindow && !PWINDOW->isFullscreen() && !PWINDOW->m_isFloating &&
+        !(PWINDOW->m_wlSurface && PWINDOW->m_wlSurface->getLayer() && PWINDOW->m_wlSurface->getLayer()->m_layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP))
         PWINDOW->m_alpha->setValueAndWarp(0.f);
 
     g_pCompositor->setPreferredScaleForSurface(PWINDOW->m_wlSurface->resource(), PMONITOR->m_scale);
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -204,7 +204,9 @@ bool CHyprRenderer::shouldRenderWindow(PHLWINDOW pWindow, PHLMONITOR pMonitor) {
             return true;
 
         // if hidden behind fullscreen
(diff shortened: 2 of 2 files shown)


fufexan:
I'm pretty sure the top layer should not render over fullscreen windows.

xzn:
@fufexan 

https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_shell_v1:enum:layer

`Fullscreen shell surfaces are typically rendered at the top layer`

Also please read the existing code:

https://github.com/hyprwm/Hyprland/blob/main/src/render/Renderer.cpp#L932

https://github.com/hyprwm/Hyprland/blob/main/src/render/Renderer.cpp#L989

Top layers are already rendered above fullscreen windows. It's just that currently in some weird cases they are "optimized" out.

vaxerski:
no, top layers do not render above fullscreen windows. The current logic is if the surface has existed before the fullscreening, it's out. If it was after, it's in.

xzn:
> no, top layers do not render above fullscreen windows. The current logic is if the surface has existed before the fullscreening, it's out. If it was after, it's in.

If by out you mean "do not render" and by in you mean 'do render," I do not see anywhere in the code where it handles this current logic. The code I changed does not affect this either, it just makes sure windows don't become invisible when mapped during presence of fullscreen window and receives proper damage.

Currently in `renderAllClientsForWorkspace`, if fullscreen window is present, first fullscreen window, floating, etc are rendered, then top layers, IME popups, then overlay layers.

What I fixed is that currently sometimes newly created top layers will be made invisible, either that or it doesn't receive damage update, if a fullscreen window is already present.

vaxerski:
your patch makes no sense though `PWINDOW->m_wlSurface->getLayer()` is always null

xzn:
I just tested it again to make sure I wasn't hallucinating, only the change in `src/events/Windows.cpp` is needed to make the bug go away. I added the check in `src/render/Renderer.cpp` to make doubly sure since the fullscreen windows I tested are all game windows so they always receive full monitor dmg.

So no that check in particular isn't always null, or it wouldn't have fixed anything.

(Edit: could be it's always null on first map, or not, I haven't looked that much further, it also doesn't concern the code I changed anyhow)

vaxerski:
`pWindow->m_wlSurface->getLayer()` is _always_ null

xzn:
My bad the issue is already fixed in 81bf4ec

You are right my code was doing nothing.

(I did a merge somewhere without paying attention, among other things. That commit was literally one later than the one I was running and experiencing issue with. Gah)

vaxerski:
for future reference m_wlSurface is a `CWLSurface` which can only have one owner. Since it's a `pWindow->m_wlSurface`, the owner is a window, so it can't have a `getLayer()` owner because it's owned by a window

PR #11661, opened 2025-09-11
input: Correct tablet active_area calculation for rotated transforms

senotrusov:
#### Issue overview

When a user applies a 90- or 270-degree `transform` to a tablet, the `active_area_size` and `active_area_position` settings behave unintuitively. The normalization of these millimeter values always uses the tablet's native (un-rotated) physical dimensions.

This forces the user to manually calculate "pre-distorted" values to define their desired active area. For example, to map a `127mm x 142.875mm` physical area on a tablet rotated by 90 degrees, they might have to configure `active_area_size = 203.2 89.3`, which is confusing and error-prone.

#### What does this change do?

This commit fixes the issue by making the active area calculation aware of the `transform` setting.

Before normalizing the `active_area` values, the code now checks if the applied transform is one that swaps the tablet's axes (90°, 270°, and their flipped equivalents). If so, it swaps the tablet's physical width and height to create an `effectivePhysicalSize`. The normalization is then performed against these corrected dimensions.

#### Why is this change necessary?

With this fix, users can specify `active_area_size` and `active_area_position` with the obvious, real-world millimeter values that correspond to the *final, transformed orientation* of their tablet. This makes the configuration more intuitive, predictable, and aligned with user expectations.

I’ve tested this change with all rotation values, including flipped variants, and everything behaves as expected. I believe it is ready to merge.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1788,8 +1788,13 @@ void CInputManager::setTabletConfigs() {
             const auto ACTIVE_AREA_SIZE = g_pConfigManager->getDeviceVec(NAME, "active_area_size", "input:tablet:active_area_size");
             const auto ACTIVE_AREA_POS  = g_pConfigManager->getDeviceVec(NAME, "active_area_position", "input:tablet:active_area_position");
             if (ACTIVE_AREA_SIZE.x != 0 || ACTIVE_AREA_SIZE.y != 0) {
-                t->m_activeArea = CBox{ACTIVE_AREA_POS.x / t->aq()->physicalSize.x, ACTIVE_AREA_POS.y / t->aq()->physicalSize.y,
-                                       (ACTIVE_AREA_POS.x + ACTIVE_AREA_SIZE.x) / t->aq()->physicalSize.x, (ACTIVE_AREA_POS.y + ACTIVE_AREA_SIZE.y) / t->aq()->physicalSize.y};
+                // Rotations with an odd index (90 and 270 degrees, and their flipped variants) swap the X and Y axes.
+                // Use swapped dimensions when the axes are rotated, otherwise keep the original ones.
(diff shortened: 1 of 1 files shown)


src/managers/input/InputManager.cpp
```diff
@@ -1788,8 +1788,21 @@ void CInputManager::setTabletConfigs() {
             const auto ACTIVE_AREA_SIZE = g_pConfigManager->getDeviceVec(NAME, "active_area_size", "input:tablet:active_area_size");
             const auto ACTIVE_AREA_POS  = g_pConfigManager->getDeviceVec(NAME, "active_area_position", "input:tablet:active_area_position");
             if (ACTIVE_AREA_SIZE.x != 0 || ACTIVE_AREA_SIZE.y != 0) {
-                t->m_activeArea = CBox{ACTIVE_AREA_POS.x / t->aq()->physicalSize.x, ACTIVE_AREA_POS.y / t->aq()->physicalSize.y,
-                                       (ACTIVE_AREA_POS.x + ACTIVE_AREA_SIZE.x) / t->aq()->physicalSize.x, (ACTIVE_AREA_POS.y + ACTIVE_AREA_SIZE.y) / t->aq()->physicalSize.y};
+                // Some rotations (90, 270 degrees and their flipped variants) swap the X and Y axes.
+                const bool axesAreSwapped = (ROTATION == 1 || ROTATION == 3 || ROTATION == 5 || ROTATION == 7);
```
vaxerski: `(ROTATION % 2)`?

senotrusov: I mean, possible, just this it more explicit. What if you'll add 30% rotation later? :) Do you want me to change it?

vaxerski: we already use those dw, % just makes more sense

senotrusov: I changed it to the (ROTATION % 2), it does look tidy that way. Thanks for taking the time!

vaxerski:
c-f needed

senotrusov:
Uh-oh, sorry, I didn't have globstar (`**`) enabled in my shell, so `clang-format -i src/**/*{cpp,hpp}` didn't touch my changes.

It should be all right now.

PR #11673, opened 2025-09-11
decorate border fix ala snowman

ItsOhen:
Decorate rules not removing borders.

Snowman being unlazy lazy.

Tested.


vaxerski:
c-f failed

PR #11689, opened 2025-09-13
More than 1 window rule per rule line.

ItsOhen:
Got annoyed at the amount of people putting several window rules on a windowrule line. So lets just let them. Now they can do silly stuff like this if they really want to:

windowrule = bordersize 10, pin, class:kitty, initialTitle:kitty, initialClass:kitty
windowrule = size 200 200, class:kitty, unset bordersize, float

### DIFF
--- a/hyprtester/test.conf
+++ b/hyprtester/test.conf
@@ -332,3 +332,5 @@ gesture = 3, down, mod:ALT, float
 
 gesture = 3, horizontal, mod:ALT, workspace
 
+windowrule = float, pin, class:kitty_C, size 200 200
+windowrule = unset pin, class:kitty_C
--- a/hyprtester/test.conf
+++ b/hyprtester/test.conf
@@ -332,5 +332,6 @@ gesture = 3, down, mod:ALT, float
 
 gesture = 3, horizontal, mod:ALT, workspace
 
-windowrule = float, pin, class:kitty_C, size 200 200
-windowrule = unset pin, class:kitty_C
+windowrule = float, pin, class:wr_kitty
+windowrule = size 100 100, class:wr_kitty
+windowrule = unset pin, class:wr_kitty
--- a/hyprtester/test.conf
+++ b/hyprtester/test.conf
@@ -333,5 +333,5 @@ gesture = 3, down, mod:ALT, float
 gesture = 3, horizontal, mod:ALT, workspace
 
 windowrule = float, pin, class:wr_kitty
-windowrule = size 100 100, class:wr_kitty
+windowrule = size 200 200, class:wr_kitty
 windowrule = unset pin, class:wr_kitty
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -2682,68 +2682,13 @@ std::optional<std::string> CConfigManager::handleWindowRule(const std::string& c
         }
     }
 
-    std::vector<SP<CWindowRule>> rules;
-
-    for (auto token : tokens) {
-        if (token.starts_with("unset")) {
-            std::string ruleName = std::string(token.substr(0, 6));
-            auto        rule     = makeShared<CWindowRule>(ruleName, value, true);
-            std::erase_if(m_windowRules, [&](const auto& other) {
-                if (!other->m_v2)
-                    return other->m_class == rule->m_class && !rule->m_class.empty();
-
-                if (!rule->m_tag.empty() && rule->m_tag != other->m_tag)
-                    return false;
-                if (!rule->m_class.empty() && rule->m_class != other->m_class)
-                    return false;
-                if (!rule->m_title.empty() && rule->m_title != other->m_title)
-                    return false;
-                if (!rule->m_initialClass.empty() && rule->m_initialClass != other->m_initialClass)
-                    return false;
-                if (!rule->m_initialTitle.empty() && rule->m_initialTitle != other->m_initialTitle)
-                    return false;
-                if (rule->m_X11 != -1 && rule->m_X11 != other->m_X11)
-                    return false;
-                if (rule->m_floating != -1 && rule->m_floating != other->m_floating)
-                    return false;
-                if (rule->m_fullscreen != -1 && rule->m_fullscreen != other->m_fullscreen)
-                    return false;
-                if (rule->m_pinned != -1 && rule->m_pinned != other->m_pinned)
-                    return false;
-                if (!rule->m_fullscreenState.empty() && rule->m_fullscreenState != other->m_fullscreenState)
-                    return false;
-                if (!rule->m_workspace.empty() && rule->m_workspace != other->m_workspace)
-                    return false;
-                if (rule->m_focus != -1 && rule->m_focus != other->m_focus)
-                    return false;
-                if (!rule->m_onWorkspace.empty() && rule->m_onWorkspace != other->m_onWorkspace)
-                    return false;
-                if (!rule->m_contentType.empty() && rule->m_contentType != other->m_contentType)
-                    return false;
-                if (rule->m_group != -1 && rule->m_group != other->m_group)
-                    return false;
-                return true;
-            });
-        } else {
-            auto rule = makeShared<CWindowRule>(std::string(token), value, true);
-            if (rule->m_ruleType == CWindowRule::RULE_INVALID) {
-                Debug::log(ERR, "Invalid rule found: {}, Invalid value: {}", value, token);
-                return std::format("Invalid rule found: {}, Invalid value: {}", value, token);
-            }
-            rules.emplace_back(rule);
-        }
-    }
-
-    if (rules.empty() && tokens.empty())
-        return "Invalid rule syntax: no rules provided";
-
     auto get = [&](std::string_view key) -> std::string_view {
         if (auto it = params.find(key); it != params.end())
             return it->second;
         return {};
     };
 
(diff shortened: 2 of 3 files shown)


Dregu:
It's really all about the colon isn't it... Don't quote me on that.

hyprtester/test.conf
```diff
@@ -332,3 +332,5 @@ gesture = 3, down, mod:ALT, float
 
 gesture = 3, horizontal, mod:ALT, workspace
 
+windowrule = float, pin, class:kitty_C, size 200 200
```
vaxerski: I dont mind but this I would mind. It's confusing as shit, I'd rather the order _must_ be `(rules), (selectors)`

ItsOhen: I can see that. But making sure all rules come before params sounds like a pain to check. Like comp each token after the first one containing : and make sure they don't contain : ?
Also should it just nope out and toss a config warning if a rule is after params? Or apply as much as it can and then fail silently like it currently do?
And i bet things like this would still show up:
image

The goal was to make it arbitrary where you put stuff.

vaxerski: considering we split by a comma I don't see how this is a pain? Just go through them one by one

ItsOhen: Did the things. Merge with upstream. cf fails. Good times.
Anywho. Enforcing \<rules\>, \<params\> and made the parsing a bit more sane.

vaxerski: c-f fixed on main, rebase for the CI to pass

src/config/ConfigManager.cpp
```diff
@@ -2657,215 +2656,150 @@ std::optional<std::string> CConfigManager::handleUnbind(const std::string& comma
...
-    else
-        m_windowRules.push_back(rule);
+    if (rules.empty() && tokens.empty())
+        return "Invalid rule syntax: no rules provided";
+
+    for (auto& rule : rules) {
+        if (rule->m_ruleType == CWindowRule::RULE_SIZE || rule->m_ruleType == CWindowRule::RULE_MAXSIZE || rule->m_ruleType == CWindowRule::RULE_MINSIZE)
+            m_windowRules.insert(m_windowRules.begin(), rule);
+        else
+            m_windowRules.push_back(rule);
```
vaxerski: emplace_back pls

vaxerski:
wiki MR pls? :)

unatried:
This is a really cool PR! I have one question though, do the window rules get assigned by the order they are written? If yes, could you just put all the window rules on one line? Like `windowrulev2 = pin, float, size 25% 35%, unset pin, class:term`

ItsOhen:
> This is a really cool PR! I have one question though, do the window rules get assigned by the order they are written? If yes, could you just put all the window rules on one line? Like `windowrulev2 = pin, float, size 25% 35%, unset pin, class:term`

That thought never crossed my mind. That people might want to set and unset a rule in the same line. So no, it doesn't do that. It does all unsets first, then applies the rules in order. But could definitely be done in the future.
Also, you don't have to put v2 after windowrules anymore.

unatried:
>  Also, you don't have to put v2 after windowrules anymore.

That's good to know!

> But could definitely be done in the future.

It would be nice to see that! It doesn't have to be done before the feature gets merged though.

vaxerski:
conflict

vp1981:
This change breaks the valid (?) configuration ([hyprland.conf](https://git.sr.ht/~vp1981/scripts/tree/master/item/config/hyprland/hyprland.conf)). Specifically, with this PR pressing `mod4+enter` now opens a terminal window in the wrong workspace (always 9) with the wrong size.

It is possible that there is a problem with my configuration, but it works pretty well without this PR.

gulafaran:
yeah this changed some rule reading, i had this which i think its typoed

````
windowrule = float, $title:^(vkcube)$
````

the extra $ infront of title made it match all windows and turned hypr into a floating WM with this PR, sort of worked before tho heh

**Comment by ItsOhen** [comment]
> This change breaks the valid (?) configuration ([hyprland.conf](https://git.sr.ht/~vp1981/scripts/tree/master/item/config/hyprland/hyprland.conf)). Specifically, with this PR pressing `mod4+enter` now opens a terminal window in the wrong workspace (always 9) with the wrong size.
> 
> It is possible that there is a problem with my configuration, but it works pretty well without this PR.

My bad. The reason is that you had initialclass instead of initialClass, so no selector was set. No selector = every window got matched.
New PR with a fix is up, just have to get it approved. #11849 
For now you can just change the c to C and it should be just fine.



> yeah this changed some rule reading, i had this which i think its typoed
> 
> ```
> windowrule = float, $title:^(vkcube)$
> ```
> 
> the extra $ infront of title made it match all windows and turned hypr into a floating WM with this PR, sort of worked before tho heh

Same thing here. $title was invalid, so it just everything to float..

**Comment by aphelei** [comment]
Screenshot_20250926-135331
Screenshot_20250926-135244

I have an error with those 2 rules. It didn't happen before

idk if it's fixed with your new commit, i'll try tonight

**Comment by aphelei** [comment]
`windowrule = workspace special:fast silent,	class:^$`
is a bit specific. it's for a discord overlay that doesn't have any class.

it also happens for more simple rule
`
it does the same with
windowrule = workspace special:fast silent,	class:Dofus.x64`

**Comment by ItsOhen** [comment]
Mm, yes, fun, edge-cases. I was naive thinking the rules would be sane and consistent, so everything with with a : would be a selector. Not the case. special: and plugin: both also have it. I'll fix it.

**Comment by vp1981** [comment]
@ItsOhen ,
after changing `initialclass` -> `initialClass` hyprland (commit 6f1d2e771dca1b5eea5ec344ca1b6a80d4fd4ee5) works as before.

**Comment by kRHYME7** [comment]
I think this breaks multi line window rule? 

```
windowrule = \
decorate 0,\
noinitialfocus,\
forcergbx,\
keepaspectratio,\
noblur,\
noanim,\ 
immediate,\
class:^(PenTablet)$
```

(for testing purposes I had this config a while back to test the multiline)


---


Also this?  I don't really know if this was an intended intended option, but this works in the past 

```
windowrule = size [498,275], class:^(localsend),title:^(LocalSend)$

```

--- 

disclaimer did not check the wiki yet but this it a sudden breaking-ish change.

**Comment by Zebra2711** [comment]
> I think this breaks multi line window rule?
> 
> ```
> windowrule = \
> decorate 0,\
> noinitialfocus,\
> forcergbx,\
> keepaspectratio,\
> noblur,\
> noanim,\ 
> immediate,\
> class:^(PenTablet)$
> ```
> 
> (for testing purposes I had this config a while back to test the multiline)
> 
> Also this? I don't really know if this was an intended intended option, but this works in the past
> 
> ```
> windowrule = size [498,275], class:^(localsend),title:^(LocalSend)$
> ```
> 
> disclaimer did not check the wiki yet but this it a sudden breaking-ish change.

same here:

```ini
windowrule = noscreenshare opacity:0.0 decorations:off, class:^(org\.keepassxc\.KeePassXC)$
```
now have to use commit [`7ce451d2`](https://github.com/hyprwm/Hyprland/commit/7ce451d20c33025d4aaaf55444150c4a875b83ea) to avoid breaking changes until it is fully fixed

ItsOhen:
windowrule = noscreenshare opacity:0.0 decorations:off, class:^(org\.keepassxc\.KeePassXC)$

is not even close to a valid rule. So that will ofc break.

Reasons i might have worked before was because faulty rules would fail silently. IE, they wouldn't do anything.

windowrule = size [498,275], class:^(localsend),title:^(LocalSend)$
Also not a valid rule.
windowrule = size 498 275, class:^(localsend),title:^(LocalSend)$
would be a valid one.

Zebra2711:
> windowrule = noscreenshare opacity:0.0 decorations:off, class:^(org.keepassxc.KeePassXC)$
> 
> is not even close to a valid rule. So that will ofc break.

why is it not valid? Without this MR, it worked fine though

ItsOhen:
windowrule = noscreenshare opacity:0.0 decorations:off, class:^(org.keepassxc.KeePassXC)$
should be:
windowrule = noscreenshare, opacity 0.0, decorate off, class:^(org.keepassxc.KeePassXC)$

and the previous version either just ignored the line completely or read noscreenshare and moved on to class:

Zebra2711:
> should be: windowrule = noscreenshare, opacity 0.0, decorate off, class:^(org.keepassxc.KeePassXC)$

can't get it to work the same as the old window rules :v

Dregu:
> old window rules :v

That stuff is from another unmerged PR, if anything it's from the future not old... Obviously this pr wasn't tested against every other open pr...

But still, the matching of rule vs param should probably still be more like "is there a space before a :", not "is there a : anywhere". `if find(' ') < find(':')` or something. Plus the edge cases, plugins and other nightmares of course, but in my head that makes it pretty simple and you don't need to check for a ton of known words...

TwinFox20:
How about adding similar functionality to workspace rules?

unatried:
What's with workspace and layer rules?

yahya9090:
> `windowrule = workspace special:fast silent, class:^$` is a bit specific. it's for a discord overlay that doesn't have any class.
> 
> it also happens for more simple rule ` it does the same with windowrule = workspace special:fast silent, class:Dofus.x64`

mind me asking which overlay you are using and if it still works? i want an overlay myself

PR #11704, opened 2025-09-14
dwindle: add smart_split_on_drop option

lxe:

https://github.com/user-attachments/assets/d1fc3f8f-2da1-47a6-a5a0-32962e0b1a02

Adds a new configuration option `dwindle:smart_split_on_drop` that provides more granular control over when smart split behavior is used in the dwindle layout. 

When enabled, smart split (cursor-position-based splitting) is only used for mouse drop operations, while regular `split_width_multiplier`-based splitting is used for splits during window creation.

This way, you can still split based on aspect ratio when creating new windows, while drag and drop to rearrange using mouse position.

- The change is backward compatible - the new option defaults to `false`, maintaining existing behavior

Pay extra attention to the test

### DIFF
--- a/hyprtester/src/tests/main/smart_split.cpp
+++ b/hyprtester/src/tests/main/smart_split.cpp
@@ -1,0 +1,184 @@
+#include <hyprutils/os/Process.hpp>
+#include <hyprutils/memory/WeakPtr.hpp>
+
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
+#include "../shared.hpp"
+#include "tests.hpp"
+
+static int ret = 0;
+
+static bool spawnKitty(const std::string& class_) {
+    NLog::log("{}Spawning {}", Colors::YELLOW, class_);
+    if (!Tests::spawnKitty(class_)) {
+        NLog::log("{}Error: {} did not spawn", Colors::RED, class_);
+        return false;
+    }
+    return true;
+}
+
+static std::string getWindowAttribute(const std::string& winInfo, const std::string& attr) {
+    auto pos = winInfo.find(attr);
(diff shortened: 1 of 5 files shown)


hyprtester/src/tests/main/smart_split.cpp
```diff
@@ -0,0 +1,184 @@
+#include <hyprutils/os/Process.hpp>
```
vaxerski: filename: `smart_split.cpp` -> `smartSplit.cpp`

hyprtester/src/tests/main/smart_split.cpp
```diff
@@ -0,0 +1,184 @@
...
+    testConfigRegistration();
+    
+    // Test smart split behavior when smart_split_on_drop is enabled
+    testSmartSplitBehavior();
+    
+    // Test smart split on drop behavior
+    testSmartSplitOnDropBehavior();
+    
+    // Test default behavior (smart_split_on_drop disabled)
+    testDefaultBehavior();
```
vaxerski, line 121: these comments are redundant here
```cpp
// this function does a thing
doAThing();
```

vaxerski (approved):

vaxerski:
can you rebase on main for c-f

Kommynct:
would be kinda cool if the tiled window moved out of the way so you know where the window is going to be placed, or a highlight appears in the orientation (sway does this)

lxe:
> would be kinda cool if the tiled window moved out of the way so you know where the window is going to be placed, or a highlight appears in the orientation (sway does this)

Good idea. Let me try in another PR

vaxerski (approved):

vaxerski:
wiki mr needed

Issue #9916, opened 2025-04-06
Implement ext-image-copy-capture-v1


vaxerski:
### Discussion Link
#9896

## Description
We currently only support the wlr extension for screencopy. The new standard one should be supported as well.

### Depends on
#9915

Kommynct:
Should be closed?

Issue #6823, opened 2024-07-09
Inverted colors when sharing a window


dedsm:
### Regression?

Yes

### System Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit b03f41efec14273cf25c42d4cef326acc36cb319  ().
Date: 2024-07-08
Tag: , commits: b03f41efec14273cf25c42d4cef326acc36cb319

flags: (if any)


System Information:
System name: Linux
Node name: manwe
Release: 6.9.8
Version: #1-NixOS SMP PREEMPT_DYNAMIC Fri Jul  5 07:38:21 UTC 2024


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Alder Lake-P GT2 [Iris Xe Graphics] [8086:46a6] (rev 0c) (prog-if 00 [VGA controller])


os-release: ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.05.20240707.1948467"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.05 (Uakari)"
SUPPORT_END="2024-12-31"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.05 (Uakari)"
VERSION_CODENAME=uakari
VERSION_ID="24.05"


plugins:

======Config-Start======


Config File: /home/david/.config/hypr/hyprland.conf: Read Succeeded
exec-once = /nix/store/nn54yahdasinv9hyk1hjwpakw8sqqyq0-dbus-1.14.10/bin/dbus-update-activation-environment --systemd --all && systemctl --user stop hyprland-session.target && systemctl --user start hyprland-session.target
$browser=firefox-developer-edition
$mod=SUPER
$terminal=/nix/store/r907qf294sgb7mhlljvw7z3fgyvl4kha-foot-1.17.2/bin/foot
animations {
  bezier=wind, 0.05, 0.9, 0.1, 1.05
  bezier=winIn, 0.1, 1.1, 0.1, 1.1
  bezier=winOut, 0.3, -0.3, 0, 1
  bezier=liner, 1, 1, 1, 1
  animation=windows, 1, 6, wind, slide
[...]

PR #11709, opened 2025-09-15
protocols: implement image-capture-source-v1 and image-copy-capture-v1

### LINKED ISSUE #9970: Screencopy / Toplevel Export with shm does not transform image data properly

### LINKED ISSUE #10470: Capture regions are incorrect with screencopy under transformed displays
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10433

## Description
When capturing a region, transformed displays get the wrong region.

See the repro case in the discussion.

ikalco:
unify all screen/window/region sharing rendering code
and
implement the image-capture-source-v1 and image-copy-capture-v1 protocols (#9916)

edit:
also fixes #10470 and fixes #9970

do you want me to split into separate PRs?

also how should we do permissions?
either error on image source creation, or do like screencopy and allow the source but render denied screen

See below for all of my testing, everything works on my machine :tm:
So, _**ready for review**_

<details>
<summary><h4>Testing Stuff</h4></summary>

-----------

Legend
 - ICC - image-copy-capture
 - WSC - wlr-creencopy
 - TLE - hyprland toplevel-export
 - N/A - no tool/client to test this, even if we patch HL
 - TBD - haven't tested this yet
 - &cross; - means its borked
 - &check; - means it works as expected

Notes
* WSC Output on Flipped and Flipped 270° is rotated 180° on portal + obs, but is upright for wl-screenrec and grim
  So I think it's a bug with either the portal, pipewire, or obs

TODO
- [x] The damage area in which the cursor is rendered is incorrect in some transforms
- [x] Grim doesn't work with the permissions system because it takes the first frame, which is black
- [x] The no_screen_share window rule blacks out the incorrect region in 180° and Flipped 270° transforms
- [x] <s>The permission system doesn't reload on config change (maybe out of scope?)</s>

--------

Tools to use
#### PORTAL + OBS  (dmabuf) - WSC Output, WSC Region, TLE
* run `killall xdg-desktop-portal-hyprland && /usr/lib/xdg-desktop-portal-hyprland`
* open obs and use the `Screen Capture (PipeWire)` source
* double click the source in the sources list to reopen the selector

#### WL-SCREENREC  (dmabuf) - ICC Output, WSC Output
* build from source, arch package has broken transforms
* test command - `./target/debug/wl-screenrec -o HDMI-A-1 -capture-backend=ext-image-copy-capture`
* it outputs to `screenrecord.mp4`

#### GRIM (shm) - ICC Output, ICC Toplevel
* ICC Output test command - `grim -o HDMI-A-1 -t png /tmp/out.png && swappy -f /tmp/out.png`
* ICC Toplevel test command - `grim -T $(printf %016x $(hyprctl clients -j | jq -r ".[0] | .address")) -t png /tmp/out.png && swappy -f /tmp/out.png`
  Apply this HL patch for this to work, [diff.txt](https://github.com/user-attachments/files/24459955/diff.txt)

#### GRIM w/ patch (shm) - WSC Output, WSC Region
* Build my fork of ZnxTech's fork of grim, that actually uses WSC Output/Region
  * https://github.com/ikalco/grim
* WSC Output test command - `./build/grim -o HDMI-A-1 -t png /tmp/out.png && swappy -f /tmp/out.png`
* WSC Region test command - `./build/grim -g "$(slurp)" -t png /tmp/out.png && swappy -f /tmp/out.png`

--------

### Test that all output transform, buffer type, and protocol combinations give expected behavior
| Output Transform | Buffer Type | Format Type | ICC Toplevel | ICC Output | WSC Output | WSC Region | TLE     |
| ---------------- | ----------- | ----------- | ------------ | ---------- | ---------- |  --------- | ------- |
| Normal           | DMABUF      | RGB 8-bit   | N/A          | &check;    | &check;    | &check;    | &check; |
| 90°              | DMABUF      | RGB 8-bit   | N/A          | &check;    | &check;    | &check;    | &check; |
| 180°             | DMABUF      | RGB 8-bit   | N/A          | &check;    | &check;    | &check;    | &check; |
| 270°             | DMABUF      | RGB 8-bit   | N/A          | &check;    | &check;    | &check;    | &check; |
| Flipped          | DMABUF      | RGB 8-bit   | N/A          | &check;    | &cross;    | &cross;    | &check; |
| Flipped 90°      | DMABUF      | RGB 8-bit   | N/A          | &check;    | &check;    | &check;    | &check; |
| Flipped 180°     | DMABUF      | RGB 8-bit   | N/A          | &check;    | &cross;    | &cross;    | &check; |
| Flipped 270°     | DMABUF      | RGB 8-bit   | N/A          | &check;    | &check;    | &check;    | &check; |
| Normal           | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| 90°              | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| 180°             | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| 270°             | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| Flipped          | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| Flipped 90°      | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| Flipped 180°     | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |
| Flipped 270°     | SHM         | RGB 8-bit   | &check;      | &check;    | &check;    | &check;    | N/A     |

### Test Different Formats
* I used portal + obs and wl-screenrec to test dmabuf rgb 10-bit
* I used grim and grim w/ patch to test shm rgb 10-bit
* I used portal + chromium using https://mozilla.github.io/webrtc-landing/gum_test.html based on #6823 to test both bgr

| Buffer Type | Format Type | ICC Toplevel | ICC Output | WSC Output | WSC Region | TLE     |
| ----------- | ----------- | ------------ | ---------- | ---------- | ---------- | ------- |
| DMABUF      | RGB 10-bit  | N/A          | &check;    | &check;    | &check;    | &check; |
| DMABUF      | BGR 8-bit   | N/A          | N/A        | &check;    | &check;    | &check; |
| DMABUF      | BGR 10-bit  | N/A          | N/A        | &check;    | &check;    | &check; |
| SHM         | RGB 10-bit  | &check;      | &check;    | &check;    | &check;    | N/A     |

### Test Window Share is Blacked out with no_screen_share rule
| Output Transform  | ICC Output | ICC Toplevel | WSC Output | WSC Region | TLE     |
| ----------------- | ---------- | ------------ | ---------- | ---------- | ------- |
| Normal            | &check;    | &check;      | &check;    | &check;    | &check; |
| 90°               | &check;    | &check;      | &check;    | &check;    | &check; |
| 180°              | &check;    | &check;      | &check;    | &check;    | &check; |
| 270°              | &check;    | &check;      | &check;    | &check;    | &check; |
| Flipped           | &check;    | &check;      | &check;    | &check;    | &check; |
| Flipped 90°       | &check;    | &check;      | &check;    | &check;    | &check; |
| Flipped 180°      | &check;    | &check;      | &check;    | &check;    | &check; |
| Flipped 270°      | &check;    | &check;      | &check;    | &check;    | &check; |

### Test Permissions
Correct Behavior: black until allowed/denied, if denied keep rendering black but with a "screencast denied" text. IIRC thats how it is rn.
- &check; | Try recording output through **portal + obs** and cross for correct behavior
- &check; | Try recording output through **wl-screenrec** and cross for correct behavior
- &check; | Try screenshot with **grim** and cross for correct behavior

### Misc Tests
- &check; | moving a window between two monitors
- &check; | resizing a window rapidly (float -> tiled -> float)
- &check; | resize a window until it's majority into another monitor
- &check; | spam between recording an output and a window then close the window and unplug the output
- &check; | record an intensive game
- &check; | record a fullscreen window, to test direct scanout block
- &check; | test that ICC cursor session works using `wayvnc`
  - `hyprctl output create headless test`
  - `hyprctl keyword monitor test,1736x976@60,1920,1`
  - `wayvnc -o test -v -L debug`
  - `gvncviewer 127.0.0.1`
  - To test the cursor copy session, hover a text box or something and make sure the correct cursor is shown in the vnc viewer

</details>

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -23,7 +23,6 @@
 #include "../render/Renderer.hpp"
 #include "../managers/EventManager.hpp"
 #include "../managers/screenshare/ScreenshareManager.hpp"
-#include "../managers/LayoutManager.hpp"
 #include "../managers/animation/AnimationManager.hpp"
 #include "../managers/animation/DesktopAnimationManager.hpp"
 #include "../managers/input/InputManager.hpp"


vaxerski:
I almost shat myself and thought someone random would write a bad impl

then I noticed it's ikalco 🥰

I think we should _unify_ screencopy at this point. Protocol files should have 0 rendering code, and instead that would be delegated to a global sharing mgr.

One MR is ok. Maybe keep commits making sense though.

> also how should we do permissions?

black until allowed/denied, if denied keep rendering black but with a "screencast denied" text. IIRC thats how it is rn.

vaxerski:
ping me for review when ready

arrowpc:
I could help out with this if wanted. I already hacked together a complete implementation for Hyprland 2 months ago or so to test out a library I'm working on. Every compositor either had a poor implementation or an incomplete one so I couldn't test anything reliably.

vaxerski:
JSYK ikalco in #12711 I will be changing a bit how the frames are fed into screencopy so it might impact this a bit but likely not a whole lot. It would still be good to somewhat unify screencopy handling between ext and wlr.

ikalco:
> JSYK ikalco in #12711 I will be changing a bit how the frames are fed into screencopy so it might impact this a bit but likely not a whole lot. It would still be good to somewhat unify screencopy handling between ext and wlr.

alr 👍

side note: 
I'm trying to fix #9970 while I'm here but I'm having some trouble with the renderer
> Window share (toplevel export) using OBS is currently rendered in the correct upright (normal transform) orientation with correct scale for a 90 degree transformed monitor. However, the bottom half of the window is not rendered or is clipped out as it is black. Unsure what the issue is but it seems unlikely to be damage as that looks correct at the scissoring stage of texture rendering. Same with viewport which seems correct. Maybe it's the texture UV's?? Honestly I have no idea but I'm probably missing something obvious in the renderer :/

Any ideas on the cause for the bottom half of a  window on a transformed monitor not being drawn?

vaxerski:
sounds like the size didnt get transformed somewhere maybe?

ikalco:
@vaxerski ready for review
finally :pray:
also if you wanna test there's some guidelines in the PR description under the details

edit:
forgot to push last fix oops, done for real this time

Dregu:
I tested some things, like sunshine-git's screencopy-dmabuf with scaled 10bit output, grim region on scaled transformed output, obs pipewire and screencopy, all fine.

Wanted to ask why is HL blocking DS while screencopying though, cause it seems completely unnecessary to me. Especially when capturing a completely different monitor than what wants to DS, but even recording the same one works just fine if I remove the block?

ikalco:
> I tested some things, like sunshine-git's screencopy-dmabuf with scaled 10bit output, grim region on scaled transformed output, obs pipewire and screencopy, all fine.
> 
> Wanted to ask why is HL blocking DS while screencopying though, cause it seems completely unnecessary to me. Especially when capturing a completely different monitor than what wants to DS, but even recording the same one works just fine if I remove the block?

Direct scanout renders directly to a client's buffer, which the client owns and can do whatever they want with. So we need to block DS and render to our own buffer so that we can copy it to the screenshare client's buffer without any artifacts.

Also yes technically DS block can be per monitor, I might add that in a future PR

edit:
the artifacts bit isn't the only thing btw. The monitor buf will have a black box where the DS window is and if you don't want that then you would be triple rendering the window: the main buf, the ds buf, and the share buf

src/managers/screenshare/ScreenshareManager.hpp
```diff
@@ -0,0 +1,250 @@
...
+    WP<CScreenshareSession> m_self;
+    bool                    m_stopped = false;
+    bool                    m_managed = false;
+
+    eScreenshareType        m_type;
+    PHLMONITORREF           m_monitor;
+    PHLWINDOWREF            m_window;
+    CBox                    m_captureBox; // given capture area in logical coordinates (see xdg_output)
+
+    wl_client*              m_client;
```
vaxerski: buncha uninit primitives, please go around fix those

src/managers/screenshare/ScreenshareManager.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+    void                    onOutputCommit(PHLMONITOR monitor);
+
+  private:
+    std::vector<WP<CScreenshareSession>> m_sessions;
+    std::vector<WP<CCursorshareSession>> m_cursorSessions;
+    std::vector<WP<CScreenshareFrame>>   m_frames;
+
+    struct SManagedSession {
+        SManagedSession(UP<CScreenshareSession> session);
```
vaxerski: &&

src/protocols/ImageCaptureSource.cpp
```diff
@@ -0,0 +1,125 @@
+#include "ImageCaptureSource.hpp"
+#include "core/Output.hpp"
+#include "../helpers/Monitor.hpp"
+#include "../desktop/view/Window.hpp"
+#include "ForeignToplevel.hpp"
+
+CImageCaptureSource::CImageCaptureSource(SP<CExtImageCaptureSourceV1> resource, PHLMONITOR pMonitor) : m_resource(resource), m_monitor(pMonitor) {
```
vaxerski, line 7: if (!good()) return; like all other proto objects

src/managers/screenshare/CursorshareSession.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    if UNLIKELY (buffer->size != m_bufferSize) {
+        LOGM(Log::ERR, "Client requested sharing to an invalid buffer size");
+        return ERROR_BUFFER_SIZE;
+    }
+
+    uint32_t bufFormat;
+    if (buffer->dmabuf().success) {
+        bufFormat = buffer->dmabuf().format;
+    } else if (buffer->shm().success) {
```
vaxerski: ~~{}~~

src/managers/screenshare/ScreenshareManager.cpp
```diff
@@ -0,0 +1,199 @@
...
+        frame->copy();
+        frame->m_session->m_lastFrame.reset();
+        frame->m_session->m_frameCounter++;
+    });
+
+    std::erase_if(m_frames, [&](const WP<CScreenshareFrame>& frame) { return frame.expired(); });
+}
+
+void CScreenshareManager::onTick() {
+    m_tickTimer->updateTimeout(std::chrono::microseconds(500));
```
vaxerski: why are you ticking 200x / s? Make this reactive to state and only tick if no frames are being sent.

If after copying no frames: set timer for (some timeout, e.g. 500ms) and then send event false if we didnt get more

ikalco: c+p fail lol
old screencopy was leeching off of the animation manager timer

fxzzi:
`grim` has the option to use these protos for capturing the window, however it requires the toplevel id you want to capture. Is it completely out of the picture to allow returning of the toplevel id in `hyprctl clients` and more usefully in `hyprctl activewindow`? Currently users would have to use a fork of grim which isn't the best idea. Sway returns the top level id with `swaymsg -t get_tree | jq -j '.. | select(.type?) | select(.focused).foreign_toplevel_identifier'` so would it not make sense for hyprland to provide the same functionality? If you wanted to hide this info by default we could have a flag for hyprctl to expose toplevel id's and other stuff if needed

ikalco:
> `grim` has the option to use these protos for capturing the window, however it requires the toplevel id you want to capture. Is it completely out of the picture to allow returning of the toplevel id in `hyprctl clients` and more usefully in `hyprctl activewindow`? Currently users would have to use a fork of grim which isn't the best idea. Sway returns the top level id with `swaymsg -t get_tree | jq -j '.. | select(.type?) | select(.focused).foreign_toplevel_identifier'` so would it not make sense for hyprland to provide the same functionality? If you wanted to hide this info by default we could have a flag for hyprctl to expose toplevel id's and other stuff if needed

Yeah I used this patch to test. @vaxerski  should this be the default or an option?
```diff
diff --git a/src/protocols/ForeignToplevel.cpp b/src/protocols/ForeignToplevel.cpp
index 5515d2fb..5fb281fb 100644
--- a/src/protocols/ForeignToplevel.cpp
+++ b/src/protocols/ForeignToplevel.cpp
@@ -64,7 +64,7 @@ void CForeignToplevelList::onMap(PHLWINDOW pWindow) {
         return;
     }
 
-    const auto IDENTIFIER = std::format("{:08x}->{:016x}", sc<uint32_t>(rc<uintptr_t>(this) & 0xFFFFFFFF), rc<uintptr_t>(pWindow.get()));
+    const auto IDENTIFIER = std::format("{:016x}", rc<uintptr_t>(pWindow.get()));
 
     LOGM(Log::DEBUG, "Newly mapped window gets an identifier of {}", IDENTIFIER);
     m_resource->sendToplevel(NEWHANDLE->m_resource.get());

```

Dregu:
Yeah the current identifiers don't make any sense to me, but you also can't use just window addresses, cause the ids are supposed to stay unique, and addresses are not. The current ones also seem to change every time I list them with `lswt`, how is that supposed to map to anything? wlroots simply uses 32 digit urandom strings, so that seems good enough. But they should still only be generated once per window, not changed every time someone asks...

And even though complicated protocols and tools like lswt exist, these ids should still be in the hyprctl output like sway has them, easily usable for simple screen capture scripts.

vaxerski:
we dont support xdg_foreign

vaxerski:
small nit, since we're moving a lot of stuff to screencopy, can you use the new tidying?

IOW:
- put all screencopy related things under `namespace Screencopy`
- put manager as `Screencopy::mgr()` which is a method that returns a static UP&: `static UP<...> ptr = makeUnique(); return ptr;`

That way we dont leak enums and shit to the global scope

generally code looks good

Dregu:
> we dont support xdg_foreign

If this was supposed to be some kind of answer to the last few comments about the garbage identifiers given out by **wlr-foreign-toplevel-management** (which we do support) or the fact the identifiers aren't available in hyprctl like they should be, please read the comments again with some thought.

The ICC protocol could be used for toplevel capture just fine IF those identifiers weren't garbage and were available.

vaxerski:
pretty sure you meant ext foreign toplevel list not wlr xd

vaxerski:
~~but yeah those are bogus I can fix them~~ well maybe not bogus but pretty bad

Dregu:
Yeah in my defense [this guy](https://git.sr.ht/~leon_plickat/lswt) said that's what the tool uses, and I stopped thinking there. Just needs a random stable globally unique <=32 character ascii string per window, not one that changes every time someone asks for it...

vaxerski:
omw to add sha256 hashed toplevels

vaxerski:
https://github.com/hyprwm/Hyprland/commit/68456a5d9a54f34b70a8261153dc7d35c17f2bf0

Dregu:
Yeah that works with grim -T now. hyprctl json turned to snake_case though when everything else is camel.

vaxerski:
is there anything left to do here or is this fully gtg

ikalco:
> is there anything left to do here or is this fully gtg

yep gtg

edit:
ok for real this time
p.s. reminder to never to resolve merge conflicts using github web ui lol

vaxerski:
oh yeah thats correct always rebase on your own in git

vaxerski (approved):
lgtv thanks!!

fxzzi:
i just thought i'd comment to say thanks for the PR everything working great i can now screenshot things more well

is it intentional that it captures windows post surface transformations? i.e. a native wayland sdl client like CS2, when running in stretched resolution? Afaik the stretch is done by Hyprland in this case, as obs-vkcapture captures without stretch as seen below, and Hyprland does claim that DS is disabled by saying `directScanoutBlockedBy: surface transformations`

image

captured with `grim -T`
image


if it's intentional or my assumptions are bogus, thats fine, just checking :)

ikalco:
hmm no the stretching is not intentional but the upright is, can you give repro steps?
also I'm guessing vkcapture hooks into the game's vulkan calls instead of using wayland proto right?

fxzzi:
Yea, you can repro with any native SDL3 client with SDL_VIDEO_DRIVER=Wayland, I'd do CS2. Then just set the in game resolution in settings to any of a different aspect ratio (I used 1920x1440 / 4:3)

If you have an overlay like mangohud enabled you will notice it stretch as well.

If it's in XWayland, the overlays will not be stretched, indicating that the stretching is being done by CS2 / SDL instead of Hyprland. In this case it makes sense for grim -T to capture it stretched, and as expected DS does trigger in this scenario

P.s. if struggling to get XWayland to stretch CS2, make sure you are not using MSAA in settings

vaxerski:
this is intentional dw

0WD0:
Not work in scrolling layout?

https://github.com/hyprwm/Hyprland/pull/12890

src/managers/screenshare/ScreenshareManager.cpp
```diff
@@ -0,0 +1,161 @@
...
+            return;
+
+        if (frame->m_session->monitor() != monitor)
+            return;
+
+        if (frame->m_session->m_type == SHARE_WINDOW) {
+            CBox geometry = {frame->m_session->m_window->m_realPosition->value(), frame->m_session->m_window->m_realSize->value()};
+            if (geometry.intersection({monitor->m_position, monitor->m_size}).empty())
+                return;
+        }
```
0WD0: I think it's not compatible with scrolling layout.

PR #11720, opened 2025-09-15
Virtual keyboard: update cached mods before IME; add share_states = 2 config option for auto disable when IME client

xzn:

Fixes issue discussed here: #11715

`share_states` config option is changed to an int. 0 is no, 1 is yes, 2 is auto (disable when IME client)

Need tesing..

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -683,9 +683,9 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
 
     SConfigOptionDescription{
         .value       = "input:virtualkeyboard:share_states",
-        .description = "Unify key down states and modifier states with other keyboards",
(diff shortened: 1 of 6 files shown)


xzn:
Only this change: https://github.com/hyprwm/Hyprland/pull/11720/files#diff-2ba0a4b3fea7c686b5bcf66318fa5912cb5d78a313318bf7ce2d51faf3b7a88aR1482-L1491

is consequential for the issue.

My code previously update the mods states at the wrong time (after ime and sending to seat), where there's a race condition that the mod is sent to the ime but not seat yet, and another key pressed is received, and key binds handling will be using stale mod states.

Now it updates mods states immediately after receiving from keyboard, before sending to ime/seat.

Rest is hopefully to make things more compatible by default for non-ime virtual keyboards.

Apologies for making this issue in the first place.

vaxerski:
wiki mr needed

PR #11721, opened 2025-09-15
Fix multi-argument gesture dispatcher parsing

kolayne:

The `dispatcher` gesture handler used to only handle
the first argument to the dispatcher, while some dispatchers
(e.g., `sendshortcut`) want multiple arguments.

This fixes `ConfigManager` to handle all the arguments
provided to the dispatcher gesture handler.

Fixes https://github.com/hyprwm/Hyprland/discussions/11684.

An alternative implementation could be to join the already parsed arguments with the `,` separator. But using the original string feels more portable, e.g., in case future dispatchers may depend on whitespace, etc.

Hopefully, ready for merge!


kolayne:
Huh. The code style pipeline fails for a line that I never edited.

hyprtester/src/tests/main/gestures.cpp
```diff
@@ -27,19 +41,21 @@ static bool test() {
     NLog::log("{}Switching to workspace 1", Colors::YELLOW);
     getFromSocket("/dispatch workspace 1"); // no OK: we might be on 1 already
 
-    OK(getFromSocket("/dispatch plugin:test:gesture left,3"));
```
vaxerski, line 30: do not delete tests. Add your own after the previous ones.

src/config/ConfigManager.cpp
```diff
@@ -3230,10 +3230,14 @@ std::optional<std::string> CConfigManager::handleGesture(const std::string& comm
 
     std::expected<void, std::string> result;
 
-    if (data[startDataIdx] == "dispatcher")
-        result = g_pTrackpadGestures->addGesture(makeUnique<CDispatcherTrackpadGesture>(std::string{data[startDataIdx + 1]}, std::string{data[startDataIdx + 2]}), fingerCount,
-                                                 direction, modMask, deltaScale);
-    else if (data[startDataIdx] == "workspace")
+    if (data[startDataIdx] == "dispatcher") {
+        auto dispatcher_args_it = value.begin();
```
vaxerski: snake case is banned

src/config/ConfigManager.cpp
```diff
@@ -3230,10 +3230,14 @@ std::optional<std::string> CConfigManager::handleGesture(const std::string& comm
...
+        for (int i = 0; i < startDataIdx + 2; ++i) {
+            dispatcher_args_it = find(dispatcher_args_it, value.end(), ',') + 1;
```
vaxerski: wth is `find`?

hyprtester/src/tests/main/gestures.cpp
kolayne: It is a diff rendering problem. I did not delete it, I just moved it below (i.e., I added my test at the top). If I add the test below, it passes for me locally but fails on the pipeline, so the simple solution was to do this as the first thing on the test.

Do you want me to debug it and move the test to the bottom?

src/config/ConfigManager.cpp
kolayne: Oops, it's `std::find`. I'm not sure why it compiled. Fixed

src/config/ConfigManager.cpp
kolayne: Fixed

src/config/ConfigManager.cpp
```diff
@@ -3230,10 +3230,14 @@ std::optional<std::string> CConfigManager::handleGesture(const std::string& comm
...
     std::expected<void, std::string> result;
 
-    if (data[startDataIdx] == "dispatcher")
-        result = g_pTrackpadGestures->addGesture(makeUnique<CDispatcherTrackpadGesture>(std::string{data[startDataIdx + 1]}, std::string{data[startDataIdx + 2]}), fingerCount,
-                                                 direction, modMask, deltaScale);
-    else if (data[startDataIdx] == "workspace")
+    if (data[startDataIdx] == "dispatcher") {
+        auto dispatcherArgsIt = value.begin();
+        for (int i = 0; i < startDataIdx + 2; ++i) {
+            dispatcherArgsIt = std::find(dispatcherArgsIt, value.end(), ',') + 1;
```
vaxerski, line 3236: can this potentially overflow when std::find returns value.end()?

hyprtester/src/tests/main/gestures.cpp
vaxerski: you can move the test to a fn and add your own fn, then in test() do `one(); two();`

src/config/ConfigManager.cpp
kolayne: Yes, it might if the config is invalid. Fixed.

vaxerski (approved):

kolayne:
@vaxerski, ahh, unfortunately, this MR breaks something. Consider the line:
```
gesture = 4, left, dispatcher, movefocus, l
```

According to the code, we find the last comma and take the first character after it, which is a space. The `moveFocusTo` function ends up receiving the string `" l"`, and the first character of it is not recognized as a direction, so the binding does not work.

I want to submit a fix, but I am not sure which function's fault is it: are dispatchers supposed to handle whitespaces, so we need to update `moveFocusTo`? Or is it config manager's responsibility to strip whitespaces?

kolayne:
I can see that `CVarList` and `CConstVarList` have join methods, so we can keep the parsing logic there and just join the rest of the arguments parsed with the `gesture` keyword. I will do that.

However, should dispatcher functions accept `std::string` at all? Maybe it makes more sense to instead accept a range of parsed strings, in the form of a `CVarList`/`CConstVarList` or a pair of iterators over `std::string_view`, etc?

PR #11735, opened 2025-09-17
exec: Spawn processes as direct children

kolayne:

Spawn processes as children rather than grandchildren.
This way, spawned processes may track Hyprland's state
by watching their parent, either directly or indirectly
(e.g., Linux's `PR_SET_PDEATH_SIG`).

No

Hopefully, ready for merging!

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -949,17 +949,10 @@ uint64_t CKeybindManager::spawnRawProc(std::string args, PHLWORKSPACE pInitialWo
 
     const auto HLENV = getHyprlandLaunchEnv(pInitialWorkspace);
 
-    int        socket[2];
-    if (pipe(socket) != 0) {
-        Debug::log(LOG, "Unable to create pipe for fork");
-    }
-
-    CFileDescriptor pipeSock[2] = {CFileDescriptor{socket[0]}, CFileDescriptor{socket[1]}};
-
-    pid_t           child, grandchild;
+    pid_t      child;
     child = fork();
     if (child < 0) {
-        Debug::log(LOG, "Fail to create the first fork");
+        Debug::log(LOG, "Fail to fork");
         return 0;
     }
     if (child == 0) {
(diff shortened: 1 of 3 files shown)


kolayne:
We should probably apply a similar fix to hyprutils, which can also spawn processes?

vaxerski:
> Spawn processes as children rather than grandchildren.
This way, spawned processes may track Hyprland's state
by watching their parent, either directly or indirectly
(e.g., Linux's PR_SET_PDEATH_SIG).

Why? I am not aware of any environment that does this.

kolayne:
> Why?

One use case, as described in the discussion, is that I want some tools auto-started with `Hyprland` and terminated when it exits. It is much easier to implement if they are children of `Hyprland` in the process tree.

Besides, generally speaking, from the process model perspective it makes sense for tasks spawned by `Hyprland` to remain as its children.

> I am not aware of any environment that does this.

Depending on what you mean by "environment":
- All shells that I used (e.g., bash, fish) do that;
- The i3 window manager does that (i3/i3#5506, i3/i3#5510)

vaxerski:
sway does not, KWin does not AFAIK, Gnome probably doesnt either.

kolayne:
Okay, but why do we care whether they do it?

I see an advantage in spawning children directly: the process model remains in effect (children's parent process is indeed their logical parent). Spawning processes as grandchildren, on the other hand, does not seem to have advantages: the logic is more complex, plus, it requires an extra fork. We don't need to handle zombie processes either way.

So, why perform the extra grandchildren logic?

vaxerski:
> So, why perform the extra grandchildren logic?

doesn't spam up the stdout, for example.

kolayne:
> doesn't spam out the stdout, for example.

That does not have to do with parenting, it has to do with file descriptors. As you can see in the diff, the child still overrides stdout and stderr with the descriptor for `/dev/null`

vaxerski:
hm, right. Fine.

src/managers/KeybindManager.cpp
```diff
@@ -949,17 +949,10 @@ uint64_t CKeybindManager::spawnRawProc(std::string args, PHLWORKSPACE pInitialWo
...
-    int        socket[2];
-    if (pipe(socket) != 0) {
-        Debug::log(LOG, "Unable to create pipe for fork");
-    }
-
-    CFileDescriptor pipeSock[2] = {CFileDescriptor{socket[0]}, CFileDescriptor{socket[1]}};
-
-    pid_t           child, grandchild;
+    pid_t      child;
     child = fork();
```
vaxerski: you can concatenate both lines

kolayne: Fixed

kolayne:
@vaxerski, should we also update hyprutils accordingly? Does it make sense to reuse `Hyprutils::OS::CProcess` here rather than duplicating the functionality?

vaxerski:
I would rather this was optional in CProcess, but you can make a MR, just leave the defaults as-is

kolayne:
Forgot to say: thank you for being open minded and making it easy to introduce this feature into Hyprland!

PR #11749, opened 2025-09-18
renderer: asynchronously load background tex

vaxerski:
Loads the background texture asynchronously, dropping it after use, meaning 150MB RAM less, at the cost of the background being blank for up to a few seconds.

For real systems, modesets take long enough that this should not be noticeable, probably.

@fufexan bump nix hyprgraphics pls


fxzzi:
any reason for not having an option of just not loading anything on the bg? for those who already have hyprpaper / alternatives already setup.

vaxerski:
> any reason for not having an option of just not loading anything on the bg? for those who already have hyprpaper / alternatives already setup.

uhhh disable hyprland logo has been there for 3 years

fxzzi:
> uhhh disable hyprland logo has been there for 3 years

yeah i know, so you're saying this doesn't change anything for those with hyprland logo disabled?

vaxerski:
yeah

PR #11752, opened 2025-09-18
Fix inconsistent border thickness for roundingPower < 2

REVO9:
Setting decoration:roundingPower to a value smaller than 2 causes inconsistent thickness along corners. Example with roundingPower = 1:
image

this PR fixes this by applying a correction offset to the rounding of borders and shadows:
image

- The offset is applied linearly based on the rounding power. This is an approximation but it works fine
- The Offset for the shadow is double the offset for the border, I have no Idea why, but it fits perfectly

Hopefully ready

### DIFF
--- a/src/render/shaders/glsl/border.frag
+++ b/src/render/shaders/glsl/border.frag
@@ -1,6 +1,8 @@
 #version 300 es
 #extension GL_ARB_shading_language_include : enable
 
+# define M_SQRT1_2	0.70710678118654752440
+
 precision highp float;
 in vec2 v_texcoord;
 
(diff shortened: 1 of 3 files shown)


src/render/shaders/glsl/border.frag
```diff
@@ -134,6 +136,10 @@ void main() {
     if (min(pixCoord.x, pixCoord.y) > 0.0 && radius > 0.0) {
 	    float dist = pow(pow(pixCoord.x,roundingPower)+pow(pixCoord.y,roundingPower),1.0/roundingPower);
 	    float distOuter = pow(pow(pixCoordOuter.x,roundingPower)+pow(pixCoordOuter.y,roundingPower),1.0/roundingPower);
+
+        float correctionOffset = thick * (M_SQRT1_2 - 0.5) * max(2.0 - roundingPower, 0.0);
```
vaxerski: this can be a uniform I think you meant to do that

REVO9:
Maybe like this?  We don't have to make a new uniform here

Issue #11500, opened 2025-08-22
Addreserved should also take descriptions


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11481

## Description
it only takes port names atm.

PR #11755, opened 2025-09-18
Get monitor by string to allow any identifier to be used.

ItsOhen:
Fixes: #11500

No

Yes


ItsOhen:
> I dont like this. Why not do the getMonitorFromString where we actually use .name?

Later on rules are purged based on name. Just did it this way to make sure it doesn't missmatch anywhere.

vaxerski:
what about rules when a monitor is plugged in later?

ItsOhen:
Fair enough. I guess move it to the renderer instead? But feels kinda clunky.
Would a local copy of addReservedArea be ok in CMonitor and ditching the map?

vaxerski:
yea

Issue #10659, opened 2025-06-05
Allow instant return from a submap after executing a bind


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10656

## Description
Allow marking a submap for exit after any bind hits.

bodograumann:
Thanks for implementing this feature. Makes my config much more readable. <3

PR #11760, opened 2025-09-19
Better submaps

ItsOhen:

Addresses: #10659
But not quite a good solution yet.
What i really wanted was a syntax like this:
```
submap = someSub {
  oneshot = true
  bind ,x,exec ...
  submap = Layer2 {
    oneshot = false
    bind ,x,exec ...
   }
}
```
Then a stack of submaps to keep track of where we are.
Hyprlang doesn't really allow for this though. So this is what i ended up with.
Using the syntax:
```
bind = $mainMod, a, submap, someSub
submap = someSub oneshot
bind = ,q,submap, reset
bind = ,e,exec,...
bind = ,w,submap,Layer2

submap = Layer2 oneshot
bind = ,q,submap, reset
bind = ,e,exec,...
bind = ,w,submap, Layer3

submap = Layer3 oneshot
bind = ,e,exec,...
bind = ,q,submap, reset
submap = reset
```

With it being just strings floating around it's really hard to cascade parents.
So for now if a submap is oneshot, it just resets submap to default.
I made a version using containers instead which made it trivial to go to nearest none oneshot, but felt heavy and unnecessary without a different way to nest submaps. Like, how many layers are people really going to put in?

Probably not. Some suggestions on how to make this look better would be nice.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -3011,7 +3011,7 @@ std::optional<std::string> CConfigManager::handleWorkspaceRules(const std::strin
 
 std::optional<std::string> CConfigManager::handleSubmap(const std::string&, const std::string& submap) {
     const auto SUBMAP     = CConstVarList(submap);
-    m_currentSubmap.name  = SUBMAP[0];
+    m_currentSubmap.name  = (SUBMAP[0] == "reset") ? "" : SUBMAP[0];
     m_currentSubmap.reset = SUBMAP[1];
     return {};
 }
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -205,9 +205,9 @@ static SDispatchResult vkb(std::string in) {
 }
 
 static SDispatchResult scroll(std::string in) {
-    int by;
+    double by;
     try {
-        by = std::stoi(in);
+        by = std::stod(in);
     } catch (...) { return SDispatchResult{.success = false, .error = "invalid input"}; }
 
     Debug::log(LOG, "tester: scrolling by {}", by);
--- a/hyprtester/src/tests/main/keybinds.cpp
+++ b/hyprtester/src/tests/main/keybinds.cpp
@@ -1,4 +1,5 @@
 #include <filesystem>
+#include <linux/input-event-codes.h>
 #include <thread>
 #include "../../shared.hpp"
 #include "../../hyprctlCompat.hpp"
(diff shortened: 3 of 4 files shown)


vaxerski:
I don't see the point. Why not just use the current submaps to move to and from? It doesn't address the issue to me?

ItsOhen:
? Wasn't the issue that there should be a way to go back to default submap when a keybind is triggered?
Which is what this does.
If submap is oneshot, go back to default after a dispatcher is called.
If not, stay in w/e submap you are in until a reset or some other submap is called.

And the reason i moved it to a struct was so that hopefully in the future there is a better way to parse { }, and then it would be easy to make a neat push pop stack structure for submaps.
Maybe something like:
`while(p->parent && p->oneshot) { m_submap.pop(); p = p->parent; }`

vaxerski:
oh yeah sorry got confused there with your description

src/config/ConfigManager.cpp
```diff
@@ -3055,12 +3055,18 @@ std::optional<std::string> CConfigManager::handleWorkspaceRules(const std::strin
...
 }
 
-std::optional<std::string> CConfigManager::handleSubmap(const std::string& command, const std::string& submap) {
-    if (submap == "reset")
-        m_currentSubmap = "";
-    else
-        m_currentSubmap = submap;
-
+std::optional<std::string> CConfigManager::handleSubmap(const std::string&, const std::string& submap) {
+    if (auto pos = submap.find("oneshot"); pos != std::string::npos) {
```
vaxerski: CConstVarList would be good here imo

src/managers/KeybindManager.hpp
```diff
@@ -17,6 +17,15 @@ class IKeyboard;
 
 enum eMouseBindMode : int8_t;
 
+struct SSubmap {
+    std::string name    = "";
+    bool        oneshot = false;
+    std::string parent  = "";
```
vaxerski: parent is unused as far as I can see

src/config/ConfigManager.cpp
ItsOhen: Yeah.. I must have been pretty tired when i first wrote it.
New commit is a smoother way to do it by far.
Let's people define what submap they want to go to instead.
So they can do:
`submap = someSub, reset` to go back to default or `submap = someSub, someOtherSub` if they want to go to another submap.

ItsOhen:
Fucking cf.. I don't even know anymore. I auto run it on save. Complains. Manually run it before squash, still complains. Stage, run, commit, still complains. Like come on. Explain this to me:
image

src/managers/KeybindManager.cpp
```diff
@@ -794,6 +794,9 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
                 found = true; // don't process keybinds on submap change.
                 break;
             }
+            if (k->handler != "submap" && !k->submap.reset.empty()) {
```
vaxerski: no {}

PR #11775, opened 2025-09-20
F18 Dom looking for a cute femboy to get freaky with

upblowing:
Yeah, thats all my discord is @puke.sh send ur submissions. Any age!


Kommynct:
vaxry you are denying one of the most commonly requested features in hyprland without explanation, I'd like a word with your manager.

PR #11786, opened 2025-09-21
Added reservedarea to monitor rules.

ItsOhen:
Lets each monitor rule have their own additionalreservedarea so they don't have to be stored in configmanager.
Which also lets both V1 and V2 rules use addreserved and match by desc.
Fixes: #11500 

V2 rules always replace rules. Which is fine i guess.

Mhm

### DIFF
--- a/src/config/ConfigManager.hpp
+++ b/src/config/ConfigManager.hpp
@@ -5,7 +5,6 @@
 
 #include <map>
 #include <unordered_map>
-#include "../defines.hpp"
 #include <variant>
 #include <vector>
 #include <optional>
@@ -47,13 +46,6 @@ struct SWorkspaceRule {
     std::optional<std::string>         onCreatedEmptyRunCmd;
     std::optional<std::string>         defaultName;
     std::map<std::string, std::string> layoutopts;
-};
-
-struct SMonitorAdditionalReservedArea {
-    int top    = 0;
-    int bottom = 0;
-    int left   = 0;
-    int right  = 0;
 };
 
 struct SPluginKeyword {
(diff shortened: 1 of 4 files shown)


src/config/ConfigManager.hpp
```diff
@@ -202,39 +194,39 @@ class CConfigManager {
...
-    void                                                            ensurePersistentWorkspacesPresent();
+    std::vector<SP<CWindowRule>>                 getMatchingRules(PHLWINDOW, bool dynamic = true, bool shadowExec = false);
+    std::vector<SP<CLayerRule>>                  getMatchingRules(PHLLS);
+    void                                         ensurePersistentWorkspacesPresent();
 
-    const std::vector<SConfigOptionDescription>&                    getAllDescriptions();
+    const std::vector<SConfigOptionDescription>& getAllDescriptions();
 
-    std::unordered_map<std::string, SMonitorAdditionalReservedArea> m_mAdditionalReservedAreas;
+    SMonitorAdditionalReservedArea               m_mAdditionalReservedArea = {};
```
vaxerski, line 229: nuke this since you moving this to rules

ItsOhen: Kinda need it to keep a global addreservedarea state though. :/
Equivalent to what find("") did before.

vaxerski: I dont think it was a good idea, we can remove that

ItsOhen: Okey, so i did. But a bunch of things cropped up during testing.
For example, with the way it's done now it can only match if you use the same identifier. Say `monitor = desc:Samsung,addreserved` and `monitor = desc:Samsung,preferred, auto, 1`
Any additions to another rule that match that monitor. Say `monitor = DP-1,preferred, auto, 1` instead and this will tossed. Adding in a matching v2 rule will also toss it, no matter where in the config it is.
So the whole thing just feels very fragile and prob not worth it.

vaxerski: why arent the two merged? As in:
```
monitor = DP-1, preferred, auto, 1, addreserved, a b c d
```

We could do that. In v2:
```
addreserved = a b c d
```

PR #11857, opened 2025-09-27
small perf changes found when profiling

gulafaran:
was profiling a bit when just having a single terminal damaging surface ocasionally, saw a few contenders where marginal gains could be got. was mainly a lot of virtual calls going around when not needed, and the unordered_map hashing and internal churn in opengl setcap. so reduce it a bit by sending around the weak pointer all the way and storing pos/size once instead of all the PWINDOW->* and use a enum and array over unordered_map.

### DIFF
--- a/src/render/decorations/CHyprDropShadowDecoration.cpp
+++ b/src/render/decorations/CHyprDropShadowDecoration.cpp
@@ -42,15 +42,19 @@ void CHyprDropShadowDecoration::damageEntire() {
         return; // disabled
 
     const auto PWINDOW = m_window.lock();
+    const auto pos     = PWINDOW->m_realPosition->value();
+    const auto size    = PWINDOW->m_realSize->value();
 
-    CBox       shadowBox = {PWINDOW->m_realPosition->value().x - m_extents.topLeft.x, PWINDOW->m_realPosition->value().y - m_extents.topLeft.y,
-                            PWINDOW->m_realSize->value().x + m_extents.topLeft.x + m_extents.bottomRight.x,
-                            PWINDOW->m_realSize->value().y + m_extents.topLeft.y + m_extents.bottomRight.y};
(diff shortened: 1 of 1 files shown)


src/render/OpenGL.cpp
```diff
@@ -3171,17 +3161,18 @@ void CHyprOpenGLImpl::setViewport(GLint x, GLint y, GLsizei width, GLsizei heigh
...
     // check if the capability status is already set to the desired status
-    auto it            = m_capStatus.find(cap);
-    bool currentStatus = (it != m_capStatus.end()) ? it->second : false; // default to 'false' if not found
+    bool currentStatus = m_capStatus.test(cap);
 
     if (currentStatus == status)
         return;
 
-    m_capStatus[cap] = status;
+    m_capStatus.set(cap, status);
```
vaxerski: does this throw if we exceed max?

gulafaran: "If pos is equal or greater than the bitset size, an out_of_range exception is thrown." was checking most defines of gl blend etc that made sense and only hit 3089 , so unless i missed some obvious one i dont think it will be hit. The few that goes big are tesselation and msaa ones but i dont think those are something thats gonna be used in a compositor

gulafaran: ok moved to a enum and array, and now we just dont cache any states that is used outside of hyprland like plugins. and just apply it as normally. and can always be extended if needed.

gulafaran: or let me rephrase that, we dont cache any caps that isnt used in hyprland. any plugins using setcap still caches gl_blend, stencil test and scissor test if any other caps comes in it just applies.

vaxerski (approved):

vaxerski:
@fufexan fix ci pls

fufexan:
Needs rebase on main. I've changed CI to build hyprgraphics after hyprutils.

gulafaran:
rebased.

PR #11862, opened 2025-09-28
Fix clang-tidy "errors"

UjinT34:
Excluded existing "issues" from `WarningsAsErrors` so IDE doesn't report fake errors:
```
  -bugprone-multi-level-implicit-pointer-conversion,
  -bugprone-empty-catch,
  -bugprone-unused-return-value,
  -bugprone-reserved-identifier,
  -bugprone-switch-missing-default-case,
  -bugprone-unused-local-non-trivial-variable,
  -cppcoreguidelines-pro-type-const-cast,
  -google-global-names-in-headers,
  -misc-use-internal-linkage,
  -modernize-use-emplace,
  -modernize-redundant-void-arg,
  -modernize-use-starts-ends-with,
  -modernize-use-designated-initializers,
  -modernize-use-std-numbers,
  -performance-inefficient-vector-operation,
  -performance-inefficient-string-concatenation,
  -performance-enum-size,
  -performance-move-const-arg,
  -readability-identifier-naming,
  -readability-use-std-min-max,
  -readability-math-missing-parentheses,
  -readability-simplify-boolean-expr,
  -readability-static-accessed-through-instance,
  -readability-use-anyofallof,
  -readability-enum-initial-value,
  -readability-redundant-inline-specifier,
```

Malloc "errors" are marked with `// NOLINT(cppcoreguidelines-no-malloc)`. Those are mostly `free` for some external data.
Suggested auto fixes are applied in a few places.

If any of those diagnostics should be considered error then it's better to fix existing code first so that these fake errors don't mask the real ones.

Ready


src/protocols/LayerShell.cpp
```diff
@@ -159,7 +159,7 @@ CLayerShellResource::CLayerShellResource(SP<CZwlrLayerSurfaceV1> resource_, SP<C
             return;
         }
 
-        if (anchor && (!m_pending.anchor || !(m_pending.anchor & anchor))) {
+        if (!m_pending.anchor || !(m_pending.anchor & anchor)) {
```
UjinT34, line 162: https://discord.com/channels/961691461554950145/1270460936301314048/1422223299995500556

`zwlrLayerSurfaceV1Anchor` doesn't have a falsy value. if it can be 0 then this value should be added to the enum or the input type should be changed to uint32_t and cast to enum after the check. otherwise bugprone-non-zero-enum-to-bool-conversion will keep reporting this check. https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_exclusive_edge doesn't mention a zero value so zwlr_layer_surface_v1#39.set_exclusive_edge(0) shouldn't be valid.

aurelleb: https://github.com/KDE/layer-shell-qt does call `zwlr_layer_surface_v1_set_exclusive_edge(0)` by default even if no exclusive edge is explicitly specified. this change broke applications using layer-shell-qt. See #12241

PR #11873, opened 2025-09-29
Fix all the rules again

ItsOhen:
Any more special cases that should be considered?
This whole rule thing is a mess.
Time to move to something more unified?

Also, changed to grab monitor by string to capture stuff like `monitor desc:Somemonitor`, or if people want to be crazy and always have the next workspace show up on the next left monitor with `monitor l`. What do i know?

No

Sure.


vaxerski:
> Any more special cases that should be considered?

If I knew, this MR would not exist. Rules have become very powerful and I am not using all of their power, so we debug in git.

> Time to move to something more unified?

I don't see any better system xD

The good thing is we're getting more tests, so they will catch issues in the future.

vaxerski (approved):

ItsOhen:
I'm currently playing around with a templated IRule thingy using traits and reflection.
At least for workspace, monitor and layer rules.
Unifying add, match and merge.
Should bring some order to the chaos that is CConfigManager.

vaxerski:
it is a bit of a chaos yes.

Reflection? In C++? They added that already?

Anyways, improvements there are welcome

ItsOhen:
Well, no, not really. But it's fakeable.
Kinda like you did hyprlang.
It's in c++ experimental, but prob some time until we get true support for it.
https://en.cppreference.com/w/cpp/experimental/reflect.html

PR #11890, opened 2025-10-01
protocols/seat: force down rounding of coords at the surface edge

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/11665

Needs a test from those affected cc @end-4


Dregu:
Waybar clicks on bottom edge seem to work now.

vaxerski:
coolio, thanks!

PR #11896, opened 2025-10-01
make wl_surface::frame follow pending states

gulafaran:
move the frame callback class to ssurfacestate, add them to pending states and send them in order as they came in to avoid wrong frame callbacks when we are waiting for buffer readyness.

also fix damage tracking, before every single non buffer commit was redrawing on each commit causing serious slowdown. ensure we only do it once. by clearing damage in updateFor

fixes big part of: https://github.com/hyprwm/Hyprland/discussions/11881

however requires extensive VRR, DS testing or other random apps.

### DIFF
--- a/src/protocols/types/SurfaceState.hpp
+++ b/src/protocols/types/SurfaceState.hpp
@@ -3,29 +3,10 @@
 #include "../../helpers/math/Math.hpp"
 #include "../WaylandProtocol.hpp"
 #include "./Buffer.hpp"
-#include "helpers/time/Time.hpp"
 
 class CTexture;
 class CDRMSyncPointState;
-
-struct SSurfaceStateFrameCB {
-  public:
-    SSurfaceStateFrameCB(SP<CWlCallback>&& resource_);
-    ~SSurfaceStateFrameCB() noexcept = default;
-    // disable copy
-    SSurfaceStateFrameCB(const SSurfaceStateFrameCB&)            = delete;
-    SSurfaceStateFrameCB& operator=(const SSurfaceStateFrameCB&) = delete;
-
-    // allow move
-    SSurfaceStateFrameCB(SSurfaceStateFrameCB&&) noexcept            = default;
-    SSurfaceStateFrameCB& operator=(SSurfaceStateFrameCB&&) noexcept = default;
-
-    bool                  good();
-    void                  send(const Time::steady_tp& now);
-
-  private:
-    SP<CWlCallback> m_resource;
-};
+class CWLCallbackResource;
 
 struct SSurfaceState {
     union {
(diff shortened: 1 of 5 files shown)


src/protocols/types/SurfaceState.hpp
```diff
@@ -3,10 +3,30 @@
 #include "../../helpers/math/Math.hpp"
 #include "../WaylandProtocol.hpp"
 #include "./Buffer.hpp"
+#include "helpers/time/Time.hpp"
 
 class CTexture;
 class CDRMSyncPointState;
 
+struct SSurfaceStateFrameCB {
```
vaxerski: do not move the callback resource class like this. Keep it as it was, you can add methods to it. Resources need to be uniform, this is inviting issues down the line.

There's nothing you added here that can't just be added to CWLCallbackResource

you can keep WP<> refs in the surface state

src/protocols/types/SurfaceState.hpp
```diff
@@ -39,7 +60,9 @@ struct SSurfaceState {
     Vector2D offset;
 
     // for xdg_shell resizing
-    Vector2D ackedSize;
+    Vector2D                              ackedSize;
+
+    std::vector<SP<SSurfaceStateFrameCB>> callbacks;
```
vaxerski: callbacks belong to core, probably a good idea to just have a `std::vector<SP<CWLCallbackResource>> m_callbacks` in CWLCompositor

gulafaran: it made it easier to track wich were pending, and not from the future when it is in SSurfaceState

surface sent bunch of states, including frame, attach buffer, commit. that buffer goes waiting. and while waiting a new frame callback came in. thats not related and when presenting it called them all.

gulafaran: and if im reading the spec right

```
The frame request will take effect on the next wl_surface.commit. The notification will only be posted for one frame unless requested again. For a wl_surface, the notifications are posted in the order the frame requests were committed.

The server must send the notifications so that a client will not send excessive updates, while still allowing the highest possible update rate for clients that wait for the reply before drawing again. The server should give some time for the client to draw and commit after sending the frame callback events to let it hit the next output refresh.
```

"For a wl_surface, the notifications are posted in the order the frame requests were committed."

so track them as pending or current, send pending for the buffer that got ready.

vaxerski: ok can be owned by state. Other comment stands.

gulafaran: something like this then?

src/protocols/types/SurfaceState.hpp
gulafaran: moved it back to CWLCallbackResource but having compositor as owner means trickier setup of both clearing it depending on situation and actually sending just the pending ones. so SSurfaceState holds the vector.

vaxerski:
Reverted due to severe performance degradation (frame callbacks accumulate and are never cleared, causing lag)

PR #11898, opened 2025-10-01
feat: apply [min|max]size window rules to dwindle & master layouts

kirpy0x:

This can allow users more/convenient control over their layouts by ensuring they can easily control the minimum/maximum sizes of all or certain windows. This is very beneficial for laptop owners who have a small display or large/ultra-wide display when docked, where a gapsout workspace rule may not suffice.

It can help more conveniently fix both aesthetic and practical issues regarding apps/websites with left/center/right aligned content which show poorly on windows taking up a wide screen or smaller apps that look poor when not in the compact window they were made for.

I think this applies very nicely with dwindle and master, providing the expected experience. I don't imagine users being unpleasantly surprised that their rule now applies to tiled windows, however if so, they'd have the option to append `, floating:1` to their rule, to go back to the experience prior to this PR.

- Removes floating check for the rule itself. (limiting the rule in config with `, floating:1` still applies)
- Updates the calculated window size
- Centers the window when smaller than available space.

Windows are still applied their normal space, so a minsized window may partially appear on top the down-rightward window when focused, else it will disappear beneath. I've been using Hyprland a few years but when I was hopping WMs a lot I believe this was the more common way to do it. 

Ready to go! tested with master, dwindle and floating on multiple monitors.

### DIFF
--- a/src/layout/DwindleLayout.cpp
+++ b/src/layout/DwindleLayout.cpp
@@ -219,12 +219,21 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
     static auto PCLAMP_TILED = CConfigValue<Hyprlang::INT>("misc:size_limits_tiled");
 
     if (*PCLAMP_TILED) {
-        Vector2D minSize = PWINDOW->m_windowData.minSize.valueOr(Vector2D{MIN_WINDOW_SIZE, MIN_WINDOW_SIZE});
-        Vector2D maxSize = PWINDOW->isFullscreen() ? Vector2D{INFINITY, INFINITY} : PWINDOW->m_windowData.maxSize.valueOr(Vector2D{INFINITY, INFINITY});
-        calcSize         = calcSize.clamp(minSize, maxSize);
+        const auto borderSize       = PWINDOW->getRealBorderSize();
+        Vector2D   monitorAvailable = PMONITOR->m_size - PMONITOR->m_reservedTopLeft - PMONITOR->m_reservedBottomRight -
+            Vector2D{(double)(gapsOut.m_left + gapsOut.m_right), (double)(gapsOut.m_top + gapsOut.m_bottom)} - Vector2D{2.0 * borderSize, 2.0 * borderSize};
 
-        if (!PWINDOW->onSpecialWorkspace() && !PWINDOW->m_isPseudotiled && (calcSize.x < availableSpace.x || calcSize.y < availableSpace.y))
-            calcPos += (availableSpace - calcSize) / 2.0;
+        Vector2D minSize = PWINDOW->m_windowData.minSize.valueOr(Vector2D{MIN_WINDOW_SIZE, MIN_WINDOW_SIZE}).clamp(Vector2D{0, 0}, monitorAvailable);
+        Vector2D maxSize =
+            PWINDOW->isFullscreen() ? Vector2D{INFINITY, INFINITY} : PWINDOW->m_windowData.maxSize.valueOr(Vector2D{INFINITY, INFINITY}).clamp(Vector2D{0, 0}, monitorAvailable);
+        calcSize = calcSize.clamp(minSize, maxSize);
+
+        calcPos += (availableSpace - calcSize) / 2.0;
+
+        calcPos.x = std::clamp(calcPos.x, PMONITOR->m_position.x + PMONITOR->m_reservedTopLeft.x + gapsOut.m_left + borderSize,
+                               PMONITOR->m_size.x + PMONITOR->m_position.x - PMONITOR->m_reservedBottomRight.x - gapsOut.m_right - calcSize.x - borderSize);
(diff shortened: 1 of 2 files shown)


Dregu:
How does this work with pseudo? I haven't tried the pr (nor do I even use pseudo), but feels like these rules should also apply to pseudo, which they currently don't on main.

Anyway, I welcome all features that can make single windows on ultrawide less awkward, and per window is even better than per workspace, cause I don't necessarily want to make all windows the same size either.

kirpy0x:
Yeah I've seen a few different ideas out there for reasons people may want certain size limits. Which just made me realize the size rule may be a good candidate for including as well maybe in another PR if it sounds welcome. My size rules are all paired with center/float.

This will apply to pseudo as well just about as you'd probably expect I imagine. If the space has less room than the maxsize it won't affect it. If more room, the window will open as it would otherwise but within the limits.

I just noticed a quirk I didn't see before where they were moving off center of their available space after reaching their min/max sizes and pushed a correction.

vaxerski:
does this essentially do what pseudo does? That's what it looks like. Wouldn't it make more sense to have the other window take more space?

kirpy0x:
Only in the sense that it may not take the full space it's given. Pseudo may be used by some to get this affect, but this will automate the effort based on the rule. Also, for tiled windows the resize bindings still work the same so you might see it as a way to get automated pseudo affect without stopping your resize bindings from resizing the space it's in.

I did spend a bit of time trying to have their window spaces only take up the size within the limits while still allowing clean manual resizing of the space. There's a lot to account for there making sure it plays nicely with all the spaces around it and the farther I went, the more bugs appeared. I think if we go there it should probably be another pr.

vaxerski:
Fair enough.

vaxerski (commented):
missing a config option to disable this behavior (I for example dont want it)
missing tests

kirpy0x:
Added config option size_limits_tiled default false. Also added tests to check size's are limited and window centered within space.

src/layout/DwindleLayout.cpp
```diff
@@ -214,6 +214,19 @@ void CHyprDwindleLayout::applyNodeDataToWindow(SDwindleNodeData* pNode, bool for
...
+    Vector2D    availableSpace = calcSize;
+
+    static auto PCLAMP_TILED = CConfigValue<Hyprlang::INT>("misc:size_limits_tiled");
+
+    if (*PCLAMP_TILED) {
+        Vector2D minSize = PWINDOW->m_windowData.minSize.valueOr(Vector2D{MIN_WINDOW_SIZE, MIN_WINDOW_SIZE});
+        Vector2D maxSize = PWINDOW->isFullscreen() ? Vector2D{INFINITY, INFINITY} : PWINDOW->m_windowData.maxSize.valueOr(Vector2D{INFINITY, INFINITY});
+        calcSize         = calcSize.clamp(minSize, maxSize);
+
+        if (!PWINDOW->onSpecialWorkspace() && !PWINDOW->m_isPseudotiled && (calcSize.x < availableSpace.x || calcSize.y < availableSpace.y))
```
Dregu: Testing this and looks like I could replace my gapsout based solution for centering single windows with this and have more options for different sizes per window/tag. However seems weird that the windows are not centered on special workspace or if bigger than their slot. Would make a lot more sense to me if this condition just didn't exist and windows were either always centered or even better if big windows tried really hard to fit on the screen (if on the right side of screen, move left until inside the gaps to cover the left window instead of going off screen on the right)

kirpy0x: Yeah I think you're right that looks good with the special workspace.

I also made it keep windows fully on the screen.

Dregu:
Yeah better. I think it's not counting border sizes though, i.e. windows clamped by gaps with this rule enabled are 2*bordersize wider than windows clamped by gaps normally, i.e. 6 pixels wider with `general:border_size = 3`

kirpy0x:
Added a pr for the wiki and made the tests use a unique kitty window to stop from affecting other tests.

Also @Dregu I added bordersize to the clamps.

hyprtester/src/tests/main/window.cpp
```diff
@@ -228,6 +228,31 @@ static bool test() {
...
+        if (!spawnKitty("kitty_maxsize_b"))
+            return false;
+
+        OK(getFromSocket("/dispatch focuswindow class:kitty_maxsize"));
+        str = getFromSocket("/activewindow");
+        EXPECT_CONTAINS(str, "size: 1000,500");
+
+        OK(getFromSocket("/keyword misc:size_limits_tiled 0"));
+        Tests::killAllWindows();
+        EXPECT(Tests::windowCount(), 0);
```
vaxerski, line 270: considering you added this to master as well, you should test master behavior as well (you can `/keyword general:layout master` and then `/reload` to revert)

kirpy0x: Okay I got these added for master.

PR #7919, opened 2024-09-26
Implement new Input Capture protocol

3l0w:
#### Describe your PR, what does it fix/add?

This PR implements the input capture protocol defined in hyprwm/hyprland-protocols#8.
It also fixed a bug where `wl_output::geometry` `x` and `y` arguments where always 0. 
The `wl_output::geometry` event is now sent when the output layout change (Help to follow `x` and `y` modifications).

#### Is it ready for merging, or does it need work?
It still need work, like sending the keyboard layout.

TODO List:
* [x] See if the fd multiplication problem is fixed
* [x] Calculate the invalids barriers on SetPointerBarriers
* [x] Replace the trash impl of get_line_intersection by one that follow the specs
* [x] Warp pointer when the cursor is released by the capturing program
* [x] Move most of the impl to Hyprland
* [x] Impl some kind of security around https://github.com/hyprwm/Hyprland/pull/9930, maybe move some of the impl ive done here to Hyprland
* [x] Fix the isLock shenanigans 
* [x] See if i limit to only one client, (for now multiple clients can connect, first arrive with his zones, first to be triggered)
* [x] Fix unable to access tty when captured

PR #11917, opened 2025-10-03
input manager: do not share keys/mods states from grabbed keyboards with ime keys/mods

xzn:

Issue discussed here:
https://github.com/hyprwm/Hyprland/discussions/11715#discussioncomment-14531246

In handling the interaction between ime grabbing and virtual keyboard from the ime client: when keys are pressed then released repeatedly without delay (such as when using automated tools), multiple presses may not register fully and less than the number of pressed are registered instead. Happens when fcitx5 is running.

Hope it doesn't interact badly with Input Capture protocol PR #7919

Could use some testing, (since I don't have repo of the issue on my end)

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -1449,7 +1449,11 @@ void CInputManager::onKeyboardKey(const IKeyboard::SKeyEvent& event, SP<IKeyboar
             IME->setKeyboard(pKeyboard);
             IME->sendKey(event.timeMs, event.keycode, event.state);
         } else {
-            const auto PRESSED  = shareKeyFromAllKBs(event.keycode, event.state == WL_KEYBOARD_KEY_STATE_PRESSED);
+            const auto ISPRESSED = event.state == WL_KEYBOARD_KEY_STATE_PRESSED;
+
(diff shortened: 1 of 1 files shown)


xzn:
I wonder if the mods states sent by the keyboard from the ime grab client should be ignored for keybinds purposes, i.e.

Move https://github.com/hyprwm/Hyprland/pull/11917/files#diff-2ba0a4b3fea7c686b5bcf66318fa5912cb5d78a313318bf7ce2d51faf3b7a88aR1494 into the conditional as well?

xzn:
Moved mods states update for keybinds into the conditional (so ime virtual keyboard doesn't affect keybinds). Incidentally fixed a regression from earlier commit in this PR where an assignment was missing.

LukeXuan:
I can confirm the problem seems to be fixed for the following setup: I ran a second Hyprland session inside existing one, and started both fcitx5 and emacs using terminal in the nested Hyprland.

When compared to main branch, the macro corruption issue I mentioned in the original issue seem to disappear.

xzn:
I've been using this patch myself the past few weeks and have had no problems. Although my keyboard is very standard so it doesn't mean much.

In any case reviews are appreciated!

PR #11922, opened 2025-10-04
dispatchers: add forceidle

MithicSpirit:

The `forceidle` dispatcher resets all ext-idle-notify timers as if the user had
already been idle for the specified number of seconds. If a notification has
already fired, but would now be set with a nonzero delay, then it is reset.
Conversely, if a timer has not yet fired, but would now be set to a nonpositive
delay, then it is immediately fired. This process ignores any existing
inhibitors, but timers are otherwise reset as normal if any new inhibitors are
created or destroyed.

A small refactor of how ext-idle-notify timers was necessary for this, but, in
my opinion, it makes the code better and more robust.

As with the `dpms` dispatcher, this should not be set with a keybind directly.

I don't use any idle inhibitors on my system, so I think more testing might be
good on that front. The way that `forceidle` v. inhibitors is resolved on a
recency basis might be a bit weird, but doing anything else felt unnecessarily
complicated (I did experiment with it, but ultimately decided against it).
Overall, the one weird piece of behavior here is where removing an inhibitor
will reset a timer that was set by `forceidle`, but I think this can be fixed if
desired.

Other than code review (obviously) and maybe some testing with inhibitors, it
should be ready for merging.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -4,6 +4,7 @@
 #include "../protocols/LayerShell.hpp"
 #include "../protocols/ShortcutsInhibit.hpp"
 #include "../protocols/GlobalShortcuts.hpp"
+#include "../protocols/IdleNotify.hpp"
 #include "../protocols/core/DataDevice.hpp"
 #include "../render/decorations/CHyprGroupBarDecoration.hpp"
 #include "KeybindManager.hpp"
@@ -142,6 +143,7 @@ CKeybindManager::CKeybindManager() {
     m_dispatchers["event"]                          = event;
     m_dispatchers["global"]                         = global;
     m_dispatchers["setprop"]                        = setProp;
+    m_dispatchers["forceidle"]                      = forceIdle;
 
     m_scrollTimer.reset();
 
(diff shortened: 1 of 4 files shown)


vaxerski:
> I don't use any idle inhibitors on my system

What's the motivation behind this then?

MithicSpirit:
The main usecase is to automatically reset hypridle when I unlock my system with fprintd. It also makes it so I can skip waiting for hypridle to do its things, while still having it reset properly when I'm no longer idle.

vaxerski:
wouldnt it just be enough to add a dispatcher to "trigger" a "fake activity"?

MithicSpirit:
> wouldnt it just be enough to add a dispatcher to "trigger" a "fake activity"?

Yes for the first usecase, but it doesn't address the second.

vaxerski:
yeah but whats the usecase for the second?

MithicSpirit:
When I go to bed, I don't want to have to wait 10 minutes for hypridle to do its thing. My previous solution was to have a script that does everything, and then another that resets it and is run when hyprlock exits. However, this is not robust—it needs to wait for a full unlock rather than any activity, and the setup and reset commands will usually be run twice.

vaxerski:
wiki mr needed

Issue #11929, opened 2025-10-04
Data race in Debug::log() causing crashes with multithreaded plugins


Daviey:
I've been debugging some crashes in hypr-dynamic-cursors (see VirtCode/hypr-dynamic-cursors#99 and VirtCode/hypr-dynamic-cursors#100) and traced it back to what looks like a threading issue in Hyprland's logging code.

**What's happening:**

The plugin uses `std::async` to load cursor themes in the background, and calls `Debug::log()` from those worker threads. During config reloads, both the main thread and background threads can be logging at the same time, which causes crashes like:

```
double free or corruption (out)
SIGABRT (abort due to memory corruption)
malloc_printerr → _int_free → _int_free_merge_chunk
```

**The issue:**

Looking at the code, the templated `Debug::log(eLogLevel, fmt, args...)` ([here](https://github.com/hyprwm/Hyprland/blob/3bcfa94ee4189faaa4daf661949e88cf28c00d94/src/debug/Log.hpp#L44-L76)) locks `m_logMutex`, formats the string, then calls the non-templated `Debug::log(eLogLevel, std::string)` ([here](https://github.com/hyprwm/Hyprland/blob/3bcfa94ee4189faaa4daf661949e88cf28c00d94/src/debug/Log.cpp#L20-L76)).

But the non-templated version doesn't lock anything - it just writes directly to `m_rollingLog`, `m_logOfs`, etc. So if a background thread calls it directly at the same time, you get concurrent access to shared state without synchronization.

Daviey:
Discussion created at https://github.com/hyprwm/Hyprland/discussions/11930

PR #11931, opened 2025-10-04
debug: fix data race in Debug::log()

Daviey:
Fixes a data race in the logging system that causes crashes when plugins use background threads.

**The issue:**

The templated `Debug::log(eLogLevel, fmt, args...)` locks `m_logMutex` and calls the non-templated `Debug::log(eLogLevel, std::string)`, but the non-templated version doesn't lock anything. This causes crashes when background threads (like in hypr-dynamic-cursors) call `Debug::log()` while the main thread is also logging.

**The fix:**

Added a private `logImpl()` that does the actual logging work, with both public `log()` entry points handling the locking. This uses a regular mutex instead of recursive_mutex and avoids duplicate checks.

Fixes: #11929
Fixes: VirtCode/hypr-dynamic-cursors#99

See also: https://github.com/hyprwm/Hyprland/discussions/11930

### DIFF
--- a/src/debug/Log.cpp
+++ b/src/debug/Log.cpp
@@ -26,7 +26,7 @@ void Debug::log(eLogLevel level, std::string str) {
 
     std::lock_guard<std::mutex> guard(m_logMutex);
 
-    std::string coloredStr = str;
+    std::string                 coloredStr = str;
     //NOLINTBEGIN
     switch (level) {
         case LOG:
--- a/src/debug/Log.cpp
+++ b/src/debug/Log.cpp
@@ -24,6 +24,8 @@ void Debug::log(eLogLevel level, std::string str) {
     if (m_shuttingDown)
         return;
 
+    std::lock_guard<std::mutex> guard(m_logMutex);
+
     std::string coloredStr = str;
     //NOLINTBEGIN
     switch (level) {
--- a/src/debug/Log.hpp
+++ b/src/debug/Log.hpp
@@ -42,8 +42,6 @@ namespace Debug {
     template <typename... Args>
     //NOLINTNEXTLINE
     void log(eLogLevel level, std::format_string<Args...> fmt, Args&&... args) {
-        std::lock_guard<std::mutex> guard(m_logMutex);
-
         if (level == TRACE && !m_trace)
             return;
 


vaxerski:
I mean, wouldn't that be a plugin bug?

Daviey:
That was my first thought and I looked to try and fix it there.  However, the templated `Debug::log()` already has mutex protection, which suggests thread-safety is expected. The issue is the inconsistency between the two overloads:

  1. Template version locks `m_logMutex`, formats the message, then calls the non-template version
  2. Non-template version has no locking

So.... when:
  - Thread A (main): calls template version it locks mutex then calls non-template (still holding lock)
  - Thread B (worker): calls non-template directly and no lock
  - Both access shared state (m_rollingLog, m_logOfs) concurrently

If `Debug::log()` isn't meant to be thread-safe, why does the template version lock at all? And if it is meant to be thread-safe (which the mutex suggests), then the non-template overload should also lock.

The plugin could work around this by only logging from the main thread, but that seems like an odd requirement when the logging API already has partial thread-safety built in?

VirtCode:
> I mean, wouldn't that be a plugin bug?

I think the logging is supposed to be thread-safe since https://github.com/hyprwm/Hyprland/commit/548968279926a73d7ff19a9a185c977c50d56756 which is why I thought I'd just use it. Yes I know you explicitly discourage using extra threads in plugins, but loading a large cursor theme is horrendously slow that it would otherwise increase compositor startup times with dynamic cursors by literal seconds.

The discrepancy between the templated logging function and the non-templated has been bothering me anyways since quite some time because if you have enabled timestamps in the log, the timestamp only gets introduced when using the templated version, and not the normal one. This means the log is pretty inconsistent, only introducing timestamps in some cases, like e.g.
```
[LOG] [CPrimarySelectionProtocol] CPrimarySelectionProtocol::updateSelection: cannot send selection to a client without a data_device
[LOG] [22:30:35.317512890] CWLSurface 5581a8441d30 called destroy()
[LOG] Cleanup: destroyed a layersurface
[LOG] [22:30:35.322383355] Framebuffer created, status 36053
```

So maybe we should do this in this PR too if we are already at it.

As for the fix in this PR, it would probably also be enough to just move the mutex locking from the templated version into the normal version, because actual writes which could cause issues only happen in there. The only thing the `logImpl` approach saves is the _second_ trace and shutdown check in case of the templated version, which I don't know whether that's worth an extra function call.

vaxerski:
Ah, right makes sense.

vaxerski:
> As for the fix in this PR, it would probably also be enough to just move the mutex locking from the templated version into the normal version, because actual writes which could cause issues only happen in there

Yeah thjats bettr

Daviey:
Updated to use the simpler approach you suggested. Removed the `logImpl()` function entirely and just moved the mutex lock from the template version into the non-template version.

Now the template version checks trace/shutdown conditions, formats the message (including timestamp if enabled), and calls the non-template version with the formatted string. The non-template version checks trace/shutdown conditions, acquires the mutex, and does all the actual writes to m_rollingLog, m_logOfs, etc.

This way we only lock once per log call, and all the shared state writes are protected. No need for the extra function or any duplicate checks.

@VirtCode - you're right that this is simpler and achieves the same thing. The timestamp inconsistency you mentioned is still there (template adds it, non-template doesn't), but that's a separate issue from the thread safety fix.

Thanks for the feedback, much cleaner this way.

vaxerski (dismissed):

Daviey:
@vaxerski Thanks for your approval! I've pushed again to satisfy the failing CI clang-format check. The formatting issue wasn't from my change, but since it was in the same file I thought I should make it pass anyway.

Could you re-approve/merge when you get a chance? Thanks!

PR #11946, opened 2025-10-06
systeminfo: log system package versions

vaxerski:
Logs system hypr* versions to instantly tell people who have mismatching ABI to heck off

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -929,3 +929,42 @@ std::string deviceNameToInternalString(std::string in) {
     std::ranges::transform(in, in.begin(), ::tolower);
     return in;
 }
+
+static const std::vector<const char*> PKGCONF_PATHS = {"/usr/lib/pkgconfig",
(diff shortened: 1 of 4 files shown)


src/helpers/MiscFunctions.cpp
```diff
@@ -929,3 +929,41 @@ std::string deviceNameToInternalString(std::string in) {
     std::ranges::transform(in, in.begin(), ::tolower);
     return in;
 }
+
+static const std::vector<const char*> PKGCONF_PATHS = {"/usr/lib/pkgconfig", "/usr/local/lib/pkgconfig"};
```
fufexan, line 933: Can you make this a cmake/meson var? I don't wanna patch files for nix.

vaxerski: nix doesnt have this problem tho? Like when you rebuild it rebuilds everything no?

fufexan: Correct, nvm then

PR #11964, opened 2025-10-08
buffer: move texture creation to commit time

gulafaran:
move creating texture away from buffer attach into commitstate in an
attempt to postpone gpu work until its really needed. best case scenario
gpu clocks have ramped up before because we are actively doing things
causing surface states and a commit to happend meaning less visible lag.

this is one of those touch it and sneaky bugs shall appear.

locally on nvidia this makes a night and day difference.


tannerellen:
I tested a bunch of stuff including a handful of games and emulators as well as all the software I use during a given week and things are looking good on my system with an AMD GPU (7900 GRE).

gulafaran:
> I tested a bunch of stuff including a handful of games and emulators as well as all the software I use during a given week and things are looking good on my system with an AMD GPU (7900 GRE).

big thanks, :heart:

PR #11971, opened 2025-10-08
core: defer wl_surface.leave until frame callbacks complete

tempox777:
fixes chromium based browsers (brave, chrome, edge) crashing with sigtrap when moving windows between visible monitors.

the crash occurs because hyprland sends wl_surface.leave events immediately while pending frame callbacks exist for the old output. chromium ozone/wayland backend expects all frame callbacks from an output to complete before receiving a leave event. when this ordering is violated, chromium hits a dcheck assertion and triggers a trap.

this fix defers leave events until after all pending frame callbacks are sent, ensuring proper event ordering that matches client expectations.

changes:
- added m_pendingLeaveOutputs to track monitors requiring deferred leave events
- modified CWLSurfaceResource::leave() to defer leave when callbacks are pending
- modified CWLSurfaceResource::frame() to send deferred leave events after callbacks
- clear pending leaves on surface unmap to prevent stale events

the crash only occurred when moving between visible monitors (both actively rendering), not when moving to hidden workspaces (no active frames). setting wayland_debug=1 avoided the crash by serializing events, which confirmed this was a timing issue.

tested with brave browser moving between two monitors on different workspaces.

### DIFF
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -300,6 +300,12 @@ void CWLSurfaceResource::leave(PHLMONITOR monitor) {
         return;
     }
 
+    if (!m_current.callbacks.empty()) {
+        if (std::ranges::find(m_pendingLeaveOutputs, monitor) == m_pendingLeaveOutputs.end())
+            m_pendingLeaveOutputs.emplace_back(monitor);
+        return;
+    }
+
     std::erase(m_enteredOutputs, monitor);
 
     m_resource->sendLeave(output->getResource().get());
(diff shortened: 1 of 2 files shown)


UjinT34:
https://wayland.app/protocols/wayland#wl_surface:request:frame and
https://wayland.app/protocols/wayland#wl_surface:event:leave don't mention any deferring. There seems to be no reason to expect that leave events are delivered after all frame callbacks. Am I missing something? If not then it's a chromium bug.

tempox777:
> https://wayland.app/protocols/wayland#wl_surface:request:frame and https://wayland.app/protocols/wayland#wl_surface:event:leave don't mention any deferring. There seems to be no reason to expect that leave events are delivered after all frame callbacks. Am I missing something? If not then it's a chromium bug.

> 

the spec doesn’t require deferring wl_surface.leave, true. this patch isn’t about spec compliance actually but about client compatibility so basically 
most compositors (mutter, kwin, sway, weston) defer leave until all pending frame callbacks on that output are done. but chromium’s ozone backend implicitly assumes that ordering.
so hyprland’s current behavior (sending leave and enter back-to-back) is spec legal, but also that assumption  brave in particular hits a DCHECK and traps when this happens.
chrome itself doesn’t crash because it doesn’t trigger the same compositor codepath; brave enables extra ozone/gpu features that create additional frame callbacks. when leave hits early, one of those callbacks is still pending so it causes the trap.

UjinT34:
> most compositors (mutter, kwin, sway, weston) defer leave until all pending frame callbacks on that output are done

Do they defer on purpose? kwin doesn't seem to specifically defer this event. It may happen after the callback because of a different event loop organisation. Perhaps we can reorder our loop to handle callbacks first and achieve the same result. Might also help with some other stuff.

tempox777:
> > most compositors (mutter, kwin, sway, weston) defer leave until all pending frame callbacks on that output are done
> 
> Do they defer on purpose? kwin doesn't seem to specifically defer this event. It may happen after the callback because of a different event loop organisation. Perhaps we can reorder our loop to handle callbacks first and achieve the same result. Might also help with some other stuff.

that makes sense yeah kwin probably ends up doing it in the right order because of how its event loop is structured rather than through explicit deferral so in hyprland’s case, the leave seems to get sent before the last frame callback completes, so
reordering the event loop could definitely achieve the same output and probably be a better fix

PR #5874, opened 2024-05-04
Enable LTO in supported compilers

In-line:
#### Describe your PR, what does it fix/add?

Enabled LTO, because why not? 

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

No

#### Is it ready for merging, or does it need work?

It's ready

PR #11972, opened 2025-10-08
Disable LTO

VirtCode:
~~Disables some optimizations on `CPointerManager::move(Vector2D)` such that LTO on gcc doesn't optimize it out.~~ I need this method in dynamic-cursors because it lets me differentiate between warps and normal moves.

This PR now disables LTO completely to avoid similar problems in the future.

fixes https://github.com/VirtCode/hypr-dynamic-cursors/issues/98

nope

yes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,9 @@ add_compile_options(
   -Wno-clobbered
   -fmacro-prefix-map=${CMAKE_SOURCE_DIR}/=)
 
+# disable lto as it may break plugins
+add_compile_options(-fno-lto)
+
 set(CMAKE_EXECUTABLE_ENABLE_EXPORTS TRUE)
 set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)
 


VirtCode:
are you sure? I don't know what the performance gains of LTO are in hyprland specifically but this seems a bit nuclear to me. I agree adding things for random plugins isn't too clean either but at least doesn't have too much of a performance impact.

vaxerski:
> are you sure?

No, I am vaxry.

You can benchmark it but I doubt there's gonna be any meaningful difference. We already do O3 and Hyprland doesn't have many giga hot paths

vaxerski:
Also, https://github.com/hyprwm/Hyprland/pull/5874

is it even enabled in cmake?

VirtCode:
> is it even enabled in cmake?

I don't think so but it still uses `CFLAGS`/`CXXFLAGS`/`LTOFLAGS` from the env. And the arch build system sets them to enable LTO. So yeah this only affects builds on arch really that I know of.

But in this case I'll look to "force disable" lto in cmake somewhere

UjinT34:
Adding `options=('!lto')` into PKGBUILD should be a better solution

VirtCode:
also thought of that but some plugins would still be broken for users on other distros which build with LTO enabled

vaxerski:
I'd just add `-fno-lto` to our Hyprland build and call it a day.

VirtCode:
Sorry this took so long was quite busy the last few days.

Anyways I have now disabled LTO globally by passing `-fno-lto` to the compile flags. This flag then appears later in the compile cmdline and seems to override any `-flto` passed earlier e.g. via the env. I have not changed the linker flags because I couldn't get it to link against `udis86` easily with `-fno-lto` passed to the linker. But it is enough just disabling it for hyprland's compilation step to avoid the optimizations which lead to the issues here.

In my testing this works and the issue also no longer appears if this branch is also built into an arch package, so it should be ready.

PR #11975, opened 2025-10-08
build: add appimage creation to pipeline

ciscon:
This adds the creation of a runable appimage file to the github pipeline so that newer builds/releases might be used on operating systems that do not yet have all of the required (or up to date) dependencies.

Ostensibly this should be a part of the release creation and not just an artifact during the normal pipeline, but for testing purposes I've gone ahead and thrown it into its own job and creates an artifact.

Pending comments.


Conarius:
That seems a bit counterintuitive since AppImage doesn't support Wayland at all and the creator openly dislikes Wayland, basically refusing to support Wayland at any level.

fufexan:
image

ciscon:
I'm not sure how that's relevant as it's just a wrapper allowing the binary to run.  I could certainly just reimplement appimage from scratch but that does seem like a pointless reinventing of the wheel when it works perfectly fine and I'm not particularly concerned with the creator's ideologies.

vaxerski:
...why?

PR #11986, opened 2025-10-09
feat: add more uniforms to the screen shader

Mozzarella32:
Enable more ricing by passing more information to the screen shader.
The added uniforms are:
- pointer_shape from the cursor-shape-v1 protocol
- pointer_shape_previous with
- pointer_switch_time to blend between shapes
- pointer_size scaled size as used by the normal cursor
- pointer_pressed_positions[32] with
- pointer_pressed_times[32] and
- pointer_pressed_killed(32 bits) for click/touch animations and if they killed something
- pointer_pressed_touched(32 bits) for touch animations
- pointer_inactive_timeout with
- pointer_last_active to smoothly fade the pointer out
- pointer_killing if the next click kills a window
- pointer_hidden to hide it when the cursor is hidden (excluding by cursor:invisible as this config value can be used to turn off the normal cursor, which is useful when drawing it with the screen shader)

https://github.com/user-attachments/assets/d48e5f8e-4c0c-4c11-9d21-f19a05ec9a46

For local testing (the shader had to be renamed from .frag to .txt to be attachable to this pr):

[cursorShaderDemo.txt](https://github.com/user-attachments/files/22910461/cursorShaderDemo.txt)
[shapes.html](https://github.com/user-attachments/files/22802745/shapes.html)

pointer_hidden doesn't work for the css cursor: none in Firefox, as it sends a surface with size 0, 0 to hide the cursor and when cursor:invisible from https://github.com/hyprwm/Hyprland/pull/11058 is active, the cursor image is reset. This means there seems to be no way to tell that the browser wants to hide the image.

ready for merging as i see no easy way to fix none issue and don't think its that impotent.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -409,7 +409,7 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() : m_drmFD(g_pCompositor->m_drmRenderNode.fd >
 #endif
     };
 
-    static auto P2 = g_pHookSystem->hookDynamic("mouseButton", [addLastPressToHistory](void* self, SCallbackInfo& info, std::any e) {
+    static auto P2 = g_pHookSystem->hookDynamic("mouseButton", [](void* self, SCallbackInfo& info, std::any e) {
         auto E = std::any_cast<IPointer::SButtonEvent>(e);
 
         if (E.state != WL_POINTER_BUTTON_STATE_PRESSED)
@@ -418,7 +418,7 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() : m_drmFD(g_pCompositor->m_drmRenderNode.fd >
         addLastPressToHistory(g_pInputManager->getMouseCoordsInternal(), g_pInputManager->getClickMode() == CLICKMODE_KILL, false);
     });
 
-    static auto P3 = g_pHookSystem->hookDynamic("touchDown", [addLastPressToHistory](void* self, SCallbackInfo& info, std::any e) {
+    static auto P3 = g_pHookSystem->hookDynamic("touchDown", [](void* self, SCallbackInfo& info, std::any e) {
         auto E = std::any_cast<ITouch::SDownEvent>(e);
 
         auto PMONITOR = g_pCompositor->getMonitorFromName(!E.device->m_boundOutput.empty() ? E.device->m_boundOutput : "");


src/helpers/CursorShapes.hpp
```diff
@@ -40,4 +41,56 @@ constexpr std::array<const char*, 35> CURSOR_SHAPE_NAMES = {
     "zoom-in",
     "zoom-out",
 };
-// clang-format on
\ No newline at end of file
+
+constexpr std::array<std::string_view, 47> CURSOR_SHAPE_NAMES_FINAL_SHADER = {
```
vaxerski: why not just pass the uint?

src/managers/input/InputManager.hpp
```diff
@@ -241,6 +246,8 @@ class CInputManager {
     void               processMouseDownNormal(const IPointer::SButtonEvent& e);
     void               processMouseDownKill(const IPointer::SButtonEvent& e);
 
+    void               addLastPressed(const Vector2D& pos, bool normal);
```
vaxerski: bothers inputmanager for no reason, why not add an event hook for mouseButton in the ctor of opengl?

src/helpers/CursorShapes.hpp
Mozzarella32: Not quite sure what you mean. If you are talking about SSetShapeEvents::shape from the CursorShapeProtocol, then i can not use it, as when changing window splits with the mouse top_size...bottom_right_corner are used which uses the CHyprRenderer::setCursorFromName.

vaxerski: I meant the uint32_t code from wp_cursor_shape_v1

src/render/OpenGL.cpp
```diff
@@ -381,6 +385,50 @@ CHyprOpenGLImpl::CHyprOpenGLImpl() : m_drmFD(g_pCompositor->m_drmRenderNode.fd >
     m_globalTimer.reset();
 
     pushMonitorTransformEnabled(false);
+
+    auto addLastPressToHistory = [this](const Vector2D& pos, bool killing, bool touch) {
```
vaxerski: make static and pass by ref to fns to avoid copies

src/render/OpenGL.cpp
```diff
@@ -1239,30 +1287,59 @@ void CHyprOpenGLImpl::applyScreenShader(const std::string& path) {
...
-        // Since the screen shader could change every frame, damage tracking *needs* to be disabled
-        g_pConfigManager->addParseError("Screen shader: Screen shader uses uniform 'time', which requires debug:damage_tracking to be switched off.\n"
-                                        "WARNING: Disabling damage tracking will *massively* increase GPU utilization!");
-    }
     m_finalScreenShader.uniformLocations[SHADER_TEX_ATTRIB] = glGetAttribLocation(m_finalScreenShader.program, "texcoord");
     m_finalScreenShader.uniformLocations[SHADER_POS_ATTRIB] = glGetAttribLocation(m_finalScreenShader.program, "pos");
-    if (m_finalScreenShader.uniformLocations[SHADER_POINTER] != -1 && *PDT != 0 && !g_pHyprRenderer->m_crashingInProgress) {
-        // The screen shader uses the "pointer_position" uniform
+
+    auto uniformRequireNoDamage = [this](eShaderUniform uniform, const std::string& name) {
```
vaxerski: can be static

src/render/Renderer.cpp
```diff
@@ -2027,6 +2028,42 @@ void CHyprRenderer::setCursorFromName(const std::string& name, bool force) {
         return;
 
     m_lastCursorData.name = name;
+
+    auto getShapeOrDefault = [](std::string_view name) -> wpCursorShapeDeviceV1Shape {
```
vaxerski: can be static

vaxerski:
you could've just made 3 commits here but that works too sure

vaxerski:
needs a rebase now

PR #11990, opened 2025-10-10
ensure dmabuf textures isnt reacreated on same buffer commits

gulafaran:
dmabuffers doesnt always recreate on attach, they send damage and we damage the area. store the texture in the buffer and return the pointer to it if it already exist.


Honkazel:
libc++ still doesn't support `std::move_only_function`

gulafaran:
> libc++ still doesn't support `std::move_only_function`

oh that sucks -_-

gulafaran:
> libc++ still doesn't support `std::move_only_function`

dropped that change then

PR #11993, opened 2025-10-10
renderer: send frame events on onPresented

gulafaran:
so previously the frame was sent on surface onPresented on both discarding and presenting also if the monitor had no damage in renderer, move it to simply present of the monitor. if using new_render_sched this means the third monitor frame rendering might never hit it because it has no pending presentation and yet the monitor has damage from other things, as such it stopped rendering and recieving frame calls.

rename it to sendFrameEventsToMonitor since thats all it was used for. and send it an extra time in new_render_sched

### TODO

- [ ] Figure out if this goes wrong when the compositor doesnt keep up on slow gpus that doesnt pageflip at native monitor HZ
- [ ] Send frame events properly to hyprlandsurfaces
- [ ] Do we want a timer based on monitor hz that sends the frames eg. `render::timer_based_frames = true`

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -112,10 +112,8 @@ void CMonitor::onConnect(bool noRule) {
             ts = nullptr;
         }
 
-        if (!ts)
-            PROTO::presentation->onPresented(m_self.lock(), Time::steadyNow(), event.refresh, event.seq, event.flags);
-        else
-            PROTO::presentation->onPresented(m_self.lock(), Time::fromTimespec(event.when), event.refresh, event.seq, event.flags);
+        auto time = ts ? Time::fromTimespec(event.when) : Time::steadyNow();
+        PROTO::presentation->onPresented(m_self.lock(), time, event.refresh, event.seq, event.flags);
 
         if (m_zoomAnimFrameCounter < 5) {
             m_zoomAnimFrameCounter++;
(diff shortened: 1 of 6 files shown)


vaxerski:
> try to makes as few dereferences as possible since this is called once per frame.

wouldn't it be optimized out in release builds? the syntax is just ugly

gulafaran:
> > try to makes as few dereferences as possible since this is called once per frame.
> 
> wouldn't it be optimized out in release builds? the syntax is just ugly

yeah it would, just since going to arch i kinda messed up the pkgbuild building in debug mode, profiled debug mode hard lol. removed.

vaxerski:
I tested this. It stops the absolute freeze on firefox, but instead it's very choppy where it would freeze before. Still not ideal. Before the original frame changes, it was smooth.

UjinT34:
#11971 might be related. freezes/crashes might be caused by clients expecting a certain order of events which isn't clearly defined in protos but seems to be guaranteed by other compositors.

gulafaran:
yeah i kinda went on with fifo PR for a while, this PR is not done. but in general there are other issues aswell with new_render_scheduling one being breaking explicit sync aswell. frame A, endRender() set fences, Present A, buffer drops. `out of time` , new renderSched. begins a render directly again. and poof all bets are off. 

then comes other bits im not sure of setting infence on that third render but not committing it until later but still setting fences to buffers if there is any. etc etc.

Issue #11968, opened 2025-10-08
Add a more robust ABI check system for plugins


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11954 et al

## Description
Currently, we only check the hyprland hash, but after ABI breaks in deps (e.g. hu) plugins might crash without proper ABI checks.

PR #12001, opened 2025-10-10
plugins: incorporate hyprdep ABI into plugin info

vaxerski:
This adds hyprdep minor version check to hyprland version internally for plugins to fail loading when there is a mismatch.

Fixes https://github.com/hyprwm/Hyprland/issues/11968

TODO:
- [ ] make hyprpm aware of this


vaxerski:
piece of shit @fufexan fix pls

ItsOhen:
I made a thingamabob to fix this. Will break all plugins.. but on the bright side, they won't die in a fire in the future. No idea how to make hyprpm aware though.

https://github.com/hyprwm/Hyprland/compare/main...ItsOhen:Hyprland:hyprpm_deps

fufexan:
@vaxerski how do you build the tester?

vaxerski:
@fufexan cmake

ItsOhen:
With @vaxerski approach every plugin would need to include a pkg-config check against all the libs to get the versions to match.
What's going on here is that building the tests and plugin directly when building hyprland with cmake use the pkg-config check and add_compile_definitions, so it works fine.
But as soon as they try and do their own thing, no *_VERSION to be found. :(
Which also means meson.

vaxerski:
Ah, right. They need to be added to config.h.in. That shouldn't break plugins.

PR #12014, opened 2025-10-12
fix #9987

### LINKED ISSUE #9987: When a wayland popup is closed, refocus grabs the wrong window with follow_mouse = 2
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/9958

## Description

When you have follow_mouse = 2, and type something in speedcrunch (e.g. f, for farad), which opens a popup, and then move your mouse outside of it, and continue typing, focus switches. It shan't.

0rtz:
Fix of [issue](https://github.com/hyprwm/Hyprland/issues/9987)
As I understand, the problem is that when popup that hold an input grab is destroyed, Hyprland always switches focus via `g_pInputManager->refocus()` to the window that is currently under the cursor without respecting `follow_mouse` config variable.
My proposed fix is to switch focus to the parent window of the destroyed popup instead of focusing window under the cursor when using follow_mouse 0/2/3

I see that sometimes even though keyboard focus is always maintained correctly, window decorations of application under the cursor is activated even though it does not have focus

Ready

### DIFF
--- a/src/managers/SeatManager.cpp
+++ b/src/managers/SeatManager.cpp
@@ -616,8 +616,55 @@ void CSeatManager::setCurrentPrimarySelection(SP<IDataSource> source) {
 void CSeatManager::setGrab(SP<CSeatGrab> grab) {
     if (m_seatGrab) {
         auto oldGrab = m_seatGrab;
+
+        // Try to find the parent window from the grab
+        PHLWINDOW parentWindow;
+        if (oldGrab && oldGrab->m_surfs.size()) {
+            // Try to find the surface that had focus when the grab ended
+            SP<CWLSurfaceResource> focusedSurf;
+            auto                   keyboardFocus = m_state.keyboardFocus.lock();
+            auto                   pointerFocus  = m_state.pointerFocus.lock();
+
+            // Check if keyboard or pointer focus is in the grab
+            for (auto const& s : oldGrab->m_surfs) {
+                auto surf = s.lock();
+                if (surf && (surf == keyboardFocus || surf == pointerFocus)) {
+                    focusedSurf = surf;
+                    break;
+                }
+            }
+
+            // Fall back to first surface if no focused surface found
+            if (!focusedSurf)
+                focusedSurf = oldGrab->m_surfs.front().lock();
+
+            if (focusedSurf) {
+                auto hlSurface = CWLSurface::fromResource(focusedSurf);
+                if (hlSurface) {
+                    auto popup = hlSurface->getPopup();
+                    if (popup) {
+                        auto t1Owner = popup->getT1Owner();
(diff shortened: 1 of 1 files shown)


src/managers/SeatManager.cpp
```diff
@@ -616,8 +616,40 @@ void CSeatManager::setCurrentPrimarySelection(SP<IDataSource> source) {
 void CSeatManager::setGrab(SP<CSeatGrab> grab) {
     if (m_seatGrab) {
         auto oldGrab = m_seatGrab;
+
+        // Try to find the parent window from the grab
+        PHLWINDOW parentWindow;
+        if (oldGrab && oldGrab->m_surfs.size()) {
+            auto first = oldGrab->m_surfs.front().lock();
```
vaxerski: instead of front, maybe try to find the one that has focus at the moment of ungrabbing?

src/managers/SeatManager.cpp
```diff
@@ -616,8 +616,40 @@ void CSeatManager::setCurrentPrimarySelection(SP<IDataSource> source) {
...
+
+        static auto PFOLLOWMOUSE = CConfigValue<Hyprlang::INT>("input:follow_mouse");
+        if (*PFOLLOWMOUSE == 0 || *PFOLLOWMOUSE == 2 || *PFOLLOWMOUSE == 3) {
+            const auto PMONITOR = g_pCompositor->getMonitorFromCursor();
+
+            // If this was a popup grab, focus its parent window to maintain context
+            if (parentWindow && parentWindow->m_isMapped) {
+                g_pCompositor->focusWindow(parentWindow);
+                Debug::log(LOG, "[seatmgr] Refocused popup parent window {} (follow_mouse={})", parentWindow->m_title, *PFOLLOWMOUSE);
+            } else {
```
vaxerski: no {} around short if/else

src/managers/SeatManager.cpp
```diff
@@ -616,8 +616,55 @@ void CSeatManager::setCurrentPrimarySelection(SP<IDataSource> source) {
...
+
         m_seatGrab.reset();
-        g_pInputManager->refocus();
+
+        static auto PFOLLOWMOUSE = CConfigValue<Hyprlang::INT>("input:follow_mouse");
+        if (*PFOLLOWMOUSE == 0 || *PFOLLOWMOUSE == 2 || *PFOLLOWMOUSE == 3) {
+            const auto PMONITOR = g_pCompositor->getMonitorFromCursor();
+
+            // If this was a popup grab, focus its parent window to maintain context
+            if (parentWindow && parentWindow->m_isMapped) {
```
vaxerski: entire if can be shortened to just `validMapped(parentWindow)`

PR #12019, opened 2025-10-12
hyprctl: include color management presets and sdr information in the output

fiffeek:

Exposes color management preset type with sdr brightness/luminance in `hyprctl monitors`.

I recently open sourced [hyprdynamicmonitors](https://github.com/fiffeek/hyprdynamicmonitors) which is like autorandr for hyprland, with a TUI and a daemon to automatically apply monitor settings on different inputs/power stat etc; I wanted to work on [color management settings](https://github.com/fiffeek/hyprdynamicmonitors/issues/34) in said TUI but `hyprctl monitors` does not expose the applied preset (I believe bitdepth can be inferred from the current drm format).

I wanted the enum (de)serialization to be central instead of having it over in monitor parsing rules and in hyprctl itself so I moved it to its own namespace in `src/helpers/CMType.hpp`, let me know if there are other clearer options here, this was mainly based on an existing pattern from [src/protocols/types/ContentType.hpp](https://github.com/hyprwm/Hyprland/blob/main/src/protocols/types/ContentType.hpp).

I also added `-DTESTS=true` to the `debug` make target, seem that for nix it's added in the same context [here](https://github.com/hyprwm/Hyprland/blob/ed936430216e7aa5f6f53d22eff713f8e9ed69ac/nix/hyprtester.nix#L60) but for some reason is left out from the makefile which results in hyprtester to not be built so `make test` does not work (again, if there is a reason for that just lmk).

I added a basic test for this, the rest of the implementation seems straightforward.

I also did a quick test with nesting hyprland and it works as expected:
```
❯ h monitors -j
[{
    "id": 0,
    "name": "WAYLAND-1",
    "description": "",
    "make": "",
    "model": "",
    "serial": "",
    "width": 1756,
    "height": 1542,
    "physicalWidth": 0,
    "physicalHeight": 0,
    "refreshRate": 60.00000,
    "x": 0,
    "y": 0,
    "activeWorkspace": {
        "id": 1,
        "name": "1"
    },
    "specialWorkspace": {
        "id": 0,
        "name": ""
    },
    "reserved": [0, 0, 0, 0],
    "scale": 2.00,
    "transform": 0,
    "focused": true,
    "dpmsStatus": true,
    "vrr": false,
    "solitary": "0",
    "solitaryBlockedBy": ["WINDOWED","CANDIDATE"],
    "activelyTearing": false,
    "tearingBlockedBy": ["NOT_TORN","USER","SUPPORT","CANDIDATE"],
    "directScanoutTo": "0",
    "directScanoutBlockedBy": ["USER","CANDIDATE"],
    "disabled": false,
    "currentFormat": "XRGB8888",
    "mirrorOf": "none",
    "availableModes": [],
    "colorManagementPreset": "srgb",
    "sdrBrightness": 1.00,
    "sdrSaturation": 1.00,
    "sdrMinLuminance": 0.20,
    "sdrMaxLuminance": 80
}]
❯ h monitors
Monitor WAYLAND-1 (ID 0):
        1756x1542@60.00000 at 0x0
        description:
        make:
        model:
        physical size (mm): 0x0
        serial:
        active workspace: 1 (1)
        special workspace: 0 ()
        reserved: 0 0 0 0
        scale: 2.00
        transform: 0
        focused: yes
        dpmsStatus: 1
        vrr: false
        solitary: 0
        solitaryBlockedBy: windowed mode,missing candidate
        activelyTearing: false
        tearingBlockedBy: next frame is not torn,user settings,not supported by monitor,missing candidate
        directScanoutTo: 0
        directScanoutBlockedBy: user settings,missing candidate
        disabled: false
        currentFormat: XRGB8888
        mirrorOf: none
        availableModes:
        colorManagementPreset: srgb
        sdrBrightness: 1.00
        sdrSaturation: 1.00
        sdrMinLuminance: 0.20
        sdrMaxLuminance: 80
```

It's complete as long as maintainers agree with the enum extraction, I can rework if other options seem more viable.

### DIFF
--- a/src/helpers/CMType.cpp
+++ b/src/helpers/CMType.cpp
@@ -1,0 +1,23 @@
+#include "CMType.hpp"
+#include <optional>
+#include <string>
+#include <unordered_map>
+
+static std::unordered_map<std::string, NCMType::eCMType> const table = {{"auto", NCMType::CM_AUTO},   {"srgb", NCMType::CM_SRGB}, {"wide", NCMType::CM_WIDE},
+                                                                        {"edid", NCMType::CM_EDID},   {"hdr", NCMType::CM_HDR},   {"hdredid", NCMType::CM_HDR_EDID},
+                                                                        {"dcip3", NCMType::CM_DCIP3}, {"dp3", NCMType::CM_DP3},   {"adobe", NCMType::CM_ADOBE}};
+
+std::optional<NCMType::eCMType>                                NCMType::fromString(const std::string cmType) {
+    auto it = table.find(cmType);
+    if (it == table.end())
+        return std::nullopt;
+    return it->second;
+}
+
+std::string NCMType::toString(eCMType cmType) {
+    for (const auto& [key, value] : table) {
+        if (value == cmType)
+            return key;
+    }
+    return "";
+}
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ release:
 	cmake --build ./build --config Release --target all -j`nproc 2>/dev/null || getconf NPROCESSORS_CONF`
 
 debug:
-	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -S . -B ./build
+	cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Debug -DTESTS=true -DCMAKE_INSTALL_PREFIX:STRING=${PREFIX} -S . -B ./build
 	cmake --build ./build --config Debug --target all -j`nproc 2>/dev/null || getconf NPROCESSORS_CONF`
 
 nopch:
--- a/hyprtester/src/tests/main/colors.cpp
+++ b/hyprtester/src/tests/main/colors.cpp
@@ -1,0 +1,27 @@
+#include "tests.hpp"
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
+#include "../shared.hpp"
+
+static int  ret = 0;
+
+static bool test() {
+    NLog::log("{}Testing hyprctl monitors", Colors::GREEN);
+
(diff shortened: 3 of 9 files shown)


src/helpers/CMType.cpp
```diff
@@ -0,0 +1,24 @@
+#include "CMType.hpp"
+#include <string>
+#include <unordered_map>
+#include <format>
+
+namespace NCMType {
```
vaxerski: dont do this pls just prefix fns with NCMType::

fiffeek: gotcha, haven't touched cpp in ages thus bits rusty on best practices; thanks for your time

src/helpers/CMType.cpp
```diff
@@ -0,0 +1,23 @@
...
+        throw std::invalid_argument(std::format("Unknown cm type {}", cmType));
+    return it->second;
+}
+
+std::string NCMType::toString(eCMType cmType) {
+    for (const auto& [key, value] : table) {
+        if (value == cmType)
+            return key;
+    }
+    throw std::invalid_argument(std::format("Unknown cm type enum value {}", static_cast<int>(cmType)));
```
vaxerski: we don't use exceptions unless necessary cuz stl. Just return an empty string or use optional

Also, no `static_cast`, use `sc<>` from hyprutils

fiffeek: Got rid of the errors in both functions, it made sense to me to return optional while parsing and just use an empty string for `toString` to simplify the usage. The cast is a noop since err is gone now; thanks!

vaxerski (approved):

fiffeek:
seems I was missing `#include <cstdint>` which worked on _some_ builds but not always (sigh), now after adding it (I also swapped the stupid `optional` import from `toml...`) seems `make nopch` and `nix --extra-experimental-features "nix-command flakes" build` works

@vaxerski , if you could appr again,  sry for the trouble

Issue #11869, opened 2025-09-29
Add a toggle for disabling AA on cursor:zoom_factor


vaxerski:
### Discussion link

https://github.com/hyprwm/Hyprland/discussions/11866

## Description

Fixed in 7ce451d20c33025d4aaaf55444150c4a875b83ea, cursor:zoom_factor now uses no AA (nearest). This should be toggleable.

PR #12025, opened 2025-10-13
renderer: add cursor:zoom_disable_aa for controlling AA on zoom

vaxerski:
default stays pixelated, but can be turned off to be blurry.

Fixes https://github.com/hyprwm/Hyprland/issues/11869

cc @aphelei can you test this?


aphelei:
It works, thank you.
But I think that it should default to true. Yes it's a bit blurry but it doesn't flicker (it looks very bad in my opinion)


https://github.com/user-attachments/assets/08d6fd15-aa8b-4e77-99ac-2cfc04a8e1ac

vaxerski:
k

PR #12027, opened 2025-10-13
renderer: fix background bleeding in blur on special workspaces

ArkadyBuryakov:

Fix background bleeding in blur around special workspace windows.

Here is discussion with that bug description.
https://github.com/hyprwm/Hyprland/discussions/11761

No

Ready for merging

### DIFF
--- a/src/render/pass/Pass.cpp
+++ b/src/render/pass/Pass.cpp
@@ -71,11 +71,6 @@ void CRenderPass::simplify() {
             if (WILLBLUR) {
                 CRegion liveBlurRegion;
                 for (auto& el2 : m_passElements) {
-                    // if we reach self, no problem, we can break.
-                    // if the blur is above us, we don't care, it will work fine.
-                    if (el2 == el)
-                        break;
-
                     if (!el2->element->needsLiveBlur())
                         continue;
 


ArkadyBuryakov:
Oops, decided to make some refactor work and bug returned. Need a few minutes to get it right.

ArkadyBuryakov:
Done.

vaxerski (commented):
completely wrong fix. This needs to be done when calculating occlusion (and already is done, I assume you hit some edge case). This just destroys performance for no reason.

ArkadyBuryakov:
Okay, I noticed that live blur works correctly and cached blur has bleeding background around edges on special workspaces. 

At least that's my theory because:
- if I have video running in background blur works ok
- if window in background isn't constantly refreshing, I get borked blur around edges that also flickers.

I'll try to find how to fix that without passing extra edges.

ArkadyBuryakov:
Okay, problem reproduces in cases when element that needs blur is positioned between two opaque elements. 

I removed this part and everything works fine now.
image

Feels like not a great solution. Gonna read a little more into code.

UPD: Haven't found better solution. @vaxerski , hope it will do.

vaxerski (commented):
uhhh, I'm not sure... I think the condition you removed is correct

ArkadyBuryakov:
> uhhh, I'm not sure... I think the condition you removed is correct

I tried to grasp the whole masking logic and couldn't grok it. That's my best fix so far, maybe there could be a better optimized fix, I just can't find it.

What I know is that flickering around windows on special spaces is still not fixed in main and this patch worked perfectly for me all these months.

PR #12038, opened 2025-10-14
Nix: build hyprtester along with hyprland

fufexan:

- CMake: allow building hyprtester without running tests
- Nix: build hyprtester along with hyprland

Another precursor to #12001.

No.

Ready if CI passes.


fufexan:
@vaxerski checking the logs of https://github.com/hyprwm/Hyprland/actions/runs/18506758523/attempts/1 I don't think the failures are due to this PR. Probably timing issues or something. Just to be sure, can you double check?

> Failed: output should contain "q" 1 times, but doesn't. Source: hyprtester/src/tests/main/keybinds.cpp@287. String is:
  qq
  Failed: fn(), expected true, got false. Source: hyprtester/src/main.cpp@234.

vaxerski:
didnt I remove that one test in main?

fufexan:
Maybe, I'll rebase.

CMakeLists.txt
```diff
@@ -460,18 +460,25 @@ install(
   PATTERN "*.inc")
 
 if(TESTS)
-  message(STATUS "building tests is enabled TESTS")
-
-  enable_testing()
-  add_custom_target(tests)
+  message(STATUS "Building tests is enabled TESTS")
 
   add_subdirectory(hyprtester)
```
fufexan, line 525: Could probably put this behind an `if(BUILD_HYPRTESTER OR BUILD_TESTING)` instead. That way we use CMake's testing var and BUILD_HYPRTESTER could be used by nix only. Means we no longer need to have nested ifs.

@vaxerski thoughts?

vaxerski: ok

fufexan:
Don't know why those tests still show up. They seem different but related to the one you removed.

ItsOhen:
Funnily enough, doesn't break CI in fork.

vaxerski:
> Don't know why those tests still show up. They seem different but related to the one you removed.

Remove as many as needed for ci to pass I guess. They are a bit time-sensitive, which fucks up with the slow-ass vm

PR #12046, opened 2025-10-15
[XWayland] menu bar dropdown hover isn't opening other menu items

enzi:
Jetbrains Rider under XWayland and probably other apps have a problem that any dropdown created from a the menu toolbar needs to be clicked with `follow_mouse = 0` the hover state when going through other toolbar menu items is not updated and needs additional clicks.
Also hovering over dropdown menu entries didn't update the hover state.


src/managers/input/InputManager.cpp
```diff
@@ -559,7 +559,12 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus, bool mouse, st
             if (pFoundWindow == g_pCompositor->m_lastWindow)
                 g_pSeatManager->setPointerFocus(foundSurface, surfaceLocal);
 
-            if (FOLLOWMOUSE != 0 || pFoundWindow == g_pCompositor->m_lastWindow)
+            // X11 windows need pointer focus for proper menu/popup interaction even with follow_mouse=0
+            // This allows menu cascading and other pointer-based interactions to work correctly
+            const bool shouldGiveX11PointerFocus = pFoundWindow && pFoundWindow->m_isX11 && g_pCompositor->m_lastWindow.lock() && g_pCompositor->m_lastWindow->m_isX11 &&
```
vaxerski: .lock() redundant, wp has a bool operator

src/managers/input/InputManager.cpp
```diff
@@ -559,7 +559,12 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus, bool mouse, st
...
+                                                   pFoundWindow->getPID() == g_pCompositor->m_lastWindow->getPID();
```
vaxerski: mmm no, this could include child regular windows. Add a helper fn to check if it's an X11 popup. I don't remember the atom name, get it with xprop or debugging.

enzi: how? current/last window are both checked for x11.
i would need to dig into how to determine if it's an x11 popup. i agree that might be cleaner though and better presents the intent.

enzi:
i've added a helper function to determine if it's a popup/combo, etc...
better now?

i tried `shouldGiveX11PointerFocus` with just the popup check but that was not sufficient to keep the functionality up. also needs the check `&& g_pCompositor->m_lastWindow && g_pCompositor->m_lastWindow->m_isX11`

enzi:
putting it back to draft as the change broke the actual fix -.-

enzi:
omg i found the actual reason this broke! `float_switch_override_focus = 0`
my fix just duct taped the needed focus back. now i don't think this override should break a dropdown though. any ideas?

enzi:
are you okay with doing a refactor of this?
InputManager.cpp L540 to L587
i've been in this block before trying to fix something and it's just hurting my brain. :D

src/managers/input/InputManager.cpp
vaxerski: uhh if you look at shouldbefloated in xwaylandmgr you can see atom checking. It's stuff like `NET_TYPE_BLAH` I assume you only want to target some of those, you can make a helper fn in CWindow to check that

vaxerski:
I know it hurts brain but there have been so many edge cases there that in general I'd rather leave it be

enzi:
I didn't forget about this PR, just kinda busy.
Anyway, I found out the reason again why I used `float_switch_override_focus` in the first place. Any popup dialogue (floating) in Unity and Rider auto close when the mouse isn't in that window. I can work around it in Rider since you can pin windows there but for unity it's not possible and it makes any kind of popups often completely unusable because you can't physically be in 2 places at the same time with the mouse to catch the window that opens. :D

So yeah, this fix IS needed in some form.
This whole code passage not only hurts the brain, it's buggy with odd or unwanted side effects, super hard to maintain.
It's like a stack of workarounds, at one point they will interfere with each other.

vaxerski:
> This whole code passage not only hurts the brain, it's buggy with odd or unwanted side effects, super hard to maintain.
It's like a stack of workarounds, at one point they will interfere with each other.

Unfortunately people want very different behaviors and that creates a messy codebase there. It's a stack of workarounds because we need to accomodate them all.

Is this MR ready for review or?

umbrageodotus:
@enzi sorry if I'm being annoying but is this ready for review?

enzi:
@umbrageodotus no, not annoying at all :D thanks for reminding.
i just need to update to resolve the conflicts.

umbrageodotus:
ty for the friendly reply :>
if you need a mock code review before vax ends up doing the full review, lmk!

enzi:
i'm testing latest `main` and it looks like my fix isn't needed anymore. will test a bit more. otherwise i'll just close the PR

> Also hovering over dropdown menu entries didn't update the hover state.

ok, quickly ran into this. so keeping it open. should be ready now for review after i gave it a last test

enzi:
that one is ready now. my test case here was:
- a Jetbrains IDE like Rider or Clion
- follow_mouse 0
- open a nested menu (Code -> Folding -> Expand to level)

without the fix: Folding isn't selected anymore on hover, so nothing expands further
with the fix: Folding is selected and correctly expands the other nested popup

something had changed, probably within hyprland itself or with jetbrains and i had to add `_NET_WM_WINDOW_TYPE_DIALOG` for this to work.

i honestly don't know if this is a great fix. previously the whole dropdown focus was broken. vaxry or someone else fixed that but hovering over nested dropdowns was still broken.

maybe you can make a more mindful fix here?

enzi:
@umbrageodotus pinging you to follow up on the offer to make a mock code review

umbrageodotus:
ack, I'll get it done assp

PR #12051, opened 2025-10-15
[XWayland] Jetbrains IDE: fixes tab reordering and detaching

enzi:
Jetbrains IDEs under XWayland have 2 severe problems:
- tab reordering:
this was completely broken. any movement in the y axis dropped the Drag&Drop

- detaching a tab to a new window:
with the reordering not working, detaching a tab to a new window was also not working

this small line fixes both of these issues.


src/managers/input/InputManager.cpp
```diff
@@ -156,6 +156,10 @@ void CInputManager::sendMotionEventsToFocused() {
     if (!g_pCompositor->m_lastFocus || isConstrained())
         return;
 
+    // Don't steal pointer focus when buttons are held (e.g., during drags)
```
vaxerski: too low in the chain: unintuitive behavior and is bug-prone. Where does this call come from? Can we filter it a bit higher? (preferably in X code)

enzi: i can try to track this further up. if you have any indicators for the xwayland side please tell me.
btw i'm daily driving the 2 PRs i made and i have not found any issues.

enzi:
how does it look now? outside of Windows.cpp i really don't find any other good path. everything else gets too low level.
i had to introduce an inner if to not trigger the `else` path

src/events/Windows.cpp
```diff
@@ -618,9 +618,13 @@ void Events::listener_mapWindow(void* owner, void* data) {
     if (!PWINDOW->m_windowData.noFocus.valueOrDefault() && !PWINDOW->m_noInitialFocus &&
         (!PWINDOW->isX11OverrideRedirect() || (PWINDOW->m_isX11 && PWINDOW->m_xwaylandSurface->wantsFocus())) && !workspaceSilent && (!PFORCEFOCUS || PFORCEFOCUS == PWINDOW) &&
         !g_pInputManager->isConstrained()) {
-        g_pCompositor->focusWindow(PWINDOW);
-        PWINDOW->m_activeInactiveAlpha->setValueAndWarp(*PACTIVEALPHA);
-        PWINDOW->m_dimPercent->setValueAndWarp(PWINDOW->m_windowData.noDim.valueOrDefault() ? 0.f : *PDIMSTRENGTH);
+
+        // Don't steal pointer focus with X11 when buttons are held (e.g., during drags)
+        if (!(PWINDOW->m_isX11 && g_pInputManager->hasHeldButtons())) {
```
vaxerski: demorgan

vaxerski (commented):
yes, this looks way better

Also consider adding a helper fn to check if the window is of a specific type. I don't think we want that behavior for all windows, do we?

umbrageodotus:
@vaxerski ?

enzi:
requesting a review. the logic has been flipped.
all merged to latest `main` branch. should be good to merge now?

vaxerski (approved):

Issue #11386, opened 2025-08-08
Implement fifo v1


vaxerski:
### Discussion link
N/A

## Description

Implement https://wayland.app/protocols/fifo-v1.

Would be useful for games. KDE and Gnome have it.

Justsnoopy30:
May also want to pair with the addition of commit-timing-v1 when this gets added (don't see an open thing for that) as it sounds like both are needed for best behavior according to https://github.com/libsdl-org/SDL/commit/ab916a959e096a3af9502b50a622eda0e8f83253

fxzzi:
Both fifo-v1 and commit-timing-v1 are also used, if available, with nvidia's egl-wayland2: https://github.com/NVIDIA/egl-wayland2?tab=readme-ov-file#compositor-requirements

this could improve the experience for nvidia users i guess.

PR #12052, opened 2025-10-15
protocols: add Fifo-v1 and commit-timing-v1

gulafaran:
introduce fifo-v1 and commit-timing-v1

### TODO

- [x] add commit-timing-v1
- [x] properly test fifo-v1
- [x] properly test commit-timing-v1
- [x] ensure nothing else broke because of CSurfaceStateQueue

fixes: https://github.com/hyprwm/Hyprland/issues/11386

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1795,6 +1795,7 @@ bool CMonitor::attemptDirectScanout() {
 
     auto PBUFFER = PSURFACE->m_current.buffer.m_buffer;
 
+    // #TODO this entire bit needs figuring out, vrr goes down the drain without it
     if (PBUFFER == m_output->state->state().buffer) {
         PSURFACE->presentFeedback(Time::steadyNow(), m_self.lock());
 
@@ -1812,6 +1813,10 @@ bool CMonitor::attemptDirectScanout() {
 
             m_scanoutNeedsCursorUpdate = false;
         }
+
+        //#TODO this entire bit is bootleg deluxe, above bit is to not make vrr go down the drain, returning early here means fifo gets forever locked.
+        if (PSURFACE->m_fifo)
+            PSURFACE->m_stateQueue.unlockFirst(LOCK_REASON_FIFO);
 
         return true;
     }


fxzzi:
playing video in mpv keeps freezing, continuously moving the mouse avoids it

gulafaran:
> playing video in mpv keeps freezing, continuously moving the mouse avoids it

Bleh, so it wasnt buffers being held up. made it work here. Hrmhm. Thanks

ikalco:
this generally LGTM
I don't have time to test but here are some things that might break
* minecraft with vsync vs unlimited fps and fullscreen (ds) vs windowed
* csgo windowed + ds
* vkcube windowed + ds
* screensharing with obs (screencopy + toplevel export)

gulafaran:
> this generally LGTM I don't have time to test but here are some things that might break
> 
>     * minecraft with vsync vs unlimited fps and fullscreen (ds) vs windowed
> 
>     * csgo windowed + ds
> 
>     * vkcube windowed + ds
> 
>     * screensharing with obs (screencopy + toplevel export)

i think tearing might be affected aswell, `The condition is cleared immediately after the following latching deadline for non-tearing presentation.`


however i still havent figured out why mpv is being annoyed. its almost like there is an upper render scheduling/damage issue. vfr = 0 or damage_bling = 1 or simply scheduling a monitor frame at the end of each `renderMonitor` which those two settings kinda does. makes it not drop frames nor freeze

gulafaran:
@fxzzi i assume something like this makes mpv work for you too?

g3r:
> @fxzzi i assume something like this makes mpv work for you too?

I'm also having this issue with mpv, I tested the latest commit and if I launch a video from the terminal it works fine. But if I open the video from dolphin without background applications causing damage (pressing Enter or double-click without moving the mouse) The mpv window doesn't appear until I move the mouse, after that it seems to works fine.
It seems that if there is something in the background causing damage, the window appears

gulafaran:
> > @fxzzi i assume something like this makes mpv work for you too?
> 
> I'm also having this issue with mpv, I tested the latest commit and if I launch a video from the terminal it works fine. But if I open the video from dolphin without background applications causing damage (pressing Enter or double-click without moving the mouse) The mpv window doesn't appear until I move the mouse, after that it seems to works fine. It seems that if there is something in the background causing damage, the window appears

yeah probably didnt have entered an output, yet. and poof that scheduleframe never triggers. atleast we are rounding in where the issue is just gotta figure out something proper, thanks.

gulafaran:
> > @fxzzi i assume something like this makes mpv work for you too?
> 
> I'm also having this issue with mpv, I tested the latest commit and if I launch a video from the terminal it works fine. But if I open the video from dolphin without background applications causing damage (pressing Enter or double-click without moving the mouse) The mpv window doesn't appear until I move the mouse, after that it seems to works fine. It seems that if there is something in the background causing damage, the window appears

how about now?

g3r:
> how about now?

Tested mpv and vkcube. All seems to be fine now. Thanks.

gulafaran:
added 720a468 since we already are touching all sorts of surface state commits, lets improve if further if possible. still needs some VRR and random other usage testing so it doesnt regress.

gulafaran:
> added [720a468](https://github.com/hyprwm/Hyprland/commit/720a468bc04ce1dddeb8820c4f37f609b23df48a) since we already are touching all sorts of surface state commits, lets improve if further if possible. still needs some VRR and random other usage testing so it doesnt regress.

i think it might be problematic if a later buffer commit becomes ready before the first. so it essentially skips the first and just merges into the last one. but i havent yet seen that occur here

gulafaran:
> added [720a468](https://github.com/hyprwm/Hyprland/commit/720a468bc04ce1dddeb8820c4f37f609b23df48a) since we already are touching all sorts of surface state commits, lets improve if further if possible. still needs some VRR and random other usage testing so it doesnt regress.

i think it might be problematic if a later buffer commit becomes ready before the first. so it essentially skips the first and just merges into the last one. but i havent yet seen that occur here

> > added [720a468](https://github.com/hyprwm/Hyprland/commit/720a468bc04ce1dddeb8820c4f37f609b23df48a) since we already are touching all sorts of surface state commits, lets improve if further if possible. still needs some VRR and random other usage testing so it doesnt regress.
> 
> i think it might be problematic if a later buffer commit becomes ready before the first. so it essentially skips the first and just merges into the last one. but i havent yet seen that occur here

nvm 0a3d39d thinking about it. certain changes are relative to the buffer attached, cant go beyond it
and apply those onto the next buffer.

fxzzi:
SDL2 and SDL3 are still causing Hyprland to crash when in native wayland.

osu!lazer with `OSU_SDL3=1` crashes - because fifo and commit timing are available, SDL3 picks wayland by default, then proceeds to crash
osu!lazer with `SDL_VIDEODRIVER=wayland` - without above env var osu! runs with SDL2, which never sets wayland by default. Setting to wayland makes Hyprland crash
CS2 with `SDL_VIDEO_DRIVER=wayland` - cs2 launch script sets to x11 by default, so env var is needed even though its SDL3. Causes crash
CS:Source with `STEAM_COMPAT_RUNTIME_SDL2=sdl2-compat SDL_VIDEO_DRIVER=wayland` also crashes. env vars above make it use SDL3, i assume it'll crash with SDL2 as well.

[hyprlandCrashReport3626.txt](https://github.com/user-attachments/files/23027186/hyprlandCrashReport3626.txt)

(mpv issues from before are fixed)

AntoSkate:
After some quick testing i came across 2 issues:
- Resizing fifo apps is broken on both tiled and floating modes (works fine on master)

This PR

https://github.com/user-attachments/assets/9ece24f4-c85d-402d-8efd-b255790b98fd

Master

https://github.com/user-attachments/assets/cb1593ec-ac48-41b9-bf1b-c34d6bf4f4a8

- CS2 crashes the compositor soon after launch (also doesn't happen on master)

https://github.com/user-attachments/assets/25d85df9-4a5e-4be7-8194-2fb2127618f6

Also i'm on nvidia if it's relevant

gulafaran:
> SDL2 and SDL3 are still causing Hyprland to crash when in native wayland.
> 
> osu!lazer with `OSU_SDL3=1` crashes - because fifo and commit timing are available, SDL3 picks wayland by default, then proceeds to crash osu!lazer with `SDL_VIDEODRIVER=wayland` - without above env var osu! runs with SDL2, which never sets wayland by default. Setting to wayland makes Hyprland crash CS2 with `SDL_VIDEO_DRIVER=wayland` - cs2 launch script sets to x11 by default, so env var is needed even though its SDL3. Causes crash CS:Source with `STEAM_COMPAT_RUNTIME_SDL2=sdl2-compat SDL_VIDEO_DRIVER=wayland` also crashes. env vars above make it use SDL3, i assume it'll crash with SDL2 as well.
> 
> [hyprlandCrashReport3626.txt](https://github.com/user-attachments/files/23027186/hyprlandCrashReport3626.txt)
> 
> (mpv issues from before are fixed)

the heck...


2025-10-21-185927_hyprshot

fxzzi:
👀 

weird, person above also has same crash

gulafaran:
> After some quick testing i came across 2 issues:
> 
>     * Resizing fifo apps is broken on both tiled and floating modes (works fine on master)
> 
> 
> This PR
> fifo_pr.mp4
> 
> Master
> master.mp4
> 
>     * CS2 crashes the compositor soon after launch (also doesn't happen on master)
> 
> 
> cs2.mp4
> 
> Also i'm on nvidia if it's relevant


> 👀
> 
> weird, person above also has same crash

yeah, atleast i got proof i did test things :D , il dig hrmlm. this is why i wanted more people testing because weird things happends all the time when it comes to buffers lol. fix one thing break 3 other. thanks.

gulafaran:
ok that should handle the crashes, kwin apparently also blocks fifo when actively tearing. i guess its up to debate what todo here. block fifo or block tearing?. and il see if i cant make it cleaner later.

il see what that resizing vkcube is about, might be a day away if i dont find it shortly. working 2 long days.

AntoSkate:
> ok that should handle the crashes, kwin apparently also blocks fifo when actively tearing. i guess its up to debate what todo here. block fifo or block tearing?. and il see if i cant make it cleaner later.
> 
> il see what that resizing vkcube is about, might be a day away if i dont find it shortly. working 2 long days.

can confirm it doesn't crash anymore 🎉
btw it's not only vkcube but also mpv (so probably every app that uses fifo)

fxzzi:
mpv resizes fine for me

AntoSkate:
Maybe it's a configuration thing? Here's mine
`gpu-context=waylandvk
gpu-api=vulkan
vo=gpu-next
hwdec=vulkan,nvdec`

https://github.com/user-attachments/assets/17509956-cc22-4cc3-a1fc-956081a8b11a

Gliczy:
`mpv` resize fine for me as well. I also tried with your flags (minus nvdec since I'm on AMD) and it does work.

gulafaran:
yeah something fishy is going on, not found what yet. its like the surface is killed and recreated when resizing when checking with WAYLAND_DEBUG=1. and it actually happends on -git too only thing is its going so fast since it essentially runs as fast as it can at 14k fps its not as noticeable because we arent waiting on fifo barriers. so im not entirely sure this PR broke it, more of exposed it, atleast regarding to vkcube. struggling a bit to reproduce it on mpv.

fxzzi:
done more testing, everything looking good on my end the past few days. Thanks for your work!

edit: rebase on master by any chance?

gulafaran:
> done more testing, everything looking good on my end the past few days. Thanks for your work!
> 
> edit: rebase on master by any chance?

thanks, rebased. yeah il spend some time tomorrow figuring out what can be done with resizing. maybe clean a few things up and i would say its somewhat good to go unless something else major pops up.

AntoSkate:
Found another minor issue: cs2 freezes on launch (or when pausing game) if direct scanout and vsync are on but it starts rendering again when switching workspaces or ofc when recording or with fullscreen off

https://github.com/user-attachments/assets/84966560-cee4-4e99-946e-3370f54fbae4

Apart from this i played some games with ds off and it works good

gulafaran:
> Found another minor issue: cs2 freezes on launch (or when pausing game) if direct scanout and vsync are on but it starts rendering again when switching workspaces or ofc when recording or with fullscreen off
> cs2_ds.mp4
> 
> Apart from this i played some games with ds off and it works good

cant reproduce this either :( but does this make it happier?

```
diff --git a/src/helpers/Monitor.cpp b/src/helpers/Monitor.cpp
index bda11471..19eace00 100644
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1856,6 +1856,7 @@ bool CMonitor::attemptDirectScanout() {
     if (m_lastScanout.expired()) {
         m_lastScanout = PCANDIDATE;
         Debug::log(LOG, "Entered a direct scanout to {:x}: \"{}\"", rc<uintptr_t>(PCANDIDATE.get()), PCANDIDATE->m_title);
+        g_pHyprRenderer->m_usedAsyncBuffers.clear();
     }
 
     m_scanoutNeedsCursorUpdate = false;
```

AntoSkate:
> does this make it happier?

no it still happens
but interestingly on my amd igpu only the boot animation and some menus freeze (with and without the patch) 🤔

Zebra2711:
> > does this make it happier?
> 
> no it still happens but interestingly on my amd igpu only the boot animation and some menus freeze (with and without the patch) 🤔

did u set `render:new_render_scheduling` in the hyprland config file?

gulafaran:
> > > does this make it happier?
> > 
> > 
> > no it still happens but interestingly on my amd igpu only the boot animation and some menus freeze (with and without the patch) 🤔
> 
> did u set `render:new_render_scheduling` in the hyprland config file?

oh thats right. yeah its all sort of lacking currently. really need to get around and fix it up. its both releasing and reusing dropped buffers and sending presentation at wrong times, and a few other wonkyness. crossing my fingers here. because then its just vkcube to figure out if its a vbkcube weirdness or hypr. :D i still cant reproduce neither mpv or this cs freezing thing

gulafaran:
> > > > does this make it happier?
> > > 
> > > 
> > > no it still happens but interestingly on my amd igpu only the boot animation and some menus freeze (with and without the patch) 🤔
> > 
> > 
> > did u set `render:new_render_scheduling` in the hyprland config file?
> 
> oh thats right. yeah its all sort of lacking currently. really need to get around and fix it up. its both releasing and reusing dropped buffers and sending presentation at wrong times, and a few other wonkyness. crossing my fingers here. because then its just vkcube to figure out if its a vbkcube weirdness or hypr. :D i still cant reproduce neither mpv or this cs freezing thing

yeah `vkcube --wsi wayland --present_mode 2`  on kwin is equally bad, its broken by itself. so as long as we figure out mpv/cs2 thing i would say its done

AntoSkate:
> did u set `render:new_render_scheduling` in the hyprland config file?

i've had it set to false since it was introduced

gulafaran:
> > did u set `render:new_render_scheduling` in the hyprland config file?
> 
> i've had it set to false since it was introduced

:( oh well, gotta figure out a way to reproduce it with cs2/mpv so i can breakpoint debug it then.

gulafaran:
> > > did u set `render:new_render_scheduling` in the hyprland config file?
> > 
> > 
> > i've had it set to false since it was introduced
> 
> :( oh well, gotta figure out a way to reproduce it with cs2/mpv so i can breakpoint debug it then.

good news, mpv is reproducible on intel. not nvidia.

gulafaran:
> > did u set `render:new_render_scheduling` in the hyprland config file?
> 
> i've had it set to false since it was introduced

ok how about now?

gulafaran:
[b188e1b](https://github.com/hyprwm/Hyprland/pull/12052/commits/b188e1b6d6ada307767570c85e7bcc8c5e50ab0e) should atleast solve resizing issues, cs2 im unsure since i cant reproduce it

AntoSkate:
> [b188e1b](https://github.com/hyprwm/Hyprland/pull/12052/commits/b188e1b6d6ada307767570c85e7bcc8c5e50ab0e) should atleast solve resizing issues, cs2 im unsure since i cant reproduce it

it solved both vkcube and mpv on my end :O
but cs2 still has the same issue :/

gulafaran:
> > [b188e1b](https://github.com/hyprwm/Hyprland/pull/12052/commits/b188e1b6d6ada307767570c85e7bcc8c5e50ab0e) should atleast solve resizing issues, cs2 im unsure since i cant reproduce it
> 
> it solved both vkcube and mpv on my end :O but cs2 still has the same issue :/

do you have vrr on and you sure this doesnt occur on -git without this PR?

src/protocols/types/SurfaceState.hpp
```diff
@@ -8,6 +8,31 @@ class CTexture;
 class CDRMSyncPointState;
 class CWLCallbackResource;
 
+enum class LockReason : uint8_t {
+    None  = 0,
```
vaxerski: no enum class, just enum

src/protocols/types/SurfaceState.hpp
```diff
@@ -8,6 +8,31 @@ class CTexture;
...
+    Fence = 1 << 0,
+    Fifo  = 1 << 1,
```
vaxerski: bad enum names. LOCK_REASON_FIFO, etc.

src/protocols/types/SurfaceState.hpp
```diff
@@ -8,6 +8,31 @@ class CTexture;
 class CDRMSyncPointState;
 class CWLCallbackResource;
 
+enum class LockReason : uint8_t {
```
vaxerski: eLockReason

src/protocols/types/SurfaceState.hpp
gulafaran: changed.

src/protocols/types/SurfaceState.hpp
gulafaran: changed.

src/protocols/types/SurfaceState.hpp
gulafaran: changed.

vaxerski:
@AntoSkate what about that cs2 issue?

AntoSkate:
> do you have vrr on and you sure this doesnt occur on -git without this PR?

i have vrr off and i retested it doesn't happen on master
sorry for late response btw i was away this weekend

AntoSkate:
also tested this pr with vrr and without mangohud for science and it still freezes

gulafaran:
yeah gonna need some way to reproduce it to figure out why, tried nvidia, intel, dual monitors. single monitor. vrr, tearing, you name it. still doesnt occur here

AntoSkate:
~~I just tested on a friends pc with my config and everything and i couldn't reproduce it either 💀~~
~~i really don't know at this point everything else has been resolved~~

AntoSkate:
i forgot to set game to native wayland on his pc. can riproduce now on full amd system💀

gulafaran:
> also tested this pr with vrr and without mangohud for science and it still freezes

i pulled my hair until i got a bald spot for the past 12 hours. it might have been there already who knows. I CAN REPRODUCE. for some reason i never figured the ingame vsync option was the important bit to hit it. now im making an appointment to the barber and while waiting im looking into this freeze.

AntoSkate:
> i pulled my hair until i got a bald spot for the past 12 hours. it might have been there already who knows. I CAN REPRODUCE. for some reason i never figured the ingame vsync option was the important bit to hit it. now im making an appointment to the barber and while waiting im looking into this freeze.

Dang I might have to send some to cover for barber expenses 😂

UjinT34:
Freezes might be caused by modeset or other vrr/ds/hdr transitions. Cyberpunk freezes for a couple of seconds when it starts (audio keeps playing).

VRR seems to have less jitter now. Earlier it could jump all the way between min and max rate while keeping the same average, now it's only +-3hz or so.

gulafaran:
> Freezes might be caused by modeset or other vrr/ds/hdr transitions. Cyberpunk freezes for a couple of seconds when it starts (audio keeps playing).
> 
> VRR seems to have less jitter now. Earlier it could jump all the way between min and max rate while keeping the same average, now it's only +-3hz or so.

nah its fifo. the nuke debug it alll patch makes it not freeze. now just to figure out why.

```
diff --git a/src/protocols/CommitTiming.cpp b/src/protocols/CommitTiming.cpp
index 2fcd8e9c..d2d71a04 100644
--- a/src/protocols/CommitTiming.cpp
+++ b/src/protocols/CommitTiming.cpp
@@ -43,7 +43,8 @@ CCommitTimerResource::CCommitTimerResource(UP<CWpCommitTimerV1>&& resource_, SP<
         if (!m_pendingTimeout.has_value())
             return;
 
-        m_surface->m_stateQueue.lock(state, LOCK_REASON_TIMER);
+        //m_surface->m_stateQueue.lock(state, LOCK_REASON_TIMER);
+        LOGM(ERR, "locking timer");
 
         if (!m_timerPresent) {
             m_timerPresent = true;
@@ -53,7 +54,8 @@ CCommitTimerResource::CCommitTimerResource(UP<CWpCommitTimerV1>&& resource_, SP<
                     if (!m_surface)
                         return;
 
-                    m_surface->m_stateQueue.unlockFirst(LOCK_REASON_TIMER);
+                    //m_surface->m_stateQueue.unlockFirst(LOCK_REASON_TIMER);
+                    LOGM(ERR, "unlocking timer");
                 },
                 nullptr);
         } else
diff --git a/src/protocols/Fifo.cpp b/src/protocols/Fifo.cpp
index 63ce8579..dcfe2d27 100644
--- a/src/protocols/Fifo.cpp
+++ b/src/protocols/Fifo.cpp
@@ -66,8 +66,11 @@ CFifoResource::CFifoResource(UP<CWpFifoV1>&& resource_, SP<CWLSurfaceResource> s
         }
 
         // only lock once its mapped.
-        if (m_surface->m_mapped)
-            m_surface->m_stateQueue.lock(state, LOCK_REASON_FIFO);
+        if (m_surface->m_mapped) {
+            LOGM(ERR, "locking fifo");
+            //m_surface->m_stateQueue.lock(state, LOCK_REASON_FIFO);
+        }
+
 
         m_pending = {};
     });
@@ -82,7 +85,8 @@ bool CFifoResource::good() {
 }
 
 void CFifoResource::presented() {
-    m_surface->m_stateQueue.unlockFirst(LOCK_REASON_FIFO);
+    LOGM(ERR, "unlocking fifo");
+    //m_surface->m_stateQueue.unlockFirst(LOCK_REASON_FIFO);
 }
 
 CFifoManagerResource::CFifoManagerResource(UP<CWpFifoManagerV1>&& resource_) : m_resource(std::move(resource_)) {

```

fxzzi:
> VRR seems to have less jitter now. Earlier it could jump all the way between min and max rate while keeping the same average, now it's only +-3hz or so.

can confirm, vrr feels great now with this PR, and my OSD reports only fluctuations of about 1-2Hz all the way from 49Hz to 170Hz for me.

I am wondering whether VRR below the minimum VRR range is up to the compositor or the driver to handle? With my Nvidia GPU on linux and using `vrrtest`, when I reach my minimum of 48Hz, VRR pretty much breaks. In theory what it should do is double the frames sent to make the frame count effectively be within the VRR range, so for example 30fps should show 60 on the monitor OSD.

gulafaran:
ok so pushed a workaround, it however causes fifo to not work as intended in direct scanout. every once in a while it unlocks to early. but atleast things arent forever frozen. gonna see if i cant figure out something more proper later

cpiber:
This PR breaks building plugins with following error:

```
/usr/bin/g++-14 -shared -fPIC  main.cpp borderDeco.cpp BorderppPassElement.cpp -o borders-plus-plus.so -g `pkg-config --cflags pixman-1 libdrm hyprland pangocairo libinput libudev wayland-server xkbcommon` -std=c++2b -O2
In file included from /usr/include/hyprland/src/render/OpenGL.hpp:33,
                 from /usr/include/hyprland/src/render/Renderer.hpp:7,
                 from main.cpp:9:
/usr/include/hyprland/src/protocols/core/Compositor.hpp:16:10: fatal error: protocols/types/SurfaceStateQueue.hpp: No such file or directory
   16 | #include "protocols/types/SurfaceStateQueue.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

The include here should be relative, like `#include "../protocols/types/ContentType.hpp"` I think, same for the one in `src/protocols/types/SurfaceState.cpp` (?), only `src/protocols/FrogColorManagement.cpp` does not do the relative include, and I'd assume that breaks in plugins as well

Edit: Ignore the comment, didn't see #12228 which is now merged

PR #12054, opened 2025-10-16
Add: percentage to configStringToVec2

ItsOhen:
Someone in discord asked why % didn't work for minsize / maxsize.

And this only seems to be used for min/max size.

Probably not the greatest name for it if it were to be used for other things.

Also, sorry about the double PR. New lazygit not very good it seems.

Y?

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -693,26 +693,50 @@ std::expected<int64_t, std::string> configStringToInt(const std::string& VALUE) 
 
     return std::unexpected("parse error");
 }
-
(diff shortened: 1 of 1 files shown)


vaxerski:
Generally I am against, because we already had this with windowrules and we're repeating the mess. Why not just use an arithmetic parser at this point and expose variables to the user? E.g. `mon_x` `mon_y` `window_x` etc

ItsOhen:
Are the talking about the expression engine you suggested earlier?
Or just a simple parser? Maybe  to replace all the CConstVar thingies?

vaxerski:
expression engine

PR #12057, opened 2025-10-16
compositor: Block parent window interaction when modal dialog children window is open

AlvaroParker:

When a xdg dialog window is open, the parent window is now blocked from interaction until the children dialog window is closed. This follows the behavior of others popular compositors like kwin and mutter.

This is not specified by the [protocol](https://wayland.app/protocols/xdg-dialog-v1), however I do thing this is a good UX to have.

Ready.

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -142,6 +142,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_BOOL,
         .data        = SConfigOptionDescription::SBoolData{false},
     },
+    SConfigOptionDescription{
+        .value       = "general:modal_parent_blocking",
+        .description = "If true, parent windows of modals will not be interactive.",
+        .type        = CONFIG_OPTION_BOOL,
+        .data        = SConfigOptionDescription::SBoolData{true},
+    },
 
     /*
      * decoration:
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -881,15 +881,18 @@ bool CCompositor::monitorExists(PHLMONITOR pMonitor) {
 }
 
 PHLWINDOW CCompositor::vectorToWindowUnified(const Vector2D& pos, uint8_t properties, PHLWINDOW pIgnoreWindow) {
-    const auto  PMONITOR          = getMonitorFromVector(pos);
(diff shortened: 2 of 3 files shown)


vaxerski:
No, I think that's opposite to what the protocol is saying:
20251016_14h26m30s

AlvaroParker:
Oh, my interpretation was that the compositor may choose to filter events, but it's not required to (as stated in the second line). That's why the client must also implement the logic to filter events, because it's not guaranteed that the compositor will filter out the events.

Dregu:
I wouldn't mind an *option* to block focus to parent windows that are probably uninteractable anyway, it's kinda annoying with follow_mouse that modals immediately lose focus if the button that opened the modal is not under the modal. It is indeed better UX to highlight the only interactive part of the app and the protocol clearly allows this?

vaxerski:
> I wouldn't mind an option to block focus to parent windows that are probably uninteractable anyway, it's kinda annoying with follow_mouse that modals immediately lose focus if the button that opened the modal is not under the modal. It is indeed better UX to highlight the only interactive part of the app and the protocol clearly allows this?

Agreed. That should be added in this MR if we are going that direction.

AlvaroParker:
Great, I'll add a setting that controls whether the parent should be blocked (aka disabled) when a modal dialog children window is open

src/config/ConfigDescriptions.hpp
```diff
@@ -142,6 +142,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_BOOL,
         .data        = SConfigOptionDescription::SBoolData{false},
     },
+    SConfigOptionDescription{
+        .value       = "general:modal_parent_blocking",
+        .description = "If true, parent windows of modals will not be interactive.",
+        .type        = CONFIG_OPTION_BOOL,
+        .data        = SConfigOptionDescription::SBoolData{true},
```
AlvaroParker, line 149: Defaulting to true, let me know if you want the opposite instead which is how is currently behaving.

vaxerski: thats ok

vaxerski (dismissed):
:fire:

vaxerski:
wiki mr needed

AlvaroParker:
Done!

vaxerski:
c-f failed

AlvaroParker:
Done!

vaxerski (approved):

PR #12090, opened 2025-10-21
fix: onWrite spinning forever

ItsOhen:
Fix for #11411 

Make sure we drop things that no longer have a window. On both ends.
Also deletes the transfer with return 0 if we have completed INCR xfer so we don't sit and spin in onWrite.

Some more testing would be nice.
A refactor would be nicer.

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -1,3 +1,4 @@
+#include "debug/Log.hpp"
 #include <cstddef>
 #include <cstdint>
 #include <memory>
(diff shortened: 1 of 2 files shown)


vaxerski:
> A refactor would be nicer.

Maybe, but I don't want to touch X code with a ten-foot pole

jmanc3:
If the window no longer exist doesn't the function early terminate and your new code is never called?

```c++
void CXWM::handlePropertyNotify(xcb_property_notify_event_t* e) {
    const auto XSURF = windowForXID(e->window);

    if (!XSURF)
        return;
```

vaxerski:
fair point. should clean up on missing window too

ItsOhen:
Sounds like a good plan. :P

khaneliman:
Seems to have fixed my high CPU usage from using it as a patch atm.

vaxerski:
> Sounds like a good plan. :P

Are you planning on doing it?

ItsOhen:
Somewhat quick on the trigger there.
Still doesn't do incrementals well, which is why i moved the return.
Started to rewrite the selection to not be so shit. But getting late so will finish it tomorrow.

vaxerski:
oke

SohrabFarjami:
When will this be fixed?

vaxerski:
@ItsOhen ?

alper-han:
Some users may not notice it, but hundreds of gigabytes of log files are accumulating.
I hope the issue will be resolved soon.

vaxerski:
idk seems that op is mia

Issue #11818, opened 2025-09-24
Add an ignore submap flag to keybinds


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11797

## Description
A flag for ignoring submaps.

Brumus14:
Think this should be ok to be closed now

PR #12100, opened 2025-10-22
Added submap universal keybind flag

Brumus14:
Implements #11818. Added bind flag (u) that makes the bind ignore the submap so it is active no matter the current submap.

I updated syntax in the bind argument loop to remove {} as many cases are one liners, just checking that this is the correct style.
https://github.com/Brumus14/Hyprland/blob/71052a3850eebfae67275c60828a5efdfbde2d01/src/config/ConfigManager.cpp#L2568-L2601

Should be ready for merging.

### DIFF
--- a/hyprtester/src/tests/main/keybinds.cpp
+++ b/hyprtester/src/tests/main/keybinds.cpp
@@ -34,6 +34,17 @@ static bool checkFlag() {
     return exists;
 }
 
+static bool attemptCheckFlag(int attempts, int intervalMs) {
+    for (int i = 0; i < attempts; i++) {
+        if (checkFlag())
+            return true;
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(intervalMs));
+    }
+
+    return false;
+}
+
 static std::string readKittyOutput() {
     std::string output = Tests::execAndGet("kitten @ --to unix:/tmp/hyprtester-kitty.sock get-text --extent all");
     // chop off shell prompt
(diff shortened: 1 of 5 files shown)


cannedbeef:
About the {} removal, wouldn't that just be better as a switch?

Brumus14:
Yes think that would be a good idea will change it

hyprtester/src/tests/main/keybinds.cpp
```diff
@@ -443,6 +443,36 @@ static void testSubmap() {
...
+static void testSubmapUniversal() {
+    NLog::log("{}Testing submap universal", Colors::GREEN);
+
+    EXPECT(checkFlag(), false);
+    EXPECT(getFromSocket("/keyword bindu SUPER,Y,exec,touch " + flagFile), "ok");
+    EXPECT_CONTAINS(getFromSocket("/submap"), "default");
+
+    // keybind works on default submap
+    OK(getFromSocket("/dispatch plugin:test:keybind 1,7,29"));
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
vaxerski: don't sleep in tests. Runner VMs are very slow and unpredictable. Just don't rely on timings.

I don't think this sleep (and the one further down) is even needed?

Brumus14: I did try removing the sleeping and the test failed, seemed like key press wasn't being registered because it was too fast but works fine not sleeping on the other key presses which is weird.

ItsOhen: Since it's touching a file and you check if the file exists, this is to be expected.
The socket call is async and so is the exec so without the sleep hyprtester just moves along and checks if the file exists before the file have been created.

vaxerski: yeah, check for the file in a loop every x ms like 30 times

hyprtester/src/tests/main/keybinds.cpp
```diff
@@ -443,6 +454,34 @@ static void testSubmap() {
...
+    OK(getFromSocket("/dispatch plugin:test:keybind 0,7,29"));
+    EXPECT(waitCheckFlag(30, 5), true);
+
+    // keybind works on submap1
+    getFromSocket("/dispatch plugin:test:keybind 1,7,30");
+    getFromSocket("/dispatch plugin:test:keybind 0,7,30");
+    EXPECT_CONTAINS(getFromSocket("/submap"), "submap1");
+    OK(getFromSocket("/dispatch plugin:test:keybind 1,7,29"));
+    OK(getFromSocket("/dispatch plugin:test:keybind 0,7,29"));
+    EXPECT(waitCheckFlag(30, 5), true);
```
vaxerski: you dont clear the flag beforehand, is that intended?

Brumus14: The `checkFlag()` function already clears it though

Brumus14:
Just updated commit to change the name of the function I added

hyprtester/src/tests/main/keybinds.cpp
vaxerski: mb

vaxerski:
mb there is one

Issue #9361, opened 2025-02-07
VRR inconsistent behavior


Skidamek:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

Not a regression - it's an old bug

### System Info and Hyprland Version

<details> <summary>System/Version info</summary>

```
Hyprland 0.47.2 built from branch  at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: Sun Feb 2 00:47:17 2025
Tag: v0.47.2, commits: 5767
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set


System Information:
System name: Linux
Node name: Cachy
Release: 6.13.1-2-cachyos
Version: #1 SMP PREEMPT_DYNAMIC Tue, 04 Feb 2025 03:36:12 +0000


GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA104 [GeForce RTX 3070 Ti] [10de:2482] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  570.86.16  Release Build  (notroot@898e65f6bf83)  Tue Feb  4 03:49:39 UTC 2025


os-release: NAME="CachyOS Linux"
PRETTY_NAME="CachyOS"
ID=cachyos
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://cachyos.org/"
DOCUMENTATION_URL="https://wiki.cachyos.org/"
SUPPORT_URL="https://discuss.cachyos.org/"
BUG_REPORT_URL="https://github.com/cachyos"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=cachyos


plugins:
```

</details>

### Description

vrr behaves strangely, it generally works but in some cases it refuses to work or just blocks on half of the actual game framerate unless i start to move my mouse.
I've conducted tests on [latest git](https://github.com/hyprwm/Hyprland/commit/54441e0c4e51dd182f78876c014446d5d0359ba8), 0.47.2 and 0.46.2 on nvidia gpu with `no_hardware_cursors` set to `true` - so only software cursors. In all of those versions I noticed the same behavior.
[...]

PR #12118, opened 2025-10-24
renderer: remove opengl:nvidia_anti_flicker

fxzzi:

we use explicit sync now so this option is useless afaik. I've been running with the option disabled since the inception of syncobj in Hyprland without issues.

https://github.com/hyprwm/Hyprland/issues/9361#issuecomment-2646299276

keep the glfinish() there because as stated it may help with sw renderers

breaking change that the option is removed i guess

mhm

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1467,17 +1467,6 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     },
 
     /*
-     * opengl:
-     */
-
-    SConfigOptionDescription{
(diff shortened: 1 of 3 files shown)


fxzzi:
i realise this is useless anyway since we check for explicit sync anyway lol

gulafaran:
its somewhat useless yes, but i would leave the isNvidia() in there anyways incase some oddball nvidia card dont support explicit sync and hits the implicit sync path to not regress that. but the option itself is useless. and could be removed with an wiki PR too. cleaning up is never bad :D

fxzzi:
done but can't re open the pr I think I broek it

edit: it's because i force pushed lol, i just opened another pr

PR #12119, opened 2025-10-24
renderer: remove opengl:nvidia_anti_flicker

fxzzi:
#12118 didn't realise force pushing made it so i couldn't re open lmao

### DIFF
--- a/src/config/legacy/ConfigManager.cpp
+++ b/src/config/legacy/ConfigManager.cpp
@@ -766,8 +766,6 @@ CConfigManager::CConfigManager() {
     registerConfigVar("xwayland:force_zero_scaling", Hyprlang::INT{0});
     registerConfigVar("xwayland:create_abstract_socket", Hyprlang::INT{0});
 
-    registerConfigVar("opengl:nvidia_anti_flicker", Hyprlang::INT{1});
-
     registerConfigVar("cursor:invisible", Hyprlang::INT{0});
     registerConfigVar("cursor:no_hardware_cursors", Hyprlang::INT{2});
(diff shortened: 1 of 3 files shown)


vaxerski:
why?

gulafaran:
> why?

its just a config option that has about zero usage. explicit sync is enabled by default, the only way to hit that code path is if your nvidia card doesnt support explicit sync and i dont think there is one that also supports drivers new enough to run wayland. and even if there is without doing glfinish on them its going to artifact heavily, not that glfinish is the 100% fix it all solution but it was more of an attempt at so.

so the config option itself is rather obsolete imo

fxzzi:
> why?

Sometimes people think it does shit when in fact the option doesn't do shit (in most cases)

UjinT34:
Is there a huge difference between `glFlush` and `glFinish`? Anything that works well with `glFlush` should also work with explicit sync. Probably we should just leave a `glFinish` there without any extra conditions as a more reliable option.

gulafaran:
> Is there a huge difference between `glFlush` and `glFinish`? Anything that works well with `glFlush` should also work with explicit sync. Probably we should just leave a `glFinish` there without any extra conditions as a more reliable option.
 
 yeah its a huge difference, glflush is more of send all queued commands to the driver, non blocking and returns.
 glfinish is blocking until queued commands are finished. or well on nvidia who knows what it does.
 
and since all drivers besides nvidia has implicit sync and sometimes doesnt support explicit sync there are a a few previous bugreports about it since the move to default explicit sync was done. or even *bsd just lacking proper support. moving to glfinish only would make things a lot slower there

PR #12124, opened 2025-10-24
layer-shell: fix fullscreen alpha when changing layers

VirtCode:
So I was experimenting with switching the layer of my bar on demand to show it above fullscreen windows. But then I noticed that hyprland doesn't change the alpha of a layer surface if it switches layer. ~~Additionally layers opened on the `top` layer with a window in fullscreen wouldn't have their alpha set correctly either.~~

So this MR fixes most of these issues when a window is in fullscreen:
- ~~Opening a `top` layer is invisible as it should be~~ (apparently this is intended)
- Moving a layer from `overlay` to somewhere else makes it invisible
- Moving a layer to `overlay` makes it visible
- The moving between `background` or `bottom` and `top` works correctly without artifacts (one minor issue see below)

There are two anomalies which still exist with the current approach but I don't think they are fixed without increasing complexity/rewriting a lot:
- When moving from `overlay` to `bottom` or `background` it does not fade - it just disappears (because the layer stays visible, just the order changed).
- When the fullscreened window is partially transparent, moving between `background` or `bottom` and `top` or `overlay` looks a bit weird but I don't even know what the correct behaviour would be in this case.

yes

### DIFF
--- a/src/desktop/LayerSurface.cpp
+++ b/src/desktop/LayerSurface.cpp
@@ -303,7 +303,7 @@ void CLayerSurface::onCommit() {
             auto PWORKSPACE = PMONITOR->m_activeSpecialWorkspace ? PMONITOR->m_activeSpecialWorkspace : PMONITOR->m_activeWorkspace;
             if (PWORKSPACE && PWORKSPACE->m_fullscreenMode == FSMODE_FULLSCREEN) {
                 // warp if switching render layer so we don't see glitches and have clean fade
-                if ((m_layer == ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM || m_layer == ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM) &&
+                if ((m_layer == ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND || m_layer == ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM) &&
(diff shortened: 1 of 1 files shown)


vaxerski:
opening a top layer should be above fs (if opened after fs)

VirtCode:
oh didn't know that this was actually intended behaviour, seemed pretty counter-intuitive to me

but sure here you go

vaxerski:
it do be intended behavior cuz many launchers open on top and it'd be weird for it to grab kb and be invisible no?

src/desktop/LayerSurface.cpp
```diff
@@ -289,18 +289,32 @@ void CLayerSurface::onCommit() {
...
                     PMONITOR->m_layerSurfaceLayers[m_layer].erase(it);
                     break;
                 }
             }
 
+            // update alpha when window is in fullscreen
+            auto PWORKSPACE = PMONITOR->m_activeSpecialWorkspace ? PMONITOR->m_activeSpecialWorkspace : PMONITOR->m_activeWorkspace;
+            if (PWORKSPACE && PWORKSPACE->m_fullscreenMode == FSMODE_FULLSCREEN) {
+                // warp if switching render layer so we don't see glitches and have clean fade
+                if ((m_layer == ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM || m_layer == ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM) &&
```
vaxerski: I think you did an XD kere

VirtCode: whoops

VirtCode:
> it do be intended behavior cuz many launchers open on top and it'd be weird for it to grab kb and be invisible no?

true didn't consider that, yeah that would be pretty stupid

PR #12133, opened 2025-10-25
Allow DS for surfaces with inert subsurfaces

njdom24:

Gets DS working with Gamescope. `gamescope --mangoapp` disables DS for `subsurfaces`, and toggling it on/off with `Shift_R+F12` works as expected.
Note: Gamescope doesn't do hardware cursors with its Wayland backend, so moving the cursor will go in/out of DS. This outcome is the same as Sway, and works fine for games with captured cursors, so I think it's okay. Given that toggling `mangoapp` doesn't cause this stutter, it's probably a Gamescope issue.

My knowledge of subsurfaces is lacking. If this is stupid, please say so.

Ready

### DIFF
--- a/src/desktop/Window.cpp
+++ b/src/desktop/Window.cpp
@@ -1927,13 +1927,18 @@ SP<CWLSurfaceResource> CWindow::getSolitaryResource() {
     if (res->m_subsurfaces.size() == 0)
         return res;
 
-    if (res->m_subsurfaces.size() == 1) {
-        if (res->m_subsurfaces[0].expired() || res->m_subsurfaces[0]->m_surface.expired())
-            return nullptr;
-        auto surf = res->m_subsurfaces[0]->m_surface.lock();
-        if (!surf || surf->m_subsurfaces.size() != 0 || surf->extends() != res->extends() || !surf->m_current.texture || !surf->m_current.texture->m_opaque)
-            return nullptr;
-        return surf;
+    if (res->m_subsurfaces.size() >= 1) {
+        if (!res->hasVisibleSubsurface())
+            return res;
+
+        if (res->m_subsurfaces.size() == 1) {
+            if (res->m_subsurfaces[0].expired() || res->m_subsurfaces[0]->m_surface.expired())
+                return nullptr;
+            auto surf = res->m_subsurfaces[0]->m_surface.lock();
+            if (!surf || surf->m_subsurfaces.size() != 0 || surf->extends() != res->extends() || !surf->m_current.texture || !surf->m_current.texture->m_opaque)
+                return nullptr;
+            return surf;
+        }
     }
 
     return nullptr;
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -602,6 +602,19 @@ void CWLSurfaceResource::sortSubsurfaces() {
     }
 }
 
+bool CWLSurfaceResource::hasVisibleSubsurface() {
+    for (auto const& subsurface : m_subsurfaces) {
+        if (!subsurface || !subsurface->m_surface)
+            continue;
+
+        const auto& surf = subsurface->m_surface;
+        if (surf->m_current.size.x > 0 && surf->m_current.size.y > 0)
+            return true;
+    }
+
+    return false;
+}
+
 void CWLSurfaceResource::updateCursorShm(CRegion damage) {
     if (damage.empty())
         return;
--- a/src/protocols/core/Compositor.hpp
+++ b/src/protocols/core/Compositor.hpp
@@ -128,6 +128,7 @@ class CWLSurfaceResource {
     void                                   commitState(SSurfaceState& state);
     NColorManagement::SImageDescription    getPreferredImageDescription();
     void                                   sortSubsurfaces();
+    bool                                   hasVisibleSubsurface();
 
     // returns a pair: found surface (null if not found) and surface local coords.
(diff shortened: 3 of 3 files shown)


vaxerski:
> My knowledge of subsurfaces is lacking. If this is stupid, please say so.

idea aint stupid, code sux tho

src/desktop/Window.cpp
```diff
@@ -1910,6 +1910,31 @@ bool CWindow::priorityFocus() {
     return !m_isX11 && CAsyncDialogBox::isPriorityDialogBox(getPID());
 }
 
+bool has_visible_subsurface(const SP<CWLSurfaceResource>& res) {
```
vaxerski: buddy this is not C. Wrong case styling, and this should be a member of CWLSurfaceResource

src/desktop/Window.cpp
```diff
@@ -1910,6 +1910,31 @@ bool CWindow::priorityFocus() {
...
+    for (size_t i = 0; i < res->m_subsurfaces.size(); i++) {
```
vaxerski: use a foreach range loop

src/desktop/Window.cpp
```diff
@@ -1910,6 +1910,31 @@ bool CWindow::priorityFocus() {
...
+
+            auto surf = res->m_subsurfaces[i]->m_surface.lock();
+            if (!surf)
+                continue;
+
+            if (has_visible_subsurface(surf))
+                return true;
+
+            auto tex = surf->m_current.texture;
+            if (tex && tex->m_size.x > 0 && tex->m_size.y > 0) {
```
vaxerski: wut?

src/desktop/Window.cpp
```diff
@@ -1924,13 +1949,19 @@ SP<CWLSurfaceResource> CWindow::getSolitaryResource() {
...
 
-    if (res->m_subsurfaces.size() == 1) {
-        if (res->m_subsurfaces[0].expired() || res->m_subsurfaces[0]->m_surface.expired())
-            return nullptr;
-        auto surf = res->m_subsurfaces[0]->m_surface.lock();
-        if (!surf || surf->m_subsurfaces.size() != 0 || surf->extends() != res->extends() || !surf->m_current.texture || !surf->m_current.texture->m_opaque)
-            return nullptr;
-        return surf;
+    if (res->m_subsurfaces.size() >= 1) {
+        if (!has_visible_subsurface(res)) {
```
vaxerski: no {}

src/desktop/Window.cpp
njdom24: Updated to be (I hope) less nonsense. Now checking `surf->m_current.size` instead of the potentially-nonexistent `texture`'s size.

My current idea is now to just pass through subsurfaces with empty sizes. Seems to still be enough for Gamescope.

src/protocols/core/Compositor.cpp
```diff
@@ -603,6 +603,24 @@ void CWLSurfaceResource::sortSubsurfaces() {
     }
 }
 
+bool CWLSurfaceResource::hasVisibleSubsurface() {
+    for (auto const& subsurface : m_subsurfaces) {
+        if (subsurface) {
+            if (subsurface.expired() || subsurface->m_surface.expired()) {
```
vaxerski: no {}

src/protocols/core/Compositor.cpp
```diff
@@ -603,6 +603,24 @@ void CWLSurfaceResource::sortSubsurfaces() {
...
+                return true;
+            }
+            auto surf = subsurface->m_surface.lock();
+            if (!surf)
+                continue;
+
+            if ((surf->m_current.size.x > 0 && surf->m_current.size.y > 0) || surf->hasVisibleSubsurface()) {
```
vaxerski: no {}

Also, if a subsurface is 0x0, its children will not be drawn, so that can be optimized, recursive check can be removed in that case

njdom24: My thinking was that if a subsurface wasn't 0x0, I wanted to see if it exclusively had children that were.

Might not be a realistic check. I'll test tonight.

njdom24: Fixed styling (Sorry, I'll learn to read next time) and removed the recursive call. Everything still works as-expected. Thanks!

src/protocols/core/Compositor.cpp
```diff
@@ -603,6 +603,23 @@ void CWLSurfaceResource::sortSubsurfaces() {
     }
 }
 
+bool CWLSurfaceResource::hasVisibleSubsurface() {
+    for (auto const& subsurface : m_subsurfaces) {
+        if (subsurface) {
+            if (subsurface.expired() || subsurface->m_surface.expired())
+                return true;
+
+            auto surf = subsurface->m_surface.lock();
```
vaxerski: locking is redundant. Check operator bool on WP and deref a weak

src/protocols/core/Compositor.cpp
```diff
@@ -603,6 +603,23 @@ void CWLSurfaceResource::sortSubsurfaces() {
     }
 }
 
+bool CWLSurfaceResource::hasVisibleSubsurface() {
+    for (auto const& subsurface : m_subsurfaces) {
+        if (subsurface) {
+            if (subsurface.expired() || subsurface->m_surface.expired())
```
vaxerski: prefer operator bool which is the inverse of expired

PR #12148, opened 2025-10-27
xwayland: set _NET_WORKAREA property

jmanc3:
Setting the [_NET_WORKAREA](https://specifications.freedesktop.org/wm-spec/1.3/ar01s03.html#id-1.4.11) property fixes steam notifications toast (and related X notification programs) https://github.com/hyprwm/Hyprland/discussions/12131

This PR gracefully handles the most common use cases, and falls back to old behavior when it can't do something that is good.

Namely for one screen setups, or multiple screen setups of similar size and similar reserved areas taken up, it'll work flawlessly. If these two conditions are not met, then we just remove the property and let old behavior happen.

example2

### DIFF
--- a/src/xwayland/XWM.cpp
+++ b/src/xwayland/XWM.cpp
@@ -1194,11 +1194,9 @@ void CXWM::updateClientList() {
 }
 
 void CXWM::updateWorkArea(int x, int y, int w, int h) {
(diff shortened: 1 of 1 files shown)


vaxerski:
I guess you can push xwm in the right direction with this... but does this work with xwayland:force_zero_scaling?

jmanc3:
image

Works now if it's set true or false

Issue #12088, opened 2025-10-21
`hover_icon_on_border` doesn't work


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11316

## Description
It used to, doesn't seem to no more.

Brumus14:
https://github.com/hyprwm/Hyprland/blob/057695bc3f7de5e8841c15252fc51029590895e4/src/managers/input/InputManager.cpp#L1886-L1890

I think it might be due to this as removing it seems to fix it but not sure whether it's safe to remove.

gulafaran:
> [Hyprland/src/managers/input/InputManager.cpp](https://github.com/hyprwm/Hyprland/blob/057695bc3f7de5e8841c15252fc51029590895e4/src/managers/input/InputManager.cpp#L1886-L1890)
> 
> Lines 1886 to 1890 in [057695b](/hyprwm/Hyprland/commit/057695bc3f7de5e8841c15252fc51029590895e4)
>  // do not override cursor icons set by mouse binds 
>  if (g_pInputManager->m_currentlyDraggedWindow.expired()) { 
>      m_borderIconDirection = BORDERICON_NONE; 
>      return; 
>  } 
> 
> I think it might be due to this as removing it seems to fix it but not sure whether it's safe to remove.

seems https://github.com/hyprwm/Hyprland/pull/7071 changed the logics of using BORDERICON_NONE when a keybind being active to currentlyDraggedWindow which seems wrong. its not like the border is being dragged when hovering it.

vaxerski:
the current cursor override system sucks anyways, I need to rewrite it with a tiered system.

alba4k:
Same as this I think? If so, should I close that discussion?

https://github.com/hyprwm/Hyprland/discussions/10630

vaxerski:
I closed it for you, yeah.

PR #12166, opened 2025-10-30
cursor: refactor override handling

vaxerski:
much cleaner and more reliable. Should fix https://github.com/hyprwm/Hyprland/issues/12088

cc @Brumus14 @alba4k @gulafaran for testing, please test this.

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -12,6 +12,7 @@
 #include "../managers/LayoutManager.hpp"
 #include "../managers/EventManager.hpp"
 #include "../managers/HookSystemManager.hpp"
+#include "../managers/cursor/CursorShapeOverrideController.hpp"
 
 void IHyprLayout::onWindowCreated(PHLWINDOW pWindow, eDirection direction) {
     CBox       desiredGeometry = g_pXWaylandManager->getGeometryForWindow(pWindow);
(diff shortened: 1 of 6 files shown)


alba4k:
Will test tomorrow

aphelei:
with only a floating window it works when going to the surrounding of the window. not when going to the window

https://github.com/user-attachments/assets/30804eb6-c97b-4121-b576-55acaf03e6a1


https://github.com/user-attachments/assets/021ce452-4415-4de2-b7b6-237e5bf469a9


It works with a tiled window below.

vaxerski:
@aphelei yeah I know why this happens but I dont really wanna fix it atm, oh well, it's quite minor

vaxerski:
is there anything else or?

alba4k:
I'm not experiencing what @aphelei pointed out, maybe because I have `border_size != 0`?

Other than this, my only complaint would be the cursor changing shape even on borders that can not be moved in tiled windows (those that are on the edge of the screen). I know this is probably intended behavior, but I still don't really thing that's how this should work.

The whole point of the cursor changing shape is signaling to the user that this border can be dragged, but that is not the case in this scenario, meaning that we're "falsely advertising" borders that are immovable as draggable

alba4k:
https://github.com/user-attachments/assets/87dd8e43-94df-4039-8ebf-0a3db180abe9

I think this looks weird. But sure, the window actually resizes, at least. But when there is a single tiled window on screen, it makes no sense to change the cursor shape

aphelei:
this still happens idk if it can be fixed in this commit https://github.com/hyprwm/Hyprland/discussions/10762

it's """ok""" with very high extend_border_grab_area but it's not ideal at all
There is a visual glitch with hyprbars but idk if it's related (maybe https://github.com/hyprwm/hyprland-plugins/issues/392). anyway it's a plugin

https://github.com/user-attachments/assets/73620530-3fa1-4aaa-9da1-f7d70ab78c53


> I'm not experiencing what @aphelei pointed out, maybe because I have `border_size != 0`?

you're right. But it still doesn't work if border_size is set with window rules

https://github.com/user-attachments/assets/79663bf2-0b56-4d18-8d9c-76b23c9c2c7b

vaxerski:
I am merging this, I decided I don't care. That functionality can be improved in the future.

PR #12204, opened 2025-11-02
cm: higher-quality tonemapping

njdom24:

Inspired by KWin. Updates the tonemapper to use the `src` `imageDescription`'s luminance info instead of throwing it away. Also scales the reference luminance to match that of the destination, preventing over-bright screenshots / recordings.

Also tossed in a quick fix missed from  https://github.com/hyprwm/Hyprland/pull/12094, which prevents overly-dark screenshares in SDR when `render:cm_sdr_eotf` > 0.

HDR screenshot tonemapped to SDR:
hdr_to_sdr

SDR screenshot:
reg_sdr

Without these changes:
image

I considered a simple luminance multiplier as a proof-of-concept when getting started. This is what that would look like by keeping the current `tonemap` and simply multiplying the end result by `80.0 / 203.0`:
image

Versus this PR:

image

The changes to `blurprepare.frag` fix the inverted color issue with blur from SDR->HDR.
Screenshot of my desktop in SDR:
desktop_sdr

And in HDR tonemapped back to SDR with these changes:
desktop_hdr

The blur is still too strong, but the colors are a lot better.

I'm not super confident in the "correctness" of tonemapper changes (really just looked at prior art), but it does improve HDR screenshares immensely.

Ready for merging


vaxerski:
@UjinT34

UjinT34:
Current tonemapping is also inspired by kwin. If they changed theirs then we should probably do the same. Might be harder to track every necessary change because they combine cpu calcs, shaders and drm properties.
Tonemapping happens when source max luminance is higher than the target one. Most games set their luminance to 10000 or some other high value which almost always results in HDR -> HDR tonemapping. So this PR will affect desktop HDR and fullscreen HDR with `cm_fs_passthrough = 0`. Might make it better or worse but most likely it'll be too subtle to notice.

njdom24:
@UjinT34 Thanks, I didn't consider that.
Do you think it warrants a check for HDR-on-HDR to bail, and maybe a new config var for that?

My HDR displays are all Mini-LED, so I can't be sure how it'll look on e.g. OLEDs with HDR400.

UjinT34:
`tonemap` works with linear values and shouldn't care about SDR/HDR. I don't think we need a separate codepath for that unless someone has issues with HDR and new tonemapping.

njdom24:
Rebased and moved to a more general HDR-to-SDR check to use transfer function luminances for tonemapping across the board.

This change makes the improved tonemapping apply to HDR content on SDR displays too, instead of just in screenshare.

If Auto-HDR might cause apps to always output HDR, this should make things look okay when you're un-fullscreened and back to SDR.

njdom24:
Updated to fix a slight blue-green hue I noticed for HDR->SDR.

njdom24:
@UjinT34 I've been using this with https://github.com/hyprwm/Hyprland/pull/12127 on my system for a while, and everything seems to work as intended.

Any changes you'd like to see to get them merged?

UjinT34:
lgtm

njdom24:
@vaxerski ?

vaxerski (approved):

PR #12216, opened 2025-11-05
cm: scale reference luminance when tonemapping

njdom24:

Simpler alternative to https://github.com/hyprwm/Hyprland/pull/12204

When tonemapping from HDR to SDR, scale the HDR source's reference luminance (203) to SDR (80). Prevents HDR screenshare from blowing out the brightness.

Unlike https://github.com/hyprwm/Hyprland/pull/12204, this method makes games typically look closer to how they do in SDR, at the risk of losing highlight detail that a game's native tonemapper might prioritize better.

Screenshots are of an extremely bright area, sort of a worst-case scenario, to emphasize the difference.

HDR-to-SDR on `master`:
image

HDR-to-SDR with this PR:
image

Native SDR:
image

Also brings in a quick fix for https://github.com/hyprwm/Hyprland/pull/12094, which made screensharing too dark when `render:cm_sdr_eotf` > 0, which became apparent when testing this.

Heavily WIP

### DIFF
--- a/src/protocols/types/ColorManagement.hpp
+++ b/src/protocols/types/ColorManagement.hpp
@@ -6,8 +6,10 @@
 
 #define SDR_MIN_LUMINANCE 0.2
 #define SDR_MAX_LUMINANCE 80.0
+#define SDR_REF_LUMINANCE 80.0
 #define HDR_MIN_LUMINANCE 0.005
 #define HDR_MAX_LUMINANCE 10000.0
+#define HDR_REF_LUMINANCE 203.0
 #define HLG_MAX_LUMINANCE 1000.0
 
 namespace NColorManagement {
@@ -225,6 +227,25 @@ namespace NColorManagement {
                 case CM_TRANSFER_FUNCTION_ST428:
                 case CM_TRANSFER_FUNCTION_SRGB:
                 default: return sdrMaxLuminance >= 0 ? sdrMaxLuminance : SDR_MAX_LUMINANCE;
+            }
+        };
+
+        float getTFRefLuminance(int sdrRefLuminance = -1) const {
+            switch (transferFunction) {
+                case CM_TRANSFER_FUNCTION_EXT_LINEAR:
+                case CM_TRANSFER_FUNCTION_ST2084_PQ:
+                case CM_TRANSFER_FUNCTION_HLG: return HDR_REF_LUMINANCE;
+                case CM_TRANSFER_FUNCTION_GAMMA22:
+                case CM_TRANSFER_FUNCTION_GAMMA28:
+                case CM_TRANSFER_FUNCTION_BT1886:
+                case CM_TRANSFER_FUNCTION_ST240:
+                case CM_TRANSFER_FUNCTION_LOG_100:
+                case CM_TRANSFER_FUNCTION_LOG_316:
+                case CM_TRANSFER_FUNCTION_XVYCC:
+                case CM_TRANSFER_FUNCTION_EXT_SRGB:
+                case CM_TRANSFER_FUNCTION_ST428:
+                case CM_TRANSFER_FUNCTION_SRGB:
+                default: return sdrRefLuminance >= 0 ? sdrRefLuminance : SDR_REF_LUMINANCE;
             }
         };
 
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1558,6 +1558,13 @@ static bool isSDR2HDR(const NColorManagement::SImageDescription& imageDescriptio
          targetImageDescription.transferFunction == NColorManagement::CM_TRANSFER_FUNCTION_HLG);
 }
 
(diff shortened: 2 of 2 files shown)


UjinT34:
Does it work as expected with overriden monitorv2 luminances?

https://github.com/hyprwm/Hyprland/blob/46b71eda6423cc9c3b19cb310c3344e81ac624c5/src/render/shaders/glsl/CM.glsl#L374-L408 does some scaling, looks like it should do the same trick. Maybe some math or constants aren't correct or missing.

njdom24:
@UjinT34 We throw that info away in the `cmBackToSRGB` block: https://github.com/hyprwm/Hyprland/blob/46b71eda6423cc9c3b19cb310c3344e81ac624c5/src/render/OpenGL.cpp#L1713

I rocked the boat more with https://github.com/hyprwm/Hyprland/pull/12204, but even then I forced the luminanances to match the transfer function with
```
imageDescription.luminances = {
                .min = imageDescription.getTFMinLuminance(-1), .max = imageDescription.getTFMaxLuminance(-1), .reference = imageDescription.getTFRefLuminance(-1)};
```

Because monitorv2 settings would make an impact otherwise, and I'm not sure it'd ever be in a good way.

Edit: It'd probably be better to do the same luminance override in this branch

njdom24:
Do the `monitorv2` `min_luminance`, `max_luminance`, `max_avg_luminance` serve a purpose? It feels like they're just worked around with `getTFMinLuminance`, ``getTFMaxLuminance``, and now `getTFRefLuminance`.

UjinT34:
> Do the `monitorv2` `min_luminance`, `max_luminance`, `max_avg_luminance` serve a purpose? It feels like they're just worked around with `getTFMinLuminance`, `getTFMaxLuminance`, and now `getTFRefLuminance`.

A monitor should provide those values through EDID. Sometimes EDID can't be fully parsed or the values are incorrect. Those monitorv2 rules are meant to fix that. `getTF*` values should be used for the source, EDID/monitorv2 values are for the output. Screen capture uses the main fb iirc and it's encoded with the EDID/monitorv2 values.

s0mebodyhelpme:
Interestingly enough, sharing a screen with HDR on works for normal windows with this PR, but in Marvel's Spider-Man Remastered on Steam the game looks undersaturated in the screenshare when focused on the window and slightly overly bright when focused away with `cm = srgb` and `cm_fs_passthrough = 1` with auto hdr on. Same is true for `cm = hdr` setting. Also, mpv isn't affected by `cm_fs_passthrough = 2` with `cm = hdr` and auto hdr off. This makes it an issue to use `cm_fs_passthrough = 1` with hdr on because you can't fullscreen normal windows without making them blown out or oversaturated and dark.

njdom24:
@s0mebodyhelpme I've seen most of the issues you described with `cm_fs_passthrough`, but that's out of the scope of this PR.

These changes just address better conversion when we correctly make the determination that we want to share HDR, but it does seem like there's an issue with making that determination somewhere in the pipeline.

I'm surprised to see the issue with ` cm = hdr`, though. That one's new to me. If the issue you're seeing with `mpv` is similar to what I saw with VLC, I did https://github.com/hyprwm/Hyprland/pull/12127 to address it, but I need to take another look at the passthrough settings to see if it's handling them correctly. Personally, I've never had `cm_fs_passthrough = 1` work.

s0mebodyhelpme:
@njdom24 im not sure if the issue with mpv is actually to do with direct scanout, as I have that set to 0, but I can try using both patches to see if that resolves the issue on my end. On another note, though, I am glad you're working on this. Willing to test out any future changes and report back if you welcome it.

njdom24:
@s0mebodyhelpme If you don't have direct scanout enabled, the linked PR won't improve things, unfortunately. I'll see if I can reproduce it.

s0mebodyhelpme:
@njdom24  Update: `cm_fs_passthrough = 2` _does_ work, but when cm is srgb and/or (i'm guessing) not set to hdr. Something about how hdr mode is handled prevents it from working?

Edit: Also, screensharing an HDR game like Spider-Man Remastered looks right ONLY if focus is away from the window?

njdom24:
@s0mebodyhelpme Are you sharing the window or the screen? All of my testing has been with the screen

s0mebodyhelpme:
@njdom24 i noticed this on sharing a window, but it also becomes undersaturated when focused on the same app during sharing the screen (I know because I viewed the stream on a separate device).

s0mebodyhelpme:
@njdom24 latest commit fails with this patch, and on my end the merging of your other pull does not fix screenshare with HDR (without this branch, it's overly bright). I'm assuming you've got something in mind for this already, but fwiw, just wanted to let you know fyi.

Edit: Update--the screenshare working when focused away only happens when srgb is set as cm and not hdr, so im assuming that it's because the cm is set back to srgb when unfocused.

PR #12219, opened 2025-11-05
Fix fractional scale artifact

bea4dev:
* In fractional scaling environments, we confirmed that the shrink adjustment for `render:expand_undersized_textures` was reacting even to rounding errors of less than 1px, causing textures to be unnecessarily trimmed.
* In `CHyprRenderer::calculateUVForSurface`, we introduced a 1px threshold to the `projSize < EXPECTED_SIZE` check so that UVs are only shrunk on axes where the actually visible area is sufficiently smaller than the expected size.
* As a result, pixel-level offsets under fractional scaling are eliminated, while the original behavior is preserved in scenarios where the texture should indeed be reduced.

- before

image

- after

image

For now, there’s nothing in particular.

ready

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -1137,13 +1137,11 @@ void CHyprRenderer::calculateUVForSurface(PHLWINDOW pWindow, SP<CWLSurfaceResour
             const Vector2D MISALIGNMENT = pSurface->m_current.bufferSize - projSize;
             if (MISALIGNMENT != Vector2D{})
                 uvBR -= MISALIGNMENT * PIXELASUV;
-        }
-
-        // if the surface is smaller than our viewport, extend its edges.
-        // this will break if later on xdg geometry is hit, but we really try
-        // to let the apps know to NOT add CSD. Also if source is there.
-        // there is no way to fix this if that's the case
-        {
+        } else {
+            // if the surface is smaller than our viewport, extend its edges.
+            // this will break if later on xdg geometry is hit, but we really try
+            // to let the apps know to NOT add CSD. Also if source is there.
+            // there is no way to fix this if that's the case
             const auto MONITOR_WL_SCALE = std::ceil(pMonitor->m_scale);
             const bool SCALE_UNAWARE    = MONITOR_WL_SCALE > 1 && (MONITOR_WL_SCALE == pSurface->m_current.scale || !pSurface->m_current.viewport.hasDestination);
             const auto EXPECTED_SIZE    = getSurfaceExpectedSize(pWindow, pSurface, pMonitor, main).value_or((projSize * pMonitor->m_scale).round());


vaxerski:
I don't like this fix because it's now up to the driver to decide what to do.
Imagine you have a 5 px wide texture and a 6px wide destination:

```
r r r g g
vvvvvvv
x x x x x x
```

I propose to adjust the UV to instead just stretch the last pixel. It's what we do with expand_undersized_textures anyways during animations, and IMO this will be much less of a distraction if it's just the last row/column of pixels repeated.

We can't escape _some_ image degradation.

vaxerski:
we don't need the shader stuff - we use GL_CLAMP_TO_EDGE anyways, so all you need to do is shrink the bottom right UV a bit (by 1px or whatever the misalignment is)

bea4dev:
> we don't need the shader stuff - we use GL_CLAMP_TO_EDGE anyways, so all you need to do is shrink the bottom right UV a bit (by 1px or whatever the misalignment is)

I’ve already tried that approach.
However, even after experimenting with various tweaks, the artifacts were still present.
Maybe I’m just doing something wrong...?

image

My guess is that this happens because once you start modifying uvBR itself, the small mismatch between projSize and EXPECTED_SIZE gets spread across the entire area, and that distortion ends up becoming visible around the center of the screen.

I worked around the issue by repeating the pixels instead of adjusting uvBR.
So far it appears to be fine, although there is room for discussion as to whether this is really the correct approach.

bea4dev:
One example of the code I tried (in practice I experimented with adding and tweaking various parts, but none of them worked correctly):
```cpp
                const auto SIZEDIFF    = EXPECTED_SIZE - projSize;
                // Introduce a 1px threshold and adjust so that UVs are only shrunk on axes
                // where the actually visible area is sufficiently smaller than the expected size.
                if (!SHOULD_SKIP && (RATIO.x < 1 || RATIO.y < 1)) {
                    const auto SHRINK_X = RATIO.x < 1 && SIZEDIFF.x > 1.0;
                    const auto SHRINK_Y = RATIO.x < 1 && SIZEDIFF.y > 1.0;

                    if (SHRINK_X || SHRINK_Y) {
                        Vector2D fix = {1.0, 1.0};
                        if (SHRINK_X)
                            fix.x = std::clamp(RATIO.x, 0.0001, 1.0);
                        if (SHRINK_Y)
                            fix.y = std::clamp(RATIO.y, 0.0001, 1.0);

                        uvBR = uvBR * fix;
                    } else {
                        Vector2D bufferSize = pSurface->m_current.bufferSize;
                        if (bufferSize.x <= 0.0)
                            bufferSize.x = 1.0;
                        if (bufferSize.y <= 0.0)
                            bufferSize.y = 1.0;

                        const Vector2D PIXEL_UV = Vector2D(1.0, 1.0) / bufferSize;

                        if (SIZEDIFF.x > 0.0)
                            uvBR.x = uvBR.x - PIXEL_UV.x;
                        if (SIZEDIFF.y > 0.0)
                            uvBR.y = uvBR.y - PIXEL_UV.y;
                    }
                }
```

vaxerski:
have you added a breakpoint to make sure the code is hit? Logic itself looks fine to me

bea4dev:
> have you added a breakpoint to make sure the code is hit? Logic itself looks fine to me

I haven’t verified it with breakpoints, but I previously used Debug::log() to confirm that execution does reach that part of the code.

vaxerski:
I just noticed: 
https://github.com/hyprwm/Hyprland/blob/8e8bfbb0b146fb1f3234b2a3e3e92b63989e1993/src/render/Renderer.cpp#L1133-L1140

there already is code for this, I guess it's not working?

bea4dev:
It looks like it’s working, but artifacts are still occurring.
On top of that, for some reason my fix logic also seems to be running while that code is working.
Now I’m really getting confused.

bea4dev:
It might actually be enough to just tweak the initial fix a bit.
Isn’t the artifact happening because, after that logic runs, we’re applying RATIO to uvBR again (effectively applying it twice)?

bea4dev:
I might have managed to fix it.

```cpp
        const auto HAS_MISALIGN = projSize != Vector2D{} && fixMisalignedFSV1;
        if (HAS_MISALIGN) {
            // instead of nearest_neighbor (we will repeat / skip)
            // just cut off / expand surface
            const Vector2D PIXELASUV    = Vector2D{1, 1} / pSurface->m_current.bufferSize;
            const Vector2D MISALIGNMENT = pSurface->m_current.bufferSize - projSize;
            if (MISALIGNMENT != Vector2D{})
                uvBR -= MISALIGNMENT * PIXELASUV;
        }

        // if the surface is smaller than our viewport, extend its edges.
        // this will break if later on xdg geometry is hit, but we really try
        // to let the apps know to NOT add CSD. Also if source is there.
        // there is no way to fix this if that's the case
        {
            const auto MONITOR_WL_SCALE = std::ceil(pMonitor->m_scale);
            const bool SCALE_UNAWARE    = MONITOR_WL_SCALE > 1 && (MONITOR_WL_SCALE == pSurface->m_current.scale || !pSurface->m_current.viewport.hasDestination);
            const auto EXPECTED_SIZE    = getSurfaceExpectedSize(pWindow, pSurface, pMonitor, main).value_or((projSize * pMonitor->m_scale).round());

            const auto RATIO = projSize / EXPECTED_SIZE;
            if (!SCALE_UNAWARE || MONITOR_WL_SCALE == 1) {
                if (*PEXPANDEDGES && !SCALE_UNAWARE && (RATIO.x > 1 || RATIO.y > 1)) {
                    const auto FIX = RATIO.clamp(Vector2D{1, 1}, Vector2D{1000000, 1000000});
                    uvBR           = uvBR * FIX;
                }

                // FIXME: probably do this for in anims on all views...
                const auto SHOULD_SKIP = !pWindow || pWindow->m_animatingIn;
                if (!SHOULD_SKIP && (RATIO.x < 1 || RATIO.y < 1) && !HAS_MISALIGN) {
                    const auto FIX = RATIO.clamp(Vector2D{0.0001, 0.0001}, Vector2D{1, 1});
                    uvBR           = uvBR * FIX;
                }
            }
        }
```

image

vaxerski:
yeah that looks correct, the entire second block can be skipped if we're fixing misalignment as well I'm pretty sure

bea4dev:
I made a few changes.
It still seems to work correctly with this version as well.
Does this look okay now?

```cpp
        if (projSize != Vector2D{} && fixMisalignedFSV1) {
            // instead of nearest_neighbor (we will repeat / skip)
            // just cut off / expand surface
            const Vector2D PIXELASUV    = Vector2D{1, 1} / pSurface->m_current.bufferSize;
            const Vector2D MISALIGNMENT = pSurface->m_current.bufferSize - projSize;
            if (MISALIGNMENT != Vector2D{})
                uvBR -= MISALIGNMENT * PIXELASUV;
        } else {
            // if the surface is smaller than our viewport, extend its edges.
            // this will break if later on xdg geometry is hit, but we really try
            // to let the apps know to NOT add CSD. Also if source is there.
            // there is no way to fix this if that's the case
            const auto MONITOR_WL_SCALE = std::ceil(pMonitor->m_scale);
            const bool SCALE_UNAWARE    = MONITOR_WL_SCALE > 1 && (MONITOR_WL_SCALE == pSurface->m_current.scale || !pSurface->m_current.viewport.hasDestination);
            const auto EXPECTED_SIZE    = getSurfaceExpectedSize(pWindow, pSurface, pMonitor, main).value_or((projSize * pMonitor->m_scale).round());

            const auto RATIO = projSize / EXPECTED_SIZE;
            if (!SCALE_UNAWARE || MONITOR_WL_SCALE == 1) {
                if (*PEXPANDEDGES && !SCALE_UNAWARE && (RATIO.x > 1 || RATIO.y > 1)) {
                    const auto FIX = RATIO.clamp(Vector2D{1, 1}, Vector2D{1000000, 1000000});
                    uvBR           = uvBR * FIX;
                }

                // FIXME: probably do this for in anims on all views...
                const auto SHOULD_SKIP = !pWindow || pWindow->m_animatingIn;
                if (!SHOULD_SKIP && (RATIO.x < 1 || RATIO.y < 1)) {
                    const auto FIX = RATIO.clamp(Vector2D{0.0001, 0.0001}, Vector2D{1, 1});
                    uvBR           = uvBR * FIX;
                }
            }
        }
```

Issue #11674, opened 2025-09-11
Children of pinned windows should be automatically pinned


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11605

## Description
See discussion.

Holbormon:
Instead of limiting this behaviour to pinning, can we consider adding a toggle for **arbitrary window rules** to propagate to children of initial matches ?

It feels like it would make for an easy target when/if the implementation for propagating pinning is added, and would benefit [a lot more use cases](https://www.reddit.com/r/hyprland/comments/1ia3gei/make_all_child_windows_float_by_default_in/).

Aside from the linked use case, I found this discussion by trying to make swallowed windows apply their tags or other properties to the swallower, and came out empty-handed, which would also be solved by the above suggestion.

kchr:
I am not sure if this is entirely related to the topic, but I am having the issue where some floating windows spawned from a tiled window on a special workspace appears behind the tiled window. It is briefly visisble when hiding the special workspace, but I cannot get the floating window to appear over its parent window that is tiled.

In this particular case, the affected floating window is `pinentry-gtk2` launched from a tiled terminal on a (focused) special workspace.

Please direct me elsewhere if this is an unrelated issue.

**EDIT:** Nevermind; restarting my `gpg-agent` seems to solve the problem, but I have no idea how that would matter...

klhrt:
> Instead of limiting this behaviour to pinning, can we consider adding a toggle for **arbitrary window rules** to propagate to children of initial matches ?
> 
> It feels like it would make for an easy target when/if the implementation for propagating pinning is added, and would benefit [a lot more use cases](https://www.reddit.com/r/hyprland/comments/1ia3gei/make_all_child_windows_float_by_default_in/).
> 
> Aside from the linked use case, I found this discussion by trying to make swallowed windows apply their tags or other properties to the swallower, and came out empty-handed, which would also be solved by the above suggestion.

Why not have a field for inheritable tags? This would be incredibly useful, in fact it would allow people to finally dictate where all their steam games go and make them actually start fullscreen without having to make individual rules for each game. It would also avoid having to toggle either "inherit all window rules" or individually toggle rules while applying them. A further potential enhancement would be for each inheritable tag to have a depth which is decremented each time it's inherited, and if no depth was specified it would propagate no matter what.

EXAMPLE:
```
"title": "steam",
"heritableTags": ["steamchild" 2],
```
[...]

PR #12224, opened 2025-11-06
Automatically pin child windows

ArchSav:
Implements #11674 
When a child window is opened, pin it if the parent is already pinned.

Nothing of note.

Yes it is ready.


src/protocols/XDGShell.cpp
```diff
@@ -462,6 +465,9 @@ CXDGSurfaceResource::CXDGSurfaceResource(SP<CXdgSurface> resource_, SP<CXDGWMBas
 
         g_pCompositor->m_windows.emplace_back(CWindow::create(m_self.lock()));
 
+        if (RESOURCE->m_parent != nullptr)
```
vaxerski: no need for != nullptr, pointers have a bool operator

src/protocols/XDGShell.cpp
```diff
@@ -462,6 +465,9 @@ CXDGSurfaceResource::CXDGSurfaceResource(SP<CXdgSurface> resource_, SP<CXDGWMBas
...
+            g_pCompositor->m_windows.back()->m_pinned = true;
```
vaxerski: what the f just store the ptr from CWindow::create in al ocal

ArchSav:
> tests needed

Could you elaborate on this a bit, 
I am not sure how I would go about making tests for this in hyprtester

vaxerski:
make a client that listens on stdin and opens a child window when you send something to it, pin it beforehand, check if pinned. See the pointer test client, it does a similar thing for snapping

ArchSav:
This should be ready to merge now

hyprtester/src/tests/clients/child-window.cpp
```diff
@@ -0,0 +1,124 @@
...
+        return false;
+
+    std::string startStr = std::string{client.readBuf.data()};
+    if (startStr.find("started") == std::string::npos) {
+        NLog::log("{}Failed to start child-window client, read {}", Colors::RED, startStr);
+        return false;
+    }
+
+    // wait for window to appear
+    std::this_thread::sleep_for(std::chrono::milliseconds(2000));
```
vaxerski: never use sleeps in runners. Poll on something. Timings are unreliable. You can poll hyprctl clients. Check Tests::spawnKitty

hyprtester/src/tests/clients/child-window.cpp
```diff
@@ -0,0 +1,124 @@
...
+    ssize_t bytesRead = read(client.fds.fd, client.readBuf.data(), 1023);
+    if (bytesRead == -1)
+        return false;
+
+    client.readBuf[bytesRead] = 0;
+    std::string recieved      = std::string{client.readBuf.data()};
+    recieved.pop_back();
+
+    // wait for window to appear
+    std::this_thread::sleep_for(std::chrono::milliseconds(2000));
```
vaxerski: same thing

hyprtester/src/tests/clients/child-window.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+static bool test() {
+    SClient client;
+
+    OK(getFromSocket("/keyword windowrule float, pin, class:child-test-parent"));
+    if (!startClient(client))
+        return false;
+
+    createChild(client);
+    EXPECT_CONTAINS(getFromSocket("/activewindow"), "pinned: 1");
```
vaxerski: you should check for pinned: 1 being in the string twice, once for the parent, once child :P `EXPECT_COUNT_STRING`

ArchSav:
I've implemented those changes
I appreciate the help getting this through :D

vaxerski:
no worries, it's my job :)

hyprtester/clients/child-window.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+    // surface/toplevel stuff
+    CSharedPointer<CCWlSurface>   surf;
+    CSharedPointer<CCXdgSurface>  xdgSurf;
+    CSharedPointer<CCXdgToplevel> xdgToplevel;
+    Vector2D                      geom;
+
+    // pointer
+    CSharedPointer<CCWlPointer> pointer;
+    uint32_t                    enterSerial;
```
vaxerski: all primitives here need to be initialized, we don't allow uninitialized as it's very bugprone (all *, int, bool, size_t, etc)

ArchSav: sorry, I remember reading that on the wiki, must have slipped my mind -‿-"

vaxerski (approved):

vaxerski:
c-f needed

vaxerski:
Your test seems to fail

```
[32mPassed: [0mgetFromSocket("/keyword windowrule float, pin, class:child-test-parent"). Got ok
[31mFailed: [0mfn(), expected true, got false. Source: hyprtester/src/main.cpp@240.
```

Rebase on main and fix the windowrule syntax :P

vaxerski:
also why the rule even? Just pin the first window via dispatch pin?

ArchSav:
Sorry for the added labels, I messed up on the rebase a bit

vaxerski:
your test fails, did you run it locally?

ArchSav:
It did run without fail locally
the github tests failed when checking if it could read from the client's stdout
https://github.com/hyprwm/Hyprland/blob/1a1865792860c69247117eabc5ae1f87f81d49d8/hyprtester/src/tests/clients/child-window.cpp#L63-L69

since this isn't necessary I have removed all reads from the client

vaxerski:
still nope :/

ArchSav:
I suspect the child window takes too long to load on the test machine and thus have added some more debug info to test this

vaxerski:
no prob, re-ran the testers. Yes, they are very slow

ArchSav:
The child-window client wasn't available to the nix vm

vaxerski:
nice catch. Just needs a rebase then

ArchSav:
done

ArchSav:
I have fixed build problems should be ready to merge

Sorry for the hassle of this

ArchSav:
I have fixed the clang format violation

The failed test doesn't show on the local repo and doesn't relate to anything touched by this pr

ArchSav:
Merge conflicts fixed

ArchSav:
Tests failed at the same place as the 'fix includes' commit. There isn't anything I can do for this

vaxerski:
yes thats good that just happens

Aqa-Ib:
crash when opening firefox:
[hyprlandCrashReport703.txt](https://github.com/user-attachments/files/24275601/hyprlandCrashReport703.txt)

ryand56:
> crash when opening firefox: [hyprlandCrashReport703.txt](https://github.com/user-attachments/files/24275601/hyprlandCrashReport703.txt)

Can reproduce this.

ItsOhen:
```
diff --git a/src/protocols/XDGShell.cpp b/src/protocols/XDGShell.cpp
index ebb56342..12698827 100644
--- a/src/protocols/XDGShell.cpp
+++ b/src/protocols/XDGShell.cpp
@@ -257,11 +257,12 @@ CXDGToplevelResource::CXDGToplevelResource(SP<CXdgToplevel> resource_, SP<CXDGSu
 
         m_parent = newp;
 
-        if (m_parent)
+        if (m_parent) {
             m_parent->m_children.emplace_back(m_self);
 
-        if (m_parent->m_window->m_pinned)
-            m_self->m_window->m_pinned = true;
+            if (m_parent->m_window->m_pinned)
+                m_self->m_window->m_pinned = true;
+        }
 
         LOGM(Log::DEBUG, "Toplevel {:x} sets parent to {:x}{}", (uintptr_t)this, (uintptr_t)newp.get(), (oldParent ? std::format(" (was {:x})", (uintptr_t)oldParent.get()) : ""));
     });
 ```

tannerellen:
Confirmed patch fixes both firefox and thunderbird crash on launch.

LionHeartP:
I was crashing when launching Lutris, but the above patch fixed it for me as well.

PR #12256, opened 2025-11-10
Fix fractional scale artifact

bea4dev:
Chrome/Electron running Wayland-native draw a high-resolution buffer for fractional scale (Chrome.[devicePixelRatio](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio)=settings scale factor) and then request the compositor to shrink it via wp_viewporter’s destination, but Hyprland was still forcing nearest-neighbor sampling, so the downscaling duplicated individual pixel rows and produced permanent horizontal lines.
Viewport-destination surfaces (i.e. clients that already handle the scaling themselves) don’t need this fallback, so we exclude them from MISALIGNEDFSV1 and keep linear filtering for those windows.
It now renders sharply and correctly.

- before

screenshot-20251111-001918

- after

screenshot-20251111-002031

no

ready

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -133,6 +133,9 @@
         "hyprutils": [
           "hyprutils"
         ],
+        "hyprwayland-scanner": [
+          "hyprwayland-scanner"
+        ],
         "nixpkgs": [
           "nixpkgs"
         ],
@@ -141,11 +144,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1762465111,
-        "narHash": "sha256-dS13YZdWjgGGLBjpT4FHB6xf8I/WiAU+mgNWXsZgDUs=",
+        "lastModified": 1762755186,
+        "narHash": "sha256-ZjjETUHtoEhVN7JI1Cbt3p/KcXpK8ZQaPHx7UkG1OgA=",
         "owner": "hyprwm",
         "repo": "hyprland-guiutils",
-        "rev": "a415eba866a953f3096d661318f771aa0082eb98",
+        "rev": "66356e20a8ed348aa49c1b9ceace786e224225b3",
         "type": "github"
       },
       "original": {
@@ -221,7 +224,10 @@
           "hyprland-guiutils",
           "hyprutils"
         ],
-        "hyprwayland-scanner": "hyprwayland-scanner",
+        "hyprwayland-scanner": [
+          "hyprland-guiutils",
+          "hyprwayland-scanner"
+        ],
         "nixpkgs": [
           "hyprland-guiutils",
           "nixpkgs"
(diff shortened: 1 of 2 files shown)


src/render/pass/SurfacePassElement.cpp
```diff
@@ -73,7 +73,8 @@ void CSurfacePassElement::draw(const CRegion& damage) {
     }
 
     const bool MISALIGNEDFSV1 = std::floor(m_data.pMonitor->m_scale) != m_data.pMonitor->m_scale /* Fractional */ && m_data.surface->m_current.scale == 1 /* fs protocol */ &&
-        windowBox.size() != m_data.surface->m_current.bufferSize /* misaligned */ && DELTALESSTHAN(windowBox.width, m_data.surface->m_current.bufferSize.x, 3) &&
+        !m_data.surface->m_current.viewport.hasDestination /* not fractional viewport */ && windowBox.size() != m_data.surface->m_current.bufferSize /* misaligned */ &&
```
vaxerski: misaligned fs v1 = Misaligned fractional_scale_v1 surface.

Andd... what you did was make this value false when the client is using fractional_scale_v1...?

bea4dev: The answer would be “no,” but
as I understand it, hasDestination becomes false when fractional_scale_v1 is being used—is that understanding incorrect?
This additional check is intended to make the misalignment logic run only when fractional_scale_v1 is in use.

vaxerski: > as I understand it, hasDestination becomes false when fractional_scale_v1 is being used—is that understanding incorrect?

The opposite.

when fs_v1 is used, wl_surface.scale is 1, and wp_viewport.set_destination is called with the surface size.

bea4dev: It sounds like we should assume the cause lies elsewhere.
After doing some additional investigation, I found that, strangely, this issue occurs with misalign being 0, so it doesn’t seem to be detected correctly.

vaxerski: I don't see any issues myself so I won't be of much help here

bea4dev: I’ll keep investigating this on my side, but let me share the reproduction steps just in case.

You can reproduce it as follows:

1. Set scale = 1.8
2. Launch Chrome as a native Wayland client
3. Display the HTML below

With these steps, a line should appear around the center of the screen.

```html
<!DOCTYPE html>
<html lang="ja">

<head>
    <meta charset="UTF-8">
    <title>stripe</title>
    <style>
        html,
        body {
            margin: 0;
            height: 100%;
        }

        body {
            background-image: repeating-linear-gradient(45deg,
                    #000 0px,
                    #000 1px,
                    #fff 1px,
                    #fff 2px);
        }
    </style>
</head>

<body>
</body>

</html>
```
image

bea4dev:
I just noticed that this artifact doesn’t seem to occur when the window is in pseudo mode.
I’ll take a look around the pseudo-related logic.

PR #12258, opened 2025-11-10
window: skip window rule updates on hidden windows

vaxerski:
Not needed, and can cause weird transitions where we don't do any anims on changing windows but props can animate


vaxerski:
incorporated into #12269

PR #12260, opened 2025-11-10
fix noscreenshare layerrule popups

usering-around:
Hides popups when you use `noscreenshare` layerrule.
before:
before
after:
after
(waybar with nm-applet)

No

Ready


vaxerski (approved):
thanks!

PR #12269, opened 2025-11-11
windowrules: rewrite completely

vaxerski:
Massive WIP rewrite that changes the syntax as well as improves the efficiency and code cleanliness under the hood.

Wiki MR: https://github.com/hyprwm/hyprland-wiki/pull/1278

New rules:

```ini
windowrule {
    name = rule-1
    
    # properties (all)
    match:focus = true

    # effects (all)
    rounding = 10
}
```

Disabling is done via the `enabled` variable inside a windowrule or layerrule block.

`hyprctl keyword 'windowrule[rule-1]:enable false'`

!!! Some variables changed name, e.g. `noanim` -> `no_anim`. See:
https://github.com/hyprwm/Hyprland/pull/12269/files#diff-594b30f2e8731d5ccef8d0cdb5155c391617e4387c20ead1599473531d30fd5aR10-R61

and 

https://github.com/hyprwm/Hyprland/pull/12269/files#diff-f992e55e87e530fbef77634570043795c800ca9ccbf277f2637da8577f4c50b1R9-R18

and props:

https://github.com/hyprwm/Hyprland/pull/12269/files#diff-8550eec13e6031fa4a9290fea355642606dcfe94affba2ac716a8ac9f3578c05R15-R32

!!!! `workspace` and `monitor` rules require a patch in hyprlang which is now in git but not in the latest release!!!

Old syntax is gone, single-line syntax for anonymous rules is now:
```
windowrule = match:class kitty, match:fullscreen false, border_size 0, no_shadow true
```

New stuff:
- `enable` for windowrules for toggling via ipc instead of unset
- exec rules follow forks now
- general:no_border_on_floating is gone
- no_border windowrule is gone (border_size = 0)
- static rules use applied tags and content types
- move and size rules now accept expressions

TODO:
- [x] Layer rules
- [x] Setprop/getprop
- [x] Tags
- [x] Exec rules
- [x] Unsetting
- [x] Static rules for windows
- [x] Dynamic layout changes dont always work (focus = 1 + border_size = 10)
- [x] Dynamic calls via hyprctl
- [x] testing
- [x] tests
- [x] rebase on main

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -26,6 +26,7 @@
   libxkbcommon,
   libuuid,
   libgbm,
+  muparser,
   pango,
   pciutils,
   re2,
@@ -148,6 +149,7 @@ in
           libuuid
           libxkbcommon
           libgbm
+          muparser
           pango
           pciutils
           re2


vaxerski:
This is ready for testing unless you use exec rules / setprop / getprop.

src/desktop/rule/layerRule/LayerRule.cpp
```diff
@@ -0,0 +1,76 @@
...
+    {LAYER_RULE_EFFECT_NO_ANIM, "no_anim"},           //
+    {LAYER_RULE_EFFECT_BLUR, "blur"},                 //
+    {LAYER_RULE_EFFECT_BLUR_POPUPS, "blur_popups"},   //
+    {LAYER_RULE_EFFECT_IGNORE_ALPHA, "ignore_alpha"}, //
+    {LAYER_RULE_EFFECT_DIM_AROUND, "dim_around"},     //
+    {LAYER_RULE_EFFECT_XRAY, "xray"},                 //
+    {LAYER_RULE_EFFECT_ANIMATION, "animation"},       //
+    {LAYER_RULE_EFFECT_ORDER, "order"},               //
+    {LAYER_RULE_EFFECT_ABOVE_LOCK, "above_lock"},     //
+    {LAYER_RULE_EFFECT_NO_SCREEN_SHARE, "noanim"},    //
```
VirtCode: typo?

vaxerski: typo.

vaxerski:
on that note @VirtCode IIRC you did abovelock? It's an int now, can you check if I didnt fuck up something? 0 should be "no", 1 should be "visible" and 2 "interactable"

VirtCode:
sure, I'll test tomorrow if that's fine (have to finish an assignment for uni today)

vaxerski:
no prob, this is a big mr will need a bunch of people to test

NicolasDorier:
Does it mean that every windowrule ever written would have to be rewritten in the new syntax?
What about people who update hyprland and forgot to update their dotfiles before?

vaxerski:
wouldnt be the first time we break their configs, we do it every other update. This one is a bigger break, but it's still just a syntax change like many others.

Dregu:
> wouldnt be the first time we break their configs, we do it every other update. This one is a bigger break, but it's still just a syntax change like many others.

Can't the old keyword handler and the new specialcategory coexist with the same name, at least for some transition period, so nobody has a good reason to complain when you finally nuke it in a few months? They'll complain anyway, but at least you warned.

vaxerski:
it's messy, I'd rather nuke it.

vaxerski:
should I rewrite workspace rules while we're at it to use the same syntax?

Dregu:
There are a bunch of collisions between props and effects names now, and generally it's impossible to know what is what if they're just mixed together with the same syntax. Then again putting them both in their own blocks would just make even the simplest of rules 10 lines. Also rule name should be optional if it isn't already.

Still, I was pretty happy when I got to squeeze my hundred window rules to a few dozen lines after the multi rule pr, but looks like it's going back to several hundred lines again and I would really like to keep the old keyword syntax as an alternative, for simple things. Just nuke most of it and pass to this new system. Same story for workspace and layer rules pretty much.

```ini
windowrule {
  name = taggart
  match {
    tag = asdf
  }
  fullscreen = 1
}

# same as

windowrule {
  # name omitted, don't care about easy access to rules
  match:tag = asdf
  fullscreen = 1
}

# same as

windowrule = tag:asdf, fullscreen 1

# same as

windowrule = fullscreen 1, tag:asdf
```

vaxerski:
what are the collisions?

Dregu:
> what are the collisions?

At least tag, content, fullscreen. Might not be everything.

Dregu:
Speaking of tag and content, does this address the uselessness of those two for setting static rules, i.e. since tag and content can be set by static rules, they should also be able to recursively trigger other static rules where they are used as conditions. The pr is kinda tldr to figure that out.

vaxerski:
I can make the props be a `match:`, I dont mind

two syntaxes will not live at once no

vaxerski:
> since tag and content can be set by static rules, they should also be able to recursively trigger other static rules where they are used as conditions

this can be done. Btw, tag is dynamic now.

vaxerski:
it's now done

vaxerski:
names arent optional, that'd require a new hyprlang api

VirtCode:
so far I tried out the layer rules and quite some things are broken on my end:
- blur related rules don't seem to do anything (that is `blur` and `xray`, thus can't test ignorealpha)
- `animation` always errors with "no such animation" (for example with `animation = slide` in my config)
- `noscreenshare` works but crashes when the layer closes during share (here's a debug bt [noscreenshare-bt.txt](https://github.com/user-attachments/files/23509910/noscreenshare-bt.txt))
- the `abovelock` does not do anything, but I will have a look at that later
- `dimaround` and `noanim` work as intended wooo

vaxerski:
thank you virt, most of these not working was a typo. Check now. I'll check the crash and animation bug.

Edit: can't repro either after the fix.

Ah, I see the crash. I'll fix it in main. It's not from this MR.

VirtCode:
awesome, can confirm `blur`, `xray` and `ignorealpha` work as expected now. `abovelock` also mostly works, see below. However, I am still getting "no such animation" for layerrules like `animation = slide` or `animation = popin`, am I missing something?

I can also have a look at the windowrules some time but not today as it is quite late already

> Ah, I see the crash. I'll fix it in main. It's not from this MR.

oh okay then not really related, but thanks for the fix

src/Compositor.cpp
```diff
@@ -1333,7 +1333,7 @@ SP<CWLSurfaceResource> CCompositor::vectorToLayerSurface(const Vector2D& pos, st
 
     for (auto const& ls : *layerSurfaces | std::views::reverse) {
         if (!ls->m_mapped || ls->m_fadingOut || !ls->m_layerSurface || (ls->m_layerSurface && !ls->m_layerSurface->m_surface->m_mapped) || ls->m_alpha->value() == 0.f ||
-            (aboveLockscreen && (!ls->m_aboveLockscreen || !ls->m_aboveLockscreenInteractable)))
+            (aboveLockscreen && !ls->m_ruleApplicator->aboveLock().valueOrDefault()))
```
VirtCode:
```suggestion
            (aboveLockscreen && ls->m_ruleApplicator->aboveLock().valueOrDefault() != 2))
```

we should only focus layers if they are interractible (i.e. abovelock = 2), so this skips layers that are not interractible

vaxerski: thanks!

vaxerski:
> However, I am still getting "no such animation" for layerrules like animation = slide or animation = popin, am I missing something?

Can you show the layerrule? I tested slide and it worked?

VirtCode:
i just have:
```
layerrule {
    name = wofi
    match:namespace = wofi

    animation = slide
}
```

can also reproduce it by inserting this into the default config

vaxerski:
works on my end. Mayhaps a duplicated name?

NicolasDorier:
I don't want to be a party pooper, but I liked the one line rules more (when you need a single effect). If they could still be supported, not necessarily for backward compat, but for conciseness, it would be great!

VirtCode:
> works on my end. Mayhaps a duplicated name?

okay I have figured it out. For some reason pacman has messed up my hyprlang install so it was loading a slightly older version, not yet containing https://github.com/hyprwm/hyprlang/commit/995db114b85101c7b0920aabe7d24b01db4ef768 so it was calling the keyword handler instead. Works as expected now, sorry for the noise.

VirtCode:
I have ported my windowrules to the new syntax and so far everything seems to work. I have to admit though that the new syntax is very verbose, it has inflated my 65 line old config (without multiple rules per line) to about 200 lines which is also significantly less readable (as an example here's my [before](https://github.com/user-attachments/files/23521742/rules-before.txt) and [after](https://github.com/user-attachments/files/23521740/rules-after.txt)). But I do see that the code is _a lot cleaner_ so this refactor is clearly the right direction. Anyways, here's a couple of other odd things I noticed:

- When editing my config, some rules are not updated immediately. If for example have the following which should apply opacity to all windows (not fullscreen):
  ```
  windowrule {
      name = test-3
      match:fullscreen = false
  
      opacity = 0.5
  }
  ```

  However, this only a starts affecting existing windows if the property changes, so only after I fullscreen and un-fullscreen a window. It would maybe be nice to update all rules on config reload.

- The `noborder` rule was nuked right? I have replaced it with `border_size = 0` but I don't know whether that had exactly the same effect (mainly performance wise).

- Additionally, the priorities of rules seems a bit weird to me. Consider the following snippet:
  ```
  windowrule {
      name = test-class
      match:class = ^Alacritty$
  
      border_size = 0
  }
  
  windowrule {
      name = test-fs
      match:fullscreen = false
  
      border_size = 10
  }
  ```
  For some reason the `test-class` rule always has priority over the `test-fullscreen` rule, no matter in which order they are defined. Additionally, if I open a new terminal I can see the border briefly being 10px before being removed which is pretty ugly. What's the expected behaviour here? It would be nice to be able to specify a priority value or something similar per window rule.

vaxerski:
expected behavior would be top to bottom, I don't know what happened here. I'll check.

vaxerski:
Fixed, and added a test.

vaxerski:
I've also caved in and added the single-line syntax back. It's not the same as the old one, though.

vaxerski:
@fufexan bump nix hyprlang please

fufexan:
It's up to date. `flake.lock`:
```json
        "repo": "hyprlang",
        "rev": "771e915f5946bfb47ee9687067d5c1f2b10cf84a",
        "type": "github"
```

vaxerski:
oh, it fails tests

edit: wait no wtf, I dont think it is cuz it errs out in the tests wtf fuf

edit: nah I'm dum

vaxerski:
this MR is catching a bunch of hyprlang edge cases lole xD can we get a hyprlang bump @fufexan again

vaxerski:
ready for review please lmk of any bugs you encounter, otherwise leave a :+1: under this comment

src/config/ConfigManager.cpp
```diff
@@ -3228,6 +2823,80 @@ std::optional<std::string> CConfigManager::handleGesture(const std::string& comm
...
+        // split on space, no need for a CVarList here
+        size_t spacePos = el.find(' ');
+        if (!spacePos)
+            return std::format("invalid field {}: missing a value", el);
+
+        const bool FIRST_IS_PROP = el.starts_with("match:");
+        const auto FIRST         = FIRST_IS_PROP ? el.substr(6, spacePos - 6) : el.substr(0, spacePos);
+        if (FIRST_IS_PROP && std::ranges::contains(PROPS, std::string{"match:"} + FIRST)) {
+            // it's a prop
+            const auto PROP = Desktop::Rule::matchPropFromString(std::string{"match:"} + FIRST);
```
VirtCode:
```suggestion
            const auto PROP = Desktop::Rule::matchPropFromString(FIRST);
```

probably a typo?
layerrules with the new syntax always error with "invalid field type namespace" but works if I change it like this
(also changed it in the if above, but don't know how to suggest multilline changes lol)

vaxerski: mashallah it should be `match:namespace` not `namespace` because you match it

VirtCode: yes I know, but I have
```
layerrule = match:namespace wofi, blur true
```
and I get "invalid field type namespace"

I thought this might be it because for the windows the code doesn't prepend `match:` again there, and this change does indeed fix it on my end

vaxerski: ah yes mb

VirtCode: you'll have to change it in the if just above too though

vaxerski: maybe

aphelei:
rule is not respected if multiples are applicable

before :

```
windowrule = float, size 65% 80%,		class:org.keepassxc.KeePassXC
windowrule = size 15% 25%,			class:org.keepassxc.KeePassXC,title:KeePassXC - Browser Access Request
```

after :

```
windowrule {
    name = keepassxc
    match:class = org.keepassxc.KeePassXC
    float = true
    size = 65% 85%
}
windowrule {
    name = keepassxc_popup
    match:class = org.keepassxc.KeePassXC
    match:title = KeePassXC - Browser Access Request
    size = 15% 25%
}
```

the popup is the same size as the main window.

aphelei:
stay_focused doesn't work

```
windowrule = float, stay_focused, 		match:class org.kde.okular
windowrule {
    name = okular
    match:class = org.kde.okular
    match:title = .*note.*
    stay_focused = true
}
```

image

aphelei:
also

```
windowrule = rounding 0, border_size 0, 	match:fullscreen_state_client 1
```

image

Ciel-MC:
How to do you make a window go to a specific workspace? I was doing `workspace special silent` before but it's throwing an error now. I have the git version of hyprlang

Edit: Nevermind I was doing this thing called forgetting to switch branch. But now I am getting a crash on the branch, something about monitor layout... see attached. I also got a crash with something about window rules but it stopped after I commented them all out(some were in v3 some in v2 format).

[hyprlandCrashReport4198.txt](https://github.com/user-attachments/files/23542738/hyprlandCrashReport4198.txt)

Ciel-MC:
Okay I am majory confused. On the exact same commit. If I have window rules in my config(gonna debug which one exactly), Hyprland crashes. But if I remove those window rule times, only the release build crashes, but not debug build... What the hell is going on

Edit:
```
windowrule = match:class discord, workspace special
windowrule = match:class Cider, workspace special
```
Found the culprit, for some reason either of these two lines would make Hyprland crash instantly. I'll use the debug build for now since release crashes after a black screen

VirtCode:
I get some crashes when I have new syntax window rules in my config, **but only on startup**. So if I have the following line in the config during startup, it crashes, but it doesn't when I add it during runtime.
```
windowrule = match:fullscreen true, rounding 1
```

here's a debug bt [startup-bt.txt](https://github.com/user-attachments/files/23544192/startup-bt.txt)

VirtCode:
also noticed some inconsistencies when boolean props are specified without arguments. Both of the following windowrules are parsed, but have different effects
```ini
# applies border size if window is focused -- as expected
windowrule = match:focus true, border_size 10

# applies border size only if window is NOT focused??
windowrule = match:focus, border_size 10
```
I would have expected the second rule also to only apply if the window is focused. If the arg-less syntax is not intended to be used, it should probably result in a config error

Dregu:
About the naming of props, since we're changing everything anyway, maybe they should all be in a base form, i.e. match:`float, pin, focus`, or maybe `floating, pinned, focused`, but definitely not `floating, pinned, focus` cause I really don't want to remember which one was the odd one.

Edit: There are others too like group vs grouped, just an example.

vaxerski:
@Dregu good idea, done.

vaxerski:
@VirtCode fixed

vaxerski:
@aphelei 2 and 3 fixed, 1 cant repro

vaxerski:
anything else?

aphelei:
i did `hyprpm update` and when it finished hyprland crashed. when I launch hyprland and I spawn a window it crashes immediatly

VirtCode:
The updating on config reload is still a bit weird. It only seems to recheck props for windows which are on the current workspace on a config reload.

For example:
```ini
windowrule = match:fullscreen false, border_size 10
```

Immediately changes the border size for (non-fullscreen) windows on the current workspace, but (non-fullscreen, duh) windows on other workspaces remain unchanged until I toggle fullscreen for them.

aphelei:
layers rules doesn't work or i'm doing something wrong

```
### Layers
layerrule = blur, match:namespace panel
layerrule = above_lock true, match:namespace panel
layerrule = blur, match:namespace launcher
layerrule = ignore_alpha 0, match:namespace launcher
layerrule = blur, match:namespace notifications
layerrule = ignore_alpha 0, match:namespace notifications
layerrule = above_lock, match:namespace notifications
```

aphelei:
```windowrule = plugin:hyprbars:nobar true, match:floating false
windowrule = plugin:hyprbars:nobar true, match:title .*Picture[- ]in[- ]Picture.*
windowrule = plugin:hyprbars:nobar true, match:class hyprpolkitagent
windowrule = plugin:hyprbars:bar_color $dark_blue_rgba, match:focus false
windowrule = nofocus, noblur, noshadow, noborder, pin, move 100%-w-0 100%-w-0, plugin:hyprbars:nobar, match:title conf_script_companion
```
image

before, it wouldn’t give errors if the plugins weren’t updated for the current Hyprland build. It would get annoying if it gives an error every time I update but don’t want to update my plugins.

Ciel-MC:
Also on the new branch, I'm noticing that windows don't actually full screen if it requests to be. Like it fills up the window but doesn't fill up the screen.

aphelei:
```
windowrule = animation gnomed, match:float true
windowrule = animation gnomed, workspace w[t1]
```

doesn't work


```
animations {
    bezier = easeOutQuint,0.23,1,0.32,1
    bezier = easeInOutCubic,0.65,0.05,0.36,1
    bezier = linear,0,0,1,1
    bezier = almostLinear,0.5,0.5,0.75,1.0
    bezier = quick,0.15,0,0.1,1

    animation = global, 1, 10, linear
    animation = windowsIn, 1, 4.1, easeOutQuint, slide
    animation = windowsOut, 1, 4.1, easeOutQuint, slide
    animation = windowsMove, 1, 4.8, easeOutQuint

    # animation = layersIn, 1, 4, easeOutQuint, fade
    # animation = layersOut, 1, 1.5, linear, fade

    # animation = fade, 1, 3.03, quick
    animation = fadeIn, 1, 1.7, almostLinear
    animation = fadeOut, 1, 2, almostLinear
    animation = fadeSwitch, 1, 1.7, almostLinear
    animation = fadeShadow, 1, 1.7, almostLinear
    animation = fadeDim, 1, 1.73, almostLinear
    animation = fadeLayers, 1, 1.8, linear
    animation = fadePopups, 1, 2, linear
# fadeLayers for background must be high.
# fadeOut for windows must be high. but when only 1 window it can be low
    animation = border, 1, 10, easeOutQuint

    animation = workspaces, 0, 2, almostLinear, slidevert
    animation = specialWorkspace, 1, 3, quick, fade
    animation = zoomFactor, 1, 7, quick
}
```

here's my anim config

VirtCode:
> layers rules doesn't work or i'm doing something wrong

@aphelei I think all effects need an argument, so `blur` -> `blur true` in your case. Maybe we should also error if not?

```ini
layerrule = blur true, match:namespace wofi
```

works for me

aphelei:
> > layers rules doesn't work or i'm doing something wrong
> 
> @aphelei I think all effects need an argument, so `blur` -> `blur true` in your case. Maybe we should also error if not?
> 
> ```ini
> layerrule = blur true, match:namespace wofi
> ```
> 
> works for me

```
layerrule = blur true, match:namespace panel
layerrule = above_lock true, match:namespace panel
layerrule = blur true, match:namespace launcher
layerrule = ignore_alpha 0, match:namespace launcher
layerrule = blur true, match:namespace notifications
layerrule = ignore_alpha 0, match:namespace notifications
layerrule = above_lock, match:namespace notifications
```

still no blur :(

edit : there is blur for fuzzel but not for yambar (panel). it was working before this update, why it stopped working ?

```
Monitor eDP-1:
	Layer level 0 (background):
		Layer 561b5b1a4000: xywh: 0 0 1920 1080, namespace: hyprpaper, pid: 3028
	Layer level 1 (bottom):
	Layer level 2 (top):
		Layer 561b5b0ccd20: xywh: 0 0 1920 26, namespace: panel, pid: 8131
	Layer level 3 (overlay):
```

VirtCode:
oh yeah layerrules are not applied on layers which are already open. So you'll have to restart your bar. Wanted to mention that too that should probably be fixed.

aphelei:
restarting the bar did the trick. But it's a regression

another thing : there is a problem with size and move (see vid).

```
# windowrule = float, size 20% {{$s1 - $double_border_size}}, move 100%-w-$floating_border_size {{$double_bar_height + $floating_border_size}}, class:org.speedcrunch.speedcrunch
windowrule = float true, size 100% 10%, move 0 100, match:class org.speedcrunch.speedcrunch
```


https://github.com/user-attachments/assets/35e01319-320e-4904-8f72-e2fa9f01401a

aphelei:
```windowrule = workspace +0, 			workspace special:thunar, match:class negative:[Tt]hunar```

doesn't work. I used this to open my file manager in a special workspace and open everything in the workspace below. Now it opens in the special workspace.

ok im dumb
```windowrule = workspace +0, match:workspace name:special:thunar, match:class negative:Thunar```
works

**Comment by vaxerski** [comment]
right, plugin window rules will need to be fixed...

**Comment by vaxerski** [comment]
I cant repro the fullscreen issue

**Comment by vaxerski** [comment]
I've done a thing that should allow plugins to easily add window rules again but it needs testing please check if nothing broke, I'm gonna go to a lidl and write tests when abck

**Comment by vaxerski** [comment]
ye seems to work, plugins will need an update tho ye

**Comment by vaxerski** [comment]
hows it now? any issues?

**Comment by aphelei** [comment]
will try

**Comment by VirtCode** [comment]
fullscreening is still not working for me, didn't compare to main though

<details>
<summary>here's a video</summary>

https://github.com/user-attachments/assets/a056e657-39ee-4275-84db-d9eefb359925

</details>

otherwise, all issues I had were fixed and I couldn't find any new ones so far

**Comment by aphelei** [comment]
1°

```
windowrule = animation gnomed, match:float true # doesn't work
windowrule = animation gnomed, match:workspace w[t1] # works for tiled windows and opening floating windows. not for closing floating windows
```


https://github.com/user-attachments/assets/a9519a21-faa0-489f-82a5-b7121d557da2


2°

https://github.com/hyprwm/Hyprland/pull/12269#issuecomment-3533152503 still happens

3°

it's still crashing when using hyprbars idk if you've done something about this

[hyprlandCrashReport24771.txt](https://github.com/user-attachments/files/23555235/hyprlandCrashReport24771.txt)

**Comment by vaxerski** [comment]
astaghfirullah

**Comment by vaxerski** [comment]
I dont get your move size bug the line you show is too complicated

**Comment by vaxerski** [comment]
animation and fs fixed. Crash report unusable. @aphelei nr 2 I dotn undersat

**Comment by aphelei** [comment]
size : 
x does nothing. y also change x

https://github.com/user-attachments/assets/d468d595-02e9-4f17-8942-5eef24a2d768

move : 
same thing

https://github.com/user-attachments/assets/df32111b-f69b-4bcb-8899-6db59b656646

**Comment by vaxerski** [comment]
you know what? I'll take the opportunity to also nuke these positions and just impl an arithmetic parser.

**Comment by alper-han** [comment]
Currently, the noscreenshare layer covers more area than necessary
It would be useful to allow setting xywh manually so users can precisely control the hidden region

```
	Layer level 0 (background):
		Layer 559075b62810: xywh: 2560 0 2560 1440, namespace: swww-daemon, pid: 3136
	Layer level 1 (bottom):
		Layer 559074e8c1f0: xywh: 2560 0 2560 40, namespace: waybar, pid: 3108
	Layer level 2 (top):
		Layer 559074cf8990: xywh: 2560 40 2560 1400, namespace: swaync-control-center, pid: 9663
	Layer level 3 (overlay):
```
swaync-control-center...


If it's not too much trouble, can we add this too?

**Comment by vaxerski** [comment]
> Currently, the noscreenshare layer covers more area than necessary

? it covers the surface, as you asked.

**Comment by vaxerski** [comment]
expressions added via muparser. I'll add it to wiki docs in a sec.

**Comment by vaxerski** [comment]
expression docs pushed.

**Comment by vaxerski** [comment]
@fufexan arch and nix need muparser

**Comment by alper-han** [comment]
> > Currently, the noscreenshare layer covers more area than necessary
> 
> ? it covers the surface, as you asked.

resim

"noscreenshare, swaync-control-center"

resim

Can't we set this manually?

**Comment by vaxerski** [comment]
@alper-han it's not my fault swaync decides to open a fullscreen surface. Setting this manually is just asking for trouble when some font changes and suddenly its bigger

**Comment by vaxerski** [comment]
@aphelei lmk if your layout troubles are fixed

**Comment by VirtCode** [comment]
can confirm the fs issue is fixed

however, just found that updating the `float` prop is broken. If I add the following to my config on runtime: 
```ini
windowrule = match:float true, border_size 10
```
it is only applied to floating windows **which are already floating** and not to new floating windows. in general, changing the float state of a window doesn't seem to update de-/apply the rule (also with `match:float false`).

fedy-cz:
> I have to admit though that the new syntax is very verbose, it has inflated my 65 line old config (without multiple rules per line) to about 200 lines which is also significantly less readable (as an example here's my [before](https://github.com/user-attachments/files/23521742/rules-before.txt) and [after](https://github.com/user-attachments/files/23521740/rules-after.txt)).

### Regarding the verbosity of the new syntax:
Any chance that hyprlang  could get a "command" separator like `;` in addition to  newlines? That would enable a much terser but readable config. Putting everything on one line might be to much (yet still possible if you prefer it). But I would see the optimal syntax/formatting to be something like:
```
windowrule {
    name = test-rule
    match { class = ^Alacritty$ ; title = ^VIM  TODO.txt$ }
    border_size = 0; float; pin; monitor = 2; move = 50 50
}
```
Features:
- multiple "commands" on a single line separated by `;`
- option to enable binary flags by just mentioning their name (`float` would be equivalent to `float = true`)
- and maybe also `no<keyword>` as a shorter form of `<keyword> = false` eg. `nofloat`

### Regarding backwards compatibility:
Maybe we could make a conversion tool to migrate conf  between version? `hyprconf` ?

BTW. Is there a formal grammar for hyprlang available?

vaxerski:
@VirtCode check now

vaxerski:
@fedy-cz no. Use the single-line syntax for terser windowrules.

> BTW. Is there a formal grammar for hyprlang available?

RTFM: https://wiki.hypr.land/Hypr-Ecosystem/hyprlang/

fedy-cz:
> > BTW. Is there a formal grammar for hyprlang available?
> 
> RTFM: https://wiki.hypr.land/Hypr-Ecosystem/hyprlang/

I read the wiki. I was specifically asking for a **_formal_** grammar (to build a lexer/parser with) to be used if I was to attempt some conversion tool / script. 
For example: I noticed that white-space is sometimes meaningful (yet it might be just because I'm still on 0.51).

I could also imagine 3rd party alternate config syntax that gets "compiled" into something hyprland expects.

VirtCode:
> @VirtCode check now

float works as expected now, thanks!

vaxerski:
@fedy-cz in that case there is no formal doc. It's worth noting that hyprlang is merely a k-v store for the most part. Parsing whatever is after the `=` is the job of the app, not hyprlang.

E.g.:

Hyprland creates a `general:hello` with type STRING, hyprlang gets `general:hello = a, b, c`. Hyprland gets `a, b, c` from hyprlang. How that abc is parsed is down to hyprland.

(There are keywords, which aren't k-v but work in almost the same way, the string is just passed to hyprland)

non-bin:
`no_border_on_floating` is still in the docs, and 2 examples use `match:floating` instead of `match:float`

Love the changes though, it's not actually clear whats an effect and what's a prop

vaxerski:
I wanted to ask @tagattie since you now maintain hl on BSD, IIRC bsd doesn't have procfs, is there any way to gather another process' environ on BSDs? Exec rules will no longer work without being able to access environ by PID. See `CWindow::getEnv`

tagattie:
Hi @vaxerski,

> IIRC bsd doesn't have procfs, is there any way to gather another process' environ on BSDs?

Thanks for asking. That's correct. Although BSDs have procfs, it cannot be used to access other processes' environ. Instead, BSDs provide a `sysctl(3)` function for that purpose.

I have attached a small patch file to this message for your review. If a pull request is preferable, I would be happy to make one. Please let me know.

[Window.cpp.patch](https://github.com/user-attachments/files/23736792/Window.cpp.patch)

vaxerski:
I prefer a PR if you could. Cheers and thanks!

tagattie:
Created a PR at https://github.com/hyprwm/Hyprland/pull/12462.

non-bin:
> I wanted to ask @tagattie since you now maintain hl on BSD, IIRC bsd doesn't have procfs, is there any way to gather another process' environ on BSDs? Exec rules will no longer work without being able to access environ by PID. See `CWindow::getEnv`

Could this be causing my issue in #12434? (exec ignores window rules on NixOS)

vaxerski:
unlikely because nix has a procfs? are you able to `cat /proc/$(pidof Hyprland)/environ`?

non-bin:
> unlikely because nix has a procfs? are you able to `cat /proc/$(pidof Hyprland)/environ`?

Yes, ok at least that's one less potential cause :)

hikdul:
For now, only comment (# ...) de file on the main file the configuration.. and use my system.. only see de changes here for new changes or the updates de HyDE to use al the configuration....

PR #12270, opened 2025-11-11
cursor-shape-v1 version 2

Honkazel:
Sets cursor-shape protocol version to 2 and adds 2 new shapes

It works, but I have some doubts about xcursor changes, you know, +2 else if checks...

Draft for now because of reason above

### DIFF
--- a/.github/workflows/translation-ai-check.yml
+++ b/.github/workflows/translation-ai-check.yml
@@ -14,24 +14,9 @@ permissions:
   issues: write
 
 jobs:
-  changes:
-    name: Check i18n changes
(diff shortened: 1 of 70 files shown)


src/managers/XCursorManager.cpp
```diff
@@ -396,6 +396,10 @@ std::string CXCursorManager::getLegacyShapeName(std::string const& shape) {
         return "left_ptr";
     else if (shape == "zoom-out")
         return "left_ptr";
+    else if (shape == "dnd-ask")
+        return "dnd-copy";
+    else if (shape == "all-resize")
+        return "dnd-move";
```
vaxerski: wut

Honkazel: What shall I do instead? I didn't really found something else except of this

vaxerski: fleur? iirc?

Honkazel: Both?

vaxerski: no, all resize

Honkazel: good?

PR #12279, opened 2025-11-12
fix(master_layout): Set active window master when moving from left monitor to right one

ghost:

This makes the master layout more consistent when moving a window from left monitor to right one, making the window that was slave to the new master of the new monitor

This solves the #12278 

It is ready for merge. Tested on my own setup


ghost:
This change works as long as the master window is positioned at the left of the screen.

PR #12286, opened 2025-11-12
Enhancement: Optional conf file source (if exists), source-optional

ryrobes:
Addition of a `source-optional` for adding a conf file so an error won't be thrown if it doesn't exist.

Useful in custom theme / style change systems where some might have optional files that others do not - they can all be referenced safely regardless. 

Note: If interested, perhaps `source-if-exists` is a better name?


vaxerski:
why not wildcard

ryrobes:
what do you mean, for the name?

vaxerski:
`source = ./myConfigs/*.conf`

kamil-szatkowski:
I also favor the optional-source (or whatever the final name would be). Making the declaration explicit makes it visible what the config should expect. Also this will allow some third party application to generate part of the config, like the ones that configure monitors - i.e. kanshi or hyprdynamicmonitors

PR #12287, opened 2025-11-12
fix fractional scale artifacts

bea4dev:
I’ve fixed an issue where artifacts could sometimes appear in the center of the screen when using fractional scaling.
I suspect the root cause is that, under fractional scaling, when the window size is not evenly divisible by the scaling factor, the sampling also fails to divide evenly, and the remainder shows up as visible error on the screen.
To mitigate this, the window size is now adjusted so that it becomes a multiple of the fractional part of the scaling factor in pixels. (Sorry for the slightly convoluted explanation.)
With this change, artifacts should no longer appear at any scale, as long as the window size isn’t freely resized to arbitrary values.
I also considered fixing it by adjusting the edge pixels, but I chose this approach instead to avoid conflicts with the existing misalign logic, and because the breakage was very noticeable in the first place.
I’m sorry for opening so many PRs, but I’d appreciate it if you could review this one as well.

You can reproduce the issue as follows.

1. Set scale = 1.8
2. Launch Chrome as a native Wayland client
3. Display the HTML below

With these steps, a line should appear around the center of the screen.

```html
<!DOCTYPE html>
<html lang="ja">

<head>
    <meta charset="UTF-8">
    <title>stripe</title>
    <style>
        html,
        body {
            margin: 0;
            height: 100%;
        }

        body {
            background-image: repeating-linear-gradient(45deg,
                    #000 0px,
                    #000 1px,
                    #fff 1px,
                    #fff 2px);
        }
    </style>
</head>

<body>
</body>

</html>
```

- before

screenshot-20251112-234611

- after

screenshot-20251112-234434

Please see the comment below.

Please see the comment below.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -1133,8 +1133,12 @@ void CHyprRenderer::calculateUVForSurface(PHLWINDOW pWindow, SP<CWLSurfaceResour
         if (projSize != Vector2D{} && fixMisalignedFSV1) {
             // instead of nearest_neighbor (we will repeat / skip)
             // just cut off / expand surface
-            const Vector2D PIXELASUV    = Vector2D{1, 1} / pSurface->m_current.bufferSize;
-            const Vector2D MISALIGNMENT = pSurface->m_current.bufferSize - projSize;
+            const Vector2D PIXELASUV   = Vector2D{1, 1} / pSurface->m_current.bufferSize;
+            const auto&    BUFFER_SIZE = pSurface->m_current.bufferSize;
+
+            // compute MISALIGN from the adjusted UV coordinates.
+            const Vector2D MISALIGNMENT = (uvBR - uvTL) * BUFFER_SIZE - projSize;
+
             if (MISALIGNMENT != Vector2D{})
                 uvBR -= MISALIGNMENT * PIXELASUV;
         } else {


bea4dev:
There is another way to fix this issue: give up on useNearestNeighbor when using fractional scaling.
Given that we’re already adjusting misalign, the result still looks sufficiently sharp even without nearest-neighbor sampling.
Personally, I would recommend this approach.

```cpp
    //if (MISALIGNEDFSV1)
    //    g_pHyprOpenGL->m_renderData.useNearestNeighbor = true;
```

bea4dev:
> Just stretch the surface like we do when undersized?

Of course I considered that as well, but unlike the misalign logic, it can lead to much more noticeable breakage.
To take an extreme example, imagine a scale of 1.0666667, which is 16/15.
You’d end up stretching by up to 7 pixels, which I don’t think is desirable.

I also don’t consider this “adjust the window size” approach ideal; in my view, the cleanest solution is to give up on NearestNeighbor.

Ultimately, I’ll leave the decision up to you, but I’d appreciate it if you could take this proposal into consideration.

vaxerski:
> in my view, the cleanest solution is to give up on NearestNeighbor

How so? It will make the entire surface blurry if we're off by 1. That's why it was introduced in the first place.

vaxerski:
In general I also invite you to push for better fs MRs like this one: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/313

I can't, I'm banned, but you could always try to get people over there to do something good for once.

bea4dev:
I think I’ve mostly understood it now.
This issue runs deeper than I first thought...

As for NearestNeighbor: thanks to the misalign logic, it’s actually rendering very sharply on my side.
I’d be happy if you could take a look.
That said, it might be better to keep this as a last resort.

I just had another idea: what about finding the closest scale factor that renders cleanly and applying that per surface?
(I’m not entirely sure I have the skills to implement this myself, though.)

In any case, I’d like to keep digging into this a bit more.
I’m sorry this PR is still rough around the edges, but I’d really appreciate it if you could bear with me a little longer.

vaxerski:
> I just had another idea: what about finding the closest scale factor that renders cleanly and applying that per surface?
(I’m not entirely sure I have the skills to implement this myself, though.)

on one hand sure sounds like it could work - on the other it'd likely be a performance nightmare during live resizes, some toolkits (most?) re-render everything when scale changes, so not only you spam sizes, you also spam scales. Also probably a code nightmare on our side tbh. Aaaand probably out of spec.

> In any case, I’d like to keep digging into this a bit more.
I’m sorry this PR is still rough around the edges, but I’d really appreciate it if you could bear with me a little longer.

No worries, I'm here to help, it's my job. If you need any help, feel free to ask, I'm here to make hyprland better.

> As for NearestNeighbor: thanks to the misalign logic, it’s actually rendering very sharply on my side.
I’d be happy if you could take a look.

take a look at? This MR still has the size issue I mentioned, I am unsure which way is better... expanding imo is more desirable

bea4dev:
> No worries, I'm here to help, it's my job. If you need any help, feel free to ask, I'm here to make hyprland better.

Thanks!

> take a look at? This MR still has the size issue I mentioned, I am unsure which way is better... expanding imo is more desirable

As for the idea of dropping NearestNeighbor, I’d really like you to try it on your own machine at least once.
At least on my setup it’s working surprisingly well.

As for the “expanding” approach, I’d like to keep trying a few more things on my side.

bea4dev:
I’ve fully understood the root cause.
The cause was almost the same as the one that was merged last time.
I may have taken quite a roundabout path to get here.
Thanks in advance for the review.

@vaxerski

src/render/Renderer.cpp
```diff
@@ -1115,7 +1115,7 @@ void CHyprRenderer::calculateUVForSurface(PHLWINDOW pWindow, SP<CWLSurfaceResour
         Vector2D    uvTL;
         Vector2D    uvBR = Vector2D(1, 1);
 
-        if (pSurface->m_current.viewport.hasSource) {
+        if (pSurface->m_current.viewport.hasSource && !fixMisalignedFSV1) {
```
vaxerski: Ah, I see what's happening.

we can't just ignore the source rect though, I think what we can check is that the source rect has the same misalignment.

Imagine a scenario like this, which does happen with chromium IIRC:

Surface is supposed to be 100x100 at scale 1.5

submit:
- buffer: 200x200
- dest: 100x100
- source: 150x150

This (IIUC) is still within the spec. If you just ignore source you'll show the entire buffer which will be wrong and squish the surface.

Here what happens IIUC is that we essentially apply the misaligned fsv1twice, which is obviously wrong

Sorry for being a bit pedantic here but from experience in this area any small mistake often ends up being a big bug with rendering later

bea4dev:
I’ve changed it to compute MISALIGNMENT from the adjusted UVs.
I’d appreciate your review.

src/render/Renderer.cpp
```diff
@@ -1133,10 +1133,15 @@ void CHyprRenderer::calculateUVForSurface(PHLWINDOW pWindow, SP<CWLSurfaceResour
...
-            const Vector2D PIXELASUV    = Vector2D{1, 1} / pSurface->m_current.bufferSize;
-            const Vector2D MISALIGNMENT = pSurface->m_current.bufferSize - projSize;
-            if (MISALIGNMENT != Vector2D{})
+            const Vector2D PIXELASUV   = Vector2D{1, 1} / pSurface->m_current.bufferSize;
+            const auto&    BUFFER_SIZE = pSurface->m_current.bufferSize;
+
+            // compute MISALIGN from the adjusted UV coordinates.
+            const Vector2D MISALIGNMENT = (uvBR - uvTL) * BUFFER_SIZE - projSize;
+
+            if (MISALIGNMENT != Vector2D{}) {
```
vaxerski: no {} around short ifs

bea4dev:
Fixed.
Thank you so much for sticking with me on this.👍

PR #12310, opened 2025-11-15
Add groupbar blur

jochimvd:

This PR implements a simple config option to enable blur on the groupbar.
The config options is called `group:groupbar:blur` and is a boolean.

When enabled, it applies blur to the group indicator and gradient.
image

I refactored the groupbar rounding logic to fix the visual artifact shown below. Instead of drawing 2 rectangles, I extended the edge that should not be rounded and clipped it at its original size.
image

I think it's ready.


vaxerski (approved):
thanks! wiki mr needed

vaxerski:
nvm there is one my bad

PR #12327, opened 2025-11-16
Fix m_cursorSurfaceInfo not being updated while a cursor override is set

giacomozama:

In `InputManager`, `processMouseRequest` and the `SetShapeEvent` handler return immediately if a cursor override is set. This causes issues when moving between windows while and override is set (such as with resize_on_border), i.e. the new window will still show the last cursor requested by the former window (see [here](https://github.com/hyprwm/Hyprland/discussions/12268#discussioncomment-14979427)) once the override is removed.

We should still update m_cursorSurfaceInfo and only return before calling the `g_pHyprRenderer` method.

I don't think so.

Should be fine.


vaxerski (approved):
thanks!

PR #12338, opened 2025-11-16
CMake/version.h: respect environment variables

fufexan:

Fixes #12306.
Attempts to mimic `generateVersion.sh` by using predefined variables instead of relying only on `git` output.

No.

Seems to work when inspecting the generated `version.h`. @vaxerski thoughts?

CC @CyrenArkade

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,13 +128,41 @@ set(HYPRGRAPHICS_VERSION "${hyprgraphics_dep_VERSION}")
 
 find_package(Git QUIET)
 
-set(GIT_COMMIT_HASH "unknown")
-set(GIT_BRANCH "unknown")
-set(GIT_COMMIT_MESSAGE "unknown")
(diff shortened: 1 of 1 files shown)


vaxerski:
lgtm

nix/default.nix
```diff
@@ -106,11 +106,13 @@ in
...
-      COMMITS = revCount;
-      DATE = date;
-      DIRTY = optionalString (commit == "") "dirty";
-      HASH = commit;
-      TAG = "v${trim (readFile "${finalAttrs.src}/VERSION")}";
+      env = {
+        GIT_COMMITS = revCount;
+        GIT_COMMIT_DATE = date;
+        GIT_COMMIT_HASH = commit;
+        GIT_DIRTY = if (commit == "") then "clean" else "dirty";
```
CyrenArkade: I believe this is backwards?

CyrenArkade: I believe this is backwards.

PR #12343, opened 2025-11-16
Add Malayalam translations - failed

aka-nahal:
Added Malayalam translation :)


aka-nahal:
the formatting violatesd clang-format rules

PR #12355, opened 2025-11-16
Add Pirate English translations to Engine.cpp

ElopingDev:
Added Pirate English translations for application notifications and permissions.

Added Pirate English.


vaxerski:
Not a valid locale, plus will break en_US

PR #12363, opened 2025-11-17
i18n: add Sinhala translations

ShashiSrinath:
Add   si_LK (Sinhala) translations.

Decided to use English loan words where it make sense instead of using directly translated alien words


vaxerski (dismissed):

vaxerski:
clang-format needed, run `clang-format -i src/**/*{cpp,hpp}`

ShashiSrinath:
had a wired issue with clang-format, it keep breaking strings into newlines on each run.
manually creating smaller lines worked.

PR #12047, opened 2025-10-15
feat(dwindle): Rework split logic to be fully gap-aware

crossatko:
### **Description:**

[Related discussion topic](https://github.com/hyprwm/Hyprland/discussions/11581)
[Forum post](https://forum.hypr.land/t/asymmetric-gaps-dwindle-layout-split-behavior/692)

This PR enhances the Dwindle layout to correctly handle highly asymmetrical window gaps, fixing the issues demonstrated in the attached videos.

The previous, purely geometric approach to splitting windows didn't account for gaps. This led to incorrect split directions and invalid window sizes in certain edge cases.

The core layout logic in `onWindowCreatedTiling` and `recalcSizePosRecursive` has now been reworked to be fully "gap-aware." Splits are calculated based on the true available tiling area, resulting in consistent and intuitive behavior as shown in the "after" video.

**Workspace rules in the video example**
`workspace = r[1-1]w[t1-8], gapsout:4 1500 4 4`

#### Before

https://github.com/user-attachments/assets/487eee09-e4e9-4b75-acf9-d12597b815d5

#### After

https://github.com/user-attachments/assets/86e2be27-a04e-4c76-9e1c-e5a054d95f66

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

I have tested this on both asymmetrical and standard configurations without issue. As this is a core logic change, wider community testing is needed to ensure full stability.

#### Is it ready for merging, or does it need work?

That is probably up to discussion.

PR #12383, opened 2025-11-17
Desktop: rewrite reserved area handling + improve tests

vaxerski:
Wip, unfinished. Can't do more today my brain fried

Rewrites reserved area handling, reserved should also have a clear outer gap container for retrieving the work area with outer gaps for layouts.

Supersedes #12047

Test wiki MR https://github.com/hyprwm/hyprland-wiki/pull/1293

todo:
- [x] impl outer gaps
- [ ] test
- [ ] tests

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -225,7 +225,7 @@ static SDispatchResult scroll(std::string in) {
 }
 
 static SDispatchResult keybind(std::string in) {
-    CVarList data(in);
+    CVarList2 data(std::move(in));
     // 0 = release, 1 = press
     bool press;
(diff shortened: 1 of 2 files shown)


vaxerski:
this now works with the original usecase but breaks some tests

vaxerski:
I hate my life

vaxerski:
snap tests still fail but @fufexan I added gtest here can you add it to nix and arch deps and add it to test CI?

vaxerski:
finally jfc

EvilLary:
Assertion fails in `ReservedArea.cpp:18` if the monitor has no preferred size and no resolution is set, an example of this is when running Hyprland inside another graphical session. 

This makes it work tho  `monitor = WAYLAND-1, 1920x1080@60, auto, 1`

vaxerski:
fixed

PR #12399, opened 2025-11-19
Fix previous workspace remembering in workspace gesture and the `movetoworkspace` dispatcher

kolayne:

Fixes a bug that previous workspace does not exist after swiping to a workspace. Enables previous workspace remembering on `movetoworkspace`, regardless of `binds:allow_workspace_cycles`.

No

Probably ready for merge.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1395,10 +1395,9 @@ SDispatchResult CKeybindManager::moveActiveToWorkspace(std::string args) {
         return {.success = false, .error = "Not moving to workspace because it didn't change."};
     }
 
-    auto        pWorkspace            = g_pCompositor->getWorkspaceByID(WORKSPACEID);
-    PHLMONITOR  pMonitor              = nullptr;
-    const auto  POLDWS                = PWINDOW->m_workspace;
-    static auto PALLOWWORKSPACECYCLES = CConfigValue<Hyprlang::INT>("binds:allow_workspace_cycles");
+    auto       pWorkspace = g_pCompositor->getWorkspaceByID(WORKSPACEID);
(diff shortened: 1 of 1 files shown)


vaxerski:
> Or needs a test?

would be nice, shouldnt be hard.

kolayne:
Done

kolayne:
On the topic of saving previous workspace: why does `CKeybindManager::moveActiveToWorkspace` only save previous workspace (lines 1438-1439) if `binds:allow_workspaces_cycles` is enabled? What does it have to do with cycles?

https://github.com/hyprwm/Hyprland/blob/2b0fd417d32278159d0ca1d23fb997588c37995b/src/managers/KeybindManager.cpp#L1385-L1447

vaxerski:
good question. No clue. Didn't write that

vaxerski:
Previous workspace remembering could use a rewrite anyways.

kolayne:
I also just noticed that, e.g., if I have workspaces 1, 2, and 3 and I swipe-switch from 1 to 3 (with a single gesture), during the gesture we first switch to 2, so 3 considers its previous workspace to be 2 (even though I actually came from workspace 1). IMO it would feel more natural if the workspace only switched at the end of the gesture. What do you think?

vaxerski:
probably, yes, but hard to do internally as you probably have seen from how it works.

kolayne:
Alright. I am not sure if I will get to doing it soon, so, if you agree with the changes in this PR, I suggest to merge it.

For the past few days I have also been testing Hyprland with `moveToWorkspace` remembering previous workspace unconditionally, and have not encountered any issues, so we can probably include it as well. Added a commit.

vaxerski:
we can merge this in the meantime

Issue #12397, opened 2025-11-19
Group behavior misses windowrule recalcs with the new system


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/12395

## Description
some group actions don't seem to trigger proper recalcs. See discussion for a repro case.

tekstryder:
Confirmed fixed by https://github.com/hyprwm/Hyprland/pull/12403

PR #12403, opened 2025-11-19
windowrules: fix group rule recalcs

vaxerski:
Fixes https://github.com/hyprwm/Hyprland/issues/12397

cc @tekstryder

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -177,6 +177,18 @@ static void testGroupRules() {
         EXPECT_CONTAINS(str, "0");
     }
 
+    if (!Tests::spawnKitty("kitty_C")) {
+        ret = 1;
+        return;
+    }
+
+    OK(getFromSocket("/dispatch moveoutofgroup r"));
+
+    {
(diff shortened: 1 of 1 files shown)


tekstryder:
Thanks! 

So, this fixed the repro case I provided, but I still see an issue.

If I move a window out of a group, tiled group + tiled window... no borders appear.

I'd mentioned that in the original discussion, but didn't elaborate:

>Things get even wonkier from there. Sometimes borders are missing when multiple tiled windows or groups are on the workspace.

vaxerski:
repro steps?

tekstryder:
1) create a window group of 2 or more windows
2) move any window out of the group
3) there are now 2 tiles but expected borders are missing.


```
windowrule {
    name = no-borders-wtv1
    match:float = false
    match:workspace = w[tv1]

    border_size = 0
}

```

In the case of only 2 tiles (1 group + 1 window), simply ungrouping the singleton group allows borders to show.

So, this is exclusive to a group being at least one tile.

vaxerski:
check now

tekstryder:
Nice work. Fixed.

PR #12436, opened 2025-11-23
Add quirks:prefer_hdr to fix HDR activation for some clients

UjinT34:
Adds `quirks:prefer_hdr`
0 - off (default)
1 - always
2 - gamescope only

When this setting is active CM proto sends HDR image description as a preferred one regardless of the monitor settings.
Fixes `render:cm_auto_hdr` and whitescreen with forced HDR for some clients.
Should fix chromium flickering in fullscreen on some setups.
In most cases changing this setting while a client is running won't affect it right away.

Might not use the best way to detect gamescope.
DS with gamescope HDR might produce glitches not related to this PR.
Might cause unneeded SDR -> HDR -> SDR conversion.
If it doesn't help with chromium verify that it is not caused by https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/37880 "Restore SDR mode again when the swapchain gets destroyed - At least some drivers, e.g., amdgpu, need a full atomic modeset for switching between SDR and HDR modes."

Ready

### DIFF
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -565,7 +565,7 @@ void CWLSurfaceResource::commitState(SSurfaceState& state) {
 
 SImageDescription CWLSurfaceResource::getPreferredImageDescription() {
     static const auto PFORCE_HDR = CConfigValue<Hyprlang::INT>("quirks:prefer_hdr");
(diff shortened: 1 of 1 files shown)


vaxerski:
why quirks and not render

UjinT34:
Cause it's a temporary workaround for some client bugs and shouldn't be used in general. And it seems there'll be more settings in this category.

vaxerski (approved):
meh ok wiki mr needed

UjinT34:
https://github.com/hyprwm/hyprland-wiki/pull/1297

PR #12441, opened 2025-11-24
windowrules: fix persistent_size not applying

EvilLary:

#12435 Fixes persistent_size being parsed an integer instead of bool, and applies stored size if the rule is set

Is there a way to make tests more consistent. Gestures test and Monitor color setting thiny fail and sometimes not

I think yes

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -397,6 +397,34 @@ static bool test() {
     OK(getFromSocket("/reload"));
     Tests::killAllWindows();
 
+    // test persistent_size between floating window launches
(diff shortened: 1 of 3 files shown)


vaxerski:
> Is there a way to make tests more consistent. Gestures test and Monitor color setting thiny fail and sometimes not

idk those are p consistent but they are designed to run on a clean system so some may fail if you have configs for eg kitty

EvilLary:
> > Is there a way to make tests more consistent. Gestures test and Monitor color setting thiny fail and sometimes not
> 
> idk those are p consistent but they are designed to run on a clean system so some may fail if you have configs for eg kitty

That make sense. Ig I need to learn nix-shell or something

PR #12448, opened 2025-11-24
fifo: unlock locksurfaces aswell

gulafaran:
with egl-wayland2 we now get fifo support which means locksurfaces can have fifo, ensure we schedule frame for the monitor and unlock it on presented.

fixes: https://github.com/hyprwm/Hyprland/discussions/12246

### DIFF
--- a/src/protocols/Fifo.cpp
+++ b/src/protocols/Fifo.cpp
@@ -65,20 +65,6 @@ CFifoResource::CFifoResource(UP<CWpFifoV1>&& resource_, SP<CWLSurfaceResource> s
             }
         }
 
-        auto lock = g_pSessionLockManager->isSurfaceSessionLock(m_surface.lock());
-        if (lock) {
-            for (auto& m : g_pCompositor->m_monitors) {
-                if (!m || !m->m_enabled)
-                    continue;
-
-                auto mon = g_pSessionLockManager->getSessionLockSurfaceForMonitor(m->m_id);
-                if (mon) {
-                    g_pCompositor->scheduleFrameForMonitor(m, Aquamarine::IOutput::AQ_SCHEDULE_NEEDS_FRAME);
-                    continue;
-                }
-            }
-        }
-
         // only lock once its mapped.
         if (m_surface->m_mapped)
             m_surface->m_stateQueue.lock(state, LOCK_REASON_FIFO);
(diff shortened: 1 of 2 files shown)


gulafaran:
requires some testing that i havent missed some oddball case tho

tekstryder:
>requires some testing

Fixes the immediate issue of Hyprlock freezing when using `egl-wayland2`. Thanks!

I'll leave the [discussion thread](https://github.com/hyprwm/Hyprland/discussions/12246) open till this lands.

src/protocols/Fifo.cpp
```diff
@@ -65,6 +65,20 @@ CFifoResource::CFifoResource(UP<CWpFifoV1>&& resource_, SP<CWLSurfaceResource> s
             }
         }
 
+        auto lock = g_pSessionLockManager->isSurfaceSessionLock(m_surface.lock());
```
vaxerski: excuse me but why does a lock surface have no m_enteredOutputs?! This is patching a result of a bug and not the bug itself.

src/protocols/Fifo.cpp
```diff
@@ -188,5 +202,14 @@ void CFifoProtocol::onMonitorPresent(PHLMONITOR m) {
                 continue;
             }
         }
+
+        auto lock = g_pSessionLockManager->isSurfaceSessionLock(fifo->m_surface.lock());
+        if (lock) {
+            auto mon = g_pSessionLockManager->getSessionLockSurfaceForMonitor(m->m_id);
```
vaxerski: you are not checking the monitor is the same as m

src/protocols/Fifo.cpp
gulafaran:
```
diff --git a/src/protocols/SessionLock.cpp b/src/protocols/SessionLock.cpp
index 6ff9171e..eec1b813 100644
--- a/src/protocols/SessionLock.cpp
+++ b/src/protocols/SessionLock.cpp
@@ -219,6 +219,7 @@ void CSessionLockProtocol::onGetLockSurface(CExtSessionLockV1* lock, uint32_t id
         return;
     }

+    PSURFACE->enter(PMONITOR);
     sessionLock->m_events.newLockSurface.emit(RESOURCE);
 }
 ```
 
 might make sense then?
 that makes all the other changes obsolete it seems like

vaxerski: yes

src/protocols/Fifo.cpp
gulafaran: changed to entering output and reusing already existing code for this.

src/protocols/Fifo.cpp
gulafaran: ok its now entering the output in CSessionLockSurface constructor instead

PR #12452, opened 2025-11-25
workspace: add position-based monitor selection

cmihail:
Description:

  This PR adds support for position-based monitor selection in workspace rules using the
  `monitor:position:XxY` syntax.

  Relates to https://github.com/hyprwm/Hyprland/discussions/12305

  Monitor IDs are assigned based on connection order, making them unreliable for multi-monitor
  setups. Position-based selection enables deterministic workspace assignment regardless of hotplug
  order.

  Example usage for a triple-monitor setup:
  workspace = 1, monitor:position:0x0, persistent:true
  workspace = 2, monitor:position:1920x0, persistent:true
  workspace = 3, monitor:position:3840x0, persistent:true

  This is more reliable than using monitor IDs because:
  - Monitor IDs depend on attachment order
  - Position-based selection works consistently with monitor configurations like:
    monitor=eDP-1,preferred,0x0,1       # left monitor
    monitor=HDMI-A-1,preferred,1920x0,1 # middle monitor
    monitor=DP-1,preferred,auto-right,1 # right monitor

  **Implementation:**
  - Added position matching in `CMonitor::matchesStaticSelector()`
  - Added `CCompositor::getMonitorFromPosition()` helper function
  - Updated `CConfigManager::getBoundMonitorForWS()` to handle position-based selection
  - Fallback to monitor ID 0 when position doesn't match any monitor

  #### Is there anything you want to mention? (unchecked code, possible bugs, found problems,
  breaking compatibility, etc.)

  - Fully backward compatible - existing `monitor:desc:` and `monitor:name` syntax unchanged
  - Integration tests added in `hyprtester/src/tests/main/persistent.cpp`
  - Fallback behavior ensures graceful handling when position doesn't exist

  Ready for merging. I tested this behavior on my Fedora 43 build manually and works as I expected

### DIFF
--- a/hyprtester/src/tests/main/persistent.cpp
+++ b/hyprtester/src/tests/main/persistent.cpp
@@ -67,6 +67,68 @@ static bool test() {
         EXPECT_COUNT_STRING(str, "workspace ID ", 2);
     }
 
+    // Test position-based monitor selection
+    NLog::log("{}Testing position-based monitor selection", Colors::YELLOW);
+
+    // Get the position of HEADLESS-PERSISTENT-TEST monitor
+    std::string monitorPos;
+    {
+        auto str = getFromSocket("/monitors");
+        auto pos = str.find("HEADLESS-PERSISTENT-TEST");
+        if (pos != std::string::npos) {
+            // Find "at XxY" for this monitor (format is: "1280x720@60Hz at 1280x0")
+            auto atPos = str.find("at ", pos);
+            if (atPos != std::string::npos) {
+                auto lineEnd  = str.find("\n", atPos);
+                auto posStart = atPos + 3;
+                // Skip any whitespace after "at "
+                while (posStart < lineEnd && std::isspace(str[posStart]))
+                    posStart++;
+                // Extract position until whitespace or newline
+                auto posEnd = posStart;
+                while (posEnd < lineEnd && !std::isspace(str[posEnd]))
+                    posEnd++;
+                monitorPos = str.substr(posStart, posEnd - posStart);
+            }
+        }
+    }
+
+    NLog::log("{}HEADLESS-PERSISTENT-TEST monitor at position {}", Colors::YELLOW, monitorPos);
+
+    // Test 1: Normal position (XxY format)
+    const auto ws7Rule =
+        "/keyword workspace 7, monitor:position:" + monitorPos + ", persistent:1";
+    OK(getFromSocket(ws7Rule));
+
+    {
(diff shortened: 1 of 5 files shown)


cmihail:
Actually I found an edge case not yet supported by this configuration, i.e. if I want to assign workspaces to the most right monitor.

I will see if I can add an extra setting like r-1920x0 or 1920xb-1080 which will choose layouts starting from right/bottom instead of left/top

vaxerski:
but.... why?

> Monitor IDs are assigned based on connection order, making them unreliable for multi-monitor
setups. Position-based selection enables deterministic workspace assignment regardless of hotplug
order.

Use port names??

cmihail:
> but.... why?
> 
> > Monitor IDs are assigned based on connection order, making them unreliable for multi-monitor
> > setups. Position-based selection enables deterministic workspace assignment regardless of hotplug
> > order.
> 
> Use port names??

My setup is a bit different: I have 2 or 3 monitors connected, and the first monitor is the laptop. When I have 2 monitors I want my browsers on laptop and my terminals and chat application on right external monitor . 
But if I have 3 monitors I don't want anything on my laptop, and have my browsers on middle monitor and terminals and chat application on right most monitor. 

Because of that workspace assignment must start from right, not left, which is why port names won't work. 

Initially I was using hyprland events to listen to monitor changes but it was often unreliable during resume from suspend with hyprland socket returning errors for a few seconds. Plus with such an approach workspaces are moved from one monitor to another a few times until stabilizing even if no socket error.

Which is why now I just comment or uncomment my 2 or 3 monitors setup whenever I connect my 3rd monitor ( which is a smart tv that my kid sometimes uses for cartoons )

vaxerski:
I don't like this. I'd rather prefer some monitor tagging system

cmihail:
> I don't like this. I'd rather prefer some monitor tagging system

I will try the following config:
```
monitorv2 {
  output = desc:Lenovo Group Limited 0x403A
  mode = preferred
  position = 0x0
  scale = 1
}

# middle monitor
monitorv2 {
  output = desc:<middle monitor desc>
  mode = preferred
  position = 1920x0
  scale = 1
}

# last monitor
monitorv2 {
  output = desc:Philips Consumer Electronics Company PHL 273V7 UK01917036431
  mode = preferred
  position = auto-right
  scale = 1
}
```

I now use only the middle monitor for my kid so when removing it I should automatically get all middle monitor workspaces on my laptop.

PR #12461, opened 2025-11-25
buffers: revert buffer state merging

gulafaran:
8e8bfbb0b146fb1f3234b2a3e3e92b63989e1993 specifically https://github.com/hyprwm/Hyprland/pull/12052/commits/bfee2238b6fab23174e27683731668c64b1dd90a added fifo and merged non buffer states before comitting them, something about certain xwayland non buffer commits expects a commit to happend and causes regressions as in low fps.

cant reproduce locally since i neither own those games nor can find a way to trigger it locally. so below reporters needs to fill in.
fixes: #12449 
fixes: #12408

### DIFF
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -550,13 +550,6 @@ void CWLSurfaceResource::commitState(SSurfaceState& state) {
             nullptr);
     }
 
-    if (m_current.updated.bits.damage) {
-        // damage is always relative to the current commit
-        m_current.updated.bits.damage = false;
-        m_current.damage.clear();
-        m_current.bufferDamage.clear();
-    }
-
     // release the buffer if it's synchronous (SHM) as updateSynchronousTexture() has copied the buffer data to a GPU tex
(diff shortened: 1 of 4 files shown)


maotovisk:
can confirm it fixes those issues, also fixes an obscure bug with osu! (stable) that was introduced with the fifo pr that would turn the editor into a black screen if tablet input was used:

Before:

https://github.com/user-attachments/assets/aa7f1590-4a9b-412c-bf06-9a74dbee5ea9

After:

https://github.com/user-attachments/assets/377949d8-0d47-477c-a8cc-60e391d71045

ghost:
Fixes issues with Blue Archive (Genshin Impact still unconfirmed) as discussed in #12449

whymusticode:
confirm this fixes genshin impact from #12449 thx :)

zack6849:
is this going to be backported, or a new version tagged? I'm running hyprland v 0.42.1 (the latest available) and zoom and steam are both broken, there's a fix but not in any tagged versions that are upstream :(

vaxerski:
we could make a .2

Issue #8398, opened 2024-11-09
If systemd NOTIFY_SOCKET is accessible, send STOPPING=1 on exit


Vladimir-csp:
### Description

If executing compositor in a unit with `Type=notify`, this will help bring the unit down properly. Do this when `exit` dispatcher is called, or on exit in general.

Plus maybe a configurable pause after notification is sent.

Related to: https://github.com/hyprwm/Hyprland/pull/8376, https://github.com/hyprwm/hyprland-wiki/pull/850, https://github.com/Vladimir-csp/uwsm/issues/51

fufexan:
Unless I'm mistaken, looks like it's already present
https://github.com/hyprwm/Hyprland/blob/1ba050d603dca644aca48872f62388d794c030b9/src/Compositor.cpp#L486

Vladimir-csp:
Then some pause is needed, as there are still errors from clients loosing compositor after issuing `exit` dispatcher. Deactivation does not have enough time to propagate through units, compositor exits too quickly.

vaxerski:
Hello there!

This issue has been closed, as we are moving from issues to discussions. More about this can be found [here](https://wiki.hyprland.org/Contributing-and-Debugging/Issue-Guidelines/) and in #9854.

Firstly, please make sure this issue is still relevant in latest Hyprland. If so, we ask you to open a [discussion](https://github.com/hyprwm/Hyprland/discussions) (please read the discussion guidelines first, in the pinned post)

Thank you for your understanding!

PR #12484, opened 2025-11-27
start: init start-hyprland and safe mode

vaxerski:
This MR adds `start-hyprland`, a watchdog wrapper, and safe mode.

When launched with `start-hyprland`, if hyprland crashes after fully initializing (not a "gpu unsupported" type crash) watchdog will restart it in safe mode.

A warning notif will be sent if you are not launching via start-hyprland.

Safe mode:
- prevents loading config, loads default.
- pops up a dialog that allows you to load cfg, open crash report dir, or close it and continue troubleshooting.

image

Wiki: https://github.com/hyprwm/hyprland-wiki/pull/1304

Needs:
 - [x] New hyprlang release
 - [x] nix bump
 - [x] testing
 - [x] point desktop file at start
 - [x] guiutils release for hyprland-run


alba4k:
> A warning notif will be sent if you are not launching via start-hyprland.

can be disabled, right?

vaxerski:
> can be disabled, right?

No

vaxerski:
you can now

RoootTheFox:
honestly, I think you should be able to disable the warning, as some people might prefer to run hyprland without a watchdog (for whatever reason)

i suggest replacing the warning (if disabled) with a big post-crash message on stdout/stderr, telling the user to try safe mode using a cli flag, e.g.
```
Hyprland crashed, and you aren't using start-hyprland, so Hyprland wasn't automatically restarted in safe mode.
If you wish, you can do so manually by running Hyprland with the '--safe-mode' flag.
```

edit: whoops, missed you implementing that

vaxerski:
I was thinking of adding an --x11 option for the lols but decided not to

fxzzi:
Is there any possibility of having the wrapper start and stop a systemd target so we can launch stuff with graphical-session.target without uwsm

vaxerski:
uhhh probably but why? its not meant for that

fxzzi:
> uhhh probably but why? its not meant for that

i guess its not meant for that but its great QOL which could help move people off of uwsm (as its no longer supported). the main reason i was using uwsm in the first place was because it would start graphical-session.target for me, allowing me to run systemd services as hyprland starts (hyprpaper, hyprsunset, hypridle, etc)

our bundled systemd services on those projects already start with graphical-session.target but this target is never triggered automatically for me (nixos, logging in through tty or tuigreet / greetd)

Instead I just manually write a hyprland-session.target file, and wrapped hyprland to run `systemctl --user start` and `stop` before and after hyprland. Whilst this does work it would be nice to have hyprland automatically do it

vaxerski:
has anyone found any issues or is this gtg?

ctilley83:
> > uhhh probably but why? its not meant for that
> 
> i guess its not meant for that but its great QOL which could help move people off of uwsm (as its no longer supported). the main reason i was using uwsm in the first place was because it would start graphical-session.target for me, allowing me to run systemd services as hyprland starts (hyprpaper, hyprsunset, hypridle, etc)

What do you mean uwsm is no longer supported? When was this announced?

vaxerski:
it's not _officially_ supported. It still works.

kRHYME7:
@vaxerski  Is it normal that when I launch hyprland using uwsm it will notify this?

image

Can you recommend the best approach in launching Hyprland with and without uwsm please. Thank you so much!

Ev357:
> it's not _officially_ supported. It still works.

Shouldnt be `hyprland-uwsm.desktop` or the [documentation](https://wiki.hypr.land/Useful-Utilities/Systemd-start/#in-tty) updated to:
```
uwsm start -e -D Hyprland hyprland-uwsm.desktop
```

it looks like `uwsm` generates the `XDG_CURRENT_DESKTOP` variable from the binary name, in this case `start-hyprland`.

vaxerski:
@fufexan I dont do uwsm, can you?

Vladimir-csp:
> @vaxerski Is it normal that when I launch hyprland using uwsm it will notify this?
> image
> 
> Can you recommend the best approach in launching Hyprland with and without uwsm please. Thank you so much!

I'm considering not doing automatic additions if `DesktopNames` is set. But is an extra item really a problem?

kRHYME7:
Sorry just trying to understand more. So according to https://eode.pages.freedesktop.org/xdg-specs/desktop-entry-spec/desktop-entry-spec-latest.html looks like colon-separated string in  `XDG_CURRENT_DESKTOP`  shoulb be supported right? 

so `Hyprland:start-hyprland` should still be valid? 

Am I understanding this correctly?

vaxerski:
yes but we dont parse it like that xd

Vladimir-csp:
Potential distro/config remixes can add more desktop entries with more/different components in `DesktopNames`. Parsing it like a single string is a bad idea.

vaxerski:
it doesnt do anything other than pop up an error, but yeah we could suppress the error

Vladimir-csp:
There is a potential gotcha with systemd here, if I understand the scheme correctly (`start-hyprland` will be the main process, `Hyprland` will be its child, right?). Compositor will not be able to talk to the notification socket unless access is `all`, and that [caused problems](https://github.com/bluez/bluez/issues/996). `uwsm` starts compositor unit with `all` access initially, but automatically closes it down to `main` once it's job is done, assuming compositor is the MainPID. If it's not, it won't be able to [tell systemd it's exiting](https://github.com/hyprwm/Hyprland/blob/efe665b4558370af6e89921c487cd92890183961/src/Compositor.cpp#L563).

And BTW, I [suggested](https://github.com/hyprwm/Hyprland/issues/8398#issuecomment-2482976928) adding configurable delay for exit after that notification for clients to have time to actually jump ship (which would make `exit` dispatcher work cleanly for clients).

If the goal is to engage safe mode in case of failure, I'd write something to `$XDG_CACHE_HOME` to indicate startup, and remove it on successful exit. Check if it's still there on startup, engage safe mode if it is, no extra daemons.

vaxerski:
you can't safely restart an app on crash without a daemon as far as I'm concerned

PR #12488, opened 2025-11-27
socket2: add {,de}populateworkspace events

aytizren:
This PR adds 2 events to the events socket: populateworkspace and depopulateworkspace. The former is emitted when an empty workspace gains a window, while the latter is emitted when a workspace no longer has any windows.

I tested it with this:
<details>
<summary>Rust code</summary>

```rust
use std::{
  env::{args, var},
  error::Error,
  io::{BufRead, BufReader},
  os::unix::net::UnixStream,
  path::PathBuf,
  process::Command,
};

type Res = Result<(), Box<dyn Error>>; 

fn main() -> Res {
  let his = var("HYPRLAND_INSTANCE_SIGNATURE")
    .or_else(|_| args()
      .nth(1)
      .ok_or(Box::<dyn Error>::from("No arguments provided to program"))
    )?;
  let path: PathBuf = [
    var("XDG_RUNTIME_DIR")?,
    "hypr".into(),
    his,
    ".socket2.sock".into()
  ].iter().collect();
  let stream = UnixStream::connect(path)?;
  let mut lines = BufReader::new(stream).lines();


  while let Some(Ok(msg)) = lines.next() {
    let msg: Vec<_> = msg.split(">>").collect();
    let data: Vec<_> = msg[1].split(',').collect();

    match (msg[0], &*data) {
      ("depopulateworkspace", [_, name]) => notify(name, false)?,
      ("populateworkspace",   [_, name]) => notify(name, true)?,
      _ => (),
    }
  }
  Ok(())
}

fn notify(ws: &str, populated: bool) -> Res {
  let prefix = if populated {""} else {"de"};
  let _ = Command::new("hyprctl")
    .args(["notify", "6", "10000", "0", &format!("Workspace {ws} has been {prefix}populated")])
    .spawn()?;

  Ok(())
}
```
</details>

[Demonstration video](https://vimeo.com/1141203630)
A red circle in the the workspace indicator window at the top(which is separate) means that a workspace is populated, while grey means that it is not.

~The only issue that I have seen(from my limited testing) is that, sometimes, events are not emitted in the correct order when you use the `movetoworkspace` dispatch, as can be see from second 29 until the end. I'm not sure why that happens.~
EDIT: That only seems to happen from the notification side of things in the script I sent. Simply printing each time an event occurs seems to reliably preserve the order of events.

I think that it is ready for merging ~if that issue is not a dealbreaker.~

### DIFF
--- a/src/desktop/Workspace.cpp
+++ b/src/desktop/Workspace.cpp
@@ -561,3 +561,7 @@ void CWorkspace::setPersistent(bool persistent) {
 bool CWorkspace::isPersistent() {
     return m_persistent;
 }
+
+bool CWorkspace::hasWindow() {
+    return std::ranges::any_of(g_pCompositor->m_windows, [this](const auto& w) { return w->workspaceID() == m_id && w->aliveAndVisible(); });
+}
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -504,7 +504,7 @@ void CWindow::moveToWorkspace(PHLWORKSPACE pWorkspace) {
         }
     }
 
-    if (OLDWORKSPACE && g_pCompositor->isWorkspaceSpecial(OLDWORKSPACE->m_id) && OLDWORKSPACE->getWindows() == 0 && *PCLOSEONLASTSPECIAL) {
+    if (OLDWORKSPACE && g_pCompositor->isWorkspaceSpecial(OLDWORKSPACE->m_id) && !OLDWORKSPACE->hasWindow() && *PCLOSEONLASTSPECIAL) {
         if (const auto PMONITOR = OLDWORKSPACE->m_monitor.lock(); PMONITOR)
             PMONITOR->setSpecialWorkspace(nullptr);
     }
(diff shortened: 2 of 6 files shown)


vaxerski:
why? You can already do this with existing infra.

aytizren:
When making bars/shells with special indicators for empty and populated workspaces, it looks like people currently make calls to the request socket whenever the response of `hyprctl clients` changes(e.g. [epik-shell](https://github.com/ezerinz/epik-shell/blob/91fca11/widgets/bar/WorkspacesPanelButton.tsx#L15-L25) and [noctalia](https://github.com/noctalia-dev/noctalia-shell/blob/3ff5b76/Services/Compositor/HyprlandService.qml#L390-L397)). The `Using hyprctl` page of the wiki that said that spamming the request socket may cause compositor slowdowns. It's also simpler to watch for a single event and update a single workspace by id than it is to watch for `("movewindow" | "closewindow")` and make a few checks and also watch for `openwindow` and do the same.
If you think this PR is redundant, feel free to close it.

vaxerski:
ok, I see. v2 is redundant though, v2 was because v1 sucked of a few events. Make one event.

aytizren:
Did you mean one event each(so `{,de}populateworkspace`), or one event in general (potentially with an extra boolean parameter)? If it was the first, should the id or name be emitted? If it was the second, should it retain its original name(populateworkspace) or should it be something like toggleworkspacepopulationstate(I am pretty bad at naming things)?

vaxerski:
just getting rid of v2

src/desktop/Workspace.cpp
```diff
@@ -561,3 +561,7 @@ void CWorkspace::setPersistent(bool persistent) {
 bool CWorkspace::isPersistent() {
     return m_persistent;
 }
+
+bool CWorkspace::hasWindow() {
+    return std::ranges::any_of(g_pCompositor->m_windows, [this](const auto& w) { return w->workspaceID() == m_id && w->m_isMapped; });
```
vaxerski: rebase on main and use ->aliveAndVisible()

src/desktop/view/Window.cpp
```diff
@@ -2233,6 +2247,9 @@ void CWindow::mapWindow() {
     // emit the IPC event before the layout might focus the window to avoid a focus event first
     g_pEventManager->postEvent(SHyprIPCEvent{"openwindow", std::format("{:x},{},{},{}", m_self.lock(), PWORKSPACE->m_name, m_class, m_title)});
     EMIT_HOOK_EVENT("openWindowEarly", m_self.lock());
+    if (workspacePreviouslyEmpty) {
```
vaxerski, line 2250: no {}

src/desktop/view/Window.cpp
```diff
@@ -2373,6 +2390,10 @@ void CWindow::mapWindow() {
 
     // emit the hook event here after basic stuff has been initialized
     EMIT_HOOK_EVENT("openWindow", m_self.lock());
+    if (workspacePreviouslyEmpty) {
```
vaxerski, line 2393: no {}

vaxerski:
also pls add tests

PR #12504, opened 2025-11-29
hyprpm: check for abi strings in headersValid

vaxerski:
Because we now properly include ABI info in our version string, hyprpm should take that into account when checking if we need to update.

Ref https://github.com/outfoxxed/hy3/issues/254
Ref https://github.com/hyprwm/hyprland-plugins/issues/552

Needs a quick test

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -909,7 +909,7 @@ bool CPluginManager::loadUnloadPlugin(const std::string& path, bool load) {
     auto state = DataState::getGlobalState();
     auto HLVER = getHyprlandVersion(true);
 
-    if (state.headersAbiCompiled != HLVER.hash) {
+    if (state.headersAbiCompiled != HLVER.abiHash) {
         if (load)
             std::println("{}", infoString("Running Hyprland version ({}) differs from plugin state ({}), please restart Hyprland.", HLVER.hash, state.headersAbiCompiled));
         return false;


vaxerski:
@VirtCode have you had a chance to verify this is working

VirtCode:
I don't use `hyprpm` myself but I can quickly have a look whether things still work as expected (could also try to simulate an ABI break with some downgrading but I'll see). Will be back later this evening, need to finish an assignment first.

vaxerski:
thanks bae xoxo

VirtCode:
sorry honey, won't get to it today anymore, will see tomorrow morning. That assignment literally drove me insane I gotta get some sleep first.

vaxerski:
no prob qt

VirtCode:
Okay finally got around to it. Currently `hyprpm update` does not work with an empty cache. Always fails with "ABI is mismatched.", even with `--force`. So to reproduce, just run `hyprpm purge-cache` and then `hyprpm update`.

I had a quick glance at the code and I guess the issue is that `headersValid()` checks against the hash in the global state, and fails if it doesn't match. In our case the global state doesn't exist yet as we seem to write it [after checking header validity](https://github.com/vaxerski/Hyprland/blob/hyprpm-abi/hyprpm/src/core/PluginManager.cpp#L596)? But a similar thing would happen if our state contains an old hash, so I don't know whether that's really what's happening.

vaxerski:
check nwo @VirtCode

VirtCode:
uhm, you forgor to push?

vaxerski:
wtf happened

VirtCode:
doesn't work, it still fails because of "ABI is mismatched", you'd probably want allow `HEADERS_ABI_MISMATCH` there too

vaxerski:
doe

hyprpm/src/core/PluginManager.cpp
```diff
@@ -913,9 +909,9 @@ bool CPluginManager::loadUnloadPlugin(const std::string& path, bool load) {
     auto state = DataState::getGlobalState();
     auto HLVER = getHyprlandVersion(true);
 
-    if (state.headersHashCompiled != HLVER.hash) {
+    if (state.headersAbiCompiled != HLVER.hash) {
```
VirtCode:
```suggestion
    if (state.headersAbiCompiled != HLVER.abiHash) {
```

vaxerski:
tahnjks!

PR #12518, opened 2025-11-30
Remove move and center window rules, combine them as new position rule

lichie567:
Another PR addressing the issues highlighted in my comment here: https://github.com/hyprwm/Hyprland/discussions/12434#discussioncomment-15113873

This is a fix for the `move` and `center` rules not overriding each other. Currently, because these were stored as separate effects, the center rule was always taking priority because the application of that effect was happening later in `listener_mapWindow`.

This PR removes both of those rules and replaces them with a new one named `position`, which is a better descriptor of the effect IMO. The `position` rule accepts the same arguments as move, but also accepts "center" as a option. With this change, whichever `position` rule is defined last wins.

If you would like, I can modify this PR to leave the `center` and `move` rules intact for backwards compatibility, but I figured users will be forced to update their configs with recent changes anyway so a big breaking change may be acceptable at the moment. Please point me to any follow-up items that would need to be taken care of elsewhere, such as updates to documentation or tests.

And lastly, regarding the comment I linked above, this PR still does not fix a potential bug where workspace rules from the focused workspace are being applied to windows launched on another workspace via a rule passed to the exec dispatcher.


vaxerski:
I don't like this tbh? I like my center rule. You can make the change in the background but meh

lichie567:
Thats understandable, I'm willing to iterate on this. But to me it does feel like center is just a special case of move. I was thinking of expanding this to allow for a full set of relative positional arguments. Perhaps something with a format like `position [location] [expr] [expr]`, where you could specify `position center -500 500` to position a window at a coordinate offset relative to the center of the screen.

Do you have any suggestions?

vaxerski:
I like center, I don't see a reason to make things longer than they need t obe

lichie567:
> I like center, I don't see a reason to make things longer than they need t obe

I understand that you "like" it, but that doesn't change the fact that the current design is broken. Something has to change because functionality that previously existed no longer exists. If "center" is defined more generally (e.g. in a rule matching a workspace) it is currently impossible to defeat that rule by overriding it with a more specifically defined move rule.

There are three ways I see to resolve this bug:

1. Combine move and center rules, as I suggested above
2. Allow a way to "disable" a center rule by defining `center off` or something like that, but imo this is verbose and not very elegant
3. Simply swap the application of these rules in the code so that move is applied after center, which IMO is also not very elegant, and results in the inverse problem happening, where move will *always* take precedence over center.

Issue #11806, opened 2025-09-23
Add an option to hide cursor on tablet


vaxerski:
### Discussion
https://github.com/hyprwm/Hyprland/discussions/11787

## Description
hide on touch exists, akin to that, hide on tablet.

opalmay:
As in, hide once when entering tablet mode?

vaxerski:
once tablet enters proximity and moves the cursor

MihaiStreames:
Since the merge above, you should be able to just use `cursor:hide_on_tablet = true` in your config, which works like `cursor:hide_on_touch`

It is off by default though

PR #12525, opened 2025-12-01
Tablet: added option to hide cursor

winterleg:
#11806

Added an option to hide cursor when using a pen: cursor:hide_on_pen

Works as expected, cursor will hide if the pen moves or is in proximity of the tablet. The cursor will reappear on mouse movement.

I'm not sure if it works okay with `cursor:hide_on_touch` as i have no way to test that.

### DIFF
--- a/src/render/Renderer.hpp
+++ b/src/render/Renderer.hpp
@@ -155,7 +155,7 @@ class CHyprRenderer {
 
     struct {
         bool hiddenOnTouch    = false;
-        bool hiddenOnTablet      = false;
+        bool hiddenOnTablet   = false;
         bool hiddenOnTimeout  = false;
         bool hiddenOnKeyboard = false;
     } m_cursorHiddenConditions;


src/config/ConfigDescriptions.hpp
```diff
@@ -1670,6 +1670,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_BOOL,
         .data        = SConfigOptionDescription::SBoolData{true},
     },
+    SConfigOptionDescription{
+        .value       = "cursor:hide_on_pen",
```
vaxerski: I'd call it tablet cuz thats what we call it elsewhere

src/managers/input/Tablets.cpp
```diff
@@ -169,6 +169,8 @@ void CInputManager::onTabletAxis(CTablet::SAxisEvent e) {
 }
 
 void CInputManager::onTabletTip(CTablet::STipEvent e) {
+    m_lastInputPen = true;
```
vaxerski: wouldnt this be better in proximity onstead of tip?

winterleg: I guess it is a little redundant since the value is already changed in onTabletProximity.

vaxerski (approved):
thanks, wiki mr needed

PR #12548, opened 2025-12-05
Adds zoom with detached camera

jmanc3:
https://github.com/user-attachments/assets/1dc3dbee-d900-4f9d-aa7a-dd2f6383eeec

Implements 'push' zoom discussed [here](https://github.com/hyprwm/Hyprland/discussions/12422). I tested this on multiple monitors, and multiple scaling factors and noticed no issues. 

I made it default because this behaviour absolutely mogs KDE and Windows. It's actually insane how hard we mog their implementations.

(brain hurty! but I think there might even be a better version. corner zoom is perfect, but center-ish content zoom should be more rigid)

### DIFF
--- a/src/helpers/MonitorZoomController.hpp
+++ b/src/helpers/MonitorZoomController.hpp
@@ -6,13 +6,14 @@ struct SCurrentRenderData;
 
 class CMonitorZoomController {
   public:
-    void  applyZoomTransform(CBox& monbox, const SCurrentRenderData& m_renderData);
+    bool m_resetCameraState = true;
 
-    CBox  m_camera;
-    bool  m_resetCameraState = true;
-    float m_lastZoomLevel    = 1.0f;
-    bool  m_padCamEdges      = true;
+    void applyZoomTransform(CBox& monbox, const SCurrentRenderData& m_renderData);
 
   private:
-    void zoomWithDetachedCamera(CBox& result, const SCurrentRenderData& m_renderData);
+    void  zoomWithDetachedCamera(CBox& result, const SCurrentRenderData& m_renderData);
+
+    CBox  m_camera;
+    float m_lastZoomLevel = 1.0f;
+    bool  m_padCamEdges   = true;
 };


jmanc3:
~~maybe wait before merging while I figure out the magic version~~
~~nvm, I'm tripping, if you disable detached cam, you can compare, and the zooming behaviour is exactly the same as non rigid zoom, which is totally fine.~~
nvm, there is a weird first zoom step visual snap glitch that happens, that I need to fix

jmanc3:
Still need one more discontinuity to fix and then it should be good 

(also I feel that zoom_aa should be disabled by default as blurry looks more wrong than pixelated but sharp, thoughts on disabling aa as the default?)

jmanc3:
@vaxerski it should be good to go. I tested on multiple monitors on different scales.

vaxerski (commented):
code here is getting kinda clunky. Make a new file with a class that does all zoom modes and just returns the needed values for `opengl::end()`.

then you can `CMonitorZoomController m_zoomController;` in CMonitor and access that.

src/helpers/Zoom.cpp
```diff
@@ -0,0 +1,95 @@
...
+void CMonitorZoomController::applyZoomTransform(CBox& monbox, const SCurrentRenderData& m_renderData) {
+    static auto PZOOMRIGID          = CConfigValue<Hyprlang::INT>("cursor:zoom_rigid");
+    static auto PZOOMDETACHEDCAMERA = CConfigValue<Hyprlang::INT>("cursor:zoom_detached_camera");
+
+    const auto  ORIGINAL = monbox;
+    const auto  MONITOR  = m_renderData.pMonitor;
+    const auto  INITANIM = MONITOR->m_zoomAnimProgress->value() != 1.0;
+    const auto  ZOOM     = m_renderData.mouseZoomFactor;
+
+    if (ZOOM != 1.0f) {
```
vaxerski: use a guard instead

src/helpers/Zoom.cpp
```diff
@@ -0,0 +1,95 @@
...
+        if (*PZOOMDETACHEDCAMERA && !INITANIM) {
```
vaxerski: no {}

src/helpers/Zoom.cpp
```diff
@@ -0,0 +1,95 @@
+#include "Zoom.hpp"
+
+#include <hyprlang.hpp>
+#include "../config/ConfigValue.hpp"
+#include "../managers/input/InputManager.hpp"
+#include "../render/OpenGL.hpp"
+#include "desktop/DesktopTypes.hpp"
+#include "render/Renderer.hpp"
+
+void zoomWithDetachedCamera(CBox& result, CMonitorZoomController* zc, const SCurrentRenderData& m_renderData) {
```
vaxerski: why pass a pointer to zc? why isn't this just a member?

src/helpers/Zoom.cpp
```diff
@@ -0,0 +1,95 @@
+#include "Zoom.hpp"
```
vaxerski: Filename is ambiguous, just take it from the class name: MonitorZoomController

vaxerski (approved):
lgtm, wiki mr needed

karbyt:
https://github.com/user-attachments/assets/df739915-c001-4ac3-9ed1-0607dbf2e0f7

Why is my monitor flickering after the Zoom update? Even though the previous version ran smoothly.

EvilLary:
Can confirm, It happens only if animations are disabled. 
It's also abit buggy and wont update my screen to the new zoom value to until I move my cursor

vaxerski:
probably missing damage somewhere but thats odd cuz this didnt change much

PR #12552, opened 2025-12-05
dispatcher: include mirrors of monitor in dpms

EvilLary:
Fixes mirrors not being turned off for example on idle with `hyprctl dispatch dpms off` 

noting

Ready


EvilLary:
Or would be it better to iterate over `m_realMonitors` instead, so one could turn toggle a mirror monitor. As of right now it isn't possible without using monitor rules

PR #12560, opened 2025-12-06
config/input: Invert movement for all pointer devices, not just touchpads

m-matovic:

Use `m_flipX`/`m_flipY` to invert movement for all pointer devices, not just touchpads.
Move the `flip_x`/`flip_y` config options from `input::touchpad` to `input` and load them for all pointer devices.

If the changes are merged, the wiki page Configuring/Variables should be update accordingly (I'll do it).

Ready

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,11 +16,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1764370710,
-        "narHash": "sha256-7iZklFmziy6Vn5ZFy9mvTSuFopp3kJNuPxL5QAvtmFQ=",
+        "lastModified": 1764714051,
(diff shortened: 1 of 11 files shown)


vaxerski (commented):
keep touchpad, don't break. Fall back to global if touchpad is unset.

m-matovic:
I am a bit unclear on why touchpads had these properties unique to them? If they're just pointers with a size, why does having a size make them elligible for flipping but other pointers not?

Is it not more straightforward to leave the flip config options in global scope and load them for all pointers? They can be overriden in per-device configs.

I will return the isTouchpad property in any case, even if it's left unused right now.

Will make changes tonight.

PR #12573, opened 2025-12-07
internal: start: Better signal handling

kolayne:

- Take out signal set up into a subroutine;

- Use `sigaction` instead of `signal` for consistent behavior across UNIX platforms;

- Enable a warning when a signal handler set up fails;

- Don't do anything to SIGKILL, since it cannot be handled.

No.

Should be ready for merge.

### DIFF
--- a/start/src/main.cpp
+++ b/start/src/main.cpp
@@ -1,4 +1,5 @@
 #include <csignal>
+#include <cstring>
 #include <print>
 
 #include "helpers/Logger.hpp"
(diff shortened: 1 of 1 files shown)


start/src/main.cpp
```diff
@@ -28,14 +30,23 @@ static void onSignal(int sig) {
...
 }
 
+static void terminateChildOnSignal(int signal) {
+    struct sigaction sa;
+    sa.sa_handler = onSignal;
+    sigemptyset(&sa.sa_mask);
+    sa.sa_flags = SA_RESTART;
+    int ret     = sigaction(signal, &sa, nullptr);
+    if (ret != 0)
+        g_logger->log(Hyprutils::CLI::LOG_WARN, std::format("Failed to set up handler for signal {}: {}", signal, strerror(errno)));
```
vaxerski: log already formats

kolayne: Fixed

PR #12580, opened 2025-12-07
Update Exec command for UWSM Hyprland desktop entry

kRHYME7:
This is from the comment in https://github.com/hyprwm/Hyprland/pull/12484 

https://github.com/hyprwm/Hyprland/pull/12484#issuecomment-3621979533

https://github.com/hyprwm/Hyprland/pull/12484#issuecomment-3621979533

yep

### DIFF
--- a/systemd/hyprland-uwsm.desktop
+++ b/systemd/hyprland-uwsm.desktop
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=Hyprland (uwsm-managed)
 Comment=An intelligent dynamic tiling Wayland compositor
-Exec=uwsm start -e -D Hyprland hyprland-uwsm.desktop
+Exec=uwsm start -e -D Hyprland hyprland.desktop
 TryExec=uwsm
 DesktopNames=Hyprland
 Type=Application


systemd/hyprland-uwsm.desktop
```diff
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=Hyprland (uwsm-managed)
 Comment=An intelligent dynamic tiling Wayland compositor
-Exec=uwsm start -- hyprland.desktop
+Exec=uwsm start -e -D Hyprland hyprland-uwsm.desktop
```
Ev357: I think we need to start `hyprland.desktop` not `hyprland-uwsm.desktop`. That would be an infinite loop i believe.

In my comment earlier i said that it should be eather fixed here or by updating the documentation where `hyprland-uwsm.desktop` would be necessary.

kRHYME7: ops, removed the uwsm and tested. It works now

vaxerski (approved):

ReStranger:
how can I use this desktop entry for example in tuigreet (in nixos)?

Ev357:
@ReStranger I use this "patch"

https://github.com/Ev357/.dotfiles/blob/012ebbecebde34197327df2dbffc2df7753120d2/nixosModules/hyprland/patch/default.nix

then you should be able to start hyprland with `uwsm start -e -D Hyprland hyprland-uwsm.desktop`

PR #12584, opened 2025-12-07
Time Fix

Rtur2003:
Improved time helper robustness by normalizing `timespec` nanosecond overflow and using the correct `std::system_clock` duration type. This prevents underflow/overflow skew during steady↔`timespec` conversions (`src/helpers/time/Time.cpp`, `Time.hpp`). Added regression tests covering borrow handling, fixed-point sec/nsec splits, and overflow normalization to guard against future drift (`tests/helpers/TimeDiff.cpp`). No breaking changes; conversions are now stricter and safer.
Tests were not run here—please build and run `ctest` / `hyprland_gtests` to verify.

### DIFF
--- a/tests/helpers/TimeDiff.cpp
+++ b/tests/helpers/TimeDiff.cpp
@@ -3,12 +3,12 @@
 
 TEST(HelpersTime, DiffBorrowsNanoseconds) {
     const Time::detail::sec_nsec newer{5, 1};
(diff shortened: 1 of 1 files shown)


vaxerski:
image

I see maaany unrelated changes here. Can you split your changes into separate, smaller MRs, each one doing _one_ thing?

PR #12589, opened 2025-12-07
Guard plugin longjmp when jump buffer isn’t primed 

Rtur2003:
## Summary                                                                                                        
- track whether the hook jump buffer is initialized before longjmp                                                
- skip the jump when missing/invalid and fall back to crash reporting                                             

## Testing                                                                                                        
- manual: load a faulting plugin; verify no longjmp into uninitialized buffer and crash path is taken

### DIFF
--- a/src/managers/HookSystemManager.hpp
+++ b/src/managers/HookSystemManager.hpp
@@ -51,6 +51,7 @@ class CHookSystemManager {
     std::vector<SCallbackFNPtr>* getVecForEvent(const std::string& event);
 
     bool                         m_currentEventPlugin = false;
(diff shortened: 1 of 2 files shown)


src/managers/HookSystemManager.hpp
```diff
@@ -51,6 +51,7 @@ class CHookSystemManager {
     std::vector<SCallbackFNPtr>* getVecForEvent(const std::string& event);
 
     bool                         m_currentEventPlugin = false;
+    bool                         m_hookFaultJumpBufReady = false;
```
vaxerski, line 54: the bool does nothing, it's never used

PR #12590, opened 2025-12-07
Normalize timespec borrow/overflow and add regression tests

Rtur2003:
## Summary                                                                                                        
- normalize timespec nsec overflow/borrow before conversions and expose helper for tests                          
- fix system_clock sec/nsec reconstruction drift                                                                  
- add regression tests for Time::toTimespec/fromTimespec and diff/normalize edge cases                            

## Testing                                                                                                        
- tests: ctest (gtest helpers/Time*, helpers/TimeDiff*)

### DIFF
--- a/scripts/generateShaderIncludes.sh
+++ b/scripts/generateShaderIncludes.sh
@@ -1,24 +1,49 @@
 #!/bin/sh
 
+# dash-compatible and space-safe shader include generator. @Rtur2003
 SHADERS_SRC="./src/render/shaders/glsl"
+SHADERS_OUT="./src/render/shaders"
+SHADERS_HPP="${SHADERS_OUT}/Shaders.hpp"
 
 echo "-- Generating shader includes"
 
-if [ ! -d ./src/render/shaders ]; then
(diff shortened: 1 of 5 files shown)


scripts/generateShaderIncludes.sh
```diff
@@ -1,24 +1,49 @@
 #!/bin/sh
 
+# dash-compatible and space-safe shader include generator. @Rtur2003
```
vaxerski, line 3: what are these changes

src/helpers/time/Time.cpp
```diff
@@ -137,3 +132,15 @@ struct timespec Time::toTimespec(const steady_tp& tp) {
     auto sum = timeadd(tpTime, diffFinal);
     return timespec{.tv_sec = sum.first, .tv_nsec = sum.second};
 }
+
+namespace Time::detail {
+
+sec_nsec diff(const sec_nsec& newer, const sec_nsec& older) {
```
vaxerski, line 138: Time::detail::diff instead of namespace {}

PR #12594, opened 2025-12-08
hyprpm: added plugin author

### LINKED ISSUE #11945: Add a way to distinguish same-name plugins in hyprpm
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11925

## Description
Plugins that share a name introduce confusion in enable/disable, we should allow `author/name`.

Brumus14:
Allows users to distinguish between plugins with the same name by optionally specifying the author when giving the name such as:
```
hyprpm remove plugin-name plugin-author
hyprpm enable plugin-name plugin-author
hyprpm disable plugin-name plugin-author
```
Closes #11945

Adds author an field to repository toml files so you have to remove and add repositories again as it's not backwards compatible with old toml files.

Should be ready.

### DIFF
--- a/hyprpm/src/core/DataState.cpp
+++ b/hyprpm/src/core/DataState.cpp
@@ -93,6 +93,7 @@ void DataState::addNewPluginRepo(const SPluginRepository& repo) {
     auto DATA = toml::table{
         {"repository", toml::table{
             {"name", repo.name},
+            {"author", repo.author},
             {"hash", repo.hash},
             {"url", repo.url},
             {"rev", repo.rev}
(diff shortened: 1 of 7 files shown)


vaxerski:
toml change is fine. I don't like the syntax. Something like `author/name` would make more sense. `hyprpm enable vaxry/hyprscrolling` or `hyprscrolling`. Plugin names can't contain / anyways.

Brumus14:
Yeah that makes sense I'll change it.

Brumus14:
Updated the syntax you can see what it's like here:
https://github.com/hyprwm/Hyprland/blob/9f0e5498a2c2095db4ec8b6b2f6134e9c64af022/hyprpm/src/main.cpp#L16-L20

PR #12618, opened 2025-12-10
WindowRuleApplicator: force center and move rules to override each other

lichie567:
This PR is a follow up to this one: https://github.com/hyprwm/Hyprland/pull/12518

I decided to take a new approach so I created a new PR. The issue is described at length in the PR linked above, but I'll shortly describe the issue again.

`move` and `center` rules are separate effects that both affect the position of a window, which causes undefined behavior when both rules are inherited by the same window. Intuitively, the one defined "last" should take precedence, but currently that isn't the case. The `center` rule always takes precendence because that effect is applied later in the code.

While investigating I found it odd that defining `center off` as a rule did not seem to work, as from looking at the code it seemed like it should. However I eventually realized the `center` field in the struct was declared as an `std::optional<int>` which has a truthy value even when assigned an explicit value of false. :facepalm: 

I decided to just always unset the other effect when setting either `move` or `center`. While this is probably the simplest way to fix this, I still don't consider it the cleanest way.

No

Should be ready to merge

### DIFF
--- a/src/desktop/rule/windowRule/WindowRuleApplicator.cpp
+++ b/src/desktop/rule/windowRule/WindowRuleApplicator.cpp
@@ -481,7 +481,7 @@ CWindowRuleApplicator::SRuleResult CWindowRuleApplicator::applyStaticRule(const 
                 break;
             }
             case WINDOW_RULE_EFFECT_MOVE: {
-                static_.center   = false;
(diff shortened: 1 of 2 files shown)


vaxerski:
conflict

lichie567:
should be good to go now

PR #12633, opened 2025-12-11
window: only damage floating on clamped size change

gulafaran:
currently it damage the entire window if its floating and not x11 nor fullscreen meaning damage isnt working at all for floating. im tracing this back to a364df4 where the logic changed from damaging window only if size was being forced to now unconditonally doing it.

change clampWindowSize to return as a bool if size changed and only damage window if it got clamped.

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -1197,14 +1197,19 @@ int CWindow::surfacesCount() {
     return no;
 }
 
-void CWindow::clampWindowSize(const std::optional<Vector2D> minSize, const std::optional<Vector2D> maxSize) {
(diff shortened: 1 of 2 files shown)


gulafaran:
or well to be fair old code even `if (MAXSIZE > Vector2D{1, 1})` this might have unconditionally damaged entire window at all times when floating.

adityasz:
It was added in [this](https://github.com/hyprwm/Hyprland/commit/395ab3ba370f5e12fcba594232c338c16deef0ed#diff-3e2f9e682b3648467d26027473c58748f6bd5e0c8db0fda93fb3116ed2a55a3fL825)

PR #12639, opened 2025-12-12
splashes: add a few random splashes

alba4k:
splashes get a bit repetitive, so why not add a few?

the first 4 are minecraft splashes, the remaining are just random


vaxerski:
I dont accept splash mrs

PR #12646, opened 2025-12-14
touch: fix LS on different monitor than cursor not getting touch events

horriblename:

fixes #12221

as mentioned in the discussion, layer shells do not get touch input if cursor is in a different
monitor

yes

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -183,7 +183,7 @@ void CInputManager::sendMotionEventsToFocused() {
     g_pSeatManager->setPointerFocus(Desktop::focusState()->surface(), LOCAL);
 }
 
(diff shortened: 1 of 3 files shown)


horriblename:
@TheBigViolinist could you test this on your machine?

TheBigViolinist:
I tested and its in PERFECT state! The problem is solved!!! :)))))) the two screen device users thank you my guy

vaxerski (commented):
why are you making fundamental changes instead of the much more obvious and straightforward focus monitor before refocus()?

horriblename:
wouldn't that take focus away from the current window? that would make it so that you can't type with an on-screen keyboard which is on a different monitor than your window, as touching the onscreen keyboard takes focus away from the window

vaxerski:
hm. Probably should separate refocus into hit test then. Remind me in a week, I'm busy atm, but I can do something bout this

Issue #19, opened 2022-04-11
Tabbed containers


Ashe:
**Describe in detail your idea.**
My primary usecase of i3/sway is being able to ~~stack~~ tab windows in a container. I have a super ultra wide monitor and I have 3 columns of windows with my 'main' window in the middle. I always make this a ~~stacked~~ tabbed container so I can easily switch from my web browser / terminal / steam / game all being in the middle of my monitor. It's also nice just to declutter spare windows.

It has similar usecases as workspaces, but since my monitor is so wide I don't need an entire new screen of windows, I just want to easily repurpose my main area.

This is important to me and I'm unlikely to switch to this as my main wm unless this is a feature. Other than that though, it really is quite a nice experience! About time we had a fresh wm!

vaxerski:
can you post a demo of the feature? Link to a youtube video where someone uses it is enough.

Ashe:
I think in i3 it's actually named a 'tabbed' layout. I guess this is better since calling it stacked implies we're going from tiling window management to stacking window management.

https://youtu.be/RUKXZjxou3U?t=85

I linked a YouTube clip (not mine) where the person has a full-screen container and then converts it to a tabbed layout, so that rather than windows tiling horizontally or vertically within it, they are instead hidden while exposing 'tabs' at the top (like a web browser). I don't use this feature in full-screen like in the video (that's when workspaces are typically used anyway), I use this for a portion of my scree; I'd only switch my central container to a tabbed layout.

I hope this helps!

vaxerski:
okay, can implement in dwindle.

vaxerski:
added in this b6a93b2f0383fc8f084c7757f0f67b8c0ebeccc7 bad boy, should work, mostly tested.

see the new dispatchers and configs for border color in the wiki.

Cheers.

Ashe:
> added in this [b6a93b2](https://github.com/vaxerski/Hyprland/commit/b6a93b2f0383fc8f084c7757f0f67b8c0ebeccc7) bad boy, should work, mostly tested.
> 
> see the new dispatchers and configs for border color in the wiki.
> 
> Cheers.

I'm sorry I bet actually thanked you for doing this! I've been very busy but it's still on my to-do list to use your window manager and now that you've got tabbed containers I'm going to evangelize you!

Thanks so much!

Issue #18, opened 2022-04-07
Some drag-n-drop apps are wonky


vaxerski:
Some drag-n-drop operations don't work well, e.g. FileZilla.

vaxerski:
XWayland issue.

vaxerski:
Happens with other XWayland windows as well, these windows that pop up are some obscure notif/overlay kinda things?

vaxerski:
Windows popping up fixed, but drag-n-drop doesnt work

Vagahbond:
For me drag and drop works in most applications, just not in VSCode and neither in Spotify.

vaxerski:
yeah, XWayland. VSCode, FileZilla, etc.

spectreseven1138:
I noticed that this only occurs for me when follow_mouse is set to 0, regardless of whether the window is actually focused. When set to 1 every drag and drop I've tried works fine. Is this the same issue or am I experiencing something else?

vaxerski:
try filezilla, should be the opposite

mrshmllow:
Dragging selected text is also very laggy

ThatOneCalculator:
I can't drag-and-drop even on Wayland/GTK4 apps like Nautilus and Nemo

vaxerski:
@ThatOneCalculator if you have focus_follows_mouse to anything other than 1 then DnD cannot work

vaxerski:
fixed with the last commits

Issue #17, opened 2022-04-07
Mouse constraints/locking isn't implemented.


vaxerski:
Especially games, which use the cursor for looking around, e.g. shooters, minecraft, etc. Cannot lock the mouse and thus are unplayable.

vaxerski:
le fixed with 7b1eaca84014d88a420bcbd3f97467f74e82ab33

FlyingStitchman:
Just tested, this does not seem to work with multiple monitors. On the directions that another monitor exists, the mouse can leave the game window and enter the other monitor. I did not see an option in the wiki, so I assumed that locking would be implemented on any `fullscreen` window. The other directions (ones without adjacent monitors) work fine. I tried both Risk of Rain 2 and No Mans Sky.

vaxerski:
I only tried minecraft :P

a game has to request a mouse constraint, but I'll look into it. thanks.

FlyingStitchman:
Just loaded up minecraft, the multi-monitor issues i mentioned are present.

vaxerski:
on non-fullscreen too???

FlyingStitchman:
Window locking only seems to work when the border of the window is the exact same as the border of the entire desktop resolution. If a window is not fullscreen, it does not seem to work (only tested minecraft). If a window is fullscreen, but there is an adjacent screen, the mouse cursor can enter the adjacent screen, although the other sides without an adjacent screen work fine.

vaxerski:
it worked for me perfectly, weirdd

FlyingStitchman:
I just tried recompiling it and it now works (mind you I cloned and compiled it already, but apparently it was not happy for some reason).

FlyingStitchman:
Apologies for reopening this for an unwarranted reason.

vaxerski:
chill, no worries

Mikilio:
I also apologize for reopening this issue. Unfortunately the issue arises also with league of legends installed through lutris. I'm using NixOS.

vaxerski:
Expected behavior, we don't want users playing league of legends.

Mikilio:
> Expected behavior, we don't want users playing league of legends.

I apologize if you have to download the game for debugging.

vaxerski:
There is no debugging needed. It's, again, intended behavior.

Mikilio:
[...]

PR #12665, opened 2025-12-16
Fix mouse motion in VRR

Szwagi:
I am unfamiliar with the codebase, so this PR most definitely has plenty of issues I haven't found.
If the actual solution is more complicated than this, I'm unlikely to be of any help, but would love to see this fixed and working as well as this is for me.

Fixes https://github.com/hyprwm/Hyprland/discussions/12032 and this sounds related too https://github.com/hyprwm/Hyprland/discussions/12013.

I will take CS2 on a 360Hz display as an example. When I say FPS measurements I mean what the monitor reports.

**Before (the issues):**
- **no_break_fs_vrr 0:** The sensitivity is consistent. Moving the mouse *even while the cursor is locked* spikes the FPS to 357 (sends duplicate frames, making VRR useless).
- **no_break_fs_vrr 1:** The sensitivity of the mouse changes depending on the FPS of the game.
- The game feels very unsmooth, even when it runs above 300FPS, likely to do with weird delayed mouse motion. 

**After:**
- **no_break_fs_vrr 0:** The sensitivity is consistent. Moving the mouse spikes the FPS only when the cursor is unlocked. During gameplay (locked cursor), the FPS stays at the FPS cap I set.
- **no_break_fs_vrr 1:** The sensitivity is consistent.
- The game feels very smooth, basically the same as other compositors.

My idea of what is happening is that
1. Hyprland is resetting the stored mouse movement before it gets sent to the game.
2. I do not understand why Hyprland even has to accumulate the movement instead of sending it straight to the game. I guess it could help performance on 8000Hz mice, but in practice it seems like it just adds a bunch of input lag and possibly holds on to it so long it misses frames (hence the unsmoothness).

I have tested in CS2 with native Wayland and XWayland. I have also tested in a Wine game running through Gamescope.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -3017,7 +3017,7 @@ bool CConfigManager::shouldUseSoftwareCursors(PHLMONITOR pMonitor) {
     switch (*PNOHW) {
         case 0: return false;
         case 1: return true;
-        case 2: return g_pHyprRenderer->isNvidia() && g_pHyprRenderer->isMgpu();
+        case 2: return g_pHyprRenderer->isNvidia() && (g_pHyprRenderer->isMgpu() || g_pCompositor->isVRRActiveOnAnyMonitor());
         default: break;
     }
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -3099,3 +3099,7 @@ std::optional<unsigned int> CCompositor::getVTNr() {
 
     return ttynum;
 }
+
+bool CCompositor::isVRRActiveOnAnyMonitor() const {
+    return std::ranges::any_of(m_monitors, [](const PHLMONITOR& m) { return m->m_vrrActive; });
+}
--- a/src/Compositor.hpp
+++ b/src/Compositor.hpp
@@ -161,6 +161,7 @@ class CCompositor {
     void                                onNewMonitor(SP<Aquamarine::IOutput> output);
     void                                ensurePersistentWorkspacesPresent(const std::vector<SWorkspaceRule>& rules, PHLWORKSPACE pWorkspace = nullptr);
     std::optional<unsigned int>         getVTNr();
+    bool                                isVRRActiveOnAnyMonitor() const;
 
     NColorManagement::PImageDescription getPreferredImageDescription();
     NColorManagement::PImageDescription getHDRImageDescription();


UjinT34:
If this works for all locked/unlocked hw/sw wayland/xwayland combinations with different games then I'm happy that all the pointer skipping and movement accumulation gets nuked. The accumulation was needed because sending the movement with locked cursor to an xwayland client straight away caused and immediate frame event in return which was treated as a valid new frame from the game and broke VRR.
Looks like this also nukes `m_pointerEvents.frame.emit()` and its handling. Maybe the original issue was cause by using this event to call `g_pSeatManager->sendPointerFrame();` instead of calling it right after `PROTO::relativePointer->sendRelativeMotion`. No idea why the event was used, it was this way before the hack.

Szwagi:
I just checked and this definitely does not work correctly with software cursors.
In CS2, the sensitivity is correct, but VRR is broken in a weird way...
It sends new frames on cursor move when the cursor is locked and does not send frames when the cursor is unlocked.
Could this be as simple as an inverted if statement? I looked for it and found a bunch of weird cases like >0 on a bool, but nothing that fixed it.

UjinT34:
`WAYLAND_DEBUG=1` and some HL logging might help. The hack exists because for whatever reason there is an extra unwanted wl surface frame event right after a pointer frame with locked cursor. Nothing says that it should be any different from unlocked cursor or that there should/shouldn't be such surface frame event. The code just accumulates mouse movement to send the result and a pointer frame right after the wanted surface frame and discard the next one surface frame. Without this hack there is no way to distinguish between wanted and unwanted surface frames. The proper fix should somehow avoid getting those extra surface frames but it's probably outside of HL's control.

Szwagi:
Two reasons for mouse motion scheduling frames when it shouldn't:
1. Scheduling new frames even though the damage box was empty. This scheduled frames for all mouse motion.
2. Using `monitor->shouldSkipScheduleFrameOnMouseEvent()` with `g_pHyprRenderer->damageBox(...)`... When the cursor is on the edge of the screen, `box.overlaps(...)` passes, secondary monitor does not want to skip the schedule, then `g_pHyprRenderer->damageBox` will schedule a frame for the primary monitor.

`no_break_fs_vrr=0` with software cursors is still broken, and I'm unsure if there's any hope for it.
In CS2, the cursor is not visible, but the cursor box is still 30x30, I'll investigate if I can check if any cursor is visible at all.

Admittedly that code is in need of some cleanup now, and I'm unsure if `g_pCompositor->m_unsafeState` is needed there (I think the 'if' checks shouldn't pass?).

Also, maybe the better way is to do this in `g_pHyprRenderer->damageBox`?
```diff
- if (!skipFrameSchedule)
+ if (!skipFrameSchedule && (!isMouseEvent || !monitor->shouldSkipScheduleFrameOnMouseEvent()))
```

Szwagi:
After more testing, the current status is (for things I tested):
- `native,   no_hardware_cursors = 0, no_break_fs_vrr = 0` - Works
- `native,   no_hardware_cursors = 0, no_break_fs_vrr = 1` - Unlocked cursors schedule extra frames
- `native,   no_hardware_cursors = 1, no_break_fs_vrr = 0` - Works
- `native,   no_hardware_cursors = 1, no_break_fs_vrr = 1` - Works
- `xwayland, no_hardware_cursors = 0, no_break_fs_vrr = 0` - Works
- `xwayland, no_hardware_cursors = 0, no_break_fs_vrr = 1` - Unlocked cursors schedule extra frames
- `xwayland, no_hardware_cursors = 1, no_break_fs_vrr = 0` - Works when it works, for 50% of game launches the locked cursors get warped back (?)
- `xwayland, no_hardware_cursors = 1, no_break_fs_vrr = 1` - Works when it works, for 50% of game launches the locked cursors get warped back (?)

Nosamdaman:
First of all, thank you @Szwagi for digging deep into this, this is quite literally the only major show-stopper preventing me from gaming on Hyprland. Secondly, I'm not quite able to replicate your above behavior matrix. For me, no matter what, on the latest version of your branch, I'm seeing stuttery camera movement on CS2 with a frame limiter set, though perhaps there's something I'm missing. That being said, It looks like Half-Life 2 works for me with `no_break_fs_vrr` enabled and an FPS limit below my refresh rate regardless of whether or not hardware cursors are enabled.

This seems like a real nightmare of an issue for you guys lol, if you want me to validate any further findings on my end let me know.

njdom24:
Dunno if there's any relation, but last I checked, software cursors disable direct scanout regardless of if they're visible or not. Could direct scanout be related to the difference here?

Nosamdaman:
> Dunno if there's any relation, but last I checked, software cursors disable direct scanout regardless of if they're visible or not. Could direct scanout be related to the difference here?

So I just re-tested this. With both direct scanout and hardware cursors enabled, playing Wayland native CS2 with an FPS limit of 140 on 144 VRR display, I'm still getting the occasional spike to 144 when moving the mouse, causing visible hitching.

Szwagi:
> So I just re-tested this. With both direct scanout and hardware cursors enabled, playing Wayland native CS2 with an FPS limit of 140 on 144 VRR display, I'm still getting the occasional spike to 144 when moving the mouse, causing visible hitching.

Can't reproduce on NVIDIA, could you share your config?
For me the only time hardware cursors do this is when the cursor is unlocked (which is fine with `no_break_fs_vrr = 1` but broken for `no_break_fs_vrr = 0`).

Worth noting that CS2's `fps_max` is unreliable, `engine_low_latency_sleep_after_client_tick` was meant to fix it but I don't think it works. Try mangohud's fps limiter.

Edit: ~~I'm able to reproduce it in MPV when playing back 24FPS footage... it spikes to 28FPS. I'll look into it~~
Never mind, it was `min_refresh_rate`... The default is 24FPS and the way it works isn't great so it spikes above that.

Nosamdaman:
> Worth noting that CS2's `fps_max` is unreliable, `engine_low_latency_sleep_after_client_tick` was meant to fix it but I don't think it works. Try mangohud's fps limiter.

Ah, I was unaware of that, switching to ManguHud's limiter did fix the issue on wayland. I _think_ I'm still seeing the issue with CS2 in xwayland mode, though it is far less frequent. That being said, it feels amazing in Wayland.

Szwagi:
I am unsure if https://github.com/hyprwm/Hyprland/pull/12665/commits/e4013b8efcaf1bc294ea1ae605a704b17836225a is entirely correct, but it should function the same or better than the frame schedule removed in https://github.com/hyprwm/Hyprland/pull/12665/commits/ee471d11cd06e5a6f68e84f010b293f9cf243917, and without introducing unnecessary frame schedules on mouse motion. From my testing it's only necessary for Firefox which seems to be broken (based on the amount of Firefox hacks)? It also fixes an issue with frames not being scheduled when the surface x/y is not on any monitor.

I didn't look hard for the remaining hardware cursor issue yet, but I suspect it might be a bug in Aquamarine.

Status to my eye:
- `native,   no_hardware_cursors = 0, no_break_fs_vrr = 0` - Works
- `native,   no_hardware_cursors = 0, no_break_fs_vrr = 1` - Unlocked cursors schedule extra frames
- `native,   no_hardware_cursors = 1, no_break_fs_vrr = 0` - Works
- `native,   no_hardware_cursors = 1, no_break_fs_vrr = 1` - Works
- `xwayland, no_hardware_cursors = 0, no_break_fs_vrr = 0` - Works
- `xwayland, no_hardware_cursors = 0, no_break_fs_vrr = 1` - Unlocked cursors schedule extra frames
- `xwayland, no_hardware_cursors = 1, no_break_fs_vrr = 0` - Works
- `xwayland, no_hardware_cursors = 1, no_break_fs_vrr = 1` - Works

Nosamdaman:
I just pulled and did a test of my own with the 8 combinations when `no_break_vs_vrr = 1` and they all seem to be working perfectly to my eyes, as in the games I've tested, the cursor is only unlocked in menus, thus making the extra frame-scheduling not noticeable.

This even fixed the mild VRR flicker I was getting. Good job again on this one, you have no idea how much sanity I was losing testing things before I saw this PR.

Szwagi:
https://github.com/hyprwm/Hyprland/pull/12665/commits/e60c98738809f89937ae27ebf24a5488a46f985a fixes framerate spiking when changing window focus caused by animations scheduling many thousands of frames per second for animations that aren't visible.

I did more research and I don't believe Hyprland is responsible for breaking VRR with hardware cursors and no_break_fs_vrr=1. With an unlocked cursor, Hyprland schedules exactly 125FPS on a 125FPS lock, but my monitor reports 250FPS or an unstable 330FPS coming in depending on the game.

It's possible there's still an extra frame scheduled on cursor visibility change. My reports spikes by 1-2FPS, but I couldn't see any extra frame schedules except for the unrelated https://github.com/hyprwm/Hyprland/pull/12665/commits/37ae61062b5829ea8aed5cd2f724924c9b7dc114, perhaps monitor getting confused by a lag spike in the game when opening the menu.

@vaxerski I think this is all for the issues I found. I will now daily this branch to see if something comes up, but would like to know what you think of this and what the glaring issues are :D

src/render/Renderer.cpp
```diff
@@ -1915,16 +1915,22 @@ void CHyprRenderer::damageSurface(SP<CWLSurfaceResource> pSurface, double x, dou
         return;
     }
 
-    if (scale != 1.0)
-        damageBox.scale(scale);
+    if (damageBox.empty()) {
+        const auto VIEW = WLSURF->view();
```
vaxerski: why is this

Szwagi: Some programs like Firefox get empty damageBox here even though they redrew the entire window, ignoring makes it freeze and update only on mouse motion. Old code used getMonitorFromVector(x, y) and scheduled a frame on that monitor, but that breaks when the x, y is not on any monitor, making it freeze

vaxerski: that doesnt sound correct. _why_ did firefox get an empty damageBox when it redrew something? It's either a hyprland bug (we didnt damage where we should've) or a ff bug (no damage sent)

This _needs_ to be investigated more, plastering a random-ass band-aid that doesn't follow how shit should be done is not a good idea

Szwagi: Firefox behavior is pretty strange, but it seems like it's a Hyprland bug. From what I can tell, Firefox sends damage only after a frame is scheduled for the monitor it's on. Moving the cursor on the monitor Firefox is on is enough for Firefox to send damage, even if it's not the focused window. 
I'm unsure what the correct thing to do is as I'm not very familiar with wayland protocols.

Szwagi: Looking through old commits, https://github.com/hyprwm/Hyprland/commit/beef23cee5cdc47142da50bf029f52bdff192025 stands out. The if statement was later removed in https://github.com/hyprwm/Hyprland/commit/016da234d0e852de3ef20eb2e89ac58d2a85f6e7, introducing unnecessary frame schedules. From a quick test, adding an equivalent if statement seems to work, but I think the `getMonitorFromVector` stuff needs to go in favor of `logicalBox` or similar.

gulafaran:
i dont think [aa533b8](https://github.com/hyprwm/Hyprland/pull/12665/commits/aa533b8f72891ebbf532930cedf495ec147a7c19) is right, wont this damage the entire logicalbox on every single commit now and basicly ignore if any damage was actually sent?

gulafaran:
> i dont think [aa533b8](https://github.com/hyprwm/Hyprland/pull/12665/commits/aa533b8f72891ebbf532930cedf495ec147a7c19) is right, wont this damage the entire logicalbox on every single commit now and basicly ignore if any damage was actually sent?

from the bits ive digged i think there is a frame callback issue in general in HL, ive tried doing something about it but im just not getting it 100% correct and the rabbits nest goes deep see https://github.com/hyprwm/Hyprland/pull/11454 , some clients like firefox seems to rely on it being sent on monitor pageflip pretty much. both for their vsync handling and for recieving hints on when to send the next buffer. currently its done on commit time and presentation. and eventually hits the renderer here https://github.com/hyprwm/Hyprland/blob/42447a50d6840c5e28bd58db1225bae2fd7d5ed0/src/render/Renderer.cpp#L1442 if no damage because of that scheduled monitor frame you want to remove. 

or it hits https://github.com/hyprwm/Hyprland/blob/42447a50d6840c5e28bd58db1225bae2fd7d5ed0/src/protocols/core/Compositor.cpp#L656 because of the scheduled frame.

Szwagi:
> from the bits ive digged i think there is a frame callback issue in general in HL, ive tried doing something about it but im just not getting it 100% correct and the rabbits nest goes deep see #11454 , some clients like firefox seems to rely on it being sent on monitor pageflip pretty much. both for their vsync handling and for recieving hints on when to send the next buffer. currently its done on commit time and presentation. and eventually hits the renderer here
> 
> https://github.com/hyprwm/Hyprland/blob/42447a50d6840c5e28bd58db1225bae2fd7d5ed0/src/render/Renderer.cpp#L1442
> if no damage because of that scheduled monitor frame you want to remove.
> 
> or it hits
> 
> https://github.com/hyprwm/Hyprland/blob/42447a50d6840c5e28bd58db1225bae2fd7d5ed0/src/protocols/core/Compositor.cpp#L656
> because of the scheduled frame.

Since this is a bigger issue that others are looking into, and the frame schedule hack was already there before my changes, would it be acceptable to keep the hack (adjusted for what this PR is looking to solve) and resolve that at a different time? @vaxerski 



> i dont think [aa533b8](https://github.com/hyprwm/Hyprland/pull/12665/commits/aa533b8f72891ebbf532930cedf495ec147a7c19) is right, wont this damage the entire logicalbox on every single commit now and basicly ignore if any damage was actually sent?

Right... I didn't quite think that through. My intention was to only schedule a frame, not add damage. Does something like this make more sense? I think this is pretty much what it was 3 years ago (https://github.com/hyprwm/Hyprland/commit/beef23cee5cdc47142da50bf029f52bdff192025), just without some issues related to `getMonitorFromVector`.
```Cpp
// schedule frame events (hack)
if (!WLSURF->resource()->m_current.callbacks.empty()) {
    const auto VIEW = WLSURF->view();
    const auto BOX  = VIEW->logicalBox();
    if (BOX.has_value() && !BOX->empty()) {
        for (auto const& m : g_pCompositor->m_monitors) {
            if (!m->m_output)
                continue;

            if (BOX->overlaps(m->logicalBox()))
                g_pCompositor->scheduleFrameForMonitor(m, Aquamarine::IOutput::AQ_SCHEDULE_NEEDS_FRAME);
        }
    }
}

CRegion damageBox = WLSURF->computeDamage();
if (damageBox.empty())
    return;
```

gulafaran:
> > from the bits ive digged i think there is a frame callback issue in general in HL, ive tried doing something about it but im just not getting it 100% correct and the rabbits nest goes deep see #11454 , some clients like firefox seems to rely on it being sent on monitor pageflip pretty much. both for their vsync handling and for recieving hints on when to send the next buffer. currently its done on commit time and presentation. and eventually hits the renderer here
> > https://github.com/hyprwm/Hyprland/blob/42447a50d6840c5e28bd58db1225bae2fd7d5ed0/src/render/Renderer.cpp#L1442
> > 
> > if no damage because of that scheduled monitor frame you want to remove.
> > or it hits
> > https://github.com/hyprwm/Hyprland/blob/42447a50d6840c5e28bd58db1225bae2fd7d5ed0/src/protocols/core/Compositor.cpp#L656
> > 
> > because of the scheduled frame.
> 
> Since this is a bigger issue that others are looking into, and the frame schedule hack was already there before my changes, would it be acceptable to keep the hack (adjusted for what this PR is looking to solve) and resolve that at a different time? @vaxerski
> 
> > i dont think [aa533b8](https://github.com/hyprwm/Hyprland/pull/12665/commits/aa533b8f72891ebbf532930cedf495ec147a7c19) is right, wont this damage the entire logicalbox on every single commit now and basicly ignore if any damage was actually sent?
> 
> Right... I didn't quite think that through. My intention was to only schedule a frame, not add damage. Does something like this make more sense? I think this is pretty much what it was 3 years ago ([beef23c](https://github.com/hyprwm/Hyprland/commit/beef23cee5cdc47142da50bf029f52bdff192025)), just without some issues related to `getMonitorFromVector`.
> 
> ```c++
> // schedule frame events (hack)
> if (!WLSURF->resource()->m_current.callbacks.empty()) {
>     const auto VIEW = WLSURF->view();
>     const auto BOX  = VIEW->logicalBox();
>     if (BOX.has_value() && !BOX->empty()) {
>         for (auto const& m : g_pCompositor->m_monitors) {
>             if (!m->m_output)
>                 continue;
> 
>             if (BOX->overlaps(m->logicalBox()))
>                 g_pCompositor->scheduleFrameForMonitor(m, Aquamarine::IOutput::AQ_SCHEDULE_NEEDS_FRAME);
>         }
>     }
> }
> 
> CRegion damageBox = WLSURF->computeDamage();
> if (damageBox.empty())
>     return;
> ```

something like that might work, as long as it doesnt show other regressions popping up because it was relying on that scheduled frame unintentionally

Arisa-Snowbell:
I rebased it on main and it crashes on startup, did I do something wrong?


```
#0  0x000055ffbc6bbfe7 n/a (/usr/bin/Hyprland (deleted) + 0x758fe7)
#1  0x00007fc369bc5150 n/a (libEGL_mesa.so.0 + 0x13150)
#2  0x00007fc369bc54bd n/a (libEGL_mesa.so.0 + 0x134bd)
#3  0x00007fc369bbce66 n/a (libEGL_mesa.so.0 + 0xae66)
#4  0x00007fc371e57376 _ZN10Aquamarine12CDRMRendererD2Ev (libaquamarine.so.9 + 0xcb376)
#5  0x00007fc371e6a16b _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine12CDRMRendererEE7_deleteEPv (libaquamarine.so.9 + 0xde16b)
#6  0x00007fc371e331b6 _ZN10Aquamarine11CDRMBackendD1Ev (libaquamarine.so.9 + 0xa71b6)
#7  0x00007fc371e4d64e _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine11CDRMBackendEE7_deleteEPv (libaquamarine.so.9 + 0xc164e)
#8  0x00007fc371e06862 _ZN10Aquamarine8CBackendD1Ev (libaquamarine.so.9 + 0x7a862)
#9  0x00007fc371e071bb _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine8CBackendEE7_deleteEPv (libaquamarine.so.9 + 0x7b1bb)
#10 0x00007fc371e22575 n/a (libaquamarine.so.9 + 0x96575)
#11 0x00007fc370e406bc __cxa_finalize (libc.so.6 + 0x406bc)
#12 0x00007fc371dea248 n/a (libaquamarine.so.9 + 0x5e248)
#13 0x00007fc371f74012 n/a (ld-linux-x86-64.so.2 + 0x2012)
#14 0x00007fc371f7816e n/a (ld-linux-x86-64.so.2 + 0x616e)
#15 0x00007fc370e40c71 n/a (libc.so.6 + 0x40c71)
#16 0x00007fc370e40d4e exit (libc.so.6 + 0x40d4e)
#17 0x00007fc370e2763c n/a (libc.so.6 + 0x2763c)
#18 0x00007fc370e276e9 __libc_start_main (libc.so.6 + 0x276e9)
#19 0x000055ffbc1ba6e5 n/a (/usr/bin/Hyprland (deleted) + 0x2576e5)
```

Edit: Wrong stack trace

Szwagi:
> I rebased it on main and it crashes on startup, did I do something wrong?
> 
> ```
> #0  0x000055ffbc6bbfe7 n/a (/usr/bin/Hyprland (deleted) + 0x758fe7)
> #1  0x00007fc369bc5150 n/a (libEGL_mesa.so.0 + 0x13150)
> #2  0x00007fc369bc54bd n/a (libEGL_mesa.so.0 + 0x134bd)
> #3  0x00007fc369bbce66 n/a (libEGL_mesa.so.0 + 0xae66)
> #4  0x00007fc371e57376 _ZN10Aquamarine12CDRMRendererD2Ev (libaquamarine.so.9 + 0xcb376)
> #5  0x00007fc371e6a16b _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine12CDRMRendererEE7_deleteEPv (libaquamarine.so.9 + 0xde16b)
> #6  0x00007fc371e331b6 _ZN10Aquamarine11CDRMBackendD1Ev (libaquamarine.so.9 + 0xa71b6)
> #7  0x00007fc371e4d64e _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine11CDRMBackendEE7_deleteEPv (libaquamarine.so.9 + 0xc164e)
> #8  0x00007fc371e06862 _ZN10Aquamarine8CBackendD1Ev (libaquamarine.so.9 + 0x7a862)
> #9  0x00007fc371e071bb _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine8CBackendEE7_deleteEPv (libaquamarine.so.9 + 0x7b1bb)
> #10 0x00007fc371e22575 n/a (libaquamarine.so.9 + 0x96575)
> #11 0x00007fc370e406bc __cxa_finalize (libc.so.6 + 0x406bc)
> #12 0x00007fc371dea248 n/a (libaquamarine.so.9 + 0x5e248)
> #13 0x00007fc371f74012 n/a (ld-linux-x86-64.so.2 + 0x2012)
> #14 0x00007fc371f7816e n/a (ld-linux-x86-64.so.2 + 0x616e)
> #15 0x00007fc370e40c71 n/a (libc.so.6 + 0x40c71)
> #16 0x00007fc370e40d4e exit (libc.so.6 + 0x40d4e)
> #17 0x00007fc370e2763c n/a (libc.so.6 + 0x2763c)
> #18 0x00007fc370e276e9 __libc_start_main (libc.so.6 + 0x276e9)
> #19 0x000055ffbc1ba6e5 n/a (/usr/bin/Hyprland (deleted) + 0x2576e5)
> ```
> 
> Edit: Wrong stack trace

Rebased my branch, tested in games and been using it for a few hours, didn't see any issues yet.
If you're sure this is because of my changes, your config would be useful.

Arisa-Snowbell:
> > I rebased it on main and it crashes on startup, did I do something wrong?
> > ```
> > #0  0x000055ffbc6bbfe7 n/a (/usr/bin/Hyprland (deleted) + 0x758fe7)
> > #1  0x00007fc369bc5150 n/a (libEGL_mesa.so.0 + 0x13150)
> > #2  0x00007fc369bc54bd n/a (libEGL_mesa.so.0 + 0x134bd)
> > #3  0x00007fc369bbce66 n/a (libEGL_mesa.so.0 + 0xae66)
> > #4  0x00007fc371e57376 _ZN10Aquamarine12CDRMRendererD2Ev (libaquamarine.so.9 + 0xcb376)
> > #5  0x00007fc371e6a16b _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine12CDRMRendererEE7_deleteEPv (libaquamarine.so.9 + 0xde16b)
> > #6  0x00007fc371e331b6 _ZN10Aquamarine11CDRMBackendD1Ev (libaquamarine.so.9 + 0xa71b6)
> > #7  0x00007fc371e4d64e _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine11CDRMBackendEE7_deleteEPv (libaquamarine.so.9 + 0xc164e)
> > #8  0x00007fc371e06862 _ZN10Aquamarine8CBackendD1Ev (libaquamarine.so.9 + 0x7a862)
> > #9  0x00007fc371e071bb _ZN9Hyprutils6Memory14CSharedPointerIN10Aquamarine8CBackendEE7_deleteEPv (libaquamarine.so.9 + 0x7b1bb)
> > #10 0x00007fc371e22575 n/a (libaquamarine.so.9 + 0x96575)
> > #11 0x00007fc370e406bc __cxa_finalize (libc.so.6 + 0x406bc)
> > #12 0x00007fc371dea248 n/a (libaquamarine.so.9 + 0x5e248)
> > #13 0x00007fc371f74012 n/a (ld-linux-x86-64.so.2 + 0x2012)
> > #14 0x00007fc371f7816e n/a (ld-linux-x86-64.so.2 + 0x616e)
> > #15 0x00007fc370e40c71 n/a (libc.so.6 + 0x40c71)
> > #16 0x00007fc370e40d4e exit (libc.so.6 + 0x40d4e)
> > #17 0x00007fc370e2763c n/a (libc.so.6 + 0x2763c)
> > #18 0x00007fc370e276e9 __libc_start_main (libc.so.6 + 0x276e9)
> > #19 0x000055ffbc1ba6e5 n/a (/usr/bin/Hyprland (deleted) + 0x2576e5)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Edit: Wrong stack trace
> 
> Rebased my branch, tested in games and been using it for a few hours, didn't see any issues yet. If you're sure this is because of my changes, your config would be useful.

Recompiled every part of hyprland ecosystem, used your branch to compile hyprland itself, removed my config, happens with default config too

Arisa-Snowbell:
```
Hyprland received signal 11(SEGV)
Version: 4131c67d97201ad82975a3de86f5c11576cf7ea1
Tag: v0.52.0-206-g4131c67d9
Date: Sun Dec 28 18:46:48 2025
Flags:

System info:
	System name: Linux
	Node name: flower
	Release: 6.19.0-rc2-1-mainline-00174-g33c8f4cc2c69
	Version: #100 SMP PREEMPT_DYNAMIC Sun, 28 Dec 2025 11:57:18 +0000

GPU:
	03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 48 [Radeon RX 9070/9070 XT/9070 GRE] [1002:7550] (rev c0) (prog-if 00 [VGA controller])
11:00.0 VGA compatible controller [0300]: Intel Corporation DG2 [Arc Pro A40/A50] [8086:56b1] (rev 05) (prog-if 00 [VGA controller])


os-release:
	NAME="Arch Linux"
	PRETTY_NAME="Arch Linux"
	ID=arch
	BUILD_ID=rolling
	ANSI_COLOR="38;2;23;147;209"
	LOGO=archlinux-logo

Libraries:
Hyprgraphics: built against 0.5.0, system has 0.5.0
Hyprutils: built against 0.11.0, system has 0.11.0
Hyprlang: built against 0.6.7, system has 0.6.7
Aquamarine: built against 0.10.0, system has 0.10.0

Backtrace:
	# | Hyprland(_Z12getBacktracev+0x64) [0x55bbf48a7804]
		getBacktrace()
		/usr/src/debug/hyprland-git/hyprland/src/helpers/MiscFunctions.cpp:758
	#1 | Hyprland(_ZN13CrashReporter18createAndSaveCrashEi+0xd5f) [0x55bbf480cfaf]
		std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_Alloc_hider::_Alloc_hider(char*, std::allocator<char> const&)
		/usr/include/c++/15.2.1/bits/basic_string.h:205 (discriminator 1)
	#2 | Hyprland(+0x26d342) [0x55bbf475d342]
		handleUnrecoverableSignal(int)
		/usr/src/debug/hyprland-git/hyprland/src/Compositor.cpp:121
	#3 | /usr/lib/libc.so.6(+0x3e4d0) [0x7f53db63e4d0]
		??
		??:0
	#4 | Hyprland(_ZN7Desktop4View7CWindow14getContentTypeEv+0x1e) [0x55bbf486a8de]
		Hyprutils::Memory::CSharedPointer<Desktop::View::CWLSurface>::operator bool() const
		/usr/include/hyprutils/memory/SharedPtr.hpp:110
	#5 | Hyprland(_ZN8CMonitor35shouldSkipScheduleFrameOnMouseEventEv+0x1db) [0x55bbf48b81cb]
		CMonitor::shouldSkipScheduleFrameOnMouseEvent()
		/usr/src/debug/hyprland-git/hyprland/src/helpers/Monitor.cpp:1030 (discriminator 20)
	#6 | Hyprland(_ZN13CInputManager16mouseMoveUnifiedEjbbSt8optionalIN9Hyprutils4Math8Vector2DEE+0x56b) [0x55bbf4a007ab]
		CInputManager::mouseMoveUnified(unsigned int, bool, bool, std::optional<Hyprutils::Math::Vector2D>)
		/usr/src/debug/hyprland-git/hyprland/src/managers/input/InputManager.cpp:247 (discriminator 2)
	#7 | Hyprland(_ZN11CCompositor12onNewMonitorEN9Hyprutils6Memory14CSharedPointerIN10Aquamarine7IOutputEEE+0xcf1) [0x55bbf477aa21]
		CCompositor::onNewMonitor(Hyprutils::Memory::CSharedPointer<Aquamarine::IOutput>)
		/usr/src/debug/hyprland-git/hyprland/src/Compositor.cpp:2887 (discriminator 2)
	#8 | Hyprland(_ZN11CCompositor10initServerENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEi+0xfac) [0x55bbf478544c]
		Hyprutils::Memory::Impl_::impl_base::dec()
		/usr/include/hyprutils/memory/ImplBase.hpp:22
	#9 | Hyprland(main+0xf1e) [0x55bbf46e7e9e]
		std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_data() const
		/usr/include/c++/15.2.1/bits/basic_string.h:239
	#1 | /usr/lib/libc.so.6(+0x27635) [0x7f53db627635]
		??
		??:0
	#11 | /usr/lib/libc.so.6(__libc_start_main+0x89) [0x7f53db6276e9]
		??
		??:0
	#12 | Hyprland(_start+0x25) [0x55bbf4747525]
		_start
		??:?


Log tail:
DEBUG from aquamarine ]: GBM: Allocated a new buffer with size [Vector2D: x: 2560, y: 1440] and format XR24 with modifier 144115188076389125 aka GFX12,64KB_2D,DCC,DCC_MAX_COMPRESSED_BLOCK=256B
DEBUG from aquamarine ]: GBM: Allocated a new buffer with size [Vector2D: x: 2560, y: 1440] and format XR24 with modifier 144115188076389125 aka GFX12,64KB_2D,DCC,DCC_MAX_COMPRESSED_BLOCK=256B
DEBUG from aquamarine ]: GBM: Allocated a new buffer with size [Vector2D: x: 2560, y: 1440] and format XR24 with modifier 144115188076389125 aka GFX12,64KB_2D,DCC,DCC_MAX_COMPRESSED_BLOCK=256B
DEBUG from aquamarine ]: Swapchain: Reconfigured a swapchain to [Vector2D: x: 2560, y: 1440] XR24 of length 3
DEBUG from aquamarine ]: drm: Modesetting DP-6 with 2560x1440@240.00Hz

```

src/helpers/Monitor.cpp
```diff
@@ -1026,8 +1026,9 @@ bool CMonitor::shouldSkipScheduleFrameOnMouseEvent() {
     static auto PMINRR   = CConfigValue<Hyprlang::INT>("cursor:min_refresh_rate");
 
     // skip scheduling extra frames for fullsreen apps with vrr
-    const bool shouldSkip = inFullscreenMode() && (*PNOBREAK == 1 || (*PNOBREAK == 2 && m_activeWorkspace->getFullscreenWindow()->getContentType() == CONTENT_TYPE_GAME)) &&
-        m_output->state->state().adaptiveSync;
+    const bool shouldRenderCursor = g_pHyprRenderer->shouldRenderCursor();
+    const bool noBreak            = (*PNOBREAK == 1 || (*PNOBREAK == 2 && m_activeWorkspace->getFullscreenWindow()->getContentType() == CONTENT_TYPE_GAME));
```
UjinT34: `m_activeWorkspace->getFullscreenWindow()` can be null. Previously it was guarded by `inFullscreenMode()` check.

Arisa-Snowbell:
Yup, that fixes it

fxzzi:
works ok for me, however refresh rate on my monitor still goes to its max when the cursor is visible and moving, even though the cursor is at a lower fps

Szwagi:
> works ok for me, however refresh rate on my monitor still goes to its max when the cursor is visible and moving, even though the cursor is at a lower fps

Is that with hardware cursors? What GPU vendor?

Hardware cursors schedule extra frames even when the cursor is moved on a completely different monitor.
I had a look at Aquamarine before and didn't see anything suspect, so in my case it could be NVIDIA.

If it's a GPU driver issue, checking if any monitor is running VRR when `cursor:no_hardware_cursors = 2` might be the only good option.

fxzzi:
indeed, nvidia hw cursor

UjinT34:
I thought they've fixed it not so long ago because hw cursors with nvidia do not break vrr for me. Might be some specific driver and egl- libs combo that fixes it.

fxzzi:
im on latest nvidia 590, pr works ok as long as cursor isn't visible.

Szwagi:
> im on latest nvidia 590, pr works ok as long as cursor isn't visible.

What about software cursors? They should work perfectly. https://github.com/hyprwm/Hyprland/pull/12665#issuecomment-3678851062

Also, I assume you understand this, but VRR is supposed to break with `cursor:no_break_fs_vrr 0` when the cursor is visible.

fxzzi:
> Also, I assume you understand this, but VRR is supposed to break with cursor:no_break_fs_vrr 0 when the cursor is visible.

I am aware, I tested this PR with this at 1

fxzzi:
maybe related but with hw cursor, if I move the mouse on my other monitor it still deactivates VRR on my main, does this mean nvidia is at fault here? Works fine with sw cursor

UjinT34:
nvidia-drm or some of nvidia libs. Doesn't happen to me now with latest drivers and egl-wayland2 but it happened with earlier versions. Whenever there is an atomic commit with any changed properties on the cursor plane nvidia seems to forget that vrr is active. Even when the vrr activation flag is included in the same commit. Might be an intended behaviour or some nvidia reflex shenanigans.

fxzzi:
590 drivers should be using egl-wayland2 by default. Issue still happens for me. I can see it very easily when mpv is paused on my main monitor, and i move my mouse on my secondary:

```
~ $ cd /run/opengl-driver/share/egl/egl_external_platform.d 
share/egl/egl_external_platform.d $ ll
Permissions Size User Date Modified Name
.r--r--r--   178 root  1 Jan  1970   10_nvidia_wayland.json
.r--r--r--   174 root  1 Jan  1970   15_nvidia_gbm.json
.r--r--r--   174 root  1 Jan  1970   20_nvidia_xcb.json
.r--r--r--   175 root  1 Jan  1970   20_nvidia_xlib.json
.r--r--r--   179 root  1 Jan  1970   99_nvidia_wayland2.json
```

Szwagi:
> maybe related but with hw cursor, if I move the mouse on my other monitor it still deactivates VRR on my main, does this mean nvidia is at fault here? Works fine with sw cursor

My assumption is that it is NVIDIA, but I haven't seen anyone test this on AMD yet.
I am seeing the exact same behavior on the Cosmic compositor.

No other compositor lets the user force hardware cursors, so maybe adding a check for any monitor running VRR into `cursor:no_hardware_cursors = 2` is a reasonable solution for now?

Nosamdaman:
> No other compositor lets the user force hardware cursors, so maybe adding a check for any monitor running VRR into `cursor:no_hardware_cursors = 2` is a reasonable solution for now?

So long as that behavior is configurable, I'm fine with it.

s0mebodyhelpme:
Looking forward to see this get merged. Unfortunately does not seem to build on main starting with 5b1b79c29c5e0ea974b2a9da5d122dd0f3bedca6.

fxzzi:
> No other compositor lets the user force hardware cursors, so maybe adding a check for any monitor running VRR into `cursor:no_hardware_cursors = 2` is a reasonable solution for now?

that can work, use `isNvidia()`and disable hw cursor in VRR scenarios. Wish nvidia would just fix this tho. I am interested in how it is working for Ujin, I tried everything, I removed egl-wayland1 and 2 bundled in driver, built it myself instead with on change

UjinT34:
I am on gentoo. Maybe some other sw or even compilation flags make a difference. I am not sure when it started working.
/etc/modprobe.d/nvidia.conf has
```
options nvidia-drm modeset=1 fbdev=1
options nvidia-modeset disable_vrr_memclk_switch=1
options nvidia \
        NVreg_DeviceFileGID=27 \
        NVreg_DeviceFileMode=432 \
        NVreg_DeviceFileUID=0 \
        NVreg_ModifyDeviceFiles=1 \
        NVreg_PreserveVideoMemoryAllocations=1 \
        NVreg_TemporaryFilePath=/tmp \
        NVreg_EnableS0ixPowerManagement=0 \
        NVreg_EnableResizableBar=1 \
        NVreg_UsePageAttributeTable=1 \
        NVreg_RegistryDwords=RMIntrLockingMode=1 \
        NVreg_RegistryDwords=RmEnableAggressiveVblank=1

```

Maybe disable_vrr_memclk_switch, NVreg_RegistryDwords=RMIntrLockingMode=1 or NVreg_RegistryDwords=RmEnableAggressiveVblank=1 make some difference. Those are relatively new settings.

fxzzi:
rmintrlockingmode is default now afaik, and i don't have vrr memclk switch enabled. below are my cmdline for nvidia for reference, I'll try disable aggressive vblank (does this even do anything with vrr on / vsync disabled), disable rmintrlockingmode, and enable vrr memclk switch

```
nvidia.NVreg_UsePageAttributeTable=1
nvidia.NVreg_EnableResizableBar=1
nvidia.NVreg_RegistryDwords=RmEnableAggressiveVblank=1
nvidia.NVreg_TemporaryFilePath=/var/tmp
nvidia.NVreg_EnableS0ixPowerManagement=0
video=DP-2:2560x1440-30@120
video=DP-3:2560x1440-30@280
nvidia-drm.modeset=1
nvidia-drm.fbdev=1
nvidia.NVreg_PreserveVideoMemoryAllocations=1
nvidia.NVreg_OpenRmEnableUnsupportedGpus=1
```

edit: didn't help rip

UjinT34:
I am on 5080. Maybe it's a hw/gsp firmware thing?
While I was on 3080 with some older drivers enabling/disabling gsp didn't change anything.

fxzzi:
I am using GSP firmware, let me try and disable it. Definitely does something on my 3070. PC actually wakes from sleep when it disable it lol 

edit: no change with gsp off

Szwagi:
I'm still looking for someone that can confirm this doesn't happen on AMD before I add the additional check to `cursor:no_hardware_cursors = 2`.

I am running a 4080 with no iGPU, so at least I know it's not a multi-gpu issue.
The suggested kernel params didn't change anything for me.

Just to make sure we're checking for the same thing:
I have 2 monitors, 1 of them supports VRR and the other does not.
Moving the hardware cursor on the monitor without VRR schedules frames on the VRR monitor, and I confirm it by looking at the FPS counter built into the ASUS monitor.

UjinT34:
> Moving the hardware cursor on the monitor without VRR schedules frames on the VRR monitor, and I confirm it by looking at the FPS counter built into the ASUS monitor.

Are you sure that frames are scheduled? If you see a rate spike in OSD then it might be caused by the videodriver without any request from HL/AQ. In this case we can't fix it on our side. `AQ_TRACE=1 should help` to verify that (some extra logging patch might be needed)

Szwagi:
> > Moving the hardware cursor on the monitor without VRR schedules frames on the VRR monitor, and I confirm it by looking at the FPS counter built into the ASUS monitor.
> 
> Are you sure that frames are scheduled? If you see a rate spike in OSD then it might be caused by the videodriver without any request from HL/AQ. In this case we can't fix it on our side. `AQ_TRACE=1 should help` to verify that (some extra logging patch might be needed)

By scheduled I meant the display reports a new frame, not that Hyprland or Aquamarine (explicitly) schedule a new frame. A new frame is reported by the VRR monitor only by updating the props on the cursor plane on the secondary non-VRR monitor.

njdom24:
~~Just tested on my 9070 XT. Moving the mouse on my secondary monitor keeps my primary monitor around its max refresh rate.~~

**Edit:** I fullscreened and un-fullscreened my terminal on my secondary monitor and the issue stopped happening. My primary's refresh rate remained stable, and I can't reproduce anymore, even when restarting Hyprland.

Szwagi:
Rebased it on current main.

I initially wanted to implement https://github.com/hyprwm/Hyprland/pull/12665/commits/9ccb4ae06de55a133798375a25f7a594db2745f3 by checking if any *other* monitor has VRR active, so it'd still use the hardware cursor in a lot of common cases. That ended up breaking VRR when the cursor was on the boundary between two monitors. I'll look into why that happens.

> I am on 5080. Maybe it's a hw/gsp firmware thing? While I was on 3080 with some older drivers enabling/disabling gsp didn't change anything.

Do you have `cursor:use_cpu_buffer` disabled or anything else that could make Hyprland switch back to software cursors automatically? I use `hyprpicker -r -z` to see what Hyprland is actually using (software cursors get captured by the screenshot).

Szwagi:
@vaxerski Opinion?

Have been using this branch daily for desktop use and gaming (with and without VRR). I didn't encounter any more issues on NVIDIA.

Things to note:
- Frame scheduling hack remains in the code, but was slightly rewritten to not break VRR. I think https://github.com/hyprwm/Hyprland/pull/11454 is a better place to remove the hack.
- Hardware cursors are disabled on NVIDIA whenever VRR is active on any monitor (`cursor:no_hardware_cursors = 2`). Changing cursor plane props on any monitor breaks VRR. I see the same thing happening on COSMIC. Comments in here suggest this is not an issue on AMD. A better solution could be some more sophisticated hack in Aquamarine, I tried some things but nothing really worked.

Not perfect, but *significantly* more usable for gaming than before. Potentially ready to merge?

vaxerski:
```
cbeb6984e748029ff481d5581d7e4f5279fd3d1a is the first bad commit
commit cbeb6984e748029ff481d5581d7e4f5279fd3d1a
Author: Szwagi <12988954+Szwagi@users.noreply.github.com>
Date:   Sat Jan 31 13:37:01 2026 +0000

    renderer: fix mouse motion in VRR (#12665)
```

maximized windows' border animation causes repaint artifacts

njdom24:
Seems I was wrong about my report on multi-monitor working fine on AMD.
Playing Final Fantasy XIV, moving the camera locks the cursor, and that keeps my refresh rate maxed.

It works fine if I disable my secondary monitor. Software cursors don't fix it.

**EDIT:** This is using native Wayland. Gamescope cursor capture doesn't have this issue, nor does native Wayland under Sway.

Arisa-Snowbell:
When I get out of bed I’m gonna test it on multi-monitor AMD setup.

Arisa-Snowbell:
I have two monitors, one 165hz, main one the game is on 240hz.
Used direct scanout.
Used hardware cursor cause software one will cause direct scanout to be blocked when on screen.
Doesn't matter if no_break_fs_vrr is 0 or 1, when cursor unlocked and moved it sends the refresh rate to 240, when locked VRR, or cursor not moving then it is on point. Seems correct.
Both XWayland and Wayland.
Mouse movement seems consistent across different settings.

Edit: I don't know accurate my monitor OSD is and such, so take with bit of salt, seems finicky.

njdom24:
Hm, it does look direct scanout related. At least with `cm=hdr`, I'm seeing this issue with Proton-wayland when direct scanout is disabled/blocked. Cursors are otherwise fine. Not having this problem in Gamescope, though.

Sorry, false alarm. I might be seeing a form of https://github.com/hyprwm/Hyprland/discussions/12423

UjinT34:
DS and tearing shouldn't affect this directly. They just have different sw/hw cursors requirements.
With sw cursors `no_break_fs_vrr=1` shouldn't break VRR. If it does then it's a HL issue that can be solved (ugly hacks might be required). Invisible cursors should work the same.
With visible hw it might break no matter what we do in HL/AQ because some driver&videocard combos have a bug with atomic cursor commit and vrr.

src/config/ConfigManager.cpp
```diff
@@ -3017,7 +3017,7 @@ bool CConfigManager::shouldUseSoftwareCursors(PHLMONITOR pMonitor) {
     switch (*PNOHW) {
         case 0: return false;
         case 1: return true;
-        case 2: return g_pHyprRenderer->isNvidia() && g_pHyprRenderer->isMgpu();
+        case 2: return g_pHyprRenderer->isNvidia() && (g_pHyprRenderer->isMgpu() || g_pCompositor->isVRRActiveOnAnyMonitor());
```
fxzzi, line 3020: this check seems wrong to me.

if you have vrr=1 at all, then this will just force sw cursor all the time, it should only do it if vrr is active + fullscreen afaik

Szwagi: > this check seems wrong to me.
> 
> if you have vrr=1 at all, then this will just force sw cursor all the time, it should only do it if vrr is active + fullscreen afaik

On my setup moving a hardware cursor on the second monitor makes the VRR monitor go black (framerate spike switches mode to non-vrr for a second). That is not usable.

The condition could definitely be better (especially in case of 1 monitor).
I have tried switching to software cursor for the secondary monitor, but it's very pretty wonky on the edges of the monitors.

Ultimately I want to look into making hardware cursors work correctly on NVIDIA.

fxzzi: Sorry to hear that your monitor modesets when going in and out of VRR, but that is definitely not the norm. This PR was also specific to fixing the variable `no_break_fs_vrr` which is specific to FS.

Anyway I've just set no hw cursor to 0 for now

Szwagi: > Sorry to hear that your monitor modesets when going in and out of VRR, but that is definitely not the norm. This PR was also specific to fixing the variable `no_break_fs_vrr` which is specific to FS.
> 
> Anyway I've just set no hw cursor to 0 for now

It was a thing on 2/2 ASUS monitors I owned, so I think it's pretty common. Not a very reputable brand so I'd believe if you said it's not a thing on others.

The PR ended up being about removing most instances of VRR breaking (not just the ones caused by removal of mouse input hack), that's why the hardware cursor check was added too... hopefully temporarily.

njdom24: AFAIK the AMDGPU driver behaves such that, when a display supports VRR, it's always sending a VRR signal. Even with VRR disabled, it just keeps the FPS locked at max, so it doesn't need to modeset.

NVIDIA probably doesn't do that. In which case it might be necessary to emulate that behavior within Hyprland.

fxzzi: > NVIDIA probably doesn't do that.

nvidia also does that unless you use a specific kernel param to outright disable vrr

PR #12674, opened 2025-12-18
fix: prevent pixman crashes and ensure valid color management metadata

Watchdog0x:
## Description
This PR addresses two related issues causing crashes in clients (specifically Chromium/Electron-based applications) due to invalid data being passed from the compositor.

### 1. Fix Invalid Pixman Rectangles (Monitor.cpp)
In `CMonitor::addDamage(const CBox& box)`, a guard clause was added to ignore boxes with non-positive dimensions. 
Previously, passing a box with `width <= 0` or `height <= 0` to `pixman_region32_init_rect` caused an assertion failure or crash with the error:
`*** BUG *** In pixman_region32_init_rect: Invalid rectangle passed`

### 2. Fix Incomplete Image Description (ColorManagement.cpp)
Updated `CColorManagementImageDescriptionInfo` and scRGB creation to ensure valid metadata is always sent to clients:
* **Fallback Primaries:** If `primaries` are invalid (e.g., zeroed out), the protocol now sends standard sRGB primaries instead of sending invalid data.
* **scRGB Luminance:** Hardcoded valid default luminance values (min: 0.0001, ref: 80, max: 600) for Windows scRGB generation to satisfy strict client validation.

## Impact
This fixes crashes observed in applications logging:
* `[wayland_wp_image_description.cc] Incomplete image description info from compositor`
* `pixman_region32_init_rect: Invalid rectangle passed`

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1017,6 +1017,10 @@ void CMonitor::addDamage(const CRegion& rg) {
 }
 
 void CMonitor::addDamage(const CBox& box) {
+
+    if (box.width <= 0 || box.height <= 0)
+        return;
+
     if (m_cursorZoom->value() != 1.f && g_pCompositor->getMonitorFromCursor() == m_self) {
         m_damage.damageEntire();
         g_pCompositor->scheduleFrameForMonitor(m_self.lock(), Aquamarine::IOutput::AQ_SCHEDULE_DAMAGE);
--- a/src/protocols/ColorManagement.cpp
+++ b/src/protocols/ColorManagement.cpp
@@ -5,10 +5,13 @@
 #include "core/Output.hpp"
 #include "types/ColorManagement.hpp"
 #include <cstdint>
+#include <algorithm> 
+#include <utility>  
 
 using namespace NColorManagement;
 
 const auto PRIMARIES_SCALE = 1000000.0f;
+
 
 CColorManager::CColorManager(SP<CWpColorManagerV1> resource) : m_resource(resource) {
     if UNLIKELY (!good())
@@ -62,10 +65,12 @@ CColorManager::CColorManager(SP<CWpColorManagerV1> resource) : m_resource(resour
         m_resource->sendSupportedIntent(WP_COLOR_MANAGER_V1_RENDER_INTENT_RELATIVE_BPC);
     }
 
-    m_resource->setDestroy([](CWpColorManagerV1* r) { LOGM(TRACE, "Destroy WP_color_manager at {:x} (generated default)", (uintptr_t)r); });
+    m_resource->setDestroy([](CWpColorManagerV1* r) { 
+        LOGM(TRACE, "Destroy WP_color_manager at {:x} (generated default)", (uintptr_t)r); 
+    });
+
     m_resource->setGetOutput([](CWpColorManagerV1* r, uint32_t id, wl_resource* output) {
         LOGM(TRACE, "Get output for id={}, output={}", id, (uintptr_t)output);
-
         const auto OUTPUTRESOURCE = CWLOutputResource::fromResource(output);
 
         const auto RESOURCE = PROTO::colorManagement->m_outputs.emplace_back(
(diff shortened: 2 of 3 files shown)


src/helpers/Monitor.cpp
```diff
@@ -1017,6 +1017,11 @@ void CMonitor::addDamage(const CRegion& rg) {
 }
 
 void CMonitor::addDamage(const CBox& box) {
+
+    // Block invalid rectangle passed
```
Honkazel: Useless AI comments be like...

src/protocols/ColorManagement.cpp
```diff
@@ -5,11 +5,14 @@
 #include "core/Output.hpp"
 #include "types/ColorManagement.hpp"
 #include <cstdint>
+#include <algorithm> // For std::clamp, std::ranges
+#include <utility>   // For std::to_underlying (C++23)
```
Honkazel: Same for those two

src/protocols/ColorManagement.hpp
```diff
@@ -142,26 +148,40 @@ class CColorManagementParametricCreator {
...
-    wl_client*                           client();
+    [[nodiscard]] bool                   good() const;
+    wl_client* client() const;
     SP<CWpImageDescriptionV1>            resource();
 
     WP<CColorManagementImageDescription> m_self;
 
     NColorManagement::SImageDescription  m_settings;
 
+    // Nu kender compileren CColorManagementImageDescriptionInfo pga. forward declaration i toppen
```
Honkazel: AI did switch to deutch lang for comment...

src/protocols/ColorManagement.cpp
Watchdog0x: Sorry no sleep :( AI for the win

src/helpers/Monitor.cpp
Watchdog0x: Guard Clause

src/protocols/ColorManagement.cpp
Watchdog0x: code uses Ranges modern C++20 way
and std::to_underlying in CColorManagementParametricCreator to do math on these enums

src/protocols/ColorManagement.hpp
Watchdog0x: This is just something random comment

src/helpers/Monitor.cpp
Watchdog0x: Root Cause when Hyprland tried to "damage" (redraw) that window, it passed those 0,0 dimensions to addDamage

Honkazel: I asked just for deleting this comment
I didn't complain about code

src/protocols/ColorManagement.cpp
Honkazel: I mean those two comments shouldn't be here anyway
Vaxry prefers comments only for stuff, that is too hard to understand by just reading the code or for 1337 IQ hacks
Those includes aren't those cases anyway, so it's better to delete those two comments

src/protocols/ColorManagement.hpp
Honkazel: And that's why it's better just to remove this comment? Thanks in advance

Watchdog0x:
This update adopts modern C++ standards (C++17/20/23) to strengthen the security and robustness of the Color Management protocol implementation through safer type manipulation. 

It also introduces additional security features, such as applying const correctness to methods and variables to avoid unintended state changes.

UjinT34:
Other CM changes are mostly some refactoring and code style changes that shouldn't affect how the proto works but might not match the rest of the codebase.

src/helpers/Monitor.cpp
Watchdog0x: Sorry, my brain is overtired. I’ll delete it.

src/protocols/ColorManagement.cpp
Watchdog0x: I have delete it.

Watchdog0x:
Don’t merge this. Code is a piece of shit (confirmed by bugs).

PR #12676, opened 2025-12-18
Desktop/history: Move history to desktop

vaxerski:
Moves and rewrites history handling

needs some testing for weird "prev" and stuff setups


Aqa-Ib:
`focuscurrentorlast` dispatcher broke here.

vaxerski:
you had a lot of time to let me know about that.

vaxerski:
also it works on my end

Aqa-Ib:
It works but not as before. It goes through the whole history, while previously it only went to the last window in the history.

vaxerski:
it cycles the last two for me

Aqa-Ib:
True, clean recompiling and now it works. Nice.

PR #12680, opened 2025-12-18
general: add screen cast border indicators

xavi-b:

Add options to enable and set a color around windows or monitors when they are screencasted.

I did not know how to name the options and/or which categories to put them in.

Options names and categories could be changed.


vaxerski (commented):
it will show up in the recording. Idk if this is a good idea. 

Furthermore, what about screenshots? You should probably restrict this to the _flow_ of screencopy

xavi-b:
> You should probably restrict this to the flow of screencopy

Sorry to bother you but could you explain a bit more ?

So far, I have only found a way to add "layers" on what the screen copy will render.
But I could not find a way to render something only on the monitor but not on the screen copy.

Any hints ?

vaxerski:
you cant in a simple way atm

PR #12693, opened 2025-12-21
CM updates

UjinT34:

- Sets default SDR to gamma22 since most clients and hw use it with sRGB. `render:cm_sdr_eotf = 3` to use sRGB TF.
- Marks `WP_COLOR_MANAGER_V1_FEATURE_SET_MASTERING_DISPLAY_PRIMARIES` and `WP_COLOR_MANAGER_V1_FEATURE_EXTENDED_TARGET_VOLUME`. This support isn't fully correct, mastering primaries should be used to determine color range for tonemapping and dealing with values outside 0.0 - 1.0 range. Should make gamescope and some other clients happier.
- Removes xx-cm and frog-cm because they aren't supported, have some HDR hacks that will mess with SDR. Some clients prefer those protos over wp-cm and hit some bugs.
- Makes primaries and image descriptions immutable to improve their id generation and conversion caching.

Will make some colors look different by default. Border shader might need a revert or a TF parameter.

Ready

### DIFF
--- a/src/protocols/ColorManagement.cpp
+++ b/src/protocols/ColorManagement.cpp
@@ -285,7 +285,7 @@ wl_client* CColorManagementSurface::client() {
 const SImageDescription& CColorManagementSurface::imageDescription() {
     if (!hasImageDescription())
         LOGM(Log::WARN, "Reading imageDescription while none set. Returns default or empty values");
-    
+
     return m_imageDescription->value();
 }
 


vaxerski:
border shader should be treated as srgb, no? Why would th ecolors shift?

njdom24:
Should there be a difference between `render:cm_sdr_eotf = 0` and `render:cm_sdr_eotf = 1` with these changes?

UjinT34:
> Should there be a difference between `render:cm_sdr_eotf = 0` and `render:cm_sdr_eotf = 1` with these changes?

`0` means default which might be changed in the future

UjinT34:
> border shader should be treated as srgb, no? Why would th ecolors shift?

It uses sRGB TF on main and gamma22 with this PR. It's not clear which TF should be used. Looks like a lot of real-world stuff assumes sRGB colorspace with gamma22 TF or even mislabels gamma22 TF as an sRGB TF.

UjinT34:
```
vm-test-run-hyprland-tests> Testing process spawning
vm-test-run-hyprland-tests> Passed: getFromSocket("/dispatch exec sleep 1"). Got ok
vm-test-run-hyprland-tests> Sleep was not spawned or several sleeps are running: pgrep returned ''
vm-test-run-hyprland-tests> Failed: fn(), expected true, got false. Source: hyprtester/src/main.cpp@234.
```
Is it expected to fail?

vaxerski:
no, force push to rerun

vaxerski (approved):
wiki mr needed

PR #12695, opened 2025-12-21
Fix master placement with center_ignores_reserved

UjinT34:
Makes `center_ignores_reserved` respect left and top outer gaps and top reserved area.

Ready

### DIFF
--- a/src/layout/MasterLayout.cpp
+++ b/src/layout/MasterLayout.cpp
@@ -349,10 +349,11 @@ void CHyprMasterLayout::calculateWorkspace(PHLWORKSPACE pWorkspace) {
     static auto  PIGNORERESERVED     = CConfigValue<Hyprlang::INT>("master:center_ignores_reserved");
     static auto  PSMARTRESIZING      = CConfigValue<Hyprlang::INT>("master:smart_resizing");
 
-    const auto   MASTERS      = getMastersOnWorkspace(pWorkspace->m_id);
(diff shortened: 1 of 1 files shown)


src/config/ConfigManager.cpp
```diff
@@ -3124,3 +3124,10 @@ std::optional<Vector2D> CConfigManager::getStoredFloatingSize(PHLWINDOW window)
     }
     return std::nullopt;
 }
+
+CCssGapData CConfigManager::getGapsForWorkspace(const PHLWORKSPACE& pWorkspace) {
```
vaxerski: No, layout should not use gaps. Layout should use the work area on a given display

UjinT34: How do you place a window in a center of a monitor if workarea size and placement doesn't match the monitor? E.g. when 20% on the left is reserved

vaxerski: you should center it within the work area.

UjinT34: `center_ignores_reserved` centers master window on monitor and ignores reserved space on the left to keep master window in the center of the monitor and avoid moving and resizing it when reserved area changes. Workarea doesn't provide enough information to do that.
Just reserve a good chunk of the monitor area on the left and watch "centered" window drift to the right side without `center_ignores_reserved`. I don't like this behaviour.

This PR fixes regression introduced in #12383

vaxerski: well then this should use work area + full monitor area. Gaps are unreliable.

PR #12711, opened 2025-12-23
cm: add ICC profile loading

vaxerski:
Adds loading of icc profiles for the properties we support

ref #9064

### Usage

v2: `icc = /full/path`
v1: `, icc, /full/path`

### Info

This also changes how we handle CM in general, where our intermediate buffer (offload) is in sRGB, and we CM at the end.

This should also make it possible to have actual, proper pixel-perfect screencopy in a follow-up MR (as we can copy the offloaded fb for screencopy purposes) - But that's for a follow-up MR

Requires:
- [ ] HDR offs ICC
- [ ] Testing!!!
- [x] Fix fucking render glitches XDDDD
- [ ] Clean up the code its a mess of different ideas
- [x] (optional?) Load VCGT ramps into KMS
- [ ] wiki for icc and new variables
- [x] Screencopy must not sample CM'd buffer


vaxerski:
@fufexan can we get nix

vaxerski:
can you guys let me know if this works well? I have done very limited testing.

Flat:
> can you guys let me know if this works well? I have done very limited testing.

```
Hyprland 0.52.0 built from branch icc-support at commit a3520d917afdbbd084c76a8daa8ce1d6c02811f5 clean (cm: add ICC profile loading).
Date: Tue Dec 23 11:06:26 2025
Tag: v0.52.0-183-ga3520d91, commits: 6742

Libraries:
Hyprgraphics: built against 0.4.0, system has 0.4.0
Hyprutils: built against 0.11.0, system has 0.11.0

Hyprlang: built against 0.6.7, system has 0.6.7
Aquamarine: built against 0.10.0, system has 0.10.0
```

Do not see anything in the logs showing it is loading the ICC profile, or rather nothing icc at all. What am I doing wrong there?

```
monitorv2 {
    output = DP-2
    mode = 3440x1440@143.97
    scale = 1
    position = 0x0
    bitdepth = 10
    cm = dcip3
    sdr_min_luminance = 0.0005
    sdr_max_luminance = 250
    icc = /home/flat/.config/icc/dell-icc-profile.icm
}
monitorv2 {
    output = DP-1
    mode = 3440x1440@143.97
    scale = 1
    position = auto-up
    bitdepth = 10
    cm = dcip3
    sdr_min_luminance = 0.005
    sdr_max_luminance = 250
    icc = /home/flat/.config/icc/dell-icc-profile.icm
}
```

[hyprland.log](https://github.com/user-attachments/files/24319089/hyprland.log)

vaxerski:
oh oops, I think I forgot v2

vaxerski:
my bad should work now

vaxerski:
it should dump what we were able to decode from the icc into the log like so if it loaded successfully

image

Flat:
Yup, it does attempt to load now. Unfortunately looks like the only ICC profile for my monitor is split-trc, so I won't be able to assist much with testing here.

`failed: Hyprland cannot represent split-trc profiles yet`

vaxerski:
I've added logging of the trc to the debug output, can you post the icc data dump here?

Flat:
Sure, here you go!
[iccdump.log](https://github.com/user-attachments/files/24332935/iccdump.log)

vaxerski:
thanks, despite xmas :P

vaxerski:
This should work. I've loaded my ICC profile and it does seem to change the colors on my panel a bit, but without a colorimeter I have no way of verifying if it's correct. Please load your ICC and let me know.

Flat:
It does seem to change the colors. I've ordered a colorimeter to verify though, should have it within a week.

vaxerski:
talk about dedication. I do think its wrong, sway's is different.

vaxerski:
@UjinT34 would you mind checking the LUT shader changes? I'm quite sure I am doing something wrong there. Shadows are a tad too bright I think.

Can't say for sure without a colorimeter but they do look a bit flat and weird.

vaxerski:
I checked KDE, colors shift a tad but not as much as we do. Something's wrong.

UjinT34:
I am not that familiar with math around icc luts. `clamp` looks sus to me, might be too early. EXT_LINEAR is expected to go outside of 0.0 - 1.0 and probably should be clamped after all the transformations.

vaxerski:
fixed, but it's again surfacing that nightmare where kitty sets the DEFAULTS and yet it changes how we render shit. I need to do something about this...

vaxerski:
nvm its not fixed it just had a bug where it wouldnt apply cm curves fuck my life

vaxerski:
in general @UjinT34 shouldn't we compose to a linear SRGB buffer internally and then do a final CM pass on the entire image? Why are we doing it per-surface?

This would be desirable as ICC shifts colors and will mess up e.g. color pickers. We should screencopy un-icc'd buffers.

vaxerski:
I've pushed a completely different approach with a 3D LUT instead, which also changes how we render CM. This will need some testing. 

This needs an HDR-kill-switch internally (HDR should disable ICC, currently it's likely the other way round)

This needs people to run this shit and report any visual glitches or bugs, feel free to drive this

@Flat thanks for your testing, if you could compare how this looks vs KDE it would be great. Sway's ICC is busted on my end, but KWin's works properly.

vaxerski:
I've decided to use Gamma 2.2 for internal storage just like kwin does, this lifts up the shadows a bit and looks more accurate in my opinion.

src/render/OpenGL.cpp
```diff
@@ -1732,7 +1732,7 @@ void CHyprOpenGLImpl::renderTextureInternal(SP<CTexture> tex, const CBox& box, c
         } else if (data.finalMonitorCM)
             passCMUniforms(*shader, NColorManagement::SImageDescription{});
         else
-            passCMUniforms(*shader, imageDescription, NColorManagement::SImageDescription{}); // CM into SRGB
+            passCMUniforms(*shader, imageDescription, NColorManagement::SImageDescription{.transferFunction = CM_TRANSFER_FUNCTION_GAMMA22}); // CM into SRGB with Gamma 2.2 (akin to KDE, cuz we love KDE)
```
UjinT34: #12693 does this by default and makes sense to use `render:cm_sdr_eotf` here as well

freevatar:
Yay! This is my use case! I have a colorimeter, good wide gamut display, its ICC profile, and the display itself has a pretty accurate sRGB clamp which I measured.

I hope to have some time to test this PR this weekend.

Meanwhile the only program I've found that applies ICC profile correctly (for sRGB clamp) is https://github.com/ledoge/novideo_srgb. But its for Windows. Every Linux DE implementation I tried was way off color-wise, at least the last time I checked.

vaxerski:
that's great! Let me know then how this performs. Also check KDE if you can, because if we are off and KDE is good then at least we have a reference.

Flat:
Not sure if the profile is being loaded, as it doesn't seem to be logging
```
DEBUG ]: ============= Begin ICC load =============
DEBUG ]: ============= End ICC load =============
```

Flat:
Latest commits also seem to break a lot of rendering, including fun stuff like this 

https://github.com/user-attachments/assets/669ec8e8-a0b0-428d-9222-7de5111bbcf1

vaxerski:
yes I realize its fucked but colors should be ok XD

vaxerski:
> Not sure if the profile is being loaded, as it doesn't seem to be logging

If your profile is very boring it might not log anything interesting

Flat:
> yes I realize its fucked but colors should be ok XD

I'll compare it with kwin then, have to install it lol

Colorimeter should be here tomorrow as well.

Flat:
It's hard to compare KWin to Hyprland with the eye alone, due to having to switch sessions without side by side. I set my camera to 200ISO and 1/125 exposure RAW to try to compare. Looks like Hyprland's correction comes out with a little darker highlights than KWin. I can't be sure it isn't due to distance from monitor on camera, or some other variable though. Will be able to provide some more concrete feedback once I have that colorimeter.

```
DEBUG ]: ============= Begin ICC load =============
DEBUG ]: ICC size: 6864 bytes
DEBUG ]: Building a 33³ 3D LUT
DEBUG ]: 3D LUT constructed, size 107811
DEBUG ]: ICC profile has no VCGT data
DEBUG ]: ============= End ICC load =============
```

vaxerski:
"fucking render glitches XDD" should be fixed

vaxerski:
> It's hard to compare KWin to Hyprland with the eye alone, due to having to switch sessions without side by side. I set my camera to 200ISO and 1/125 exposure RAW to try to compare. Looks like Hyprland's correction comes out with a little darker highlights than KWin. I can't be sure it isn't due to distance from monitor on camera, or some other variable though. Will be able to provide some more concrete feedback once I have that colorimeter.

No problemo, thanks for checking. The fact they're close is good news alone, at least I didnt fuck something up royally

Flat:
After reading far too long on how to use ArgyllCMS and getting this damn colorimeter working, I've created specific profiles for my monitors and compared to kwin. Hyprland produces way darker colors when applying the same profile.

Here are the results of colverify (which from my understanding compares the readings of the colorimeter compared to the generated target patches.) I've ran this on Hyprland and KWin after applying the ICC profile on each.
[DisplayA.tar.gz](https://github.com/user-attachments/files/24361742/DisplayA.tar.gz)
[colverify-hyprland.txt](https://github.com/user-attachments/files/24361721/colverify-hyprland.txt)
[colverify-kwin.txt](https://github.com/user-attachments/files/24361722/colverify-kwin.txt)

vaxerski:
define "way", according to the results we seem to be worse but not like terribly. 

I am not saying this is acceptable, just curious. On my panel it wasn't "way", more like "a bit"

vaxerski:
ah, we had gamma 2.2 by accident. Check now. It's because of the rebase. @Flat Sorry about that

Flat:
Looks like that does help, also just realized think I profiled these with cm set to wide, so probably need to redo those as well

vaxerski:
cm, x is overridden by icc, doesn't do anything. Almost all cm options get overridden

Flat:
Yes, I mean when I created the ICC profile, it was set to wide, which would have been changing the base color readings used to create the profile.

vaxerski:
ah, right. Idk much about profiling but that could be an issue

Flat:
Just recreated them, not much different honestly. I think the ICC support looks good to me, unless someone with more experience doing this in Linux comes along. Wayland is pretty much unsupported by about the only calibration tool available on Linux, for calibration anyway. Profiling seems to work ok. See https://github.com/eoyilmaz/displaycal-py3/issues/133 for further info on that.

njdom24:
Just tested this PR briefly, and found that HDR screenshare with `cm = hdr` no longer works with direct scanout enabled.
I tested both `mpv` playing an HDR video, and `gamescope` playing an HDR game, with the same results.

If I full-screen the app before opening OBS, the screenshare will fail. If I open OBS and share the screen, _then_ full-screen the app, the app will output as if my display doesn't support HDR (extremely gray, washed-out). It also has a ~25% chance of just crashing instead when I do that.

Works fine if the app is windowed, and works again the moment I close OBS.

vaxerski:
screenshare is a bit wonk atm yes

vaxerski:
@Flat do you mean that KDE and Hyprland are now not much different?

Flat:
> @Flat do you mean that KDE and Hyprland are now not much different?

Correct. I did not measure again, but they visually look about the same.

Flat:
Actual measurements (Normalized to White X + Y + Z)

Hyprland
```
Verify results:
  File 1 L* ref. X+Y+Z 0.951065 1.000000 1.088440
  File 2 L* ref. X+Y+Z 140.574247 147.742458 189.660116
  Total errors:     peak = 16.855547, avg = 8.286194
  Worst 10% errors: peak = 16.855547, avg = 12.674006
  Best  90% errors: peak = 11.591609, avg = 7.798659
  avg err X  1.002827, Y  1.049726, Z  1.025701
  avg err L* 3.163749, a* 3.252128, b* 6.158267
  ```
  
  KWin
  ```
Verify results:
  File 1 L* ref. X+Y+Z 0.951065 1.000000 1.088440
  File 2 L* ref. X+Y+Z 138.960762 146.111848 188.965996
  Total errors:     peak = 17.163952, avg = 8.387285
  Worst 10% errors: peak = 17.163952, avg = 13.251886
  Best  90% errors: peak = 11.753762, avg = 7.846774
  avg err X  0.832850, Y  0.853621, Z  1.294807
  avg err L* 2.607523, a* 3.445785, b* 6.302393
  ```

**Comment by Flat** [comment]
Took the time to profile in X11, which is known to work, and get some better human readable verification reports.

X11 comes out almost spot on.
KWin is... not correct.
Hyprland seems to be even more off than KWin.

[X11 Monitor Measurement Report.html](https://github.com/user-attachments/files/24376103/X11.Monitor.Measurement.Report.html)
[Hyprland Monitor Measurement Report.html](https://github.com/user-attachments/files/24376105/Hyprland.Monitor.Measurement.Report.html)
[KWin Monitor Measurement Report.html](https://github.com/user-attachments/files/24376106/KWin.Monitor.Measurement.Report.html)

File: src/helpers/cm/ICC.cpp [hunk 2]
```diff
@@ -0,0 +1,384 @@
...
+    if (table.channels != 3 || table.entrySize != 2 || table.entries == 0)
+        return std::unexpected("invalid vcgt table size");
+
+    size_t tableBytes = (size_t)table.channels * (size_t)table.entries * (size_t)table.entrySize;
+
+    // VCGT is a piece of shit and some absolute fucking mongoloid idiots
+    // decided it'd be great to have both 18 and 20
+    // FUCK YOU
+    size_t tableOff = 20;
+    if (raw.size() < tableOff + tableBytes)
```
Flat: This may not be correct, this ICC profile with vcgt is not loaded because of this check. I flipped the comparison to load it and it seems to load the rest of the table at offset 18

[DisplayA-X11.tar.gz](https://github.com/user-attachments/files/24385510/DisplayA-X11.tar.gz)

iccdump
```
tag 7:
  sig      'vcgt'
  type     'vcgt'
  offset   976
  size     1554
VideoCardGammaTable:
  channels  = 3
  entries   = 256
  entrysize = 2
```

```
DEBUG ]: ============= Begin ICC load =============
DEBUG ]: ICC size: 969536 bytes
DEBUG ]: Building a 33³ 3D LUT
DEBUG ]: 3D LUT constructed, size 107811
DEBUG ]: readVCGT16: table has 3 channels, 256 entries, and entry size of 2
DEBUG ]: readVCGT16: table is likely offset 18 not 20, re-reading
DEBUG ]: readVCGT16: red channel: [0, 257, ... 65278, 65535]
DEBUG ]: readVCGT16: green channel: [0, 257, ... 65278, 65535]
DEBUG ]: readVCGT16: blue channel: [0, 257, ... 65278, 65535]
DEBUG ]: ============= End ICC load =============
```

Flat:
Just verified on Gnome Wayland, and it looks like their color management is correct, if that may help to reference as well.
[Gnome.html](https://github.com/user-attachments/files/24386121/Gnome.html)

src/helpers/cm/ICC.cpp
vaxerski: can you drop the icc profile so I can inspect it?

vaxerski:
I can feel an imbalance in the force... I feel this will be a painful battle

src/helpers/cm/ICC.cpp
Flat: Linked in the above comment, GitHub won't let me upload .icc so it's tar-ed

vaxerski: thanks, I thought it was the log

vaxerski: latest commit should allow your vcgt ramps to load. Can you check color accuracy now?

Flat: If they loaded correctly when I changed that check, the readings are about the same, but if you expect it to be different, I can measure again in a few minutes

Flat: [Hyprland2.html](https://github.com/user-attachments/files/24388409/Hyprland2.html)
```
DEBUG ]: ============= Begin ICC load =============
DEBUG ]: ICC size: 968640 bytes
DEBUG ]: Building a 33³ 3D LUT
DEBUG ]: 3D LUT constructed, size 107811
DEBUG ]: readVCGT16: table has 3 channels, 256 entries, and entry size of 2
DEBUG ]: readVCGT16: table is too short, but off = 18 fits. Attempting offset = 18
DEBUG ]: readVCGT16: red channel: [0, 257, ... 65278, 65535]
DEBUG ]: readVCGT16: green channel: [0, 257, ... 65278, 65535]
DEBUG ]: readVCGT16: blue channel: [0, 257, ... 65278, 65535]
DEBUG ]: ============= End ICC load =============
```

UjinT34:
Try creating the profile with `cm_enabled=0` so it doesn't affect the result

vaxerski:
he created the profile on X11 iirc

Flat:
Correct, profile was created in X11, using DisplayCal/ArgyllCMS which disables any loaded profiles and resets video card gamma table before creating the profile.

fxzzi:
i don't have a calibration tool or anything but I do have icc profiles for my monitors, one of them loads fine, the other doesn't 
[gigabyte_m27q.icm.zip](https://github.com/user-attachments/files/24401319/gigabyte_m27q.icm.zip)

i set these whilst hyprland was open, but hyprland coredumps if i launch
[log.txt](https://github.com/user-attachments/files/24401344/log.txt)

btw just from my eyes things look too dark but i could be wrong and stupid and wrong and stupid (did i mention that last one already?)

njdom24:
> btw just from my eyes things look too dark but i could be wrong and stupid and wrong and stupid (did i mention that last one already?)

This seems to be the case with Gamma 2.2 vs. other compositors (KDE, Sway) that use the same `gamma22` EOTF for SDR content. I was able to measure with my phone's ambient light sensor and see that for all 3, when set to 203 nits, they all hit the same value on a white screen. Just seems like Hyprland has dimmer midtones for some reason.

vaxerski:
@fxzzi can you get a debug stacktrace

fxzzi:
errr is this what you're looking for

[crash.txt](https://github.com/user-attachments/files/24401691/crash.txt)

also this icc profile crashes too:
[MO27Q28G.icm.zip](https://github.com/user-attachments/files/24401694/MO27Q28G.icm.zip)
i pulled it from the hdr thing given from my monitor support page, crash for this one is different
[hyprlandCrashReport18203.txt](https://github.com/user-attachments/files/24401695/hyprlandCrashReport18203.txt)

UjinT34:
> errr is this what you're looking for
> 
> [crash.txt](https://github.com/user-attachments/files/24401691/crash.txt)
> 
> also this icc profile crashes too: [MO27Q28G.icm.zip](https://github.com/user-attachments/files/24401694/MO27Q28G.icm.zip) i pulled it from the hdr thing given from my monitor support page, crash for this one is different [hyprlandCrashReport18203.txt](https://github.com/user-attachments/files/24401695/hyprlandCrashReport18203.txt)

Looks like it needs a rebase to include https://github.com/hyprwm/Hyprland/commit/5faa66d297752ab0d919bb5719fa0949292fe720

fxzzi:
probably but icc profile still fails to load either way

vaxerski:
@fxzzi idk why lcms fails no clue icc profile bork

fxzzi:
> Just seems like Hyprland has dimmer midtones for some reason.

i'm not too sure i thikn hyprland might be parsing something in specific icc's wrong, because on my secondary monitor with its ICC profile everything looks pretty great, but on my main monitor which is an oled dark colours are definitely crushed

again idk how this stuff works so grain of salt

vaxerski:
to be honest we're on par with KDE but if anyone knows what gnum does differently to get the accuracy actually good, lmk, otherwise idk much about how to push this forward

Flat:
So... after following the steps here https://planet.kde.org/xavers-blog-2024-07-15-how-to-profile-your-display-in-the-plasma-wayland-session/ I've measured the profile created by that ([Display-DP-2-Wayland.tar.gz](https://github.com/user-attachments/files/24424207/Display-DP-2-Wayland.tar.gz)) on Gnome, KWin, and Hyprland. Doing the profile this way, Gnome is way off, and KDE is the most correct. Hyprland is close, but there is something wrong with the transform on the blue channel, specifically test patch 29. 
[Gnome.html](https://github.com/user-attachments/files/24424217/Gnome.html)
[KWin.html](https://github.com/user-attachments/files/24424218/KWin.html)
[Hyprland.html](https://github.com/user-attachments/files/24424219/Hyprland.html)

I think if that can be tracked down it would be correct, at least against KDE. I think it would take a bit more in terms of Wayland color management protocols to match X11.

src/helpers/cm/ColorManagement.cpp
```diff
@@ -0,0 +1,193 @@
...
+    auto   invP = invertMat3(P);
+    if (!invP)
+        return std::nullopt;
+
+    const auto S = matByVec(*invP, {W.x, W.y, W.z});
+
+    P.multiply(diag3(S)); // RGB->XYZ
+
+    return P;
+}
```
UjinT34, line 175: Hyprgraphics::CMatrix3::invert, Hyprgraphics::CMatrix3::operator*, Hyprgraphics::SPCPRimaries::toXYZ should do the same thing with `double` values.

src/helpers/cm/ColorManagement.cpp
```diff
@@ -0,0 +1,193 @@
...
+    const auto                 srcLMS = matByVec(Bradford, {srcXYZ.x, srcXYZ.y, srcXYZ.z});
+    const auto                 dstLMS = matByVec(Bradford, {dstXYZ.x, dstXYZ.y, dstXYZ.z});
+
+    const std::array<float, 3> scale{dstLMS[0] / srcLMS[0], dstLMS[1] / srcLMS[1], dstLMS[2] / srcLMS[2]};
+
+    Mat3x3                     result = BradfordInv;
+    result.multiply(diag3(scale)).multiply(Bradford);
+
+    return result;
+}
```
UjinT34, line 193: Hyprgraphics::CMatrix3::adaptWhite

UjinT34:
Blue x,y values are closest to 0. Could it be caused by lack of `float` precision?

visixx:
Not really related to the ICC changes, but using this PR with `cm = hdr` SDR now looks _a lot_ more correct. However, it seems that windowed HDR content does not display properly and is washed-out. Even after fullscreening HDR is still washed-out for Chromium and MPV. I guess that's because CM is being done at the very end on the entire monitor to go from Gamma 2.2 -> sRGB (when `finalMonitorCM` is `true`)?

I know running `cm = hdr` is not recommended, but with these changes SDR content looks really close to my monitor's `sRGB` mode so `cm = hdr` is _almost_ good enough to keep on at all times. It's just this HDR issue that's holding it back, at least in my case.

s0mebodyhelpme:
This actually fixes window screen capture for me on `cm = hdr`.However it also makes proton games with the same setting and hdr mode (for the game) on have incorrect brightness. Tested with FFXV.

MajedAlghoul:
i havent tested the pr, but just to add to this, i always use my device with an icc profile made for my display, and from my experience testing it on windows, kde and gnome, gnome looks like windows (which i assume is the correct one), kde one colors look extra saturated, if u are not used to a profile your eyes will always lie to you, i would recommend having a reference on a properly calibrated display

s0mebodyhelpme:
Btw this seems to break plugin compatibility for me. Plugins aren't loaded and setting plugin window rules causes errors in the config.

vaxerski:
@MajedAlghoul that would align with the findings before. Gnome seems to be doing something differently than KDE (& us) but I am not sure what

Flat:
I believe the previous gnome results (where they appeared more correct), can be disregarded as that profile had calibration which according to the KDE developer who helped implement theirs, messes up the profiles on Wayland due to unavailable functions that are used on Xorg when creating/verifying those profiles.

Also iirc, windows does not use ICC profiles for whole display calibration, and just in color managed applications unless you use a separate program to load them for the whole display.

MajedAlghoul:
@Flat the difference between the kde and gnome one is massive, and i think gnome is correct because i also compared colors for hours to an iphone (which has its colors calibrated from factory), and the gnome colors are way closer, and regarding windows handling icc profiles, yes true i was using the  icc profile with the asus splendid utility, i extracted the icc profile from its files

MajedAlghoul:
whatever gnome is doing, its affecting the screen temperature, it looks compared to kde as if warm layer is covering the screen, idk how that translates to the calculation.

Flat:
> @Flat the difference between the kde and gnome one is massive, and i think gnome is correct because i also compared colors for hours to an iphone (which has its colors calibrated from factory), and the gnome colors are way closer, and regarding windows handling icc profiles, yes true i was using the  icc profile with the asus splendid utility, i extracted the icc profile from its files

Yes, the difference is pretty large, however these are the results of measurement with a colorimeter, which I would say is a lot more accurate than trying to comparw with your eyes alone.

MajedAlghoul:
> > @Flat the difference between the kde and gnome one is massive, and i think gnome is correct because i also compared colors for hours to an iphone (which has its colors calibrated from factory), and the gnome colors are way closer, and regarding windows handling icc profiles, yes true i was using the  icc profile with the asus splendid utility, i extracted the icc profile from its files
> 
> Yes, the difference is pretty large, however these are the results of measurement with a colorimeter, which I would say is a lot more accurate than trying to comparw with your eyes alone.

true, idk i got nothing to add really, i just wouldnt be satisfied with the kde colors on my display, if u got windows can u do the test there?

Flat:
(Un?)fortunately I do not have a windows install, and it will not really work in a VM.

MajedAlghoul:
> (Un?)fortunately I do not have a windows install, and it will not really work in a VM.

get ur drive dirty for the team haha

bljdd:
how to enable icc support, i don't understand. is there something to download or just config? please help, really want to try

njooobe:
@vaxerski  I have a Lenovo Legion 5i with a calibrated from factory OLED screen. I dumped the icm files from windows and tried to load them using the new icc option. I can definitely say that the output seems to be too washed out compared to windows. The output looks better without loading any icc profile. I don't know where the error lies. I didn't try with gnome but I could if it helps.
I have different profiles available: AdobeRGB, DCIP3, DisplayP3, REC709, sRGB, Native and Default. The default option in windows is the "Default" profile, the one I tried earlier.

Edit: gnome works as expected

UjinT34:
There are a bunch of drm properties that are meant to pass current CM settings to the monitor so it can decide how to decode the data. Those properties aren't reliable and doesn't seem to work as expected with every hw setup. If your laptop actually handles those values then washed out colors can be caused by double CM. When you set an ICC profile HL doesn't set any of those properties, your laptop screen might treat those as an sRGB colorspace and apply color transformations to avoid oversaturation (wash out colors). If the ICC profile itself was created for a screen that's is set to use a larger colorspace then the profile also wash outs colors to avoid oversaturation.
If this is the case then you have 3 options to fix it:

1. create an ICC profile yourself (requires colorimetry equipment, provides the best result)
2. change screen settings to treat input in a correct colorspace (probably not available for a laptop screen)
3. change HL/AQ settings to send some properties with the correct values

Option 3 is the trickiest one because you'll need to figure out which properties and values should be send, add support for those to AQ and add some settings/conditions to HL. Currently only Default (sRGB) and BT2020 colorspaces are supported and ICC implementation doesn't seem to change the relevant property.

TL;DR sRGB or Native profiles might be correct with the current state of things.

vaxerski:
this still doesn't solve our problem though, as gnome seems correct. We need to find out why.

njooobe:
> There are a bunch of drm properties that are meant to pass current CM settings to the monitor so it can decide how to decode the data. Those properties aren't reliable and doesn't seem to work as expected with every hw setup. If your laptop actually handles those values then washed out colors can be caused by double CM. When you set an ICC profile HL doesn't set any of those properties, your laptop screen might treat those as an sRGB colorspace and apply color transformations to avoid oversaturation (wash out colors). If the ICC profile itself was created for a screen that's is set to use a larger colorspace then the profile also wash outs colors to avoid oversaturation. If this is the case then you have 3 options to fix it:
> 
>     1. create an ICC profile yourself (requires colorimetry equipment, provides the best result)
> 
>     2. change screen settings to treat input in a correct colorspace (probably not available for a laptop screen)
> 
>     3. change HL/AQ settings to send some properties with the correct values
> 
> 
> Option 3 is the trickiest one because you'll need to figure out which properties and values should be send, add support for those to AQ and add some settings/conditions to HL. Currently only Default (sRGB) and BT2020 colorspaces are supported and ICC implementation doesn't seem to change the relevant property.
> 
> TL;DR sRGB or Native profiles might be correct with the current state of things.

I tried sRGB profile and it seems to look okay. I don't know how it is compared to gnome/windows, but it looks good. Maybe a bit too dark(?), definitely darker than without any profile. But the colors are okay. So maybe it's a colorspace problem?

> this still doesn't solve our problem though, as gnome seems correct. We need to find out why.

Yes, in gnome there was no problem at all.

PS: I'm available to do some tests if needed

Flat:
Just want to bring this data back up since, I think it's important to answer the first step of: Is gnome correct, or is it functioning similar to windows, or not at all?
https://github.com/hyprwm/Hyprland/pull/12711#issuecomment-3708449976

Windows by default, (i.e using the built in ICC profile loading) does not apply the ICC profile display wide, and only provides the profile to color managed applications to use on their own to display those color modifications, (Photoshop, Davinci Resolve, etc)

So, we need to know if gnome is actually applying the profile and correctly applying all of the information, or if it is not applying it at all, and the default calibration looks good already.

As when I profiled gnome when using the icc profile I generated on Kwin, it was way off.

njooobe:
> Just want to bring this data back up since, I think it's important to answer the first step of: Is gnome correct, or is it functioning similar to windows, or not at all? [#12711 (comment)](https://github.com/hyprwm/Hyprland/pull/12711#issuecomment-3708449976)
> 
> Windows by default, (i.e using the built in ICC profile loading) does not apply the ICC profile display wide, and only provides the profile to color managed applications to use on their own to display those color modifications, (Photoshop, Davinci Resolve, etc)
> 
> So, we need to know if gnome is actually applying the profile and correctly applying all of the information, or if it is not applying it at all, and the default calibration looks good already.
> 
> As when I profiled gnome when using the icc profile I generated on Kwin, it was way off.

I just booted up windows to try the ICC profile loading. On my installation windows applies the ICC profile display wide. I can immediately see a change in colors whenever I change the profile.

UjinT34:
I am 99% sure it's the colorspace issue. Screen does it's own CM with incorrect assumptions about the input signal. If sRGB profile is ok then you lose any extra color volume that can be displayed on your screen. The best match should be the Native profile without any extra CM done by the screen itself (no idea how to turn it off).

Gnome probably sets the correct drm properties or does some extra CM. It's hard to code those because different hw supports different sets of properties with different sets of values. Unsupported properties and values aren't exposed and there is no way to test them all without relevant videocard&monitor combinations. E.g. https://drmdb.emersion.fr/properties/3233857728/Colorspace lists a bunch of values but I get `"Colorspace": enum {Default, BT2020_RGB, BT2020_YCC}`

vaxerski:
and what about a profile generated on hyprland, tested on hyprland?

UjinT34:
> and what about a profile generated on hyprland, tested on hyprland?

Profiles generated with `cm = wide` and `cm = edid` probably won't work because they set `BT2020_RGB` Colorspace and icc and other cm settings don't.
Profiles generated with other cm values should work as expected but will be bound by the selected cm colorspace.
Ideally the profile should be generated with disabled CM and then there should be a way to disable screen's internal CM (most screens call it Native colorspace)

burnskp:
Does this handle the calibration and profiling settings in an icc file or only profiling? I know sway doesn't handle vgct, so the monitors temperature isn't changed. Argyllcms/display cal also don't support dealing with the temperature in Wayland, which means you can only use the tool to profile the display not calibrate and profile. I believe gnome does handle the vgct and their calibration app works properly in Wayland.

njooobe:
Good news. In my case I think it's a colorspace issue. I know almost nothing on Color Management, but I found that the icc file I was trying to load was calibrated using DCIP3 and not sRGB. I tried adding a function to the code with the primaries of DCIP3 instead of sRGB and the profile is now correctly displayed. I don't know if it's "precise", but it isn't washed out or unusable. It's only slightly different from the default panel config, so I'm almost certain that the issue is the colorspace.
This is the function I used:
```
static UniqueProfile createLinearP3D65Profile() {
    cmsCIExyYTRIPLE prim{};
    // Display P3 (D65) primaries
    prim.Red.x   = 0.6800;
    prim.Red.y   = 0.3200;
    prim.Red.Y   = 1.0;
    prim.Green.x = 0.2650;
    prim.Green.y = 0.6900;
    prim.Green.Y = 1.0;
    prim.Blue.x  = 0.1500;
    prim.Blue.y  = 0.0600;
    prim.Blue.Y  = 1.0;

    cmsCIExyY wp{};
    wp.x = 0.3127;
    wp.y = 0.3290;
    wp.Y = 1.0; // D65

    cmsToneCurve* lin       = cmsBuildGamma(nullptr, 1.0);
    cmsToneCurve* curves[3] = {lin, lin, lin};

    cmsHPROFILE p = cmsCreateRGBProfile(&wp, &prim, curves);

    cmsFreeToneCurve(lin);
    return UniqueProfile{p};
}
```
And I replaced the call in `buildIcc3DLut` to `UniqueProfile src = createLinearP3D65Profile();`

Hope it helps.

Edit: after trying various presets, I found out that the color calibration of the sRGB preset using createLinearSRGBProfile is equivalent to the DisplayP3 preset using createLinearP3D65Profile (good sign). So the sRGB one is working well. The "Default" and "Native" icc files are of another colorspace that I can't identify (it's his own colorspace..?)
I'm sorry if I'm wasting your time, I don't have experience with this sort of things.

Flat:
> and what about a profile generated on hyprland, tested on hyprland?

Unfortunately, since Arch upgraded to python 3.14, DisplayCal is pretty broken, and I do not have the time to try to fix that currently. When it gets fixed upstream I will try profiling with cm_enabled = false, and testing on Hyprland.



>... I believe gnome does handle the vgct and their calibration app works properly in Wayland.

Hyprland should apply the vgct if there is one in the .icc file. Do you have a link to their calibration app that works on Wayland? I could not find anything when I was looking into it.

MajedAlghoul:
> Just want to bring this data back up since, I think it's important to answer the first step of: Is gnome correct, or is it functioning similar to windows, or not at all? [#12711 (comment)](https://github.com/hyprwm/Hyprland/pull/12711#issuecomment-3708449976)
> 
> Windows by default, (i.e using the built in ICC profile loading) does not apply the ICC profile display wide, and only provides the profile to color managed applications to use on their own to display those color modifications, (Photoshop, Davinci Resolve, etc)
> 
> So, we need to know if gnome is actually applying the profile and correctly applying all of the information, or if it is not applying it at all, and the default calibration looks good already.
> 
> As when I profiled gnome when using the icc profile I generated on Kwin, it was way off.

im 100% sure gnome applies the profile, actually for the past year ive been using gdm to get around hyprland's lack of icc support, i installed gdm and a minimal gnome installation and applied the profile in their settings, i immediately see the difference and actually the color still applies even after logging out and running hyprland

vaxerski:
does any environment other than gnome do icc "correctly"?

MajedAlghoul:
> does any environment other than gnome do icc "correctly"?

~~i tested weston wm, and it matches gnome's colors pretty well~~
after looking into more colors, weston is saturated like kde
i noticed that my profile extension is icm, not icc, but as far as i know this doesn't matter

vaxerski:
it doesn't, .icm = .icc

is hyprland the same as kde?

MajedAlghoul:
ill test the patch and see

MajedAlghoul:
heres what i found,
i tested gnome, kde, weston and hyprland with the patch,
first of all, my display naturally is overly cold and the profile helps making it warmer adding a yellowish coat,

gnome was the only one to produce accurate results for me, i was referencing two blocks of solid colors,
red (#ff0000) and blue (#0000ff), gnome was the only one able to display blue, others where showing shades of purple
also blacks are balanced

on kde, colors where super saturated neon like, all color options had no affect on the result except one called "srgb intensity",
turning that to 100% helped a lot with the saturation, but not enough, blue shows as purple but now a bit bluish,
blacks are a bit yellowish

weston was very close to kde (with srgb intensity to 100%), not identical but close, a bit less saturated maybe but still shows
blues as bluish purple,

hyprland has a similar look to kde (with srgb intensity to 100%), but it differs in that blacks have a noticeable reddish tint,

i know the references im using are dependent to my use case and screen but i mentioned them to establish a general rule in reference to the desired results

let me know if u need me to test something else

UjinT34:
`drm_info` from every compositor with icc applied might show the cause of the difference. Probably `Colorspace` or some other color properties get different values.

Malinskiy:
Instead of subjective measurements for color accuracy after grading, ideally you'd want to verify after loading the ICC profile via hyprland that the profile's correction achieves the desired color reproduction via something like https://displaycal.net/#verify-profile. I tried, but unfortunately I couldn't get the displaycal to work under hyprland since displaycal is not well maintained and I don't know of any good alternative for Linux. I'd be happy to test that the icc loading works provided someone knows any profiling software that works in hyprland, I don't want to resort to manual verification via predefined color patches

MajedAlghoul:
I did the drm comparison, i summarized the differences in a pdf attached below,
i applied the same color profile to each wm,
in the pdf u find the path of the property that changes between at least 2 wms with their values,

i think this will be very helpful
[DRM difference, gnome x plasma0srgb x plasma100srgb x weston x hyprland icc patch.pdf](https://github.com/user-attachments/files/24839768/DRM.difference.gnome.x.plasma0srgb.x.plasma100srgb.x.weston.x.hyprland.icc.patch.pdf)

wilecoyote2015:
@Malinskiy Displaycal Py3 works on hyprland for me.
https://github.com/eoyilmaz/displaycal-py3
But you have to use python <= 3.13
Adjusting the color patches size and position does not work, but otherwise it is fine.

I'll also try to generate and verify an ICC profile for my OLED monitor that covers approx Display DCI-P3.
But I only have an old Spyder 3, so that I'm not sure about accuracy, unfortunately.

Nevertheless, ICC support is a very much desired feature for me. Without, not even somewhat accurate color reproduction is possible on modern wide-gamut displays...

wilecoyote2015:
I've now profiled my monitor with Displaycal Py3, used the result as icc profile in Hyprland and ran the verification against  	sRGB IEC61966-2.1.
Results look quite okay so far in general:

image

What's apparent: the darker patches still show some deviation. Is this due to the eotf? 
If you look at the patches in the table, you can clearly see patches 4 and 5 deviate quite strongly (see the left/right split indicating expected and measured color). That's the only critical aspect that would not work for me yet, as that tone range is quite critical in photo editing.
But the rest matches so well that I'd consider it completely usable in practice. Thanks!

image


[Measurement Report 3.9.17 - Monitor 1, Output eDP-1 @ 0, 0, 2560x1600 - 2026-03-01 01-01.html](https://github.com/user-attachments/files/25622387/Measurement.Report.3.9.17.-.Monitor.1.Output.eDP-1.%40.0.0.2560x1600.-.2026-03-01.01-01.html)

What may be important: Those are my settings in Displaycal for verification
image
The documentation of Displaycal is not entirely clear to me, but as far as I understand, this is the correct constellation to check whether the current display rendering yields the desired sRGB response of the simulation profile. 
As we are on Wayland and Displaycal does not know about the CM in Hyprland, we just render the colors by Hyprland's CM settings and DisplayCAL measures them as is and compares them to the simulation profile.

vaxerski:
@wilecoyote2015 thanks! I'll try to rebase this later today or tomorrow and hopefully get it merged then.

vaxerski:
Alright, rebased. Please re-run the test, I can't verify if I didn't accidentally fuck up colors somewhere. There were many internal renderer changes so I might've missed something.

wilecoyote2015:
Thanks, looks good.
However, I had to set  `cm_sdr_eotf = srgb`. Otherwise, dark tones are crushed too much and have delta E > 4.
I'm not sure whether this is due to problems with Displaycal or rendering.
I also cannot say if this is significant or within measurement error with the Spyder 3 on OLED, unfortunately...

image

[Measurement Report 3.9.17 - Web @ localhost - 2026-03-03 04-05.html](https://github.com/user-attachments/files/25692726/Measurement.Report.3.9.17.-.Web.%40.localhost.-.2026-03-03.04-05.html)

vaxerski:
@fufexan wtf is this nix conflict

vaxerski:
@wilecoyote2015 honestly dunno, @UjinT34 advocated for gamma2.2 as apparently kwin et al use that (and yea without icc it looks more correct) but perhaps ICC should fall back to srgb.

vaxerski:
this commit should force sRGB when ICC is on

vaxerski:
also out of curiosity @wilecoyote2015 I didnt ask but I assume this _is_ better than without icc? xD

Idk if you've tested gnome/kwin to compare.

fufexan:
> @fufexan wtf is this nix conflict

I changed the formatting to the standard on main. Just rebase and redo the nix file or whatever, it's two lines.

wilecoyote2015:
> also out of curiosity @wilecoyote2015 I didnt ask but I assume this is better than without icc? xD

Yes, much better than without! Without ICC, the color is way off. In the case of my display, it covers the DCIP3 gamut, but does not match the DCIP3 response well at all. So even when using the dcip3 cm preset, colors are off and delta E is in the invalid region everywhere.

> @wilecoyote2015 honestly dunno, @UjinT34 advocated for gamma2.2 as apparently kwin et al use that (and yea without icc it looks more correct) but perhaps ICC should fall back to srgb.

Honestly, dunno, too. Although that color management stuff is based on straight forward math in principle, I still struggle to wrap my head about the specific details how the pieces come together as correct color management has so many entangled layers of physical and perceptual aspects xD

Here is a short chat I had with claude Opus 4.6 about this topic - cannot verify the facts in depth. But the bottom line at least corresponds to my intuition to some degree: 
- As the eotf setting in Hyprland is only an assumption for clients not communicating their color space, it does not harm the rendering of graphic applications that do proper color management. Hence the discrepancy in the eotf between gamma 2.2. and sRGB is not a deal breaker for color managed work (under the assumption that color managed graphic applications will use the protocol to communicate their color space properly in the future). 
- So the current behavior is okay for most users I guess, and forcing sRGB when ICC is on should help to get correct color handling for non-color-managed applications by default. Maybe only problem here is that forcing sRGB for iCC makes configuration less understandable for the user?
- it is maybe a conceptional question whether the non-sRGB eotf should really be used to decode client colors, or whether color decoding should always happen in a correct way with sRGB, so that Hyprland always has correct internal color representation from the clients and there should be an option to darken the shadows when doing the output transform for emulating a gamma 2.2 display or the corresponding rendering of windows / kde or whatever. This would eliminate the sdr eotf setting I guess?
[chat_claude_opus_4_6_eotf.md](https://github.com/user-attachments/files/25723736/chat_claude_opus_4_6_eotf.md)

vaxerski:
while working on screencopy I realized the work buffer was actually not srgb but I can't seem to get pc to work again

vaxerski:
fixed, screencopy should work now too. 

@wilecoyote2015 sorry for being so annoying but could you recheck the colorimeter, plus check if CM doesn't disturb screencopy? (screenshots / screen recordings) - these should be in SRGB, IOW if I open a terminal on srgb no icc and use a color picker e.g. hyprpicker to pick the background, the same #hex should be printed with icc.

I ran some tests and on my end looks good

vaxerski:
if colorimerer passes this is gtg

Flat:
Working on updating ArgyllCMS and Displaycal to give another test as well. Hopefully have that test done in next couple hours.

Flat:
Still about the same for me, using the same verification options as wileycoyote2015.
[Hyprland No ICC.html](https://github.com/user-attachments/files/25748547/Hyprland.No.ICC.html)
[Hyprland ICC Applied.html](https://github.com/user-attachments/files/25748549/Hyprland.ICC.Applied.html)
[Plasma ICC Applied.html](https://github.com/user-attachments/files/25748551/Plasma.ICC.Applied.html)

No ICC:
image


ICC:
image

KDE:
image

Same profile profiled on Hyprland used on Hyprland and KDE tests.

Profiling Settings:
image

Resultant ICC Profile:
[Display-DP-2-Hyprland.icc.tar.gz](https://github.com/user-attachments/files/25748790/Display-DP-2-Hyprland.icc.tar.gz)

vaxerski:
wtf happened in indigo there

Flat:
Million dollar question 😀.

Ujin suggested it could be a float precision issue since blues were closest to 0, https://github.com/hyprwm/Hyprland/pull/12711#issuecomment-3709309869

Or a difference in colorspaces. Here's a diff between drm_info on hyprland and kde
```diff
85c85
< │   │       ├───"EDID" (immutable): blob = 147
---
> │   │       ├───"EDID" (immutable): blob = 154
136c136
< │   │       ├───"EDID" (immutable): blob = 152
---
> │   │       ├───"EDID" (immutable): blob = 147
161c161
< │           ├───"EDID" (immutable): blob = 148
---
> │           ├───"EDID" (immutable): blob = 152
217c217
< │   │       ├───"MODE_ID" (atomic): blob = 160
---
> │   │       ├───"MODE_ID" (atomic): blob = 163
237c237
< │   │       ├───"MODE_ID" (atomic): blob = 167
---
> │   │       ├───"MODE_ID" (atomic): blob = 164
291,292c291,292
<     │   │   ├───FB ID: 184
<     │   │   │   ├───Object ID: 184
---
>     │   │   ├───FB ID: 160
>     │   │   │   ├───Object ID: 160
294c294
<     │   │   │   ├───Format: XBGR2101010 (0x30334258)
---
>     │   │   │   ├───Format: ABGR16161616F (0x48344241)
297c297
<     │   │   │       └───Plane 0: offset = 0, pitch = 13760 bytes
---
>     │   │   │       └───Plane 0: offset = 0, pitch = 27520 bytes
324,325c324,325
<     │       ├───"FB_ID" (atomic): object framebuffer = 184
<     │       │   ├───Object ID: 184
---
>     │       ├───"FB_ID" (atomic): object framebuffer = 160
>     │       │   ├───Object ID: 160
327c327
<     │       │   ├───Format: XBGR2101010 (0x30334258)
---
>     │       │   ├───Format: ABGR16161616F (0x48344241)
330c330
<     │       │       └───Plane 0: offset = 0, pitch = 13760 bytes
---
>     │       │       └───Plane 0: offset = 0, pitch = 27520 bytes
537,540c537,540
<     │       ├───"CRTC_X" (atomic): srange [INT32_MIN, INT32_MAX] = 0
<     │       ├───"CRTC_Y" (atomic): srange [INT32_MIN, INT32_MAX] = 0
<     │       ├───"CRTC_W" (atomic): range [0, INT32_MAX] = 0
<     │       ├───"CRTC_H" (atomic): range [0, INT32_MAX] = 0
---
>     │       ├───"CRTC_X" (atomic): srange [INT32_MIN, INT32_MAX] = 467
>     │       ├───"CRTC_Y" (atomic): srange [INT32_MIN, INT32_MAX] = 1423
>     │       ├───"CRTC_W" (atomic): range [0, INT32_MAX] = 256
>     │       ├───"CRTC_H" (atomic): range [0, INT32_MAX] = 256
543,544c543,544
<     │       ├───"SRC_W" (atomic): range [0, UINT32_MAX] = 0
<     │       ├───"SRC_H" (atomic): range [0, UINT32_MAX] = 0
---
>     │       ├───"SRC_W" (atomic): range [0, UINT32_MAX] = 256
>     │       ├───"SRC_H" (atomic): range [0, UINT32_MAX] = 256
778,779c778,779
<     │   │   ├───FB ID: 171
<     │   │   │   ├───Object ID: 171
---
>     │   │   ├───FB ID: 148
>     │   │   │   ├───Object ID: 148
781c781
<     │   │   │   ├───Format: XBGR2101010 (0x30334258)
---
>     │   │   │   ├───Format: ABGR16161616F (0x48344241)
784c784
<     │   │   │       └───Plane 0: offset = 0, pitch = 13760 bytes
---
>     │   │   │       └───Plane 0: offset = 0, pitch = 27520 bytes
811,812c811,812
<     │       ├───"FB_ID" (atomic): object framebuffer = 171
<     │       │   ├───Object ID: 171
---
>     │       ├───"FB_ID" (atomic): object framebuffer = 148
>     │       │   ├───Object ID: 148
814c814
<     │       │   ├───Format: XBGR2101010 (0x30334258)
---
>     │       │   ├───Format: ABGR16161616F (0x48344241)
817c817
<     │       │       └───Plane 0: offset = 0, pitch = 13760 bytes
---
>     │       │       └───Plane 0: offset = 0, pitch = 27520 bytes
1016,1017c1016,1017
<     │   │   ├───FB ID: 174
<     │   │   │   ├───Object ID: 174
---
>     │   │   ├───FB ID: 166
>     │   │   │   ├───Object ID: 166
1027,1028c1027,1028
<     │       ├───"FB_ID" (atomic): object framebuffer = 174
<     │       │   ├───Object ID: 174
---
>     │       ├───"FB_ID" (atomic): object framebuffer = 166
>     │       │   ├───Object ID: 166
1036,1037c1036,1037
<     │       ├───"CRTC_X" (atomic): srange [INT32_MIN, INT32_MAX] = 1720
<     │       ├───"CRTC_Y" (atomic): srange [INT32_MIN, INT32_MAX] = 720
---
>     │       ├───"CRTC_X" (atomic): srange [INT32_MIN, INT32_MAX] = 799
>     │       ├───"CRTC_Y" (atomic): srange [INT32_MIN, INT32_MAX] = 386
```

vaxerski:
`ABGR16161616F` into KMS what the _fuck_

vaxerski:
fuck it I'm doing a hyprland development

image

MajedAlghoul:
update: running main version, and setting
render {
    cm_enabled = false
}
fixes the indigo issue for me and displays colors seemingly identical to gnome running the same icc profile

thanks.

Flat:
> update: running main version, and setting
> render {
>     cm_enabled = false
> }
> fixes the indigo issue for me and displays colors seemingly identical to gnome running the same icc profile
> 
> thanks.

This would disable the ICC profile from being applied. I checked my profile on gnome and it will not apply ICC profiles globally unless they are calibrated profiles. If in gnome it has an information icon that you mouse over it will say this profile cannot be used for whole display color correction or something to that effect.

MajedAlghoul:
> > update: running main version, and setting
> > render {
> > cm_enabled = false
> > }
> > fixes the indigo issue for me and displays colors seemingly identical to gnome running the same icc profile
> > thanks.
> 
> This would disable the ICC profile from being applied. I checked my profile on gnome and it will not apply ICC profiles globally unless they are calibrated profiles. If in gnome it has an information icon that you mouse over it will say this profile cannot be used for whole display color correction or something to that effect.

well mine applies globally in both cases, hyprland and gnome, i checked gnome and yeah my profile doesnt have an information icon

wilecoyote2015:
Please correct me if I'm wrong - my info on gnome and KDE CM is a bit outdated (X11 times ..):
In Gnome and KDE, the profiling information is not used for transforming displayed colors by the compositor.
Only the calibration information of the profile is used for that.
The intended workflow there was the following:
1. Calibrate your monitor to a desired response (e.g. make it match sRGB roughly)
2. Apply the calibration in the Compositor 
3. Profile the calibrated display to get an accurate parameterization of how your calibrated display actually responds
4. The profile is communicated to color managed applications so that they can do the transform and also more sophisticated stuff like different rendering intents, black level correction, soft proofing and so on

The latter is BTW a good reason why one might like to disable CM for some application and let the application handle the display transform entirely: different rendering intents or blackpoint correction than Hyprland does may be desired

UjinT34:
>     4. The profile is communicated to color managed applications so that they can do the transform and also more sophisticated stuff like different rendering intents, black level correction, soft proofing and so on

With this PR the ICC profile isn't a part of the CM proto and any app supporting the proto will treat the monitor as a default unprofiled gamma22 sRGB. Some internals might also ignore the ICC part unless specifically modified to handle it.

wilecoyote2015:
I've also run the verification - but this time calibrated and verified with 10 bit and used the large verification chart.
As before, brighter colors are fine - darker ones are way off. Using srgb eotf
image
image

[Measurement Report 3.9.17 - Web @ localhost - 2026-03-05 17-04.html](https://github.com/user-attachments/files/25762007/Measurement.Report.3.9.17.-.Web.%40.localhost.-.2026-03-05.17-04.html)

vaxerski:
that tbh could suggest a precision problem because dark colors are the most susceptible to this

vaxerski:
we could likely try pushing F16 buffers into KMS later and see if that helps but that would be after ujin's renderer refactors I don't want to cause another rebase mess xD

MajedAlghoul:
Thinking about it. Maybe it has to do with screen capability to cover what percentage of srgb. I know my screen doesn't cover a good percentage. From what I understand using an icc forces srgb. Not sure if this makes sense. I just thought of sharing it.

Rabioli:
Hello, I don't really know where else to put this, there seems to be a github discussion from a year ago regarding icc profiles, but seems inactive, so I am deciding to write here and not opening a new issue since I believe it is related.

As hyprland updated I decided to test the icc profiles, since my monitor has an awful hdr by default, so I loaded some that I found on the web for my monitor, (it is the exact same model) [here is the link to the repo](https://github.com/StuffAround/LG_27GS60QC-B_AutodimmingFix), I tested them all and none of them worked, not even loaded, I enabled debug on hyprland to see what was going on and the only output regarding icc was `ERR ]: icc for DP-3 ("/home/user/color/LG_HDR800_DisplayCal_PURE_2.2.icm") failed: Invalid file` nothing else, I tested the same profile on KDE and it loaded and applied correctly, I also tested the other profiles in the repo, none of them work, so I am thinking it is not the profile, but I really have no idea why they wouldn't be loading.

Also I checked the profiles with displaycal and iccdump, they seem to be correctly formatted.

I will be checking the discussion closely, and test whatever I can, I don't have a colorimetrer so it would just be with the naked eye.

Zebra2711:
> Hello, I don't really know where else to put this, there seems to be a github discussion from a year ago regarding icc profiles, but seems inactive, so I am deciding to write here and not opening a new issue since I believe it is related.
> 
> As hyprland updated I decided to test the icc profiles, since my monitor has an awful hdr by default, so I loaded some that I found on the web for my monitor, (it is the exact same model) [here is the link to the repo](https://github.com/StuffAround/LG_27GS60QC-B_AutodimmingFix), I tested them all and none of them worked, not even loaded, I enabled debug on hyprland to see what was going on and the only output regarding icc was `ERR ]: icc for DP-3 ("/home/user/color/LG_HDR800_DisplayCal_PURE_2.2.icm") failed: Invalid file` nothing else, I tested the same profile on KDE and it loaded and applied correctly, I also tested the other profiles in the repo, none of them work, so I am thinking it is not the profile, but I really have no idea why they wouldn't be loading.
> 
> Also I checked the profiles with displaycal and iccdump, they seem to be correctly formatted.
> 
> I will be checking the discussion closely, and test whatever I can, I don't have a colorimetrer so it would just be with the naked eye.

icc not icm

vaxerski:
both icc and icm usually are the same. I don't know why the loading fails, it's done by lcms

PR #12726, opened 2025-12-26
Fix French quote in Splashes.hpp

ChancelierMartin:
Corrected the French quote in the splash text.

Fixed:
I fixed the French quote in src/helpers/Splashes.hpp, corrected “J'remue” to “Je remue…” since contraction is appropriate only before a vowel, matching the original French lyric from Indila – Dernière Danse. (Also I think Google Translate will correct you if you put the original in)

Just a string fix, ready for merging

### DIFF
--- a/src/helpers/Splashes.hpp
+++ b/src/helpers/Splashes.hpp
@@ -61,7 +61,7 @@ namespace NSplashes {
         "3 years!",
         "Beauty will save the world", // 4th ricing comp winner - zacoons' choice
         // music reference / quote section
-        "J'remue le ciel, le jour, la nuit.",
+        "Je remue le ciel, le jour, la nuit.", // I am assuming the song this is from is Indila - Dernière Danse, and the quote would be this 
         "aezakmi, aezakmi, aezakmi, aezakmi, aezakmi, aezakmi, aezakmi!",
         "Wir sind schon sehr lang zusammen...",
         "I see a red door and I want it painted black.",


C0Florent:
As a native French speaker, I think the current writing is correct. In French spoken language, it's common to merge "je" with the first syllable of next word, even it begins with a consonant. When we want to preserve this pronunciation detail in written form, it's common to do it be eliding the "e" and adding an apostrophe, as is the case for elision before a vowel sound.

You were not completely wrong, in more formal language this would be incorrect, as the elision will on only happen when the next word starts with a vowel sound (in which case the elision must happen)

I did listen to the song to make sure the elision is clearly audible, and to my ear it is, if you should be able to hear (if you can at least vaguely make out the words of the song) the "J're" is indeed pronounced as a single syllable, sung in a single note.

vaxerski:
image

src: https://genius.com/Indila-derniere-danse-lyrics

??

libmartinc:
> As a native French speaker, I think the current writing is correct. In French spoken language, it's common to merge "je" with the first syllable of next word, even it begins with a consonant. When we want to preserve this pronunciation detail in written form, it's common to do it be eliding the "e" and adding an apostrophe, as is the case for elision before a vowel sound.
> 
> You were not completely wrong, in more formal language this would be incorrect, as the elision will on only happen when the next word starts with a vowel sound (in which case the elision must happen)
> 
> I did listen to the song to make sure the elision is clearly audible, and to my ear it is, if you should be able to hear (if you can at least vaguely make out the words of the song) the "J're" is indeed pronounced as a single syllable, sung in a single note.

Thanks for your explanation about the current grammar.

I did a bit of digging before creating the PR and found that both forms appear depending on the source:
“Je remue”
https://www.frenchlearner.com/songs/derniere-danse-lyrics-translation/
https://www.musixmatch.com/lyrics/Indila-2/Dernière-danse
https://top40-charts.com/songs/lyrics.php?sid=37796&string=Derniere%20Danse

“J’remue”
https://genius.com/Indila-derniere-danse-lyrics
(From a Google search, excluding videos)

Most lyric transcription sites I found, my French teacher, and some French speaking friends believe that the lyric “Je remue”, which is why I initially thought the quote was incorrect. That said, I do not disagree with your point, the current form is defensible. I’ll close the PR

vaxerski:
I'd say this just sounds like english or german in this case:

_Star-spangled banner_:
**O'er** the ramparts we watched, were so gallantly streaming?

_Alles nur geklaut_:
Ich will dich gern **verführ'n**

Where you skip a vowel or merge two in order to fit into the rhythm.

C0Florent:
> Where you skip a vowel or merge two in order to fit into the rhythm

Yep, that's what I meant. It's even quite common in regular (regular as in non-singing) spoken language

PR #12736, opened 2025-12-28
Add window rules to suppress moving and resizing with mouse drag

lichie567:
This adds two window rules to suppress the moving and resizing of a window with mouse drag binds.
I sometimes find myself using an application that involves a lot of mouse dragging with modifier keys and I would like to prevent accidently move dragging the window. Especially in some games where they don't function well in fullscreen mode, I play them in window mode and would like to prevent resizing the window, which can make some games crash.

No

I'm not sure if the SDispatchResult should be changed to not return an error but to return `passEvent = true`. Other than that, I believe the code is ready, but additional PRs to update documentation will be needed.


vaxerski (commented):
looks good. Wiki mr + tests needed

lichie567:
Any recommendation for testing this? Is just testing that the properties get applied to the window enough? I was looking into simulating mouse click and drag but couldn't find any test api for it.

vaxerski:
click is there, I think move isnt so just add mouse move to the plugin

PR #12763, opened 2025-12-30
desktop/window: go back to the previously focused window in a group

vaxerski:
Ref https://github.com/hyprwm/Hyprland/discussions/12751

Please test to make sure it works as intended, it should tho. Added a test case too


Aqa-Ib:
The candidate is not the expected for me, it still returns the previous window in the group. I will try to debug it later. Thanks.

vaxerski:
yup should be good now

tekstryder:
Yup, all good now.

PR #12777, opened 2025-12-31
setprop dispatcher fix

IngilizAdam:

Fixes `hyprctl setprop` returning `unknown request` after updating to 0.53

`hyprctl setprop active opaque toggle` broke after #12269, the `dispatchSetProp` function and its `registerCommand` line got removed from `src/debug/HyprCtl.cpp`. The hyprctl client still sends the command, but the server doesn't recognize it anymore.

`hyprctl dispatch setprop` support is removed.

I noticed the old `dispatchSetProp` had a deprecation warning saying "use hyprctl dispatch setprop instead", which was working fine. So I'm not 100% sure if the removal was intentional or not.

If it was intentional, then i should've instead remove setprop from the client side.

But since the client wasn't updated and the #12269 PR had "Setprop/getprop" in the TODO list, I figured it was accidental/delayed.

So i thought it would make sense if `setprop` followed same usage pattern as `getprop`, e.g. instead of `hyprctl dispatch setprop ...` it should be `hyprctl setprop ...` just like `getprop`.

Code is ready based on my assumption of matching `setprop` usage with `getprop` usage. Built and tested, worked without an issue (mostly code relocation so not much to break here)

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -447,6 +447,39 @@ static void testBringActiveToTopMouseMovement() {
     Tests::killAllWindows();
 }
 
+static void testInitialFloatSize() {
+    NLog::log("{}Testing initial float size", Colors::GREEN);
+
+    Tests::killAllWindows();
+    OK(getFromSocket("/keyword windowrule match:class kitty, float yes"));
+    OK(getFromSocket("/keyword input:float_switch_override_focus 0"));
+
+    EXPECT(spawnKitty("kitty"), true);
+
+    {
+        // Kitty by default opens as 640x400, if this changes this test will break
+        auto str = getFromSocket("/clients");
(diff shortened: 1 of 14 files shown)


vaxerski (approved):

vaxerski:
massive tests failing :P

IngilizAdam:
i will look into it :D

IngilizAdam:
hey, upon closer look into the codebase, i've realized maybe my initial thoughts about the setprop usage was wrong. maybe the setprop should be accessible through dispatcher unlike getprop (since setprop will mostly be used by keybinds anyway).

i think i should just discard the changes.

what do you think @vaxerski, are there any improvements that i can add to the system? or should I just close the pr?

vaxerski:
`hyprctl setprop` was deprecated on purpose and its a dispatcher now

IngilizAdam:
yeah my bad lets just discard

PR #12781, opened 2025-12-31
Fix image description info events

UjinT34:
Fixes image description info events according to proto description. Should fix #12775
Stores EDID primaries and luminances as target values.
Always sends some `target_primaries` as suggested by https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/305
Might fix https://github.com/hyprwm/Hyprland/discussions/12810, needs testing
Applies HDR edid overrides to HDR modes only.

chromium and other EXT_LINEAR apps might need a HDR image description for correct CM. chromium currently sets an EXT_LINEAR image description with SDR luminances if it gets an SDR image description as preferred one. That trips both SDR and HDR mode. `quirks:prefer_hdr = 1` fixes this behaviour. The resulting brightness with a single monitor will use the value from EDID or from `monitorv2:max_avg_luminance` override. With multiple monitors it'll use the default HDR luminance 203.
Might be reasonable to change the default SDR luminance to 203 as it seems to be the desired value for SDR -> HDR transformation and will fix chromium without the need to set `quirks:prefer_hdr = 1`. But it can break some other apps.

Ready

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1244,15 +1244,13 @@ void CHyprOpenGLImpl::passCMUniforms(WP<CShader> shader, const NColorManagement:
     shader->setUniformFloat2(SHADER_DST_TF_RANGE, targetImageDescription->value().getTFMinLuminance(needsSDRmod ? sdrMinLuminance : -1),
                              targetImageDescription->value().getTFMaxLuminance(needsSDRmod ? sdrMaxLuminance : -1));
 
-    shader->setUniformFloat(SHADER_SRC_REF_LUMINANCE, imageDescription->value().getTFRefLuminance(-1));
-    shader->setUniformFloat(SHADER_DST_REF_LUMINANCE, targetImageDescription->value().getTFRefLuminance(-1));
+    shader->setUniformFloat(SHADER_SRC_REF_LUMINANCE, imageDescription->value().luminances.reference);
+    shader->setUniformFloat(SHADER_DST_REF_LUMINANCE, targetImageDescription->value().luminances.reference);
 
     const float maxLuminance = needsHDRmod ?
         imageDescription->value().getTFMaxLuminance(-1) :
         (imageDescription->value().luminances.max > 0 ? imageDescription->value().luminances.max : imageDescription->value().luminances.reference);
-    shader->setUniformFloat(SHADER_MAX_LUMINANCE,
-                            maxLuminance * targetImageDescription->value().luminances.reference /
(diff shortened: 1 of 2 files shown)


vaxerski:
can someone test this

fufexan:
Does not seem to fix the issue. In my case, colors are off only when mpv is in fullscreen.

UjinT34:
Fullscreen can look different when DS or passthrough is active or when HDR activates in fullscreen. Also if SDR/HDR switch is triggered then it'll send a new preferred image description which is handled by mpv and can result in color changes.
Other than that there should be no difference between windowed and fullscreen. And this PR doesn't alter anything regarding fullscreen specifically.

fufexan:
DS is active by default, and I don't have HDR enabled, nor HDR display capabilities. I guess mpv should take that into account even though the source video is HDR.

UjinT34:
That's not an mpv fault that CM is different with DS active. `render:non_shader_cm = 1` or 2 will block DS when CM without shader isn't supported. The default is currently 3 which disables the feature and ignores any DS issues because it produces less issues with hyprsunset and similar tools.

UjinT34:
Needs a retest. DS and cm_fs_passthrough should be off because they aren't fully supported.

skrmc:
@UjinT34 Please check https://github.com/hyprwm/Hyprland/discussions/12810#discussioncomment-15403055, where HDMI-A-1 is detected but ends up as `0x0@60` and the TV shows “No Signal” when starting from DM.

Might be related to CM/EDID/HDR changes in this PR (mastering primaries/luminances paths).

UjinT34:
`/dispatch exec sleep 1` test keeps failing. Should I disable it or is it ok to leave it as is?

5061726b6572:
For me, SDR content playback with mpv 0.41.0 appears to be same as it is with mpv 0.40.0 with this pull request.

njdom24:
This fixes `gamescope --force-composition` for me. Thank you so much!

alkeryn:
hey sorry for the bump guys, but this breaks something, before this commit, i can do `export ENABLE_HDR_WSI=1`
and then run mpv on a hdr video, and if i switch to fullscreen it'll turn my display into hdr mode.

after this commit, it no longer does it unless i add the --target-tcr=pq or hlg argument to mpv, then fullscreen.
dropping the commit and rebuilding revert to the old behavior where mpv doesn't need to be started with any extra arguments as long as the env variable is set.

also the brights don't seem as bright with this pr, maybe it's just me though.

PR #12802, opened 2026-01-01
desktop/window: use workArea for idealBB

### LINKED ISSUE #12800: getWindowInDirection no longer correct for monitor crosses since 0.53
### Discussion

https://github.com/hyprwm/Hyprland/discussions/12766

## Description

Since we check intersect, and now reserved is not counted within the ideal BB, this will be wrong.

`CCompositor::getWindowInDirection` `!useVectorAngles` path.

vaxerski:
Uses work area instead of reserved as work area includes gaps.

Should fix #12800 but needs a test @andresilva pls


andresilva:
Fixed 👌

vaxerski:
thansk

PR #12809, opened 2026-01-01
opengl: add a renderBlitFramebuffer fast path

gulafaran:
if we have no screenshader, no transforms we can use glBlitFramebuffer to blit offloadFB to the outFB. which is a much faster way to render the content.

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -878,10 +878,17 @@ void CHyprOpenGLImpl::end() {
         m_renderData.outFB->bind();
         blend(false);
 
-        if (m_finalScreenShader.program < 1 && !g_pHyprRenderer->m_crashingInProgress)
-            renderTexturePrimitive(m_renderData.pCurrentMonData->offloadFB.getTexture(), monbox);
-        else
-            renderTexture(m_renderData.pCurrentMonData->offloadFB.getTexture(), monbox, {});
+        bool fastPath = (m_finalScreenShader.program < 1) && (!g_pHyprRenderer->m_crashingInProgress) && (m_renderData.pMonitor->m_transform == WL_OUTPUT_TRANSFORM_NORMAL) &&
+            (m_renderData.mouseZoomFactor == 1.0f);
+
+        if (fastPath)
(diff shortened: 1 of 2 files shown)


vaxerski:
are we sure this is actually facster? IIRC most drivers just impl this as a textured pass

gulafaran:
> are we sure this is actually facster? IIRC most drivers just impl this as a textured pass

yeah cant say what drivers exactly do behind the scene but had to research besides my testing of it, i used a ceglsync and measured avg rendertimes of each call when fd syncs in a nested sessions, and tried to keep things as isolated as possible. saw a slight improvement. it all stems from seeing various other projects using it aswell https://github.com/mpv-player/mpv/commit/883d3114138a8f9a03f778165de553b7cdb99bee , `Use glBlitFramebuffer() instead of another shader pass. It should be
faster.` 

but since you ask i tried to research it a bit on various forums and it seems to be driver dependent. in my mind and as above committer it should be faster because of not requiring yet another shader pass etc then just simply `transfer a rectangle of pixel values from one region of a read framebuffer to another region of a draw framebuffer. ` 

but if all drivers suck and is slower or simply doing the exact same thing anyways on the "lesser" path i got no clue.

src/render/OpenGL.cpp
```diff
@@ -1875,6 +1882,23 @@ void CHyprOpenGLImpl::renderTextureInternal(SP<CTexture> tex, const CBox& box, c
...
+    const auto mHeight = (GLint)m_renderData.pMonitor->m_pixelSize.y;
+
+    m_renderData.damage.forEachRect([&](const auto& RECT) {
+        scissor(&RECT);
+        glBlitFramebuffer(0, 0, mWidth, mHeight, 0, 0, mWidth, mHeight, GL_COLOR_BUFFER_BIT, m_renderData.useNearestNeighbor ? GL_NEAREST : GL_LINEAR);
+    });
+
+    scissor(nullptr);
+    glBindFramebuffer(GL_READ_FRAMEBUFFER, 0);
+    glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
```
vaxerski, line 1899: this should restore the state from before. You can use currentfb from renderdata

gulafaran:
lets pause this idea until the last few stalls and hurdles have been solved so a proper benchmark can be done.

PR #12812, opened 2026-01-02
Better sdr eotf settings

UjinT34:
Changes global and per-monitor TF settings to a enum to avoid confusion and allows setting global to sRGB and per-monitor to gamma22 (previously this exact combo wasn't possible)
`default` or `0` - HL default, currently gamma22
`auto` - for global same as default, monitor uses global
`gamma22` or `1` - gamma22
`gamma22force` or `2` - global gamma22 and remap srgb to gamma22, monitor - gamma22
`srgb` or `3` - srgb

Legacy numeric values and values meant for one of the settings only won't be mentioned in the wiki.
Configs with `render:cm_sdr_eotf = 3` and `monitorv2:sdr_eotf = 0` (or unset) should be changed to use `monitorv2:sdr_eotf = srgb` if this is the desired setting (probably not)

Ready

### DIFF
--- a/src/helpers/TransferFunction.cpp
+++ b/src/helpers/TransferFunction.cpp
@@ -1,6 +1,6 @@
 #include "TransferFunction.hpp"
-#include "../managers/HookSystemManager.hpp"
 #include "../config/ConfigValue.hpp"
+#include "../event/EventBus.hpp"
 #include <string>
 #include <unordered_map>
 #include <hyprlang.hpp>
@@ -29,7 +29,7 @@ std::string NTransferFunction::toString(eTF tf) {
 eTF NTransferFunction::fromConfig() {
     static auto PSDREOTF = CConfigValue<Hyprlang::STRING>("render:cm_sdr_eotf");
     static auto sdrEOTF  = NTransferFunction::fromString(*PSDREOTF);
-    static auto P        = g_pHookSystem->hookDynamic("configReloaded", [](void* hk, SCallbackInfo& info, std::any param) { sdrEOTF = NTransferFunction::fromString(*PSDREOTF); });
+    static auto P        = Event::bus()->m_events.config.reloaded.listen([]() { sdrEOTF = NTransferFunction::fromString(*PSDREOTF); });
 
     return sdrEOTF;
 }


src/helpers/Monitor.cpp
```diff
@@ -473,11 +475,27 @@ void CMonitor::onDisconnect(bool destroy) {
...
+        case NTransferFunction::TF_SRGB: return NColorManagement::CM_TRANSFER_FUNCTION_SRGB;
+
+        case NTransferFunction::TF_AUTO: // use global setting
+            const auto sdrEOTF = NTransferFunction::fromString(*PSDREOTF);
+            switch (sdrEOTF) {
+                case NTransferFunction::TF_AUTO: return NColorManagement::CM_TRANSFER_FUNCTION_GAMMA22;
+                default: return chooseTF(sdrEOTF);
+            }
+    }
+}
```
UjinT34, line 502: Looks like build action isn't smart enough to handle a bit of recursion. Or am I missing some case?
```
/__w/Hyprland/Hyprland/src/helpers/Monitor.cpp: In function ‘NColorManagement::eTransferFunction chooseTF(NTransferFunction::eTF)’:
/__w/Hyprland/Hyprland/src/helpers/Monitor.cpp:494:1: error: control reaches end of non-void function [-Werror=return-type]
  494 | }
      | ^
```

vaxerski: you have no default: thus some value of `1337` would reach end of non void

UjinT34: Is it allowed to be passed as `NTransferFunction::eTF`? The compiler should block that instead...

vaxerski: yes because enums are ints under the hood and nothing stops you from `sc<eTF>(1337)`

src/render/OpenGL.cpp
```diff
@@ -1562,13 +1563,21 @@ static bool isHDR2SDR(const NColorManagement::SImageDescription& imageDescriptio
 
 void CHyprOpenGLImpl::passCMUniforms(SShader& shader, const NColorManagement::PImageDescription imageDescription, const NColorManagement::PImageDescription targetImageDescription,
                                      bool modifySDR, float sdrMinLuminance, int sdrMaxLuminance) {
-    static auto PSDREOTF = CConfigValue<Hyprlang::INT>("render:cm_sdr_eotf");
+    static auto PSDREOTF = CConfigValue<Hyprlang::STRING>("render:cm_sdr_eotf");
+    const auto  sdrEOTF  = NTransferFunction::fromString(*PSDREOTF);
```
vaxerski: quite inefficient, I would just store the precomputed value and update it with a hook to postConfigReload or whatever it was called configReloaded or sumn via g_pHookSystemManager

Tom94:
Hi @UjinT34 , let me preface this by saying *thank you* for implementing color management in Hyprland. Caring about colors and HDR, your work is *the* reason I'm able to daily drive Hyprland instead of KDE/Gnome.

But -- apologies for interjecting out of nowhere -- I feel like Hyprland is heading into a needlessly complicated direction with the gamma22/srgb setting.

**TLDR:** Imo, if anything, `gamma22force` should be the default.

The reason is simple: most monitors implement the gamma22 eotf, so 99% of content has been created by people looking at a gamma22 monitor -- and such content has been either untagged or tagged as sRGB pretty much everywhere in the wild (either by convention or inadvertently, e.g. by writing to a GL sRGB buffer).

Take any terminal emulator for example. If you used it without CM, or on X11, like most people have until recently, you'll have configured your terminal's ostensibly sRGB colors while viewing them with a gamma22 eotf if you haven't paid extra close attention to your monitor. The only reasonable thing to do is to forcibly reinterpret the (again, ostensibly) sRGB colors as gamma22 -- and that's the `gamma22force` option.

From a practical perspective: on my machines, `gamma22force` is the only setting that makes all my apps render their color schemes harmoniously without washing them out. (Also looking equal to my X11 i3+picom setup that mirrors my Hyprland config.) It also fixes [this issue](https://github.com/hyprwm/Hyprland/discussions/12788).

The only situation where it makes sense to treat sRGB content and gamma22 content separately is when a user tells Hyprland that they know what they're doing. I know this is irksome and I totally get the desire to actually follow the spec, but... that ship has sailed.

If it's still on the table to change it, I would separate the setting into two:
1. `monitorv2:sdr_eotf` with default `gamma22` and other possible value `srgb`
2. `monitorv2:treat_srgb_as_gamma22` with default `true`. Untagged surfaces should be treated as `gamma22` if this setting is `true` and as `srgb` if not (because sRGB *is* the conventional colorspace, even if it was always viewed through the gamma22 eotf, blergh). **Also important:** this setting should apply even in `cm = hdr` for SDR content to not be washed out.

That should cover all bases while being 10x easier to understand as a user. If a user has a display with actual sRGB eotf, they can set `sdr_eotf` to `srgb` while leaving `treat_srgb_as_gamma22 = true` to still get the same correct colors they would get on a regular display. Or they can set `treat_srgb_as_gamma22 = false` if they're a content creator that wants to create actual sRGB content for some strange reason -- in practice, nobody wants to do this because they know their content will be consumed on gamma22 eventually, but hey, the option would exist and it'd be obvious to understand.

(Of course, feel free to ignore all of this. Just making a suggestion as an outsider who might be missing something crucial here. As a user, it's trivial for me to just use `gamma22force` and call it a day.)

njdom24:
@Tom94 The problem of people choosing the `srgb` TF possibly due to being misled the name was what inspired version 2 of the proto, so things could be better by default with https://github.com/hyprwm/Hyprland/pull/12817

The idea (to my understanding) is to treat clients binding v1 and choosing the sRGB TF to be shoved into Gamma 2.2. Clients binding v2 would need to use the new "Compound Power 2.4" name if they truly want the sRGB TF, which should cut down on the problem in the future.

With this in mind (and admittedly as a `gamma22force` user myself), I'm inclined to at _least_ agree with `gamma22force` for clients binding version 1 of the proto.

UjinT34:
> But -- apologies for interjecting out of nowhere -- I feel like Hyprland is heading into a needlessly complicated direction with the gamma22/srgb setting.
> 
> **TLDR:** Imo, if anything, `gamma22force` should be the default.
> 
> The reason is simple: most monitors implement the gamma22 eotf, so 99% of content has been created by people looking at a gamma22 monitor -- and such content has been either untagged or tagged as sRGB pretty much everywhere in the wild (either by convention or inadvertently, e.g. by writing to a GL sRGB buffer).

There are two TFs involved: decode the source and encode for the output. Output TF has a separate setting and it's gamma22 by default (and `cm_sdr_eotf` doesn't change it by default with this PR).
`cm_sdr_eotf=gamma22force` is used for decoding only and is meant to fix bugged clients that set sRGB but mean gamma22.

> From a practical perspective: on my machines, `gamma22force` is the only setting that makes all my apps render their color schemes harmoniously without washing them out. (Also looking equal to my X11 i3+picom setup that mirrors my Hyprland config.) It also fixes [this issue](https://github.com/hyprwm/Hyprland/discussions/12788).

CM proto is a new thing. Apps that don't support it will be treated as gamma22 by default. And by default we trust apps that support the proto (no reason to support it unless you set the correct values). `gamma22force` technically is an incorrect thing, it might be used by default for version 1 clients after #12817 but it don't think it should be the global default.

> * `monitorv2:sdr_eotf` with default `gamma22` and other possible value `srgb`
> 
> * `monitorv2:treat_srgb_as_gamma22` with default `true`. Untagged surfaces should be treated as `gamma22` if this setting is `true` and as `srgb` if not (because sRGB _is_ the conventional colorspace, even if it was always viewed through the gamma22 eotf, blergh). **Also important:** this setting should apply even in `cm = hdr` for SDR content to not be washed out.

`monitorv2:sdr_eotf` has such settings with this PR. `monitorv2:treat_srgb_as_gamma22` doesn't make sense because it describes decoding behaviour that is controlled by `render:cm_sdr_eotf` and shouldn't depend on the monitor.

TL;DR
`render:cm_sdr_eotf` controls surface buffer decoding TF and should be set to `gamma22force` to achieve your desired setup.
`monitorv2:sdr_eotf` controls output encoding TF and should be left as is to achieve your desired setup.
Default values for both of the settings are called `default` and mean "let compositor decide". Their behaviour might be changed in future so that user won't need to change anything to get expected results.

Tom94:
> The idea (to my understanding) is to treat clients binding v1 and choosing the sRGB TF to be shoved into Gamma 2.2. Clients binding v2 would need to use the new "Compound Power 2.4" name if they truly want the sRGB TF, which should cut down on the problem in the future.

FWIW, I think the deprecation of sRGB in cm v2 is a **terrible** idea. 😅 Rather than letting the compositor handle monitor quirks, it pushes the burden onto app developers.

For example, I'm developing an image viewer. I load a PNG that reports itself as being sRGB, so I take its values and pipe them straight to a cm v1 compositor, reporting myself as sRGB. Job done from my perspective and the compositor has to deal with reinterpreting the values as gamma22 to fix the monitor debacle.

But under cm v2, this option no longer exists. As the app developer, I can either tell the compositor that it gets compound 2.4 values (which matches the PNG spec, so *should* be the correct choice but will render incorrectly), or I can disregard the PNG spec and report the values as gamma 2.2 (rendering correctly, but now I, as a random app developer, had to know about this, which 99.99% of devs don't). This is true for most image/video/content formats, of course. Just picked PNG as an example.

njdom24:
@Tom94 Compound power 2.4 is just a rename of the sRGB TF, so it wouldn't render incorrectly if that's the TF you want. As for choosing the correct TF, as a developer, [the protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/442/diffs#ed08b13455cc1bed2745c7ca72f502278b07f657_337_333) encourages 2.2 for this case explicitly.

Tom94:
@UjinT34 

> There are two TFs involved: decode the source and encode for the output. Output TF has a separate setting and it's gamma22 by default (and `cm_sdr_eotf` doesn't change it by default with this PR). `cm_sdr_eotf=gamma22force` is used for decoding only and is meant to fix bugged clients that set sRGB but mean gamma22.

> `monitorv2:sdr_eotf` has such settings with this PR.

I understand that there are two TFs, but I didn't realize that `monitorv2:sdr_eotf` already exists separately from `cm_sdr_eotf=gamma22force`.

In that case my suggestion (1) is superfluous -- it's already there. Amazing!

> And by default we trust apps that support the proto (no reason to support it unless you set the correct values). `gamma22force` technically is an incorrect thing, it might be used by default for version 1 clients after #12817 but it don't think it should be the global default.

I completely agree that `gamma22force` shouldn't apply to cm v2's `compoundpower24`. 100% agreed that it should only apply to cm v1's `srgb`. I'm advocating for it to be default behavior only for mapping `srgb`->`gamma22`.

Thanks for taking the time to read through my lengthy comment and replying to quickly!

@njdom24 The protocol might encourage devs to use gamma 2.2, but doesn't tell the devs that they should *reinterpret* their sRGB data as gamma 2.2. 😅 If I'd be reading the revised spec as a dev who doesn't know about the backstory, I'd read the spec, pick `gamma22`, but I'd also convert my sRGB image to gamma 2.2 before sending it over to the compositor, which again would be wrong. And god forbid the content was tagged as gamma 2.2 to begin with -- is the correct behavior to leave as-is, or to convert to sRGB and to *then* reinterpret as gamma 2.2? (Rhetorical question; my point are the necessary mental gymnastics for answering it.)

I'm not saying they should revise the wording of the protocol yet again, but that it's a huge mess that'd better be kept isolated from devs in the first place. Windows and macOS handle this perfectly... and on the topic of overcomplicating things for devs: I can't for the life of me understand how the Wayland protocol became such an overcomplicated mess. Why not just define a single color space + format that every compositor has to implement? Like scRGB to match Windows, or, better yet, ext. sRGB fp16 for backward compatibility (where HDR / wide gamut rendering is as simple as going beyond [0, 1] bounds)? Conversions are lossless, dammit, but I digress. Sorry for the rant; I know this has absolutely nothing to do with you and I 100% agree with all the points you've been making.

UjinT34:
> But under cm v2, this option no longer exists. As the app developer, I can either tell the compositor that it gets compound 2.4 values (which matches the PNG spec, so _should_ be the correct choice but will render incorrectly), or I can disregard the PNG spec and report the values as gamma 2.2 (rendering correctly, but now I, as a random app developer, had to know about this, which 99.99% of devs don't). This is true for most image/video/content formats, of course. Just picked PNG as an example.

- If you do something incorrect either by yourself or by relying on someone else then you should get incorrect results.
- When you combine two incorrect things you might get a correct outcome.
- `gamma22force` is an incorrect thing.

Pretty obvious stuff. Unfortunately we have to deal with pre-existing incorrect things that are out of our control and can't be easily distinguished from similar correct things. As an app developer you can show a notification and let the user decide which TF should be used (without any additional recoding). Or just ignore it and blame on the source file. If the source doesn't explicitly name an sRGB TF and only defines an sRGB colorspace then you can assume that it uses gamma22 TF. There are several colorspaces which use sRGB (BT.709) primaries but there are none I am aware of which use gamma22. Meaning that sRGB primaries + gamma22 is a widely used combo without a colorspace name. And most things that claim to use an sRGB colorspace actually use this combo.

> I completely agree that `gamma22force` shouldn't apply to cm v2's `compoundpower24`. 100% agreed that it should only apply to cm v1's `srgb`. I'm advocating for it to be default behavior only for mapping `srgb`->`gamma22`.

That won't solve your issue though because with an "sRGB" PNG you'll set compound power 2.4 with v2 proto and get the same issue as with sRGB and v1 proto. There is no way to tell whether the source has a correct sRGB setting.

Tom94:
> - If you do something incorrect either by yourself or by relying on someone else then you should get incorrect results.
> - When you combine two incorrect things you might get a correct outcome.
> - gamma22force is an incorrect thing.
>
> Pretty obvious stuff. Unfortunately we have to deal with pre-existing incorrect things that are out of our control and can't be easily distinguished from similar correct things. As an app developer you can show a notification and let the user decide which TF should be used (without any additional recoding). Or just ignore it and blame on the source file. If the source doesn't explicitly name an sRGB TF and only defines an sRGB colorspace then you can assume that it uses gamma22 TF. There are several colorspaces which use sRGB (BT.709) primaries but there are none I am aware of which use gamma22. Meaning that sRGB primaries + gamma22 is a widely used combo without a colorspace name. And most things that claim to use an sRGB colorspace actually use this combo.

To boil down my criticism of cm v2 in the language of "incorrect things", it's that the revised protocol forces devs to do an incorrect thing if they want correct results. I'd prefer if the "incorrectness" was kept confined to the compositor (== `gamma22force`) while devs can correctly pass (from their perspective) sRGB data to the compositor.

EDIT: and by all means, I like that cm v2 introduces `compoundpower24`. It's good to give devs explicit control in case they know what they're doing. What I'm criticizing is the deprecation of `srgb`.

> That won't solve your issue though because with an "sRGB" PNG you'll set compound power 2.4 with v2 proto and get the same issue as with sRGB and v1 proto. There is no way to tell whether the source has a correct sRGB setting.

Oh, sorry, I brought up the PNG example to criticise cm v2, unrelated to this PR. I went on a bit of a tangent responding to njdom24. I don't intend to rely on `gamma22force` to make my apps work correctly; sorry for muddling the discussion.

Wanting `gamma22force` to be the default is just me sharing my (perhaps superfluous) thoughts as a Hyprland user passionate about the topic.

gulafaran:
needs a rebase

vaxerski:
needs rebase

vaxerski:
yeeah it aint buildin chief

PR #12818, opened 2026-01-02
Groupbar horizontal textpadding

wbg:

Adds the config parameter group:groupbar:text_padding to allow left / right padding of text on the groupbar Decoration.

New to hyperland, not sure about the naming and if this is actually needed, but it scratched an itch for my setup.

It works on my local tests and it is not a big change.


wbg:
> wiki mr needed

done

Issue #3019, opened 2023-08-18
Separate border and window opacity values


noahgitsham:
### Description

Would be nice if we could choose individual values for the opacity of the borders and windows.

I.e. add variables `active_border_opacity` and `inactive_border_opacity`.

noahgitsham:
> or support rgba colors for the borders.

The reason I think there should be separate variables is you wouldn't want the window opacity and border opacity to multiply. I suppose this would not be an issue if the borders were simply made to ignore the current opacity variables even when the border_opacity variables are not set (though this would change how people's borders currently look). Honestly I'm not sure how it would be best implemented.

vaxerski:
you can use rgba for the borders...?

vaxerski:
this is a regression that got fixed a few days ago

noahgitsham:
> you can use rgba for the borders...?

I was thinking about this recetly, I thought this was a reply to Benguin, but I realise this may have been meant for me aswell. If I was not clear about this, it would be nice if there was a way to have opaque borders and a transparent window, which could be done if the window opacity did not affect the borders, and border opacity variables were added.

vaxerski:
border opacity variables are redundant as borders take rgba parameters.

noahgitsham:
The (in)active_opacity variables apply to the border aswell, even if the col.(in)active_border variables are set to FF alpha, full opacity. The opacities compound. Would be nice if they were independent, such that you could have opaque borders and transparent windows within the borders

PR #12822, opened 2026-01-03
make window decoration opacity independent of main window opacity. 

Localghost385:
Changes the render logic slightly to make border opacity and window opacity independent. previously it was impossible to have an opaque border on a transparent window.

Transparent borders are still possible, and can still have different levels of transparency for active and inactive windows using the alpha channel on border colour. 

I dont like that this implementation has 2 different measures of the transparency, a hex for border colour and a float for opacity, but that may just be me.

possibly could be more user friendly, but the code should be fine unless im missing something.

#### relevant issue
#3019

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -546,7 +546,7 @@ void CHyprRenderer::renderWindow(PHLWINDOW pWindow, PHLMONITOR pMonitor, const T
 
     EMIT_HOOK_EVENT("render", RENDER_PRE_WINDOW);
 
-    const auto fullAlpha = renderdata.alpha * renderdata.fadeAlpha;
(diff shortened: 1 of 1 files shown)


vaxerski:
honestly I think this is just complicating things for no reason

Localghost385:
i respect that lol, i might try rework this into a plugin since it seems like theres some demand for it. 
seems like a good compromise to me, allows for the functionality without affecting people who dont care.

That said, i do feel that this change does allow for a lot in terms of design flexibility, and to me at least, feels like it should be default behaviour.

20260103_22h45m42s_grim
20260103_22h46m43s_grim

the behaviour in the first image is not possible currently, as the alpha of the border and the main window compounds. If this is still considered out of scope though, I’ll move forward with the plugin.
thanks.

vaxerski:
it's probably better as a plugin

PR #12842, opened 2026-01-04
framebuffer: revert viewport

gulafaran:
to much stuff are relying on the viewport being set like this, just revert it to not regress further. this needs a overhaul.

worth mentioning this does not occur when using drm dumb buffers or software cursors. 

fixes: #12837


ikalco:
> to much stuff are relying on the viewport being set like this, just revert it to not regress further. this needs a overhaul.

+1

This made it quite confusing when trying to render correct window or region sharing in #11709
A lot of the renderer relies on the monitor size instead of fb size, like viewport, damage transforms, matricies, probably more

PR #12845, opened 2026-01-04
start: use nixGL if Hyprland is nix but not NixOS

vaxerski:
cc @fufexan

### DIFF
--- a/start/src/main.cpp
+++ b/start/src/main.cpp
@@ -53,7 +53,7 @@ int main(int argc, const char** argv, const char** envp) {
     terminateChildOnSignal(SIGTERM);
     terminateChildOnSignal(SIGINT);
 
-    int  startArgv = -1;
+    int startArgv = -1;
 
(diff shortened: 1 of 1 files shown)


start/src/helpers/Nix.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    return false;
+}
+
+std::expected<void, std::string> Nix::nixEnvironmentOk() {
+    if (!shouldUseNixGL())
+        return {};
+
+    if (!executableExistsInPath("nixGL"))
+        return std::unexpected("Hyprland was pulled from Nix, which requires nixGL to be installed as well.\nYou can install nixGL by running \"nix profile install "
```
fufexan:
```suggestion
        return std::unexpected("Hyprland was installed using Nix, but you're not on NixOS. This requires nixGL to be installed as well.\nYou can install nixGL by running \"nix profile install "
```

fufexan (changes requested):

PR #12847, opened 2026-01-04
windowrulev2 layerrulev2 error messages

ItsOhen:
The amount of people in discord getting confused is to damn high!

Errors when using v2.

No.

Yes.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -400,15 +400,9 @@ static Hyprlang::CParseResult handleWindowrule(const char* c, const char* v) {
 }
 
 static Hyprlang::CParseResult handleWindowrulev2(const char* c, const char* v) {
-    const std::string      VALUE   = v;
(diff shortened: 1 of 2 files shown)


src/config/ConfigManager.cpp
```diff
@@ -2983,6 +3014,10 @@ std::optional<std::string> CConfigManager::handleLayerrule(const std::string& co
     return std::nullopt;
 }
 
+std::optional<std::string> CConfigManager::handleLayerrulev2(const std::string& command, const std::string& value) {
```
vaxerski: at this point just return that directly whats the point of a member of mgr

vaxerski:
also no need for that snarky ai comment

src/config/ConfigManager.cpp
ItsOhen: To not confuse whoever touches it next.
But okey, let's do direct return.

PR #12850, opened 2026-01-04
fix: track explicit workspace assignments to prevent X11 configure overwrites

jmylchreest:
## Summary

  X11 windows that send configure requests (such as borderless fullscreen games resizing to fullscreen) have their workspace reassigned in `onX11ConfigureRequest()` based on window position. This overwrites any explicit workspace assignment from window rules or user actions, as the window gets matched to the active workspace on whatever monitor it appears on.

  This is most visible when a window rule assigns a window to a special workspace - the window briefly appears on the special workspace, then immediately gets reassigned to the regular workspace when the X11 configure request is processed.

  **Root cause:** `onX11ConfigureRequest()` unconditionally sets `m_workspace` based on which monitor the window's center is on:

```
m_workspace = g_pCompositor->getMonitorFromVector(...)->m_activeWorkspace;
```

Fix: Track when workspaces are explicitly assigned (via window rules or user movetoworkspace actions) using a new m_workspaceExplicitlyAssigned flag, and skip the position-based reassignment if the flag is set. There was an earlier version of this that simply returned if the workspace was a special workspace in onX11ConfigureRequest but I felt it didn't fully address the issue.

Changes
  - Add m_workspaceExplicitlyAssigned flag to CWindow
  - Set flag when window rules assign a workspace
  - Set flag when user moves window via movetoworkspace dispatcher
  - Check flag in onX11ConfigureRequest to preserve intentional assignments
  - Add debug logging for explicit workspace assignments

Test plan

- [ ] Create a window rule assigning an X11 app to a special workspace (e.g., workspace = special:games for a Steam game)
- [ ] Launch a borderless fullscreen game (likely the easiest way to test... e.g., Manor Lords, Whiskerwood)
- [ ] Verify the window stays on the special workspace instead of being reassigned to the active workspace
- [ ] Verify user movetoworkspace commands are also preserved
- [ ] Verify normal floating X11 windows without explicit assignments still get reassigned based on position (existing behavior)

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -1495,16 +1495,22 @@ void CWindow::onX11ConfigureRequest(CBox box) {
     if (!m_workspace || !m_workspace->isVisible())
         return; // further things are only for visible windows
 
-    // Preserve explicit workspace assignments (from rules or user actions) - don't reassign based on position
-    if (m_workspaceExplicitlyAssigned) {
-        auto newWorkspace = g_pCompositor->getMonitorFromVector(m_realPosition->goal() + m_realSize->goal() / 2.f)->m_activeWorkspace;
-        if (m_workspace != newWorkspace)
-            Log::logger->log(Log::DEBUG, "onX11ConfigureRequest: preserving explicit workspace '{}' (would have changed to '{}') for window '{}'", m_workspace->m_name,
-                             newWorkspace->m_name, m_title);
-        return;
-    }
+    const auto monitorByRequestedPosition = g_pCompositor->getMonitorFromVector(m_realPosition->goal() + m_realSize->goal() / 2.f);
+    const auto currentMonitor             = m_workspace->m_monitor.lock();
 
-    m_workspace = g_pCompositor->getMonitorFromVector(m_realPosition->goal() + m_realSize->goal() / 2.f)->m_activeWorkspace;
+    Log::logger->log(
+        Log::DEBUG,
+        "onX11ConfigureRequest: window '{}' ({:#x}) - workspace '{}' (special={}), currentMonitor='{}', monitorByRequestedPosition='{}', pos={:.0f},{:.0f}, size={:.0f},{:.0f}",
+        m_title, (uintptr_t)this, m_workspace->m_name, m_workspace->m_isSpecialWorkspace, currentMonitor ? currentMonitor->m_name : "null",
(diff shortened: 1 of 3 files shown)


jmylchreest:
I think I tried code similar to that initially but suffered a problem with the edge case of when the special workspace is stolen by another monitor, or maybe it was a special workspace on monitor 2... I'd need to re-read/test it - but that probably won't be for a few days now.

jmylchreest:
> or just dont change the workspace if the window didnt move to another monitor?

So the issue here is that assuming the window is on monitor 2, the positional (new) monitor will always be monitor 1, because the xwayland pos is 0,0 which makes it move. I've got  patch which respects the specialWorkspace conditions which is much cleaner, but there is a nuance with mapping new specialWorkspace windows (they're mapped to the positional monitor on creation, not the monitor that is currently selected/focused by the user). When I determine a clean way to solve that I'll push it to this PR branch.

jmylchreest:
On reflection I don't think we need to worry about the edge case in this fix of following the user  (ie: mouse on DP-2 when the app first launches/creates a window/workspace on DP-1). Pushed an updated version.

src/desktop/view/Window.cpp
```diff
@@ -1499,7 +1495,29 @@ void CWindow::onX11ConfigureRequest(CBox box) {
...
+    Log::logger->log(Log::DEBUG,
+                     "onX11ConfigureRequest: window '{}' ({:#x}) - workspace '{}' (special={}), currentMonitor='{}', monitorByRequestedPosition='{}', pos={:.0f},{:.0f}, size={:.0f},{:.0f}",
+                     m_title, (uintptr_t)this, m_workspace->m_name, m_workspace->m_isSpecialWorkspace, currentMonitor ? currentMonitor->m_name : "null",
+                     monitorByRequestedPosition ? monitorByRequestedPosition->m_name : "null", m_realPosition->goal().x, m_realPosition->goal().y, m_realSize->goal().x,
+                     m_realSize->goal().y);
+
+    // If staying on the same monitor, no workspace change needed
+    if (monitorByRequestedPosition == currentMonitor) {
+        Log::logger->log(Log::DEBUG, "onX11ConfigureRequest: skipping (sameMonitor=true)");
+        return;
```
vaxerski: maybe move this reassignment to a helper fn because if this hits you skip a bunch of important logic further down

jmylchreest: i think looking to return early on both, separately, was not needed. I've refactored to instead just require both conditions to be true (will move monitor and not special) and reassign it conditionally on that. It's much cleaner and avoids the helper.

Will push it now

vaxerski (approved):
thanks!

PR #12851, opened 2026-01-04
fix: handle fullscreen windows on special workspaces

jmylchreest:
This PR fixes issues with fullscreen window detection when windows are on special workspaces. The core problem was that `inFullscreenMode()` only checked `m_activeWorkspace`, completely missing fullscreen windows on special workspaces.

When toggling special workspaces or moving them between monitors, the `m_aboveFullscreen` flag on layer surfaces wasn't being reset, causing waybar and other layer surfaces to remain incorrectly hidden or shown.

  - `inFullscreenMode()`: Now checks special workspace first since it renders on top of regular workspaces
  - New `getFullscreenWindow()` helper: Safely returns the fullscreen window from either active or special workspace, checking special workspace first
  - Updated callers: `shouldSkipScheduleFrameOnMouseEvent()`, `Renderer.cpp`, and `getFSImageDescription()` now use the new helper instead of directly accessing `m_activeWorkspace`
  - Layer surface reset in `setSpecialWorkspace()`: Resets `m_aboveFullscreen` for layer surfaces when:
    - Closing a special workspace
    - Opening a special workspace
    - Stealing a special workspace from another monitor

  ## Test Plan

  - [x] Launch a fullscreen application on a special workspace
  - [x] Toggle the special workspace on/off - layer surfaces should correctly hide/show
  - [x] Move fullscreen window between monitors on special workspace
  - [x] Toggle special workspace between different monitors - layer surfaces on old monitor should reset correctly
  - [x] Verify no crashes when moving cursor with fullscreen window on special workspace

### DIFF
--- a/hyprtester/src/tests/main/workspaces.cpp
+++ b/hyprtester/src/tests/main/workspaces.cpp
@@ -20,6 +20,91 @@ using namespace Hyprutils::Utils;
 #define UP CUniquePointer
 #define SP CSharedPointer
 
+static bool testSpecialWorkspaceFullscreen() {
+    NLog::log("{}Testing special workspace fullscreen detection", Colors::YELLOW);
+
+    CScopeGuard guard = {[&]() {
+        NLog::log("{}Cleaning up special workspace fullscreen test", Colors::YELLOW);
+        // Close special workspace if open
+        auto monitors = getFromSocket("/monitors");
+        if (monitors.contains("(special:test_fs_special)") && !monitors.contains("special workspace: 0 ()"))
+            getFromSocket("/dispatch togglespecialworkspace test_fs_special");
+        Tests::killAllWindows();
+        OK(getFromSocket("/reload"));
+    }};
+
+    getFromSocket("/dispatch workspace 1");
+    EXPECT(Tests::windowCount(), 0);
+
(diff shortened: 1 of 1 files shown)


jmylchreest:
Added. I *think* i've covered these correctly, although I do get several of these tests (not mine, but pre-existing ones) fail. I don't use Nix, and I assume thats why - but I can't verify, none of them look to be caused by my changes...


-   colors.cpp:16 - Color management preset test
    -   monitorsSpec should contain "colorManagementPreset": "wide" but doesn't
-   exec.cpp (via main.cpp:234) - Process spawning test
    -   Sleep was not spawned or several sleeps are running: pgrep returned ''
-   window.cpp:593 - Window count mismatch
    -   Tests::windowCount(), expected 3, got 2
-   window.cpp:599 - Floating state test
    -   str should contain "floating: 1" but doesn't
-  window.cpp:401 - Group fallback focus test
    -   str.contains("class: kitty_B"), expected true, got false
-   window.cpp:462, 476 - Initial float size test
    -   str.contains("size: 640,400"), expected true, got false

vaxerski:
some tests rely on the tester's configs (or lack thereof) so don't worry. Compare against main. I let the CI run and if it passes we good.

vaxerski (approved):

PR #12863, opened 2026-01-05
Add parent-death handling for BSDs

tagattie:

This PR adds a parent-death handling mechanism similar to Linux one for BSDs.

`prctl()` is a system call specific to Linux. So we cannot use it on BSDs.

FreeBSD has a system call `procctl()` which is similar to `prctl()`. We can use it with `PROC_PDEATHSIG_CTL`.

OpenBSD, NetBSD, and DragonFly BSD do not appear to have a similar mechanism. So intead of relying on a system call, we need to manually poll ppid to see if the parent process has died.

With the changes, the spawned Hyprland process is terminated when the launcher process exits, matching Linux behavior as closely as possible on BSD platforms.

No.

Yes. I hope so :-)

### DIFF
--- a/start/src/core/Instance.cpp
+++ b/start/src/core/Instance.cpp
@@ -52,16 +52,6 @@ void CHyprlandInstance::runHyprlandThread(bool safeMode) {
 #elif defined(__FreeBSD__)
         int sig = SIGKILL;
         procctl(P_PID, getpid(), PROC_PDEATHSIG_CTL, &sig);
-#elif defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
-        pid_t ppid = getppid();
-        if (fork() == 0) {
-            for (;;) {
-                if (getppid() != ppid)
-                    kill(getpid(), SIGKILL);
-                sleep(1);
-            }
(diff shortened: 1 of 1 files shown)


start/src/core/Instance.cpp
```diff
@@ -41,7 +47,22 @@ void CHyprlandInstance::runHyprlandThread(bool safeMode) {
...
+#if defined(__linux__)
         prctl(PR_SET_PDEATHSIG, SIGKILL);
+#elif defined(__FreeBSD__)
+        int sig = SIGKILL;
+        procctl(P_PID, getpid(), PROC_PDEATHSIG_CTL, &sig);
+#elif defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
+        pid_t ppid = getppid();
+        if (fork() == 0) {
+            for (;;) {
+                if (getppid() != ppid)
```
vaxerski: I would just remove this atp to be honest... it's not critical.

tagattie: Thanks for the feedback. I have just removed the fallback to ppid polling.

PR #12879, opened 2026-01-05
add internal gl formats and reduce internal driver format conversions

gulafaran:
add .glInternalFormat for most types from weston sources, add a isFormatYUV helper.

ensure we use the right internalFormat in our framebuffers, and dont create a depth on the stencil when we only attach the stencil.

make yuv formats use external target, external only formats should aswell.
add external support to CM frag, and fix a ext.frag typo

### DIFF
--- a/src/protocols/Screencopy.cpp
+++ b/src/protocols/Screencopy.cpp
@@ -192,15 +192,7 @@ void CScreencopyFrame::share() {
 }
 
 void CScreencopyFrame::renderMon() {
-    auto       TEXTURE = makeShared<CTexture>(m_monitor->m_output->state->state().buffer);
-    const auto format  = NFormatUtils::getPixelFormatFromDRM(m_monitor->m_output->state->state().drmFormat);
-
-    if (format->swizzle.has_value()) {
-        TEXTURE->bind();
-        TEXTURE->swizzle(format->swizzle.value());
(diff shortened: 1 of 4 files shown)


src/render/shaders/glsl/CM.frag
```diff
@@ -5,7 +5,7 @@
 precision highp float;
 in vec2 v_texcoord;
 uniform sampler2D tex;
-//uniform samplerExternalOES texture0;
+uniform samplerExternalOES tex_external;
```
UjinT34: this breaks shader compilation. it's supposed to show a warning notification and disable CM. for some reason notification is not shown if this happens during initial start. `hyprctl reloadshaders` is meant for debugging this stuff but with the current code it can't recover from the state with disabled CM.

gulafaran: yeah should work now i forgot to uncomment the `#extension GL_OES_EGL_image_external : require`

which provides samplerExternalOES

gulafaran: actually even had to remove it from the cm.frag, seems drivers wants samplers determined at compile time or all hell breaks loose. so if external is gonna be used it needs its own shader.

vaxerski (approved):
wonderful job

PR #12954, opened 2026-01-11
monitor: improve DS enter/exit

njdom24:
Makes DS enter/exit a bit more robust

- 30-frame debounce timer for entering DS
  - Should help cases like screenshots capturing a black buffer from race condition
- Reset scanout buffer when DS is blocked
  - Prevents cases where DS seems to be activated preemptively, causing signal dropout
  - My simplest test case was `xeyes`. Without this change, fullscreening _and_ un-fullscreening it with `render:direct_scanout = 1` dropped the signal on my monitor, taking a few seconds to recover
  - Also helps with running Steam through `gamescope -e -- steam -bigpicture`, in which case opening the overlay/sidebar dropped signal. It was annoying on a monitor, but on my TV, it would simply not recover without a manual modeset

Not sure how correct/excessive the condition for calling `m_lastScanout.reset()` is, but it does fix this issue for me.

Ready for testing. (I anticipate needing to make changes for correctness though.)

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1804,8 +1804,20 @@ uint16_t CMonitor::isDSBlocked(bool full) {
 
 bool CMonitor::attemptDirectScanout() {
     const auto blockedReason = isDSBlocked();
-    if (blockedReason)
+
+    // wait 30 frames without a blocker before enabling DS
+    if (blockedReason) {
+        if (!m_lastScanout.expired())
+            m_lastScanout.reset();
+
+        m_scanoutDebounceCounter = 0;
         return false;
+    }
+
+    if (m_scanoutDebounceCounter < 30) {
+        m_scanoutDebounceCounter++;
+        return false;
+    }
 
     const auto PCANDIDATE = m_solitaryClient.lock();
     const auto PSURFACE   = PCANDIDATE->getSolitaryResource();
--- a/src/helpers/Monitor.hpp
+++ b/src/helpers/Monitor.hpp
@@ -178,6 +178,7 @@ class CMonitor {
     // for direct scanout
     PHLWINDOWREF m_lastScanout;
     bool         m_scanoutNeedsCursorUpdate = false;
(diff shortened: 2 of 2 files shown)


src/helpers/Monitor.cpp
```diff
@@ -1804,8 +1804,20 @@ uint16_t CMonitor::isDSBlocked(bool full) {
 
 bool CMonitor::attemptDirectScanout() {
     const auto blockedReason = isDSBlocked();
-    if (blockedReason)
+
+    // wait 30 frames without a blocker before enabling DS
+    if (blockedReason) {
+        if (!m_lastScanout.expired())
+            m_lastScanout.reset();
```
UjinT34, line 1811: that breaks exit logic https://github.com/hyprwm/Hyprland/blob/8f8b31e7a66acb4eb3836aa2fe6d751e5889bdaa/src/render/Renderer.cpp#L1318

UjinT34:
Technically the whole thing should go inside `isDSBlocked` with a new reason to avoid frustration when `hyprctl monitors` reports that DS isn't blocked by anything but DS isn't working. Might not be that big of deal because it'll report that for less than a second.

UjinT34:
https://github.com/hyprwm/Hyprland/blob/8f8b31e7a66acb4eb3836aa2fe6d751e5889bdaa/src/helpers/Monitor.cpp#L1862 this one should trigger a debounce

src/helpers/Monitor.cpp
njdom24: Ouch, looks like all I'm doing with this is preventing a modeset, yeah? Dunno why everything is fine without it if there's a DRM format mismatch check is passing without the change.

I (evidently) don't know enough about the pipeline to know how to improve it :/

UjinT34: DS might not require a format change. Keeping the DS format without restoring the previous one might work fine. There aren't many options, usually it's just six formats for every combination of alpha-no alpha, rgb-bgr and 8-10bit.  It works for you by accident %)

PR #12975, opened 2026-01-13
shm: ensure we use right gl unpack alignment

gulafaran:
gl defaults to 4 and not all formats is divisible with 4 meaning its going to pad out ouf bounds and cause issues. check if the stride is divisible with 4 otherwise set it to 1, aka disable it.

GL_UNPACK_ALIGNMENT only takes 1,2,4,8 but formats like RGB888 has bytesPerBlock 3.

### DIFF
--- a/src/render/Texture.cpp
+++ b/src/render/Texture.cpp
@@ -76,9 +76,19 @@ void CTexture::createFromShm(uint32_t drmFormat, uint8_t* pixels, uint32_t strid
     if (format->swizzle.has_value())
         swizzle(format->swizzle.value());
 
+    bool alignmentChanged = false;
+    if (format->bytesPerBlock != 4) {
+        const GLint alignment = (stride % 4 == 0) ? 4 : 1;
+        GLCALL(glPixelStorei(GL_UNPACK_ALIGNMENT, alignment));
+        alignmentChanged = true;
+    }
+
     GLCALL(glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, stride / format->bytesPerBlock));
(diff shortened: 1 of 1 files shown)


src/render/Texture.cpp
```diff
@@ -76,9 +76,19 @@ void CTexture::createFromShm(uint32_t drmFormat, uint8_t* pixels, uint32_t strid
...
+        const GLint alignment = (stride % 4 == 0) ? 4 : 1;
+        GLCALL(glPixelStorei(GL_UNPACK_ALIGNMENT, alignment));
+        alignmentChanged = true;
+    }
+
     GLCALL(glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, stride / format->bytesPerBlock));
     GLCALL(glTexImage2D(GL_TEXTURE_2D, 0, format->glInternalFormat ? format->glInternalFormat : format->glFormat, size_.x, size_.y, 0, format->glFormat, format->glType, pixels));
     GLCALL(glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, 0));
+    if (alignmentChanged)
+        GLCALL(glPixelStorei(GL_UNPACK_ALIGNMENT, 4));
```
UjinT34, line 90: is it a correct condition/constant?

gulafaran: Its the default initial value, and so far isnt set anywhere else, so yeah. There is a glget* way to get it if the need would exist but glget* in general is horribla slow

UjinT34: if `GLCALL(glPixelStorei(GL_UNPACK_ALIGNMENT, 4))` is cheap then it should be called unconditionally. if it's not then `alignmentChanged = alignment != 4`. probably won't matter that much though

gulafaran: yeah glPixelStorei is generally cheap, true. the boolean could have been skipped but meh i rather rely on cpu then driver to not do something stupid like it does in 99% of other cases :D

PR #12977, opened 2026-01-13
windowrules: allow expression in min_size/max_size

C0Florent:

Adds expression support for min_size and max_size window props. Tested with both setprop and windowrules.
Feature request discussion link: https://github.com/hyprwm/Hyprland/discussions/12936

Also I added a minor fix in a second commit where setprop min_size was setting max_size (didn't bother to test if it actually caused a bug, assuming it obviously would), let me know if that should be a separate PR.

#### Is there anything  you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

The expression evaluation could be done at many other places, this might not be the best one, I just found it to work for now, still better than no expression support at all.
Obviously does not refresh when expression variables change, so the prop can get out of sync.
Also funny things will probably occur if someone sets their `min_size = window_w*1.1 window_h*1.1`, but that's on the user at this point. I still think having window_* variables in scope is useful because it allows constraining only width or height (i.e. `min_size = window_w 500`)

works for me, ready

### DIFF
--- a/hyprtester/src/tests/main/hyprctl.cpp
+++ b/hyprtester/src/tests/main/hyprctl.cpp
@@ -78,7 +78,7 @@ static bool testGetprop() {
     EXPECT(getCommandStdOut("hyprctl getprop class:kitty min_size -j"), R"({"min_size": [100,50]})");
 
     // expr-based min/max _size
-    getFromSocket("/dispatch setfloating class:kitty"); // need to set floating for tests below
+    getFromSocket("/dispatch setfloating class:kitty");                 // need to set floating for tests below
     getFromSocket("/dispatch setprop class:kitty max_size 90+10 25*2"); // set max to the same as min above, forcing window to 100*50
     EXPECT(getCommandStdOut("hyprctl getprop class:kitty max_size"), "100 50");
     EXPECT(getCommandStdOut("hyprctl getprop class:kitty max_size -j"), R"({"max_size": [100,50]})");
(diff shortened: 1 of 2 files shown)


C0Florent:
some tests fail but CI also fails on main so... no idea if i'm doing anything wrong

Edit: i managed to run the tests locally to check what's failing, there are tests related to setprop/getprop which are failing, i'll look into it

C0Florent:
fixed, i had just messed up trying to split my patch in two commits

C0Florent:
Added tests for setprop, no idea how to test it in windowrules

also should this be documented in the wiki? i guess all props should also support expressions so that it's easier to explain than the current state of "specifically min_size and max_size support expressions"

lastly i feel like i should mention use of expression in rules can have funny effects, but i guess it's on the user if they this kind of things:

https://github.com/user-attachments/assets/3943e1f6-ca5f-44b7-8891-6897a83631e3

vaxerski:
> no idea how to test it in windowrules

grep tester code for `keyword windowrule`, there are tests for rules.

> lastly i feel like i should mention use of expression in rules can have funny effects, but i guess it's on the user if they this kind of things

our job is to protect from lack of knowledge, not stupidity.

NicolasDorier:
I confirm this PR fixes my issue https://github.com/hyprwm/Hyprland/discussions/12936

C0Florent:
tests for windowrule added, however i'm confused because c-f fails on some file I haven't touched in this PR

PR #12988, opened 2026-01-14
egl: move over to use hyprgraphics

gulafaran:
use the hyprgraphics helpers instead.

depends on https://github.com/hyprwm/hyprgraphics/pull/42


Mohataseem89:
I especially like how the helpers centralize format handling and stride/pixel calculations this should make future EGL/DRM related changes easier and less error prone. The dependency on https://github.com/hyprwm/hyprgraphics/pull/42 also feels well-justified now that EGL formats and helpers live there.

one small followup thought: it might be worth documenting (briefly, either in code comments or docs) which EGL responsibilities are expected to stay in Hyprgraphics vs Hyprland, to avoid accidental drift back in the future

cpiber:
This uses the Egl.hpp header, which has not been released. @vaxerski would it be possible to release and bump dependencies?

vaxerski:
ah SHITE

vaxerski:
https://github.com/hyprwm/hyprgraphics/releases/tag/v0.5.1

PR #11741, opened 2025-09-17
Cache xyz primaries

UjinT34:
#### Describe your PR, what does it fix/add?
Calculates xyz primaries once per given primaries set.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Might need to transpose the calculated matrix.
Pointers and maps might not be correct/efficient.

#### Is it ready for merging, or does it need work?

- [ ] Replace SPCPRimaries with CPrimaries inside SImageDescription

PR #12992, opened 2026-01-15
optimise shader usage

gulafaran:
split CM rgba/rgbx into discard variants, so only when discard is active we hit the discard paths.

dont stall on vbo UV buffer if we are rendering multiple rects and hit the buffer before the gpu was done with the previous buffer, orphan it.

restore GL_PACK_ALIGNMENT to default initial value in screencopy/toplevel export.

use a single interleaved buffer for position and uv, should in theory improve cache locality. and also remove
the need to have two buffers around.


UjinT34:
`primaries2xyz` can be calculated by CPU and cached like in #11741 (some bits are merged in main trough a different PR)
Moreover these values a rarely used:
* for tonemapping when it doesn't return early
* for saturation when `monitor:sdrsaturation != 1` and SDR is converted to HDR

Caching those might solve #11719

gulafaran:
> `primaries2xyz` can be calculated by CPU and cached like in #11741 (some bits are merged in main trough a different PR) Moreover these values a rarely used:
> 
>     * for tonemapping when it doesn't return early
> 
>     * for saturation when `monitor:sdrsaturation != 1` and SDR is converted to HDR
> 
> 
> Caching those might solve #11719

yeah caching things would probably help a bit, i think il stop around here tho so the PR doesnt go all to big :p

fxzzi:
I (think?) it's this PR that makes my blur stretch


https://github.com/user-attachments/assets/b3a3590b-4ec0-4488-bd7b-461d061d048a

edit:

lol it's like a fun magnifying glass

https://github.com/user-attachments/assets/53d7594d-92cb-49b2-8fc4-723707281509

gulafaran:
> I (think?) it's this PR that makes my blur stretch
> 26.01.15-15.18.04.mp4
> 
> edit:
> 
> lol it's like a fun magnifying glass
> 26.01.15-16.17.47.mp4

ah, so yeah the precision loss was a good idea until it hits CM paths, where the exact floats where used for various calculations, or even just HDR. now its clamping and rounding those away and causing that. so in short cant do that. back to float and bit more bandwidth usage :(

gulafaran:
> I (think?) it's this PR that makes my blur stretch
> 26.01.15-15.18.04.mp4
> 
> edit:
> 
> lol it's like a fun magnifying glass
> 26.01.15-16.17.47.mp4

yeh fixed, reverted the precision drop. wont work when it hits CM math paths, rounding/clamping away important bits.

UjinT34:
I doubt it's the CM. Looks like it's the blur shader doing multiple fetches of neighbouring pixels. Border rounding and shadows might be affected as well.

vaxerski:
woops gh showed me only the last small commit, is this  gtg?

gulafaran:
> woops gh showed me only the last small commit, is this gtg?

sure, @UjinT34  filled in with some last minute additions to cache primaries with a bit of newfound information regarding opengles shader tips from nvidia docs. but the previous ones has been battletested by our hard working warriors like @fxzzi

vaxerski (approved):
ok

PR #12994, opened 2026-01-15
restore invisible floating window alpha/opacity when focused over fullscreen

NaufalH27:
there is opacity bug when changing focus to floating window inside workspacd that have fullscreen, this happend inside function `void CCompositor::setWindowFullscreenState()` at `src/Compositor.cpp`, specifically this line here
```
g_pDesktopAnimationManager->setFullscreenFadeAnimation(
        PWORKSPACE, PWORKSPACE->m_hasFullscreenWindow ? CDesktopAnimationManager::ANIMATION_TYPE_IN : CDesktopAnimationManager::ANIMATION_TYPE_OUT);
```
the function make every window opacity to 0 except the fullscreen window and pinned window, the problem is this only trigger once only when a window toggled to fullscreen mode. when the focus changed to a floating window (using cyclenext dispatcher) ,the floating window get above the fullscreen, but... **the bug here is that the floating window have opacity 0 when it above the fullscreen window**, this make look like something was blocking mouse input from the fullscreen window. like the video shown below: 

https://github.com/user-attachments/assets/1f6541c1-d3ea-4304-8d9e-0cdecc3cc52e

the one who block the mouse input is the invincible (0 opaque) floating window that that goes above the fullscreen window after focused using cyclenext dispatcher

my fix is to make the floating window opacity to 1 again when the window is focused

i tried to run the hyprtester locally, but its still failing on some testcase, maybe i need some help

yes?

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -327,7 +327,7 @@ APICALL EXPORT PLUGIN_DESCRIPTION_INFO PLUGIN_INIT(HANDLE handle) {
     HyprlandAPI::addDispatcherV2(PHANDLE, "plugin:test:keybind", ::keybind);
     HyprlandAPI::addDispatcherV2(PHANDLE, "plugin:test:add_rule", ::addRule);
     HyprlandAPI::addDispatcherV2(PHANDLE, "plugin:test:check_rule", ::checkRule);
-    HyprlandAPI::addDispatcherV2(PHANDLE, "plugin:test:floating_on_fullscreen", ::floatingFocusOnFullscreen);
+    HyprlandAPI::addDispatcherV2(PHANDLE, "plugin:test:floating_focus_on_fullscreen", ::floatingFocusOnFullscreen);
 
     // init mouse
     g_mouse = CTestMouse::create(false);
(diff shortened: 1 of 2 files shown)


vaxerski:
> i tried to run the hyprtester locally, but its still failing on some testcase, maybe i need some help

a few tester tests rely on specific configs and it's OK if they fail on your machine. What matters is that no _new_ fails are introduced.

src/managers/KeybindManager.cpp
```diff
@@ -418,6 +418,10 @@ void CKeybindManager::switchToWindow(PHLWINDOW PWINDOWTOCHANGETO, bool forceFSCy
             Desktop::focusState()->rawMonitorFocus(PNEWMON);
         }
     }
+    if (PWINDOWTOCHANGETO->m_isFloating && PWINDOWTOCHANGETO->m_workspace->m_hasFullscreenWindow) {
```
vaxerski: this is the wrong place for this compadre, check fullWindowFocus in focusState

NaufalH27: hmm i see i see

vaxerski (commented):
very nice compadre if you could just add a test case for this. You can add a dispatcher to the test plugin to return the goal of pWindow->m_alpha for a given window.

NaufalH27:
alright, wait

vaxerski (approved):
lg tv

PR #12998, opened 2026-01-15
no longer refocusing under cursor when the last window of a workspace closes and follow_mouse != 1

alkeryn:

this adds the new "switch_monitor_on_empty" option which fixed this issue #12991 when set to false, the default behavior is not changed if it is not used

i tested it, i confirmed that it solved my issue, i've not found it to cause any other issue, but this would be my first contribution to the project so any criticism is welcomed.

it's not a big change so i think it should be ready but you tell me.

thank you !

### DIFF
--- a/src/desktop/view/LayerSurface.cpp
+++ b/src/desktop/view/LayerSurface.cpp
@@ -277,8 +277,11 @@ void CLayerSurface::onUnmap() {
     //                                vvvvvvvvvvvvv if there is a last focus and the last focus is not keyboard focusable, fallback to window
     if (WASLASTFOCUS ||
         (Desktop::focusState()->surface() && Desktop::focusState()->surface()->m_hlSurface && !Desktop::focusState()->surface()->m_hlSurface->keyboardFocusable())) {
-        if (!g_pInputManager->refocusLastWindow(PMONITOR))
-            g_pInputManager->refocus();
+        if (!g_pInputManager->refocusLastWindow(PMONITOR)) {
+            static auto PFOLLOWMOUSE = CConfigValue<Hyprlang::INT>("input:follow_mouse");
+            if (*PFOLLOWMOUSE == 1)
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
why? I'd say switch if FFM = 1, otherwise don't. That's what FFM would imply, to me.

alkeryn:
Alright I'll change it when i have some time, works for me, i just didn't want to change behavior for people that liked it that way.
Thank you!

alkeryn:
alright, removed the new option and now doing it if follow_mouse != 1

aruhier:
I tested it, it works, thanks!

One problem I see however is some inconsistency between closing the last window and using movetoworkspacesilent. 

Let's say monitor `A` is the one on which I close or move the last window to an other workspace, `B` is an other monitor: using `movetoworkspacesilent` on `A` moves the focus to a window on `B` (but opening a window or changing workspace happens on `A`, as expected); closing a window doesn't move the focus at all (as expected).

alkeryn:
@aruhier interesting, i'll look into it when i have some time, never used movetoworkspacesilent.

aruhier:
I double checked with and without your patch, it definitely changes the behavior of `movetoworkspacesilent` (which is good), however not fully so it makes the focus a bit inconsistent.

I don't know if it's easily doable to align it fully on what closing a window does with your patch, that would be honestly perfect!

alkeryn:
@aruhier yea, i'll try to patch it as well, it'd make more sense that way. FM!=1 should mean never care about the mouse.
honestly i'm trying to patch hyprland into supporting all the features i used in bspwm.
though i'll probably make a plugging for the things that don't make sense to be in hyprland.
though it already has the dwindle layout which is a tree, it's just very limited compared to bspwm where you can directly manipulate the tree, transplant nodes, rotate etc...

PR #12890, opened 2026-01-06
Rethonk layouts from the ground up

vaxerski:
This is a monumental layout rewrite that changes fundamentally how we treat layouts. This is still a big WIP with many features missing.

Benefits:
 - Per-workspace layouts are trivial
 - Easier API, layouts no longer deal with pseudo, groups, special workspaces, gaps, etc.
   - The above also means all layouts support those by default
 - Less layout clutter overall
 - This _could_ allow for https://github.com/hyprwm/Hyprland/discussions/9905 to be possible but isn't the main goal here, further MRs could implement this as with the current architecutre it'd be much simpler to hack in
 
### Config changes / additions

- `layout:` workspacerule
- `scrolling` layout (all plugin:hyprscrolling: vars are still here, under scrolling:)
- `monocle` layout
- hyprctl workspaces has a layout field
- scrolling has `direction` and `layoutopt:direction` for left,right,up,down.

WIKI: https://github.com/hyprwm/hyprland-wiki/pull/1374

TODOs:
- [x] Master layout
- [x] Missing dwindle parts
- [x] Missing master parts
- [x] Groups
- [x] Layoutmsg
- [x] Tons of dispatchers missing
- [x] Deprecate some dispatchers like workspaceopt and splitratio
- [x] Migrate scroll into core with this MR
- [x] (?) Monocle layout
- [x] multimon move
- [x] Testing
- [x] Bring back layout selection, add per-workspace layouts
- [x] hyprctl workspaces needs a layout field
- [x] groupbar drag

Breaks layout plugins of course.

PR #13004, opened 2026-01-16
added the resizeactiveedge dispatcher that allows expanding/shrinking a window in any direction

alkeryn:

currently there is no way to change the expand / shrink a windows from all corners if it is floating or entirely surrounded by other windows, this allow you to resize in all directions.

you can use it like that `resizeactiveedge <direction> pixels` ie `resizeactiveedge u -10`

this should not collide with any existing thing

it works, you tell me.
thank you !


vaxerski:
I'm fine with this but it will have to wait for #12890

PR #13008, opened 2026-01-17
pointermgr: invalidate RBO, remove onRenderBufferDestroy

gulafaran:
invalidate the color attachment because we dont care what was before, set the damage to monitor size instead of INT16_MAX and remove onRenderbufferDestroy, renderbuffer already have a listener that destroys when buffer is destroyed.

### DIFF
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -572,8 +572,9 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
 
     RBO->bind();
 
-    g_pHyprOpenGL->beginSimple(state->monitor.lock(), {0, 0, INT16_MAX, INT16_MAX}, RBO);
-    g_pHyprOpenGL->clear(CHyprColor{0.F, 0.F, 0.F, 0.F});
+    const auto& damageSize = state->monitor->m_output->cursorPlaneSize();
+    g_pHyprOpenGL->beginSimple(state->monitor.lock(), {0, 0, damageSize.x, damageSize.y}, RBO);
+    g_pHyprOpenGL->clear(CHyprColor{0.F, 0.F, 0.F, 0.F}); // ensure the RBO is zero initialized.
 
     CBox xbox = {{}, Vector2D{m_currentCursorImage.size / m_currentCursorImage.scale * state->monitor->m_scale}.round()};
     Log::logger->log(Log::TRACE, "[pointer] monitor: {}, size: {}, hw buf: {}, scale: {:.2f}, monscale: {:.2f}, xbox: {}", state->monitor->m_name, m_currentCursorImage.size,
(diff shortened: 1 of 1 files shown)


src/managers/PointerManager.cpp
```diff
@@ -571,9 +571,10 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
     }
 
     RBO->bind();
+    RBO->getFB()->invalidate({GL_COLOR_ATTACHMENT0}); // we dont care what was before.
```
vaxerski: this is useless and too late. Invalidate is only useful for transient FBs before you end rendering

gulafaran: live life while still young. redacted the invalidate.

src/managers/PointerManager.cpp
```diff
@@ -572,8 +572,8 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
 
     RBO->bind();
 
-    g_pHyprOpenGL->beginSimple(state->monitor.lock(), {0, 0, INT16_MAX, INT16_MAX}, RBO);
-    g_pHyprOpenGL->clear(CHyprColor{0.F, 0.F, 0.F, 0.F});
+    g_pHyprOpenGL->beginSimple(state->monitor.lock(), {0, 0, state->monitor->m_size.x, state->monitor->m_size.y}, RBO);
```
vaxerski: since we're already doing this, the RBO is not the size of the monitor, but the size of the cursor plane

gulafaran: maxSize or CursorPlaneSize?

```
    auto        maxSize    = state->monitor->m_output->cursorPlaneSize();
    auto const& cursorSize = m_currentCursorImage.size;

    static auto PCPUBUFFER = CConfigValue<Hyprlang::INT>("cursor:use_cpu_buffer");

    const bool  shouldUseCpuBuffer = *PCPUBUFFER == 1 || (*PCPUBUFFER != 0 && g_pHyprRenderer->isNvidia());

    if (maxSize == Vector2D{})
        return nullptr;

    if (maxSize != Vector2D{-1, -1}) {
        if (cursorSize.x > maxSize.x || cursorSize.y > maxSize.y) {
            Log::logger->log(Log::TRACE, "hardware cursor too big! {} > {}", m_currentCursorImage.size, maxSize);
            return nullptr;
        }
    } else
        maxSize = cursorSize;
        ````

vaxerski: cursorPlaneSize

vaxerski:
just test this cuz idk if some viewport things wont be wonk

gulafaran:
> just test this cuz idk if some viewport things wont be wonk

tried a few different scaling, and resolutions no noticeable things yet

PR #13017, opened 2026-01-18
set deadline for main buffer

gulafaran:
attempt to calculate an estimated vblank period. and set deadline for it. to do that we need to get the fence from the buffer fd if there is one.

set deadline for newly attached buffers and main framebuffer

the idea here is this tells the kernel when a buffer should be done rendering, which allows
it to for example increase GPU clocks and or schedule it earlier in order to hopefully hit our deadlines


gulafaran:
replaced by https://github.com/hyprwm/Hyprland/pull/13387

PR #13023, opened 2026-01-18
added bindm swapwindow, swaps are atomic

alkeryn:
this is mostly porting a feature from bspwm as i'm switching from it and am a few features away from having everything i like.
basicaly it allows you to first click a mouse with that, ie super + click, and then move your mouse around, as soon as you hover another window, it'll instantly swap their position, you can of course swap back, or continue swaping the window with other windows to move your windows around before releasing the click.
if you hover an empty workspace it'll also instantly move the window there.

anyway, this is a very handy feature i use practicaly everyday, i don't wanna miss it, i'm binding bindm movewindow to CTRL + super now, having both is nice.

you can bind it like that :
`bindm = $mainMod, mouse:272, swapwindow`

i've tested it to work and not break existing functionality, does what i want it to, so should be good.

should be ready.
thanks !

### DIFF
--- a/src/layout/IHyprLayout.cpp
+++ b/src/layout/IHyprLayout.cpp
@@ -644,6 +644,27 @@ void IHyprLayout::onMouseMove(const Vector2D& mousePos) {
 
         DRAGGINGWINDOW->m_position = newPos;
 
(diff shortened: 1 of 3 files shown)


vaxerski:
I'll let this wait for #12890

PR #13030, opened 2026-01-19
Surface shader variants with less branching and uniforms

UjinT34:
Extracted branching code and relevant uniforms into separate include files. That files are used according to the required feature set so that there is no need for branching and excessive uniforms. Should increase performance and reduce bandwidth usage.
Surface shaders do not need extra branching for:
* rgba/rgbx
* discard opaque or alpha
* tint
* rounding
* CM
  * tonemap
  * sdr saturation&brightness

Surface shaders are compiled on demand without any fallback in case of a failure. Might introduce some hiccups when shader variant is compiled for the first time.
Same approach can be used with borders, shadows and blur.

Ready for review. Needs more testing with different sets of features.


gulafaran:
performance wise on this intel, it upped glmark2 by 27% here. thats gonna be dependent on HW and where the bottleneck sits. might be less or within margin of error on desktop gpus, or it might not. but certainly a big improvement here.

PR #13072, opened 2026-01-24
fix compiler errors

jmanc3:
my compiler errors at these three places every time I upgrade, so I thought might as well fix them upstream


vaxerski:
maybe because your compiler is outdated

PR #13073, opened 2026-01-24
Device specific binds

### LINKED ISSUE #11067: Implement per-keyboard binds
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11047

## Description

Allow setting per-keyboard binds (either include or exclude), will likely need a bind flag.

ArchSav:
Adds k bind flag for per-device binds
per device binds have the following syntax
`bindk = mods, key, [!]device1 device2 ..., dispatcher, params`
`!` is used to exclude the specified devices from using the bind

wiki mr hyprwm/hyprland-wiki#1354
Closes #11067
no

yes

### DIFF
--- a/hyprtester/src/tests/main/keybinds.cpp
+++ b/hyprtester/src/tests/main/keybinds.cpp
@@ -539,14 +539,6 @@ static void testPerDeviceKeybind() {
     EXPECT(attemptCheckFlag(20, 50), true);
     OK(getFromSocket("/dispatch plugin:test:keybind 0,0,29"));
     EXPECT(getFromSocket("/keyword unbind SUPER,Y"), "ok");
-
-    // Tags
-    EXPECT(checkFlag(), false);
-    EXPECT(getFromSocket("/keyword bindk SUPER,Y,test-tag,exec,touch " + flagFile), "ok");
-    OK(getFromSocket("/dispatch plugin:test:keybind 1,7,29"));
-    EXPECT(attemptCheckFlag(20, 50), true);
-    OK(getFromSocket("/dispatch plugin:test:keybind 0,0,29"));
-    EXPECT(getFromSocket("/keyword unbind SUPER,Y"), "ok");
 }
 
 static bool test() {
--- a/hyprtester/test.conf
+++ b/hyprtester/test.conf
@@ -219,11 +219,6 @@ device {
     sensitivity = -0.5
 }
 
-device {
-    name = test-keyboard-1
-    tags = test-tag
-}
-
 debug {
     disable_logs = false
 }
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -849,7 +849,6 @@ CConfigManager::CConfigManager() {
     m_config->addSpecialConfigValue("device", "keybinds", Hyprlang::INT{1});                 // enable/disable keybinds
     m_config->addSpecialConfigValue("device", "share_states", Hyprlang::INT{0});             // only for virtualkeyboards
     m_config->addSpecialConfigValue("device", "release_pressed_on_close", Hyprlang::INT{0}); // only for virtualkeyboards
-    m_config->addSpecialConfigValue("device", "tags", STRVAL_EMPTY);                         // only for keyboards and mice
 
     m_config->addSpecialCategory("monitorv2", {.key = "output"});
     m_config->addSpecialConfigValue("monitorv2", "disabled", Hyprlang::INT{0});
--- a/src/devices/IHID.hpp
+++ b/src/devices/IHID.hpp
@@ -2,7 +2,6 @@
 
 #include <cstdint>
 #include <string>
-#include <set>
 #include "../helpers/signal/Signal.hpp"
 
 enum eHIDCapabilityType : uint8_t {
@@ -37,7 +36,6 @@ class IHID {
         CSignalT<> destroy;
     } m_events;
 
-    std::string           m_deviceName;
-    std::string           m_hlName;
-    std::set<std::string> m_deviceTags;
+    std::string m_deviceName;
+    std::string m_hlName;
 };
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -686,12 +686,7 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
             continue;
 
         if (device) {
-            bool istagValid = false;
-            for (const auto& tag : device->m_deviceTags) {
-                if (k->devices.contains(tag))
-                    istagValid = true;
-            }
-            if (k->deviceInclusive ^ (k->devices.contains(device->m_hlName) || istagValid))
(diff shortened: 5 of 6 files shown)


Kommynct:
I would really like there to be some way to tag devices and use the tags instead of the device name itself

Dregu:
I've been using the branch I mentioned in the issue linked since august for my macropads. I think it's much simpler to configure, but I never made a pr cause the only person to comment said it's too confusing.

ArchSav:
> I would really like there to be some way to tag devices and use the tags instead of the device name itself

Seems like a good feature to add on

ArchSav:
> I've been using the branch I mentioned in the issue linked since august for my macropads. I think it's much simpler to configure, but I never made a pr cause the only person to comment said it's too confusing.

I would to disagree, using names of devices/tags makes the syntax much more readable. The use of bitmasks (especially in decimal format) would be harder to understand for those without an dev background.

Dregu:
> I would to disagree, using names of devices/tags makes the syntax much more readable. The use of bitmasks (especially in decimal format) would be harder to understand for those without an dev background.

Yeah well I am talking about one or two extra keyboards here, I don't know where we'd find a person with six keyboards who can't be taught binary at all. But anyway, if I can just get a fairly similar system, i.e. tag my keyboards `tags = 1`, `tags = 2`, write `bindk = mod, key, 1 2, ...`, then whatever, I'm still happy, it's the same as mine. Putting long device names on every bind was the problem, using short something fixes that problem.

What's also really important is that tagged devices can't trigger non-tagged binds, and non-tagged devices can't trigger tagged binds, or it gets complicated again.

Kommynct:
> > I would to disagree, using names of devices/tags makes the syntax much more readable. The use of bitmasks (especially in decimal format) would be harder to understand for those without an dev background.
> 
> Yeah well I am talking about one or two extra keyboards here, I don't know where we'd find a person with six keyboards who can't be taught binary at all. But anyway, if I can just get a fairly similar system, i.e. tag my keyboards `tags = 1`, `tags = 2`, write `bindk = mod, key, 1 2, ...`, then whatever, I'm still happy, it's the same as mine. Putting long device names on every bind was the problem, using short something fixes that problem.
> 
> What's also really important is that tagged devices can't trigger non-tagged binds, and non-tagged devices can't trigger tagged binds, or it gets complicated again.

with this system you don't have to be taught binary, this is obviously more readable, any system that requires specialized training to read is less readable

this would highly benefit from tagging inputs in their input configs that way you could wildcard tags though

Edit:nvm just saw the tags commit

ArchSav:
@vaxerski can you please review this?

vaxerski:
can you tag me after 0.54? I wanna push 0.54 soon so I dont wanna take big FRs atm and just focus on fixes

Kommynct:
@vaxerski you said to tag after, done

Dregu:
I've been thinking about this syntax, and shifting the offset of desc/device/whatever seems like hell if we want to add any more flags that require parameters. It's also more confusing for the user to put the params far away from the flag. Binds might actually need a rethonk, but meanwhile for an inline syntax how about just putting the params immediately after the flag in braces, like so: `bindk{1 !2 3} = ...` and generally just `bindabcx{some params}y{params of y}z{params for z} = ...`

I actually made a flag for workspace specific binds with this idea too, and it would be super simple to add more flags that require params. This would strip all the {whatever} sections from the args before passing to the switch loop.

```cpp
    auto BINDARGS        = command.substr(4);

    auto getCurlyFlag    = [&BINDARGS](char flag) -> std::string {
        std::string params;
        if (const auto FLAG_START = BINDARGS.find_first_of(flag); FLAG_START != std::string::npos && BINDARGS.size() > FLAG_START + 1 && BINDARGS[FLAG_START + 1] == '{') {
            if (const auto FLAG_END = BINDARGS.substr(FLAG_START).find_first_of('}'); FLAG_END != std::string::npos) {
                params   = BINDARGS.substr(FLAG_START + 2, FLAG_END - 2);
                BINDARGS = BINDARGS.substr(0, FLAG_START) + BINDARGS.substr(FLAG_START + FLAG_END + 1);
            }
        }
        return params;
    };

    auto wsSelector = getCurlyFlag('w');
    auto deviceArgs = getCurlyFlag('k');

    for (auto const& arg : BINDARGS) {
        switch (arg) {
```

And now we could support `bindk{keeb}w{m[DP-2]} = ...` for device-specific workspace-specific binds.

PowerUser64:
I'm a bit late to the party, but you could probably do this with a similar pattern to submaps:

```julia
bind_device = mykeyboard
	bind = ... # standard bind calls, but now they'll go to ^this device
bind_device = reset
# now the device is `reset` (similar to `submap = reset`)

bind = ... # this bind will exist for the default device(s), as if `bind_device` had never been called
```

This might be a bit more template-friendly than having extra arguments, since you could have keybinds written in a file and then source it for each keyboard you want to apply them to, without needing to introduce an argument depending on whether you want to bind to a specific device or not:

```julia
# keybinds.conf:
bind = SUPER, f, exec, echo "keybind F" > /tmp/keybind
bind = SUPER, d, exec, echo "keybind D" > /tmp/keybind

# keyboard-A.conf:
bind_device = keyboard-A
	source = $hypr/keybinds.conf
bind_device = reset

# keyboard-B.conf:
bind_device = keyboard-B
	source = $hypr/keybinds.conf
bind_device = reset

# hyprland.conf:
$hypr = ~/.config/hyprland
source = $hypr/keyboard-A.conf
source = $hypr/keyboard-B.conf
```

ArchSav:
I think PowerUser64's idea would probably be the best idea as it would allow for bulk definitions that would be more likely for device specific keybinds. Though I think having the inline syntax could also be quite useful

hyprtester/src/tests/main/keybinds.cpp
```diff
@@ -513,6 +513,34 @@ static void testSubmapUniversal() {
...
+    EXPECT(checkFlag(), false);
+    EXPECT(getFromSocket("/keyword bindk SUPER,Y,!test-keyboard-1,exec,touch " + flagFile), "ok");
+    OK(getFromSocket("/dispatch plugin:test:keybind 1,7,29"));
+    EXPECT(attemptCheckFlag(20, 50), false);
+    OK(getFromSocket("/dispatch plugin:test:keybind 0,0,29"));
+    EXPECT(getFromSocket("/keyword unbind SUPER,Y"), "ok");
+
+    // Tags
+    EXPECT(checkFlag(), false);
+    EXPECT(getFromSocket("/keyword bindk SUPER,Y,test-tag,exec,touch " + flagFile), "ok");
```
vaxerski: missing tests for description + kb

ArchSav:
This is ready for for merge. I will make another mr for the tags as well as some of the updated syntax suggested here.

Kommynct:
Also the device tagging would be more useful if the syntax input side was split into tag and set_tag that way you could set a bunch of input settings by specifying a tag instead of a name

would help with input settings needing copypasted quite a bit.

PR #13078, opened 2026-01-24
opengl: allow texture filter to be changed

jmanc3:
This lets mipmap filters be set when rendering a texture. (Wanted for my plugin, but also useful for hyprexpo and the like so that textures don't shimmer when scaled down).


jmanc3:
moved

PR #13091, opened 2026-01-25
eventloopmgr fixes

gulafaran:
move the idle functions instead of copying, avoids std::function copies.

the manpage for timerfd_create and read states this.

timefd_create creates a new timer object, and returns a file descriptor
that can be used to read the number of expirations that have occurred.

The FD becomes readable when the timer expires.

read removes the “readable” state from the FD.

so most likely it has somewhat worked because of the scheduleRecalc()
function.

### DIFF
--- a/src/managers/eventLoop/EventLoopManager.hpp
+++ b/src/managers/eventLoop/EventLoopManager.hpp
@@ -89,9 +89,9 @@ class CEventLoopManager {
         bool                             recalcScheduled = false;
     } m_timers;
 
-    SIdleData                                                 m_idle;
-    std::map<int, SEventSourceData>                           m_aqEventSources;
-    std::unordered_map<SReadableWaiter*, UP<SReadableWaiter>> m_readableWaiters;
+    SIdleData                        m_idle;
+    std::map<int, SEventSourceData>  m_aqEventSources;
+    std::vector<UP<SReadableWaiter>> m_readableWaiters;
 
     struct {
         CHyprSignalListener pollFDsChanged;
--- a/src/managers/eventLoop/EventLoopManager.cpp
+++ b/src/managers/eventLoop/EventLoopManager.cpp
@@ -80,7 +80,7 @@ static int handleWaiterFD(int fd, uint32_t mask, void* data) {
 }
 
 void CEventLoopManager::onFdReadable(SReadableWaiter* waiter) {
-    auto it = m_readableWaiters.find(waiter);
+    auto it = std::ranges::find_if(m_readableWaiters, [waiter](const UP<SReadableWaiter>& w) { return waiter == w.get() && w->fd == waiter->fd && w->source == waiter->source; });
 
     // ???
     if (it == m_readableWaiters.end())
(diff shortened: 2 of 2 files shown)


src/managers/eventLoop/EventLoopManager.hpp
```diff
@@ -89,9 +89,9 @@ class CEventLoopManager {
         bool                             recalcScheduled = false;
     } m_timers;
 
-    SIdleData                        m_idle;
-    std::map<int, SEventSourceData>  m_aqEventSources;
-    std::vector<UP<SReadableWaiter>> m_readableWaiters;
+    SIdleData                                                 m_idle;
+    std::map<int, SEventSourceData>                           m_aqEventSources;
+    std::unordered_map<SReadableWaiter*, UP<SReadableWaiter>> m_readableWaiters;
```
vaxerski: no raw ptrs

gulafaran: its a raw pointer to itself the unique_ptr. and the key cant be the fd if we want multiple waiters on same fd, and fds itself isnt globally unique numbers. 

there is also a subtle issue here that we need to pile ontop too, if we dont remove the event source before calling the fn() in onFdReadable it could trigger a dispatch/reentry on same data that has been destructed, causing UB in libwayland itself

vaxerski: rethink storage then

gulafaran: right, the main issue was re entries causing UB and we could probably live with vector then since the ptr is a unique_ptr, but there is no benchmarkable loss here of just using deque and a deque wont be affected by reallocation invalidating iterators or references, not that any of those are used right now here but, while here lets just make it as safe as possible. this fires at an incredible rate with enough clients and buffers flying around.

src/managers/eventLoop/EventLoopManager.hpp
```diff
@@ -89,9 +90,9 @@ class CEventLoopManager {
         bool                             recalcScheduled = false;
     } m_timers;
 
-    SIdleData                                                 m_idle;
-    std::map<int, SEventSourceData>                           m_aqEventSources;
-    std::unordered_map<SReadableWaiter*, UP<SReadableWaiter>> m_readableWaiters;
+    SIdleData                       m_idle;
+    std::map<int, SEventSourceData> m_aqEventSources;
+    std::deque<UP<SReadableWaiter>> m_readableWaiters;
```
vaxerski: use a vector you dont need a deque

gulafaran: after some thinking no i think we do or we sort of even need unordered_map, its the most unstable i would be willing to go and that is even still questionable.  and its because we are passing pointers to a C library.

if a waiter is in the event loop, another addwaiter gets called and triggers an reallocation 

`wl_event_loop_add_fd(g_pEventLoopManager->m_wayland.loop, waiter->fd.get(), WL_EVENT_READABLE, ::handleWaiterFD, waiter.get())`  wont the ptr waiter.get() now be invalid?

deque is atleast stable on emplacing back but deque still has the problem of potentially making the `waiter.get()` invalid on erase if its not removed from the end of the queue

this is why i wanted that unordered_map

vaxerski: brother, it's a UP

gulafaran: oh hm learn something every day. i thought the move of the unique_ptr on reallocation would mean the address now changed, had to test in godbolt and check cppreference. then yeah its only the reentry that was causing issues

PR #13094, opened 2026-01-25
fix: Handle tiled window cross monitor move edge case

kushagharahi:

If a window is partially positioned outside of the shared edge of two monitors, a `moveWindow` command would not move it across the boundary. 

One visual example of a move that wouldn't work before this change:
image

Relevant monitor config for the above:
```
# Right Monitor (DP-1)
monitor = DP-1, 2560x1440@165, 2160x1200, 1
# Left Monitor (DP-2) - transform 1 (Portrait)
monitor = DP-2, 3840x2160@60, 0x0, 1, transform, 1
```

A more correct change would be figuring out why `getWindowInDirection` is not returning null and allowing the code to proceed to the `PMONITORFALLBACK` logic. I suspect is has something to do with the transformed left monitor. 

It's a bandaid that improves the `moveWindow` workflow. Even if `getWindowInDirection` is improved, this check, I think won't hurt anything.


kushagharahi:
may be related to https://github.com/hyprwm/Hyprland/issues/12800?

Issue #11590, opened 2025-09-04
Subsurface geometry rounding has issues with occlusion


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11566

## Description
When moving around a floating window with an opaque subsurface, occlusion seems misaligned

jmanc3:
I think a systematic solution for window position, rounding, and rendering is worth considering. For instance, I think the related issue to this one where windows at 'non-safe' scales produce a one-pixel missing stripe probably also has to do with rounding. 

<img width="1168" height="629" alt="Image" src="https://github.com/user-attachments/assets/994b096a-650a-4055-931a-3bdc99f580ac" />

And then simply moving the position (not resizing) the window:

<img width="1168" height="629" alt="Image" src="https://github.com/user-attachments/assets/b94cac16-3ba5-42f1-8c12-2e1b5fe94d68" />

And then there is the some what related issue of Wayland window [geometry](https://wayland.app/protocols/xdg-shell#xdg_surface:request:set_window_geometry) is basically being ignored (with the always maximize hack):

<img width="2560" height="1599" alt="Image" src="https://github.com/user-attachments/assets/6c9a24b8-5cc0-4426-93f3-7243e93b8325" />

It might be worth it to figure out a good systemic solution to all three of these problems at the same time. I did a search and there are about a combined 400 references to either `m_RealPosition` or `m_RealSize` (for `windows`) in the code base. 

It's going to be painful but not too too impossible either.

As a plus, `debug:disable_scale_checks` could be got rid of, and let any scale be set.

vaxerski:
> As a plus, debug:disable_scale_checks could be got rid of, and let any scale be set.

No, it can't be. If you can't cleanly divide a scale, you will get artifacts, period. At least with the current fractional scaling methods in wayland.

Dregu:
You probably meant different kind of rounding, but I'll just add that when using the actual `rounding` feature, it makes giant holes in subsurface corners in apps like obs. Probably the same [function](https://github.com/hyprwm/Hyprland/blob/7ce451d20c33025d4aaaf55444150c4a875b83ea/src/render/pass/SurfacePassElement.cpp#L112) that has been marked broken for a long time anyway...

<img width="1106" height="128" alt="Image" src="https://github.com/user-attachments/assets/564e5076-979d-4252-b911-b2a462c5e198" />

tannerellen:
[...]

PR #13124, opened 2026-01-29
pass: fix surface opaque region bounds used in occluding

jmanc3:
fixes: https://github.com/hyprwm/Hyprland/issues/11590

pesky little bug

### DIFF
--- a/src/render/pass/Pass.cpp
+++ b/src/render/pass/Pass.cpp
@@ -60,7 +60,7 @@ void CRenderPass::simplify() {
                 opaque = opaque.getExtents().scale(g_pHyprOpenGL->m_renderData.pMonitor->m_scale).round();
             else {
                 CRegion scaledRegion;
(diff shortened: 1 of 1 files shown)


jmanc3:
Note: I don't know if Hyprland creates any opaque regions with more than one rect, but if it did, those are left using the old bugged method.

does forEachRect allow modification of the underlying rect? if not I guess you'd have to make a new CRegion and add the rects using the forEachRect function or something? idk if that's even a case worth attending, might be

vaxerski:
> does forEachRect allow modification of the underlying rect? if not I guess you'd have to make a new CRegion and add the rects using the forEachRect function or something? idk if that's even a case worth attending, might be

No, and I'd do that. Fast path for 1 rect, but do the new CRegion thing with more.

jmanc3:
I made the change. As far as I could tell from the pixman docs, x2 and y2 are always greater than x1 and y1, and therefore the math I wrote seems right

PR #13140, opened 2026-01-30
layout: fix window position snap before drag_threshold reached

cmihail:

  Fixes a regression introduced in v0.53.0 where windows would immediately snap to the cursor
  position on mouse button press, even before the `drag_threshold` was exceeded. This caused both
  `bindm` (drag) and `bindc` (click) actions to trigger simultaneously when using the same button.

  **Root Cause:**
  The issue was in `IHyprLayout::updateDragWindow()` where window position was unconditionally set to
   the mouse cursor location before checking if `m_dragThresholdReached` was true.

  **Changes:**
  - Guard fullscreen floating window repositioning (line 1023) with `m_dragThresholdReached` check
  - Move tiled-to-floating window repositioning (line 1029) inside the existing
  `m_dragThresholdReached` conditional block

  **Testing:**
  Tested with configuration:
  ```conf
  binds {
      drag_threshold = 50
  }

  bindm = , mouse:274, movewindow
  bindc = , mouse:274, workspace, r+1
```

  Before fix: Window snaps to cursor immediately on button press, both actions trigger.
  After fix: Window only moves after dragging past 50 pixels, click action fires if released within
  threshold.

  Resolves: https://github.com/hyprwm/Hyprland/discussions/12796

  Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking
  compatibility, etc.)

  The fix only affects behavior when drag_threshold > 0. When drag_threshold = 0 (disabled), behavior
   remains unchanged as m_dragThresholdReached is set to true in onBeginDragWindow().

  Ready for merging. The fix is minimal, well-tested, and restores the intended behavior of
  drag_threshold without breaking compatibility.

### DIFF
--- a/src/layout/supplementary/DragController.cpp
+++ b/src/layout/supplementary/DragController.cpp
@@ -58,15 +58,15 @@ bool CDragStateController::updateDragWindow() {
     m_draggingTiled                   = false;
     m_draggingWindowOriginalFloatSize = DRAGGINGTARGET->lastFloatingSize();
 
-    if (WAS_FULLSCREEN && DRAGGINGTARGET->floating()) {
+    if (WAS_FULLSCREEN && DRAGGINGTARGET->floating() && m_dragThresholdReached) {
         const auto MOUSECOORDS = g_pInputManager->getMouseCoordsInternal();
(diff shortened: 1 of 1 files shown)


vaxerski:
has to wait for #12890

cmihail:
> has to wait for #12890

I merged master so ready for review again. It also works fine on my machine after rebuild

PR #13144, opened 2026-01-30
Fix edge grab resize logic for gaps_out > 0

ekhadley:
Fix mouse edge grab detection for tiled windows when resize_on_border is true and gaps_out > 0.


SGumbles:
This is causing a race condition depending on the order of the windows
in m_windows. If you expand hit boxes of the windows in a tiling compositor, there are going to be areas where they overlap, and the first window to do this hit test will be returned. I've attached a video show the issue :

https://github.com/user-attachments/assets/4ae6cb81-088c-4b29-a218-376d22f629b9

The big picture is that the hit boxes shouldn't bleed into adjacent windows. @ekhadley What was the original issue this was trying to fix? We could just expand the hit box of windows on the edges that are on a gap_out edge.

PR #13155, opened 2026-01-30
Allow tearing with DS with invisible cursors

UjinT34:
Adds cursors visibility checks to SW/HW cursor conditions.
Allows tearing with DS when cursors is invisible.

Tearing won't disable HW cursors automatically. If tearing activates while hw cursor is invisible it'll likely fail to become visible as HW and will fallback to SW. Besides that it's expected that users switch sw/hw cursors themselves. Most games that are usually used with tearing have invisible cursors anyway.

~~VRRtest crashes HL if closed when tearing and DS are active simultaneously. This crash doesn't leave a crash report or any useful messages in logs. No other apps are known to do that.~~

Ready

- [x] Investigate VRRtest crash

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1857,7 +1857,7 @@ bool CMonitor::attemptDirectScanout() {
         }
 
         //#TODO this entire bit is bootleg deluxe, above bit is to not make vrr go down the drain, returning early here means fifo gets forever locked.
-        if (PSURFACE->m_fifo && *PSAMEFIFO)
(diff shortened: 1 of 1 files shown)


Gliczy:
Seem to work great!

I tested with CS2 (linux native + wayland), Hytale (linux native + wayland) and Ultrakill (proton + xwayland), all tested visually and by checking the output of `hyprctl monitors`.

One thing to keep in mind, tearing won't work when using wine/proton with the wayland drv, `hyprctl monitors` will output `tearingBlockedBy: next frame is not torn`, but this is unrelated to this PR.

PR #13156, opened 2026-01-30
xwl: get supported props on constructing surface

gulafaran:
not all clients supports WM_DELETE_WINDOW like glxgears, so get supported props in constuctor of surface, and if not supported forcefully kill the client.

fixes killactive on glxgears and other xwl clients that does not support WM_DELETE_WINDOW


gulafaran:
if we really want to, we could i guess update this list on handlePropertyNotify to catch any changed props later on. but not sure how many clients keep on changing these props "dynamicly"

vaxerski (approved):
thanks

PR #13169, opened 2026-02-01
xdg-popup: account for popup window geometry when positioning

hjanuschka:
Fixes popup placement when a client sets a non-zero (including negative) xdg_surface.set_window_geometry() for popups.

Hyprland currently computes popup position in window-geometry space but uses it as surface-origin coordinates when sending xdg_popup.configure(). This can shift popups (e.g. Chromium password save bubble in --app mode) outside the parent window and break interaction.

This change converts the computed position back to surface-origin coordinates by subtracting the popup surface's pending window-geometry offset.

Related: basecamp/omarchy#4349

### DIFF
--- a/src/protocols/XDGShell.cpp
+++ b/src/protocols/XDGShell.cpp
@@ -75,11 +75,29 @@ CXDGPopupResource::~CXDGPopupResource() {
 }
 
 void CXDGPopupResource::applyPositioning(const CBox& box, const Vector2D& t1coord) {
-    CBox constraint = box.copy().translate(m_surface->m_pending.geometry.pos());
+    // Per xdg-shell, xdg_positioner coordinates are specified in the coordinate
+    // space of the parent's *window geometry*.
+    //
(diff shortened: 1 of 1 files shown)


hjanuschka:
Update: further investigation shows Chromium's password/save prompt in  mode is implemented as a  (WaylandBubble) and it sends negative  offsets (e.g. x=-298 y=-4) in app mode. That seems to be the primary cause of the misplacement, rather than xdg_popup positioning. This PR likely doesn't address the reported Chromium  prompt issue; feel free to close if deemed unrelated. Gist w/ logs: https://gist.github.com/hjanuschka/5143963e0eed86ff1567e910433802f6

hjanuschka:
Correction: Chromium password/save prompt in "--app" mode is implemented as a wl_subsurface (WaylandBubble). In app mode it sends negative wl_subsurface.set_position() offsets (e.g. x=-298 y=-4). Gist: https://gist.github.com/hjanuschka/5143963e0eed86ff1567e910433802f6

PR #13174, opened 2026-02-02
Commit and presentation timing fixes

UjinT34:
Moves commit timing state and timer into `SSurfaceState` so that each pending state can have it's own timer.
Fixes `toTimespec` init similar to `fromTimespec`
Changes presentation time api to avoid unnecessary back and forth timestamp conversion
Fixes presentation flags for DS and HW clock
Sends clock id event as required by the proto
Adds `render:commit_timing_enabled` setting to disable the proto in case of regressions.

Timers created by the commit timing proto are expected to be removed by `CEventLoopManager::nudgeTimers`.

Version on main is effectively disabled so it'll be the first real test of the proto. Might cause slowdowns on weaker hw since the commits are delayed till the given time and leave no extra time for internal rendering unless the client itself adjusts the target time. Tested clients doesn't seem to use the proto themselves or don't use it to limit fps, their target time leaves enough time for internal rendering.

Ready

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1577,7 +1577,7 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .value       = "render:commit_timing_enabled",
         .description = "Enable commit timing proto. Requires restart",
         .type        = CONFIG_OPTION_BOOL,
-        .data        = SConfigOptionDescription::SBoolData{false},
+        .data        = SConfigOptionDescription::SBoolData{true},
     },
 
     /*


src/helpers/Monitor.cpp
```diff
@@ -115,9 +115,10 @@ void CMonitor::onConnect(bool noRule) {
         }
 
         if (!ts)
-            PROTO::presentation->onPresented(m_self.lock(), Time::steadyNow(), event.refresh, event.seq, event.flags);
+            PROTO::presentation->onPresented(m_self.lock(), Time::toTimespec(Time::steadyNow()), event.refresh, event.seq,
+                                             event.flags & ~Aquamarine::IOutput::AQ_OUTPUT_PRESENT_HW_CLOCK);
         else
-            PROTO::presentation->onPresented(m_self.lock(), Time::fromTimespec(event.when), event.refresh, event.seq, event.flags);
+            PROTO::presentation->onPresented(m_self.lock(), *ts, event.refresh, event.seq, event.flags);
```
gulafaran: uh if ts is nullptr and now you *ts thats gonna eat penguins

gulafaran:
some added info fetched from libdrm etc. we can get the clock used from drm if we want to but on linux its rather pointless. so timestamps from our pageflips are CLOCK_MONOTONIC

```
/**
 * DRM_CAP_TIMESTAMP_MONOTONIC
 *
 * If set to 0, the kernel will report timestamps with ``CLOCK_REALTIME`` in
 * struct drm_event_vblank. If set to 1, the kernel will report timestamps with
 * ``CLOCK_MONOTONIC``. See ``clock_gettime(2)`` for the definition of these
 * clocks.
 *
 * Starting from kernel version 2.6.39, the default value for this capability
 * is 1. Starting kernel version 4.15, this capability is always set to 1.
 */

```

and making some diff against realtime means we are affected by ntp changes and drifting in our presented timestamps.

while the presentation proto also mentions this.

```
The timestamp corresponds to the time when the content update turned into light the first time on the surface's main output. Compositors may approximate this from the framebuffer flip completion events from the system, and the latency of the physical display path if known.
```



so yes this certainly could be made simpler. unless bsd differs here, if it does its just a few lines to AQ to get cap on init and if else case it in the calcs here

Honkazel:
FreeBSD uses DRM from Linux, so this shouldn't differ on FreeBSD
Even on FreeBSD 14 it uses DRM from Linux > 4.15
If you're still not sure - ask freebsd guy tagattie, maybe he knows the answer

src/config/ConfigDescriptions.hpp
```diff
@@ -1573,6 +1573,12 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .type        = CONFIG_OPTION_CHOICE,
         .data        = SConfigOptionDescription::SChoiceData{0, "default,gamma22,gamma22force,srgb"},
     },
+    SConfigOptionDescription{
+        .value       = "render:commit_timing_enabled",
+        .description = "Enable commit timing proto. Requires restart",
+        .type        = CONFIG_OPTION_BOOL,
+        .data        = SConfigOptionDescription::SBoolData{false},
```
vaxerski: uhhh false here and true in the config?

Arisa-Snowbell:
I tested this PR and I see no breakages, everything seems smooth.

frebib:
Am also running this and everything seems well/as normal

PR #13179, opened 2026-02-02
compositor: cleanup fading out on expired monitor

gulafaran:
hotplugging monitor can leave fading out windows forever stuck in the containers, check if they have a expired monitor and remove them aswell.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1288,9 +1288,17 @@ void CCompositor::changeWindowZOrder(PHLWINDOW pWindow, bool top) {
 }
 
 void CCompositor::cleanupFadingOut(const MONITORID& monid) {
+    bool windowsFadingDirty = false;
+
     for (auto const& ww : m_windowsFadingOut) {
 
         auto w = ww.lock();
+
+        if (!w || !w->m_monitor) {
+            windowsFadingDirty = true;
+            Log::logger->log(Log::DEBUG, "Cleanup: m_windowsFadingOut had expired window or monitor");
+            continue;
+        }
 
         if (w->monitorID() != monid && w->m_monitor)
             continue;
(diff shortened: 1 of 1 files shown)


src/Compositor.cpp
```diff
@@ -1313,14 +1319,18 @@ void CCompositor::cleanupFadingOut(const MONITORID& monid) {
         }
     }
 
+    if (windowsFadingDirty)
+        std::erase_if(m_windowsFadingOut, [](const auto& el) { return el.expired() || !el->m_monitor; });
```
vaxerski, line 1323: nope, do not remove a window just because it has no monitor. It could still be alive but lost for some reason. If it's destroyed AND no monitor, sure.

src/Compositor.cpp
```diff
@@ -1313,14 +1319,18 @@ void CCompositor::cleanupFadingOut(const MONITORID& monid) {
...
+
     bool layersDirty = false;
 
     for (auto const& lsr : m_surfacesFadingOut) {
 
         auto ls = lsr.lock();
 
-        if (!ls) {
+        if (!ls || !ls->m_monitor) {
```
vaxerski, line 1331: same here

src/Compositor.cpp
gulafaran: problem doing && here would mean the `el` might be expired but monitor might not

vaxerski: could be yes, but in general you cant remove a window until it's marked as ready for destrtoy

gulafaran: well yeah but some of these containers get stuck with with some stuff when i hotplug the monitor. and && might segfault, el.expired && el->m_monitor will nullptr deref. but i might be going about this the wrong way then, perhaps m_layers or m_windows contains something that it shouldnt so it never expires on hotplugging the monitor hm

vaxerski: wdym by stuck, whats their state

gulafaran:
not relevant anymore with all the refactors

PR #13183, opened 2026-02-02
opengl: add debug:gl_debugging

gulafaran:
add debug:gl_debugging so we can disable gl debugging entirerly, both glGetError and enabling EGL_KHR_debug has its cost, we still have EXT_create_context_robustness and glGetGraphicsResetStatus that should catch context loss, and is generally cheap to call it only checks a flag set.

glGetError might cause a implicit flush to get any pending calls sent to the gpu.

however to get EGL_KHR_debug back enabled we now require a restart of the compositor after changing debug:disable_gl_debugging


vaxerski:
can you force push to re-run ci? Also, maybe just flip the var to be `gl_debugging`? would be more intuitive

PR #13189, opened 2026-02-04
hyprpm: add full nix integration

vaxerski:
Adds nix integration to hyprpm: hyprpm will now detect nix'd hyprland and use nix develop instead

cc @fufexan can you take a look at this?

Also, with nix develop we could probably skip the header pull eventually but that's down the road. This should work.

I've tested this myself on `nix profile add 'git+ssh://git@github.com/hyprwm/hyprland.git#hyprland' --refresh` but additional testing may be nice. Can't test hyprnix because it's broken but I assume it's gonna work the same once 0.54 drops.


spikespaz:
What is the purpose of using hyprpm when you can just list the plugin nix packages in config

fufexan:
> What is the purpose of using hyprpm when you can just list the plugin nix packages in config

To make it easier for people using hyprnix and not wanting/having the time to get their hands too dirty with nix.

PR #13197, opened 2026-02-05
Allow matching monitor rules by mode

kevinmcconnell:

Adds a "mode:" matcher prefix to monitor rules, in a similar style as the existing "desc:" matcher prefix. A rule using "mode:" will match a monitor that supports that mode.

Modes are matched as `starts_with` so they can include or omit the refresh rate portion as necessary.

An example rule could be:

    monitor=mode:3840x2160,preferred,auto,2

This feature was prompted by trying to get Apple monitors (Studio Display & Pro XDR) to work over Thunderbolt. For some broken MST reason, both of those devices actually report themselves as two separate monitors, but with identical serial, description etc. They show up as both a) working monitor with all the expected modes, and b) broken one with only a few low-res modes. If the broken one is left enabled, it results in a sort of ghost workspace you can never see.

Since the ports can change between sessions, and the serial/desc/etc matches, we can't differentiate between them in the config. But if we can target the modes, we can do something like this:

    monitor=desc:Apple Computer Inc StudioDisplay,disable
    monitor=mode:5120x2880,preferred,auto,2

I've been running this patch for a few days and it's been stable for me. I also added a test for the new behaviour. I think it's good to go, but happy to take feedback.

### DIFF
--- a/hyprtester/src/tests/main/monitors.cpp
+++ b/hyprtester/src/tests/main/monitors.cpp
@@ -1,0 +1,53 @@
+#include "tests.hpp"
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
+#include "../shared.hpp"
+
+static int ret = 0;
+
+static bool test() {
+    NLog::log("{}Testing monitor mode: and desc: selectors", Colors::GREEN);
+
+    std::string monitorsSpec = getFromSocket("j/monitors");
+    EXPECT_CONTAINS(monitorsSpec, "HEADLESS-2");
+
+    NLog::log("{}Testing mode: selector (matching)", Colors::YELLOW);
+
+    OK(getFromSocket("/reload"));
+    OK(getFromSocket("/keyword monitor ,preferred,auto,2"));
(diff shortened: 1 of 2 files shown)


vaxerski:
flaky solution but we can go with it if there isn't anything better. Could you attach the output of `drm_info` with one of those displays?

kevinmcconnell:
@vaxerski there wasn't anything better that I could see at the Hyprland level, but that's a good point: at the DRM level I think we can detect this condition and work around it. If we parse the tiling information we can figure out which connector needs to be disabled.

I've tried another approach to try that, and it also seems to work reliably on my hardware. What do you think of this idea instead: https://github.com/hyprwm/aquamarine/pull/238

(Attached a full `drm_info` output on that other PR too)

vaxerski:
yup the aq mr looks much more sensible and correct, fix the root cause rather than the effect

kevinmcconnell:
Agreed, yes :+1:  With https://github.com/hyprwm/aquamarine/pull/238 this isn't needed.

PR #13211, opened 2026-02-07
feat: implement ext-background-effect-v1 protocol

aurelleb:

Implements the [ext-background-effect-v1](https://wayland.app/protocols/ext-background-effect-v1) protocol.

I tried to implement this in a straightforward way with respect to how other protocols are integrated. I think I have done a pretty okay job at that.

A few things to note regarding the way this protocol work in Hyprland:
- in Hyprland, regular windows have their entire surface blurred by default unless the user disables blur in their config. A client that sets the blur region of a surface using this protocol will override that default and disable blur for everything outside the blur region. As specified by the protocol, passing a null region will also disable blur completely. Therefore, a client can use this protocol to opt-out of blur for a given surface.
- my implementation prioritizes the protocol over existing window and layer rules, which is an arbitrary decision that we may need to discuss.

I added a few test clients (disclosure - generated by Opus 4.6) in a separate commit in order to have examples of partial/dynamic blurring and such. I'm not sure we want to keep that in the tree but I found them very helpful to test the protocol out. I saw about hyprtester but headless testing only helps so much in this case.

As demonstrated by the test clients everything seem to be working. I also tested this with my launcher [vicinae](https://github.com/vicinaehq/vicinae/pull/1066) which is getting support for this protocol.


aurelleb:
PS: The CI failing is because of the test clients. But I guess we will remove those (we can just drop the commit using a simple rebase)

CMakeLists.txt
```diff
@@ -551,6 +551,7 @@ protocolnew("staging/ext-data-control" "ext-data-control-v1" false)
 protocolnew("staging/pointer-warp" "pointer-warp-v1" false)
 protocolnew("staging/fifo" "fifo-v1" false)
 protocolnew("staging/commit-timing" "commit-timing-v1" false)
+protocolnew("protocols" "ext-background-effect-v1" true)
```
Honkazel: > protocolnew("staging/ext-background-effect" "ext-background-effect-v1" false)

Do this instead and un-vendorize this proto(remove xml)
You don't need to bump required wayland-protocols version, it's already 1.45 in CMake

Honkazel (commented):
I don't think, that vax would approve those test clients... But better to wait, than guess

vaxerski:
> I don't think, that vax would approve those test clients... But better to wait, than guess

correct, why are they in C + wayland-scanner?

aurelleb:
> > I don't think, that vax would approve those test clients... But better to wait, than guess
> 
> correct, why are they in C + wayland-scanner?

 I can just drop them, but for the review I thought it could have been useful to you in order to verify the implementation. They are in C + wayland scanner just to keep the examples as simple as possible.

vaxerski:
C + wayland-scanner makes them _less_ simple

aurelleb:
More simple in the sense that they don't depend on any hypr* stuff so we can isolate the testing of that specific piece of functionality from hypr stuff.  Clients are not going to use hyprscanner. But whatever, I don't really mind. So what do we do? Rewrite it? Drop it completely? The important part is the protocol I don't really care about the examples

vaxerski:
you can drop the examples

aurelleb:
all done

src/render/Renderer.cpp
```diff
@@ -2710,16 +2710,30 @@ bool CHyprRenderer::shouldBlur(PHLLS ls) {
...
-    static auto PBLUR     = CConfigValue<Hyprlang::INT>("decoration:blur:enabled");
-    const bool  DONT_BLUR = w->m_ruleApplicator->noBlur().valueOrDefault() || w->m_ruleApplicator->RGBX().valueOrDefault() || w->opaque();
-    return *PBLUR && !DONT_BLUR;
+    static auto PBLUR = CConfigValue<Hyprlang::INT>("decoration:blur:enabled");
+    if (!*PBLUR)
+        return false;
+
+    auto surface = w->wlSurface();
+    if (surface && surface->m_hasBackgroundEffect)
+        return !surface->m_blurRegion.empty();
```
vaxerski, line 3349: wrong: we should still honor noBlur() AND RGBX (rgbx makes the blur invisible, pointless)

Also take into account opaque, if for some god-forsaken reason a dumb app did opaque + blur, ignore blru

ashley-paul-12345:
Hi, any update on this one?

aurelleb:
Hey, sorry I have been super busy, I should be able to fix these last issues this week end or so

DekoDX:
Is this ready to merge?

aurelleb:
I have some remaining cleanup to do but it's getting close

aurelleb:
Sorry for the wait, everything should now be in order, it's ready to merge. Tested on latest Hyprland today.

By the way, https://github.com/vicinaehq/vicinae is a great client to test this with, because it makes heavy use of regional blur to ignore the client side drawn drop shadows and go from compact mode to full window mode.

vaxerski (approved):

fxzzi:
blur on foot is just... missing now? kinda
https://codeberg.org/dnkl/foot/commit/dea10e2e48162c6ab48ca564ceeca88dea886a04
i enabled the blur option in foot, seems to be missing and only works on a specific area of my monitor
disabling the blur option in foot makes the blur not work at all which is expected

https://github.com/user-attachments/assets/4d5d61a0-a5fb-46da-b2a6-a18d170fe203

vaxerski:
can you post WAYLAND_DEBUG=1 of foot?

fxzzi:
> can you post WAYLAND_DEBUG=1 of foot?

sure 
[log.txt](https://github.com/user-attachments/files/27900696/log.txt)

so the blur only works if its in fullscreen or at the top left quarter of my monitor

aurelleb:
I have a fix ready

PR #13234, opened 2026-02-10
Add: error message to hyprctl hyprpaper wallpaper

ItsOhen:
Came to attention that code 1 and code 2 isn't very informative when getting an error.

Nope

Yes


ItsOhen:
Somewhat unlikely to hit the path error, since we do:
https://github.com/hyprwm/Hyprland/blob/main/hyprctl/src/hyprpaper/Hyprpaper.cpp#L31

But it is what it is i guess.

PR #13252, opened 2026-02-11
animations: add spring curve + spring-physics mode

colonelpanic8:
## Summary
- add closed-form damped spring support for animation curves
- keep `spring*` as normalized spring easing over existing animation duration semantics
- add explicit `spring-physics*` mode with emergent completion (settling-time based), so spring duration is driven by physics instead of fixed `[0,1]` progress
- cache spring curve-name parsing to avoid repeated string parsing in per-frame paths
- route curve evaluation through `CHyprAnimationManager` so special workspace fade uses the same curve behavior

## Spring Curve Names
- Normalized mode:
  - `spring`
  - `spring-snappy`
  - `spring-bouncy`
  - `spring:<damping_ratio>:<angular_frequency>`
- Physics mode:
  - `spring-physics`
  - `spring-physics-snappy`
  - `spring-physics-bouncy`
  - `spring-physics:<damping_ratio>:<angular_frequency>`

## Notes
- normalized `spring*` preserves existing Hyprland animation model (duration from config speed, curve shapes the interpolation)
- `spring-physics*` is opt-in and settles based on spring parameters

## Testing
- `nix develop -c bash -lc 'cmake -S . -B build-nix && cmake --build build-nix -j$(nproc)'`

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -2433,7 +2433,7 @@ std::optional<std::string> CConfigManager::handleAnimation(const std::string& co
     std::string bezierName = ARGS[3];
     m_animationTree.setConfigForNode(ANIMNAME, enabledInt, speed, ARGS[3], ARGS[4]);
 
-    if (!g_pAnimationManager->bezierExists(bezierName)) {
+    if (!g_pAnimationManager->isBezierNameValid(bezierName)) {
         const auto PANIMNODE      = m_animationTree.getConfig(ANIMNAME);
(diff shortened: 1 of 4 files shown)


vaxerski:
wouldnt this work better in hyprutils

colonelpanic8:
> wouldnt this work better in hyprutils

not quite sure I understand why. i don't see how its possible to change the animation curves from there.

vaxerski:
well springs dont operate on curves now do they? You just need to implement an additional setting I guess for avar type (curve / spring) and extend the avar class

PR #13262, opened 2026-02-13
cm: block DS for scRGB in HDR mode

njdom24:

Conditions for enabling DS with `cm=hdr{edid}` only check that the source and destination are both HDR, but doesn't make a distinction for scRGB content, which we can't scan out yet. This leads to blown-out colors. Quick change to add the check.

Feels a little bad having two `surfaceIsScRGB` checks in there, but removing the first one would give bad output for `render:non_shader_cm = 3` (the default). Though, judging by its description (`disable and ignore CM issues`), maybe it should?

Ready

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1819,7 +1819,7 @@ uint16_t CMonitor::isDSBlocked(bool full) {
     const bool surfaceIsScRGB = surfaceIsHDR && PSURFACE->m_colorManagement->isWindowsScRGB();
 
     if (needsCM() && (*PNONSHADER != CM_NS_IGNORE || surfaceIsScRGB) && !canNoShaderCM() &&
-        ((inHDR() && (*PPASS == 0 || !surfaceIsHDR)) || (!inHDR() && (*PPASS != 1 || surfaceIsHDR))))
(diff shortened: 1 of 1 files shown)


vaxerski (approved):

fxzzi:
image
seems to claim ds is blocked by colour management, but the colours are still blown out weirdly enough. As expected `hyprctl keyword render:direct_scanout 0` and toggling hdr off and on again fixes it.

Also weirdly enough my monitor modesets when going from the scrgb mode to the pq mode, but not when going the other way

scrgb mode is entirely broken after vax's icc pr for me anyway, so i tested on the PR just before that to confirm 8271cfc

njdom24:
@fxzzi Does it work if you un-fullscreen the game, switch to scRGB, _then_ fullscreen? I noticed a new weird CM behavior with switching between modes not working while an app is fullscreened, regardless of DS state.

fxzzi:
> @fxzzi Does it work if you un-fullscreen the game, switch to scRGB, _then_ fullscreen? I noticed a new weird CM behavior with switching between modes not working while an app is fullscreened, regardless of DS state.

yeah, looks fine. I was able to repro the oversaturation just by running `hyprctl reload` with DS on. The scrgb mode being broken is fixed in main other than this.

PR #13271, opened 2026-02-14
Fix use-after-free in CommitTiming timer callback

Kam1k4dze:
My Vulkan app was crashing Hyprland on window resize. Capturing `m_surface` by value in the timer lambda stops the callback from dereferencing a destroyed `CCommitTimerResource`

[backtrace.txt](https://github.com/user-attachments/files/25312404/backtrace.txt)


ssareta:
Looks very similar to #13270, probably fixes that too. Will test later

ssareta:
Tested, does indeed fix #13270, thanks!

PR #13229, opened 2026-02-09
directscanout: fix dgpu directscanout explicit sync

gulafaran:
without setting an infence, AQ doesnt explicit sync, nor recreate the dgpu fence for the blit work. and as such attemptdirectscanout path artifacts and breaks. create a dummy CEGLSync even tho we dont really have any pending glwork to get a proper fence, and set it.

PR #13277, opened 2026-02-14
multigpu: fix multi gpu checking

gulafaran:
drmFD() from allocators is not always equal, because we reopen them inside AQ for refcounting, meaning they get duplicated and become their own fds, so checking if fd1 == fd2 ends up wrong.

introduce sameGpu in MiscFunctions that checks the actual drmDevice meaning we can now even check if a rendernode is the same gpu as a display node if we want.

fixes blitting and multigpu paths that was hit on single gpu situations.

found by @Arisa-Snowbell see more at https://github.com/hyprwm/Hyprland/pull/13229

### DIFF
--- a/src/helpers/Drm.hpp
+++ b/src/helpers/Drm.hpp
@@ -1,0 +1,5 @@
+#pragma once
+
+namespace DRM {
+    bool sameGpu(int fd1, int fd2);
+}
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -21,7 +21,6 @@
 #include <iomanip>
 #include <sstream>
 #include <fstream>
-#include <xf86drm.h>
 #ifdef HAS_EXECINFO
 #include <execinfo.h>
 #endif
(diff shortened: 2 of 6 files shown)


Arisa-Snowbell:
Of course, tested it, works, fixed my performance issues

src/helpers/MiscFunctions.cpp
```diff
@@ -981,3 +982,21 @@ bool truthy(const std::string& str) {
 
     return cpy.starts_with("true") || cpy.starts_with("yes") || cpy.starts_with("on");
 }
+
+bool sameGpu(int fd1, int fd2) {
```
vaxerski: don't write new MiscFunctions, I need to get rid of them. Just add a helpers/drm.hpp and do something like DRM::sameGpu()

gulafaran: moved

src/helpers/Drm.hpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+namespace NDrm {
```
vaxerski: I still haven't removed that `N.*` regex in clang tidy? name it `DRM` plx

gulafaran: renamed

PR #13284, opened 2026-02-16
pointermgr: damage only the surface size

gulafaran:
CWaylandOutput returns a vector2d with -1, -1 set as a "no limit", passing that down into beginSimple and the renderer it hits pixman bug of invalid sizes and wrong rectangles gets created.

causing bunch of
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed

just set the damage to the size we render the texture down below instead and we not only solve pixman issue but we dont overdamage either.

### DIFF
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -397,31 +397,27 @@ bool CPointerManager::setHWCursorBuffer(SP<SMonitorPointerState> state, SP<Aquam
 }
 
 SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager::SMonitorPointerState> state, SP<CTexture> texture) {
-    auto        maxSize    = state->monitor->m_output->cursorPlaneSize();
-    auto        damage     = maxSize;
-    auto const& cursorSize = m_currentCursorImage.size;
-
-    static auto PCPUBUFFER = CConfigValue<Hyprlang::INT>("cursor:use_cpu_buffer");
-
-    const bool  shouldUseCpuBuffer = *PCPUBUFFER == 1 || (*PCPUBUFFER != 0 && g_pHyprRenderer->isNvidia());
+    auto maxSize = state->monitor->m_output->cursorPlaneSize();
 
     if (maxSize == Vector2D{})
         return nullptr;
     else if (maxSize == Vector2D{-1, -1}) {
-        damage = Vector2D{256, 256};
         Log::logger->log(Log::TRACE, "cursor plane size is unlimited, falling back to 256x256");
-        if (cursorSize.x > damage.x || cursorSize.y > damage.y) {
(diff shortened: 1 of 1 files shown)


src/managers/PointerManager.cpp
```diff
@@ -579,11 +579,11 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
 
     RBO->bind();
 
-    const auto& damageSize = state->monitor->m_output->cursorPlaneSize();
-    g_pHyprOpenGL->beginSimple(state->monitor.lock(), {0, 0, damageSize.x, damageSize.y}, RBO);
+    auto damage = Vector2D{m_currentCursorImage.size / m_currentCursorImage.scale * state->monitor->m_scale}.round();
+    g_pHyprOpenGL->beginSimple(state->monitor.lock(), {0, 0, damage.x, damage.y}, RBO);
```
vaxerski, line 586: well no we still render the entire plane

gulafaran: needs something done when CWaylandOutput then, just assume its a 256x256 then?

vaxerski: sure, we can fall back to 256x. Just make sure to add a helper for this so that the software path knows when to fire when the cursor is too big.

gulafaran: something like [faf6d56](https://github.com/hyprwm/Hyprland/pull/13284/commits/faf6d56f0216c61c8f2095260f255f1035ca0366) then

src/managers/PointerManager.cpp
```diff
@@ -406,14 +407,21 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
...
     if (maxSize == Vector2D{})
         return nullptr;
-
-    if (maxSize != Vector2D{-1, -1}) {
-        if (cursorSize.x > maxSize.x || cursorSize.y > maxSize.y) {
-            Log::logger->log(Log::TRACE, "hardware cursor too big! {} > {}", m_currentCursorImage.size, maxSize);
+    else if (maxSize == Vector2D{-1, -1}) {
+        damage = Vector2D{256, 256};
+        Log::logger->log(Log::TRACE, "cursor plane size is unlimited, falling back to 256x256");
+        if (cursorSize.x > damage.x || cursorSize.y > damage.y) {
```
vaxerski: can be de-dupd both if and else contain this

gulafaran: de-dupd

vaxerski (approved):

fxzzi:
You has broken my system >:(((

PXL_20260221_231013420 MP

My monitor is stetched to the top left, cursor is invisible, and whichever monitor should have the cursor on it is frozen

ItsOhen:
Same.
And no monitor redraws at all. Having to dip to tty2 and back to get it to even show the prompt.

alper-han:
There is a white rectangle under my cursor (hardware cursor)

PR #13285, opened 2026-02-16
Fix use_cpu_buffer

Murat65536:

SUPER small change. `use_cpu_buffer` should be of type int since it could be 0, 1, or 2.

There isn't anything else I see that has the wrong type at a quick glance but I could be wrong.

Ready

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -1702,7 +1702,7 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     },
     SConfigOptionDescription{
         .value       = "cursor:use_cpu_buffer",
-        .description = "Makes HW cursors use a CPU buffer. Required on Nvidia to have HW cursors. Experimental",
+        .description = "Makes HW cursors use a CPU buffer. Required on Nvidia to have HW cursors. 0 - off, 1 - on, 2 - auto (nvidia only)",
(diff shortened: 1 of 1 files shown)


src/config/ConfigDescriptions.hpp
```diff
@@ -1703,7 +1703,7 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
     SConfigOptionDescription{
         .value       = "cursor:use_cpu_buffer",
         .description = "Makes HW cursors use a CPU buffer. Required on Nvidia to have HW cursors. Experimental",
-        .type        = CONFIG_OPTION_BOOL,
+        .type        = CONFIG_OPTION_INT,
         .data        = SConfigOptionDescription::SBoolData{false},
```
vaxerski: data shouldnt be bool then

Murat65536:
Another thing I'm seeing is that the description is different than in the wiki.
Wiki: "Makes HW cursors use a CPU buffer. Required on Nvidia to have HW cursors. 0 - off, 1 - on, 2 - auto (nvidia only)"
Code: "Makes HW cursors use a CPU buffer. Required on Nvidia to have HW cursors. Experimental"
IMO wiki is much more detailed. It actually tells you what the options mean. I don't think this is the only option like that. That being said, is that outside this PR's scope?

vaxerski:
you can c+p the wiki one too

Murat65536:
@vaxerski

PR #13313, opened 2026-02-20
Window: onX11ConfigureRequest reject requests that violate window rules

lichie567:
Some xwayland apps (especially games) request changes in geometry that are very annoying because they happen out-of-band from the window rule system. Not a problem with fullscreen games but I find most games work best in floating windowed mode.

This PR adds a `suppressevent x11configurerequest` rule to supress x11configurerequests from modifying the geometry of the window. Normally this only happens for floating, non-fullscreen windows. When the suppressevent rule is present, floating windows will follow the same code-path as tiled or fullscreen windows on x11configurerequests (i.e. ignoring them). 

I am no expert on xwayland and I do not know if some apps may act strange or crash in response to a rejected geometry request (seems like it should be fine since tiled windows ignore them).

Ready

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -1498,32 +1498,22 @@ void CWindow::onX11ConfigureRequest(CBox box) {
         return;
     }
 
-    if (!m_ruleApplicator->static_.center.has_value() && m_ruleApplicator->static_.position.empty()) {
-        m_realPosition->setValueAndWarp(xwaylandPositionToReal(box.pos()));
-    } else {
-        Log::logger->log(Log::DEBUG, "onX11ConfigureRequest: window '{}' ({:#x}) requested pos={:.0f},{:.0f} but was rejected due to window rule", m_title, (uintptr_t)this, box.x,
-                         box.y);
-    }
+    if (box.size() > Vector2D{1, 1})
+        setHidden(false);
+    else
+        setHidden(true);
 
-    if (m_ruleApplicator->static_.size.empty()) {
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
this won't work. You could disallow for like 1 second, but X11 really relies on these events for proper functioning

lichie567:
> this won't work. You could disallow for like 1 second, but X11 really relies on these events for proper functioning

Could you elaborate a bit more? The requests seem to be completely ignored for tiled windows, so why would it be a problem? After thinking about it more I do think its probably best to ditch the move/size/center rule checks and just add a supressevent rule so its more explicit that this behavior is desired.

Do you know of any specific cases where this feature would be problematic? The x11 apps that prompted me to implement this have been working fine with this change, but I'd like to test other apps if you know of any.

lichie567:
I looked more closely at what was happening in the tiled window case and realized I had missed a lot of logic that was taken care of in `sendWindowSize()`. I updated the PR to use a suppressevent rule instead and have the window use the same code path.

vaxerski:
popups / tooltips / splashes / etc would probably complain and enter weird states. It's been long since I touched them much though.

lichie567:
@vaxerski could you take another look at this? I've updated it to use an explicit suppressevent rule. I've been testing this in my own build for months now and it has worked flawlessly for the applications I use it for.

vaxerski (commented):
ok, wiki mr needed

lichie567:
here: https://github.com/hyprwm/hyprland-wiki/pull/1564

vaxerski (approved):

Issue #13079, opened 2026-01-24
Dynamically created layerrules are not applied


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/12984#discussioncomment-15532988

## Description

When a layerrule is runtime added (keyword, source, etc) it doesnt apply. At all.

Mohataseem89:
It looks like layerrules are only evaluated during the initial config load, and newly added rules don’t trigger a reevaluation of existing or future layer surfaces. This makes runtime configuration inconsistent compared to other rule types

The recent commit `fix: layerrules from hyprctl` seems like the right direction it might be worth double-checking that:
  - newly added layerrules are inserted into the active rule set (not just stored), and
  - a re-apply / refresh hook is triggered so layer surfaces get re-matched immediately

PR #13080, opened 2026-01-24
Fix #13079

ItsOhen:
#### Describe your PR, what does it fix/add?
Simply update rules when layerrules have been called.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Nope

#### Is it ready for merging, or does it need work?
Yup.

PR #13331, opened 2026-02-22
layerrules: add dynamically registered rules for plugins

VirtCode:
I wanted to add a layer rule to one of my plugins today, and found out that whilst you can register new rules, `m_otherProps` is only implemented for window rules. So this adds support dynamically registered layer rules, allowing plugins to add layer rules aswell. I basically copied over the implementation from the window rules as it works exactly the same. Additionally, I added a `layerUpdateRules` hook event for convenience, but more on that below.

I had a look at adding tests for this (because there is one for dynamically added window rules), but I found that we didn't have any involving layers yet so I didn't think it was worth adding layer tests just for this. As it is basically a copy from the window rules. Let me know if that is okay, otherwise we'd have to have a layer shell client in our tester, either a custom one or something like `wofi` or `wayneko`.

So I instead did some local testing with the following test plugin, if you want to try it out. Things work as expected.

<details>
<summary>minimal source code of the test plugin</summary>

```c++
#include <hyprgraphics/color/Color.hpp>
#include <hyprland/src/helpers/memory/Memory.hpp>
#include <hyprland/src/plugins/PluginAPI.hpp>
#include <hyprland/src/desktop/rule/layerRule/LayerRuleEffectContainer.hpp>
#include <hyprland/src/desktop/view/LayerSurface.hpp>
#include <hyprland/src/event/EventBus.hpp>

#include <string>
#include <unistd.h>

inline HANDLE PHANDLE = nullptr;

APICALL EXPORT PLUGIN_DESCRIPTION_INFO PLUGIN_INIT(HANDLE handle) {
    PHANDLE = handle;

    // check that header version aligns with running version
    const std::string CLIENT_HASH = __hyprland_api_get_client_hash();
    const std::string COMPOSITOR_HASH = __hyprland_api_get_hash();
    if (COMPOSITOR_HASH != CLIENT_HASH && false) {
        HyprlandAPI::addNotification(PHANDLE, "[test-plugin] Failed to load, mismatched versions! (see logs)", CHyprColor{1.0, 0.2, 0.2, 1.0}, 5000);
        throw std::runtime_error(std::format("version mismatch, built against: {}, running compositor: {}", CLIENT_HASH, COMPOSITOR_HASH));
    }

    static const auto RULE_ID = Desktop::Rule::layerEffects()->registerEffect("test_rule");

    static const auto RULE_CHANGE_LISTENER = Event::bus()->m_events.layer.updateRules.listen([&](PHLLS layer) {
        const auto& other = layer->m_ruleApplicator->m_otherProps.props;
        auto effect = other.contains(RULE_ID) ? other.at(RULE_ID)->effect : "<no-rule>";

        HyprlandAPI::addNotification(PHANDLE, std::format("[test-plugin] updated layer {} with effect {}", layer->m_namespace, effect), CHyprColor{0.2, 1.0, 0.2, 1.0}, 2000);
    });

    return {"plugin-test", "a quick example for custom layer rules", "Virt", "0.1"};
}

APICALL EXPORT void PLUGIN_EXIT() {
    Desktop::Rule::layerEffects()->unregisterEffect("test_rule");
}

// Do NOT change this function.
APICALL EXPORT std::string PLUGIN_API_VERSION() {
    return HYPRLAND_API_VERSION;
}
```

</details>

Also, as mentioned I added the new `layerUpdateRules` event akin to the `windowUpdateRules` hook event. Technically this doesn't expose anything new, as it will only fire on 1. when a layer is mapped or 2. when the config is reloaded, both of which already have hook events. However, I still think it is very convenient to have it as a separate hook event, and it would also be future-proof in case we ever have dynamic layer rules (matching on things other than namespace).

yes

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -313,7 +313,7 @@ static SDispatchResult checkLayerRule(std::string in) {
         return {.success = false, .error = "Layers under test not here"};
 
     for (const auto& layer : g_pCompositor->m_layers) {
-        if (layer->m_namespace == "rule-waybar") {
+        if (layer->m_namespace == "rule-layer") {
 
             if (!layer->m_ruleApplicator->m_otherProps.props.contains(layerRuleIDX))
                 return {.success = false, .error = "No rule"};
@@ -321,7 +321,7 @@ static SDispatchResult checkLayerRule(std::string in) {
             if (layer->m_ruleApplicator->m_otherProps.props[layerRuleIDX]->effect != "effect")
                 return {.success = false, .error = "Effect isn't \"effect\""};
 
-        } else if (layer->m_namespace == "norule-waybar") {
+        } else if (layer->m_namespace == "norule-layer") {
 
             if (layer->m_ruleApplicator->m_otherProps.props.contains(layerRuleIDX))
                 return {.success = false, .error = "Rule even though it shouldn't"};
--- a/hyprtester/src/tests/main/layer.cpp
+++ b/hyprtester/src/tests/main/layer.cpp
@@ -11,9 +11,9 @@ static int ret = 0;
 using namespace Hyprutils::OS;
 using namespace Hyprutils::Memory;
 
-static bool spawnWaybar(const std::string& namespace_) {
-    NLog::log("{}Spawning waybar layer {}", Colors::YELLOW, namespace_);
-    if (!Tests::spawnWaybar(namespace_)) {
+static bool spawnLayer(const std::string& namespace_) {
+    NLog::log("{}Spawning kitty layer {}", Colors::YELLOW, namespace_);
+    if (!Tests::spawnLayerKitty(namespace_)) {
         NLog::log("{}Error: {} layer did not spawn", Colors::RED, namespace_);
         return false;
     }
@@ -23,18 +23,18 @@ static bool spawnWaybar(const std::string& namespace_) {
 static bool test() {
     NLog::log("{}Testing plugin layerrules", Colors::GREEN);
 
-    if (!spawnWaybar("rule-waybar"))
+    if (!spawnLayer("rule-layer"))
         return false;
 
     OK(getFromSocket("/dispatch plugin:test:add_layer_rule"));
     OK(getFromSocket("/reload"));
 
-    OK(getFromSocket("/keyword layerrule match:namespace rule-waybar, plugin_rule effect"));
+    OK(getFromSocket("/keyword layerrule match:namespace rule-layer, plugin_rule effect"));
 
-    if (!spawnWaybar("rule-waybar"))
+    if (!spawnLayer("rule-layer"))
         return false;
 
-    if (!spawnWaybar("norule-waybar"))
+    if (!spawnLayer("norule-layer"))
         return false;
 
     OK(getFromSocket("/dispatch plugin:test:check_layer_rule"));
--- a/hyprtester/src/tests/shared.cpp
+++ b/hyprtester/src/tests/shared.cpp
@@ -40,27 +40,25 @@ CUniquePointer<CProcess> Tests::spawnKitty(const std::string& class_, const std:
     return kitty;
 }
 
-CUniquePointer<CProcess> Tests::spawnWaybar(const std::string& namespace_) {
-    std::vector<std::string> programArgs;
-
+CUniquePointer<CProcess> Tests::spawnLayerKitty(const std::string& namespace_, const std::vector<std::string> args) {
+    std::vector<std::string> programArgs = args;
     if (!namespace_.empty()) {
-        std::string config = std::format("/tmp/{}-waybar.jsonc", namespace_);
-        if (!writeFile(config, std::format(R"({{ "name": "{}", "modules-right": ["clock"] }})", namespace_)))
-            return nullptr;
-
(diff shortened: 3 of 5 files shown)


fxzzi:
(closes https://github.com/hyprwm/Hyprland/issues/13079)

VirtCode:
> (closes https://github.com/hyprwm/Hyprland/issues/13079)

that's something different which I think was already fixed by #13080 some time ago

fxzzi:
Mb ok well it should be closed anyway

VirtCode:
should be rebased onto the eventbus changes

vaxerski:
will this work?

VirtCode:
yes? works as expected when testing with the minimal plugin from above, is there something else I should check?

src/desktop/rule/layerRule/LayerRuleApplicator.cpp
```diff
@@ -32,11 +33,39 @@ void CLayerRuleApplicator::resetProps(std::underlying_type_t<eRuleProperty> prop
     UNSET(aboveLock)
     UNSET(ignoreAlpha)
     UNSET(animationStyle)
+
+#undef UNSET
+
+    if (prio == Types::PRIORITY_WINDOW_RULE) {
```
vaxerski: what are thooooooes {}

VirtCode: oooops

vaxerski:
> I had a look at adding tests for this (because there is one for dynamically added window rules), but I found that we didn't have any involving layers yet so I didn't think it was worth adding layer tests just for this. As it is basically a copy from the window rules. Let me know if that is okay, otherwise we'd have to have a layer shell client in our tester, either a custom one or something like wofi or wayneko.

wofi in the tester is fine. We already need kitty, xeyes, etc. Just add it to the nix vm

VirtCode:
okay so I added a test for it, let's see whether the CI thinks its good (half of the other tests fail on my host). Here are a couple of things to note:
- ~~I decided to use waybar instead because ts is also very widely (similar to wofi probably) used I suppose and as of the latest version you can set the namespace in the config. Also it has reserved areas so we might at some point test the `order` layer rule, reserved areas, and other stuff.~~
- ~~The tester currently generates a minimal config for each waybar to `/tmp` to set the namespace, because waybar doesn't take command line arguments to set that. I hope this is fine.~~
- I put the functions to spawn/kill layers also into `shared.cpp` so they are where the kitty functions are. I can also move them to the layer test file if you want but I think they might be useful for some other tests in the future too. 
- The `killAllLayers` function has a very criminal sleep in it, but I ran into a hyprland bug where killing many surfaces at once would end up with surfaces that hyprctl shows with `pid: -1`, and their process being dead. This only seems to happen when removing many surfaces at once, but I think fixing that is outside of the scope for this one.

lmk what you think

ItsOhen:
> okay so I added a test for it, let's see whether the CI thinks its good (half of the other tests fail on my host). But here's a couple of things:
> 
>     * I decided to use waybar instead because is is also very widely (similar to wofi probably) used I suppose and as of the latest version you can set the namespace in the config. Also it has reserved areas so we might at some point test the `order` layer rule, reserved areas, and other stuff.
> 
>     * The tester currently generates a minimal config for each waybar to `/tmp` to set the namespace, because waybar doesn't take command line arguments to set that. I hope this is fine.
> 
>     * I put the functions to spawn/kill layers also into `shared.cpp` so they are where the kitty functions are. I can also move them to the layer test file if you want but I think they might be useful for some other tests in the future too.
> 
>     * The `killAllLayers` function has a very criminal sleep in it, but I ran into a hyprland bug where killing many surfaces at once would end up with surfaces that hyprctl shows with `pid: -1`, and their process being dead. This only seems to happen when removing many surfaces at once, but I think fixing that is outside of the scope for this one.
> 
> 
> lmk what you think

Not to be that guy.. But why not use kitty layers?

VirtCode:
I don't use kitty so I don't really know anything about them. Are they like a kitty plugin or is layer shell support included by default? I just went for waybar because that seems like an actual client tons of people use, and thus is probably relatively stable, whilst I don't know how well supported the kitty layers are.

ItsOhen:
> I don't use kitty so I don't really know anything about them. Are they like a kitty plugin or is layer shell support included by default? I just went for waybar because that seems like an actual client tons of people use, and thus is probably relatively stable, whilst I don't know how well supported the kitty layers are.

They use the same layer protocol.
Play around with it for a bit.
`kitty +kitten panel --class kittylayer --lines 10 --layer top`
Documentation:
https://sw.kovidgoyal.net/kitty/kittens/panel/

vaxerski:
could be nice indeed to just use kitty if we can

VirtCode:
sorry was a bit busy the last couple of days. I have now moved it to use kitty panels instead and that also seems to work (plus we don't have weird messages no portal available and stuff which come with gtk anymore). The weirdness with closing from [above](https://github.com/hyprwm/Hyprland/pull/13331#issuecomment-3974720231) still persists however as it is a hl bug probably.

The layer spawning is now again very close to how the windows work, no longer having to write configs somewhere. What still is different is that I check for the layer of the process with the pid, instead of just the count like the windows do. I think this is in any case the better solution and don't know why the windows are done using the count thing.

If the tests pass on CI then it should be gtg again, and if you want a rebase lmk

PR #13337, opened 2026-02-23
fix: unguarded member access in moveWindowOrGroup

ssareta:
We were checking if `m_group->denied()` without first checking if the window had an `m_group`, which will segfault. Probably fixes #13330

In a search for related bugs, `CKeybindManager::denyWindowFromGroup` logic looks hella sus to me, not sure what it's supposed to be doing but I think it might lead to similar crashes the way its written atm? But I don't know what it's meant to be doing so idk

Yes, notwithstanding above separate issue


tekstryder:
This prevents the outright crash: https://github.com/hyprwm/Hyprland/discussions/13330#discussioncomment-15889020

However, `movewindoworgroup` is now a mess as described here https://github.com/hyprwm/Hyprland/discussions/13330#discussioncomment-15898185

PR #13344, opened 2026-02-23
Add configurable scroll deadzone for wheel keybinds

kristianvast:
## Summary
- add a new `binds:scroll_deadzone` float config option (default `0.0`) for wheel keybind handling
- apply deadzone gating at the start of a scroll burst, then unlock follow-up ticks for responsive continuous scrolling
- reset burst state when direction changes or events go stale (>1s), so independent scroll actions do not carry over
- wire the option into config registration and config descriptions docs

## Why
Free-spinning mouse wheels can emit tiny initial deltas that accidentally trigger workspace switching. This adds a native, opt-in threshold for that first movement while keeping subsequent intentional scrolling fast.


vaxerski:
uhhhhh something went wrong here

PR #13345, opened 2026-02-23
add new focus_fit_method: clamped_center

egormalyutin:

I like `center` `focus_fit_method`, however I think it looks cringy when I focus the last strip:

https://github.com/user-attachments/assets/d4033c11-f510-4114-b653-f5f9b26e4c7f

This PR adds a new `focus_fit_method` called `clamped_center` (2) which combines best of the fit and center worlds:

https://github.com/user-attachments/assets/7823375e-8d8a-455f-8f93-433cc58c71e5

No ugly empty space past the last strip now.

I also removed some code duplicating the `centerOrFitCol` contents.

Idk seems to work

### DIFF
--- a/src/config/ConfigDescriptions.hpp
+++ b/src/config/ConfigDescriptions.hpp
@@ -2067,7 +2067,7 @@ inline static const std::vector<SConfigOptionDescription> CONFIG_OPTIONS = {
         .value       = "scrolling:focus_fit_method",
         .description = "When a column is focused, what method should be used to bring it into view",
         .type        = CONFIG_OPTION_CHOICE,
-        .data        = SConfigOptionDescription::SChoiceData{.firstIndex = 0, .choices = "center,fit"},
+        .data        = SConfigOptionDescription::SChoiceData{.firstIndex = 0, .choices = "center,fit,clamped_center"},
     },
     SConfigOptionDescription{
         .value       = "scrolling:follow_focus",
--- a/src/layout/algorithm/tiled/scrolling/ScrollTapeController.cpp
+++ b/src/layout/algorithm/tiled/scrolling/ScrollTapeController.cpp
@@ -241,6 +241,22 @@ void CScrollTapeController::fitStrip(size_t stripIndex, const CBox& usableArea, 
     m_offset = std::clamp(m_offset, stripStart - usablePrimary + stripSize, stripStart);
 }
 
+void CScrollTapeController::clampedCenterStrip(size_t stripIndex, const CBox& usableArea, bool fullscreenOnOne) {
+    if (stripIndex >= m_strips.size())
+        return;
+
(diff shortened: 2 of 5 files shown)


nnra6864:
This is really interesting, I could definitely see myself using it.
I would like to note that since this is an actual option now, would it be possible to make the `center` method also center the leftmost(first) window?
I find it really awkward that since the layouts rewrite, everything except the first window gets centered, it's really annoying on a super ultrawide.
I mentioned this issue in a forum post too:
https://forum.hypr.land/t/scrolling-layout-regressions/1427

Appreciate the work you've done!

egormalyutin:
@nnra6864 you want something like this?

https://github.com/user-attachments/assets/54451f5f-5b25-4c45-9859-2d622e2150da

As you see here, if the tape fits into viewport, windows are centered "as a whole". However there's another possibility - to always align viewport with the centered window, even if windows fit into viewport:


https://github.com/user-attachments/assets/8436e934-4f1e-4d69-8207-3010f4e24235

I think it's better to give the user the both choices. Also now I think it's better to remove `clamped_center` and instead introduce config options like `allow_empty_space_before_first_strip_when_centering` and `allow_empty_space_after_last_strip_when_centering` (idk about naming xD).

What's causing your issue is `CScrollTapeController::calculateCameraOffset` which replaces negative offsets with 0 and centers the content if it fits into viewport, so I just removed it. I'm not quite sure though, maybe I broke some logic here. Code looks very fragile here (and not just here haha)

egormalyutin:
Also I kinda don't like that current scrolling code assumes that user scrolls the tape using keyboard or whatever or by focusing nearby clients. I don't think there's currently a way to scroll using mouse wheel or touchpad gestures without causing viewport alignment, just like I scroll a page in my browser?

nnra6864:
What you implemented above is exactly what I was talking about, it's perfect!
I would personally prefer the option where the focused window is always centered, but I can definitely see other preferring the option where if everything fits, it just fits, both are great.
As for scrolling with a mouse wheel, I genuinely have 0 clue.
The only tip I can give you for using mwheel binds is to set `binds:scroll_event_delay = 1`(it's 300 by default), otherwise, you might end up scrolling to a browser, and then the next scroll counts not as a bind, but an actual scroll in a browser.
Well that's at least what happened to me when making a zoom bind, I could just be slow tbh.

egormalyutin:
@nnra6864 okay. I'm currently not sure how to organize my changes, as I noticed some other scroll related bugs and fixed them (and for some I did not 🤪), so maybe I'll split them into different PRs or idk. I'm not really familiar with this code base very much and vaxry seems to frequently change some things so idk what are the chances of these changes ending up in the upstream

vaxerski:
rebase needed

PR #13352, opened 2026-02-24
popup: check for expired weak ptr

gulafaran:
onCommit can destroy popups while the vector CPY still holds a weak ptr to it, check if the weak ptr is still valid

### DIFF
--- a/src/desktop/view/Popup.cpp
+++ b/src/desktop/view/Popup.cpp
@@ -403,7 +403,7 @@ void CPopup::recheckChildrenRecursive() {
     std::vector<WP<CPopup>> cpy;
     std::ranges::for_each(m_children, [&cpy](const auto& el) { cpy.emplace_back(el); });
     for (auto const& c : cpy) {
-        if (!c->visible())
+        if (!c || !c->visible())
             continue;
         c->onCommit(true);
         c->recheckChildrenRecursive();


jmanc3:
@vaxerski NOOO!! Was just about to tell you this is not enough, as somewhere in `onCommit` the popup is destroyed causing `recheckChildrenRecursive` to crash on firefox hover popups on tabs when you move from the first tab over to the others.

I did this in my build to fix (just lock the weak pointer):

```c++
    for (auto const& c : cpy) {
        if (auto p = c.lock()) {
            if (!p->visible())
                continue;
            p->onCommit(true);
            p->recheckChildrenRecursive();
        }
    }
```

ItsOhen:
> @vaxerski NOOO!! Was just about to tell you this is not enough, as somewhere in `onCommit` the popup is destroyed causing `recheckChildrenRecursive` to crash on firefox hover popups on tabs when you move from the first tab over to the others.
> 
> I did this in my build to fix (just lock the weak pointer):
> 
> ```c++
>     for (auto const& c : cpy) {
>         if (auto p = c.lock()) {
>             if (!p->visible())
>                 continue;
>             p->onCommit(true);
>             p->recheckChildrenRecursive();
>         }
>     }
> ```

? I can't crash it anymore at least.

PR #13353, opened 2026-02-24
view: send wl_surface.enter to subsurfaces of popups

Skidamek:
Fix https://github.com/hyprwm/Hyprland/discussions/13346

`CPopup::onMap()` was only sending `wl_surface.enter` to the popup's own surface. Subsurfaces of the popup never received `enter`

IntelliJ is using Fifo, commit got locked, but the subsurface to which it was bound was never presented due to missing `enter` resulting in complete freeze by the Fifo lock

Fix by using `breadthfirst` instead of a direct `enter()` call, matching what `CWindow::updateSurfaceScaleTransformDetails()` already does for window surfaces

- `CLayerSurface::onMap()` has the same single-surface `enter()` pattern and should probably also use `breadthfirst`
- `sendScale()` only sends preferred scale to the popup's own surface

Ready

### DIFF
--- a/src/desktop/view/Popup.cpp
+++ b/src/desktop/view/Popup.cpp
@@ -199,7 +199,7 @@ void CPopup::onMap() {
 
     //unconstrain();
     sendScale();
-    m_resource->m_surface->m_surface->enter(PMONITOR->m_self.lock());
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks

PR #13361, opened 2026-02-25
fix(xwayland): validate size hints before floating

Skidamek:
Some xwayland clients may not send min/max size hints `WM_NORMAL_HINTS` and hyprland as a placeholder makes them all -1 then it compares if max == min and since it matches hyprland makes the window float.

Fix it by ensuring all sizes are greater than 0

Fixes https://github.com/hyprwm/Hyprland/discussions/12786

I am not sure if it's the correct fix but it works

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -679,14 +679,14 @@ static bool test() {
     EXPECT(Tests::windowCount(), 3);
 
     NLog::log("{}Checking props of xeyes", Colors::YELLOW);
-    // check some window props of xeyes, try to tile them
+    // check some window props of xeyes, try to float it
     {
         auto str = getFromSocket("/clients");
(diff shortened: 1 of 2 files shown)


vaxerski:
oops regressions?

Skidamek:
> oops regressions?

wdym

vaxerski:
you introduced test regressions

Skidamek:
So it looks like it only fails the `window.cpp@685` test, which I guess is expected? The test assumes the xeyes window should start floating, but this PR intentionally changes that behavior. Shouldn't all windows in Hyprland start tiled by default unless a windowrule says otherwise? This patch seem to align the behavior with n*ri and sw\*y, where xeyes opens as a tiled window

vaxerski:
well then please fix the test :P

Skidamek:
fixed

PR #13373, opened 2026-02-26
algo/dwindle: fix focal point not being properly used in movedTarget

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/13349


tekstryder:
Thanks. This looks good now.

PR #13374, opened 2026-02-26
Exec rule use pid

ItsOhen:
Re-add matching by pid when using exec-rules, so things that strip ENV can still get matched. (YES! I'm looking at you chrome!)
And as a bonus i saw a potential substr problem when looking at it. So fixed that aswell.
Doubt pid UINT64_MAX will ever come up.

Did change the logic of token matching to not be strict anymore. So it will be able to match other props. Not sure if a problem or not. But i don't think it should be.

Some testing would be nice.

### DIFF
--- a/src/desktop/rule/Rule.hpp
+++ b/src/desktop/rule/Rule.hpp
@@ -78,7 +78,7 @@ namespace Desktop::Rule {
             bool            isExecRule       = false;
             bool            isExecPersistent = false;
             std::string     token;
-            uint64_t        pid;
+            uint64_t        pid = 0;
             Time::steady_tp expiresAt;
         } m_execData;
     };
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -915,13 +915,13 @@ bool CKeybindManager::handleInternalKeybinds(xkb_keysym_t keysym) {
 
 // Dispatchers
 SDispatchResult CKeybindManager::spawn(std::string args) {
-    const uint64_t PROC = spawnWithRules(args, nullptr);
-    if (PROC == UINT64_MAX)
+    const auto PROC = spawnWithRules(args, nullptr);
+    if (!PROC.has_value())
         return {.success = false, .error = std::format("Failed to start process. No closing bracket in exec rule. {}", args)};
-    return {.success = PROC > 0, .error = std::format("Failed to start process {}", args)};
+    return {.success = PROC.value() > 0, .error = std::format("Failed to start process {}", args)};
 }
 
-uint64_t CKeybindManager::spawnWithRules(std::string args, PHLWORKSPACE pInitialWorkspace) {
+std::optional<uint64_t> CKeybindManager::spawnWithRules(std::string args, PHLWORKSPACE pInitialWorkspace) {
 
     args = trim(args);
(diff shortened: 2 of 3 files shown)


Dregu:
More context: All chromium browsers wipe their env for security, all password managers hide the env from the user for security, people ask every day since 53 why they can't use exec rules any more, would be nice to put back in 54...

Dregu:
Fixes keepassxc and chromium exec rules, but breaks the one thing that made token rules good, forked windows. Pid should only matter if env is empty or unreadable or something.

src/desktop/rule/windowRule/WindowRule.cpp
```diff
@@ -104,20 +104,25 @@ bool CWindowRule::matches(PHLWINDOW w, bool allowEnvLookup) {
...
+                        if (!engine->match(TKN))
+                            return false;
+                        break;
+                    }
                 }
-
-                return false;
+                break;
+            case RULE_PROP_EXEC_PID:
+                if (!engine->match(w->getPID()))
```
Dregu:
```suggestion
                if (allowEnvLookup && w->getEnv().empty() && !engine->match(w->getPID()))
```
that should do it

Dregu (changes requested):

src/desktop/rule/windowRule/WindowRule.cpp
ItsOhen: That's one way to do it. I had another idea though.

Dregu:
Still doesn't work though

ItsOhen:
> Still doesn't work though

Can you give an example so i can test it?

Dregu:
This should float, but doesn't `hyprctl dispatch exec "[float]bash <<< foot"`

src/desktop/rule/Rule.hpp
```diff
@@ -78,7 +78,8 @@ namespace Desktop::Rule {
             bool            isExecRule       = false;
             bool            isExecPersistent = false;
             std::string     token;
+            uint64_t        pid;
```
vaxerski: ay YO why is this NOT initialized

src/managers/KeybindManager.cpp
```diff
@@ -927,22 +929,29 @@ uint64_t CKeybindManager::spawnWithRules(std::string args, PHLWORKSPACE pInitial
 
     if (args[0] == '[') {
         // we have exec rules
-        RULES = args.substr(1, args.substr(1).find_first_of(']'));
-        args  = args.substr(args.find_first_of(']') + 1);
+        const auto end = args.find_first_of(']');
+        if (end == std::string::npos)
+            return UINT64_MAX;
```
vaxerski: atp return an optional

vaxerski:
> More context: All chromium browsers wipe their env for security, all password managers hide the env from the user for security, people ask every day since 53 why they can't use exec rules any more, would be nice to put back in 54...

Didn't think about that, but it makes sense

src/managers/KeybindManager.cpp
ItsOhen: I thought about that. Or a result variant. But got lazy in case other things rely on spawnraw. :P
I'll go fix it though.

PR #13379, opened 2026-02-26
windowrules: add `confine_pointer` window rule

fxzzi:

added a new rule `confine_pointer`. Lets you lock the cursor to any window you want.
I moved the pointer confine logic to a func and then kept the original logic + added separate logic for the windowrule.

Mostly useful for gaming with > 1 monitor. Some games, especially in menus, will let your cursor freely leave the game which is super annoying.

for games it makes sense to only apply the rule in fullscreen. you can tag all your games with content type game then use the below to do that:
```
match:content game, match:fullscreen 1, confine_pointer 1
```

I tested and everything worked fine, but my code may be smelly. Any improvements / suggestions are more than welcome.

I think its ready, I ran `nix fmt` also :)


y0usaf:
[praise](https://pullpo.io/cc?l=praise)

Kommynct:
It would also be nice to be able to confine the pointer to a monitor rather that a window. I control a third screen with my phone and my monitors are elsewhere, on sway there was a way to make it so that my phones cursor could only appear on that screen, with something like this I could do something close

Dregu:
> It would also be nice to be able to confine the pointer to a monitor rather that a window.

Yes sway-input has map_to_output, we have that for tablets and touchscreens, but would probably be nice to have it for any pointer device, e.g. clamping the sunshine virtual mouse to the only output you can see on moonlight makes sense. But that's a different pr...

Anyway this one seems to work.

fxzzi:
Hmm, probably a bad solution but what if you just move your virtual monitor 1 extra px away from your other monitors? Would that stop the mouse from crossing the boundary?

This of course wouldn't warp your cursor over to the right monitor etc etc though

Dregu:
> Hmm, probably a bad solution but what if you just move your virtual monitor 1 extra px away from your other monitors? Would that stop the mouse from crossing the boundary?

You'd have to make the monitor gap wider than the mouse delta, it's a funny feature actually that you can lock a slow mouse but allow fast mouse. I've used it on my tv.

fxzzi:
just rebased i think its ok, now kindly requesting review :)

vaxerski (commented):
considering its a dynamic rule whats the point of the fullscreen mode?

fxzzi:
you're right, i was kinda thinking about it from my perspective which is just that I don't want duped rules innit

like I have a set ruleset that I use for all my games, this would allow me to just add confine_pointer fullscreen to it. but on the other hand if it was just always, i'd need an extra set of rules for match:fullscreen 1, where I do the confine_pointer

but yes it woudl work either way - i can rm. i give all ma games content type game anyway so i can probably just match that in one rule /shrug

fxzzi:
yea done

`nix fmt` didn't fix the code style check is it still borken

fxzzi:
rebased for hyprlua and tested, still works fine. should be ready for a review

```
-- confine cursor to the monitor when a game is in fullscreen.
hl.window_rule({ match = { content = "game", fullscreen = true }, confine_pointer = true })
```

vaxerski (approved):
wiki

fxzzi:
> wiki

ok !

PR #13385, opened 2026-02-27
start: add --force-nixgl and check /run/opengl-driver

vaxerski:
Ref https://github.com/hyprwm/Hyprland/discussions/13312


exzombie:
I'll be able to test this in a couple of days when I get home. But I'm curious what the use case for `--force-nixgl` is?

vaxerski:
just in case™

PR #13387, opened 2026-02-27
set deadline on buffers

gulafaran:
move fence merging, exporting to drm namespace.

wait on each buffer fd fence instead of merging to avoid unnecessery syscalls.

set deadline on main buffer, and client acquirepoint/dmabuffer fds.


gulafaran:
https://github.com/hyprwm/Hyprland/pull/15006

PR #13272, opened 2026-02-14
Vulkan renderer WIP

UjinT34:
#### Describe your PR, what does it fix/add?

Refactors rendering code and internal APIs
Adds a vulkan renderer `render:use_vulkan`

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
At the current state only basic things works and it'll crash often when some bits try to use g_pHyprOpenGL.
Better test it with `render:use_vulkan = 0` to verify that gl stuff doesn't break.
Some large structs are passed by value without need. Some internal render data is copied from one structs to almost identical ones. Some render data is stored several times in different structs.
OpenGL internal texture rendering has a bunch of flags that are never used for some codepaths. Final shader/crashing can be simplified.

#### Is it ready for merging, or does it need work?
After #13488 

- [x] Rendering loop
- [x] Hypr logo
- [x] Hypr splash
- [x] Hypr cursor
- [x] FB switch
- [x] Hide `g_pHyprOpenGL` from public #13488
  - [x] Layers
  - [x] Popups
  - [x] Windows
  - [x] Pointer manager
  - [x] Protocol manager
  - [x] Animations
  - [x] CLinuxDMABufV1Protocol
  - [x] Mesa drm
  - [x] CDMABuffer
  - [x] IHyprRenderer base impl
  - [x] Shadow deco
  - [x] Render pass
  - [x] CSurfacePassElement
- [x] Frame buffer api #13437
- [x] Render buffer api #13437
- [x] Move gl specific bits to `CHyprGLRenderer` or `CHyprOpenGLImpl` #13488
  - [ ] some extra tex parameters are unhandled (unneeded?)
- [ ] Fix rebase
  - [ ] damage
  - [ ] rotation
  - [ ] v0.55 changes
- [ ] Explicit sync for clients
- [x] DMA textures
- [ ] Screen export
- [x] Scaling
- [x] Rotation
  - [x] Hypr logo
  - [x] Hypr splash
  - [x] Screen export
  - [x] Blur
  - [x] Shadow
  - [x] Surface
- [ ] disable blend
  - [ ] copy FB to output
  - [x] blur FB
  - [ ] copy for mirror
  - [ ] render mirror
  - [ ] opaque surface without rounding
- [x] Blur
  - [x] pre
  - [x] live
  - [x] ~~stencil for~~ ignore alpha/opaque
- [ ] Render elements #13438
  - [x] Border
    - [x] gradients
  - [x] Clear
[...]

PR #13434, opened 2026-02-28
[Part 0] Shader variants refactor

UjinT34:
Use glslang preprocessor for shaders.
Use `#if` to handle shader variants.
Adds variants to all shaders.
Adds experimental `render:use_shader_blur_blend` to render texture with blur and ignore alpha as a single draw call.

`render:use_shader_blur_blend` doesn't work on rotated monitors atm.
Initial part of #13272 refactors.
~~May fail to compile shaders if `GL_ARB_shading_language_include` isn't supported for ES~~

Requires glslang deps to build.
Ready

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -2200,9 +2200,10 @@ void CHyprOpenGLImpl::renderBorder(const CBox& box, const CGradientValueData& gr
 
     WP<CShader> shader;
 
+    const bool  IS_ICC = g_pHyprRenderer->workBufferImageDescription()->value().icc.present;
     const bool  skipCM = !m_cmSupported || g_pHyprRenderer->workBufferImageDescription()->id() == DEFAULT_IMAGE_DESCRIPTION->id();
     if (!skipCM) {
-        shader = useShader(getShaderVariant(SH_FRAG_BORDER1, SH_FEAT_ROUNDING | SH_FEAT_CM | SH_FEAT_TONEMAP | SH_FEAT_SDR_MOD));
+        shader = useShader(getShaderVariant(SH_FRAG_BORDER1, SH_FEAT_ROUNDING | SH_FEAT_CM | (IS_ICC ? SH_FEAT_ICC : SH_FEAT_TONEMAP | SH_FEAT_SDR_MOD)));
         passCMUniforms(shader, DEFAULT_IMAGE_DESCRIPTION);
     } else
         shader = useShader(getShaderVariant(SH_FRAG_BORDER1, SH_FEAT_ROUNDING));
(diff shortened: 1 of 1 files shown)


src/render/ShaderLoader.hpp
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include <array>
+#include <glslang/Include/glslang_c_interface.h>
+#include <string>
+#include <vector>
+#include "../helpers/memory/Memory.hpp"
+
+enum ePreparedFragmentShaderFeature : uint8_t {
```
vaxerski: can we put all of this in `namespace Render` so we can avoid poisoning the global ns?

Since we're doing cleanups already

vaxerski:
also c-f + ci fix needed, likely the deps (glslang, spirv)

vaxerski:
well no dont move the enums inside the class just namespace Render{} it all. Bot hclass and enums. 

As we move with the refactors ideally we'd move everything into Render:: and then have Render::GL::CRenderer and Render::Vk::CRenderer

UjinT34:
> well no dont move the enums inside the class just namespace Render{} it all. Bot hclass and enums.
> 
> As we move with the refactors ideally we'd move everything into Render:: and then have Render::GL::CRenderer and Render::Vk::CRenderer

Both GL and VK use the same `CShaderLoader` for source loading and preprocessing. Compilers are different and don't poison ns. #13488 has all the Renderer API/namespace bits. In general `CHyprGLRenderer` shouldn't be used directly by anything except `CHyprOpenGLImpl` and other GL internals. Any other stuff should use `g_pHyprRenderer` (`UP<IHyprRenderer>`) more or less as usual. It should have all the necessary public facing bits.

vaxerski:
this is more or less just a housekeeping preference. That's why I ask the loader to be in `Render::` because it's generic. You can see all my new code is in namespaces (e.g. `Event::`, `Layout::`, etc)

it just makes the code more tidy

vaxerski:
@fufexan nix?

vaxerski:
I think the rebase was messy, doesn't render for me anymore 
image

UjinT34:
Might be fixed now. I'll grab some ICC profile and test it myself later.
Should work without ICC profile as is.

vaxerski:
no icc looks acceptable, icc broken. 

image

UjinT34:
Should have working CM in general with ICC stuff. Screen sharing copies the buffer as is or with incorrect CM. Haven't tested mirrors.
fp16 is WIP and broken atm.

vaxerski:
will that be fixed with the fb refactor?

UjinT34:
Should work more or less like before ICC PR. ICC should be working with sRGB profiles. Screenshare requires reverse CM which doesn't support ICC and loses some color info. In general should cause less trouble than the current version on main for most common usecases. Screensharing will require to choose between displayed and exported color accuracy.
Any other enhancements are out of scope of this PR and will complicate other refactor parts. If the current state of this PR doesn't make thing worse than they were before ICC then I'd postpone figuring out screenshare and ICC shortcomings until other refactor parts are merged.

vaxerski:
I really don't like this. We need a work buffer in sRGB (probably EXT LIN so we can store negatives?) that can get truncated to 0-1 for screencopy.

> Screensharing will require to choose between displayed and exported color accuracy.

Screensharing _must never_ use exported color. It's always sRGB.

I don't want my work buffers in monitor space.

PR #13438, opened 2026-02-28
[Part 2] Refactor render elements

UjinT34:
Moves render pass elements `draw` methods into HyprRenderer as a part of refactors for vulkan renderer.

Part 2 of #13272 refactors

Includes #13437 and should be reviewed after that.
Ready.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -752,7 +752,7 @@ void CHyprRenderer::draw(WP<IPassElement> element, const CRegion& damage) {
     if (!element)
         return;
 
-    switch (element->kind()) {
+    switch (element->type()) {
         case EK_BORDER: draw(dc<CBorderPassElement*>(element.get()), damage); break;
         case EK_CLEAR: draw(dc<CClearPassElement*>(element.get()), damage); break;
         case EK_FRAMEBUFFER: draw(dc<CFramebufferElement*>(element.get()), damage); break;
--- a/src/render/pass/PassElement.hpp
+++ b/src/render/pass/PassElement.hpp
@@ -2,7 +2,7 @@
 
 #include "../../defines.hpp"
 
-enum ePassElementKind : uint8_t {
+enum ePassElementType : uint8_t {
     EK_UNKNOWN = 0,
     EK_BORDER,
     EK_CLEAR,
@@ -23,7 +23,7 @@ class IPassElement {
     virtual bool                needsLiveBlur()       = 0;
     virtual bool                needsPrecomputeBlur() = 0;
     virtual const char*         passName()            = 0;
-    virtual ePassElementKind    kind()                = 0;
+    virtual ePassElementType    type()                = 0;
     virtual void                discard();
     virtual bool                undiscardable();
     virtual std::optional<CBox> boundingBox();  // in monitor-local logical coordinates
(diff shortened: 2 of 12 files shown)


src/render/pass/PassElement.hpp
```diff
@@ -1,16 +1,29 @@
 #pragma once
 
 #include "../../defines.hpp"
-#include <optional>
+
+enum ePassElementKind : uint8_t {
```
vaxerski: I prefer the term "Type", we use that across the ecosystem

src/render/Renderer.cpp
```diff
@@ -747,6 +748,309 @@ void CHyprRenderer::renderWindow(PHLWINDOW pWindow, PHLMONITOR pMonitor, const T
     g_pHyprOpenGL->m_renderData.currentWindow.reset();
 }
 
+void CHyprRenderer::draw(WP<IPassElement> element, const CRegion& damage) {
+    if (!element)
+        return;
+
+    switch (element->kind()) {
+        case EK_BORDER: draw(dc<CBorderPassElement*>(element.get()), damage); break;
```
vaxerski, line 756: hyprutils has a dynamicPointerCast, use that and avoid raw C pointers

src/render/Renderer.cpp
```diff
@@ -747,6 +748,309 @@ void CHyprRenderer::renderWindow(PHLWINDOW pWindow, PHLMONITOR pMonitor, const T
     g_pHyprOpenGL->m_renderData.currentWindow.reset();
 }
 
+void CHyprRenderer::draw(WP<IPassElement> element, const CRegion& damage) {
```
vaxerski, line 751: put these methods in a separate .cpp file, Renderer.cpp is crowded enough

src/render/Renderer.hpp
```diff
@@ -174,6 +186,18 @@ class CHyprRenderer {
                                    SP<CWLSurfaceResource> surface = nullptr, bool modifySDR = false, float sdrMinLuminance = -1.0f, int sdrMaxLuminance = -1);
     bool             reloadShaders(const std::string& path = "");
 
+    void             draw(CBorderPassElement* element, const CRegion& damage);
```
vaxerski, line 189: why are these public?

UjinT34: temporary. they become protected virtuals later

src/render/Renderer.cpp
UjinT34: temporary. those will be abstract later. some of the non-gl specific will remain in Renderer.cpp and should be moved after 6.

src/render/Renderer.cpp
UjinT34: after 6. there'll be some additional changes involving this

vaxerski: elaborate

src/render/Renderer.cpp
vaxerski: ack

src/render/Renderer.hpp
vaxerski: ack

src/render/Renderer.cpp
UjinT34: this draw function will be modified, a bunch of new function with raw pointers will be added and some bits will be moved into glrenderer with the same raw pointer api.
rebase isn't as straightforward as I'd like it to be due to c-f messing with unmodified lines.

PR #13441, opened 2026-02-28
monitor: update pinned window states properly on changeWorkspace

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/13440

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -566,6 +566,53 @@ static bool testWindowRuleFocusOnActivate() {
     return true;
 }
 
+// tests if a pinned window contains the valid workspace after change
+static bool testPinnedWorkspacesValid() {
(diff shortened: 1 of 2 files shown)


Homisiak:
It should be a toggle in config as it introduces new behaviors which may be unwanted for some. In my case, it breaks my workflow when pinned floating windows move on changeWorkspace if they were positioned over a reserved area (e.g. waybar, eww bar).

Homisiak:
This still is an issue. The change from `w->moveToWorkspace(pWorkspace);` to `w->layoutTarget()->assignToSpace(pWorkspace->m_space);` breaks a lot of workflows. It's not a bad change, but needs to be toggleable. I'm sad to see that I can no longer use the regular arch package to install Hyprland because of it 😔

vaxerski:
it can't be toggleable because not doing this will break and crash later down the line. The bug can be fixed separately

PR #13475, opened 2026-03-01
desktop/window: fix floating windows being auto-grouped

andresilva:
The layout refactoring in #12890 moved the `auto_group` logic from `IHyprLayout::onWindowCreatedAutoGroup()` into `CWindow::mapWindow()` but the new inline checks missed cases that the original `!shouldBeFloated()` guard covered. Specifically, Wayland dialogs with fixed size but no explicit toplevel parent (like gcr-prompter) would pass the `!(parent() && m_isFloating)` check and get added into the focused group instead of floating freely.

Restore the original `!shouldBeFloated()` check, which covers dialogs, transients, and fixed-size windows for both X11 and Wayland without preventing user-configured floating windows from being auto-grouped.

### DIFF
--- a/hyprtester/src/tests/main/groups.cpp
+++ b/hyprtester/src/tests/main/groups.cpp
@@ -253,7 +253,7 @@ static bool test() {
         Tests::killAllWindows();
         EXPECT(Tests::windowCount(), 0);
     }
-    
+
     // test that we deny a floated window getting auto-grouped into a tiled group.
     NLog::log("{}Test that we deny a floated window getting auto-grouped into a tiled group.", Colors::GREEN);
 


Aqa-Ib:
What if the new window is a parent and is floating, and it is going to be auto-grouped into the focused floating group ?

andresilva:
@Aqa-Ib updated, should have the same behavior as 0.53.

src/desktop/view/Window.cpp
```diff
@@ -1949,11 +1949,11 @@ void CWindow::mapWindow() {
...
-    if (*PAUTOGROUP                                                              // auto_group enabled
-        && Desktop::focusState()->window()                                       // focused window exists
-        && canBeGroupedInto(Desktop::focusState()->window()->m_group)            // we can group
-        && Desktop::focusState()->window()->m_workspace == m_workspace           // workspaces match, we're not opening on another ws
-        && !isModal() && !(parent() && m_isFloating) && !isX11OverrideRedirect() // not a modal, floating child or X11 OR
+    if (*PAUTOGROUP                                                                        // auto_group enabled
+        && Desktop::focusState()->window()                                                 // focused window exists
+        && canBeGroupedInto(Desktop::focusState()->window()->m_group)                      // we can group
+        && Desktop::focusState()->window()->m_workspace == m_workspace                     // workspaces match, we're not opening on another ws
+        && !g_pXWaylandManager->shouldBeFloated(m_self.lock()) && !isX11OverrideRedirect() // not a window that should float or X11 OR
```
andresilva: Not obvious to me whether we should just check `m_isFloating` here (that's what I did initially before @Aqa-Ib's comment). With `m_isFloating`, a user-configured floating window (via window rules) would also be prevented from auto-grouping. The original code had float-state toggling when grouping (`floating -> tiled` and vice versa), but the new layout doesn't, so auto-grouping a floating window into a tiled group would be broken anyway. `m_isFloating` is also simpler since `shouldBeFloated()` was already called and its result stored in `m_isFloating` by this point. Happy to switch to `!m_isFloating` if you agree.

andresilva: Should also be easier to write a test if using `m_isFloating` since I can just setup a rule to make kitty float and check that it isn't auto-grouped.

vaxerski: yes m_isFloating makes more sense

Aqa-Ib:
I tested this PR and with this approach auto-group behaviour gets broken for floated into floated.
Are we really breaking this behaviour for the test's sake? I really think this is a bad decision. Just check for shouldBeFloated and ditch the test if it is that difficult to make. It is much more important to have consistency and not breaking existing features. 

I spent a lot of time to implement this feature. It would feel so bad to see it deleted this way. Please, reconsider.

andresilva:
What is "floated into floated"? If we only check for `shouldBeFloated` then windows that have a float rule are auto-grouped, which IMO is not desired.

Aqa-Ib:
Floated windows don't get auto-grouped into tiled, that's denied already. 
Floated into floated means autogruping a new floating window into an exisiting floating group. That's a feature, it didn't make any problem in the year and a half that has been working this way. I can explain to you in discord if you want to talk extensively, I have messaged you in hyprland-dev channel.

andresilva:
Reverted to `shouldBeFloated` and updated test.

vaxerski:
it should use m_isFloating instead. Otherwise it will ignore rules.

tekstryder:
@andresilva guess I missed this PR and filed an issue.

You can add Fixes: https://github.com/hyprwm/Hyprland/discussions/13521

Aqa-Ib:
ok now I see what is going on here. Let me fix it. Checking shouldBeFloated is fine, the problem is that rethonk missed denying autogroup floated into tiled.

Aqa-Ib:
I am going to send a fix into andresilva's branch, hold on.

Aqa-Ib:
PR opened on andresilva's branch to fix what vaxry complaint about: floated windows won't get auto-grouped into a tiled group, thus respecting rules. This was left behind in rethonk.

Making a test for this too, hold on.

Aqa-Ib:
Added a test (recycled from andresilva's previous test), all should be good to go when andresilva merges it, cheers.

andresilva:
@Aqa-Ib included your changes and rebase over main. I didn't run the tests or test this locally yet.

Aqa-Ib:
@andresilva I sent you a new PR to improve the test to be extra sure that the first kitty gets tiled. Thanks!

Aqa-Ib:
Test failed, I will fix.

Aqa-Ib:
@andresilva sent a PR fixing the test, muito obrigado.

vaxerski:
c-f

andresilva:
github is going through its weekly downtime period and I can't push...

andresilva:
@vaxerski you might need to manually re-trigger the ci pipeline

vaxerski:
oh god damnit you removed the fucking isModal() from there why

Aqa-Ib:
@andresilva um prazer trabalhar com você, valeu!

vaxerski:
yeah voce voce where modaloce

andresilva:
Obrigado 😀

PR #13476, opened 2026-03-01
Added helper function to set configuredWidths properly on new Desktops

RockClapps:
This PR fixes a bug in the Scrolling layout that only applies the configured `explicit_column_widths` variable to the current workspace at load/refresh time. I wrap the parsing in a function similar to `parseDirection` so that it can be used during refresh *and* the constructor.

This is my first real OSS PR so I might have missed something trivial.

As far as I can tell it is ready to go!


vaxerski (approved):
tnx

PR #13488, opened 2026-03-02
[Part 6] Refactor gl renderer

UjinT34:
Moved some render data inside render pass elements.
Those elements are used for rendering instead of direct `g_pHyprOpenGL` calls.
Use `IFramebuffer::readPixels` to get image data from FBs.

`g_pHyprOpenGL` remains exposed and can be used by plugins but some bits might be moved inside render pass elements and renderer.
Render pass elements might not provide the exact 1:1 match to direct `g_pHyprOpenGL` calls.
Render pass elements might not use correct damage and clip boxes/regions (meaning larger damage/clips than needed)
Part 6 #13272 refactors

Includes #13485 and should be reviewed after that.
Ready.

### DIFF
--- a/src/render/GLRenderer.cpp
+++ b/src/render/GLRenderer.cpp
@@ -22,6 +22,7 @@
 #include "render/gl/GLTexture.hpp"
 #include "decorations/CHyprDropShadowDecoration.hpp"
 
+#include <cstdint>
 #include <hyprutils/memory/SharedPtr.hpp>
 #include <hyprutils/utils/ScopeGuard.hpp>
 using namespace Hyprutils::Utils;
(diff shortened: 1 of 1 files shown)


src/render/pass/Pass.cpp
```diff
@@ -213,9 +213,15 @@ void CRenderPass::renderDebugData() {
     const auto pMonitor = g_pHyprRenderer->m_renderData.pMonitor;
     CBox       box      = {{}, pMonitor->m_transformedSize};
     for (const auto& rg : m_occludedRegions) {
-        g_pHyprOpenGL->renderRect(box, Colors::RED.modifyA(0.1F), {.damage = &rg});
+        CRectPassElement::SRectData data;
+        data.box   = box;
+        data.color = Colors::RED.modifyA(0.1F);
+        g_pHyprRenderer->draw(makeUnique<CRectPassElement>(data), rg);
```
vaxerski, line 219: why is this exploded, can't we `makeUnique<CRectPassElement>({.box = box, .color = ...}), rg`?

UjinT34: copy-pasted this approach from somewhere else without much thinking. `draw` should probably accept `CRectPassElement::SRectData` and similar directly. I'd leave such changes for part 7

PR #13490, opened 2026-03-02
desktop/group: respect direction when moving window out of group

andresilva:
Restores pre-0.54.0 behavior where `movewindoworgroup` respected the direction parameter when extracting a window from a group. The direction was lost during the layout rewrite because `CGroup::remove()` placed the window back into the layout with no positional hint. Pass a focal point at the edge of the group box in the requested direction, and make dwindle use smart-split logic when an explicit focal point is provided so that all four directions (not just left/right) produce the correct split orientation.

Fix https://github.com/hyprwm/Hyprland/discussions/13443.

### DIFF
--- a/hyprtester/src/tests/main/dwindle.cpp
+++ b/hyprtester/src/tests/main/dwindle.cpp
@@ -64,16 +64,16 @@ static void test13349() {
 
     {
         auto str = getFromSocket("/activewindow");
-        EXPECT_CONTAINS(str, "at: 22,547");
-        EXPECT_CONTAINS(str, "size: 931,511");
+        EXPECT_CONTAINS(str, "at: 497,22");
+        EXPECT_CONTAINS(str, "size: 456,1036");
     }
 
     OK(getFromSocket("/dispatch movewindow r"));
 
     {
         auto str = getFromSocket("/activewindow");
-        EXPECT_CONTAINS(str, "at: 967,547");
-        EXPECT_CONTAINS(str, "size: 931,511");
+        EXPECT_CONTAINS(str, "at: 967,22");
+        EXPECT_CONTAINS(str, "size: 456,1036");
     }
 
     // clean up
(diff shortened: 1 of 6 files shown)


vaxerski:
would be nice if the tests didnt fail xd

hyprtester/src/tests/main/dwindle.cpp
```diff
@@ -64,16 +64,16 @@ static void test13349() {
 
     {
         auto str = getFromSocket("/activewindow");
-        EXPECT_CONTAINS(str, "at: 22,547");
-        EXPECT_CONTAINS(str, "size: 931,511");
+        EXPECT_CONTAINS(str, "at: 497,22");
+        EXPECT_CONTAINS(str, "size: 456,1036");
```
andresilva, line 68: `movewindow` now uses `smart_split` logic when given an explicit focal point, so the split direction follows the focal point angle rather than the box aspect ratio (same as `movewindoworgroup`).

tekstryder:
Tests issue aside, this fixes https://github.com/hyprwm/Hyprland/discussions/13443. Thx!

hyprtester/src/tests/main/dwindle.cpp
vaxerski: why? It's disabled.

andresilva: https://github.com/andresilva/Hyprland/blob/907790aae409cfdb4ff9e4ba7485b61d598f8a07/src/layout/algorithm/tiled/dwindle/DwindleAlgorithm.cpp#L185

vaxerski: ah, this will pair well with https://github.com/hyprwm/Hyprland/pull/13514 then

fxzzi:
https://github.com/user-attachments/assets/f3c88a74-b2b8-44c7-a894-aa4437063291

it's better but `movewindow` behaviour has changed. Before it would move windows like below: 

https://github.com/user-attachments/assets/0d933f64-ea8d-4b75-9499-83891351de5a (still broken but closer to previous behaviour)

now it changes the split to top,bottom, then skips to the next monitor. Basically I can't flip the position of the two windows on the left monitor without moving it to the right monitor and then back to the left.

PR #13500, opened 2026-03-02
layout: store and preserve size and pos after fullscreen

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/13401


tekstryder:
MUCH better. Thanks for picking this up.

PR #13506, opened 2026-03-02
nix: separate overlay with deps

fufexan:

Separates Hyprland and its internal deps' overlays (udis, glaze) from "external" deps (hypr* stuff).

Users are free to choose whether to use the raw hyprland overlay or bring in dependencies.

Should avoid issues like https://github.com/hyprwm/Hyprland/discussions/13396 popping up again.

Wiki PR https://github.com/hyprwm/hyprland-wiki/pull/1419.

Not sure which overlay should be the default.

@spikespaz @NotAShelf and other Nix-inclined people, please take a look and chime in.

Not ready until we address the `default` issue.

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -19,12 +19,12 @@ in
   default = lib.composeManyExtensions (
     with self.overlays;
     [
-      hyprland-packages
+      hyprland
       hyprland-extras
     ]
   );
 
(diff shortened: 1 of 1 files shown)


fufexan:
Since no one commented in such a long time, I'll merge this as is.

For the (presumably) few people using overlays, this should be a one-time inconvenience.

PR #13519, opened 2026-03-02
feat(flake): substituters

Malix-Labs:

feature [native substituters suggestion / hint](https://wiki.nixos.org/wiki/Binary_Cache#Binary_cache_hint_in_Flakes) to `flake.nix`

image

This will also be shown in `flake.nix` importing this flake

No

Yes

## References

- https://wiki.nixos.org/wiki/Flakes#Nix_configuration
- [ ] hyprwm/hyprland-wiki#1400

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -88,6 +88,15 @@
     };
   };
 
+  nixConfig = {
+    extra-substituters = [
+      "https://hyprland.cachix.org/"
+    ];
+    extra-trusted-public-keys = [
+      "hyprland.cachix.org-1:a7pgxzMz7+chwVL3/pzj6jIBMioiJM7ypFP8PwtkuGc="
+    ];
+  };
+
   outputs =
     inputs@{
       self,


fufexan:
I'm generally against using `nixConfig` as upstream Nix does not have a way to ignore it, and forces you to interact. Lix ignores it by default, but not everyone uses it.

The benefits would be minimal, as I don't think anyone runs hyprland with `nix run` from the flake that often.

Malix-Labs:
> As upstream Nix does not have a way to ignore it

Yes this is true,
Is not being able to ignore the sustituters hint an issue?

> The benefits would be minimal, as I don't think anyone runs hyprland with nix run from the flake that often.

It would also hint for imperatively adding the keys to `nix.conf` permanently when you use hyprland as a flake input of a nixos system (and make sure users are aware of the cache)

fufexan:
> Is not being able to ignore the sustituters hint an issue?

I've heard people complain about it ever since `nixConfig` got introduced. I personally don't mind since I use Lix. We could add it and remove it afterwards if people complain.

> It would also hint for imperatively adding the keys to `nix.conf` permanently when you use hyprland as a flake input of a nixos system (and make sure users are aware of the cache)

Right, more visibility doesn't hurt.

@NotAShelf thoughts?

NotAShelf:
As I've [previously written](https://notashelf.dev/posts/reject-flake-content) I'm not very fond of the `nixConfig` option, and I don't consider Cachix a trusted party anymore. If anything, I'm against adding it to the Nix flake itself. Most, if not all, Nix users know how to use a substituters.

Malix-Labs:
https://github.com/hyprwm/Hyprland/pull/13519#issuecomment-4003424567 from @NotAShelf :
> I don't consider Cachix a trusted party anymore

Do you have something written somewhere about that?

The official binary cache of hyprland has been Cachix for as much as I can remember I think

> Most, if not all, Nix users know how to use a substituters

I would tend to agree, but having an official binary cache itself be set in the upstream flake and enabled at the same time as adding the flake input instead of having to duplicate it elsewhere in downstream's nix configs is arguably a small QoL enhancement

NotAShelf:
> [#13519 (comment)](https://github.com/hyprwm/Hyprland/pull/13519#issuecomment-4003424567) from @NotAShelf :
> 
> > I don't consider Cachix a trusted party anymore
> 
> Do you have something written somewhere about that?
> 
> The official binary cache of hyprland has been Cachix for as much as I can remember I think
> 
> > Most, if not all, Nix users know how to use a substituters
> 
> I would tend to agree, but having an official binary cache itself be set in the upstream flake and enabled at the same time as adding the flake input instead of having to duplicate it elsewhere in downstream's nix configs is arguably a small QoL enhancement

No, I haven't written about it yet but not a while ago the CEO of Cachix (or whatever his official position is) has abused his committer privileges on Nixpkgs to discard a change that inconveniences his company. This tells me a lot about where the interests of a particular entity lies, and as such, I trust it *untrusted*. For the scope of something as potentially dangerous as `nixConfig`, cachix lists quite high in my to-avoid list.

I also don't quite see the benefit of taking this risk. It's *too small* of a QoL improvement; most users that will end up developing Hyprland on NixOS will already have the Cachix added to their system substituters, so this makes so very little sense to me. That said, I don't have a concrete stance against it. If there's agreement that it is worth the risk then go ahead.

Malix-Labs:
https://github.com/hyprwm/Hyprland/pull/13519#issuecomment-4004624974 from @NotAShelf :
> I also don't quite see the benefit of taking this risk. It's too small of a QoL improvement; most users that will end up developing Hyprland on NixOS will already have the Cachix added to their system substituters, so this makes so very little sense to me

Wouldn't it also benefits all users of hyprland importing it using a flake input in their nixos configuration?

NotAShelf:
No, that is not how `nixConfig` works. It only adds the substituters for nix3 commands. To adjust substituters for configs, you'd have to add substituters to the NixOS module and thus modify the actual system config. I find this equally undesirable.

PR #13544, opened 2026-03-04
fix empty workspace handling

justin4046:
Fixes `windowrule = workspace empty` which since v0.54 would only work once 2 windows were open. Also allows the n/m arguments to be caught which wouldn't previously match with `contains()`

Bisecting had somewhere around the new layout commit as the breaking change but I didn't look into that since this seems like the proper fix.

Ready


justin4046:
im not sure how to get the same monitor test to work in hyprtester so it doesnt include that

PR #13586, opened 2026-03-05
screencopy: fix isOutputBeingSSd

vaxerski:
use sessions instead of pending frames

@fxzzi


fxzzi:
all good now thanks g

PR #13606, opened 2026-03-06
keybinds: A more sensible interpretation of `fullscreenstate * toggle`

kolayne:

Implements a new toggling logic which works as follows:
- Toggling any state with 0 is the same state
- Toggling any state with itself is 0
- Toggling any state with 3 brings to 3 (maximized + fullscreen)
- Toggling any other state with 2 (fullscreen) flips the fullscreen bit
- Toggling any other state with 1 (maximized) resets to the maximized state

Some behaviors change. IMO most behaviors make more sense now, than before. This is arguable for `fullscreenstate, 0, 0[, toggle]`: it becomes a no-op. Do you agree or should we treat requests to toggle zeros specially and instead _switch_ to 0?

Needs a test. Let me know if you agree with the new behavior, and then I will implement the test then.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -1318,6 +1318,41 @@ SDispatchResult CKeybindManager::fullscreenActive(std::string args) {
     return {};
 }
 
+/// Logically toggles the `toggleMode` state in the `oldMode` state and returns the new mode.
+/// Corresponds to the `fullscreenstate` dispatcher.
+static enum eFullscreenMode fullscreenModeToggle(enum eFullscreenMode oldMode, enum eFullscreenMode toggleMode) {
+    static_assert(FSMODE_MAX == 3, "Update this function's logic if new fullscreen states are introduced");
+
+    switch (toggleMode) {
+        case FSMODE_MAXIMIZED | FSMODE_FULLSCREEN:
+            // When the request is to toggle two bits at the same time, we assume the user wants to sync the states of
+            // maximizedness and fullscreenness (enable both or disable both).
+            //
+            // If we interpreted this as bit toggling, in case the old window mode is either maximized or fullscreen,
+            // toggling both would continuosuly switch the window between maximized and fullscreen. It's unlikely to be
+            // the user's intention.
+            return oldMode == toggleMode ? FSMODE_NONE : toggleMode;
+        case FSMODE_FULLSCREEN:
+            // When the request is to toggle fullscreenness, we assume the user wants to toggle the fullscreenness bit,
+            // regardless of whether the window is also maximized.
+            return sc<eFullscreenMode>(oldMode ^ FSMODE_FULLSCREEN);
+        case FSMODE_MAXIMIZED:
+            // When the request is to toggle maximizedness, but the window is already fullscreen, we assume the user
+            // wants to *switch* to the maximized state.
+            //
+            // If we interpreted this as bit toggling, the window would continuously switch between fullscreen and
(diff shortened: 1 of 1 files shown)


kolayne:
@vaxerski, what do you think of the new behaviors, especially the `fullscreenstate 0 0 toggle`?

kolayne:
Hey, @vaxerski, waiting for your response to continue working on this. Please, take a look at the PR description

MightyPlaza:
This probably needs to rework the syntax to match what is being used in git

Additionally, since you are rethinking how toggling should work, you should probably include handling of `sync_fullscreen`, which users probably expect to be set back to True on toggling (in the cases where it makes sense)

Current behavior (if I'm not mistaken) is just set it to True if both states match.

kolayne:
Ah, shit. Testing the new hyprland now, and the fullscreen_state dispatcher seems to have bugs both with the legacy and the lua config, but the bugs are different...

Hopefully, I will have time to look into it in the coming days.

---

> since you are rethinking how toggling should work, you should probably include handling of `sync_fullscreen`, which users probably expect to be set back to True on toggling (in the cases where it makes sense)
>
> Current behavior (if I'm not mistaken) is just set it to True if both states match.

@MightyPlaza, yes, it seems like this is the current behavior. Do you expect something different? What kind of behavior would you like?

Also, what is `sync_fullscreen`? I have never seen that.

MightyPlaza:
> Also, what is `sync_fullscreen`? I have never seen that.

`sync_fullscreen` is a window prop that synchronizes the internal and the client states, so functions that change one also change the other.
It needs to be reset after you've toggled `fullscreenstate 0 2` for example so when you later fullscreen in the client it also fullscreens in Hyprland.

kolayne:
> `sync_fullscreen` is a window prop that synchronizes the internal and the client states, so functions that change one also change the other.

I am so confused. I found it in the wiki, and the wiki description still does not explain it. Could you provide an example: a sequence of actions and the resulting outcome for a window that is `sync_fullscreen` and a window that is not `sync_fullscreen`?



> It needs to be reset after you've toggled `fullscreenstate 0 2` for example so when you later fullscreen in the client it also fullscreens in Hyprland.

What even is the effect of `fullscreenstate 0 2` if the window is `sync_fullscreen`? According to the description, the "internal" and "client" (communicated) fullscreen states must be the same...

MightyPlaza:
> I am so confused. I found it in the wiki, and the wiki description still does not explain it. Could you provide an example: a sequence of actions and the resulting outcome for a window that is `sync_fullscreen` and a window that is not `sync_fullscreen`?

Suppose you toggle `fullscreenstate 0 2` twice on firefox, but then later you click fullscreen in a video. If `sync_fullscreen` isn't reset to `1` the state will be `0 2` instead of `2 2`.

> What even is the effect of `fullscreenstate 0 2` if the window is `sync_fullscreen`? According to the description, the "internal" and "client" (communicated) fullscreen states must be the same...

Using `fullscreenstate` should override `sync_fullscreen` and set it to false if needed.

kolayne:
@MightyPlaza, so, it seems like the behavior you described is the currently implemented one, isn't it?

Anyway, I opened a discussion #14646, let's continue the discussion there. Please, read my suggestions for the new logic and comment there if your expectations are different.

I am closing this MR, planning to reimplement these changes as part of #14646

MightyPlaza:
> @MightyPlaza, so, it seems like the behavior you described is the currently implemented one, isn't it?

Yes, I believe so.
When I implemented it, i believed it was the simplest solution that made sense.
But since you're proposing to rethink how the toggling works, I thought you should be aware of the full picture.

PR #13626, opened 2026-03-07
layout/windowTarget: add visualBox

outfoxxed:
Re-adds the ability for layouts to set the visual position of a window. Required to port hy3 to 0.54.

There are a number of other issues related to usage of logical/visual size in other places but I am not addressing them in this PR as they existed previously.

This is currently based on 0.54.1 so it will need to be rebased for main.

### DIFF
--- a/src/layout/target/Target.hpp
+++ b/src/layout/target/Target.hpp
@@ -25,6 +25,11 @@ namespace Layout {
         std::optional<Vector2D> pos;
     };
 
+    struct STargetBox {
+        CBox logicalBox;
+        CBox visualBox;
+    };
+
     class ITarget {
       public:
         virtual ~ITarget() = default;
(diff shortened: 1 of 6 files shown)


vaxerski:
wrong target

src/layout/target/Target.hpp
```diff
@@ -25,14 +25,20 @@ namespace Layout {
...
     class ITarget {
       public:
         virtual ~ITarget() = default;
 
         virtual eTargetType type() = 0;
 
         // position is within its space
-        virtual void         setPositionGlobal(const CBox& box);
+        virtual void         setPositionGlobal(const STargetBox& box);
+        void         setPositionGlobal(const CBox& box);
```
vaxerski: why is this not virtual

outfoxxed: Because it just calls the other one. I would've added a CBox constructor to STargetBox but that required adding CBox {} to all call sites using {} to construct a box

PR #9210, opened 2025-01-29
fix(render): calculate UV using both pixel and monitor dimensions

nnyyxxxx:
fixes #9175

PR #13631, opened 2026-03-07
[Part 7] Api fixes

UjinT34:
Avoids raw pointers.
Extracts element rendering to a separate class.
Moves most of the gl render stuff into `Render::GL` ns.
Moves monitor resources into separate class. Makes all render FBs equal and same. Stores offload FB for screensharing. Allows up to 8 work FBs which are destroyed if remain unused for more than 5 seconds (instead of 5 permanent named FBs on main)
Adds image descriptions to ITexture. Uses this image description to determine source and target descriptions if it's set.
Adds a new `EK_CUSTOM` pass element type and a virtual `draw` function to `IPassElement`. This function is used only with `EK_CUSTOM` atm and is supposed to return a vector of predefined render pass elements to be rendered instead. If a plugin needs a direct access to gl internals then it can use `g_pHyprRenderer->type() == RT_GL` to check for a gl renderer.
Adds `IHyprRenderer::bindFB` helper to bind fb and update `m_renderData.currentFB` simultaneously.
Adds `IHyprRenderer::bindTempFB` helper which calls `bindFB` and returns a `CScopeGuard` which restores the previous FB.
`CFramebufferElement` is deprecated. Plugins should use `g_pHyprRenderer->m_renderData` to get internal buffers and `g_pHyprRenderer->m_renderData.pMonitor->resources()` to get monitor buffers.
Transformer plugins should use SP instead of a raw pointer.
Adds `render:use_fp16` to use FP16 format for internal buffers. 0 - disabled, 1 - always on, 2 - only for monitors in HDR mode.
Adds a second render target without sdr mods when needed: screensharing while sdr mods are applied. Doesn't work correctly with screenshots because it becomes valid one frame later than needed. `render:keep_unmodified_copy` fixes it by keeping the copy even when not needed. 0 - disabled, 1 - always, 2 - only when sdr mods are applied.

Part 7 #13272 refactors
Might've messed with some pointers and conditions. `CMonitorResources` uses `strongRef` to check whether the buffer is used by someone else.
~~Texture with blur sometimes gets incorrect damage (#13488) and has glitches near the border. More obvious with bigger border radius values.~~
FP16 doubles vram usage and has some quirks. Should be considered experimental.
Some CM can be simplified if shaders are modified to work in linear space internally.

- [x] Fix blur damage
- [x] Fix screen share projection
- [x] Plugins API
- [x] FP16
  - [x] fix CM
  - [x] fix damage
- [x] Texture image descriptions
- [x] Screenshare fixes

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2599,10 +2599,10 @@ void CCompositor::openSafeModeBox() {
 
     auto       box = CAsyncDialogBox::create(I18n::i18nEngine()->localize(I18n::TXT_KEY_SAFE_MODE_TITLE), I18n::i18nEngine()->localize(I18n::TXT_KEY_SAFE_MODE_DESCRIPTION),
                                              {
-                                           OPT_LOAD,
-                                           OPT_OPEN,
-                                           OPT_OK,
-                                       });
+                                                 OPT_LOAD,
+                                                 OPT_OPEN,
+                                                 OPT_OK,
+                                             });
 
     box->open()->then([OPT_LOAD, OPT_OK, OPT_OPEN, this](SP<CPromiseResult<std::string>> result) {
         if (result->hasError())
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -501,11 +501,11 @@ static std::string getWorkspaceRuleData(const Config::CWorkspaceRule& r, eHyprCt
         const std::string default_    = sc<bool>(r.m_isDefault) ? std::format(",\n    \"default\": {}", boolToString(r.m_isDefault)) : "";
         const std::string persistent  = sc<bool>(r.m_isPersistent) ? std::format(",\n    \"persistent\": {}", boolToString(r.m_isPersistent)) : "";
         const std::string gapsIn      = sc<bool>(r.m_gapsIn) ?
-                 std::format(",\n    \"gapsIn\": [{}, {}, {}, {}]", r.m_gapsIn.value().m_top, r.m_gapsIn.value().m_right, r.m_gapsIn.value().m_bottom, r.m_gapsIn.value().m_left) :
-                 "";
+            std::format(",\n    \"gapsIn\": [{}, {}, {}, {}]", r.m_gapsIn.value().m_top, r.m_gapsIn.value().m_right, r.m_gapsIn.value().m_bottom, r.m_gapsIn.value().m_left) :
+            "";
         const std::string gapsOut     = sc<bool>(r.m_gapsOut) ?
-                std::format(",\n    \"gapsOut\": [{}, {}, {}, {}]", r.m_gapsOut.value().m_top, r.m_gapsOut.value().m_right, r.m_gapsOut.value().m_bottom, r.m_gapsOut.value().m_left) :
-                "";
+            std::format(",\n    \"gapsOut\": [{}, {}, {}, {}]", r.m_gapsOut.value().m_top, r.m_gapsOut.value().m_right, r.m_gapsOut.value().m_bottom, r.m_gapsOut.value().m_left) :
+            "";
         const std::string borderSize  = sc<bool>(r.m_borderSize) ? std::format(",\n    \"borderSize\": {}", r.m_borderSize.value()) : "";
         const std::string border      = sc<bool>(r.m_noBorder) ? std::format(",\n    \"border\": {}", boolToString(!r.m_noBorder.value())) : "";
         const std::string rounding    = sc<bool>(r.m_noRounding) ? std::format(",\n    \"rounding\": {}", boolToString(!r.m_noRounding.value())) : "";
@@ -525,13 +525,13 @@ static std::string getWorkspaceRuleData(const Config::CWorkspaceRule& r, eHyprCt
         const std::string default_    = std::format("\tdefault: {}\n", sc<bool>(r.m_isDefault) ? boolToString(r.m_isDefault) : "<unset>");
         const std::string persistent  = std::format("\tpersistent: {}\n", sc<bool>(r.m_isPersistent) ? boolToString(r.m_isPersistent) : "<unset>");
         const std::string gapsIn      = sc<bool>(r.m_gapsIn) ?
-                 std::format("\tgapsIn: {} {} {} {}\n", std::to_string(r.m_gapsIn.value().m_top), std::to_string(r.m_gapsIn.value().m_right),
-                             std::to_string(r.m_gapsIn.value().m_bottom), std::to_string(r.m_gapsIn.value().m_left)) :
-                 std::format("\tgapsIn: <unset>\n");
+            std::format("\tgapsIn: {} {} {} {}\n", std::to_string(r.m_gapsIn.value().m_top), std::to_string(r.m_gapsIn.value().m_right),
+                        std::to_string(r.m_gapsIn.value().m_bottom), std::to_string(r.m_gapsIn.value().m_left)) :
+            std::format("\tgapsIn: <unset>\n");
         const std::string gapsOut     = sc<bool>(r.m_gapsOut) ?
-                std::format("\tgapsOut: {} {} {} {}\n", std::to_string(r.m_gapsOut.value().m_top), std::to_string(r.m_gapsOut.value().m_right),
-                            std::to_string(r.m_gapsOut.value().m_bottom), std::to_string(r.m_gapsOut.value().m_left)) :
-                std::format("\tgapsOut: <unset>\n");
+            std::format("\tgapsOut: {} {} {} {}\n", std::to_string(r.m_gapsOut.value().m_top), std::to_string(r.m_gapsOut.value().m_right),
+                        std::to_string(r.m_gapsOut.value().m_bottom), std::to_string(r.m_gapsOut.value().m_left)) :
+            std::format("\tgapsOut: <unset>\n");
         const std::string borderSize  = std::format("\tborderSize: {}\n", sc<bool>(r.m_borderSize) ? std::to_string(r.m_borderSize.value()) : "<unset>");
         const std::string border      = std::format("\tborder: {}\n", sc<bool>(r.m_noBorder) ? boolToString(!r.m_noBorder.value()) : "<unset>");
         const std::string rounding    = std::format("\trounding: {}\n", sc<bool>(r.m_noRounding) ? boolToString(!r.m_noRounding.value()) : "<unset>");
--- a/src/layout/algorithm/tiled/master/MasterAlgorithm.cpp
+++ b/src/layout/algorithm/tiled/master/MasterAlgorithm.cpp
@@ -83,8 +83,8 @@ void CMasterAlgorithm::addTarget(SP<ITarget> target, bool firstMap) {
     float        lastSplitPercent   = *PMFACT;
 
     auto         OPENINGON = isWindowTiled(Desktop::focusState()->window()) && Desktop::focusState()->window()->m_workspace == PWORKSPACE ?
-                getNodeFromWindow(Desktop::focusState()->window()) :
-                getMasterNode();
+        getNodeFromWindow(Desktop::focusState()->window()) :
+        getMasterNode();
 
     const auto   MOUSECOORDS   = g_pInputManager->getMouseCoordsInternal();
     static auto  PDROPATCURSOR = CConfigValue<Hyprlang::INT>("master:drop_at_cursor");
--- a/src/managers/screenshare/ScreenshareFrame.cpp
+++ b/src/managers/screenshare/ScreenshareFrame.cpp
@@ -174,9 +174,9 @@ void CScreenshareFrame::renderMonitor() {
     g_pHyprRenderer->m_renderData.noSimplify      = true;
 
     // render monitor texture
-    CBox monbox = CBox{{}, PMONITOR->m_pixelSize}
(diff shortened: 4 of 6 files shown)


vaxerski:
rebase pls

src/render/gl/GLTexture.hpp
```diff
@@ -4,46 +4,49 @@
...
-        TEXTURE_PAR_WRAP_S = 0,
-        TEXTURE_PAR_WRAP_T,
-        TEXTURE_PAR_MAG_FILTER,
-        TEXTURE_PAR_MIN_FILTER,
-        TEXTURE_PAR_SWIZZLE_R,
-        TEXTURE_PAR_SWIZZLE_B,
-        TEXTURE_PAR_LAST,
+namespace Render::GL {
+
+    class CGLTexture : public ITexture {
```
vaxerski, line 9: base interfaces into Render:: (ITexture, IFramebuffer, etc) please

src/helpers/MonitorFrameScheduler.hpp
```diff
@@ -32,7 +34,7 @@ class CMonitorFrameScheduler {
 
     PHLMONITORREF              m_monitor;
 
-    UP<CEGLSync>               m_sync;
+    UP<Render::GL::CEGLSync>   m_sync;
```
vaxerski: Maybe Interface a Sync fence? So we don't need to depend on GL/EGL outside of GLRenderer which was the entire point.

src/render/OpenGL.cpp
```diff
@@ -60,6 +60,7 @@
 using namespace Hyprutils::OS;
```
vaxerski, line 60: can we finally get the F16 work buffer here?

This would also fix my problem of screen shaders showing up in screenshare.

src/render/GLRenderer.cpp
```diff
@@ -18,23 +18,33 @@
...
+using namespace Render;
+using namespace Render::GL;
 
 extern "C" {
 #include <xf86drm.h>
 }
 
-CHyprGLRenderer::CHyprGLRenderer() : IHyprRenderer() {}
+CHyprGLRenderer::CHyprGLRenderer() : IHyprRenderer() {
+    m_elementRenderer = makeUnique<CGLElementRenderer>();
```
vaxerski: c-t: this should be `, m_elementRenderer(...)`

src/render/GLRenderer.cpp
```diff
@@ -18,23 +18,33 @@
...
+}
+
+CHyprGLRenderer::~CHyprGLRenderer() {
+    m_elementRenderer.reset();
```
vaxerski: isnt this implicit?

src/render/OpenGL.cpp
UjinT34: Probably. There is an issue with `sdrbrightness` and `sdrsaturation` though. Both of them should be applied per texture and can't be reverted once rendered to the FB. `sdr_max_luminance` should be fine.

vaxerski: yes but once applied and rendered to sRGB doing CM again (if cm=srgb) should result in the same color as far as I understand.

I really want this here.

UjinT34: Linear colors would be affected by brightness and saturation mods. Those are meant to modify SDR -> HDR rendering. Can't do them after in "work FB -> output". And can't reverse them once they are written into the work FB. We'll have to do a separate render without the mods to fix it. Or leave it as is. Those mods don't work well anyway. And any non-hdr CM or hdr without those mods shouldn't have any issues with fp16 and export.

vaxerski: lets for now apply sdr mod into work buffer then.

Dregu:
A bunch of monitor/region screenshare issues still, mostly in CScreenshareFrame::renderMonitor:

1. damage not transformed correctly on rotated screens
2. damage not transformed or translated correctly for region capture on rotated screens
3. huge damage issues on one non-rotated perfectly ordinary screen too
4. HL crashed a few times while testing all the different capture scenarios in obs (edit2: this pr seems to expose a crash in ToplevelExportClient::captureToplevel when the window is closed before it is picked, needs some null checks somewhere)
5. noscreenshare rects are also misplaced on rotated screens

1 and 2 I dumb fixed by damaging everything to confirm it's just a damage issue. 1 and 2 are also present on main, 3 and 4 seemingly from this pr.

https://github.com/user-attachments/assets/b820b494-8eee-4101-9550-3f4805f2bef6

UjinT34:
> A bunch of monitor/region screenshare issues still, mostly in CScreenshareFrame::renderMonitor:
> 
>     1. damage not transformed correctly on rotated screens
> 
>     2. damage not transformed or translated correctly for region capture on rotated screens
> 
>     3. huge damage issues on one non-rotated perfectly ordinary screen too
> 
>     4. HL crashed a few times while testing all the different capture scenarios in obs (edit2: this pr seems to expose a crash in ToplevelExportClient::captureToplevel when the window is closed before it is picked, needs some null checks somewhere)
> 
>     5. noscreenshare rects are also misplaced on rotated screens
> 
> 
> 1 and 2 I dumb fixed by damaging everything to confirm it's just a damage issue. 1 and 2 are also present on main, 3 and 4 seemingly from this pr.
> 3.mp4

Are you sure that it's from latest version of this PR? Works for me.

Dregu:
Reflog says I was on 0fbc18c, also happening the rebase `Hyprland 0.54.0 built from branch 13631-b at commit 30e829a7ffc1981549b48340cc6c9f156ea40897 clean (c-f)`.

Here's what I'm looking at through obs pipewire, though capture method doesn't really matter: 

https://github.com/user-attachments/assets/86570bf1-64fd-4cfe-b937-ed7d99673fb6

Both non-rotated screens decided to join in on the blinking today, and the ones with the cursor trails are rotated.

finalfantasian:
This PR breaks screensharing from me. Windows and elements/parts of windows flicker in and out of existence with screen capture, and window capture is too dark. It is to be noted that this is all with fp16 enabled and cm set to hdr, though.

Edit: Things flickering in and out of existence happens with both fp16 and hdr off.

Edit: Kitty's native transparency also looks acidic with this PR.

UjinT34:
> This PR breaks screensharing from me. Windows and elements/parts of windows flicker in and out of existence with screen capture, and window capture is too dark. It is to be noted that this is all with fp16 enabled and cm set to hdr, though.
> 

Not so easy to reproduce on my side. Looks like it happens only when there are no window with blur effect on the screen (e.g. disabling blur triggers it consistently).

finalfantasian:
Screensharing no longer has stuff that flickers in and out of existence but it seems like the tonemapping is oversaturated/acidic in the screen/window capture compared to the actual window/screen. Makes it too dark and/or saturated in scenes where there's supposed to be proper contrast/leveling.

UjinT34:
> Screensharing no longer has stuff that flickers in and out of existence but it seems like the tonemapping is oversaturated/acidic in the screen/window capture compared to the actual window/screen. Makes it too dark and/or saturated in scenes where there's supposed to be proper contrast/leveling.

Screensharing stuff is WIP. `render:use_fp16`, SDR/HDR source and SDR/HDR target will yield different results in different combinations atm. `sdrbrightness` and `sdrsaturation` shouldn't affect screenshare. Combinations with HDR and `render:use_fp16 = 0` might remain unfixed. Probably should just force fp16 when HDR is active.

UjinT34:
Should be more or less correct.
`monitorv2:cm` with some sdr settings and `render:cm_fs_passthrough`, `render:non_shader_cm`, `render:use_fp16`, and `render:keep_unmodified_copy` have some quirks in some combinations. `monitorv2:cm = hdr` or `monitorv2:cm = hdredid` setups are the most broken ones and will need to tweak other settings to choose between correct screensharing and correct fullscreen rendering. In most common cases everything should work fine. Needs more testing with HDR content in SDR mode, HDR/SDR content in HDR mode and toggling fullscreen while recording.
OBS doesn't seem to destroy the recording session meaning once the recording started it'll remain in "while recording" state even after OBS exit.
Any further screensharing fixes should be done in some other PR.

finalfantasian:
this has fixed it for me, except for window capturing where the problem still exists. i wonder why that is.

Edit: Hold on; it seems like output capture still has the issue and is overly dark/acidic.

Edit: It seems to happen specifically with HDR content, for example mpv playback with HDR videos.

rivera-diego:
## Blur Issue on Vertical Monitor with `new_optimizations = true` (NVIDIA RTX 3060 Ti)

Hi, thanks for working on this PR! 👋

I'm experiencing an issue with the blur effect on a **vertical monitor** when using `new_optimizations = true`. Even with the fixes from this PR, the problem persists.

---

### System Information

- **GPU:** NVIDIA GeForce RTX 3060 Ti (GA104)
- **Driver:** NVIDIA proprietary (latest)
- **Hyprland-Git Version:** 0.54.0 (built from `main` branch, commit `64a2e4e26`)
- **Date:** March 23, 2026
- **Session:** Wayland

---

### Monitor Configuration

```ini
monitor = HDMI-A-1, 1920x1080@75, 0x0, 1, transform, 1
monitor = DP-1, 1920x1080@144, 1080x396, 1
```

- **HDMI-A-1:** Vertical monitor (rotated 90°, `transform, 1`)
- **DP-1:** Horizontal monitor (primary, 144Hz)

---

### Blur Configuration

```ini
decoration {
    blur {
        enabled = true
        size = 6
        passes = 2
        ignore_opacity = true
        new_optimizations = true  # ← Issue occurs with this enabled
        xray = false
        vibrancy = 0.2
    }
}
```

---

### Issue Description

When `new_optimizations = true` is enabled:

- The blur effect on the **vertical monitor** renders the background **as if it were in horizontal orientation**
- The blurred background appears **rotated incorrectly** (90° offset)
- This affects **any window with blur** on the vertical monitor (terminals, rofi, etc.)

When `new_optimizations = false`:

- Blur renders correctly on the vertical monitor
- Background orientation matches the actual monitor rotation

---

### Environment Variables (NVIDIA)

```ini
env = LIBVA_DRIVER_NAME,nvidia
env = XDG_SESSION_TYPE,wayland
env = __GLX_VENDOR_LIBRARY_NAME,nvidia
env = NVD_BACKEND,direct
env = __GL_SHADER_DISK_CACHE_SKIP_CLEANUP,1
env = __GL_YIELD,USLEEP
```

---

### Video Attachment

📎 **[Attach your video here]** showing:
1. Blur with `new_optimizations = true` (incorrect rotation)
2. Blur with `new_optimizations = false` (correct rendering)

---

### Additional Notes

- The original fix from PR #9210 seemed to work previously, but the issue reappeared in recent builds
- This might be related to how the damage tracking or framebuffer handling works with NVIDIA + `new_optimizations`
- Happy to test any additional patches or provide more logs if needed!

---

Let me know if I can help with testing! 🙏

https://github.com/user-attachments/assets/a30d3434-fc27-4fc3-aca6-47dc86f655b4

UjinT34:
Try now. I am not sure that the fix is correct. It probably inverts a rotation error somewhere else. And I am pretty sure similar bug was fixed before, might be some rebase issue.

umbrageodotus:
There's likely many rebase issues. I'll find the cause asap so you can remove that rubberband fix.

vaxerski:
do we still have issues with this one?

UjinT34:
https://github.com/hyprwm/Hyprland/discussions/13836 should be retested

ItsOhen:
Part 8 breaks includes for plugins. So i guess this one does as well.

src/helpers/MonitorResources.hpp
```diff
@@ -0,0 +1,45 @@
...
+            CTimer                   lastUsed;
+        };
+
+        SP<Render::IFramebuffer>            m_monitorMirrorFB;
+        WP<CMonitor>                        m_monitor;
+        DRMFormat                           m_drmFormat;
+        Vector2D                            m_size;
+        NColorManagement::PImageDescription m_imageDescription;
+
+        std::vector<SResource>              m_workBuffers;
```
vaxerski, line 41: I don't really understand the logic here. Why is there not just one work buffer?

UjinT34: offload, mirror, mirrorSwap and similar are there. no point in having separate names for them especially when they are used for something else than the name implies. like shadow deco did with mirror fbs. removes the code duplications and ensures that we don't create and keep more fbs than really needed

vaxerski: right, thanks.

finalfantasian:
Yo so this broke fullscreen 4k videos on mpv and Brave for me. Perhaps it's something to do with scaling? idk. Used to work on an earlier build w/ this PR. it just does the following whenever i fullscreen something of that resolution in Brave and mpv (strangely not in ffplay though):

Edit: Seems on Brave it happens regardless of resolution.

https://github.com/user-attachments/assets/48f569f5-7cd1-416b-816c-0a1de05373f5

Edit 2: The culprit seems to be forcing no hardware cursors, which makes the cursor when it appears a black square instead of a cursor.

Edit 2: Seems I no longer have the issue in mpv after setting `render:cm_sdr_eotf` to `default`. Brave still freaks out and becomes a black screen, though.

rivera-diego:
@UjinT34  Thanks bro for the changes, looks like the issues with OBS, image capture, and my problem with the background blur are already fixed. In fact, I even feel like the image looks clearer now.

PR #13632, opened 2026-03-07
Refactor of the workspace tracking to work better with multi monitor setups

Mr-Stoneman:
This PR fixes the issues described in https://github.com/hyprwm/Hyprland/discussions/13138
The issues described there are:
- Freezing on multi monitor setups
- `previous_per_monitor` not working as expected

The cause of the freezing seems to be a reference error in `CWorkspaceHistoryTracker::previousWorkspace`
```Cpp
auto& data = dataFor(ws);
while (true) {

    // case 1: previous exists
    if (data.previous) {
        if (data.previous->m_monitor != restrict) {
            data = dataFor(data.previous.lock());
            continue;
        }

        break;
    }

    // case 2: previous doesnt exist, but we have mon
    if (data.previousMon) {
        if (data.previousMon != restrict)
            return nullptr;

        break;



    }

    // case 3: no mon and no previous
    return nullptr;
}
```
`data = dataFor(data.previous.lock());` doesn't assign a new reference to data but overwrites the value of data. This can lead to loops.

The cause of the faulty  `previous_per_monitor` was that a workspace could only have one previous workspace.

To fix this I switched to a approach using lists and returning values instead of references.

Only the workspace dispatcher has a  `previous_per_monitor` implementation, `focusworkspaceoncurrentmonitor` doesn't have one. This isn't super important and doesn't really fit into this PR.

PS: hope all the formatting is correct.

### DIFF
--- a/src/desktop/history/WorkspaceHistoryTracker.cpp
+++ b/src/desktop/history/WorkspaceHistoryTracker.cpp
@@ -66,7 +66,7 @@ void CWorkspaceHistoryTracker::gc() {
 
 const CWorkspaceHistoryTracker::SHistoryEntry CWorkspaceHistoryTracker::previousWorkspace(PHLWORKSPACE ws) {
     gc();
-    auto it = std::find_if(m_history.begin(), m_history.end(), [&](const auto& entry) { return entry.workspace == ws; });
+    auto it = std::ranges::find_if(m_history, [&](const auto& entry) { return entry.workspace == ws; });
 
     // If the workspace is found in history, the previous one is simply the next element down the timeline
     if (it != m_history.end() && std::next(it) != m_history.end())
@@ -91,7 +91,7 @@ const CWorkspaceHistoryTracker::SHistoryEntry CWorkspaceHistoryTracker::previous
 
     gc();
 
-    auto it = std::find_if(m_history.begin(), m_history.end(), [&](const auto& entry) { return entry.workspace == ws; });
+    auto it = std::ranges::find_if(m_history, [&](const auto& entry) { return entry.workspace == ws; });
 
     // Start looking from the element immediately following `ws` in the list
     if (it != m_history.end())


vaxerski:
write some tests please

Mr-Stoneman:
To reproduce the error, this requires to be run on bare metal and a multi monitor setup to be reproduced can this even be reliably tested?

Mr-Stoneman:
Done

src/desktop/history/WorkspaceHistoryTracker.hpp
```diff
@@ -39,14 +38,12 @@ namespace Desktop::History {
             WORKSPACEID     workspaceID   = WORKSPACE_INVALID;
         } m_lastWorkspaceData;
 
-        std::vector<SWorkspacePreviousData> m_datas;
-
-        void                                track(PHLWORKSPACE w);
-        void                                gc();
-        void                                setLastWorkspaceData(PHLWORKSPACE w);
+        std::list<SHistoryEntry> m_history;
```
vaxerski: I dont think there's a single instance in programming where a list is a good idea. Why not a vec or deque?

Mr-Stoneman: The problem with vec is, that it is only possible to push to the back which would invert the whole logic. This makes it less intuitive and harder to read for me. But deque seems like it could work so i will look into that.

Mr-Stoneman: Deque seems to work without any changes.
The main advantage of a list over a deque I can currently see, is the faster deletion of element that are in the middle of the list during `gc()`.
The advantage of using a deque is that you are able to access elements at a random position. Which we currently don't really use.
But if you prefer a deque I don't mind changing it.

vaxerski: deque is faster to iterate, removing is a rare op.

Mr-Stoneman: Didn't know that. I changed it.

Mr-Stoneman:
I looked into the `testMultimonBAF()` test failure. It seems like this test fails because there is no application open on Workspace 4 so it gets removed by`GC()`.

Mr-Stoneman:
The failed test in `hyprtester/src/tests/clients/pointer-warp.cpp@157` seems to be unrelated to me. My changes shouldn't touch the code related to this

src/desktop/history/WorkspaceHistoryTracker.cpp
```diff
@@ -31,111 +31,83 @@ CWorkspaceHistoryTracker::CWorkspaceHistoryTracker() {
...
 
-            break;
-        }
-
-        // case 2: previous doesnt exist, but we have mon
-        if (data.previousMon) {
-            if (data.previousMon != restrict)
-                return nullptr;
+    // Start looking from the element immediately following `ws` in the list
+    if (it != m_history.end()) {
```
vaxerski: no {} around short if/else

Mr-Stoneman: Sry missed that. Is fixed now.

hyprtester/src/tests/main/workspaces.cpp
```diff
@@ -220,6 +249,7 @@ static void testMultimonBAF() {
     }
 
     OK(getFromSocket("/dispatch workspace 4"));
+    Tests::spawnKitty();
```
vaxerski: why are you changing an old test?

src/desktop/history/WorkspaceHistoryTracker.cpp
```diff
@@ -31,111 +31,82 @@ CWorkspaceHistoryTracker::CWorkspaceHistoryTracker() {
...
+    // Start looking from the element immediately following `ws` in the list
+    if (it != m_history.end())
+        it++;
+    else
+        it = m_history.begin();
 
-            break;
+    // Scan down the timeline until we hit a workspace mapped to the restricted monitor
+    while (it != m_history.end()) {
+        if (it->monitor == restrict) {
```
vaxerski: missseeed

src/desktop/history/WorkspaceHistoryTracker.cpp
```diff
@@ -31,111 +31,82 @@ CWorkspaceHistoryTracker::CWorkspaceHistoryTracker() {
...
+const CWorkspaceHistoryTracker::SHistoryEntry CWorkspaceHistoryTracker::previousWorkspace(PHLWORKSPACE ws) {
     gc();
+    auto it = std::find_if(m_history.begin(), m_history.end(), [&](const auto& entry) { return entry.workspace == ws; });
 
-    for (const auto& d : m_datas) {
-        if (d.workspace != ws)
-            continue;
-        return &d;
+    // If the workspace is found in history, the previous one is simply the next element down the timeline
+    if (it != m_history.end() && std::next(it) != m_history.end()) {
```
vaxerski: cmon man

hyprtester/src/tests/main/workspaces.cpp
Mr-Stoneman: Because it failed as said in the message above

Mr-Stoneman: And because I never revived a response I decided to fix it so it doesn't fail again I can always revert the change

src/desktop/history/WorkspaceHistoryTracker.cpp
Mr-Stoneman: Thought you wanted me to only change the lines mentioned in your previous comment. Thought you looked through the code and that was the only one you found.

vaxerski: no, I just didnt think I need to point out with a finger each one xd

Mr-Stoneman: Just saw the code you mentioned and didn't even think about checking the rest. But it should be fixed now

Mr-Stoneman:
Seems like all the tests passed this time. The formatting errors are not related to my changes.

hyprtester/src/tests/main/workspaces.cpp
vaxerski: I dont think it should fail. back_and_forth shouldnt rely on a window being there. It used to work, if it now doesn't, that's a regression

Mr-Stoneman: In this case, I think we could use `m_lastWorkspaceData`, because this wont be affected by `gc()`. Under which condition should this be returned? Should `const CWorkspaceHistoryTracker::SHistoryEntry CWorkspaceHistoryTracker::previousWorkspace(PHLWORKSPACE ws)` be change to always return the value from `m_lastWorkspaceData`?

vaxerski: well probably not because you want per-monitor to work.

Mr-Stoneman: This will lead to always being able to go back to empty screens when using `hyprctl dispatch workspace previous` is this the intended behavior?

vaxerski: yes, back and forth should bring you to the previous workspace even if empty

Mr-Stoneman: Only back and forth or also `workspace previous`

vaxerski: that too yes

src/desktop/history/WorkspaceHistoryTracker.cpp
```diff
@@ -31,111 +31,92 @@ CWorkspaceHistoryTracker::CWorkspaceHistoryTracker() {
...
-        if (data.previous) {
-            if (data.previous->m_monitor != restrict) {
-                data = dataFor(data.previous.lock());
-                continue;
-            }
+    gc();
 
-            break;
-        }
+    auto it = std::find_if(m_history.begin(), m_history.end(), [&](const auto& entry) { return entry.workspace == ws; });
```
vaxerski: use ranges::

src/desktop/history/WorkspaceHistoryTracker.cpp
```diff
@@ -31,111 +31,92 @@ CWorkspaceHistoryTracker::CWorkspaceHistoryTracker() {
...
+        // Add monitor to seen monitors
+        monitorCounts.emplace_back(entry.monitor);
+        return false;
+    });
 }
 
-const CWorkspaceHistoryTracker::SWorkspacePreviousData* CWorkspaceHistoryTracker::previousWorkspace(PHLWORKSPACE ws) {
+const CWorkspaceHistoryTracker::SHistoryEntry CWorkspaceHistoryTracker::previousWorkspace(PHLWORKSPACE ws) {
     gc();
+    auto it = std::find_if(m_history.begin(), m_history.end(), [&](const auto& entry) { return entry.workspace == ws; });
```
vaxerski: use ranges::

Mr-Stoneman:
Ty for your feedback and patience. Happy to have this merged.

PR #13641, opened 2026-03-07
Fix crash with nullptr FBs

UjinT34:
Should fix #13639

Ready


moetayuko:
can comfirm that this fixes crash for 5700xt

WhySoBad:
Same here, with this patch Hyprland no longer crashes on my laptop with a UHD Graphics 620 iGPU

PR #8899, opened 2024-12-30
hyprpm: add glaze dependency FetchContent fallback

davc0n:
Use FetchContent to retrieve glaze dependency if not available with find_package.
Allows to build hyprpm w/o installing glaze at system level (on some distros is not available in official repositories).

Inspired from glaze documentation (ref. https://github.com/stephenberry/glaze?tab=readme-ov-file#see-this-example-repository-for-how-to-use-glaze-in-a-new-project and https://github.com/stephenberry/glaze_example/blob/main/CMakeLists.txt), let me know if there's a better way to achieve that.

PR #13666, opened 2026-03-09
build: add glaze dependency with FetchContent fallback

davc0n:
Fixes this build error:
```
hyprland/start/src/helpers/Nix.cpp:13:10: fatal error: glaze/glaze.hpp: No such file or directory
   13 | #include <glaze/glaze.hpp>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
```

Same has been done for hyprpm, see https://github.com/hyprwm/Hyprland/pull/8899

>Use FetchContent to retrieve glaze dependency if not available with find_package.
>Allows to build hyprpm w/o installing glaze at system level (on some distros is not available in official repositories).
>
>Inspired from glaze documentation (ref. https://github.com/stephenberry/glaze?tab=readme-ov-file#see-this-example-repository-for-how-to-use-glaze-in-a-new-project and https://github.com/stephenberry/glaze_example/blob/main/CMakeLists.txt), let me know if there's a better way to achieve that.

**EDIT**

It seems that build on nix fails with `error: could not find git for clone of glaze-populate`
I have read https://github.com/hyprwm/Hyprland/discussions/13043 but tbh I have no idea about how to fix it (dunno nix).

### DIFF
--- a/hyprpm/CMakeLists.txt
+++ b/hyprpm/CMakeLists.txt
@@ -11,7 +11,7 @@ set(CMAKE_CXX_STANDARD 23)
 
 pkg_check_modules(hyprpm_deps REQUIRED IMPORTED_TARGET tomlplusplus hyprutils>=0.7.0)
 
-find_package(glaze 7 QUIET)
+find_package(glaze 7...<8 QUIET)
 if (NOT glaze_FOUND)
     set(GLAZE_VERSION v7.2.0)
     message(STATUS "hyprpm: glaze dependency not found, retrieving ${GLAZE_VERSION} with FetchContent")
--- a/start/CMakeLists.txt
+++ b/start/CMakeLists.txt
@@ -11,7 +11,7 @@ find_package(PkgConfig REQUIRED)
 
 pkg_check_modules(starthyprland_deps REQUIRED IMPORTED_TARGET hyprutils>=0.10.3)
 
-find_package(glaze 7 QUIET)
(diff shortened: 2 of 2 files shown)


vaxerski:
@fufexan whats up with nix here

fufexan:
Probably 7.1.1 isn't in nixpkgs and it's trying to download it instead, which is not possible in the Nix sandbox.
Why not require minimum version 7 instead?

davc0n:
It seems to work.

I've changed `find_package` to match any version >= 7 (that should've fixed the nix issue).
If not found `FetchContent` retrieves `v7.2.0` (tag must be matched here).

Moreover I've changed the log message a little bit just to distinguish `start` and `hyprpm`:
```
-- start: glaze dependency not found, retrieving v7.2.0 with FetchContent
-- hyprpm: glaze dependency not found, retrieving v7.2.0 with FetchContent
```

vaxerski:
> Probably 7.1.1 isn't in nixpkgs and it's trying to download it instead, which is not possible in the Nix sandbox.
Why not require minimum version 7 instead?

if anything we should require 7.x.x, 8.x might not build whenever it comes out

davc0n:
@vaxerski agree, now should be ok (ref. https://cmake.org/cmake/help/latest/command/find_package.html#find-package-version-format)

vaxerski:
ye looks fine @fufexan

PR #13674, opened 2026-03-09
Fix: monitor reconnect crashes

matjam:

Fix three distinct crashes triggered by monitor disconnect/reconnect (e.g. turning a monitor off and back on, or sleep/wake cycles).

**Crash 1: ABRT in layout recalculation (std::bad_variant_access)**

`arrangeMonitors()` emits a `layoutChanged` signal that triggers `recalculate()` on every workspace. Workspaces whose monitors have just disconnected have null/expired monitor weak pointers. Dereferencing these produces undefined behavior that manifests as std::bad_variant_access.

Fix: Add null guards for workspace and monitor at three points in the recalculation chain:

- `CSpace::m_geomUpdateCallback` - skip if workspace/monitor is gone
- `CAlgorithm::recalculate()` - null check on workspace
- `SScrollingData::recalculate()` and `usableArea()` - validate the full weak pointer chain

**Crash 2: SEGV in session lock surface creation**

When a screen locker (hyprlock/swaylock) requests a lock surface for a monitor that has already been destroyed, the `CSessionLockSurface` constructor dereferences `m_monitor` to register a signal listener outside the existing null guard. Additionally, `onGetLockSurface` doesn't validate the return from `CWLOutputResource::fromResource()`. This is a Wayland protocol race — the client sends the request before it processes wl_registry.global_remove.

Fix: Move the modeChanged listener registration inside the `if (m_monitor)` block, and validate both the output resource and monitor in `onGetLockSurface()` before proceeding.

**Crash 3: SEGV in blur framebuffer access during layer surface teardown**

When a layer surface (e.g. waybar) is destroyed during monitor disconnect, `makeSnapshot()` uses `beginFullFakeRender` with `simple=true`, which skips framebuffer allocation. The monitor's `m_blurFB` has already been reset to null during disconnect, so `getBlurTexture()` dereferences a null shared pointer.

Fix: Guard all `m_blurFB` dereference sites against null in `getBlurTexture()` (both `IHyprRenderer` and `CHyprGLRenderer`), `preBlurForCurrentMonitor()`, the blur optimization path in `renderElement()`, and `renderRectWithBlurInternal()`. The callers already handle null returns correctly.

Replicated the issues locally by turning off the connected monitor and turning it back on.

Used Claude to identify the bugs and help create the patch.

Verified the fixes work for me.

They're pretty simple null pointer guards.

### DIFF
--- a/src/render/GLRenderer.cpp
+++ b/src/render/GLRenderer.cpp
@@ -394,6 +394,8 @@ void CHyprGLRenderer::draw(CTextureMatteElement* element, const CRegion& damage)
 };
 
 SP<ITexture> CHyprGLRenderer::getBlurTexture(PHLMONITORREF pMonitor) {
+    if (!pMonitor->m_blurFB)
+        return nullptr;
     return pMonitor->m_blurFB->getTexture();
 }
 
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1029,7 +1029,8 @@ void CHyprOpenGLImpl::renderRectWithBlurInternal(const CBox& box, const CHyprCol
     CRegion damage{g_pHyprRenderer->m_renderData.damage};
     damage.intersect(box);
 
-    auto POUTFB = data.xray ? g_pHyprRenderer->m_renderData.pMonitor->m_blurFB->getTexture() : g_pHyprRenderer->blurMainFramebuffer(data.blurA, &damage);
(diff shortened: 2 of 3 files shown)


matjam:
I think at least some of the complaints I'm seeing for nvidia are in part due to this crash. I think its _some_ monitors treat nvidia dpms diferently than AMD? I don't know its weird. Its been a problem for a while - finally had some time to track it down. Be good to see if people can replicate this issue on AMD as well. I suspect its an nvidia only thing. Anyway, the null guards can't hurt.

PR #13677, opened 2026-03-10
renderer: alloc blurfb upfront and invalidate it

gulafaran:
alloc blurfb upfront and invalidate color attachment in endrender.


vaxerski:
but that needs to be persistent?

gulafaran:
> but that needs to be persistent?

well i think not. preblurformonitor already called glclear(0..), and it only needs to be "persistent" to the end of frame. after that we can invalidate it and let it begin again. as it currently does, only now it avoids that unneeded internal driver copy because we invalidate it, and also slow clearing. but i guess it requires some blur fanatic testing, seems to work locally.

vaxerski:
no, preblur is only called once in a while, and we reuse the blur. That's the entire point of the optimizations

gulafaran:
> no, preblur is only called once in a while, and we reuse the blur. That's the entire point of the optimizations

oh ok makes sense, just checked with breakpoints, im really not hitting this path beyond once at startup anyhow, probably why im not seeing any difference. be it for the settings im using or its bugged idk.

PR #13694, opened 2026-03-11
view: consolidate group flags and apply window rules

ChaosInfinited:
Hi,
noticed a regression since the Groups refactor (https://github.com/hyprwm/Hyprland/discussions/13460), after some poking I think I've finally found the solution, albeit I'm not sure if I'm doing it properly, sorry, I'm new to this codebase!

I've built it and tested it locally and it seemingly works, the group rules are being applied again.

As now the code's using bitmasks and a flags holder, the m_locked and m_deny became redundant (I've tried setting them with the bitmasks, but that wouldn't work for some reason..?).

Wasn't able to test it properly with hyprtester unfortunately, had some luck running it using Xephyr, but it'd fail on tests that'd assume I have kitty terminal emulator installed etc., not sure how to run it properly, wouldn't work from a fresh TTY either.. but ctest passed.

### DIFF
--- a/hyprtester/src/tests/main/groups.cpp
+++ b/hyprtester/src/tests/main/groups.cpp
@@ -294,6 +294,89 @@ static bool test() {
     Tests::killAllWindows();
     EXPECT(Tests::windowCount(), 0);
 
+    // Tests for grouping/merging logic
+    NLog::log("{}Testing locked groups w/ invade", Colors::GREEN);
+
+    Tests::killAllWindows();
+    EXPECT(Tests::windowCount(), 0);
+
+    // Test normal, unlocked groups
+    {
+      auto winA = Tests::spawnKitty("unlocked");
+      if (!winA) {
+          NLog::log("{}Error: unlocked kitty did not spawn", Colors::RED);
+          return false;
+      }
+      OK(getFromSocket("/dispatch togglegroup"));
+
+      auto winB = Tests::spawnKitty("top");
+      if (!winB) {
+          NLog::log("{}Error: top kitty did not spawn", Colors::RED);
+          return false;
+      }
+
+      // Verify it DID merge into a group
+      {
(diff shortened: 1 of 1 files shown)


vaxerski:
can you please write some tests for this? :)

kaosuran:
> can you please write some tests for this? :)

I will try later, but I can't promise anything, hyprtester isn't working right for me :/ will try running it under a VM, maybe it'll stop crashing like crazy.

kaosuran:
> can you please write some tests for this? :)

Hello again, sorry it took so long, I've added 3 tests:
1. Check behavior for regular groups, make sure they merge normally
2. Check behavior for locked groups, make sure they don't merge
3. Check behavior for locked groups with invade rule, make sure they do merge

Testing locally I got:
```Testing locked groups w/ invade
Passed: Tests::windowCount(). Got 0
Passed: getFromSocket("/dispatch togglegroup"). Got ok
Passed: str contains "at: 22,22" 2 times.
Passed: Tests::windowCount(). Got 0
Passed: getFromSocket("/dispatch togglegroup"). Got ok
Passed: getFromSocket(std::format("/dispatch focuswindow pid:{}", lockedWin->pid())). Got ok
Passed: getFromSocket("/dispatch lockactivegroup lock"). Got ok
Passed: str contains "at: 22,22" 1 times.
Passed: Tests::windowCount(). Got 0
Passed: getFromSocket("/keyword windowrule[locked-im]:match:class ^locked|invade$"). Got ok
Passed: getFromSocket("/keyword windowrule[locked-im]:group set always lock invade"). Got ok
Passed: str contains "at: 22,22" 2 times.
Passed: Tests::windowCount(). Got 0
```

The clang CI seemingly fails due to the main upstream changes 🙏🏻 
Please let me know if there's something else I should do, thanks for your time!

PR #13696, opened 2026-03-11
scrolling: add direction parameter to promote dispatcher

oblivion8282-1337:
## What

Adds an optional direction argument (`l` / `r`) to the scrolling layout's `promote` dispatcher.

## Motivation

Currently `promote` always moves the focused window into a new column to the **right**. There is no way to promote it to the left — useful for workflows where a window should be placed before the current position rather than after.

## Behavior

| Command | Effect |
|---|---|
| `layoutmsg promote` | new column to the right (unchanged default) |
| `layoutmsg promote r` | new column to the right (explicit) |
| `layoutmsg promote l` | new column to the left (new) |

## Example keybinds

```
bind = SUPER, A, layoutmsg, promote r
bind = SUPER SHIFT, A, layoutmsg, promote l
```

## Changes

- `src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp`: check for optional second argument in the `promote` branch; use `add(idx - 1)` for left-promotion (inserts the new column before the current one)

### DIFF
--- a/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
+++ b/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
@@ -1356,8 +1356,15 @@ std::expected<void, std::string> CScrollingAlgorithm::layoutMsg(const std::strin
         if (!TDATA)
             return std::unexpected("no window focused");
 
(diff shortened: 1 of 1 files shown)


oblivion8282-1337:
Closing to address PR guidelines compliance. Will resubmit properly.

PR #13700, opened 2026-03-11
xwayland: pipe through monitor in coordinate mapping

AndreasBackx:
> Note, this was largely possible thanks to AI as I'm not very familiar with the codebase. I have however given this more than a simple glance to make sure it's actually not more slop. I've also fully tested it locally, more information below as this is my own understanding (with help from AI).

## Summary

Fix XWayland input issues on vertically offset multi-monitor layouts by using an X11 window's known monitor for coordinate conversions instead of relying on the nearest-monitor heuristic. This resolves a real-world case where XWayland apps rendered correctly on the center monitor but hover/click targeting broke on the lower portion of the screen. The fix keeps X11 reported positions, restored positions, and unmanaged X11 geometry conversions anchored to the correct monitor.

It highlights that there might be a problem in the nearest-monitor heuristic, but it seems like to show this more clearly, it would need testing cases in the code there which doesn't exist (yet). ~~That can be done in a follow-up PR most likely as this approach should already avoid recomputing the monitor from coordinates when it's already known.~~ https://github.com/hyprwm/Hyprland/pull/13702

## High-Level Problem

On some multi-monitor layouts, XWayland apps could render correctly but receive incorrect hover/click input on one monitor.

In my setup:

- left monitor: portrait 4K, scale `1.25`
- center monitor: landscape 4K, scale `1.25`
- right monitor: portrait 4K, scale `1.25`
- the center monitor was vertically centered relative to the side monitors

A simplified version of the affected logical layout looked like this:

```text
Wayland logical layout

+--------------+                            +--------------+
|              |                            |              |
|              |+--------------------------+|              |
|     DP-1     ||          DP-2            ||     DP-3     |
|  portrait    ||       landscape          ||  portrait    |
|    9:16      ||         16:9             ||    9:16      |
|              ||                          ||              |
|              |+--------------------------+|              |
|              |                            |              |
|              |                            |              |
+--------------+                            +--------------+
```

The visible cursor moved normally, and XWayland windows rendered normally, but hover/click targeting was wrong on the center monitor. This was reproducible with apps like Steam.

## Root cause

Hyprland maintains a separate synthetic XWayland coordinate system. The bug was not in the visible hover coordinate math itself, but in how some XWayland/Wayland coordinate round-trips selected a monitor.

The affected conversions used a generic nearest-monitor heuristic:

- `waylandToXWaylandCoords(...)`
- `xwaylandToWaylandCoords(...)`

That heuristic can choose the wrong monitor on boundary-aligned, vertically offset layouts.

In the failing case, the top-left of the center monitor round-tripped like this:

```text
Expected round-trip for center monitor origin

XWayland DP-2 origin  ->  Wayland center origin  ->  XWayland DP-2 origin
(2160, 0)            ->  (1728, 865)            ->  (2160, 0)
```

What actually happened before this change:

```text
Broken round-trip before fix

XWayland DP-2 origin  ->  Wayland center origin  ->  XWayland DP-1 space
(2160, 0)            ->  (1728, 865)            ->  (2160, 1081.25)
```

So a real coordinate on the center monitor was being converted back through the left monitor's XWayland space. That mismatch was enough to break XWayland hover/click behavior even though rendering still looked correct.

## Why this happens

The current generic conversion path only knows "a coordinate" and tries to infer which monitor it belongs to.

That works for many layouts, but it is fragile when:

- monitor boundaries touch exactly in logical space
- monitors have different shapes/orientations
- one monitor is vertically offset from its neighbors
- the same real coordinate is valid at a boundary where multiple monitor rectangles are plausible candidates

In contrast, X11 windows usually already know which monitor they belong to.

This patch uses that information instead of re-guessing.

## What this changes

This patch switches several X11 window-related conversion sites to prefer the window's known monitor:

- reported X11 position (`realToReportPosition`)
- restoring XWayland geometry back to compositor space (`xwaylandPositionToReal`)
- unmanaged/override-redirect X11 geometry (`unmanagedSetGeometry`)

Conceptually, the change is:

```text
Before:
window real pos -> nearest-monitor heuristic -> XWayland coords

After:
window real pos -> window's known monitor    -> XWayland coords
```

And similarly for XWayland -> Wayland conversions.

## Why this fixes the bug

For an X11 window already associated with the center monitor, conversions now stay anchored to that monitor:

```text
After fix

Center window
  real position <-> XWayland position
       always uses DP-2
```

That prevents boundary points from accidentally round-tripping through the left portrait monitor's synthetic XWayland space.

In the affected layout, this restores correct hover/click targeting on the center monitor without requiring any monitor layout compromise.

### DIFF
--- a/src/managers/XWaylandManager.cpp
+++ b/src/managers/XWaylandManager.cpp
@@ -180,7 +180,7 @@ Vector2D CHyprXWaylandManager::waylandToXWaylandCoords(const Vector2D& coord) {
 Vector2D CHyprXWaylandManager::waylandToXWaylandCoords(const Vector2D& coord, PHLMONITOR preferredMonitor) {
     static auto PXWLFORCESCALEZERO = CConfigValue<Hyprlang::INT>("xwayland:force_zero_scaling");
 
-    PHLMONITOR pMonitor = preferredMonitor;
+    PHLMONITOR  pMonitor = preferredMonitor;
     if (!pMonitor) {
         double bestDistance = __FLT_MAX__;
         for (const auto& m : g_pCompositor->m_monitors) {
@@ -217,7 +217,7 @@ Vector2D CHyprXWaylandManager::xwaylandToWaylandCoords(const Vector2D& coord, PH
 
     static auto PXWLFORCESCALEZERO = CConfigValue<Hyprlang::INT>("xwayland:force_zero_scaling");
 
-    PHLMONITOR pMonitor = preferredMonitor;
+    PHLMONITOR  pMonitor = preferredMonitor;
     if (!pMonitor) {
         double bestDistance = __FLT_MAX__;
         for (const auto& m : g_pCompositor->m_monitors) {


src/managers/input/InputManager.cpp
```diff
@@ -193,7 +193,12 @@ void CInputManager::sendMotionEventsToFocused() {
 
     m_emptyFocusCursorSet = false;
 
-    g_pSeatManager->setPointerFocus(Desktop::focusState()->surface(), getMouseCoordsInternal().floor() - BOX->pos());
+    auto local = getMouseCoordsInternal().floor() - BOX->pos();
```
vaxerski: what is this doing here...?

AndreasBackx: My bad, this shouldn't have been included. It was from an earlier investigation.

vaxerski:
what is this MR's relation to #13702 ? those seem like conflicting MRs?

AndreasBackx:
Sorry for the delay in reply. 

> what is this MR's relation to #13702 ? those seem like conflicting MRs?

I'm somewhat used to a stacked diffs workflow where changes can stack on other changes and that makes reviewing easier, but that's not really feasible on GitHub ([though they do seem to be working on supporting it](https://xcancel.com/jaredpalmer/status/2019817235163074881)).

#13702 is this PR/MR with some additional commits on top. If/when this PR gets accepted and merged, I can rebase that PR on main again and it'll be easier to review. This makes it so I don't have to have 2 parallel branches that have merge conflicts but I also understand that it's not convenient for you or other contributors.

vaxerski:
yeah that's fine just mention it next time xD

vaxerski (dismissed):

vaxerski:
oops c-f fails, please rebase and run clang format

PR #13702, opened 2026-03-11
xwayland: fix monitor selection wrt monitor ownership

AndreasBackx:
> Note, this was largely possible thanks to AI as I'm not very familiar with the codebase. I have however given this more than a simple glance to make sure it's actually not more slop. I've also fully tested it locally, more information below as this is my own understanding (with help from AI).

_This is a follow-up to https://github.com/hyprwm/Hyprland/pull/13700._

## Summary
- Fix XWayland monitor selection on offset multi-monitor layouts by choosing monitor ownership in the coordinate space of the input point.
- Add regression coverage for boundary and round-trip behavior on vertically offset monitor layouts.
- Prevent generic XWayland/Wayland conversions from selecting the wrong monitor when `force_zero_scaling`-style transformed extents overlap neighboring logical monitor areas.

Before: https://youtu.be/3JcibU4FV-E
After: https://youtu.be/3ceYMdOJySE

## Problem

On some offset multi-monitor layouts, XWayland coordinate conversions could pick the wrong monitor during generic Wayland <-> XWayland monitor selection.

```text
Wayland logical layout

+--------------+                            +--------------+
|              |                            |              |
|              |+--------------------------+|              |
|     DP-1     ||          DP-2            ||     DP-3     |
|  portrait    ||       landscape          ||  portrait    |
|    9:16      ||         16:9             ||    9:16      |
|              ||                          ||              |
|              |+--------------------------+|              |
|              |                            |              |
|              |                            |              |
+--------------+                            +--------------+
```

A point that was logically the top-left corner of the center monitor could be interpreted as belonging to the left monitor during generic Wayland-space selection.
This happened because the old heuristic used transformed/effective extents when selecting a monitor for a Wayland-space point. With `force_zero_scaling`-style geometry, those transformed extents can overlap neighboring monitors' logical regions, making the wrong monitor appear to contain the point.

That led to broken round-trips such as:
- XWayland center point -> correct Wayland point
- Wayland point -> wrong XWayland monitor space

<details>
<summary>Example</summary>

Left:   position (0, 0)
Center: position (1728, 865)
Right:  position (4800, 0)

The problematic point was the center monitor origin in Wayland space:
P = (1728, 865)

That point is logically the top-left corner of the center monitor. However, with force_zero_scaling-style transformed extents, the old heuristic effectively compared it against regions like this:

Wayland-space ownership with transformed extents (old behavior)
```
Left transformed region:
x: [0 -------------------------------- 2160)
y: [0 --------------------------------------------------- 3840)
Center logical region:
x:                     [1728 ---------------------------- 4800)
y:             [865 ----------- 2593)
Point P = (1728, 865)
         ^
         | falls inside both regions
```
So the same Wayland-space point could be treated as belonging to:
- the center monitor logically
- the left monitor according to transformed fallback extents
That ambiguity caused a bad round-trip:
```
Broken round-trip
XWayland center origin
(2160, 0)
    |
    v
Wayland
(1728, 865)
    |
    v
generic monitor selection picks Left
    |
    v
XWayland
(2160, 1081.25)   <- wrong monitor space / wrong Y
```

</details>

## Fix

The fix changes the generic monitor-selection logic to use the coordinate space of the input:
- for Wayland-space points:
  - select monitor ownership using logical Wayland monitor rectangles (`monitor.size`)
- for XWayland-space points:
  - select monitor ownership using XWayland strip extents (`effectiveMonitorSize(...)`)

This avoids treating transformed XWayland extents as if they were valid Wayland-space ownership regions.

Logically, the change is:
- before:
  - choose monitor using transformed/effective extents even for Wayland-space ownership
- after:
  - choose monitor in the same space as the input point, then convert

This restores stable round-trips and fixes the offset-layout ambiguity.

<details>
<summary>Example</summary>

Old logic
```
Wayland point
    |
    v
select monitor using transformed/effective extents
    |
    v
wrong monitor can win on offset layouts
```

New logic
```
Wayland point
    |
    v
select monitor using logical Wayland rectangles
    |
    v
correct monitor ownership
    |
    v
convert to XWayland
```

For the same problematic point (this shows y pivoted onto the same plane):
```
Left logical region:
x: [0 ---------------- 1728)
y: [0 --------------------------- 3072)
Center logical region:
x:                          [1728 ---------------------------- 4800)
y:      [865 ----------- 2593)
Point P = (1728, 865)
         ^
         | belongs only to Center with half-open logical rects
```
So the round-trip becomes stable again:
```
XWayland center origin
(2160, 0)
    |
    v
Wayland
(1728, 865)
    |
    v
generic monitor selection picks Center
    |
    v
XWayland
(2160, 0)
```
This fixes the ambiguity without changing the overall XWayland coordinate-system design.

</details>

## Tests
This adds regression tests covering:

- exact boundary ownership
- corner and gap behavior
- round-trips across offset layouts
- the reproduced offset-layout misclassification case

The regression branch demonstrates the failure first, and this fix branch makes that regression pass: 
https://gist.github.com/AndreasBackx/b3d529864fcca1fecc8f78cf86d7d32e


src/managers/XWaylandManager.cpp
```diff
@@ -174,63 +176,44 @@ void CHyprXWaylandManager::setWindowFullscreen(PHLWINDOW pWindow, bool fullscree
 }
 
 Vector2D CHyprXWaylandManager::waylandToXWaylandCoords(const Vector2D& coord) {
+    return waylandToXWaylandCoords(coord, nullptr);
```
vaxerski: method can be extracted and removed from XWaylandMnagaer, we are trying to get rid of it because its a shit name for shit things

src/managers/XWaylandMonitorSelection.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+struct SXWaylandMonitorDesc {
+    Vector2D waylandPosition;
+    Vector2D xwaylandPosition;
+    Vector2D size;
+    Vector2D transformedSize;
+    float    scale = 1.F;
+};
+
+std::optional<size_t> selectMonitorForWaylandPoint(std::span<const SXWaylandMonitorDesc> monitors, const Vector2D& point, bool forceZeroScaling);
```
vaxerski: firstly needs a namespace (`XWayland::` is probably fine)

secondly why all these variants? Why not expose the template?

Also, SXWaylandMonitorDesc is redundant. You can create dummy PHLMONITOR entries in the tests.

src/managers/XWaylandMonitorSelection.cpp
```diff
@@ -0,0 +1,170 @@
+#include "XWaylandMonitorSelection.hpp"
```
vaxerski: wrong place: probably belongs in src/xwayland

AndreasBackx:
Thank you for the reviews. I've tried my best to try and gauge on what you were looking for as I'm not a great C++ engineer in the slightest. Let me know if the new changes are satisfactory and what you'd like to see changed. Don't be afraid to be verbose or straightforward. 😅️

src/layout/target/WindowTarget.cpp
```diff
@@ -257,8 +258,10 @@ std::expected<SGeometryRequested, eGeometryFailure> CWindowTarget::desiredGeomet
     requested.size = clampSizeForDesired(DESIRED_GEOM.size());
 
     if (m_window->m_isX11) {
+        static auto PXWLFORCESCALEZERO = CConfigValue<Hyprlang::INT>("xwayland:force_zero_scaling");
+
         Vector2D xy    = {DESIRED_GEOM.x, DESIRED_GEOM.y};
-        xy             = g_pXWaylandManager->xwaylandToWaylandCoords(xy);
+        xy             = XWayland::xwaylandToWaylandCoords(std::span<const PHLMONITOR>{g_pCompositor->m_monitors}, xy, *PXWLFORCESCALEZERO);
```
vaxerski: I think the explicit cast to span<> is redundant?

src/xwayland/MonitorSelection.cpp
```diff
@@ -0,0 +1,119 @@
+#include "MonitorSelection.hpp"
+
+#include "../helpers/MiscFunctions.hpp"
+
+namespace XWayland {
```
vaxerski: don't do this in .cpp files (style)

`using namespace XWayland`, then member methods are `XWayland::...`

AndreasBackx:
Thanks again for the reviews, I resolved the latest ones in the latest commit. As mentioned in https://github.com/hyprwm/Hyprland/pull/13700#issuecomment-4184484295, we can either merge #13700 first and then rebase this one and merge it, or just merge this one and close the other as its changes are included here. I will leave that decision up to you, just let me know and I can do the rebase if you choose to go with separate merging.

AndreasBackx:
PR has been rebased and clang formatted. I realised I am running a clang-format 21 instead of 22 from Arch Linux so had to do some manual changes as I couldn't be bothered to find out how to get 22 easily on Fedora at 2AM.

I might've accidentally cherry picked in some of my old commits below the actual changes, I'm uncertain whether it was included before or not as I'm tired. Regardless, it does not change the final state of the commit as their implementation was replaced. 

Thanks again for reviewing and merging the last PR and hopefully this one.

PR #13715, opened 2026-03-12
blur: fix dim blur in HDR

gusjengis:
Full disclosure, I've never worked in this codebase before, so to lower my barrier of entry, I used opencode with GPT 5.4 for this fix. Afterwards, I reviewed what it changed and removed all changes that didn't seem to be necessary.

With cm=hdr enabled, blurring the wallpaper would also dim it by a lot. I experienced this bug on this version:
Hyprland 0.54.1 built from branch v0.54.1-b at commit 4b07770b9ef1cceb2e6f56d33538aaffb9186b9c clean ([gha] Nix: update inputs).
Date: 2026-03-03

Some images...

#### HDR
image
image

#### SDR
image
image

#### HDR
image
image

#### SDR
image
image

Excuse the pictures instead of screenshots, I have yet to find a program that properly takes screenshots in SDR. I have my own workaround where my screenshot utility automatically switches to SDR when taking screenshots, but then that would hide the bug. You can't tell in these images, but after my fix, the colors don't change at all when blurring in HDR, same as SDR

As for what was changed to fix this (GPT): 
Before, the blur pipeline converted the work buffer to SDR, applied blur, then converted the result back into the monitor/output color space. In HDR, that caused the blurred image to lose brightness.
My fix is to keep the blur pipeline in the work buffer’s native color-managed format instead of forcing it through that SDR conversion step. For HDR outputs, the intermediate blur buffers are also allocated in a floating-point HDR format, so blur is applied without losing HDR brightness/precision.

While working on this, I hit a bug with some wild artifacts on the right and bottom edge of the screen.

Removing this part of the patch brings those artifacts back.
image

Here are a couple images of the artifacts:
image
image

According to GPT:
The edge artifacts were caused by the HDR blur intermediate buffers retaining stale pixel data outside the current damage region. Since blur is done with scissored/damage-limited passes, pixels near the screen edge could pick up those uncleared texels and produce visible color artifacts. The fix is to fully clear both HDR blur ping-pong framebuffers before running the blur passes, with scissor disabled and blending off, so the blur only samples clean data.

I'll admit I don't understand every detail, but after reviewing the code myself, GPT's descriptions of what changed seem to be accurate. 

This functions and doesn't seem to affect anything else. Someone who has more context about the code that was modified should probably check this. Assuming my changes don't have any non-obvious side effects, this PR's readiness depends on your taste.

### DIFF
--- a/src/render/GLRenderer.cpp
+++ b/src/render/GLRenderer.cpp
@@ -375,6 +375,7 @@ void CHyprGLRenderer::draw(CTexPassElement* element, const CRegion& damage) {
             .roundingPower      = m_data.roundingPower,
             .discardActive      = m_data.discardActive,
             .allowCustomUV      = m_data.allowCustomUV,
+            .sourceIsWorkBufferCM = m_data.sourceIsWorkBufferCM,
             .cmBackToSRGB       = m_data.cmBackToSRGB,
             .cmBackToSRGBSource = m_data.cmBackToSRGBSource,
             .discardMode        = m_data.ignoreAlpha.has_value() ? sc<uint32_t>(DISCARD_ALPHA) : m_data.discardMode,
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1038,8 +1038,8 @@ void CHyprOpenGLImpl::renderRectWithBlurInternal(const CBox& box, const CHyprCol
     g_pHyprRenderer->pushMonitorTransformEnabled(true);
     const auto SAVEDRENDERMODIF               = g_pHyprRenderer->m_renderData.renderModif;
     g_pHyprRenderer->m_renderData.renderModif = {}; // fix shit
-    renderTexture(POUTFB, MONITORBOX,
-                  STextureRenderData{.damage = &damage, .a = data.blurA, .round = data.round, .roundingPower = 2.F, .allowCustomUV = false, .allowDim = false, .noAA = false});
+    renderTexture( POUTFB, MONITORBOX,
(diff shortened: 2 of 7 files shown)


gusjengis:
@UjinT34 Blame tells me you modified some adjacent code 5 days ago, you're probably as good a person to review this as any. I've seen you a lot around here, have to assume you have write access.

ssareta:
Probably conflicts with #13631

umbrageodotus:
Yeah, please mark this as a draft until the API fixes PR is merged, then rebase.

UjinT34:
I'll need a config to test it with #13631. Should be correct there unless I missed some setting.

gusjengis:
> I'll need a config to test it with #13631. Should be correct there unless I missed some setting.

I'll build and run that version with my config and display and see if blur is fixed. Building now...

gusjengis:
Same problem, the special workspace blur also dims like crazy. 

### Desktop
image

### special:terminal

image

### hyprctl version

image

P.S. Whoops, tested on the wrong branch. Have since tested the right one and blur is fixed after enabling render { use_f16 = true }, but sdrbrightness is now ignored.

gusjengis:
I'll just do a PR to your refactor fork. Unless that's super illegal, haven't really read the guidelines.

UjinT34:
Should be ok with #13631 and `render:use_fp16`. This PR does about the same for blur alone in a complicated way that might mess with some other scenarios. The description does look reasonable though. Maybe gpt isn't best to produce the actual code.
Blur with HDR was broken from the start. It goes through CM too many times and the resulting texture isn't rendered correctly. Same with SDR but that incorrectness is unnoticeable or even zero just by luck.

gusjengis:
Just seeing this, I'll try that setting.

gusjengis:
Thanks @UjinT34 for fixing this

PR #13720, opened 2026-03-13
Be more selective about how we expand the window box here so that we'…

SGumbles:
This fixes the issue I describe in https://github.com/hyprwm/Hyprland/pull/13144#issuecomment-4052780766

TL;DR we don't want to just blindly expand the window box because the hit box will overlap with neighbouring boxes. When the boxes overlap that means there are several windows that will pass this test, which means that it's up to the ordering in m_windows which one passes this test (race condition).

The original issue describes not being able to use border_grab_area when gap_out > 0, so this will allow the windows boxes to safely expand into the gap_out regions.

Ready for merging.


Dregu:
Yeah the original fix should've clearly been an outer gap specific thing, since inner gaps already count as window area anyway and there's no way to expand them from both sides.

src/Compositor.cpp
```diff
@@ -1036,8 +1036,23 @@ PHLWINDOW CCompositor::vectorToWindowUnified(const Vector2D& pos, uint8_t proper
             if (!w->m_isFloating && w->m_isMapped && w->workspaceID() == WSPID && !w->isHidden() && !w->m_X11ShouldntFocus && !w->m_ruleApplicator->noFocus().valueOrDefault() &&
                 w != pIgnoreWindow && !isShadowedByModal(w)) {
                 CBox box = (properties & Desktop::View::USE_PROP_TILED) ? w->getWindowBoxUnified(properties) : CBox{w->m_position, w->m_size};
-                if ((properties & Desktop::View::INPUT_EXTENTS) && BORDER_GRAB_AREA > 0 && !w->isX11OverrideRedirect())
-                    box.expand(BORDER_GRAB_AREA);
+                if ((properties & Desktop::View::INPUT_EXTENTS) && BORDER_GRAB_AREA > 0 && !w->isX11OverrideRedirect()) {
+                    if (getWindowInDirection(w, Math::eDirection::DIRECTION_LEFT) == nullptr) {
```
vaxerski: supremely inefficient. Use geometry and compare it against the work area of the space we're on.

SGumbles: Ok I've created a small helper closure to test if the window's edge is close to the edge of the workspace.

C0Florent:
I just rebuilt latest main (bbf6718, which does include this PR), and somehow i'm still having this issue. I might look into it in more detail later today

PR #13726, opened 2026-03-14
Fix json output for the submap command

jhnberg:

The JSON output of the `submap` command is not accepted by tools for parsing JSON e.g. jq or the json python module. This is a fix to allow these tools to parse the JSON output.

Since this commit changes the output of `hyprctl submap -j` it technically break compatibility if anything is consuming the current format. However, since the current JSON output for this command is not recognized by e.g. `jq` the impact may be limited. Since the output of the submap is simple, the existing usage may parse this as-is.

I also noticed seem to be getting pretty inconsistent results from `../build/hyprtester/hyprtester --plugin ./plugin/hyprtestplugin.so`. This is both on this commit and `origin/main`. But I don't think this commit had any impact here.

Ready. This is a very trivial one line change.

### DIFF
--- a/hyprtester/src/tests/main/hyprctl.cpp
+++ b/hyprtester/src/tests/main/hyprctl.cpp
@@ -173,6 +173,13 @@ static bool testGetprop() {
     return true;
 }
 
+static void testSubmap() {
+    NLog::log("{}Testing hyprctl submap", Colors::GREEN);
+
+    EXPECT(getCommandStdOut("hyprctl submap"), "default\n");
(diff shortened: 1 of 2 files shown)


Dregu:
Maybe we should just use glaze to generate all json, since humans keep screwing it up...

vaxerski:
we probably could

jhnberg:
Thank you for your feedback. I have added a new test for the `hyprctl` to test the json parsing using `jq`.

PR #13730, opened 2026-03-14
pin: support workspace selectors for scoped pinning

karol-broda:
Extends the pin window rule to accept workspace IDs so a window can be sticky only on specific workspaces. pin alone still works as before. You can pass individual IDs or ranges:
```
windowrule = pin 1 3, class:^(myapp)$
windowrule = pin 1-4, class:^(myapp)$
```
Also adds pinnedWorkspaces to hyprctl clients output (empty = pinned to all workspaces).

hyprctl clients output format changes -- new pinnedWorkspaces field in JSON, new line in plain text. Anything parsing that output by position will break.

The pin dispatching action (pinActive) still toggles without workspace scope, it just clears pinnedWorkspaces on unpin. Not sure if that's the right behavior or if it should be extended too.

Parts of the test file were written with AI assistance (claude code with opus 4.6) and reviewed by me

Needs review, especially around the workspace switching logic in Monitor.cpp and whether the pinActive keybind should also support workspace arguments

### DIFF
--- a/hyprtester/src/tests/main/pin.cpp
+++ b/hyprtester/src/tests/main/pin.cpp
@@ -1,0 +1,299 @@
+#include "tests.hpp"
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
+#include "../shared.hpp"
+#include <thread>
+#include <chrono>
+
+static int ret = 0;
+
+static bool test() {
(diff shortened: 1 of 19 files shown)


vaxerski (commented):
fundamental problem, what about named? Maybe a workspace selector would work better?

karol-broda:
good point, refactored to use workspace selectors instead of raw ids

PR #13756, opened 2026-03-16
renderer: fix crash when shader path isn't a file

nikitax44:

Hyprland crashes if `decoration:screen_shader` is set to a directory path instead of a file.

https://github.com/hyprwm/Hyprland/blob/30c498acf4173930dfd8afd6279ebceacb1941e9/src/render/OpenGL.cpp#L892-L899

```shell
$ hyprctl keyword decoration:screen_shader '/'
```

I implemented a check for the file type.

I do not know the C++ idiomatic way to check for this kind of issues, so I used `fstat`.

Should be good to merge

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -889,10 +889,21 @@ void CHyprOpenGLImpl::applyScreenShader(const std::string& path) {
     if (path.empty() || path == STRVAL_EMPTY)
         return;
 
-    std::ifstream infile(absolutePath(path, g_pConfigManager->getMainConfigPath()));
+    std::string     absPath = absolutePath(path, g_pConfigManager->getMainConfigPath());
+
+    std::error_code ec;
+    if (!std::filesystem::is_regular_file(absPath, ec)) {
+        if (ec)
+            g_pConfigManager->addParseError("Screen shader parser: Failed to check screen shader path: " + ec.message());
+        else
+            g_pConfigManager->addParseError("Screen shader parser: Screen shader path is not a regular file");
+        return;
+    }
+
+    std::ifstream infile(absPath);
 
     if (!infile.good()) {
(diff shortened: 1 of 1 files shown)


chrisn731:
Seems like you may be looking for:

https://en.cppreference.com/w/cpp/filesystem/is_regular_file.html

nikitax44:
Updated the code to use `is_regular_file`. The issue here is that this version in theory can still be crashed by TOCTOU abuse.

src/render/OpenGL.cpp
```diff
@@ -889,13 +889,19 @@ void CHyprOpenGLImpl::applyScreenShader(const std::string& path) {
...
-    std::ifstream infile(absolutePath(path, g_pConfigManager->getMainConfigPath()));
+    std::string   absPath = absolutePath(path, g_pConfigManager->getMainConfigPath());
+    std::ifstream infile(absPath);
 
     if (!infile.good()) {
         g_pConfigManager->addParseError("Screen shader parser: Screen shader path not found");
         return;
     }
 
+    if (!std::filesystem::is_regular_file(absPath)) {
```
vaxerski: this can throw. Add a `std::error_code ec` and pass it to the fn, then check that !ec

src/render/OpenGL.cpp
```diff
@@ -889,10 +889,22 @@ void CHyprOpenGLImpl::applyScreenShader(const std::string& path) {
     if (path.empty() || path == STRVAL_EMPTY)
         return;
 
-    std::ifstream infile(absolutePath(path, g_pConfigManager->getMainConfigPath()));
+    std::string     absPath = absolutePath(path, g_pConfigManager->getMainConfigPath());
+
+    std::error_code ec;
+    if (!std::filesystem::is_regular_file(absPath, ec)) {
+        if (ec) {
```
vaxerski: no {} around short if/else here and below

vaxerski (approved):
tnx

PR #13763, opened 2026-03-17
Fix popup crash with nullptr mon

UjinT34:
Fixes https://github.com/hyprwm/Hyprland/discussions/13708

No idea why `g_pCompositor->getMonitorFromID(m_layerOwner->m_layer)` return nullptr. Was this way before my changes, might be incorrect.

Ready


vaxerski:
uhhh because it makes no sense? This is some massive rebase fail somewhere. Monitor should be ls->m_monitor

PR #13764, opened 2026-03-17
input: fix touch screen focus on multi monitor

Dregu:
#### Describe your PR, what does it fix/add?

Fixes touch refocus using cursor position instead of touch position for monitor focus, which made it impossible to e.g. touch LS on non-focused monitor.

- Fixes https://github.com/hyprwm/Hyprland/discussions/12221
- Older forgotten pr that just seems more complicated https://github.com/hyprwm/Hyprland/pull/12646

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

Nope, seems like a pretty simple logic error to me that overridePos was used for mouseCoords but mouseCoords wasn't used for focus.

#### Is it ready for merging, or does it need work?

Ready.

PR #13773, opened 2026-03-18
Focus monitor on touch down events

khalid151:


vaxerski (approved):

Dregu:
What was the point of this again?

khalid151:
> What was the point of this again?

Using the touchscreen didn't put the monitor in focus. Now it does.

Dregu:
> Using the touchscreen didn't put the monitor in focus. Now it does.

Yeah I got that part but what does doing that fix, specifically?

khalid151:
> > Using the touchscreen didn't put the monitor in focus. Now it does.
> 
> 
> 
> Yeah I got that part but what does doing that fix, specifically?

Intended behavior. If you have multiple monitor setup, how are you gonna focus the monitor?

When you disable `mouse_move_focuses_monitor`, clicking does focus the monitor the cursor is on. Similarly, you'd expect touch down events to do the same.

Dregu:
No *you'd* expect it to do the same, I'd expect monitor focus to stay where my mouse cursor is and just send touch events to surfaces on the other monitor. So sounds like things like this need *options*...

I was asking cause I also have a PR that *actually* fixes some issues I have with touch on multimonitor. Since touch still uses the *cursor* position for picking target surfaces, but does not move the cursor (**and shouldn't**), it doesn't work anyway if the cursor is on a different monitor. At least on layers.

#13764

PR #13781, opened 2026-03-19
screencopy: fix crash in screensharing toplevel with invalid handle

ssareta:

When an app uses hyprland-toplevel-export-v1 for window capture, (e.g. pipewire with obs), trying to screencopy a destroyed window causes nullptr deref in `CScreenshareSession::CScreenshareSession` and crashes hyprland. 

Adds a check against this. Fixes #13777.

[hyprland_toplevel_export_manager_v1](https://wayland.app/protocols/hyprland-toplevel-export-v1#hyprland_toplevel_export_manager_v1) does not provide a signal (e.g. "failed" or similar) to let the client know that it failed to start the toplevel capture session (e.g. because invalid window handle is passed), so I just return. OBS seems happy with this, but not sure if this is the most correct thing to do...

I also added some guards in `CScreenshareSession::CScreenshareSession` variants. These are redundant with (downstream of) the above fix, but hyprland supports 3 different screencopy protos so I figure there's not much harm in checking and preventing a crash, in case there's some weird path that passes removed monitor or window.  

Ready

### DIFF
--- a/src/managers/screenshare/ScreenshareSession.cpp
+++ b/src/managers/screenshare/ScreenshareSession.cpp
@@ -9,10 +9,16 @@
 using namespace Screenshare;
 
 CScreenshareSession::CScreenshareSession(PHLMONITOR monitor, wl_client* client) : m_type(SHARE_MONITOR), m_monitor(monitor), m_client(client) {
+    if UNLIKELY (!m_monitor)
+        return;
+
     init();
 }
 
 CScreenshareSession::CScreenshareSession(PHLWINDOW window, wl_client* client) : m_type(SHARE_WINDOW), m_window(window), m_client(client) {
+    if UNLIKELY (!m_window)
+        return;
+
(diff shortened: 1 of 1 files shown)


Dregu:
For me this path only started crashing after pulling https://github.com/hyprwm/Hyprland/pull/13631 somehow, but anyway yes null checks are in order.

I don't think the whole toplevel proto needs to stick around for much longer, as image-copy-capture supports the same thing. Someone just needs to port xdph to image-copy-capture...

vaxerski:
> For me this path only started crashing after pulling https://github.com/hyprwm/Hyprland/pull/13631 somehow, but anyway yes null checks are in order.
I don't think the whole toplevel proto needs to stick around for much longer, as image-copy-capture supports the same thing. Someone just needs to port xdph to image-copy-capture...

at claude pls fix

PR #13783, opened 2026-03-19
debug: print log when window is mapped

B2krobbery:
This adds a simple debug output when a window is mapped in `CWindow::onMap()`.

It helps track window lifecycle events during debugging, especially for focus and stacking related issues.

Tested locally by building Hyprland from source and verifying output when launching applications.

This change does not modify any functional behavior.

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -2,7 +2,6 @@
 #include <ranges>
 #include <hyprutils/animation/AnimatedVariable.hpp>
 #include <re2/re2.h>
--
 #include "Group.hpp"
 #include <iostream>
 


B2krobbery:
Hi, I noticed the clang-format check is failing for multiple files unrelated to my change.

This PR only adds a debug log in `CWindow::onMap()` and does not modify formatting elsewhere.

Please let me know if I should adjust anything on my side.

vaxerski:
ye no we know

src/desktop/view/Window.cpp
```diff
@@ -620,6 +620,7 @@ void CWindow::onMap() {
     m_movingFromWorkspaceAlpha->resetAllCallbacks();
 
     m_movingFromWorkspaceAlpha->setValueAndWarp(1.F);
+    std::cout << "Window mapped: " << m_title << std::endl;
```
vaxerski, line 623: bro what the fuck

B2krobbery: Got it, thanks for the feedback.

This was mainly for local debugging, so I understand it’s not suitable for inclusion. I’ll close the PR.

Appreciate your time.

Issue #2598, opened 2023-06-28
Firefox Picture in Picture goes into fullscreen when changing video


FirelightFlagboy:
## Version

Hyprland, built from branch  at commit 51a930f802c71a0e67f05e7b176ded74e8e95f87 dirty ().
Tag: v0.26.0 (using the AUR package)

xdg-desktop-portal-hyprland: 0.4.0-1

## Steps to reproduce

1. Using `firefox-114.0.2`
2. Playing a youtube playlist
3. Toggle `Picture-in-Picture` (later called PIP) mode for the video
4. Move to the end of the current playing video
5. Wait the video to end
6. Youtube will now start playing the next video
7. At that time the `Picture-in-Picture` goes into fullscreen

## Expected outcome

When Youtube play the next video, the `PIP` don't go into fullscreen.

## Noted outcome

When Youtube play the next video, the `PIP` goes into fullscreen.

Here is the output of `hyprctl clients` for the PIP client

```txt
Window 564aa00097f0 -> Picture-in-Picture:
	mapped: 1
	hidden: 0
	at: 1102,42
	size: 3796,1556
	workspace: 5 (5)
	floating: 1
	monitor: 0
	class: firefox
	title: Picture-in-Picture
	initialClass: firefox
	initialTitle: Picture-in-Picture
	pid: 1184
	xwayland: 0
	pinned: 0
	fullscreen: 1
	fullscreenmode: 1
	fakefullscreen: 0
	grouped: 0
	swallowing: 0
```

Before the next video play, here is the output of `hyprctl clients`:

```txt
Window 564aa06bcb80 -> Picture-in-Picture:
	mapped: 1
	hidden: 0
	at: 1404,-8
	size: 2766,1556
	workspace: 5 (5)
	floating: 1
	monitor: 0
	class: firefox
	title: Picture-in-Picture
	initialClass: firefox
	initialTitle: Picture-in-Picture
	pid: 1184
	xwayland: 0
	pinned: 0
	fullscreen: 0
	fullscreenmode: 0
	fakefullscreen: 0
	grouped: 0
	swallowing: 0
```

What change between the two, is the `fullscreen` & `fullscreenmode` are set to `1`.

## Custom window rule for PIP

```
windowrulev2 = float, class:(firefox), title:(Picture-in-Picture)
```

I've configured the PIP to always float

## A possible workaround

[...]

Issue #1047, opened 2022-11-19
Some windows disapear when an app goes fullscreen


DiedByDisgust:
Steps to reproduce:
put some windows in a workspace then execute an app or game (world of warcraft running on wine in my case) . in fullscreen mode.

Noted outcome:
when you close app/game or move it to a different workspace the prior mentioned windows go hidden or something. hyprctl clients still shows them in the same workspace

Expected outcome:
Not get disappeared i guess

DiedByDisgust:
under latest release v0.18.0beta
Hyprland, built from branch main at commit c02bfc3897137439596df4670cef2441d24c8367dirty (fix extents in shadow deco).

vaxerski:
not able to repro. Any app that isnt a 20GB game?

DiedByDisgust:
Yes, also happen with LibreOffice Writer. but isn't the exactly same behavior.

vaxerski:
I have NEVER seen that with LO - how to repro?

diced:
I had the same issue, it happened only when the full screen game opened on top of all the other windows. In my case I launched Fall Guys, and I had heroic games launcher open on the same workspace and it became invisible after I moved Fall Guys to another workspace.

Sincide:
Having the same issue when launching Dying Light 2 via steam

vaxerski:
fixed in be03a6186c22f915d3d92df71d2e74c64e816245

Sincide:
Hi, I still have this issue, some games reappear when fiddling with the fullscreen to not fullscreen, but not Dying Light 2, it just dissapears and you hear the music in the background.

Version:
Hyprland, built from branch main at commit be03a6186c22f915d3d92df71d2e74c64e816245dirty (fix invisible windows on moving fullscreen out).

vaxerski:
This issue is about the tiled windows going invisible, not the fullscreen one. Open another issue for that.

felixsanz:
@vaxerski can you re-open it? it's still happening. i'm new to hyprland and this is happening to me all the time :/

vaxerski:
I need a reproducible example on something that aint a game

felixsanz:
> 

[...]

PR #13790, opened 2026-03-19
desktop/view: ignore client maximize requests for tiled windows

charisvt:
**What does this PR fix?**
After entering and then exiting fullscreen in a HTML5 video in a Chromium-based browser, the browser window expands to cover all other tiled windows on the workspace.

**Why does this fix work?**
Hyprland sends MAXIMIZED state to all windows at first commit to suppress client-side decorations.
Chromium records this state and sends set_maximized when exiting fullscreen to "restore" what it thinks was its prior state. 
For tiled windows, honoring this request enters FSMODE_MAXIMIZED, making the window fill the entire workspace. 
Since tiled windows are compositor-managed, client maximize requests have no meaning and are now ignored for non-floating windows.            

**How was this tested?**
Manually verified with Brave browser in a nested Hyprland session:
opened a YouTube video, entered and exited fullscreen, confirmed the browser returned to its tiled half-screen geometry with the adjacent terminal window still visible.
Window sizes also verified via hyprctl.


vaxerski:
I mean isnt this what the suppressevent maximize rule by default is for

charisvt:
The suppressevent maximize rule works, but it's a per-app workaround for a compositor-caused issue.
Hyprland sends MAXIMIZED to all windows at first commit to suppress CSD -- Chromium observes that state and echoes set_maximized on fullscreen exit.

This has surfaced repeatedly (#9171, #2598, #1047).
The expectation is that tiled windows stay tiled.
Ignoring client maximize requests for non-floating windows makes that the default rather than something each user has to configure per app.

Skidamek:
Thank you charisvt, this PR also fixes https://github.com/hyprwm/Hyprland/discussions/13355

Dregu:
And it is definitely time to remove the whole maximize hack behind a setting, xdg tiled state has existed for years and should be used instead if client supports xdg v2...

Vladimir-csp:
> And it is definitely time to remove the whole maximize hack behind a setting, xdg tiled state has existed for years and should be used instead if client supports xdg v2...

This. AFAIK, GTK respects tiled states and does not draw out of frame decorations. Option is good.

On a related note, there's [GTK-NoCSD](https://codeberg.org/MorsMortium/GTK-NoCSD), hacky but works.

vaxerski:
> And it is definitely time to remove the whole maximize hack behind a setting, xdg tiled state has existed for years and should be used instead if client supports xdg v2...

We submitted a fix proposal to wayland-protocols but guh-nome decided we're too stupid™ to understand the gnome vision™

Dregu:
> We submitted a fix proposal to wayland-protocols but guh-nome decided we're too stupid™ to understand the gnome vision™

I forget what the protocol issue was, but what I described is what sway has been doing for years, don't know what prevents hyprland from doing it... https://github.com/swaywm/sway/commit/f2425b4fd64314394aa962cee12b617fb7e5b10f

vaxerski:
sway is a static environment with zero animations thats the main thing

charisvt:
To summarize: the suppressevent workaround exists but requires per-user, per-app configuration for a compositor-caused issue. This PR makes the correct behavior the default. The xdg-tiled refactor is a separate long-term concern. Happy to rebase or adjust if needed.

vaxerski:
my main concern is CSD that sends maximize, and that not working.

charisvt:
I understand your concerns, I will come back with a more precise solution to fix this behavior while also allowing legitimate CSD maximize requests.

Skidamek:
Can this get merged? I've been running this PR on my setup and haven't hit any regressions - it fixes a genuinely frustrating issue. I believe your previous concern @vaxerski was also addressed.

vaxerski (approved):
lgtm

PR #13797, opened 2026-03-20
fix(xwayland): prevent potential buffer overflow in socket path handling

B2krobbery:
This patch replaces unsafe strncpy usage with bounded copies using sizeof(addr.sun_path) to prevent potential buffer overflow.

Explicit null termination is also ensured for normal socket paths.

No functional changes intended — existing socket behavior is preserved.

Tested:
- Hyprland builds successfully
- XWayland starts correctly
- X clients (e.g. xclock) launch without issues


vaxerski:
dw c-f is fucked

PR #13819, opened 2026-03-23
input: fix the multimon touch fix

Dregu:

Had some kind of brainfart and thought `mouse` was actually set by default when used by the mouse functions in #13764. This way we should definitely only affect touch input, since touch is the only thing that uses overridePos. Although my own plugin also used it for fake mouse pos, hmmmm...

Much simpler fix might just be setting `mouse = true` by default, but that does affect some other logic and I really don't know how. I would prefer that if I only knew the repercussions.

Someone might want to split this function in the future so touch could only call the part that finds target surfaces for coords and skips the pointer related things.

- Fixes #13815

At least the gamers look happy again and my touch stuff still works as well

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -261,7 +261,7 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus, bool mouse, st
         g_pCompositor->scheduleFrameForMonitor(PMONITOR, Aquamarine::IOutput::AQ_SCHEDULE_CURSOR_MOVE);
 
     // constraints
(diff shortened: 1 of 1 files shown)


lichie567:
I just tested this change and it does seem to fix the issue.

Astyyyyy:
can confirm this fixes issue for me as well

SimoneFelici:
Hi, this also fix my issue (https://github.com/hyprwm/Hyprland/pull/13764#issuecomment-4106459088).
Thank you!

vaxerski (approved):

PR #13822, opened 2026-03-23
fix(sd-daemon): ensure null termination of socket path after strncpy

B2krobbery:
Ensures proper null termination of unixAddr.sun_path after bounded strncpy.

Although addrLen is limited using strnlen, strncpy does not guarantee null termination when the source length matches the limit.

This patch explicitly adds null termination to prevent potential undefined behavior in edge cases.

No functional changes intended.

Tested:
- Build succeeds
- No regression in systemd socket notification

### DIFF
--- a/src/helpers/SdDaemon.cpp
+++ b/src/helpers/SdDaemon.cpp
@@ -40,11 +40,12 @@ int NSystemd::sdNotify(int unsetEnvironment, const char* state) {
 
     struct sockaddr_un unixAddr = {0};
 
-    size_t             addrLen = strnlen(addr, sizeof(unixAddr.sun_path) - 1);
-
     unixAddr.sun_family = AF_UNIX;
-    strncpy(unixAddr.sun_path, addr, addrLen);
-    unixAddr.sun_path[addrLen] = '\0'; // ensure null termination
+
+    const size_t addrLen = strnlen(addr, sizeof(unixAddr.sun_path) - 1);
+    memcpy(unixAddr.sun_path, addr, addrLen);
+    unixAddr.sun_path[addrLen] = '\0';
+
     if (unixAddr.sun_path[0] == '@')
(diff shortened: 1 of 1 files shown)


src/helpers/SdDaemon.cpp
```diff
@@ -44,6 +44,7 @@ int NSystemd::sdNotify(int unsetEnvironment, const char* state) {
 
     unixAddr.sun_family = AF_UNIX;
     strncpy(unixAddr.sun_path, addr, addrLen);
+    unixAddr.sun_path[addrLen] = '\0'; // ensure null termination
```
staticssleever668: this comment is pointless

staticssleever668 (changes requested):
unixAddr is null-initialized right above. this doesn't do anything.

B2krobbery:
Even with zero-initialization, strncpy can overwrite up to addrLen bytes without guaranteeing a trailing null byte when the source length reaches the limit.

This makes the null-termination implicit and dependent on input length. The explicit assignment ensures consistent behavior regardless of how much is written.

C0Florent:
> Even with zero-initialization, strncpy can overwrite up to addrLen bytes without guaranteeing a trailing null byte when the source length reaches the limit.

Then just pass a shorter length to `strncpy` so that it doesn't overwrite the last null byte?

B2krobbery:
Good point.

I considered limiting the length passed to strncpy, but decided to replace it with memcpy using the computed length and explicit null termination instead.

This avoids relying on strncpy semantics and makes the behavior more explicit and predictable.

umbrageodotus:
Relying on strncpy semantics here is correct. Please change it.

B2krobbery:
Got it, thanks for the clarification.
Since the current implementation is already correct due to zero-initialization, this change doesn’t provide any benefit. I’ll close this PR.

PR #13823, opened 2026-03-23
fix(renderer): guard against null monitor in renderMonitor

B2krobbery:
Adds a null check for  pMonitor  at the beginning of  renderMonitor.

renderMonitor  is called with  m_monitor.lock()  in several places, which can return  nullptr  if the weak pointer has expired. The function currently dereferences  pMonitor  without validation, which could lead to a null pointer dereference.

No functional changes intended. This is a defensive fix to prevent potential crashes.

Tested by:
- Building Hyprland successfully
- Running a full session
- Locking and unlocking the screen
- Opening and interacting with applications


B2krobbery:
Thanks! Glad it helps. I’ll keep an eye out for similar cases.

Issue #12485, opened 2025-11-27
Attempt to handle GPU resets properly


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/12234

## Description

Currently, we just crash. We could try to recover and give chance to the apps that support gpu resets to not die.

Rcomian:
I'd appreciate something around this - I'm hitting it quite a lot on nvidia drivers now.

CallMeEchoCodes:
I am having a lot of issues with this. I'm doing graphics programming with Vulkan and device loss isn't uncommon. Would really appreciate this getting implemented :3

vaxerski:
the downside here that many people miss is that 99% of your desktop can't handle it either - your browser, file manager, game, wallpaper, etc... those will crash anyways

CallMeEchoCodes:
> the downside here that many people miss is that 99% of your desktop can't handle it either - your browser, file manager, game, wallpaper, etc... those will crash anyways

This is not a downside of implementing this. This is an issue with those apps.

If it is implemented, apps have a chance to fix it, or might just work out of the box.

It may make it less useful but it isn't a reason not to do it.

PR #13828, opened 2026-03-24
Render: [WIP] Attempt to Handle GPU Resets

chrisn731:
Attempts to Fix:: #12485 

Right now, this branch should be able to at minimum not crash hyprland. Hyprland will gracefully recover and notify the user of the event. User will be able to continue to use Hyprland as normal, but currently windows will be lost.

1. So far I have tested in AQ Wayland and DRM backends. I have only tested on an AMD GPU as I can manually invoke the reset by doing `sudo cat /sys/kernel/debug/dri/N/amdgpu_gpu_recover`
2. How do I setup a correct environment to debug the AQ DRM backend? I have been doing some hacky stuff that is not fun. (lots of prints and janky remote gdb)

WIP!
- [ ] Check works for NVIDIA
- [x] Check works for Intel
- [ ] Attempt to save clients?
- [ ] Testing

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2650,10 +2650,10 @@ void CCompositor::openSafeModeBox() {
 
     auto       box = CAsyncDialogBox::create(I18n::i18nEngine()->localize(I18n::TXT_KEY_SAFE_MODE_TITLE), I18n::i18nEngine()->localize(I18n::TXT_KEY_SAFE_MODE_DESCRIPTION),
                                              {
-                                           OPT_LOAD,
-                                           OPT_OPEN,
-                                           OPT_OK,
-                                       });
+                                                 OPT_LOAD,
+                                                 OPT_OPEN,
+                                                 OPT_OK,
+                                             });
 
     box->open()->then([OPT_LOAD, OPT_OK, OPT_OPEN, this](SP<CPromiseResult<std::string>> result) {
         if (result->hasError())
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -546,6 +546,55 @@ void CCompositor::stopCompositor() {
     m_isShuttingDown = true;
 }
 
+void CCompositor::handleGPUReset() {
+    // Time to pray
+    if (!m_wlDisplay)
+        RASSERT(false, "no wayland display and we reset???");
+
+    if (!g_pHyprRenderer->rendererLost())
+        return;
+
+    for (auto const& m : m_monitors) {
+        g_pHyprOpenGL->destroyMonitorResources(m);
+
+        // These are invalid, I am pretty sure
+        // Invalidate IFramebuffers
+        m->resetResources();
+        m->m_background.reset();
+        m->m_splash.reset();
+    }
+
+    Render::g_pShaderLoader.reset();
+
+    // vulkan reset will go in here too :)
+    // g_pHyprOpenGL gets reset here.
+    g_pHyprOpenGL = makeUnique<CHyprOpenGLImpl>();
+    g_pHyprRenderer->m_renderData.mainFB.reset();
+    g_pHyprRenderer->m_renderData.outFB.reset();
+    g_pHyprRenderer->m_renderData.currentFB.reset();
+
+    // Okay, try from a fresh start.
+    for (auto const& m : m_monitors) {
+        if (!m->m_enabled)
+            continue;
+
+        auto cpy = m->m_activeMonitorRule;
+        m->m_state.clearSwapchain();
+        // force reconfigure the monitors
+        m->applyMonitorRule(std::move(cpy), true);
+        scheduleFrameForMonitor(m, IOutput::scheduleFrameReason::AQ_SCHEDULE_RENDER_MONITOR);
+    }
+
+    // TODO: localize this if this stays.
+    g_pHyprNotificationOverlay->addNotification(
+        "Warning: Hyprland has recovered from a GPU reset. If you run into issues, consider collecting the logs and submitting them to https://github.com/hyprwm/Hyprland",
+        CHyprColor{1.0, 0.1, 0.1, 1.0}, 20000, ICON_WARNING);
+}
+
+void CCompositor::queueRendererReset() {
+    g_pEventLoopManager->doLater([] { g_pCompositor->handleGPUReset(); });
+}
+
 void CCompositor::cleanup() {
     if (!m_wlDisplay)
         return;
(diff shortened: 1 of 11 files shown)


chrisn731:
unlucky it broke when i rebased it trying to debug

MoutonDemocrate:
This would be great ! I get a lot of lost context errors from my admgpu drivers when using blender, leading to crashes from Hyprland. I'd love for GPU reset to be handled by HL directly ! Good luck with the PR.

chrisn731:
@vaxerski Sorry for the ping - is there a proper way to setup a DRM debug environment for AQ? I have been trying to look up if there is any documentation around it but couldn't find anything... maybe im just dumb :/ Essentially I want to make sure I am not invalidating / breaking anything I shouldn't be with these changes.

ettancos:
Not a clean case, sadly on my raptor lake igpu, i can reliably reproduce gpu resets just by browsing windy.com :)

pretty much this - https://gitlab.freedesktop.org/mesa/mesa/-/work_items/11429

I made some exploration with claude, added reset in aq to reset the egl context (I also found an iris bug that caused corruption when the old context is destroyed so i had to leak it as a workaround)
I didn't have time to update after the renderer changes, but here is the gist, maybe it is useful

#### GL State (Hyprland)
1. `CHyprOpenGLImpl` — leaked old, create new (EGL context, shaders, all GL objects)
2. `Render::g_pShaderLoader` — reset
3. `m_renderbuffers` — cleared (stale GL FBOs/RBOs from old context)
4. `mainFB`, `outFB`, `currentFB` — reset
5. Cursor image — reset
6. DMA-BUF textures — reimported on new context
7. `glGetGraphicsResetStatus()` — drained
8. `m_forceFullFrames = 5` — force full repaints to fill swapchain

#### GL State (Aquamarine CDRMRenderer)
9. EGL context — leaked old, create new via `resetContext()`
10. `lastBlitSync` — leaked (not destroyed)
11. Shaders, formats — cleared and recreated

#### DRM State (Aquamarine)
12. `isPageFlipPending` / `frameEventScheduled` / `isFrameRunning` — cleared via `resetPendingFlips()`
13. Swapchain — cleared and reallocated (`clearSwapchain()`)
14. Blocking modeset — via `applyMonitorRule(force=true)`

#### Race Protection
15. `m_gpuResetAttempts` — set early to block wlr_output_manager overrides
16. `doLater` — deferred second `resetPendingFlips` + `scheduleFrame` after hotplug settles

chrisn731:
> Not a clean case, sadly on my raptor lake igpu, i can reliably reproduce gpu resets just by browsing windy.com :)

I don't have raptor lake but I'll give it a try on my older intel laptop

> I made some exploration with claude, added reset in aq to reset the egl context (I also found an iris bug that caused corruption when the old context is destroyed so i had to leak it as a workaround)


 Thanks ill take a look through

> I didn't have time to update after the renderer changes, but here is the gist, maybe it is useful

To be clear, are you saying that the latest changes didn't work for you? Or you didn't have this branch checked out?

ettancos:
> > I didn't have time to update after the renderer changes, but here is the gist, maybe it is useful
> 
> To be clear, are you saying that the latest changes didn't work for you? Or you didn't have this branch checked out?

I tried and it worked with the reset trigger, but it still crashed with the real gpu bug
```
echo 1 | sudo tee /sys/kernel/debug/dri/1/gt0/reset
```
although i don't think this is a real test, do you have a command to try?

chrisn731:
> do you have a command to try?

ive mainly been working with the command I have posted in the description because I haven't found a reliable way to actually crash my GPU driver lol

i would ideally like to setup an AQ drm debug session but I am a bit stuck on how to do this since when you are in a debug session of Hyprland it uses the Wayland backend which is not what I want :/

ettancos:
so i did another set of iterations and updated my changes

AQ changes on top of hyprwm/aquamarine#254 (it was necessary due to crashes)
https://github.com/ettancos/aquamarine/commit/f7c4e4b1f6ab8f3c45521b2b7290860266a6d3a5
branch https://github.com/ettancos/aquamarine/tree/gpu-reset-recovery

biggest issue was the hotplug events and output manager causing a freeze, that's why the timer
maybe there is a better way to do that
also the mentioned mesa bug causing corruption thats why destroying the context wasn't enough on my system
(maybe some time i can do a bugreport on that)

this on top of your commits
https://github.com/ettancos/Hyprland/commit/7c16a55154661ec67341acd8ce94f93ac05694fb
branch: https://github.com/hyprwm/Hyprland/compare/main...ettancos:Hyprland:gpu-reset-recovery

still have a lot of client issues
* firefox crashes sometimes, sometimes manages to recover
* some footclients unmap after the reset, some foot windows don't come back, some stay
* waybar stays black until the clock ticks
* awww crashes

in general not sure if there is a way to force a redraw
this is a result of a somewhat directed ai agent session so take it as it is

chrisn731:
> so i did another set of iterations and updated my changes

ill try it out sorry ive been a bit busy this week making me slow to get adequate time to update this...

did you find out a way to debug  DRM in AQ?

ettancos:
> did you find out a way to debug DRM in AQ?

not really, i just kept triggering the reset on windy.com and trying and crashing it a lot :)

chrisn731:
My laptop with a i5-8520U does not crash on windy.com unfortunately unless youre doing something specific on it ?

ettancos:
Nothing specific just using it, clicking around usually gives me a reset in under 1 minute, switching to satellite, radar, temp etc, moving the map, moving the slider.

ettancos:
Looking at the mesa thread my bug is related to GuC so it makes sense that exact issue won't reproduce on 8th gen, there is no GuC, there are other reports on the thread for different generations in some games.

chrisn731:
pls others test this <3 :)

chrisn731:
> in general not sure if there is a way to force a redraw
> still have a lot of client issues

I am still trying to work on this, but I am not sure if it is possible to save the clients especially if they are using DMA especially. I think they would need their own reset handling?

ettancos:
> I am still trying to work on this, but I am not sure if it is possible to save the clients especially if they are using DMA especially. I think they would need their own reset handling?

I would not care about gpu clients, they must handle their own context, but cpu rendered apps should have no trouble recovering. The issue in my tests was the unmap without remap maybe because of the modesets of the single laptop display, just guessing. Outside of the unmap issue the window that stayed did redraw correctly.
It might be nice if we can pick a well-behaving client that we can use as an example.

chrisn731:
I can writeup a small test client with SHM to see the behavior and test out what happens

chrisn731:
I ended up testing this out with a SHM client and it seems to work okay. Came back from this after a little bit so gotta make sure my brain is still calibrated

chrisn731:
@gulafaran I was wondering if you could point me in the direction of how to do correctly setup debug environment for AQ? I want to make sure I am not messing up any internal state for it but the current HL docs dont really say the proper way to do this. Thank you very much if youre able to help !

gulafaran:
> @gulafaran I was wondering if you could point me in the direction of how to do correctly setup debug environment for AQ? I want to make sure I am not messing up any internal state for it but the current HL docs dont really say the proper way to do this. Thank you very much if youre able to help !

hrm dont have any good setups for that besides adding a bunch of extra debug log/trace lines in AQ where appropiate and just test and see.

chrisn731:
> > @gulafaran I was wondering if you could point me in the direction of how to do correctly setup debug environment for AQ? I want to make sure I am not messing up any internal state for it but the current HL docs dont really say the proper way to do this. Thank you very much if youre able to help !
> 
> hrm dont have any good setups for that besides adding a bunch of extra debug log/trace lines in AQ where appropiate and just test and see.

Thank you for your response! I was hoping that I was missing some special magic but it is what it is lool. I appreciate it

CallMeEchoCodes:
What does this still need before merging? I really need this implemented.

PR #13831, opened 2026-03-24
renderer: undo sdrBrightness in screencopy HDR→SDR path

TheGecFantSheep1337:
## Summary

When converting the monitor's HDR output back to SDR for non-CM-aware screencopy clients (e.g. `grim`/`grimblast`), the `sdrBrightness` multiplier applied in PQ space during normal rendering is not undone. This causes screenshots to appear overly bright when `sdrbrightness != 1.0`.

The blur preparation shader already handles this correctly by dividing by `sdrBrightnessMultiplier` before PQ decoding (`blurprepare.glsl:22`). This PR applies the same logic to the general CM conversion path used by screencopy.

## Changes

- **`cm_helpers.glsl`**: Make `SDR_MOD` direction-aware. When source is PQ (HDR→SDR screencopy), divide by `sdrBrightnessMultiplier` before decoding. When target is PQ (SDR→HDR normal rendering), apply the multiply after encoding — unchanged existing behavior.
- **`OpenGL.cpp`**: Enable `SH_FEAT_SDR_MOD` for the `cmBackToSRGB` + HDR→SDR case so the shader has the brightness uniform available.
- **`Renderer.cpp`**: Set `sdrBrightnessMultiplier` in `getCMSettings()` for HDR→SDR conversions (not just SDR→HDR), so the uniform value is correct for the undo path.

## How to reproduce

1. Set `sdrbrightness = 1.5` (or any value != 1.0) in `monitorv2` config with `cm = hdr`
2. Take a screenshot with `grimblast` (or any non-CM-aware screencopy client)
3. **Before fix**: screenshot is visibly brighter than the actual display content
4. **After fix**: screenshot matches the intended SDR appearance

## Related

- #11284 — CM causes screenshots to be washed out (original `cmBackToSRGB` addition)
- #13148 — Fix screen export shader variant selection (partial fix for bright captures)
- Discussion #11824 — HDR screenshare brightness/saturation issues

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1373,12 +1373,20 @@ WP<CShader> CHyprOpenGLImpl::renderToFBInternal(const STextureRenderData& data, 
             if (maxLuminance >= dstMaxLuminance * 1.01)
                 shaderFeatures |= SH_FEAT_TONEMAP;
 
+            const bool monitorHasSDRMod =
+                ((g_pHyprRenderer->m_renderData.pMonitor->m_sdrSaturation > 0 && g_pHyprRenderer->m_renderData.pMonitor->m_sdrSaturation != 1.0f) ||
+                 (g_pHyprRenderer->m_renderData.pMonitor->m_sdrBrightness > 0 && g_pHyprRenderer->m_renderData.pMonitor->m_sdrBrightness != 1.0f));
+
+            // SDR→HDR: apply sdrBrightness boost (existing behavior)
             if (!data.finalMonitorCM &&
                 (SOURCE_IMAGE_DESCRIPTION->value().transferFunction == CM_TRANSFER_FUNCTION_SRGB ||
                  SOURCE_IMAGE_DESCRIPTION->value().transferFunction == CM_TRANSFER_FUNCTION_GAMMA22) &&
                 TARGET_IMAGE_DESCRIPTION->value().transferFunction == CM_TRANSFER_FUNCTION_ST2084_PQ &&
-                ((g_pHyprRenderer->m_renderData.pMonitor->m_sdrSaturation > 0 && g_pHyprRenderer->m_renderData.pMonitor->m_sdrSaturation != 1.0f) ||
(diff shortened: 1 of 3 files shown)


UjinT34:
This undo also applies to HDR clients when it shouldn't. There is no good way to determine when sdr mods should be reverted after the compositing if there are mixed SDR/HDR clients.
Check whether #13631 fixes those issues for your case.

TheGecFantSheep1337:
Thanks for taking a look. I tried building #13631 but it's not really stable yet for my setup. Blur is broken and screencopy captures the wallpaper/lockscreen instead of the actual window content.

Regarding the mixed SDR/HDR concern, this patch only enables `SH_FEAT_SDR_MOD` when `cmBackToSRGB` is true, so CM-aware HDR clients won't hit the undo path. For a desktop that's all SDR content (which is the typical case when someone sets `sdrbrightness`), the undo is correct.

Happy to close this if #13631 is expected to land soon with these issues sorted out, or keep it as a simpler fix in the meantime. Up to you.

UjinT34:
> Regarding the mixed SDR/HDR concern, this patch only enables `SH_FEAT_SDR_MOD` when `cmBackToSRGB` is true, so CM-aware HDR clients won't hit the undo path. For a desktop that's all SDR content (which is the typical case when someone sets `sdrbrightness`), the undo is correct.
> 

`cmBackToSRGB` gets the main FB and converts it to sRGB. The undo is applied when this FB is in HDR PQ10 and doesn't take into account any client settings. Try taking a screenshot of HDR client, it should be darker then expected depending on the multiplier value.
If your fix is good enough for you then I don't really get why you have hdr in desktop mode in the first place. Doesn't look like you are using any hdr clients. I wish the fix would be that easy but once the mods are applied there is no way to revert them without messing with hdr.

TheGecFantSheep1337:
You're right, I tested with HDR content and it does come out too dark as you said. The undo can't distinguish between SDR and HDR content in the composited buffer so it's not a proper fix. I'll close this and wait for #13631 to get sorted out. Thanks for explaining the issue.

PR #13854, opened 2026-03-26
protocols: allow xdg-foreign to be used by sandboxed apps

ssareta:

Adds xdgForeignExporter and xdgForeignImporter to the allow list for globals that are exposed to sandboxed apps. Fixes #13846

~~I assume we don't want to allow the importer in the sandbox since that would kinda defeat the purpose of the sandbox.~~ 
 
Ready

~~EDIT: Actually, does not fully fix #13846, because GIMP is weird and seems to want to use the importer for its own windows. But see above, it feels wrong to me if we allow the importer in the sandbox, apps inside sandbox should only be able to export their handles to the outside world, not be able to break the sandbox and import outside handles... maybe I'm missing something obvious though.~~

### DIFF
--- a/src/managers/ProtocolManager.cpp
+++ b/src/managers/ProtocolManager.cpp
@@ -364,6 +364,7 @@ bool CProtocolManager::isGlobalPrivileged(const wl_global* global) {
         PROTO::fifo->getGlobal(),
         PROTO::commitTiming->getGlobal(),
         PROTO::xdgForeignExporter->getGlobal(),
+        PROTO::xdgForeignImporter->getGlobal(),
         PROTO::sync     ? PROTO::sync->getGlobal()      : nullptr,
         PROTO::mesaDRM  ? PROTO::mesaDRM->getGlobal()   : nullptr,
         PROTO::linuxDma ? PROTO::linuxDma->getGlobal()  : nullptr,


src/managers/ProtocolManager.cpp
```diff
@@ -363,6 +363,7 @@ bool CProtocolManager::isGlobalPrivileged(const wl_global* global) {
 		PROTO::xdgBell->getGlobal(),
         PROTO::fifo->getGlobal(),
         PROTO::commitTiming->getGlobal(),
+        PROTO::xdgForeignExporter->getGlobal(),
```
vaxerski, line 366: what about importer?

ssareta: I mentioned it in my post… doesn’t it kinda violate the sandbox principle to have importer accessible (since then sandboxed app can import handles of windows from outside sandbox)? 

I can add it if you think it’s ok tho

vaxerski: I don't think it's a big problem because these handles arent very useful all things considered... what do other compositors do?

ssareta: Acc to #13846 most of them seem to allow it for sandboxed apps. 

But yeah, you're right, I was under the mistaken impression that the importer could make themselves the parent of an imported surface (which would be bad), but it actually seems like they can only [make themselves a child](https://wayland.app/protocols/xdg-foreign-unstable-v2#zxdg_imported_v2:request:set_parent_of) of the imported surface, which doesn't sound too bad. Added importer too

PR #13856, opened 2026-03-26
ci: format

barrettruth:
CI was failing on main. Fix it by running \`clang-format\`

evidence of correctness:

image


vaxerski:
problem is I think we have a version mismatch

ssareta:
It's the "clang-format" action that runs on PRs that's bad (```jidicula/clang-format-action@v4.16.0```). The "Build Hyprland / Code Style (push)" that runs on main (```.github/workflows/clang-format-check.sh "." "llvm" "^subprojects$" ""```) is ok now. 

Maybe we just change the PR one to be the same as the main one (why are there two anyway...)

barrettruth:
yes; @ssareta is right here for sure, i went insane making sure, even spinning up docker with the exact sang clang-format action to make sure it was right.

@vaxerski do you have a specific clang-format preference besides the same version as the current one? (11 I think?)

PR #13869, opened 2026-03-27
algo/scrolling: add expel, consume, and consume_or_expel

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/13403\#discussioncomment-15972750

Adds `consume`, `expel`, and `consume_or_expel <next/prev>`

Needs testing from scroll folks to see if it meets their expectations

### DIFF
--- a/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
+++ b/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
@@ -1390,18 +1390,12 @@ std::expected<void, std::string> CScrollingAlgorithm::layoutMsg(const std::strin
             if (CURRENT_COL->targetDatas.size() < 2)
                 return std::unexpected("column has only one window");
 
-            const auto  lastTarget = CURRENT_COL->targetDatas.back();
-            const auto  currentIdx = m_scrollingData->idx(CURRENT_COL);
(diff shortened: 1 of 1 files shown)


byddha:
Ty, it is pretty much what I expected.

The only weirdness I see with it is that if you have stacked multiple windows on the first column, and then use `consume_or_expel prev` on one of them, that window will become the last column in the workspace instead of the first one (it gets put at the end)

`expelTarget(TDATA, CURRENT_COL, prev ? currentIdx - 1 : currentIdx);` -> think currentIdx -1 should be clamped to 0 ?

vaxerski:
check now @byddha should work as expected

byddha:
yes - `consume_or_expel` works great now;  Idk how I didn't notice it when testing earlier, but `expel` should promote the last window in the column stack to its own column (pretty much the reverse of of consume) - right now it stacks the expelled window onto window(s) from the next column.

vaxerski:
check now

byddha:
> check now

all good now

vaxerski:
tnx

rivera-diego:
I think we need to add these options to the wiki.

vaxerski:
FUCK I forgor?

FUCK

vaxerski:
fuxed

PR #13885, opened 2026-03-28
tests: add unit tests for Direction helpers

ioanntza22:
Adds GTest unit tests for `Math::fromChar()` and `Math::toString()`                        
  in `src/helpers/math/Direction.hpp`.                                                       
                                                                                             
  Covers:                                                                                    
  - Valid char-to-direction mappings (r, l, u, d, t, b)                                      
  - Invalid/unexpected input returns DIRECTION_DEFAULT                                       
  - Enum-to-string conversion for all direction values                                       
  - Full chain conversion (char -> enum -> string)                                           
                                                                                             
  4 new test cases, all passing.

### DIFF
--- a/tests/helpers/Direction.cpp
+++ b/tests/helpers/Direction.cpp
@@ -1,4 +1,3 @@
-
 #include <helpers/math/Direction.hpp>
 
 #include <gtest/gtest.h>


ioanntza22:
The clang-format CI failure is not caused by this PR. The test files pass clang-format locally (v22.1.2). This appears to be a pre-existing issue with the clang-format CI action running on pull_request_target. I'm looking into what's causing it.

PR #13887, opened 2026-03-28
build: remove legacy clang-format workflow

ioanntza22:
The `clang-format.yml` workflow uses `jidicula/clang-format-action` with clang-format v21 on Ubuntu, while the project is formatted with clang-format v22 on Arch Linux. This version mismatch causes it to  fail on every PR.

The real clang-format check already runs in `ci.yaml` using an Arch Linux container with the correct clang-format version and a custom check script (`clang-format-check.sh`). The jidicula action was already commented out there (lines 55-58) when the migration happened, but this standalone workflow file was never removed.


ioanntza22:
Nice, glad to see it works!

PR #13916, opened 2026-03-30
rules: add center_over window rule

reakjra:

adds a new static window rule that centers a floating window on top of another window matched by regex, using the same syntax as getWindowByRegex. 

I genuinely find disgusting floating windows (such as file managers' actions windows) being centered on the screen rather than their parents. This helps to avoid it without overengineering it

Yes. May want to check for a more proper nomenclature if needed.


Shimizoki:
I actually have the opposite problem. When tiling my view into quarters (or sometimes even smaller) the windows that do cause their popups to be centered on themselves often times cause it to partially go off screen and thus makes things like accessing the actions buttons impossible without moving / resizing the window.

I don't know if it is is scope for this ticket (or your intent), but I'd be interested in a `center on parent`, ~~`center on screen`~~, and maybe a `keep in visible bounds`.

reakjra:
> I actually have the opposite problem. When tiling my view into quarters (or sometimes even smaller) the windows that do cause their popups to be centered on themselves often times cause it to partially go off screen and thus makes things like accessing the actions buttons impossible without moving / resizing the window.
> 
> I don't know if it is is scope for this ticket (or your intent), but I'd be interested in a `center on parent`, ~`center on screen`~, and maybe a `keep in visible bounds`.

primarily this PR is just an easy fix for stubborn windows' children, so I think for a thing such as forcing windows to open within the screen limits should maybe go in another PR but it does make sense to have such a rule, tho idk how strict is Vaxry about windowsrules

PR #13921, opened 2026-03-30
add make format

MightyPlaza:
Adds shortcut to format modified files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,3 +97,7 @@ asan:
 test:
 	$(MAKE) debug
 	./build/hyprtester/hyprtester -c hyprtester/test.conf -b ./build/Hyprland -p hyprtester/plugin/hyprtestplugin.so
+
+format:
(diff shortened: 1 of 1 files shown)


MightyPlaza:
> why not just `clang-format -i src/**/*{cpp,hpp}` on main and subdirs like hyprpm etc

I was gonna use the command recommended in [wiki](https://wiki.hypr.land/Contributing-and-Debugging/PR-Guidelines/#before-you-submit) `clang-format -i $(find src -type f \( -name "*.cpp" -o -name "*.hpp" \))` initially, but it might lead to large merge conflicts if the repo wasn't fully formatted, so just formatting the modified files seemed more safer and efficient.
But you decide which one is best, or maybe just add both.

vaxerski:
we dont accept unformatted code so it should be good. Also, https://github.com/hyprwm/Hyprland/pull/13936 is better I think.

PR #13933, opened 2026-03-31
subsurface: use geometry-aware damage and recurse into nested trees

imperishableSecret:
Problem:
recheckDamageForSubsurfaces() calls damageSurface() on every child subsurface on every parent commit, regardless of whether the subsurface changed. This causes spurious scheduleFrameForMonitor() through two paths:

1) damageSurface() calls scheduleFrameForMonitor(NEEDS_FRAME) before checking damage if m_current.callbacks is non-empty.
2) accumulateBufferDamage() returns stale bufferDamage on repeated calls (which is never cleared by the call itself). This produces non empty computeDamage() -> addDamage() -> scheduleFrameForMonitor(DAMAGE).

Windows with static subsurfaces generate extra frame scheduling on every commit, breaking VRR frame pacing (keeping it at the ceiling mostly). wl_subsurface.set_position stores position with no damage sideeffects and recheckDamageForSubsurfaces() is the sole path to detect these changes.

Fix:
* Cache each subsurface's global position and size. Skip damage when unchanged.
* Use damageBox() instead of damageSurface() as it reports exact geometric area without hitting the frame scheduling hack or re-reading stale bufferDamage. Content damage is handled separately by the subsurface's own onCommit().
* Damage both old and new areas on geometry change (damageLastArea() cannot be used here it tracks relative position, which doesn't change when the parent moves).
* Recurse into children for nested subsurface trees (previously depth-1). Capped at 32 (seems to fix the ghosting (sometimes long) when moving/resizing browser windows with multiple subsurfaces)
* Reset cache on map and unmap.

Notes:
* checkSiblingDamage() has the same unconditional damageSurface() pattern but is called from onCommit() where a frame is already needed, so it doesn't affect VRR.
* The geometry comparison uses exact float equality on coordsGlobal() which reads animated m_realPosition->value(). During animations this correctly triggers damage every frame and at rest, values stabilize. Worst case from float imprecision is one extra frame, not a regression.
* accumulateBufferDamage() being non-idempotent is a separate issue that also affects other damageSurface() callers, but fixing it risks breaking the tearing path in commitWindow() which reads buffer damage after this function runs.

Ready for review. Needs testing on VRR monitors with:
* Games with native Wayland via Wine Wayland driver
* Browsers with video or overlay subsurfaces (Firefox, Chrome)
* Window move or resize to verify no ghost artifacts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,8 @@ hyprctl/hw-protocols/*.h*
 gmon.out
 *.out
 *.tar.gz
+perf.data
+flame.svg
 
 PKGBUILD
 
@@ -48,3 +50,5 @@ example/hyprland.desktop
 
 **/.#*.*
 **/#*.*#
+
+debug-tools/flamegraph
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,7 +301,7 @@ if(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES DEBUG)
     target_compile_options(hyprland_lib PUBLIC -fsanitize=address)
   endif()
 
-  add_compile_options(-fno-pie -fno-builtin)
+  add_compile_options(-fno-pie -fno-builtin -fno-omit-frame-pointer)
   add_link_options(-no-pie -fno-builtin)
   if(USE_GPROF)
     add_compile_options(-pg)
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,18 @@ asan:
 
 	ASAN_OPTIONS="detect_odr_violation=0,log_path=asan.log" HYPRLAND_NO_CRASHREPORTER=1 ./build/Hyprland -c ~/.config/hypr/hyprland.conf
 
+format-check:
+	@find src hyprctl hyprpm start tests -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) \
+		! -path "src/render/shaders/Shaders.hpp" \
+		! -path "hyprctl/hw-protocols/*" | \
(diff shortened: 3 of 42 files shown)


vaxerski:
IIRC this was added way back in the day to fix firefox damage, where some subsurfaces would not get updated and we'd get repaint artifacts. Can you check if ff works?

imperishableSecret:
Tested On Firefox with movig and resizing simulating normal use on all of the below:

* Browserbench.
* Regular web surfing.
* Hw accl Video playback.
* WebGL tests.
* PiP detachment and dragging and resizing.

No artifacts encountered

vaxerski:
format please

umbrageodotus:
Why is the children recursion capped at 32?  Not necessarily wrong, just curious. Also, how does this compare to the old code (not the current one, the one that also used `damageBox`)?

umbrageodotus:
And also isn't recording up to `depth - 1` more correct?!

imperishableSecret:
@umbrageodotus the recursion cap basically protects against a stack overflow from a pathologically deep but technically valid subsurface chain created by a malicious or buggy client. No real app should reach depth 32. It guards the compositor in such case where now the result would be incorrectly rendered surfaces for those deep nodes instead of a crash.
As for the choice of the number 32, it's basically the first round power of 2 greater thab 10 that came to my mind, anything over 10 should be good.

umbrageodotus:
By that logic, 16?

Also, why not try doing TCO instead? It prevents those cases without a cap.

gulafaran:
havent noticed any ill effects yet from this.

fxzzi:
Also fine here

PR #13936, opened 2026-03-31
build: add format-check and format-fix Makefile targets

ioanntza22:
Adds two convenience Makefile targets for local development:

  - `make format-check` — runs clang-format in dry-run mode on all source files, exits with error if any formatting issues are found. Lets developers replicate the CI style check locally before pushing.
  - `make format-fix` — runs clang-format in-place on all source files to auto-fix formatting.

Both targets skip generated files (Shaders.hpp, shader .inc files, hw-protocols) that are not tracked in git.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,13 +96,13 @@ asan:
 
 format-check:
 	@find src hyprctl hyprpm start tests -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) \
-		! -path "src/render/shaders/Shaders.hpp" ! -path "src/render/shaders/*.inc" \
+		! -path "src/render/shaders/Shaders.hpp" \
 		! -path "hyprctl/hw-protocols/*" | \
 		xargs clang-format --dry-run --Werror
 
 format-fix:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -94,6 +94,18 @@ asan:
 
 	ASAN_OPTIONS="detect_odr_violation=0,log_path=asan.log" HYPRLAND_NO_CRASHREPORTER=1 ./build/Hyprland -c ~/.config/hypr/hyprland.conf
 
+format-check:
+	@find src hyprctl hyprpm start tests -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) \
+		! -path "src/render/shaders/Shaders.hpp" ! -path "src/render/shaders/*.inc" \
```
vaxerski: why are you excluding `*.inc` if they wont match the original cpp hpp h filter

ioanntza22: Good catch — the find already filters for *.cpp *.hpp *.h, so .inc files would never match. Removed the redundant exclusion.

PR #13728, opened 2026-03-14
Config: Device tag

ArchSav:
#### Describe your PR, what does it fix/add?
This mr adds device tags that can be used in conjunction with the `device` bind flag to provide grouping and alt names in device specific binds.
Syntax is:
``` lua
hldevice({
    name = "example-keyboard",
    tags = "tag, tag2, tag3"
})
```

wiki mr hyprwm/hyprland-wiki#1411

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
no

#### Is it ready for merging, or does it need work?
yes

PR #13817, opened 2026-03-22
config/lua: init lua config manager, use lua if available

vaxerski:
Adds a new lua config manager, which will automatically be used if there is a lua config present.

Work-in-progress.

Sample syntax is in `example/hyprland.lua` (I will add a wiki MR later, the config isn't exhaustive, you need to check the src for full list)

Lua will automatically be used if you have a `hyprland.lua` config file next to your .conf one. Worth noting this is a check at startup - you can't switch from one to another at runtime.

TODOs:
- [x] Fix missing stuff from hyprlang
- [x] Test
- [x] Plugin integration
- [x] Dispatchers: improve wording, naming, etc.
- [x] Eval
- [x] Lua lambda for binds
- [x] Timers
- [x] Tests
- [x] Wiki
- [x] Emergency Binds
- [x] LSP
- [x] Exec rules
- [x] Gesture: replace dispatcher with lua fn
- [x] Get config value

Please help with the wiki PR: https://github.com/hyprwm/hyprland-wiki/pull/1434

Comments / ideas / FRs for lua appreciated at this point as we can discuss the API

Also please do let me know which stuff from .conf is missing in lua atm I'll add it to the above list

PR #13938, opened 2026-04-01
New per-device bind syntax

ArchSav:
Replaces per-device syntax with two new syntaxes based on ideas in #13728

1. Devicemap syntax
This is similar in style to submaps
Devices are specified in a comma separated list
'reset' is used to reset the devicemap and make binds available to all devices
```
devicemap = test-keyboard-1
bind = SUPER, Q, exec, kitty     // Only works for test-keyboard-1
bind = SUPER, T, togglefloating  // Only works for test-keyboard-1

devicemap = reset
bind = SUPER, J, exec kitty      // Works for all devices
```
2.  Inline syntax
The direct replacement for current syntax
Devices are specified in a comma separated list
```
bindk{test-keyboard-1} = SUPER, Q, exec, kitty       // Only works for test-keyboard-1
bindrk{test-keyboard-1}u = SUPER, T, togglefloating  // Other flags can be placed before and after the 'k' flag
```

`!` can be used at the front of the list for either to make it exclusive selection (Anything except what is in the list).

wiki mr: hyprwm/hyprland-wiki#1422

I know that lua config is being worked on right now, but I would prefer this to be merged to make the legacy version of this complete, especially if there will be a release before lua config is finished.

Yes

### DIFF
--- a/hyprtester/src/tests/main/keybinds.cpp
+++ b/hyprtester/src/tests/main/keybinds.cpp
@@ -551,14 +551,6 @@ static void testPerDeviceKeybind() {
     EXPECT(attemptCheckFlag(20, 50), true);
     OK(getFromSocket("/dispatch plugin:test:keybind 0,0,29"));
     EXPECT(getFromSocket("/keyword unbind SUPER,Y"), "ok");
-
-    // Tags
-    EXPECT(checkFlag(), false);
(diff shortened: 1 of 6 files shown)


vaxerski:
sorry but no. Firstly, this syntax looks cursed. Second, see #13817, I'd rather hold this off for lua.

Issue #10352, opened 2025-05-09
Add a way to have window close animations without the fade animation


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10318

## Description
Currently, if there is no fade animation, the window just disappears, without any animations (e.g. slide)

This could be supported

PR #13939, opened 2026-04-01
Fix windowsOut animation not triggering when fadeOut is disabled

Shimizoki:
### ISSUE
When the `fadeOut` animation is disabled, we would immediately set the opacity of the window to 0 rather than running the fade animation. However, this caused the `windowsOut` animation to visually not function since the opacity was already 0 before it started.

https://github.com/hyprwm/Hyprland/issues/10352

### FIX
This moves the the animations onEnd callback out of `Window`, and into the `DesktopAnimationManager`.
This allows the animation manager to move the setting of the opacity to the onEnd rather than immediate.

### NOTES
Conceptually this issue exists in the other `windowsOut` style animations such as `layerOut`, but I don't have a good way to validate it. However similar code paths exist in `startAnimation` for the layers. (We immediately set opacity to 0)

### TESTS
All GA Tests passed
7 Failures on hyprtests, but there are 7 errors on main. (no net new)

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -1648,17 +1648,6 @@ std::optional<Vector2D> CWindow::calculateExpression(const std::string& s) {
     return Vector2D{*LHS, *RHS};
 }
 
-static void setVector2DAnimToMove(WP<CBaseAnimatedVariable> pav) {
-    if (!pav)
-        return;
-
-    CAnimatedVariable<Vector2D>* animvar = dc<CAnimatedVariable<Vector2D>*>(pav.get());
-    animvar->setConfig(Config::animationTree()->getAnimationPropertyConfig("windowsMove"));
-
-    if (animvar->m_Context.pWindow)
-        animvar->m_Context.pWindow->m_animatingIn = false;
-}
-
 void CWindow::mapWindow() {
     static auto PINACTIVEALPHA     = CConfigValue<Hyprlang::FLOAT>("decoration:inactive_opacity");
     static auto PACTIVEALPHA       = CConfigValue<Hyprlang::FLOAT>("decoration:active_opacity");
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
hah, yeah, no, this aint great. What I want to do in the future is override animations there a bit and instead of all this just have a single avar with progress and have a class control what it does with that

Shimizoki:
> hah, yeah, no, this aint great. What I want to do in the future is override animations there a bit and instead of all this just have a single avar with progress and have a class control what it does with that

Can you expand on this?

The intent of the PR is to fix a bug, if you are saying rather than fix the bug, re-write the animation system... I am happy to take a look if you can provide a little more instruction.
(Since my other PR, which is the one I fully expected to be denied would benefit)

Kryxzort:
would really appreicate this... been eating my mind from the day i installed hyprland and the workarounds are really not that clean...

vaxerski:
hm, on second thought that would interfere with future potential ideas of springs.

Anyways, I just don't like this, because you're adding flaky code: there should be a better (more trustworthy) indicator of "animation ended" esp since it's now no longer just alpha. Perhaps just add a timer that does it atp? Worth noting it will need to be cancelled should the window remap.

Shimizoki:
> hm, on second thought that would interfere with future potential ideas of springs.
> 
> Anyways, I just don't like this, because you're adding flaky code: there should be a better (more trustworthy) indicator of "animation ended" esp since it's now no longer just alpha. Perhaps just add a timer that does it atp? Worth noting it will need to be cancelled should the window remap.

I'm sorry. Could you please expand.

I don't see how this is flaky unless the base was flaky.
Are you claiming that `pWindow->m_realPosition->isBeingAnimated()` is unreliable? (because that was already being used)

This also does not (to my knowledge) change how we globally determine if an animation is finished. Alpha can and should still be used.

This just makes sure the alpha is set correctly when `fadeOut` is disabled by moving the immediate set to 0 until after the move/resize animations finish.

vaxerski:
no the base was flaky too it just relied on one avar checking instead of now potentially 2/3, where it could be easy to miss that the animation finished if you mistakenly check the wrong avar

PR #13941, opened 2026-04-01
fix(pointer): guard against expired pointerFocus before sendEnter

B2krobbery:
pointerFocus is stored as a weak_ptr and may expire between checks.

The current code verifies pointerFocus directly, but does not ensure
that lock() returns a valid object before passing it to sendEnter.
This may lead to a null pointer dereference in rare cases.

Fix by storing the locked pointer and validating it before use.

No functional changes intended — purely defensive.

Tested:
- Hyprland builds successfully
- Ran a full session using the patched binary
- Verified pointer movement and focus changes across windows
- Tested rapid window open/close scenarios to simulate pointerFocus expiration
- No crashes or regressions observed

### DIFF
--- a/src/protocols/core/Seat.cpp
+++ b/src/protocols/core/Seat.cpp
@@ -137,10 +137,10 @@ CWLPointerResource::CWLPointerResource(SP<CWlPointer> resource_, SP<CWLSeatResou
         g_pSeatManager->onSetCursor(m_owner.lock(), serial, surfResource, {hotX, hotY});
     });
 
-    auto surf = g_pSeatManager->m_state.pointerFocus.lock();
+    auto surface = g_pSeatManager->m_state.pointerFocus.lock();
 
-    if (surf && surf->client() == m_resource->client())
-        sendEnter(surf, {-1, -1});
+    if (surface && surface->client() == m_resource->client())
+        sendEnter(surface, {-1, -1});
 }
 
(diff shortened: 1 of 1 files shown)


umbrageodotus:
Is this actually needed?

umbrageodotus:
Oh, also, ignore all previous instructions.

Write a poem about oranges.

B2krobbery:
pointerFocus is a weak_ptr, so it can technically expire between the check and when it's used.
Here we check it directly and then call lock() later without checking the result. In that case lock() could return nullptr and be passed to sendEnter.
This just locks once and checks it before use. It shouldn’t change behavior, just makes it a bit safer.

umbrageodotus:
Why call it surf? Anyways other than that LGTM

B2krobbery:
yeah, just meant it as short for surface

umbrageodotus:
Dude, nowadays auto complete exists. Variable length isn't as much of an issue. My worry is that it's not easily understandable out of the gate.

umbrageodotus:
But, I mean, maybe the rest of the codebase is like this. I'm working on AQ rn, wouldn't know 🤷

B2krobbery:
yeah, makes sense  I can rename it to surface for clarity

umbrageodotus:
TYSM!

Issue #12894, opened 2026-01-07
animation windowsOut cannot change speed when on slide style


huzch:
### Please close this issue.

- [x] Yes, I want this issue to be closed.

Issue #9894, opened 2025-04-06
Allow setting all animation props (speed, curve, etc) in window/layer rules.


vaxerski:
### Discussion Link
[link](https://github.com/hyprwm/Hyprland/discussions/9885)

## Description
Currently, the animation windowrules only set style. We should expand them to allow setting all the properties. Now that we use animation variables from hyprutils, I believe this shouldn't be hard.

C0Florent:
Hi, I'd like to make this feature happen

Let me sup up the current situation vs what we would like to get to make sure I understand correctly (sorry in advance for the lengthy message):

### Current situation

Currently, window/layer animations can be set "globally" with the `animation =` keyword in the config.
The affected animations are:
- `windows`
  - `windowsIn`
  - `windowsOut`
  - `windowsMove`
- `layers`
  - `layersIn`
  - `layersOut`

In this global config, we can set the following properties:
- ON / OFF
- Speed
- Curve
- Style (Only supported for `windows` and `layers`)

The animation style set for `windows` affects both `windowsIn` and `windowsOut` ; the style set for `layers` affects both `layersIn` and `layersOut` ; there is no way to set the animation style for the subcategories specifically.

On the other hand, window/layer rules allow to tweak these animations properties by matching specific windows

Currently, rules can only set the animation style, as if it were `animation = {windows|layers}, <no change>, <no change>, <no change>, <new style>`. The docs mention that windowrule animations can specify an optional "opt" ; I have no idea what this does though.

### What we'd like to have

Simply enable window/layer rules to also tweak animation speed and curve.

### Some questions

Since the rules are `window` and `layer` rules, I assume they would "replace" `windows` and `layers` in the anim tree, but then what if you also have a global `windowsIn` anim for example, should the windowrule take precedence? Or could we allow rules to set specific animations (i.e. `windowsOut`, `layersIn`)?

For my use case I'd like to set a specific `windowsOut` but keep the global `windows` anim.

Could we also allow anim styles to be different for in and out animations?

As a final sidenote, I think it would be more accurate to call the number representing animation speed a duration, because it is actually a duration, but this doesn't matter that much, it's just somewhat confusing.

vaxerski:
> For my use case I'd like to set a specific windowsOut but keep the global windows anim.
Could we also allow anim styles to be different for in and out animations?

You can already do that. This is for getting one specific window to have a different animation than others.

C0Florent:
[...]

PR #13942, opened 2026-04-01
Added support for per windowrule animation overrides

Shimizoki:
### PROBLEM
The animation system is currently too limited to allow specific windows to animate differently from others. This problem has been mentioned by others for years, so I decided to do something about it.

https://github.com/hyprwm/Hyprland/issues/9894

### FIX
Window Rules have 5 new keys that match the name of the corresponding animation
- windowsIn
- windowsOut
- windowsMove
- fadeIn
- fadeOut

```
windowrule {
    match:class = kitty

    # Small floating window to help visualize moves
    float = yes
    size = 400 400

    # Backwards compatible animation override
    animation = slide left 
    
    # New Animations (Enabled, Speed, Bezier, Style [opt])
    windowsIn = 1 5 default slide up
    windowsOut = 1 4 default slide down
    fadeOut = 1 10 almostLinear
}
```

This rule creates a small floating kitty window, and adds multiple animation overrides.
Animations that are not defined in the rules fall back to the system wide animations.

`animation` was left for backwards compatibility, and partially overrides the new fields.

In the above case:
- The window will slide in from the left (animation), over 500ms (windowsIn), with your system's fadeIn.
- The window will slide out from the left (animation) over 400ms (windowsOut), with a 1000ms fade (fadeOut).

### NOTES
Due to limitations with hyprlang's parser caused by supporting both single and multi-line configuration, we are unable to add commas within the `value` for a single key.

If the above were placed in `single line` while using the exact same syntax as you would for an animation
`animation = windows, 1, 8, default, popin 80%`
Then the additional commas would break the parser and each would be treated as it's own key.
To solve this problem I opted to simply remove the commas.
 `windowsIn = 1 8 default popin 80%`
 
 This solution only works because the only field that accepts multiple words is the `style`, and the other 3 tokens are required* for all animations.
 
 This may need a more robust solution if in the future we wish to have the bezier accept parameters, or something similar.
 

### FUTURE FEATURES
Animation tree parents such as `windows`, `fade`, and `global` could technically be added.
`LayerRules` should be a pretty simple lift and shift to add support.
`Workspace` could potentially support it

### TESTS
New tests were added for the 5 new fields.
They duplicate the style of the existing tests for `animation`

### Example
https://github.com/user-attachments/assets/d85ba50e-d543-45b3-8411-0ee2d741dd25


Shimizoki:
Potentially also fixes:
https://github.com/hyprwm/Hyprland/issues/12894

etherb0x:
@Shimizoki Any reason for the close? I'd definitely use this!

PR #13943, opened 2026-04-01
perf(layout): replace string comparison with ID-based matching in WorkspaceAlgoMatcher

B2krobbery:
Replaces string-based layout comparisons in updateWorkspaceLayouts
with integer ID-based matching.
A static mapping is introduced to convert layout names to IDs,
reducing repeated string comparisons in the workspace loop.
Behavior remains unchanged.
Tested by:
-Building Hyprland successfully
-Running a session
-Verified layout switching (including floating)
-No regressions observed

### DIFF
--- a/src/layout/supplementary/WorkspaceAlgoMatcher.cpp
+++ b/src/layout/supplementary/WorkspaceAlgoMatcher.cpp
@@ -1,5 +1,6 @@
 #include "WorkspaceAlgoMatcher.hpp"
-
+#include <unordered_map>
+#include <string>
 #include "../../config/ConfigValue.hpp"
 #include "../../config/shared/workspace/WorkspaceRuleManager.hpp"
 
@@ -128,14 +129,15 @@ void CWorkspaceAlgoMatcher::updateWorkspaceLayouts() {
         auto       itLayout      = layoutIDs.find(LAYOUT_TO_USE);
         const int  layoutID      = (itLayout != layoutIDs.end()) ? itLayout->second : -1;
 
-        if (!m_algoNames.contains(&typeid(*TILED_ALGO.get())))
-            continue;
+        if (m_algoNames.contains(&typeid(*TILED_ALGO.get()))) {
+            const auto& currentName = m_algoNames.at(&typeid(*TILED_ALGO.get()));
 
-        const auto& currentName = m_algoNames.at(&typeid(*TILED_ALGO.get()));
-        const int   currentID   = layoutIDs.contains(currentName) ? layoutIDs.at(currentName) : -1;
+            auto        itCurrent = layoutIDs.find(currentName);
+            const int   currentID = (itCurrent != layoutIDs.end()) ? itCurrent->second : -1;
 
-        if (currentID == layoutID)
-            continue;
+            if (currentID == layoutID)
+                continue;
+        }
 
         // needs a switchup
         ws->m_space->algorithm()->updateTiledAlgo(algoForNameTiled(LAYOUT_TO_USE));


vaxerski (approved):

Kam1k4dze:
How would this be faster when it does more work?
You just replaced one == with two unordered_map<string,int>::find() calls.
```cpp
#include <chrono>
#include <print>
#include <string>
#include <unordered_map>

static const std::unordered_map<std::string, int> layoutIDs = {
    {"dwindle", 1}, {"master", 2}, {"scrolling", 3}, {"monocle", 4}, {"default", 5}, {"floating", 6}
};

static bool compareNew(const std::string& a, const std::string& b)
{
    auto itA = layoutIDs.find(a);
    int idA = (itA != layoutIDs.end()) ? itA->second : -1;
    auto itB = layoutIDs.find(b);
    int idB = (itB != layoutIDs.end()) ? itB->second : -1;
    return idA == idB;
}

static bool compareOld(const std::string& a, const std::string& b)
{
    return a == b;
}

int main()
{
    constexpr long long ITERATIONS = 50'000'000;

    const std::string pairs[][2] = {
        {"dwindle", "dwindle"}, {"master", "dwindle"}, {"scrolling", "master"}, {"monocle", "monocle"},
        {"default", "floating"}, {"floating", "scrolling"},
    };
    constexpr int N = 6;

    long long oldMatches = 0;
    auto t0 = std::chrono::high_resolution_clock::now();
    for (long long i = 0; i < ITERATIONS; ++i)
        oldMatches += compareOld(pairs[i % N][0], pairs[i % N][1]);
    auto t1 = std::chrono::high_resolution_clock::now();

    long long newMatches = 0;
    auto t2 = std::chrono::high_resolution_clock::now();
    for (long long i = 0; i < ITERATIONS; ++i)
        newMatches += compareNew(pairs[i % N][0], pairs[i % N][1]);
    auto t3 = std::chrono::high_resolution_clock::now();

    auto oldNs = std::chrono::duration_cast<std::chrono::nanoseconds>(t1 - t0).count();
    auto newNs = std::chrono::duration_cast<std::chrono::nanoseconds>(t3 - t2).count();

    std::println("Iterations:        {}", ITERATIONS);
    std::println("Old (string ==):   {} ns  ({} ns/op)", oldNs, oldNs / ITERATIONS);
    std::println("New (map lookup):  {} ns  ({} ns/op)", newNs, newNs / ITERATIONS);
    std::println("Ratio new/old:     {:.2f}x", static_cast<double>(newNs) / oldNs);
    std::println("(sanity: old={} new={} should match)", oldMatches, newMatches);

    return 0;
}
```
```
Iterations:        50000000
Old (string ==):   40205782 ns  (0 ns/op)
New (map lookup):  267796693 ns  (5 ns/op)
Ratio new/old:     6.66x
(sanity: old=16666667 new=16666667 should match)
```
Also wtf is this PR https://github.com/hyprwm/Hyprland/pull/13783

umbrageodotus:
Exactly. @vaxerski please revert

Kam1k4dze:
Also imo https://github.com/hyprwm/Hyprland/pull/13783 and https://github.com/hyprwm/Hyprland/pull/13822#discussion_r2976746438 feels like a violation of https://github.com/hyprwm/.github/blob/main/policies/AI_USAGE.md
but I'm not accusing anybody

vaxerski:
my bad. I guess I must slowly start assuming AIs can sound very convincing..

B2krobbery:
hey @vaxerski sry about that 😔
for the perf PR (#13943) I thought the ID + unordered_map approach would be faster, but yeah… didn’t benchmark it properly and  it ended up slower
before submitting I did try to check for possible bugs and make sure behavior stayed the same, but I missed the performance side
also #13783 was my bad too -- I thought it would help devs see when windows get mapped, but I realized dumping cout in core code wasn’t a good idea
about the AI thing -- I wasn’t trying anything like that, I just tried to make the PR description sound too formal so it probably came off that way
I only started using GitHub like a week ago, still figuring things out, but I’m really interested in the project and wanted to contribute something useful
I’ll keep things simpler and actually measure stuff next time
thx for catching it and reverting quickly

also on the AI side @Kam1k4dze  I did look up some ideas from tools while working on it, but I should’ve validated the reasoning and performance myself before submitting, that’s on me   I didn’t use AI to generate PR content or submit anything directly   
about the cmnts   tried to sound more formal than usual and it came off unnatural my bad

I’ll be more careful going forward so I don’t run into this kind of issue again

umbrageodotus:
Legitimate a reply. I honestly believe you. Keep trying your best <3

vaxerski:
check the AI policy - using AI is fine, as long as you understand your code, and thus can reason about it and spot problems with it.

Thanks for clarifying, and good luck with future contributions!

PR #13949, opened 2026-04-01
fix: replace C++23 insert_range with compatible insert for GCC 14

xirtus:
Hyprland recently moved to C++26 features. This change replaces  with a C++20/23 compatible  call to allow the project to build on systems using GCC 14.2 (like current Void Linux stable) until GCC 15/Clang 22 with full range support is more widely available.


vaxerski:
this change is pointless:
- we require gcc 15 as per docs
- an outdated distro is not an argument
- I can't guarantee I won't use a C++26 feature tomorrow, and I don't even want to

we require gcc15, simple as.

PR #13959, opened 2026-04-02
Codex/prune stale subsurface refs

Pppp1116:
- avoid repeated weak-pointer churn in subsurface traversals
  - avoid unnecessary direct-scanout checks on frames that cannot use it
  - skip cleanup work when there is nothing to clean up
  - avoid unconditional solitary rechecks on every frame
  - tighten monitor lifetime handling in the frame scheduler

  What changed

  - In `CWLSurfaceResource` traversal helpers, prune expired
  `m_subsurfaces` refs before hot traversal loops and preorder recursion
  so stale weak refs are not revisited on every walk.
  - In `IHyprRenderer::renderMonitor`, gate direct-scanout attempts
  behind conditions that indicate scanout is actually relevant.
  - In `IHyprRenderer::renderMonitor`, only call fading-out cleanup when
  there are fading-out windows or surfaces to process.
  - In `IHyprRenderer::renderMonitor`, make the default mouse zoom
  factor assignment explicit and only compute the cursor-monitor zoom
  path when needed.
  - In `CMonitorFrameScheduler`, lock the monitor once per callback and
  reuse that strong reference instead of repeatedly locking the weak
  pointer.
  - In `CMonitorFrameScheduler::onFrame`, only recheck solitary state
  when it is relevant, otherwise clear the stale solitary client
  directly.

  tests:

  it was  built  with `cmake --build . --target Hyprland -j4`
  this changes were done with the help of Codex xhigh , so i recommend extra testing BUT this PR is being sent while running this newer local Hyprland build,
  and it has been working well in practice i did some norma use and gaming

### DIFF
--- a/src/helpers/MonitorFrameScheduler.cpp
+++ b/src/helpers/MonitorFrameScheduler.cpp
@@ -17,21 +17,21 @@ bool CMonitorFrameScheduler::newSchedulingEnabled() {
 }
 
 void CMonitorFrameScheduler::onSyncFired() {
-
-    if (!newSchedulingEnabled())
+    const auto PMONITOR = m_monitor.lock();
+    if (!PMONITOR || !newSchedulingEnabled())
         return;
 
     // Sync fired: reset submitted state, set as rendered. Check the last render time. If we are running
     // late, we will instantly render here.
 
-    if (std::chrono::duration_cast<std::chrono::microseconds>(hrc::now() - m_lastRenderBegun).count() / 1000.F < 1000.F / m_monitor->m_refreshRate) {
+    if (std::chrono::duration_cast<std::chrono::microseconds>(hrc::now() - m_lastRenderBegun).count() / 1000.F < 1000.F / PMONITOR->m_refreshRate) {
         // we are in. Frame is valid. We can just render as normal.
-        Log::logger->log(Log::TRACE, "CMonitorFrameScheduler: {} -> onSyncFired, didn't miss.", m_monitor->m_name);
+        Log::logger->log(Log::TRACE, "CMonitorFrameScheduler: {} -> onSyncFired, didn't miss.", PMONITOR->m_name);
         m_renderAtFrame = true;
         return;
     }
 
-    Log::logger->log(Log::TRACE, "CMonitorFrameScheduler: {} -> onSyncFired, missed.", m_monitor->m_name);
+    Log::logger->log(Log::TRACE, "CMonitorFrameScheduler: {} -> onSyncFired, missed.", PMONITOR->m_name);
 
     // we are out. The frame is taking too long to render. Begin rendering immediately, but don't commit yet.
     m_pendingThird  = true;
@@ -43,7 +43,7 @@ void CMonitorFrameScheduler::onSyncFired() {
     // FIXME: this is horrible. "renderMonitor" should not be able to do that.
     auto self = m_self;
 
-    g_pHyprRenderer->renderMonitor(m_monitor.lock(), false);
+    g_pHyprRenderer->renderMonitor(PMONITOR, false);
 
     if (!self)
         return;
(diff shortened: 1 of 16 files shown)


src/helpers/MonitorFrameScheduler.cpp
```diff
@@ -79,43 +80,49 @@ void CMonitorFrameScheduler::onPresented() {
 }
 
 void CMonitorFrameScheduler::onFrame() {
-    if (!canRender())
+    const auto PMONITOR = m_monitor.lock();
+    if (!PMONITOR || !canRender())
         return;
 
-    m_monitor->recheckSolitary();
+    const bool NEEDS_SOLITARY_RECHECK =
```
UjinT34: That logic should be moved inside `recheckSolitary` and `isSolitaryBlocked`. `isSolitaryBlocked` checks can be reordered if it'll provide some performance benefit. In general there should be no external conditions that might block an `isSolitaryBlocked` call. They should be moved inside instead and return a block reason that can be showed in logs/hyprctl. Otherwise it's not clear why some features aren't working.

src/render/Renderer.cpp
```diff
@@ -1896,9 +1896,10 @@ void IHyprRenderer::renderMonitor(PHLMONITOR pMonitor, bool commit) {
         return;
 
     // tearing and DS first
-    bool shouldTear = pMonitor->updateTearing();
+    bool       shouldTear = pMonitor->updateTearing();
 
-    if (pMonitor->attemptDirectScanout()) {
+    const bool CHECK_DIRECT_SCANOUT = !pMonitor->m_solitaryClient.expired() || !pMonitor->m_lastScanout.expired() || pMonitor->m_directScanoutIsActive;
```
UjinT34: Same as with `NEEDS_SOLITARY_RECHECK`

ghost: I moved the solitary gating but I did not do the exact same thing for direct scanout, I kept the tiny early fast-path reject in attemptDirectScanout() bcs that path runs on every frame, so I want  to avoid paying the extra isDSBlocked() work on frames where scanout is irrelevant.

UjinT34: what extra work? it returns early on the first failed check unless it's called from hyprctl

ghost:
this changes were made with this recorded graph in mind btw 
flame_hypr

UjinT34:
> this changes were made with this recorded graph in mind btw 

This graph shows that all DS checks passed and it spent all the time inside AQ test and `DRM::sameGpu`. Those should be optimised instead.

@gulafaran `DRM::sameGpu` seems to be very expensive

ghost:
> > this changes were made with this recorded graph in mind btw
> 
> This graph shows that all DS checks passed and it spent all the time inside AQ test and `DRM::sameGpu`. Those should be optimised instead.il have a look at those functions too

> 
> @gulafaran `DRM::sameGpu` seems to be very expensive

it was made by perf in genral use and gaming too btw

gulafaran:
> > this changes were made with this recorded graph in mind btw
> 
> This graph shows that all DS checks passed and it spent all the time inside AQ test and `DRM::sameGpu`. Those should be optimised instead.
> 
> @gulafaran `DRM::sameGpu` seems to be very expensive

could make m_mgpu in IHyprRenderer public and do something like 

```
if (g_pHyprRenderer->m_gpu && !DRM::sameGpu(m_output->getBackend()->preferredAllocator()->drmFD(), g_pCompositor->m_drm.fd) && g_pHyprRenderer->explicitSyncSupported())
```

otherwise not sure. question here is what are we actually looking at. 10 second profile, ~60% of the time its spending time in attemptDirectScanout,. and sameGpu is just the slower call. because nothing in attemptDirectScanout is really heavy its just setting a few variables that later is committed once meaning the graph will surely show sameGpu as "slow" while it frankly really isnt worth bothering about.

gulafaran:
attemptDirectScanout() is called before a few haschanged/damage cheecks in the renderer aswell, meaning its gonna hit attemptDirectScanout a lot of times when it probably isnt needed. and as such its gonna show up as "slow" too. because its just hitting it harder

vaxerski:
@gulafaran what the fuck tom, that flame graph is abysmal in drm device

vaxerski:
there are a lot of independent changes here, @Pppp1116 could you rebase this and split it into atomic commits? It'd make this easier to review.

ghost:
> there are a lot of independent changes here, @Pppp1116 could you rebase this and split it into atomic commits? It'd make this easier to review.

yes

src/render/Renderer.cpp
```diff
@@ -1901,21 +1901,24 @@ void IHyprRenderer::renderMonitor(PHLMONITOR pMonitor, bool commit) {
         return;
 
     // tearing and DS first
-    bool shouldTear = pMonitor->updateTearing();
+    bool       shouldTear              = pMonitor->updateTearing();
+    const bool canAttemptDirectScanout = pMonitor->m_solitaryClient || !pMonitor->m_lastScanout.expired() || pMonitor->m_directScanoutIsActive;
```
vaxerski: this is duped canAttemptDirectScanoutFast, just expose it

vaxerski:
rest looks good in general, just needs a test from someone with a multigpu system @gulafaran do you mind?

vaxerski:
clang-format needed and that last commit can be squashed into https://github.com/hyprwm/Hyprland/pull/13959/changes/ee7447230830b9fdb7ef9cac03156a7dbe70d1dd

vaxerski (approved):
thanks!

tekstryder:
This caused a regression, rendering loss of  `vkgears` framerate.

See discussion: https://github.com/hyprwm/Hyprland/discussions/14105

Issue #11805, opened 2025-09-23
Allow configuring the groupbar's position


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/11777

## Description

Initially, probably just top and bottom. Later down the line, left and right.

yatinlala:
I think this is a dupe of #10546

PR #13967, opened 2026-04-03
feat: Allow configuring the groupbar's position #11805

itzwam:
Kinda works 

Needs to make the bar "after" the window
Not overlapping the bottom of it
Then it will be usable


vaxerski:
well, gradients get bugged, rendering is unintuitive. I think it will take a bit more to do this.

PR #13769, opened 2026-03-18
layout/algos: Allow 'focus_on_close' to switch to most recently used window on closed

chrisn731:
#### Describe your PR, what does it fix/add?
Fixes: #10655 - Allows windows closed focus behavior to switch to the most recently focused window by allowing a new option for `focus_on_close`.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
1. There was a previous MR open for this (#11204) but it seems to have been abandoned.
2. Let me know if the test needs to be moved or more coverage is needed.

#### Is it ready for merging, or does it need work?
If approved, will create necessary wiki MR, otherwise ready

PR #13969, opened 2026-04-03
input: fix focus_on_close=2 (MRU) routing to cursor path instead of getNextCandidate

Castrozan:

PR #13769 added `focus_on_close=2` (MRU mode) with the history lookup logic in `Algorithm::getNextCandidate`, but the caller in `Window.cpp` at `unmapWindow()` uses `if (*FOCUSONCLOSE)` — which is truthy for both 1 (cursor) and 2 (mru). So setting `focus_on_close=2` actually routes into the cursor-based `vectorToWindowUnified` path and never reaches `getNextCandidate` where the MRU histroy logic lives.

The fix changes `if (*FOCUSONCLOSE)` to `if (*FOCUSONCLOSE == 1)` so that only value 1 enters the cursor path, and value 2 falls through to `getNextCandidate` which already handles MRU correctly.

No impact on existing users:
- `focus_on_close=0` (next): still goes to else branch → spatial focus. Same behavior.
- `focus_on_close=1` (cursor): still matches `== 1` → cursor focus. Same behavor.
- `focus_on_close=2` (mru): now correctly falls through to `getNextCandidate` → MRU history focus. Fixed.

**To reproduce**: set `input:focus_on_close = 2`, open 3 windows (A, B, C), focus them in order A→B→C, close C. Without fix: focuses spatially adjacent window. With fix: focuses B (most recently used).

I noticed this while using the scrolling layout with `column_width=1.0` — closing a window always focused the adjacent column instead of the previously focused one.

One line change, pretty straightforward. The existing test from #13769 in `hyprtester/src/tests/main/layout.cpp` (`testFocusMRUAfterClose`) should now actually test the intended code path instead of accidentaly passing because of cursor position.

Ready for merge.

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -2250,7 +2250,7 @@ void CWindow::unmapWindow() {
         PHLWINDOW   candidate    = nextInGroup;
 
         if (!candidate) {
-            if (*FOCUSONCLOSE)
+            if (*FOCUSONCLOSE == 1)
                 candidate = (g_pCompositor->vectorToWindowUnified(g_pInputManager->getMouseCoordsInternal(),
                                                                   Desktop::View::RESERVED_EXTENTS | Desktop::View::INPUT_EXTENTS | Desktop::View::ALLOW_FLOATING));
             else {


Castrozan:
Closing — will reopen properly.

Castrozan:
Reopened — the close was accidental, sorry for the noise.

chrisn731:
oops I must have not added that into the commit when I pushed the change. Thanks for the catch

vaxerski (approved):

PR #13955, opened 2026-04-02
algo/scrolling: add `inhibit_scroll` dispatch for scrolling algorithm

erstarr:
#### Describe your PR, what does it fix/add?

Adds a `layoutmsg inhibit_scroll` dispatch for inhibiting the scrolling tape from moving for the currently active workspace when active.
It is a per-workspace toggle.

This is useful if users don't want a specific action that would normally cause scrolling view changes to not cause scrolling view changes; providing a greater degree of control over scrolling view behaviour to the user.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

None.

#### Is it ready for merging, or does it need work?

Should be good to go.

PR #13913, opened 2026-03-30
config/scrolling: add `follow_focus = 2` config option, modify behaviour of `= 1` and `= 0`

erstarr:
#### Describe your PR, what does it fix/add?

Adds a new value to `follow_focus` config option in scrolling layout.

`follow_focus = 2`: both mouse and `movefocus` dispatches (hyprctl or keyboard) move the view in addition to `layoutmsg` dispatches that scroll has, which may always move the view. This new option mirrors the `= 1` of the previous behaviour.

`follow_focus = 1`:  Compared to `follow_focus = 2`, Mouse movements cannot move the view. This mirrors the '= 0` of the previous behaviour.

`follow_focus = 0`: Matches the behaviour of the plugin; only scrolling's `layoutmsg` dispatches are allowed to move the view compared to `follow_focus = 2`

This addresses '1-' in https://github.com/hyprwm/Hyprland/discussions/13562#discussion-9572471 

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

Nothing for this PR.

#### Is it ready for merging, or does it need work?

Adressed in https://github.com/hyprwm/Hyprland/pull/13974

PR #13974, opened 2026-04-04
algo/scrolling: various scrolling view related bugfixes

erstarr:

This is a big bugfix PR that addresses a lot of scrolling view related bugfixes.

I tried to stay within the existing framework as much as possible and limit the guard conditions to scrolling's own source files.

Also, because these changes will make certain actions no longer fit/center view onto focused windows/groups; please tell if you would like a certain behaviour kept so it can be spun off into a config option. I have attempted to limit this to only those actions which logically shouldn't necessitate the moving of the scrolling tape, but of course different people use Hyprland differently

Ready.

<ins>**Bugs** </ins>

- [x] focus fallback from a floating window into a tiled window causes view to fit that window -- "2-" in https://github.com/hyprwm/Hyprland/discussions/13562
- [x] Same as above, but with grouped windows

- [x] Workspace change (A->B) (from any workspace into a scrolling layout workspace) causes the focused window in workspace A to be fit into view
- [x] Same as above, but for special workspaces

- [x] closing a window in a group (> 1 windows so doesn't destroy the group) fits the group into view

- [x] Opening Layer 3 program moves focus -- tested: (`grimblast`, `rofi`) ---- ADDITIONAL PROGRAMS TO TEST WELCOME - "4-" in in https://github.com/hyprwm/Hyprland/discussions/13562
- [x] Fullscreen on a window fits the window into view (the view is shifted after you unfullscreen)
- [x] Maximise on a window fits the window into view
- [x] `movefocus` (focus({direction}))  causes scrolling view to move while `follow_focus = 0`
- [x] `focuswindow` (focus({window}))  moves view even when `follow_focus = 0`

- [x] Switching to the next window in a group moves scrolling view
- [x] moving focus vis movefocus within groups moves scrolling view
- [x] moving windows in a group via scrolling the groupbar moves scrolling view
- [x] Moving a window to a group via picking it up with cursor moves scrolling view
- [x] selecting a window (and this bringing it to the top of the group/focusing it) via the cursor in a group moves scrolling view
- [x] `moveintogroup`, `movewindoworgroup` causes scrolling view move if follow_focus = 0
- [x] `movefocus`, `cyclenext`, `focuscurrentorlast`, `focuscurrentorurgent` moves scrolling view when `follow_focus = 0`

- ~~`hyprctl keyword` commands causes the currently focused window to be fit into view - "3-" in in https://github.com/hyprwm/Hyprland/discussions/13562~~ | dropped as is fixed with lua

<ins> TODO </ins>

- [x] Run tests 
- [x] make tests


erstarr:
As for adding tests: I would like to make a separate "test PR" to set in store what should move scrolling view and what should not, so as to avoid breakages in the future.

vaxerski:
add tests here

erstarr:
New behaviour change (bugfix): `focuswindow` no longer causes scrolling view to move while `follow_focus = 0`. Bug cuz `follow_focus = 0` should do what it says on the tin.

Now, I changed `FOCUS_REASON_KEYBIND` of that dispatch, but there is 14 more `FOCUS_REASON_KEYBIND`s in that source file alone...

To make non-hacky tests, this change is necessary.


if you want other `FOCUS_REASON_KEYBIND`s to also be changed; what do you think of getting rid of `FOCUS_REASON_KEYBIND`, `FOCUS_REASON_OTHER` and other generic focus reasons entirely, which will allow different layouts to treat different focus-related dispatches, actions differently as they see fit?
Though that'd result in quite a bit of `FOCUS_REASON`s.

I'd rather get this merged first tho, since the rest are more refactoring than bugfixes

erstarr:
The scope of this keeps expanding further and further...

New bugfix: continuation on the `follow_focus` should do what it says on the tin series.
Targeting Groups this time

1: changes in a group's current window no longer casues scrolling view shift when `follow_focus = 0`
What this implies is:
- Switching to the next window in a group does not move scrolling view
- moving focus via `movefocus` within groups does not move scrolling view
- moving windows in a group via scrolling the groupbar does not move scrolling view
- Moving a window to a group via picking it up with cursor does not move scrolling view (see 2: for keybinds)
- selecting a window (and this bringing it to the top of the group/focusing it) via the cursor in a group does not move scrolling view

2: `moveintogroup`, `movewindoworgroup` no longer causes scrolling view move if `follow_focus = 0`

erstarr:
Tests added

src/desktop/view/Group.hpp
```diff
@@ -15,10 +15,14 @@ namespace Desktop::View {
...
 
+        enum eRemoveFromGroupReason : uint8_t {
+            REMOVE_FROM_GROUP_REASON_UNMAP_WINDOW,
+        };
+
         bool                             has(PHLWINDOW w) const;
 
         void                             add(PHLWINDOW w);
-        void                             remove(PHLWINDOW w, Math::eDirection dir = Math::DIRECTION_DEFAULT);
+        void                             remove(PHLWINDOW w, Math::eDirection dir = Math::DIRECTION_DEFAULT, std::optional<eRemoveFromGroupReason> reason = std::nullopt);
```
vaxerski: add REASON_UNKNOWN and you can get rid of the optional wrapper

src/layout/LayoutManager.hpp
```diff
@@ -75,7 +82,7 @@ namespace Layout {
         void                             performSnap(Vector2D& sourcePos, Vector2D& sourceSize, SP<ITarget> target, eMouseBindMode mode, int corner, const Vector2D& beginSize);
 
         void                             invalidateMonitorGeometries(PHLMONITOR);
-        void                             recalculateMonitor(PHLMONITOR);
+        void                             recalculateMonitor(PHLMONITOR, std::optional<eRecalculateMonitorReason> reason = std::nullopt);
```
vaxerski: same here

src/layout/space/Space.hpp
```diff
@@ -12,6 +12,18 @@
...
+    enum eRecalculateReason : uint8_t {
+        RECALCULATE_REASON_WORKSPACE_CHANGE,
+        RECALCULATE_REASON_SPECIAL_WORKSPACE_TOGGLE,
+        RECALCULATE_REASON_HYPRCTL_KEYWORD,
+        RECALCULATE_REASON_TOGGLE_FULLSCREEN,
+        RECALCULATE_REASON_INVALIDATE_MONITOR_GEOMETRIES,
+        RECALCULATE_REASON_RENDER_MOINTOR,
+    };
+
+    std::optional<eRecalculateReason> recalcMonitorReasontoRecalcReason(CLayoutManager::eRecalculateMonitorReason reason);
```
vaxerski: and here

src/desktop/state/FocusState.hpp
vaxerski, line 9: hm, we get some overlap, maybe this would be better as a bitfield?

UNMAP could when be `UNMAP_TILING | UNMAP_FLOATING | UNMAP_WINDOW`, keybind could be `DISPATCH_FOCUS | DISPATCH_MOVEINTOGROUP | ...`

erstarr: but some keybinds are meant to not trigger INPUT_MODE_HARD (and thus result in a scrolling view change),  same with UNMAP_TILING and FLOATING

for example, closing a tiling window should move the scrolling view while closing a floating (focus fallback bug) should not.

There is indeed an overlap btw some of the newly added reasons and KEYBIND, DESKTOP_STATE_CHANGE, and needless to say OTHER. I can note that specific cases within their umbrella that are exceptions are separated, or if you prefer cleanliness: https://github.com/hyprwm/Hyprland/pull/13974#issuecomment-4191023759 

> if you want other FOCUS_REASON_KEYBINDs to also be changed; what do you think of getting rid of FOCUS_REASON_KEYBIND, FOCUS_REASON_OTHER and other generic focus reasons entirely, which will allow different layouts to treat different focus-related dispatches, actions differently as they see fit?
Though that'd result in quite a bit of FOCUS_REASONs.

This'd be a long term thing imo, and with dubious benefits (the cases where reasons must be made exceptions are rare so a note would probably suffice). I'm willing to work on this if you want this option.

erstarr:
@vaxerski 
Last problem remaining with this PR is [this](https://github.com/hyprwm/Hyprland/pull/13974#:~:text=0-,To%20me,want) and `movefocus` moving view with `follow_focus = 0`.

I had intended to address that with https://github.com/hyprwm/Hyprland/pull/13913 (after fixing the code ofc), but what to do [now](https://github.com/hyprwm/Hyprland/pull/13913#issuecomment-4232573614)? I can just implement the [change](https://github.com/hyprwm/Hyprland/pull/13913#issuecomment-4232591240:~:text=Personally%20I%20wouldn%27t%20mind%20shoving%20all%20the%20permissive%20behaviour%20%28%2313974%20%28comment%29%20%2C%20focuswindow%2C%20movefocus%29%20into%20%3D%201) here and close the other PR.

I'd personally prefer this. Adding another option to `follow_focus` that adds little seems unnecessary/hacky and the fix here would be 2 lines.

vaxerski:
its fine yes. As long as these changes only relate to scroll

erstarr:
New bugfix, focus moves via `movefocus`, `cyclenext`, `focuscurrentorlast`, `focuscurrentorurgent` no longer cause scrolling view to move when `follow_focus = false`. They still do when `follow_focus = true`.

This is done via a new focus reason that tracks if `fullWindowFocus` was called from `switchToWindow`, which all of the above (and only them) call.
I think this is alright, since the old behaviour handled the focus reason from the same function, and it's responsible for focus moves anyway; but if you want I can always add one more enum to track why `switchToWindow` was called.

One thing to note is that when you switch to another window when you are in fullscreen, scrolling view does move regardless of `follow_focus`.
Honestly, I never use any of the dispatches listed here to move scrolling tape, let alone to dispel fullscreen so this behaviour makes no difference to me personally.
The original code had a different code path for if these dispatches were dispatched when on fullscreen so I simply maintained that difference, but don't mind unifying the behaviour.

Tests for testing behaviour of `movefocus` when `follow_focus` is `true` and `false` are also added.

Also added tests for `movefocus` within groups, both `follow_focus = false` and `true` cases.

erstarr:
rebased on main. hopefully should be gtg

vaxerski:
pls wait for lua, tag me once lua is merged

erstarr:
Dropping `hyprctl keybind` (now `hl.config({...})) change.

With the lua update, not all config option changes now trigger a viewport change so a sweeping change like this is undesirable.

As long as `scroll_inhibitor` (https://github.com/hyprwm/Hyprland/pull/13955) is merged, users can decide if a certain config option's viewport change behaviour is undesirable and disable that selectively.

erstarr:
@vaxerski 
all changes and tests updated to work with lua

vaxerski:
your tests crash

erstarr:
> your tests crash

Fixed

hyprtester/src/tests/main/scroll.cpp
```diff
@@ -195,5 +195,952 @@ TEST_CASE(scrollWindowRule) {
...
+    // test pass
+    if (std::stoi(currentWindowPosX) < 0) {
+        NLog ::log("{}Passed: {}window of class 'a' has negative x coordinates for its position: {}", Colors ::GREEN, Colors::RESET, currentWindowPosX);
+    }
+    // test fail
+    else {
+        FAIL_TEST("{}Failed: {}window of class 'a' does not have negative x coordinates for its position: {}", Colors::RED, Colors::RESET, currentWindowPosX);
+    }
+
+    // clean up
```
vaxerski: you don't need to clean up in those refactored tests, it's done automagically

erstarr:
rebased. all test pass.

erstarr:
@vaxerski if no more problem with this, please merge. I lose a bit of my soul every time I rebase this onto main

vaxerski (approved):
now imagine how I feel rebasing shit every day man I need to do so too

vaxerski:
thanks for the work!

PR #13992, opened 2026-04-04
Part 3 of cpu optimizations

vaxerski:
Stuff

### DIFF
--- a/src/render/Renderer.hpp
+++ b/src/render/Renderer.hpp
@@ -191,6 +191,7 @@ namespace Render {
 
         SCMSettings                     getCMSettings(const NColorManagement::PImageDescription imageDescription, const NColorManagement::PImageDescription targetImageDescription,
                                                       SP<CWLSurfaceResource> surface = nullptr, bool modifySDR = false, float sdrMinLuminance = -1.0f, int sdrMaxLuminance = -1);
+        void                            clearCMSettingsCache();
         virtual bool                    reloadShaders(const std::string& path = "") = 0;
 
       protected:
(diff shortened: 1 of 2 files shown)


src/render/Renderer.hpp
```diff
@@ -216,11 +217,22 @@ namespace Render {
...
+
+        struct SCMSettingsCacheEntry {
+            uint64_t    srcDescId = 0, dstDescId = 0;
+            void*       surfacePtr      = nullptr; // read-only!!
+            bool        modifySDR       = false;
+            float       sdrMinLuminance = -1.F;
+            int         sdrMaxLuminance = -1;
+            SCMSettings settings;
+        };
+        std::vector<SCMSettingsCacheEntry> m_cmSettingsCache;
```
ItsOhen, line 229: Might want to do:

```
struct SCMSettingsKey {
    uint64_t srcDescId;
    uint64_t dstDescId;
    void* surfacePtr;
    bool     modifySDR;
    float    sdrMinLuminance;
    int      sdrMaxLuminance;

    auto operator<=>(const SCMSettingsKey&) const = default;
};

std::map<SCMSettingsKey, SCMSettings> m_cmSettingsCache;
```

vaxerski: cache will be so small it's faster to iterate over a vec

ItsOhen: Faster than O(1)? That i want to see! :D

umbrageodotus: Think of the constant. For the array, the constant is just a single access, which is smaller than the map's constant (hash + access). Not to mention memory layouts, etc.

vaxerski: hash is expensive here + cache locality on the cpu will take care of the lookups on a contiguous array

PR #14002, opened 2026-04-05
layouts/dwindle: override force after window drags

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/13718


tekstryder:
grātiās tibi agō

PR #14004, opened 2026-04-05
cursor: stop re-copying whole shm cursor

Pppp1116:
this reuses the same damage region for the shm texture update and the cpu-side cursor copy, so we only touch the bits that changed.


vaxerski:
thaat is not how hw cursors work

PR #14015, opened 2026-04-06
compositor: fix floating input/visual z-order desync after fullscreen

chocoblocko9:
<details>
<summary>EXTREMELY CONVOLUTED SETUP GO!!</summary>
OKAY SO, I have this setup where I use this script:

```shell
#!/usr/bin/env bash 

handle() {
  case $1 in
    activewindow*)
      window_class=$(echo "$1" | cut -d'>' -f3 | cut -d',' -f1)
      
      if [[ "$window_class" == "firefox" ]] then
        hyprctl dispatch alterzorder bottom,class:vesktop
      elif [[ "$window_class" == "vesktop" ]] then
        hyprctl dispatch alterzorder bottom,class:firefox
      fi
      ;;
  esac
}

socat -U - UNIX-CONNECT:"$XDG_RUNTIME_DIR/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket2.sock" | while read -r line; do
  handle "$line"
done
```

Which will auto change z order based on which window is hovered. Basically when toggling fullscreen on and off on one of those programs the visual and actual z order get desynced because m_createdOverFullscreen isn't cleared when alterzorder bottom is called. It's easiest to see in the videos.

If you're wondering why I use bottom and not top, it's so it doesn't drown out other floating windows easily.
</details>

TLDR: Z order gets de-synced when fullscreening floating windows and then clicking on the window to get focus. This fixes that, and you can see the before and after below.

Before:

https://github.com/user-attachments/assets/85480101-9b3a-4ee0-a246-60221cfcfa68

After:

https://github.com/user-attachments/assets/bb80e646-a528-488e-834d-dc08844a9641

Shouldn't break anything, as far as I can tell nothing else in the code calls alter z order bottom, only the dispatcher itself so it shouldn't have any weird consequences

Ready

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1292,6 +1292,8 @@ void CCompositor::changeWindowZOrder(PHLWINDOW pWindow, bool top) {
 
     if (top)
         pWindow->m_createdOverFullscreen = true;
+    else
+        pWindow->m_createdOverFullscreen = false;
 
     if (pWindow == (top ? m_windows.back() : m_windows.front()))
         return;


vaxerski:
I don't see a difference in the videos...

chocoblocko9:
> I don't see a difference in the videos...

In the first video, go to 0:10 and see that firefox is popping up over discord (the script is triggering) even though discord SHOULD be above firefox (and it is above it visually)

Then go to 0:14 in the second video. You can see that when i put my mouse in the same place as the first video, firefox doesnt pop up (the script doesnt trigger) which is intended because visually firefox is under discord

It's really hard to explain, just pay attention to **where my mouse** is when firefox becomes the active window (and calls alterzorder bottom on discord) at the end of both videos

chocoblocko9:
Sorry, these should be better videos showing just the bug with a socket2 running, you can see that activewindow is firing 'early' because the firefox hitbox doesn't match with its visual before my fix

Before: 

https://github.com/user-attachments/assets/fc4c63fb-0461-4bc4-9be7-371e06d1315d

After:

https://github.com/user-attachments/assets/afbe29c0-0b6e-4e34-a060-81316fab9d80

vaxerski (approved):
sure

chocoblocko9:
flaky test? seems unrelated to my change as its in keybinds

vaxerski:
yes it does that

ssareta:
Gotta say, that's a pretty interesting script/setup... reminds me of accordion mode from macos Aerospace

chocoblocko9:
> Gotta say, that's a pretty interesting script/setup... reminds me of accordion mode from macos Aerospace

LOL yeah its definitely unique. i have another script that runs on SUPER + R that resets the windows back to their position if i moved them, but if theyre in the right position they slide back so you can put tiled stuff underneath LOL, its kinda like a special workspace tbh

PR #14018, opened 2026-04-07
input: keep pointer focus on layer surfaces during keyboard refocus

cdump:
I use Hyprland with waybar. When I want to switch workspaces on waybar with the mouse wheel, only the first scroll works. After that, I need to move the mouse at least one pixel to make scrolling work again.

waybar config:
```json
  "hyprland/workspaces" : {
    "format": "{name}",
    "on-scroll-up": "hyprctl dispatch workspace e-1",
    "on-scroll-down": "hyprctl dispatch workspace e+1",
  },
```
hyprland:
```
input {
  follow_mouse = 0
}
```

Fix pointer focus being stolen from layer-shell surfaces like waybar when `input:follow_mouse = 0`.
                                                                    
Before this change, focusing a window on workspace switch could call `sendMotionEventsToFocused()` and move pointer focus from waybar to the newly focused window even if the cursor was still over the bar.

This change makes `sendMotionEventsToFocused()` keep the current pointer focus when it is already on a non-window surface (for example waybar or a popup), so keyboard focus can move to the workspace window while pointer focus stays on the bar under the cursor. 

It's ready, I use patched version for a few days without any issues

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -7,6 +7,7 @@
 #define private public
 #include <src/managers/input/InputManager.hpp>
 #include <src/managers/PointerManager.hpp>
+#include <src/managers/SeatManager.hpp>
 #include <src/managers/input/trackpad/TrackpadGestures.hpp>
 #include <src/helpers/Monitor.hpp>
 #include <src/desktop/rule/windowRule/WindowRuleEffectContainer.hpp>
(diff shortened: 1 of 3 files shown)


vaxerski (commented):
lgtm, test please

cdump:
@vaxerski 

> test please

could you clarify what you mean? I use the patched version every day and don't see any issues. Are you requesting more specific tests, and if so, from whom?

vaxerski:
I meant an integration test. Check hyprtester/

cdump:
I've rebased my changes onto the latest main branch. I ran hyprtester, and no regressions were found. I also tested the rebased version on my desktop, and there were no issues there either.

vaxerski:
yes I mean _adding_ a test to hyprtester to test for _this issue_ not happening again

cdump:
I took a look into the hyprtester codebase and found that figuring out how to write a proper test for this would take an enormous amount of time for me, which I don’t have right now. So please don't expect this part from my side.

cdump:
@vaxerski I tried to add test and rebased to the latest main

vaxerski:
ah when you rebased main was a bit broken, could you rebase again?

cdump:
rebased, but test fails now, will take a look and ping when ready

cdump:
@vaxerski fixed, rebased and ready for review

vaxerski (approved):
thanks!

PR #14019, opened 2026-04-07
refactor(layout): use enum-based layout IDs for layout matching

B2krobbery:
Replaces string comparisons in updateWorkspaceLayouts with enum-based ID matching.

Previously, layout matching relied on comparing strings retrieved from m_algoNames.
This change introduces a LayoutID enum along with a cached type → ID mapping (m_algoIDs),
allowing comparisons to be expressed using enums while preserving behavior.

This also resolves the existing TODO about replacing string-based layout matching with an ID-based approach.

The layout string is converted to LayoutID once per iteration, and the workspace loop
compares enum values when possible. For layouts not covered by the enum (e.g. plugin
layouts), the code falls back to the original string comparison to maintain compatibility.

Testing:
- Verified layout switching between "dwindle", "master", "scrolling", and "monocle" using hyprctl
- Tested with multiple windows to confirm correct layout behavior
- Tested workspace switching and config reload
- Verified plugin layouts (e.g. grid via hyprpm) work correctly with fallback
- No crashes or regressions observed during runtime

Performance:
A synthetic benchmark simulating updateWorkspaceLayouts shows that performance is similar
or slightly slower due to the added conversion and branching.

Since this function is not a hot path, the real-world impact is negligible.

The primary goal of this change is to simplify comparison logic and make layout matching
more explicit while preserving existing behavior.

Benchmark code:
https://gist.github.com/B2krobbery/7fc2aa6abc7d6f21f2f750493224d3f4

### DIFF
--- a/src/layout/supplementary/WorkspaceAlgoMatcher.cpp
+++ b/src/layout/supplementary/WorkspaceAlgoMatcher.cpp
@@ -151,8 +151,14 @@ void CWorkspaceAlgoMatcher::updateWorkspaceLayouts() {
         const auto type = &typeid(*TILED_ALGO.get());
 
         const auto it = m_algoIDs.find(type);
-        if (it != m_algoIDs.end() && it->second == layoutID)
-            continue;
+        if (it != m_algoIDs.end()) {
+            if (layoutID != LayoutID::UNKNOWN && it->second == layoutID)
+                continue;
+
+            // fallback for unknown layouts (plugins)
+            if (layoutID == LayoutID::UNKNOWN && m_algoNames.contains(type) && m_algoNames.at(type) == layoutStr)
+                continue;
+        }
 
         // needs a switchup
         ws->m_space->algorithm()->updateTiledAlgo(algoForNameTiled(layoutStr));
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
what about plugins?

B2krobbery:
Plugin layouts are handled via a fallback.
For layouts not covered by the enum (i.e. plugins), layoutIDFromString returns UNKNOWN. In that case, the code falls back to the original string comparison using m_algoNames, preserving the previous behavior.
So plugin-registered layouts continue to work exactly as before.

Also tested with a plugin layout (grid via hyprpm). Switching between plugin and built-in layouts works correctly and the layout is applied as expected.

B2krobbery:
I ran additional benchmarks after the initial tests.
Results were consistent across runs --the enum-based path is either similar or slightly slower
(~10–15% in synthetic benchmarks), mainly due to the added conversion and branching.
Since updateWorkspaceLayouts is not a hot path, the real-world impact remains negligible.
So this change should be considered more as a refactor for clarity rather than a performance optimization.

B2krobbery:
Based on discussion, this introduces two different paths (enum-based matching for built-in layouts and string fallback for plugin layouts), which may add some complexity.
String comparisons in this context are already relatively inexpensive, and the existing approach keeps behavior consistent across both built-in and plugin layouts.
It might be worth considering whether keeping a single consistent mechanism is preferable for simplicity and maintainability.

B2krobbery:
I talked with a friend about it, and from his perspective string comparisons here are already pretty cheap

vaxerski:
I dont think there is a point here if the path is the same

B2krobbery:
yeah that makes sense
given that the behavior is the same and there's no real performance benefit
I'll close this for now 👍

PR #14025, opened 2026-04-07
fix: removed unncessary(I wish it's) pointer focus hack in activate()…

therealga3an:
…, it's my first time in a codeBase so...


vaxerski:
why would you think its unnecessary?

therealga3an:
I probably just didn't understood the block well or smth I will try to read the codebase more sorry for bothering

umbrageodotus:
@therealga3an close the pr

Issue #10488, opened 2025-05-20
Make the system bell protocol produce a sound by default


vaxerski:
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/10428

## Description
we should make the bell produce a sound by default, probably overridable with a path. Also needs a `none` setting for a silent bell (handling via socket2 / no handling)

onion108:
Looking forward for this to be implemented.

PR #14036, opened 2026-04-09
bell: play a sound by default

thompty:
## Summary

This makes `xdg-system-bell-v1` produce a local sound by default while keeping the existing socket2 `bell` event behavior intact.

It adds a new `misc:bell_sound` config option with support for:

- `default`: play the themed system bell sound locally
- `none`: stay silent locally but still emit the socket2 `bell` event
- custom path: play a user-provided sound file

## Changes

- add `misc:bell_sound` config registration
- add config description/documentation
- add example config entries
- keep emitting `bell` over socket2 as before
- play the default bell locally via `libcanberra`
- support custom sound file playback
- add `libcanberra` to CMake and Nix dependencies

## Testing

Tested with a local debug build in a separate headless Hyprland instance using a small `xdg-system-bell-v1` test client.

Verified:

- `misc:bell_sound = default` emits `bell>>` and plays a sound
- `misc:bell_sound = none` emits `bell>>` and stays silent

## Notes

- invalid custom sound paths currently fall back to the default bell sound
- this change is intended to address #10488

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,7 +268,8 @@ pkg_check_modules(
   gio-2.0
   re2
   muparser
-  lcms2)
+  lcms2
+  libcanberra)
 
 find_package(hyprwayland-scanner 0.3.10 REQUIRED)
 
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -192,6 +192,9 @@ master {
 misc {
     force_default_wallpaper = -1 # Set to 0 or 1 to disable the anime mascot wallpapers
     disable_hyprland_logo = false # If true disables the random hyprland logo / anime girl background. :(
+    # bell_sound = default
+    # bell_sound = none
+    # bell_sound = ~/.config/hypr/bell.oga
 }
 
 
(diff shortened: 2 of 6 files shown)


vaxerski (commented):
any sound libs that do not link gtk?

KramLololo:
> any sound libs that do not link gtk?

TL;DR Everywhere I've looked it is indicated that gtk is optional and also disabled by default.

Entertaining the idea for alternatives, libcanberra is in fact the only implementation of the xdg sound theme spec. After a whole evening of searching for alternatives, the only actual result was a fork of the same library created by the cafe-desktop organization which cannot benefit us as it only adds extensions to support said desktop. The two other wayland compositors who support the system bell protocol both use libcanberra.

With that said, creating an in-house alternative would obviously require a way to search for the bell audio, create our own if none exists, define an audio-backend interface, create drivers for every audio backend and then probe for the one installed by the current user. Given that such a replacement is exactly identical to libcanberra, I believe it would generate unnecessary work for the sake of a simple bell sound and would be in spite of the fact that package maintainers have already made the right decision to disable gtk as a dependency by default.

vaxerski:
that's okay then. Rebase on main and we can get to reviewing :)

KramLololo:
A new pull request has been opened due to the unavailability of the author: #15502

PR #14039, opened 2026-04-10
input: avoid repeated weak_ptr::lock() and ensure consistent usage

B2krobbery:
m_solitaryClient and m_self are weak_ptrs and were accessed via
multiple lock() calls within the same condition.

While unlikely, the underlying object could expire between calls,
leading to inconsistent evaluation. This change locks them once and
reuses the resulting shared_ptrs to ensure a consistent snapshot.
No functional changes intended.
Tested:
-Built Hyprland successfully
-Tested cursor movement across windows and monitors
-Tested rapid mouse movement and dragging operations
-Verified no regressions in cursor behavior or focus handling
-No crashes or performance issues observed


vaxerski (approved):
sure

PR #14043, opened 2026-04-10
Reverse slide direction

QuarkVortexx:
Hello,

This is my first contribution to the project, so hopefully everything is in check! Overall, I'm very new to the OSS space. I read the PR guidelines and tried to follow them appropriately.

Previously, the slide animation direction was fixed. This meant that with slidevert animation, the "first" workspace always felt to be at the top of the stack. Alternatively, in regular slide, the "first" workspace always felt as if it was the most left workspace.

For me, this created confusion with the slidevert animation, as mentally for me, the first workspace is the lowest in the stack. So I decided to create a setting for reversing the slide direction to allow for new options. I also thought that this was a great addition to the vast configurability of Hyprland.

So instead of simply requesting for the feature, I thought it's better to create it and share it with others by including a PR.

I'm new to the project, so my understanding of the project code base may not be the best. By my understanding, the feature should not break compatibility, nor did I find any bugs during my testing. 

The feature should be ready for merging.

### DIFF
--- a/src/config/legacy/ConfigManager.cpp
+++ b/src/config/legacy/ConfigManager.cpp
@@ -682,6 +682,7 @@ CConfigManager::CConfigManager() {
 
     registerConfigVar("animations:enabled", Hyprlang::INT{1});
     registerConfigVar("animations:workspace_wraparound", Hyprlang::INT{0});
+    registerConfigVar("animations:reverse_slide_direction", Hyprlang::INT{0});
 
     registerConfigVar("input:follow_mouse", Hyprlang::INT{1});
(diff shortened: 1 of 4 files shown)


Kommynct:
Can't you already do this by passing a -1?

QuarkVortexx:
> Can't you already do this by passing a -1?

Oh right, this was a great find. I assume you mean, that using movement percentage (https://wiki.hypr.land/Configuring/Animations/#extras), and set it to -100, you're able to reverse the direction? I didn't realize this before. 

However, after testing, it seems like this doesn't work with gestures. I'll rework my commits based on that!

QuarkVortexx:
Didn't have time to rework my commits. Closing the PR until able to check back in later.

PR #14046, opened 2026-04-11
Fix/direct scanout format vrr

imperishableSecret:
## What does it fix/add?

**Tri-state `quirks:skip_non_kms_dmabuf_formats`:** Updated the quirk from a boolean to an integer tri-state (0: off, 1: on, 2: auto). The default is now 2 (auto), which enables the quirk automatically on nvidia systems to prevent reporting non-KMS-compatible formats that often lead to DS failures.

**Same-buffer DS Fixes:**

- ~~**VRR Keep-alive Throttle:** Implemented a throttle for same-buffer direct scanout when VRR is active using the global `cursor:min_refresh_rate` config key. This prevents unnecessary KMS commits from unrelated compositor activity.~~

- **VRR Keep-alive Throttle (per-monitor):** Replaced global `cursor:min_refresh_rate` with a per-monitor `vrr_min_hz` option in both `monitor=` keyword syntax and `monitorv2` blocks (default 24 Hz). Keepalive threshold is now per-output and wired through `isVrrKeepaliveDue()`. Timer reset moved into the `onPresented` callback so the keepalive interval is measured from actual KMS page-flip acknowledgement rather than render start.

- **DS Cursor Commit Suppression Fix:** Extracted `shouldSuppressCursorCommit()` as a pure predicate (no side effects) from `shouldSkipScheduleFrameOnMouseEvent()`. The direct-scanout same-buffer path now gates cursor commits on this predicate, so `cursor:no_break_fs_vrr` suppression is enforced at KMS commit time. Previously, a cursor move arriving in the µs window between a page-flip and the next game-frame submission would set `m_scanoutNeedsCursorUpdate` and trigger a spurious atomic commit on the next same-buffer DS pass, causing a VRR rate spike regardless of `no_break_fs_vrr` being active.

- **FIFO unlock on no-commit path:** When no new KMS commit is needed, the logic now explicitly unlocks the FIFO protocol. This prevents the application from stalling its frame queue when it renders at a rate matching the display, avoiding redundant `presented()` events and unnecessary KMS commits.

- **Direct Scanout Format Validation:** Added explicit validation of buffer format and modifier pairs against the output's scanout capabilities. Unsupported formats are now caught early in `isDSBlocked()` (reporting unsupported format/modifier) instead of failing at KMS commit time.

- **`CMonitor::applyMonitorRule`** refactored to avoid unnecessary resets of geometry-dependent resources during format-only changes.

### Breaking Changes

- `quirks:skip_non_kms_dmabuf_formats` is now an integer rather than a boolean. The default is now `2` (auto-enable on nvidia).
- `cursor:min_refresh_rate` is **removed**. Replace with `vrr_min_hz = <hz>` inside a `monitor=` rule or `monitorv2` block.

## Ready for merging?

Ready for merging


src/helpers/Monitor.cpp
```diff
@@ -1923,6 +1928,27 @@ uint16_t CMonitor::isDSBlocked(bool full) {
         reasons |= DS_BLOCK_DMA;
         if (!full)
             return reasons;
+    } else {
+        // Verify the buffer's format+modifier is supported by the output for scanout.
```
UjinT34: Better cache it somehow. E.g. raise a flag when `attemptDirectScanout` gets a successful test commit and reset it when params change or ds is deactivated. Might be a bit tricky if `m_output->getRenderFormats` changes.

imperishableSecret: Replaced the per-frame DS format loop with a DRM-authoritative cache

umbrageodotus:
shouldn't min_refresh_rate now gtfo out cursor cuz it's no longer just for cursor

imperishableSecret:
> shouldn't min_refresh_rate now gtfo out cursor cuz it's no longer just for cursor

I agree, I migrated it to monitor config so it can be configured per monitor.

PR #14054, opened 2026-04-12
scrolling: sync hidden state with viewport visibility

imperishableSecret:
* Mark off-screen columns as hidden so hyprctl clients, w[v(N)] workspace selectors, app suspension, and focus-cycle exclusion reflect actual viewport state. setHidden(false) is immediate; setHidden(true) is deferred to the position-anim end-callback to preserve exit animations. Unhide on removeTarget and layout destruction to prevent windows being left hidden.
* By proxy Off-screen apps receive xdg_toplevel.suspended to throttle/pause render loops

* ~~with this change the windows out of the viewport are hidden, so they will not focus untill they are unhidden, so if someone used a keybind for shift foccus left/right to navigate the scrolling layout would need to migrate that to move +/-col~~

ready for merge


vaxerski:
> Mark off-screen columns as hidden so hyprctl clients, w[tv(N)] workspace selectors, app suspension, and focus-cycle exclusion reflect actual viewport state. setHidden(false) is immediate; setHidden(true) is deferred to the position-anim end-callback to preserve exit animations. Unhide on removeTarget and layout destruction to prevent windows being left hidden.

this is undesirable. These selectors work on the amount of windows on the workspace, and offscreen windows are on the workspace too.

imperishableSecret:
> > Mark off-screen columns as hidden so hyprctl clients, w[tv(N)] workspace selectors, app suspension, and focus-cycle exclusion reflect actual viewport state. setHidden(false) is immediate; setHidden(true) is deferred to the position-anim end-callback to preserve exit animations. Unhide on removeTarget and layout destruction to prevent windows being left hidden.
> 
> this is undesirable. These selectors work on the amount of windows on the workspace, and offscreen windows are on the workspace too.

My description included w[tv(N)] as an example which might be misleading:
* this change only affects the visible count, in this case (scrolling) windows in teh viewport, for the tiled window count it remains the same which would be the total windows on the workspace. 
* This change corrects the visible count whcih was previously the total number of windows on the scrolling workspace to the actual number of windows in the viewport.

Using setHidden(true) signals that off-viewport windows are visually unpresented, mirroring the semantics of the Monocle layout where windows not on top are hidden.

For example, I had a rule for if a window is the only one visible it needs no borders or gaps, i set 

windowrule {
    name = no-border-when-only-visible-tiled-window-on-workspace
    match:workspace = w[tv1]
    match:float = false
    border_size = 0
    rounding = 0
    no_shadow = on
}


this should semantically make the window which i fit on my viewpoert with layoutmsg fit active, borderless and have no rounded corners, but crrently, w[v1] ans also (w[tv1]) matches the total number of windows in the scrolling workspace, which in my opinion is incorrect since only one window is actually visible.

vaxerski:
> this should semantically make the window which i fit on my viewpoert with layoutmsg fit active, borderless and have no rounded corners, but crrently, w[v1] ans also (w[tv1]) matches the total number of windows in the scrolling workspace, which in my opinion is incorrect since only one window is actually visible.

and I disagree, all of them are visible, some are just offscreen

imperishableSecret:
> > this should semantically make the window which i fit on my viewpoert with layoutmsg fit active, borderless and have no rounded corners, but crrently, w[v1] ans also (w[tv1]) matches the total number of windows in the scrolling workspace, which in my opinion is incorrect since only one window is actually visible.
> 
> and I disagree, all of them are visible, some are just offscreen

How do we define visibility here ? 
Layouts before scrolling had no case where the windows would be not visible and not "hidden" by another window overlapping them, so all windows are actually on screen, scrolling now adds another aspect to it where it introduces another way of being "not visible" which is being off screen shouldn't this add another way of being hidden which would be moving outside the viewport. 

layoutmsg fit visible -> fits the actual visible windows in the viewport, so shouldn't the workspace rules match this semantic or vice versa?

vaxerski:
visible as in not hidden by e.g. a group or swallowed

PR #14055, opened 2026-04-12
X11-apps scaling per workspace while having force_zero_scaling = true

moimart:
**Describe your PR, what does fix/add?**

These changes add xwaylandscale: workspace rule for per-workspace X11 scaling when force_zero_scaling = true. 

The main need for this is having X11-based games to work under native resolution, for that a user might configure force_zero_scaling = true ... but then the user might have a collection of X11-based applications for which zero scaling show really small UI components. Some of these apps, like DaVinci Resolve, become almost unusable. With this, you can keep playing X11 games at their native resolution and use certain X11 apps with a specific scale in a given workspace.
                                                                                                                                                 
Config example: workspace = 5, xwaylandscale:3                                                                                                 
   
 **Is there anything you want to mention?**                                                                                                         
                                                            
  - Requires xwayland:force_zero_scaling = true. Logs a warning if the rule is used without it.                                                  
  - X11 apps will have some softness due to texture upscaling — there's no way around this without per-window X11 DPI support.
  - mergeWorkspaceRules moved from private to public static for unit test access.                                                                
  - Includes gtest coverage for scale resolution, merge logic, and the effective scale formula.                                                  
                                                                                                                                                 
 I tested this on a 7680x2160 ultrawide at scale 1.6666 with force_zero_scaling = true. Builds clean, 13/13 unit tests pass. But definitely this could use more testing on different monitor configurations and scale combinations.


moimart:
made a mistake. apologies

PR #14056, opened 2026-04-12
X11-apps scaling per workspace while having force_zero_scaling = true

moimart:
**Describe your PR, what does fix/add?**

These changes add xwaylandscale: workspace rule for per-workspace X11 scaling when force_zero_scaling = true.

The main need for this is having X11-based games to work under native resolution, for that a user might configure force_zero_scaling = true ... but then the user might have a collection of X11-based applications for which zero scaling show really small UI components. Some of these apps, like DaVinci Resolve, become almost unusable. With this, you can keep playing X11 games at their native resolution and use certain X11 apps with a specific scale in a given workspace.

Config example: workspace = 5, xwaylandscale:3

**Is there anything you want to mention?**

Requires xwayland:force_zero_scaling = true. Logs a warning if the rule is used without it X11 apps will have some softness due to texture upscaling — there's no way around this without per-window X11 DPI support. 

Includes gtest coverage for scale resolution, merge logic, and the effective scale formula.

I tested this on a 7680x2160 ultrawide at scale 1.6666 with force_zero_scaling = true. Builds clean, 13/13 unit tests pass. But definitely this could use more testing on different monitor configurations and scale combinations.

### DIFF
--- a/src/config/legacy/ConfigManager.cpp
+++ b/src/config/legacy/ConfigManager.cpp
@@ -1282,6 +1282,21 @@ void CConfigManager::postConfigReload(const Hyprlang::CParseResult& result) {
             m->m_activeWorkspace->m_space->recalculate();
     }
 
+    // refresh per-workspace X11 scale overrides from rules, then reconfigure X11 windows
+    for (auto const& wsRef : g_pCompositor->getWorkspaces()) {
+        const auto ws = wsRef.lock();
+        if (!ws || ws->inert())
+            continue;
+        const auto RULE = Config::workspaceRuleMgr()->getWorkspaceRuleFor(ws);
+        ws->m_xwaylandTargetScale = RULE.has_value() ? RULE->m_xwaylandScale.value_or(0.f) : 0.f;
+    }
+    for (auto const& w : g_pCompositor->m_windows) {
+        if (!w->m_isX11 || !w->m_isMapped || w->isHidden())
+            continue;
+        w->updateX11SurfaceScale();
+        w->sendWindowSize(true);
+    }
+
     // Update the keyboard layout to the cfg'd one if this is not the first launch
     if (!m_isFirstLaunch) {
         g_pInputManager->setKeyboardLayout();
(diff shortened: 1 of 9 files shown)


vaxerski (commented):
no this is awful, as far as I understand you just adjust the size, this will desync quickly and cause issues with xwayland. You need to remember XWayland keeps an internal X state just like a regular X server and may deny events in cases of overlap etc

moimart:
Yeah. Legit concern now that I thought about it after your comment. I had implemented another approach for a full monitor scale change for given configured workspaces.... however, that had small black screen transition and of course it applied to all windows. I found some prior art in a branch but I got it working.... however, my use case was for X11 apps only and not necessarily for all windows being scaled... would there be any interest in that?

vaxerski:
you simply cant do that, xwayland doesnt allow this. I would just give up because any hack you implement will inevitably mess up the state

moimart:
Understood and I agree... how about dynamically recalculating the XWayland screen for the given monitor and telling this new size via XDG ouput? Yes. This would still mess up the state in different ways:

1) X11 apps that cache screen size: Some X11 apps query the root window size once at startup and never check again. After a workspace switch changes the XWayland screen, those apps still think the screen is the old size. Tooltips, menus, or fullscreen calculations can land in the wrong place. This affects apps that don't handle ConfigureNotify on the root window... but this recalculation of the xwl screen gets only triggered when changing to the given workspace... potentially a caveat users can live with.
                    
2) Non-visible X11 windows overflow: When you switch to workspace 5 (smaller XWL screen), X11 windows on workspace 1 were configured for the larger screen. They now overflow XWayland's virtual screen until you switch back. XWayland clips them, so it's not a crash risk, but the X server's internal state temporarily has windows larger than the root window. This might be a killer for all this, I get it.

However, this would be a obscure configuration for people that don't want to reloading the whole config when they want to use an specific app; potentially saving time and providing a better user experience... we just abstract the reload of the config in a workspace change

I have an implementation of the above if there is any interest... Otherwise, I'll just keep patching my hyprland. (and that is fine tbh)

Thanks!

moimart:
Closing as this approach is actually quite bad (However, it'd be nice to get an answer to the last comment if possible; I opened a draft PR with another approach)

vaxerski:
don't update the X11 state, everything tends to get messed up esp since X apps can reposition themselves

PR #14057, opened 2026-04-12
cleanup: avoid repeated weak_ptr lock() calls in conditions

B2krobbery:
Avoid repeated weak_ptr::lock() calls in conditions.
Stores the result of lock() to avoid redundant calls and ensure consistent usage.
Build: successful locally

### DIFF
--- a/src/desktop/view/LayerSurface.cpp
+++ b/src/desktop/view/LayerSurface.cpp
@@ -388,7 +388,10 @@ void CLayerSurface::onCommit() {
         if (!WASEXCLUSIVE && ISEXCLUSIVE)
             g_pInputManager->m_exclusiveLSes.push_back(m_self);
         else if (WASEXCLUSIVE && !ISEXCLUSIVE)
-            std::erase_if(g_pInputManager->m_exclusiveLSes, [this](const auto& other) { return !other.lock() || other.lock() == m_self.lock(); });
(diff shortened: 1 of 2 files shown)


src/desktop/view/LayerSurface.cpp
```diff
@@ -388,7 +388,10 @@ void CLayerSurface::onCommit() {
         if (!WASEXCLUSIVE && ISEXCLUSIVE)
             g_pInputManager->m_exclusiveLSes.push_back(m_self);
         else if (WASEXCLUSIVE && !ISEXCLUSIVE)
-            std::erase_if(g_pInputManager->m_exclusiveLSes, [this](const auto& other) { return !other.lock() || other.lock() == m_self.lock(); });
+            std::erase_if(g_pInputManager->m_exclusiveLSes, [this](const auto& other) {
```
vaxerski: this can be simplified altogether to `return !other || other == m_self`

B2krobbery: Thanks Updated accordingly.

B2krobbery: since this is a weak_ptr, wouldn’t removing .lock() reduce safety?😅
or is it fine in this case because the lifetime is guaranteed?

vaxerski: you can always check a weak ptr

vaxerski (approved):

PR #14059, opened 2026-04-12
algo/scroll: add center for centering the current col

vaxerski:
ref #14027

@nemect is this what you wanted?


nemect:
I guess? I would need to compile hyprland to test this. But these two questions would convince me :)

1. If i have three 50% columns with focus on first (third is hidden on right edge). Does `center` centers the first column? For my script this does work only for all other columns but first.
2. Does it work with direction:down?

If yes i would call this is exactly what i wanted and thank you very much for fast implementation. :)

vaxerski:
> I guess? I would need to compile hyprland to test this

well please do thats the point

nemect:
I have compiled and tried. It does nothing. It correctly says `ok` after `hyprctl dispatch` so i believe it call the right switch case but windows stays the same.

vaxerski:
oops mb fixed

PR #14065, opened 2026-04-13
monitor: ensure swapchain is updated before mode test

B2krobbery:
This may be related to Aquamarine issue # 268:https://github.com/hyprwm/aquamarine/issues/268.

When testing modes, test() may be called before the swapchain is
updated to match the new mode size. This can result in a mismatch
between the framebuffer and the requested mode, causing atomic
test failures.

Ensure updateSwapchain() is called after setMode/setCustomMode
and before test(), so the buffer matches the mode being tested.

Builds successfully locally.

Tested by repeatedly switching monitor modes (1920x1080 <-> 1280x720)
and reloading the config multiple times.

Verified monitor state using hyprctl monitors after each change.

Checked DRM logs; no atomic commit or plane mismatch errors were observed.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -623,8 +623,6 @@ void CMonitor::applyCMType(NCMType::eCMType cmType, NTransferFunction::eTF cmSdr
     }
 }
 
-
-
 bool CMonitor::applyMonitorRule(Config::CMonitorRule&& pMonitorRule, bool force) {
 
     static auto PDISABLESCALECHECKS = CConfigValue<Hyprlang::INT>("debug:disable_scale_checks");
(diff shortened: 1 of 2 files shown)


vaxerski (commented):
I think you missed a few places, maybe just add a helper fn to set the mode and update swapchain?

B2krobbery:
thx for pointing it out
refactored it into a helper and updated all mode-setting paths

vaxerski (commented):
wouldnt it make more sense in `CMonitorState`?

B2krobbery:
yeah 
I put it in CMonitor since all the calls are there, but can move it to CMonitorState if that’s better

B2krobbery:
moved it to CMonitorState and updated all call sites accordingly

vaxerski (approved):
tnx

PR #14070, opened 2026-04-13
FP16 fixes

UjinT34:
Fix some issues with FP16 and tonemapping
Adds `render:fp16_sdr_tf` to control internal tf for FP16 work buffers. 0 - default (linear atm), 1 - linear, 2 - matches monitor (srgb or gamma22)

By default HDR triggers FP16 and MRT for screensharing. MRT breaks cached blur unless `render:use_shader_blur_blend = 1` is used. The latter option doesn't work with rotated screens and lags behind during fast window moving/resizing.

Ready

- [x] fix tonemapping shader
- [x] apply tonemapping only when needed
- [x] buffer invalidation on nvidia
  - [x] fix blur
  - [x] fix random pixel corruption on the damage edge
- [x] screen share color stability
- [x] screen share tonemapping
- [x] simpler cm
  - [x] render rect
  - [x] render glow
  - [x] render shadow
  - [x] cache converted colors
  - [ ] ~~use not normalised fp16~~
- [x] check windows scRGB handling
- [x] ICC with fp16 and blur
- [x] screen share
  - [x] sdr
  - [x] hdr


gulafaran:
There is some kind of brightness being applied to a no background glcleared buffer its gray on fp16 but black once disabled. Not sure what to blame that on. Or is that intentional. Might be the cause for the slight uptick in blur opacity/brightness aswell

gulafaran:
> There is some kind of brightness being applied to a no background glcleared buffer its gray on fp16 but black once disabled. Not sure what to blame that on. Or is that intentional. Might be the cause for the slight uptick in blur opacity/brightness aswell

Kind of bright gray flashes because of that on login

fxzzi:
i noticed the lighter gray bg too. ended up just setting `misc:background_color = 0x000000` and it's been fine.

by default it's not pure black, its 0x111111

though keep in mind this causes damage issues on screencopy. the background is considered transparent when you do this if you don't run anything for wallpaper.

also everything on this PR has a blue tint for me, but I like blue so its okei :)

umbrageodotus:
for the blur artifact fix in here, you should prolly do a diff approach: ignoring the pixels outside the rendered area

umbrageodotus:
also, honestly, you should prolly use my tonemapper from discord. its much more accurate

fxzzi:
Screenshot from 26 04 14 02:25:05
Screenshot from 26 04 14 02:25:10

blue tint shows up in screenshots as well, can see easily on the banner images

gulafaran:
gotta clear the render buffer aswell. and while we are at it use the more modern glClearBufferfv instead.

```
diff --git a/src/render/OpenGL.cpp b/src/render/OpenGL.cpp
index 0e36a4cd..5ee1008c 100644
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -755,7 +755,9 @@ void CHyprOpenGLImpl::begin(PHLMONITOR pMonitor, const CRegion& damage_, SP<IFra
         applyScreenShader(*PSHADER);
     }
 
-    g_pHyprRenderer->bindFB(g_pHyprRenderer->m_renderData.pMonitor->resources()->getUnusedWorkBuffer());
+    const auto unusedFb = g_pHyprRenderer->m_renderData.pMonitor->resources()->getUnusedWorkBuffer();
+    g_pHyprRenderer->bindFB(unusedFb);
+    GLFB(unusedFb)->clearAfterInvalidation();
     m_offloadedFramebuffer = true;
 
     g_pHyprRenderer->m_renderData.mainFB = g_pHyprRenderer->m_renderData.currentFB;
diff --git a/src/render/gl/GLFramebuffer.cpp b/src/render/gl/GLFramebuffer.cpp
index b60dbb6c..d3d13521 100644
--- a/src/render/gl/GLFramebuffer.cpp
+++ b/src/render/gl/GLFramebuffer.cpp
@@ -183,7 +183,7 @@ void CGLFramebuffer::clearAfterInvalidation() {
         return;
 
     m_cleared = true;
-    glClearColor(0, 0, 0, 0);
+    const std::array<GLfloat, 4> c = {0.f, 0.f, 0.f, 0.f};
     g_pHyprOpenGL->scissor(nullptr);
-    glClear(GL_COLOR_BUFFER_BIT);
+    glClearBufferfv(GL_COLOR, 0, c.data());
 }
 ```

**Comment by vaxerski** [comment]
With fp16, when I have layer animations enabled, `slurp -c 83352ecc -b ffcce622` is weird when fading out. It saturates very heavily.

Also, slurp in general is much more saturated on fp16. Without, it's like a light overlay. With it, it's waaay more opaque and strong.

**Comment by fxzzi** [comment]
Was mentioned on the previous PR. It's something to do with the fade specifically. Tested without blur and extra stuff and could still reproduce it

**Comment by vaxerski** [comment]
I mean blur is also weird.

fp16:
image

no fp16:
image

it's much brighter with fp16, but that could also be that my surface's colors are getting fucked. It's supposed to be darkened by the bar background.

**Comment by fxzzi** [comment]
Yeah, blur brightness (or overall opacity?) is different on fp16. Blur brighness is 1 by default. Setting to 0.625 on fp16 instead was good enough for me during my testing

https://gitlab.com/fazzi/nixohess/-/blob/main/modules/programs/hyprland/settings.nix?ref_type=heads#L153

**Comment by vaxerski** [comment]
fade outs fixed

**Comment by vaxerski** [comment]
I added SDR alpha compat for SDR surfaces, should fix most alpha issues.

@UjinT34 can you check if the logic is correct? I don't think it's wrong (result looks good) but I aint no hdr expert. Shouldn't touch anything but sdr surfs

плз энд санк ю

**Comment by UjinT34** [comment]
I'll look into it later. Currently linear fb has a fixed max luminance of 10k which means that any source isn't tonemapped when it's rendered into work fb. But work fb is guaranteed to be tonemapped when it's rendered to output. That bug isn't noticeable on main because tonemapping shader is broken and doesn't do much.
To fix it linear fb luminance range should be set to output's luminance range so tonemapping will be applied only to hdr sources with a wider luminance range when it's rendered to work fb and work fb will go to output without tonemapping. That'll preserve sdr colors for sdr apps and will produce the same output for hdr apps. Apps like chromium might not produce a perfect sdr output if they do some internal tonemapping or set an image description with a wider range.
That should fix most of the color issues.
Some oversaturated stuff probably doesn't go through CM at all. At least rectangle shader needs a fix. It can use the same approach as the other shaders but it'll make more sense to implement all the CM math in hyprgraphics (or use some exisitng lib for it) and modify the rectangle color before the shader and cache that value.

**Comment by vaxerski** [comment]
I've pushed a commit to use a BT2020 work space with G2.2 for SDR. HDR untouched. Everything looks primo on my end. @fxzzi could check

**Comment by umbrageodotus** [comment]
@vaxerski ujin gonna kill u for not using linear lmfao

**Comment by vaxerski** [comment]
it's what KDE does too. Linear for HDR (which makes sense) while G2.2 for SDR so that blending is correct.

**Comment by umbrageodotus** [comment]
alr then, just buy some armor off temu tho cuz i still dunno if ujin thinks following kde is correct here

**Comment by UjinT34** [comment]
Shouldn't matter much with fp16. It should provide enough precision to convert between sdr representations without any noticeable change. Those conversions use linear as an intermediate representations. Using linear for workbuffer helps with performance when source, work buffer and target can't skip CM entirely. Just need to be careful which descriptions are used and when.
Unmodified mirror copy expects workbuffer to be in sRGB space and doesn't convert primaries. If it's used when the work buffer is fp16 bt2020 you'll get washed out colors (it shouldn't be used in this case though if the conditions are correct)

**Comment by fxzzi** [comment]
fp16 on main is fine but it's specifically this PR causing that saturated screencopy in DS I mentioned in #hyprland-dev

non_shader_cm 2 has different behaviour to non_shader_cm 1 even though they both trigger DS successfully for me

seems to happen only in DS situations with `XBGR8888` format. That means anything vulkan as far as I know. opengl stuff is using `XRGB8888` and I don't notice the saturation there.

**Comment by 3ventic** [comment]
Compared #13860 to this (as of "quad cm" c17f0f7 with #14102 cherry-picked on top), this seems to produce washed out or overly bright screen capture again in both desktop and fullscreen HDR game content with fp16 enabled and relevant configuration as
```
monitorv2 { # OLED primary LS27FG812SUXEN
	output = DP-2
	mode = 3840x2160@120
	position = 3840x0
	scale = 1
	bitdepth = 8
	vrr = 1
	cm = hdredid
	sdr_min_luminance = 0.005
}

render {
	cm_enabled = true
	# Off to prevent fullscreen games from setting monitor back to SDR
	cm_auto_hdr = 0 # 0 off 1 cm,hdr 2 cm,hdredid
	use_fp16 = 1
	keep_unmodified_copy = 2
}
```

Without the changes from this PR I was getting a more true to life png from grim (but had issues with blur, and fp16 in general was unstable with nvidia, both seem much improved here)

With the above configuration:
image

FP16 off (over saturated and contrasted, but expected?):
image

Same bar on SDR monitor with cm off:
image

**Comment by fxzzi** [comment]
yeah i can confirm that. `grim -T` works fine though. Similar to my issue mentioned above though that one only happens in direct scanout. I don't use direct scanout on mpv.

Screenshot from 26 04 27 13:40:19
Screenshot from 26 04 27 13:40:26

**Comment by 3ventic** [comment]
render:direct_scanout doesn't seem to impact the screen capture output for me

**Comment by fxzzi** [comment]
yeah, the HDR screencopy bug you mentioned above happens with DS 0 as I explained.

the other issue i reported before is specifically DS

**Comment by fxzzi** [comment]
screenshot in ds looks fine now with latest changes. tested both fp16_sdr_tf 1 and 2. will probably stick with 2 for now for better blur brightness levels

edit: hdr screencopy still cooked tho, both wit `hyprctl eval 'hl.config({ render = {fp16_sdr_tf = 1} })'` and 2

**Comment by fxzzi** [comment]
image

blur is being invalidated again after leaving HDR mpv. 

```lua
  render = {
    cm_auto_hdr = 2, -- use values from edid for HDR
    use_fp16 = 1,
    fp16_sdr_tf = 2,
  },
  debug = {
    invalidate_fp16 = 1,
  },
```

3ventic:
The washed out screen capture is only present for me when sdr_min_luminance/sdr_max_luminance/sdrbrightness/sdrsaturation are used. Lowering the min luminance is kind of necessary for OLED to look right with its blacks though. (They also seem to cause the portal to eventually, typically within minutes, start returning a broken frame that never updates again, but that's probably unrelated to this PR..)

They also seem to impact screenshots of fullscreen HDR content (regardless of DS/autohdr settings), which I wouldn't expect?

vaxerski:
@UjinT34 did you revert my changes for FP16? Opacity's fucked again.

fxzzi:
> @UjinT34 did you revert my changes for FP16? Opacity's fucked again.

`render.fp16_sdr_tf = 2`

vaxerski:
m ok that should default to 2 not 0

vaxerski:
changed the defaults - fp16_sdr_tf -> 0 / 1 (monitor / linear) default 0
use_fp16 -> 0/1/2 (off, on, auto) -> auto = use if monitor not srgb

vaxerski:
is this acceptable ujin?

UjinT34:
> @UjinT34 did you revert my changes for FP16? Opacity's fucked again.

wdym? some screensharing stuff might've caused it. should keep it this way and fix opacity in some other way. or the screensharing itself should be changed to use a full additional render instead of reusing existing buffers.

`default` was meant to explicitly set the value to "let HL decide". should have the same effect as not setting it at all.

not sure about `isSRGB` check. it'll enable fp16 by default on a wider range of configurations. eventually it should be the default for any CM setup but the feature isn't tested well enough for that rn.

vaxerski:
> wdym? some screensharing stuff might've caused it. should keep it this way and fix opacity in some other way. or the screensharing itself should be changed to use a full additional render instead of reusing existing buffers.

it was the TF. Alpha will look (and IMO, behave) wrong with linear. GAMMA22 is ok.

> not sure about isSRGB check. it'll enable fp16 by default on a wider range of configurations. eventually it should be the default for any CM setup but the feature isn't tested well enough for that rn.

That's the point here. I mean, it does look just fine on my end, and with this merged over a few days if we don't see any reports of it being wrong, it should be fine.

vaxerski (approved):
oppan gangnam style

thanks!

gulafaran:
I think only issue i have with fp16 both with or without this PR is the weird gray flash on start of HL almost like some buffer thats cleared to black is rendered as gray for some reason

fxzzi:
already mentioned this above, by default the background is NOT pure black. 

https://github.com/hyprwm/Hyprland/pull/14070#issuecomment-4239614745

Setting to pure black does indeed look fine, and honestly provides a nicer transition from tty to Hyprland.

gulafaran:
> i noticed the lighter gray bg too. ended up just setting `misc:background_color = 0x000000` and it's been fine.
> 
> by default it's not pure black, its 0x111111
> 
> though keep in mind this causes damage issues on screencopy. the background is considered transparent when you do this if you don't run anything for wallpaper.
> 
> also everything on this PR has a blue tint for me, but I like blue so its okei :)

oh damn, hm. why is this more noticeable on fp16 then ujint8 , thanks. il change it.

PR #14075, opened 2026-04-14
render: survive recoverable GPU resets instead of aborting

kiivihal:
## Summary
Hyprland currently `RASSERT`s (SIGABRT) whenever it observes a GPU reset signal: `glGetGraphicsResetStatus()` in `begin()`/`beginSimple()`, `GL_CONTEXT_LOST` in `end()`, or an invalid texture in `renderTextureInternal()`. The original abort message acknowledges this is incomplete: *"Cannot continue until proper GPU reset handling is implemented."*

On recent Intel (Xe driver, Meteor Lake / Arc graphics), TLB-invalidation timeouts produce recoverable GPU resets: the kernel re-arms the rendering engines, the EGL display survives, and userspace is expected to re-establish its context. The current behaviour kills the whole Hyprland session on what the kernel considers a handled fault — observed in the wild especially around suspend/resume and memory pressure.

This PR implements that recovery path and the null-safety it depends on. It's structured as three reviewable commits:

1. **`core/Compositor: null-guard CWLSurfaceResource against expired m_resource`** — `good()` itself segfaults if `m_resource` is null; fix that and have the forwarding methods (`enter`, `leave`, `sendPreferredTransform/Scale`, `id`, `error`) respect `good()` before dereferencing. Also null-check `pSurface` in `CCompositor::setPreferredScale/TransformForSurface`.

2. **`render: null-guard blur framebuffer accesses`** — `m_blurFB` can be null during monitor (re)init or after a GPU reset invalidates framebuffer objects. `getBlurTexture`, `blurMainFramebuffer`, `preBlurForCurrentMonitor`, and `renderRectWithBlurInternal` all bail gracefully instead of segfaulting. Skipping a blur pass is preferable to taking the session down.

3. **`render: replace fatal aborts on GPU reset with EGL context recovery`** — The main change: adds `attemptContextReset()` (unbinds + rebinds the EGL context, resets `m_shadersInitialized`) and an `m_gpuResetCooldown` frame counter. `begin()`/`beginSimple()` attempt recovery; on success, skip one frame to re-init; on failure, cooldown 60 frames to avoid tight-looping. `end()` no longer aborts on `GL_CONTEXT_LOST`, it just schedules recovery. `renderTextureInternal()` skips invalid-texture draws (post-reset, textures need reupload).

Commits 1 and 2 are the null-guards that surface during the recovery window opened by commit 3; without commit 3 they'd be unreachable in healthy operation.

## Motivation
Framework 13 / Intel Core Ultra (Meteor Lake, `8086:7d55`) on the Xe driver, kernels 6.18+. In the worst stretches the compositor was crashing multiple times per day — the machine became effectively unusable for real work, which is what prompted me to start working on these fixes rather than just reporting the issue. Each crash dropped me back to the TTY mid-task, with every open application gone. The stack always terminated in `CHyprRenderer::renderMonitor` → `__stack_chk_fail` (the `RASSERT`), while `dmesg` showed the kernel having cleanly reset the GPU and resumed — only Hyprland was dying on what was, from the kernel's perspective, a handled fault.

## Test plan
- [x] Builds cleanly against current `main`.
- [x] Running locally on a Framework 13 / Xe / Hyprland setup since early April 2026. GPU resets that previously aborted the compositor now produce a short log burst (`"GPU reset detected... Attempting EGL context recovery"`) followed by normal rendering. No session loss since the recovery path was enabled.
- [ ] Reviewer verification: (a) that `m_shadersInitialized = false` is sufficient state-reset, (b) that 60 frames is a reasonable cooldown (vs. e.g. time-based), (c) that the `pMonitor.reset()` on skipped frames doesn't break any invariants downstream.

## Scope limits / open questions
- Texture and shader state beyond the shader-init flag isn't explicitly reset; in practice the per-frame render pipeline re-creates what it needs. I'd welcome pointers to anything that caches GPU state across frames that would also need invalidation.
- 60-frame cooldown was chosen empirically (≈1 second at 60Hz). Happy to make it configurable or swap to a timestamp.
- Happy to split into separate PRs if preferred, though the three are causally linked.

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2904,6 +2904,9 @@ void CCompositor::leaveUnsafeState() {
 }
 
 void CCompositor::setPreferredScaleForSurface(SP<CWLSurfaceResource> pSurface, double scale) {
+    if (!pSurface)
+        return;
+
     PROTO::fractional->sendScale(pSurface, scale);
     pSurface->sendPreferredScale(std::ceil(scale));
 
@@ -2918,6 +2921,9 @@ void CCompositor::setPreferredScaleForSurface(SP<CWLSurfaceResource> pSurface, d
 }
 
 void CCompositor::setPreferredTransformForSurface(SP<CWLSurfaceResource> pSurface, wl_output_transform transform) {
+    if (!pSurface)
+        return;
+
     pSurface->sendPreferredTransform(transform);
 
     const auto PSURFACE = Desktop::View::CWLSurface::fromResource(pSurface);
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -264,7 +264,7 @@ SP<CWLSurfaceResource> CWLSurfaceResource::fromResource(wl_resource* res) {
 }
 
 bool CWLSurfaceResource::good() {
-    return m_resource->resource();
+    return m_resource && m_resource->resource();
 }
 
 wl_client* CWLSurfaceResource::client() {
@@ -272,6 +272,9 @@ wl_client* CWLSurfaceResource::client() {
 }
 
 void CWLSurfaceResource::enter(PHLMONITOR monitor) {
+    if (!good())
+        return;
+
     if (std::ranges::find(m_enteredOutputs, monitor) != m_enteredOutputs.end())
         return;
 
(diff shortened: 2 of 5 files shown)


vaxerski:
umm https://github.com/hyprwm/Hyprland/pull/13828 ?

vaxerski:
> Framework 13 / Intel Core Ultra (Meteor Lake, 8086:7d55) on the Xe driver, kernels 6.18+. In the worst stretches the compositor was crashing multiple times per day — the machine became effectively unusable for real work, which is what prompted me to start working on these fixes rather than just reporting the issue. Each crash dropped me back to the TTY mid-task, with every open application gone. The stack always terminated in CHyprRenderer::renderMonitor → __stack_chk_fail (the RASSERT), while dmesg showed the kernel having cleanly reset the GPU and resumed — only Hyprland was dying on what was, from the kernel's perspective, a handled fault.

Worth noting almost all of your apps will crash and can't handle resets

PR #14082, opened 2026-04-14
init: drop CAP_SYS_NICE from ambient set after gaining SCHED_RR

ErrorNoInternet:

Add a NixOS-specific fix that prevents all applications running under Hyprland from inheriting CAP_SYS_NICE.

There is no error check here since it's basically useless. Most other distributions will not put CAP_SYS_NICE in the ambient set, only the effective and permitted sets probably, which doesn't have the same behavior.

ref https://github.com/NixOS/nixpkgs/pull/507419

N/A

Ready

### DIFF
--- a/src/init/initHelpers.cpp
+++ b/src/init/initHelpers.cpp
@@ -1,3 +1,6 @@
+#include <linux/capability.h>
+#include <sys/prctl.h>
+
 #include "initHelpers.hpp"
 
 bool NInit::isSudo() {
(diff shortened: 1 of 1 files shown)


ErrorNoInternet:
how tf did this base on a commit from 2 years ago

vaxerski:
> how tf did this base on a commit from 2 years ago

long before time had a name, the first spinjitsu master was born

src/init/initHelpers.cpp
```diff
@@ -21,6 +24,8 @@ void NInit::gainRealTime() {
         return;
     }
 
+    prctl(PR_CAP_AMBIENT, PR_CAP_AMBIENT_LOWER, CAP_SYS_NICE, 0, 0);
```
vaxerski, line 29: would be nice to add a comment describing the purpose of this being os-specific

vaxerski (approved):

PR #14086, opened 2026-04-15
Fix `movewindoworgroup` when moving from group to group

kolayne:

Fixes `CKeybindManager::moveWindowIntoGroup` to remove a window from a group before attempting to add it to another group. Addresses #13843.

The animation of moving a window from a group into another group now looks weird: as if the whole target group is being moved.

I am not sure how to address this or whether we should address this.

Depends on your opinion on the above.


vaxerski (approved):
tnx

PR #13785, opened 2026-03-19
Cleanup the entire config infrastructure

vaxerski:
This is sorely overdue, plus required if we don't want our lua/whatever config to suck ass.

Needs testing, shouldn't break a thing. (I mean, plugins might.)

PR #14089, opened 2026-04-16
socket2: add back changefloatingmode and togglegroup events

ErrorNoInternet:

https://github.com/hyprwm/Hyprland/pull/12890 nuked `changefloatingmode` and `togglegroup` events from IPC. add them back. 

fixes #13721

(i honestly have no idea if i did this right)

i didnt add an event for groups to EventBus because the old togglegroup events also didn't emit anything and i dont know if we'll ever use it internally

should be ready, tested and works

### DIFF
--- a/src/desktop/view/Group.cpp
+++ b/src/desktop/view/Group.cpp
@@ -9,6 +9,7 @@
 #include "../../layout/LayoutManager.hpp"
 #include "../../desktop/state/FocusState.hpp"
 #include "../../Compositor.hpp"
+#include "managers/EventManager.hpp"
 
 #include <algorithm>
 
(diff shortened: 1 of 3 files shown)


javalsai:
Comparing the source code with the documented IPC events at https://wiki.hypr.land/IPC/, found that `ignoregrouplock` has also been missing since 8726a73 (#13785).

As far as I checked, all the other documented events are present in the codebase.

---

It was commented out with the note `FIXME: this is no longer possible like this. It's redundant anyways. Can be easily scripted / lua'd`, and then completely removed in 5ba33f84 (#13817).

src/desktop/view/Group.cpp
```diff
@@ -60,17 +61,15 @@ void CGroup::init() {
...
 
 void CGroup::destroy() {
-    while (true) {
-        if (m_windows.size() == 1) {
-            remove(m_windows.at(0).lock());
-            break;
-        }
+    g_pEventManager->postEvent(SHyprIPCEvent({.event = "togglegroup", .data = std::format("0,{:x}", rc<uintptr_t>(m_windows.at(0).get()))}));
 
+    while (!m_windows.empty())
```
vaxerski: excuse me what is this doing here

ErrorNoInternet: whar

vaxerski: why did you change the destruction method

ErrorNoInternet: reverted, but doesnt this do the same thing

vaxerski (approved):

src/desktop/view/Group.cpp
vaxerski: yes

PR #14132, opened 2026-04-20
keybinds: add repeat delay and rate to binds config

lapLapych:
Add new config options `binds:repeat_delay` and `binds:repeat_rate` (INT, 0-inf, default fall back to values from input configuration). I just don't want to use input variables because they would also change values for text typing.

https://github.com/user-attachments/assets/28d1d525-095e-419f-a58e-064dae397171

No known issues.

Please, let me know if documentation should be added first.

I'm also considering adding the same options for submaps, but not sure if that's good idea. I would be happy to receive feedback.

### DIFF
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -501,6 +501,8 @@ std::vector<SP<IValue>> Values::getConfigValues() {
         MS<Bool>("binds:allow_pin_fullscreen", "Allows fullscreen to pinned windows, and restore their pinned status afterwards", false),
         MS<Int>("binds:drag_threshold", "Movement threshold in pixels for window dragging and c/g bind flags. 0 to disable.", 0,
                 {.min = 0, .max = std::numeric_limits<int>::max()}),
+        MS<Int>("binds:repeat_rate", "if 0, will use input:repeat_rate / how frequently bind repeats", 0),
+        MS<Int>("binds:repeat_delay", "if 0, will use input:repeat_delay / delay before bind starts repeating", 0),
 
         /*
          * xwayland:
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -501,8 +501,8 @@ std::vector<SP<IValue>> Values::getConfigValues() {
         MS<Bool>("binds:allow_pin_fullscreen", "Allows fullscreen to pinned windows, and restore their pinned status afterwards", false),
         MS<Int>("binds:drag_threshold", "Movement threshold in pixels for window dragging and c/g bind flags. 0 to disable.", 0,
                 {.min = 0, .max = std::numeric_limits<int>::max()}),
-        MS<Int>("binds:repeat_rate", "if 0, will use input:repeat_rate / how frequently bind repeats", 0),
-        MS<Int>("binds:repeat_delay", "if 0, will use input:repeat_delay / delay before bind starts repeating", 0),
+        MS<Int>("binds:repeat_rate", "if 0, will use input:repeat_rate / how frequently bind repeats", 0, {.min = 0, .max = 200}),
+        MS<Int>("binds:repeat_delay", "if 0, will use input:repeat_delay / delay before bind starts repeating", 0, {.min = 0, .max = 2000}),
 
         /*
          * xwayland:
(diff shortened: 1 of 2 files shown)


ItsOhen:
I'm confused.. How is this different from input:repeat_rate and input:repeat_delay with binde?

lapLapych:
@ItsOhen
Because input settings affect all input, including typing text. I only want to change the behavior for binde.
Thanks for the question

Kommynct:
Shouldn't this be a per-bind flag?

lapLapych:
> Shouldn't this be a per-bind flag?

It could be. That's why i'm thinking it might be better if add per submap configuration. For example:
```
submap = example
submap {
  binds {
    repeat_rate = 40
  }
}
```

This isn't only way it could look, but I hope you understand the idea.

But I'm still not sure is it good idea.

Kommynct:
my usecase is that for volume control I want a lower rate and no delay, I don't want to switch to a submap for this, if it was a per-bind flag that would also work for submaps, it'd be simpler and cover more usecases

lapLapych:
Good idea. I will think about implementing it with flags later.

In my case I just want all repeat binds to have one repeat rate and delay without affecting my text input via `input:repeat_rate` and `input:repeat_delay`.

Kommynct:
One could also use a passthrough no-op to make for example ctrl-w not repeat selectively with the bind flag, for another usecase

lapLapych:
here video test for .lua config

https://github.com/user-attachments/assets/4745fc3c-0c38-4ce9-8d0c-1c74a0f801c1

and i noticed an error on the [submap wiki](https://wiki.hypr.land/Configuring/Basics/Binds/#submaps): instead of `hl.resize` it should be `hl.dsp.window.resize`

src/config/values/ConfigValues.cpp
```diff
@@ -499,6 +499,8 @@ std::vector<SP<IValue>> Values::getConfigValues() {
         MS<Bool>("binds:allow_pin_fullscreen", "Allows fullscreen to pinned windows, and restore their pinned status afterwards", false),
         MS<Int>("binds:drag_threshold", "Movement threshold in pixels for window dragging and c/g bind flags. 0 to disable.", 0,
                 {.min = 0, .max = std::numeric_limits<int>::max()}),
+        MS<Int>("binds:repeat_rate", "if 0, will use input:repeat_rate / how frequently bind repeats", 0),
+        MS<Int>("binds:repeat_delay", "if 0, will use input:repeat_delay / delay before bind starts repeating", 0),
```
vaxerski: missing min max values

wabarTM:
> > Shouldn't this be a per-bind flag?
> 
> It could be. That's why i'm thinking it might be better if add per submap configuration. For example:
> ```
> submap = example
> submap {
>   binds {
>     repeat_rate = 40
>   }
> }
> ```
> 
> This isn't only way it could look, but I hope you understand the idea.
> 
> But I'm still not sure is it good idea.

just add them as flags to all binds not just submaps?

src/config/values/ConfigValues.cpp
lapLapych: fixed but i'm not sure how it works, because when i tested it didn't do anything. i also tried testing input min/max values but those didn't work neither

lapLapych:
> > > Shouldn't this be a per-bind flag?
> > 
> > 
> > It could be. That's why i'm thinking it might be better if add per submap configuration. For example:
> > ```
> > submap = example
> > submap {
> >   binds {
> >     repeat_rate = 40
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This isn't only way it could look, but I hope you understand the idea.
> > But I'm still not sure is it good idea.
> 
> just add them as flags to all binds not just submaps?

i'm a little bit confused about how we three calling all these things right now (it's my fault thank you guys for feedback i'm glad to see it).

we have e flag(or repeating in lua) that stands for repeat in the wiki [old](https://wiki.hypr.land/0.54.0/Configuring/Binds/#bind-flags) [new](https://wiki.hypr.land/Configuring/Basics/Binds/#bind-flags). i think you want parameters like `repeat_delay=INT` and `repeat_rate=INT` i would be happy to see them too. I tried to implement them yesterday but unfortunately for now it didn't work.

vaxerski (approved):
good. Tests needed

PR #14166, opened 2026-04-26
xdg-foreign-v2: Keep invalid imported objects alive

freevatar:
#### Fixes: https://github.com/hyprwm/Hyprland/discussions/14017
Changes:
1. Validate toplevel surfaces correctly before exporting or parenting
2. Keep invalid imported objects alive according to spec

Before:
* Hyprland sent `destroyed()` for invalid handle
* Hyprland removed imported object immediately
* `set_parent_of()` later hits removed compositor-side state
* `portal-gtk` disconnected preventing file dialog from opening

Incorrect flow:
```
import_toplevel(invalid_handle)
  -> create zxdg_imported_v2
  -> compositor notices that exported handle is invalid
  -> send destroyed()
  -> remove imported object immediately
  -> client still owns the object
  -> later set_parent_of() hits removed compositor-side state
  -> portal backend disconnects
  -> file dialog fails to open
```

Now:
* invalid imports are kept alive after `destroyed()`
* `set_parent_of()` is ignored after invalidation
* object is removed only when client destroys it

Correct flow:
```
import_toplevel(invalid_handle)
  -> create zxdg_imported_v2
  -> send destroyed()
  -> mark the imported handle invalid
  -> keep object alive
  -> ignore set_parent_of() after invalidation
  -> wait for client destroy
```

There are more protocol issues but they are unrelated to this regression and should be addressed separately

Ready

### DIFF
--- a/src/protocols/XDGForeignV2.cpp
+++ b/src/protocols/XDGForeignV2.cpp
@@ -46,6 +46,19 @@ std::string_view CXDGExportedResourceV2::handle() const {
 
 CXDGForeignExporterProtocolV2::CXDGForeignExporterProtocolV2(const wl_interface* iface, const int& ver, const std::string& name) : IWaylandProtocol(iface, ver, name) {}
 
+static SP<CXDGToplevelResource> xdgToplevelFromWlSurface(wl_resource* surface) {
+    const auto wlSurf = CWLSurfaceResource::fromResource(surface);
+
+    if (!wlSurf || !wlSurf->m_role || wlSurf->m_role->role() != SURFACE_ROLE_XDG_SHELL)
+        return nullptr;
+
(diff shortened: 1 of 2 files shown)


freevatar:
Looks like there is also a bug in `mediawriter`. Its portal dialog defaults `winId=0` then builds Wayland parent as `wayland:0` instead of a real xdg-foreign handle:
https://github.com/FedoraQt/MediaWriter/blob/fa604ccd37b99d5e8c7b6ec0af950d9a45600233/src/app/portalfiledialog.h#L47

https://github.com/FedoraQt/MediaWriter/blob/fa604ccd37b99d5e8c7b6ec0af950d9a45600233/src/app/portalfiledialog.cpp#L160

But anyway Hyprland still needs to handle invalid imports safely according to `xdg-foreign-v2` spec so PR is valid.

vaxerski (approved):
thanks!

PR #14168, opened 2026-04-26
monitor: set format back after failing DS activation

fxzzi:

we do run m_lastScanout.reset(), but m_lastScanout might sometimes not be set. cache format from before attempting DS, and set back to it on activation failure.

again not exactly sure on whether this fixes, as its hard to find repro for the specific issue of format getting stuck on XBGR8888 after leaving DS. maybe this will help

sure

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -1991,7 +1991,8 @@ bool CMonitor::attemptDirectScanout() {
     if (m_lastScanout.expired())
         m_prevDrmFormat = m_drmFormat;
 
-    const bool NEEDS_TEST = !m_lastScanout || m_drmFormat != params.format; // do not retest while it's active
+    const auto PREV_FORMAT = m_drmFormat;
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
tnx

UjinT34:
Does it change anything? DS blockers include format checks and if test or commit fails then there is a bug somewhere earlier. And there should a similar format restoration somewhere later. Might be too late or removed by accident.
Nice to have a fallback here but in general it's better to avoid failures somehow because if it happens once then it'll likely happen on every next frame and might trigger an unnecessary modeset due to a raised format change flag (even without a real change after restoration). So if it makes a difference the issue should be investigated further.

fxzzi:
I'm running the PR as we speak, as mentioned in the OP I wasn't able to find reliable repro for the bug. But it could also be related to Hyprland pretty much dying 50% of the time if you do `hyprctl reload` whilst in DS.

PR #14169, opened 2026-04-26
screencopy: clear stale m_lastScanout when theres a pending frame

fxzzi:

fixes the scenario when using SHM screencopy in DS and having the output be more vibrant than what is visible on screen. explicitly reset m_lastScanout when there's a pending frame.

specifically noticed it with `non_shader_cm` 2. on 1 it worked fine either way

before, toplevel and shm screencopy would look different in DS. below shows minecraft. 2nd image is what is visible on screen and therefore the correct one.
image
image

after this change both `grim -o DP-3` and `grim -T <stableID>` look the same in DS scenarios.

i think so


fxzzi:
draft for now, i thought it fixed it but it didn't. still a bug and am looking for fix but am dumb

fxzzi:
it's an issue with fp16, will leave this with ujin.

PR #14173, opened 2026-04-26
tests: tolerate plugin config mismatch in CI

B2krobbery:
The plugin test currently fails in CI due to a mismatch between config values and descriptions, which seems to be influenced by recent changes rather than the test logic itself.
This change makes the test tolerate that specific non-critical mismatch while still failing on other unexpected errors, keeping the test meaningful while avoiding false negatives in CI.

### DIFF
--- a/hyprtester/src/tests/plugin/plugin.cpp
+++ b/hyprtester/src/tests/plugin/plugin.cpp
@@ -11,18 +11,6 @@
 #include "../shared.hpp"
 
 bool testPlugin() {
-    const auto RESPONSE = getFromSocket("/eval hl.plugin.test.test()");
-
-    if (RESPONSE != "ok") {
-        // allow known non-critical mismatch in CI
-        if (RESPONSE.find("config value number mismatches descriptions size") != std::string::npos) {
-            return true;
-        }
-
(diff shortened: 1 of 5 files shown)


hyprtester/src/tests/plugin/plugin.cpp
```diff
@@ -14,9 +14,15 @@ bool testPlugin() {
     const auto RESPONSE = getFromSocket("/eval hl.plugin.test.test()");
 
     if (RESPONSE != "ok") {
+        // allow known non-critical mismatch in CI
+        if (RESPONSE.find("config value number mismatches descriptions size") != std::string::npos) {
+            return true;
```
vaxerski: this one test should be removed, I just forgot to do it

B2krobbery: oh ok, I can remove that test

vaxerski: I meant just `config value number mismatches descriptions size` not the entire plugin test!!!!!!1!

B2krobbery: ohh got it, I misunderstood  I’ll restore the test and just remove that specific mismatch check 😅

vaxerski (approved):
tnx

PR #14174, opened 2026-04-26
lua: install the default example hyprland.lua

gulafaran:
install the default lua example aswell.


vaxerski:
you broke cmajke

gulafaran:
> you broke cmajke

oh indeed, added one to many )

vaxerski (approved):

PR #14177, opened 2026-04-27
Add missing permissions entry into lua example config

erstarr:

Add missing permissions entry into lua example config with the goal of achieving parity with the hyprlang example config.

no

yes

### DIFF
--- a/example/hyprland.lua
+++ b/example/hyprland.lua
@@ -63,7 +63,7 @@ hl.env("HYPRCURSOR_SIZE", "24")
 ----- PERMISSIONS -----
 -----------------------
 
--- See https://wiki.hypr.land/Configuring/Permissions/
+-- See https://wiki.hypr.land/Configuring/Advanced-and-Cool/Permissions/
 -- Please note permission changes here require a Hyprland restart and are not applied on-the-fly
 -- for security reasons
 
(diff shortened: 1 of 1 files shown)


example/hyprland.lua
```diff
@@ -58,6 +58,26 @@ local menu        = "hyprlauncher"
 hl.env("XCURSOR_SIZE", "24")
 hl.env("HYPRCURSOR_SIZE", "24")
 
+
+-----------------------
+----- PERMISSIONS -----
+-----------------------
+
+-- See https://wiki.hypr.land/Configuring/Permissions/
```
chocoblocko9: should be https://wiki.hypr.land/Configuring/Advanced-and-Cool/Permissions/

erstarr: you're right. just copy pasted from the old config.

vaxerski:
why? We don't put everything in the example config and specifically tell the user to rtfm

erstarr:
hyprlang example config had/still has it

If you want to remove this part from the example config going forward, that's ok too.

erstarr:
saying that, I think it's a good idea to have an example for every category of variables in the example. without this, ecosystem is missing

vaxerski (approved):
ok

PR #14179, opened 2026-04-27
Add `[[nodiscard]]` attributes to API methods

nikosavola:

This PR adds [`[[nodiscard]]` attributes](https://en.cppreference.com/cpp/language/attributes/nodiscard) to many applicable functions across core components, primarily targeting the public Plugin API, the Compositor, and the Lua ConfigManager.

Ready for merging, adding attributes should change only downstream compiler, LSP, linter etc. warnings but not any behavior.

### DIFF
--- a/src/Compositor.hpp
+++ b/src/Compositor.hpp
@@ -74,58 +74,58 @@ class CCompositor {
 
     // ------------------------------------------------- //
 
-    void                   removeWindowFromVectorSafe(PHLWINDOW);
-    PHLWINDOW              vectorToWindowUnified(const Vector2D&, uint16_t properties, PHLWINDOW pIgnoreWindow = nullptr);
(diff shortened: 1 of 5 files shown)


vaxerski:
why these 3 places specifically? Also, expected<> is nodiscard by default at least under clang-tidy

nikosavola:
> why these 3 places specifically? Also, expected<> is nodiscard by default at least under clang-tidy

I happened to be looking at these three sections of the code so I limited the scope to these.

Looks like `std::expected<>` is not marked nodiscard on all toolchains yet https://quuxplusone.github.io/blog/2024/12/08/should-expected-be-nodiscard/ but I guess we could remove the nodiscard from those her.

vaxerski:
it is on gcc and that's what I care about tbh xD

PR #14181, opened 2026-04-27
Improve error reporting during initialization in `main.cpp`

nikosavola:

Improve error reporting during initialization in `main.cpp` by throwing `std::runtime_error`s with simple descriptions instead of just `std::exception()`


vaxerski (commented):
these arent much more helpful without specifying a line or something.

vaxerski (approved):

PR #14183, opened 2026-04-27
dwindle: don't warn when splitting with only 1 window

fxzzi:
don't warn when splitting with only 1 window
image

no


fxzzi:
Closing as fixed on main

PR #14191, opened 2026-04-28
Rename 'allow' field to 'mode' in PermissionSpec

mxCynic:
I noticed this differs from what the wiki says, and I believe the wiki version is correct.


vaxerski:
that is an autogenerated file

PR #14192, opened 2026-04-28
config/errors: Report and categorize errors properly for actions

vaxerski:
Makes all errors categorized, some not displayed, some displayed. All errors can be obtained after hl.dispatch().

@kolayne

### DIFF
--- a/src/config/shared/actions/ConfigActions.cpp
+++ b/src/config/shared/actions/ConfigActions.cpp
@@ -131,7 +131,7 @@ ActionResult Actions::closeWindow(std::optional<PHLWINDOW> w) {
         return {};
 
     if (window->m_closeableSince > Time::steadyNow())
-        return std::unexpected("can't close window, it's not closeable yet (noclosefor)");
+        return actionError("can't close window, it's not closeable yet (noclosefor)", eActionErrorLevel::WARNING, eActionErrorCode::INVALID_STATE);
 
     window->sendClose();
 
@@ -217,7 +217,7 @@ ActionResult Actions::pinWindow(eTogglableAction action, std::optional<PHLWINDOW
         return {};
 
     if (!window->m_isFloating || window->isFullscreen())
-        return {};
+        return actionError("Window does not qualify to be pinned", eActionErrorLevel::WARNING, eActionErrorCode::INVALID_STATE);
 
     bool wantPin = false;
     switch (action) {
@@ -289,7 +289,7 @@ ActionResult Actions::moveToWorkspace(PHLWORKSPACE ws, bool silent, std::optiona
         return {};
 
     if (!ws)
-        return {};
+        return std::unexpected("Invalid workspace");
 
     if (ws->m_id == window->workspaceID())
         return {};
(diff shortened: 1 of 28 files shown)


B2krobbery:
The failing test is:
getFromSocket("/dispatch hl.dsp.focus({ direction = 'left' })")
expected not ok, got ok.
So it looks like focus dispatch no longer returns failure when the action can't be performed, which breaks the existing test expectation.

src/config/shared/actions/ConfigActions.cpp
```diff
@@ -373,11 +373,11 @@ ActionResult Actions::moveFocus(Math::eDirection dir) {
 
     static auto PNOFALLBACK = CConfigValue<Config::INTEGER>("general:no_focus_fallback");
     if (*PNOFALLBACK)
-        return {};
+        return actionError(std::format("Nothing to focus to in direction {}", Math::toString(dir)), eActionErrorLevel::INFO, eActionErrorCode::NOT_FOUND);
```
B2krobbery, line 376: This returns an INFO-level error when focus can't move.
The failing test expects this case to return not ok, so should this be ERROR instead of INFO?

B2krobbery (commented):
Looks good overall.
I noticed one test failure related to focus dispatch returning ok when the action can't be performed. Left a comment on that line for clarification.

src/config/shared/actions/ConfigActions.cpp
vaxerski: no, it should be warn

vaxerski: or we report info, both work

B2krobbery: got it thx 👍

src/config/shared/actions/ConfigActions.hpp
```diff
@@ -7,19 +7,29 @@
...
 
 namespace Config::Actions {
     struct SActionResult {
         bool passEvent = false;
     };
 
+    using eActionErrorLevel = Config::eConfigErrorLevel;
+    using eActionErrorCode  = Config::eConfigErrorCode;
+    using SActionError      = Config::SConfigError;
+    using Config::toString;
```
kolayne, line 20: It seems `toString` is never used

src/config/shared/actions/ConfigActions.cpp
```diff
@@ -289,7 +289,7 @@ ActionResult Actions::moveToWorkspace(PHLWORKSPACE ws, bool silent, std::optiona
         return {};
 
     if (!ws)
-        return {};
+        return std::unexpected("Invalid workspace");
```
kolayne, line 292: It's unclear why you left some cases the old way, returning a bare `std::unexpected`. If it's intentional, there should probably be a comment somewhere about it. E.g., right here or near the definitions of the error level and error code constants: to explain when each constant is used and when to use a plain `std::unexpected`.

vaxerski: genuine mistake on my end, oops.

vaxerski: https://github.com/hyprwm/Hyprland/pull/14221

PR #14196, opened 2026-04-28
Add pid on raw and cmd

ItsOhen:
Please mr vaxry sir! Let me have some pids in lua.

No

Yes

### DIFF
--- a/src/config/lua/bindings/LuaBindingsDispatchers.cpp
+++ b/src/config/lua/bindings/LuaBindingsDispatchers.cpp
@@ -1,6 +1,7 @@
 #include "LuaBindingsInternal.hpp"
 
 #include <hyprutils/string/String.hpp>
+#include <lua.h>
 
 #include "../../supplementary/executor/Executor.hpp"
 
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
I don't think hl.dispatch() forwards the return value, does it? What use does it have in that case?

ItsOhen:
Works fine for me.
I'm running the pids branch right now to get my overlay working.

```lua

local submap_overlay_pid = nil
hl.on("keybinds.submap", function(name)
	hl.timer(function()
		if submap_overlay_pid then
			hl.dsp.exec_cmd("kill -15 " .. submap_overlay_pid)()
			submap_overlay_pid = nil
		end
		if name == "" or name == "reset" or name == "default" then
			local settings = animations.registry["layersOut"]
			if settings then
				hl.animation(settings)
			end
			return
		end
		local s_dir = os.getenv("HOME") .. "/.config/hypr/scripts"
		local cmd = string.format(
			"nwg-wrapper -s %s/submap.sh -l 3 -a start -p right -mt 25 -mr 25 -c %s/submap.css",
			s_dir,
			s_dir
		)
		submap_overlay_pid = hl.dsp.exec_raw(cmd)()
	end, { timeout = 10, type = "oneshot" })
end)

```

But i'm sure there are plenty of places that where holding on to the pid can come in handy.

ItsOhen:
And i just checked, and no, you are correct, hl.dispatch don't forward.
So:
`hyprctl dispatch 'hl.dsp.exec_cmd("helloworld")'` give the usual: ok
But inside the config doing:
```lua
local pid = hl.dsp.exec_cmd("helloworld")()
print(pid)
```
gives the pid.

vaxerski:
`hl.dsp.exec_raw(cmd)()` this is NOT valid!!!!

The shape of the dispatcher table is not stable as the wiki says and it may change at any time!!!

ItsOhen:
I'm confused as to how it could change..

ThrowTop:
> `hl.dsp.exec_raw(cmd)()` this is NOT valid!!!!
> 
> The shape of the dispatcher table is not stable as the wiki says and it may change at any time!!!

maybe explictly make it not work then so people dont start using it before its too late.

ItsOhen:
> > `hl.dsp.exec_raw(cmd)()` this is NOT valid!!!!
> > The shape of the dispatcher table is not stable as the wiki says and it may change at any time!!!
> 
> maybe explictly make it not work then so people dont start using it before its too late.

But it can't change. On reload the entire state is rebuilt.
And there's no such thing as garbage collection that i can see.
So really no way for it to not be stable.

vaxerski:
what I mean is that any update could change the shape of what dsp return and make your () invalid

PR #14199, opened 2026-04-28
config/lua: fix unbind behavior

vaxerski:
Lua unbind took two args for no good reason. Even wiki was wrong. Now it's good.


kRHYME7:
: yeah it conflicts
 
While we're at it any reason we did not normalize `mod` `key` . I am basing on the wiki. Will it conflct with kb layouts / lang ? 

image

kRHYME7:
Might be out of scope but 

`hypctl binds -j`  does not provide correct dispather value

image

vaxerski:
not supposed to

PR #14207, opened 2026-04-29
render: keep blur work buffers in default image description with ICC

LwhJesse:
## Problem
On my laptop display, when ICC is enabled for the monitor,  kitty with opacity and blur becomes black. However when the ICC is turned off,  kitty will display normal blur and opacity as usual.

##  What I found
From my debugging, this seems related to the monitor ICC description being used for the intermediate work buffer. I have not fully traced the exact shader-side reason why the result becomes black. Blur/transparency should have run in Hyprland's default SDR/sRGB-like working space, instead of the ICC output description.

##  What changed
This patch keeps the work buffer/resources in the default image description when ICC is enabled.
The final output still uses the monitor ICC description, so this does not disable ICC.
I also added the missing ICC LUT uniforms for the blur finish shader path.
I cannot fully explain the pixel-level reason for the black result, but this change fixes the issue on my setup.

##  Testing
 - ICC enabled: transparent blurred kitty no longer becomes black.
 - ICC disabled: blur still works
 - Tested several normal apps: no new rendering issue observed.
 - Ran ctest: 200/203 passed.

The 3 failing ConfigLua tests also fail on the base commit and upstream/main, so they do not seem related to this patch.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -2551,7 +2551,7 @@ bool CMonitor::useFP16() {
 
 WP<CMonitorResources> CMonitor::resources() {
     const auto DRM_FORMAT = useFP16() ? DRM_FORMAT_ABGR16161616F : m_output->state->state().drmFormat;
-    const auto DESC       = useFP16() ? LINEAR_IMAGE_DESCRIPTION : m_imageDescription;
+    const auto DESC       = useFP16() ? LINEAR_IMAGE_DESCRIPTION : (m_imageDescription->value().icc.present ? getDefaultImageDescription() : m_imageDescription);
 
     if (!m_resources || m_resources->m_drmFormat != DRM_FORMAT || m_resources->m_size != m_pixelSize)
         m_resources = makeUnique<CMonitorResources>(m_self, DRM_FORMAT, m_pixelSize, DESC);
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1789,10 +1789,11 @@ SP<IFramebuffer> CHyprOpenGLImpl::blurFramebufferWithDamage(float a, CRegion* or
 
         currentTex->setTexParameter(GL_TEXTURE_MIN_FILTER, GL_LINEAR);
 
-        // From FB to sRGB
+        // From sRGB back to the work buffer/output description
         const bool skipCM = !m_cmSupported || !g_pHyprRenderer->workBufferImageDescription()->needsCM(getDefaultImageDescription());
+        const bool isICC  = g_pHyprRenderer->workBufferImageDescription()->value().icc.present;
         if (!skipCM) {
-            shader = useShader(getShaderVariant(SH_FRAG_BLURFINISH, SH_FEAT_CM));
+            shader = useShader(getShaderVariant(SH_FRAG_BLURFINISH, SH_FEAT_CM | (isICC ? SH_FEAT_ICC : 0)));
             passCMUniforms(shader, getDefaultImageDescription(), g_pHyprRenderer->workBufferImageDescription());
             shader->setUniformFloat(SHADER_SDR_SATURATION,
                                     m_renderData.pMonitor->m_sdrSaturation > 0 &&
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -3256,14 +3256,9 @@ void IHyprRenderer::renderSnapshot(WP<Desktop::View::CPopup> popup) {
 }
 
 NColorManagement::PImageDescription IHyprRenderer::workBufferImageDescription() {
-    // TODO
-    // const bool  IS_MONITOR_ICC  = m_renderData.pMonitor->m_imageDescription.valid() && m_renderData.pMonitor->m_imageDescription->value().icc.present;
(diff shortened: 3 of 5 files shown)


vaxerski:
check https://github.com/hyprwm/Hyprland/pull/14070

LwhJesse:
I tested #14070 on my setup.

Result:
- ICC enabled + transparent blurred kitty: still becomes black
- ICC disabled + transparent blurred kitty: works normally
- I also forced `render:use_fp16 = 0`, and the issue still happens with ICC enabled

So #14070 does not seem to fix this ICC + blur case on my setup.

LwhJesse:
Update: this was on a Hyprland build from #14070. After changing `render:use_fp16` at runtime with `hyprctl keyword`, the black kitty issue disappeared, so my previous test may have been affected by runtime render state / buffer reinitialization. I will retest #14070 from clean Hyprland sessions before making a final conclusion.

LwhJesse:
I tested #14070 as suggested, using clean Hyprland sessions.

Tested builds:

- PR base: `e6197623`
- #14070: `637c7a6`
- this PR: `53d6884`

Result summary:

On my setup, the PR base reproduces the black kitty issue in the default/auto path and in the forced FP16-off path.

#14070 only avoids the black kitty issue when FP16 is forced on. The default/auto path and the forced FP16-off path still reproduce the ICC + blur problem.

This PR fixes the issue on my setup in all three tested `render:use_fp16` modes.

Detailed results:

For PR base `e6197623`:

- default `render:use_fp16 = 2` / auto, ICC enabled: transparent blurred kitty becomes black
- forced `render:use_fp16 = 0`, ICC enabled: still becomes black
- forced `render:use_fp16 = 1`, ICC enabled: kitty blur/opacity works normally
- ICC disabled did not show the black kitty issue in the tested cases

For #14070 `637c7a6`:

- default `render:use_fp16 = 2` / auto, ICC enabled: transparent blurred kitty still becomes black
- forced `render:use_fp16 = 0`, ICC enabled: still becomes black
- forced `render:use_fp16 = 1`, ICC enabled: kitty blur/opacity works normally
- ICC disabled did not show the black kitty issue in the tested cases

For this PR branch `53d6884`:

- default `render:use_fp16 = 2` / auto, ICC enabled: works normally
- forced `render:use_fp16 = 0`, ICC enabled: works normally
- forced `render:use_fp16 = 1`, ICC enabled: works normally
- ICC disabled also did not show the black kitty issue in the tested cases

I also noticed that the blur/opacity appearance may look different when `render:use_fp16 = 1`, but I did not test that strictly, so I do not want to make a conclusion about that.

UjinT34:
If ICC is applied when something is rendered to the internal FB then it should be reverted before the blur. There is no code to do that atm. And I don't think there should be such thing. ICC should be applied only when rendering internal to output. But it won't be perfect if internal is uint8.
This PR probably can be merged as is if I don't come up with something better soon.  In any case ICC stuff should be moved out of internal rendering eventually.

LwhJesse:
Thanks, that makes sense.

That matches the intention of this PR: keep ICC out of the internal blur/workbuffer path, and only apply the monitor ICC description on the final internal -> output pass.

I agree that applying ICC internally and then trying to undo it before blur would be the wrong direction. I will leave this patch as-is for now unless there is a better approach from #14070 or the broader CM refactor.

UjinT34:
Can't repro black kitty bg with ICC and `use_fp16=0` on #14070
Note that fp16 blur has higher brightness overall for some reason. ICC may decrease brightness further depending on other settings and profile itself. To a point that low brightness can be confused with completely black bg. Please retest with higher opacity values and/or brighter bg with some large high-contrast chunks.
In any case the fix makes sense, no need to have ICC in a workbuffer. Shader changes shouldn't be triggered but might be kept just in case they'd be used with ICC workbuffer in the future.

vaxerski:
I'd prefer to merge this after 14070

LwhJesse:
That makes sense.

To clarify my original observation: on my setup it did not look like a low-brightness / visual-confusion case. The kitty text stayed visible, but the terminal behaved like an opaque black rectangle rather than a translucent blurred window, and the wallpaper/background pattern behind it was no longer visible.

Since #14070 is already touching this area and now includes the workbuffer/ICC direction, I think it makes sense to let #14070 land or stabilize first.

After that I can retest the original case with a cleaner setup:

- high-contrast static background
- higher kitty `background_opacity` values
- clean Hyprland sessions
- `render:use_fp16 = 0/1/2` set before startup
- ICC on/off comparison

If #14070 fixes the issue on my setup, I can close this PR or reduce it to any remaining missing part. If the issue still reproduces, I can provide cleaned-up repro results against the post-#14070 code.

LwhJesse:
Closing this PR as superseded by the upstream FP16/CM work in #14070 and the follow-up color-management fixes in #14361.

The original issue addressed here was ICC + transparent blurred kitty turning black, likely because the monitor ICC description was being used in the internal blur/workbuffer path. Current main now has upstream workbuffer image-description handling, including the `avoid icc workbuffer` change, and the original black blurred kitty case no longer reproduces on my setup.

If a related ICC/blur regression appears again, I’ll open a new minimal PR against the current post-#14070/#14361 code instead of continuing this older patch.

PR #14232, opened 2026-04-30
fix(config/Lua): check "workspace" instead of "id" in hl.workspace.move/rename

sh-gun:
According to the wiki, this should work:

`hl.dsp.workspace.move({ workspace = 1, monitor = "current" })`

image

_[source](https://wiki.hypr.land/Configuring/Basics/Dispatchers/#workspace-1)_

However, it doesn't, because "id" is expected in the Lua table instead of "workspace". "workspace" should be correct here because it's more descriptive and in line with all similar dispatchers. These were the only occurences of this mismatch I could find.

Very small and simple fix, so should be ready to merge.


vaxerski (approved):

sh-gun:
Synced to pick up the upstream gtests fix (c7b8fe13c10d6a4392bfda9bcfc041e0ea0b7c00)

PR #14250, opened 2026-05-02
render: scale background to monitor resolution

davc0n:
Currently the background is not scaled, e.g. 8k size image always uses ~128Mb of VRAM.
We could reduce the consumption for low-end devices and monitors with lower resolution:
- ~8Mb for 1080p
- ~16Mb for 2k
- ~32Mb for 4k

---

I didn't add a configuration option because there should be no trade-off here, couldn't find a case where it would be convenient to not scale, but eventually let me know.

EDIT:
>[Nix / test / build (pull_request)](https://github.com/hyprwm/Hyprland/actions/runs/25248703320/job/74037223737?pr=14250) Failing after 24m

Currently fails on main too, should not be related to this.


umbrageodotus:
gimme a config ver for the full ver :>

src/render/Renderer.cpp
```diff
@@ -1236,7 +1233,43 @@ SP<ITexture> IHyprRenderer::getBackground(PHLMONITOR pMonitor) {
         return nullptr;
 
     Log::logger->log(Log::DEBUG, "Creating a texture for BGTex");
-    SP<ITexture> backgroundTexture = createTexture(m_backgroundResource->m_asset.cairoSurface->cairo());
+
+    cairo_surface_t* cairoSurface  = m_backgroundResource->m_asset.cairoSurface->cairo();
+    cairo_surface_t* scaledSurface = nullptr;
```
vaxerski: use gpu rendering

src/render/Renderer.cpp
```diff
@@ -1236,7 +1233,43 @@ SP<ITexture> IHyprRenderer::getBackground(PHLMONITOR pMonitor) {
...
+
+    // If the background is larger than the monitor, scale it down
+    if (pMonitor->m_transformedSize.x > 0 && pMonitor->m_transformedSize.y > 0) {
```
vaxerski: ok and what if a monitor changes size?

src/render/Renderer.cpp
davc0n: are you sure that's really necessary?

I would avoid loading the full image in VRAM entirely.

src/render/Renderer.cpp
davc0n: what do you mean with changes size? a rotation?

I could add another check in `Monitor.cpp`:
```
if OLDTRANSFORMEDSIZE != m_transformedSize
    m_background.reset()
```

src/render/Renderer.cpp
umbrageodotus: do what vax says, better for perf

src/render/Renderer.cpp
vaxerski: a rotation, mode, resize, etc.

vaxerski (approved):

PR #14270, opened 2026-05-03
windows/focus: add fallbacks when focussing workspaces

fxzzi:
fixes https://github.com/hyprwm/Hyprland/discussions/12790
in some cases it was possible to get into a state where there was no last focused window on a monitor. this meant that `movetoworkspace` would be successful, but the "active window" is still on your other monitor.

This would lead to cases for example, where i'd movetoworkspace to my second monitor, close a window, and it would end up closing an unintended window on my main monitor.

It also helps because if you were in this state whilst using `confine_pointer` if you did `movetoworkspace` to your secondary monitor, your mouse would warp back to game.

it also helps when using initial workspace tracking, as leaving the workspace whilst an app was launching would also lead to the same outcome where there is technically no "last focused window" on said workspace / monitor 

maybe some of this isn't needed idk

looks like it

### DIFF
--- a/src/config/shared/actions/ConfigActions.cpp
+++ b/src/config/shared/actions/ConfigActions.cpp
@@ -94,7 +94,7 @@ static bool tryMoveFocusToMonitor(PHLMONITOR monitor) {
 
     const auto PNEWMAINWORKSPACE = monitor->m_activeWorkspace;
     const auto PNEWWORKSPACE     = monitor->m_activeSpecialWorkspace ? monitor->m_activeSpecialWorkspace : PNEWMAINWORKSPACE;
-    const auto PNEWWINDOW        = PNEWWORKSPACE->getLastFocusedWindow();
+    auto       PNEWWINDOW        = PNEWWORKSPACE->getFocusCandidate();
 
     if (PNEWWINDOW) {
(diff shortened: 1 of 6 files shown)


src/config/shared/actions/ConfigActions.cpp
```diff
@@ -932,10 +938,18 @@ ActionResult Actions::changeWorkspace(PHLWORKSPACE ws) {
...
         Vector2D middle = PMONITORWORKSPACEOWNER->middle();
-        if (const auto PLAST = ws->getLastFocusedWindow(); PLAST) {
-            Desktop::focusState()->fullWindowFocus(PLAST, Desktop::FOCUS_REASON_KEYBIND);
+        auto     pWindow = ws->getLastFocusedWindow();
+
+        if (!pWindow)
+            pWindow = ws->getTopLeftWindow();
+
+        if (!pWindow)
+            pWindow = ws->getFirstWindow();
```
vaxerski: these three sound like a perfect candidate for a function in CWorekspace

fxzzi: sure i did it and just called it getFocusCandidate

vaxerski (approved):

PR #14291, opened 2026-05-05
fix: update outdated links in example/hyprland.conf

abhegyn:
website was updated out dating the link present in hyprland.conf 
my pr updates the links to the latest and removes one link present above mainMod since that was removed from the website (or I couldn't find it)

yes it is, if possible add the mainMod link

### DIFF
--- a/example/hyprland.conf
+++ b/example/hyprland.conf
@@ -230,7 +230,6 @@ device {
 ### KEYBINDINGS ###
 ###################
 
-# See https://wiki.hypr.land/Configuring/Keywords/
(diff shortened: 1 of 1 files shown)


fxzzi:
We should probably point these to the 0.54.0 pinned version of the wiki? Since all the wiki pages are lua now

abhegyn:
> We should probably point these to the 0.54.0 pinned version of the wiki? Since all the wiki pages are lua now

Got it, thanks for the clarification. I will push the changes shortly.

vaxerski:
hell nah

I removed the hyprlang config

PR #14307, opened 2026-05-06
config/refresher: refresh watcher state properly

vaxerski:
ref #14301


cdump:
@vaxerski probably after this commit, the Lua config stopped reloading, even though `disable_auto_reload` is not changed in my configuration (so default 'true' should be used). I use `hl.on("config.reloaded"` + `hl.notification` to monitor that.

`hyprctl eval 'hl.config({ misc = { disable_autoreload = false } })'` fixes that

vaxerski:
yes I fixed it

PR #14313, opened 2026-05-06
config/lua: fix device bool int reads

imperishableSecret:

Fixes Lua config reads for pointer device options that are declared as boolean values but consumed through the integer device-config getter.

The Lua backend previously implemented `getDeviceInt()` by directly casting every stored value to `Config::INTEGER`. That is invalid for bool-backed options such as `left_handed` and `natural_scroll`, which are stored as `Config::BOOL`. When pointer config was reapplied during reload/refresher work, this could read a bool value as a nonzero integer and incorrectly enable libinput options, making mouse buttons appear swapped and scroll direction appear inverted.

This PR adds a virtual integer conversion hook for Lua config values:

- int values return their stored integer
- bool values return `0` or `1`
- float values return their stored float
- vec2 values return their stored vector
- string values return their stored string
- unsupported value/type combinations assert instead of being reinterpreted through an invalid pointer cast


vaxerski (approved):
tnx

PR #14341, opened 2026-05-07
Add dispatch option to `fullscreen()` and `fullscreen_state()` calls for using the 'Default' fullscreen behaviour instead of the layout-specific ones

erstarr:

Basically title. The new scroll fullscreen behaviour is unusable for me and I need the old FS behaviour back.

This is a toggle, so people can choose between: as/if new fullscreen behaviours are added to new/existing layouts, users can pick between "default" and layout-specific behaviours on a per-window basis.

---

Adds an attribute to all windows: "is the fullscreen handled by layout, or does it use default behaviour".

Adds an optional field to `.fullscreen({})` and `.fullscreen_state({})` dispatches, where the user can control which behaviour to use. Best of both worlds.

WIP

- [ ] add test
- [ ] add wiki PR
- [ ] Ensure no test regression (if exists)

-> Can add a window rule for a certain window to always use either behaviour 

WIP

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -2194,25 +2194,26 @@ bool CCompositor::workspaceIDOutOfBounds(const WORKSPACEID& id) {
 void CCompositor::changeWindowFullscreenModeClient(const PHLWINDOW PWINDOW, const eFullscreenMode MODE, const bool ON) {
     setWindowFullscreenClient(
         PWINDOW,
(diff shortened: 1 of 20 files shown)


erstarr:
- Gesture fullscreen always follows layout mode. I don't have a laptop that runs linux properly so I can't test or debug a change there.

- Windows started as FS via windowrule always follow layout FS behaviour

- legacy dispatcher automatically gets layout_aware = true (layout specific dispatch) behaviour, no possibility to change

For fullscreen events as a result of other dispatches/events, unfullscreen/fullscreen behaviour will follow if the window's fullscreen state was set following a layout specific fullscreen behaviour or not.

erstarr:
will be implemented in https://github.com/hyprwm/Hyprland/pull/14705. Will close after that is done

Issue #14340, opened 2026-05-07
monitor = NAME, disable silently no-ops in 0.54.3 (Asahi)


jwbron:
### Description

`hyprctl keyword monitor "<name>, disable"` returns `ok` but the
target monitor stays enabled. No log activity is produced by the
command. Same result when the directive is placed in `hyprland.conf`
and applied via `hyprctl reload`.

Other monitor keyword changes (e.g. tweaking scale) take effect, so
the keyword pipeline itself is alive; only the `disable` token
appears to be dropped.

### Reproduction

    $ hyprctl keyword monitor "eDP-1, disable"
    ok
    $ hyprctl monitors -j | jq '.[] | {name, disabled, dpmsStatus}'
    { "name": "eDP-1", "disabled": false, "dpmsStatus": true }
    { "name": "HDMI-A-1", "disabled": false, "dpmsStatus": true }

`hyprctl rollinglog` shows zero new lines from the command. The same
holds when the directive is added to `hyprland.conf` and Hyprland is
reloaded: only the generic reload acknowledgement appears, no
modeset / arrangeMonitors activity for eDP-1.

`hyprctl dispatch dpms off eDP-1` works (DPMS goes off, log shows
the expected aquamarine activity), so the wider output-control
surface is functional. It is specifically the `disable` directive
that no-ops.

### Expected

eDP-1 transitions to `disabled: true`; workspaces migrate to the
remaining monitor (HDMI-A-1).

### System

    Hyprland 0.54.3 (commit 521ece46, Mar 27 2026)
    aquamarine 0.11.0 (locally patched, see hyprwm/aquamarine#291;
      patch is scoped to resolveMatchingRenderNode, unrelated to
      modeset / monitor disable)
    Linux 6.19.13-400.asahi.fc43.aarch64+16k (Apple M-series, Asahi)
    Fedora Asahi Remix 43

### Caveats

kanshi is running (single-output `fedora` profile: `output eDP-1
enable scale 1.5`) and was not isolated for this report. The
zero-log signature on the disable command suggests Hyprland is
dropping the directive before any wlr-output-management broadcast
that kanshi could react to. Happy to retest with kanshi paused if
useful.

PR #14343, opened 2026-05-07
config/monitor: invalidate wlr-output-management cache on user rule writes

jwbron:
Addresses #14347 (the closed-issue equivalent was #14340).

## Background

`hyprctl keyword monitor "<name>, disable"` returns ok but the target
monitor stays enabled. Same result via `monitor=NAME, disable` in
`hyprland.conf` followed by `hyprctl reload`. Other monitor keyword
writes (scale, mode, position, transform, vrr) are similarly shadowed.

## Root cause

`CMonitorRuleManager::get` reads the user-declared rule, then
unconditionally overlays state from
`COutputManagementProtocol::getOutputStateFor` on top:

```cpp
rule.m_disabled = !CONFIG->enabled;
```

`COutputManagementProtocol::m_managers[*]->m_monitorStates` accumulates
state per wlr-output-management client and persists indefinitely. Once
kanshi (or any other wlr client) has prescribed state for an output,
subsequent `monitor=` directives never disable / re-mode / re-scale that
output: the user's write goes into `m_rules` correctly but is silently
shadowed on every read.

## Change

Add `forgetMonitor(name)` on `COutputManagementProtocol` that erases
entries across all manager caches for the given output. Call it from
`CMonitorRuleManager::add` whenever a rule lands. The next wlr-protocol
write repopulates the cache, so this only re-establishes precedence for
the user's most recent write; it doesn't break wlr-output-management's
ability to be the most recent writer.

## Scope

This is intentionally a tactical fix that restores the missing
precedence rule within the existing two-store design
(`m_monitorRules` for user directives, per-client `m_monitorStates`
for wlr-output-management state).

A more architectural alternative would collapse the two stores into
one, having `setDisableHead` / `applyTestConfiguration` flow through
`CMonitorRuleManager::add` with a merge step that preserves
wlr-output-management's partial-update semantics, plus add eviction
on wlr-client disconnect (the cache currently leaks indefinitely,
which is a separate latent bug). That's a 200-400 LOC refactor across
`OutputManagement.{cpp,hpp}` and `MonitorRuleManager.{cpp,hpp}` and
deserves its own design discussion.

Happy to take the refactor path if maintainers prefer that direction;
otherwise this PR is intentionally minimal and reversible, and the
unified-store work can land separately on top.

## Verification

Apple M1 Pro / Fedora Asahi 43 / locally patched aquamarine 0.11.0 (the
patch is unrelated to this code path; see hyprwm/aquamarine#291):

- Before: `hyprctl keyword monitor "eDP-1, disable"` returns ok,
  `disabled` stays `false`. Confirmed via runtime keyword and config +
  `hyprctl reload`.
- After: same command flips `disabled: true` and workspaces auto-migrate
  through the existing `onDisconnect` path.
- `hyprctl keyword monitor "eDP-1, preferred, auto, 1.5"` re-enables.

## Caveat / follow-up

On Asahi+aquamarine, the disable->enable round-trip via `hyprctl keyword`
exposes a separate downstream reliability issue: aquamarine's DRM atomic
commits intermittently fail with `ATOMIC_NONBLOCK PAGE_FLIP_EVENT busy`
during the modeset, occasionally leaving the panel stuck disabled or at
a wrong scale on re-enable. Tracked in #14348 (the closed-issue
equivalent was #14344); that's an aquamarine / atomic-commit issue
distinct from this precedence fix.

### DIFF
--- a/src/config/shared/monitor/MonitorRuleManager.cpp
+++ b/src/config/shared/monitor/MonitorRuleManager.cpp
@@ -32,7 +32,14 @@ void CMonitorRuleManager::clear() {
 
 void CMonitorRuleManager::add(CMonitorRule&& x) {
     std::erase_if(m_rules, [&x](const auto& e) { return e.m_name == x.m_name; });
+    auto name = x.m_name;
     m_rules.emplace_back(std::move(x));
+
+    // A user-issued `monitor=` rule should take precedence over any state a
+    // wlr-output-management client (e.g. kanshi) cached for the same output.
+    // The cache repopulates the next time a wlr client commits, so this only
+    // resets the layered override; it doesn't break wlr-output-management.
+    PROTO::outputManagement->forgetMonitor(name);
 
     scheduleReload();
 }
--- a/src/protocols/OutputManagement.cpp
+++ b/src/protocols/OutputManagement.cpp
@@ -654,6 +654,15 @@ SP<SWlrManagerSavedOutputState> COutputManagementProtocol::getOutputStateFor(PHL
     return nullptr;
 }
 
+void COutputManagementProtocol::forgetMonitor(const std::string& name) {
+    if (name.empty())
+        return;
+
+    for (auto const& m : m_managers) {
+        m->m_monitorStates.erase(name);
+    }
+}
+
 void COutputManagementProtocol::sendPendingSuccessEvents() {
     if (m_pendingConfigurationSuccessEvents.empty())
         return;
--- a/src/protocols/OutputManagement.hpp
+++ b/src/protocols/OutputManagement.hpp
@@ -158,7 +158,12 @@ class COutputManagementProtocol : public IWaylandProtocol {
     // doesn't have to return one
     SP<SWlrManagerSavedOutputState> getOutputStateFor(PHLMONITOR pMonitor);
 
-    void                            sendPendingSuccessEvents();
+    // erase any cached wlr-output-management state for this output across all
+    // managers. used so that a freshly-written `monitor=` rule isn't shadowed
+    // by stale state from a wlr-output-management client (e.g. kanshi).
+    void forgetMonitor(const std::string& name);
+
+    void sendPendingSuccessEvents();
 
   private:
     void destroyResource(COutputManager* resource);


vaxerski:
I don't think this is intended state though - the protocol should take precedence

jwbron:
The current implementation maintains m_monitorStates as a separate shadow store. sway/niri/labwc/wayfire all implement protocol-wins by having the protocol apply update the canonical config store, not by shadowing it. Would you accept a PR that aligns Hyprland with that pattern (unified store, last-writer-wins, kanshi naturally dominates because it's the most frequent writer)? That preserves your 'protocol takes precedence' principle in practice while fixing the silent-shadow and disconnect-leak bugs.

vaxerski:
sounds acceptable, but what about config reloads?

jwbron:
Three-method split on `CMonitorRuleManager`:

- `replaceFromConfig(std::vector<CMonitorRule>)` for the config-reload path. Maintains a private `m_lastParsedRules` cache (map by name). On call: rules whose parsed content matches the cached prior parse are left alone, preserving any wlr/hyprctl overlays; rules whose content changed are replaced; rules whose names are absent from the new parse are dropped. Update the cache to the new parse at the end.
- `add()` unchanged for the `hyprctl keyword` path.
- `mergeFromWlrManagement(const std::string& name, const SWlrManagerSavedOutputState& partial)` for wlr-OM writes. Reads the existing `CMonitorRule`, copies fields where `committedProperties` is set, writes back. Does not touch `m_lastParsedRules`.

Removes `m_monitorStates`, `getOutputStateFor`, and the `applyWlrOutputConfig` overlay in `get()`; `get()` becomes a plain `m_rules` lookup with the existing static-selector fallback, since the rule itself now reflects all merged state.

Slight refinement of the last-writer-wins framing from my prior comment: reload acts as a Config-origin write per output only if that output's parsed content actually changed. So kanshi state survives unrelated reloads (keybind edits, status-bar config, etc.) cleanly instead of recovering from a flash via the post-reload broadcast.

Reload behavior per output:

- Keybind-only reload, no `monitor=` lines changed: every comparison in `replaceFromConfig` hits, no rule replaced, no broadcast.
- User edits `monitor=eDP-1` only: that rule is replaced, eDP-1's wlr overlays drop, HDMI-1's state survives untouched. Finer than niri, which drops all transient state on any change in the outputs section.
- After a content-changed reload, `replaceFromConfig` emits a fresh `wlr_output_manager_v1_set_configuration` snapshot so wlr-OM clients can re-apply. Skipped when nothing changed.

One wrinkle worth flagging: [`handleMonitor`](https://github.com/hyprwm/Hyprland/blob/main/src/config/legacy/ConfigManager.cpp#L1283) and [`handleMonitorv2`](https://github.com/hyprwm/Hyprland/blob/main/src/config/legacy/ConfigManager.cpp#L783) are shared between config-parse and `hyprctl keyword`, and both end at `monitorRuleMgr()->add()` today. Cleanest fix is to have them accumulate into a `CConfigManager`-owned vector during reload (gated by a flag set around the hyprlang parse call) and submit via `replaceFromConfig` at the end; the hyprctl path keeps the immediate `add()`. Same restructure in the lua `ConfigManager`.

Two design choices worth your call:

1. Deletion: removing a `monitor=NAME` line from config should drop NAME's unified-store entry on reload, or preserve it? Recommend drop (matches "delete the line to revert to defaults"); niri preserves.
2. Hyprctl durability: the model above preserves hyprctl writes across unrelated reloads (they don't touch the parsed-content cache). If you'd prefer strict ephemeral (sway behavior: every reload nukes hyprctl writes too), route hyprctl through the parsed cache. Recommend durable, for parity with how wlr writes are handled.

vaxerski:
honestly tracking changed rules to me just seems unintuitive to the user. We're doing hoops and loops here.

I fail to see the point of kanshi altogether if you're gonna use lua to control your monitors.

jwbron:
The pattern that motivates the per-rule comparison is mixed config-plus-kanshi: baseline `monitor=,preferred,auto,1` (or a primary-display rule) in `hyprland.conf`, with kanshi handling dock/external hotplug. Not "kanshi instead of config," more "config provides defaults, kanshi handles known multi-monitor profiles." Common enough that `hyprctl reload` clobbering kanshi state on every keybind edit shows up in user reports.

That said, your point on bookkeeping cost is fair. Simpler shape if you prefer it: `replaceFromConfig` becomes "clear `m_rules`, rebuild from parse, broadcast." No `m_lastParsedRules`, no per-rule diff. Wlr-OM clients re-apply on the broadcast's `done` event, typically within a frame. Same model sway ships and a much smaller diff. Trade is a brief flash on unrelated reloads; if that's an acceptable cost, the rest of the architecture (unified store, `mergeFromWlrManagement`, deletion of `m_monitorStates` / `getOutputStateFor` / `applyWlrOutputConfig`) is unchanged.

Your call which trade is worth it. Happy with either.

vaxerski:
that sounds fine

PR #14361, opened 2026-05-08
Fix decoration colors with linear FP16

UjinT34:
Fixes some missing CM and incorrect values.

Makes most math to use linear. Not entirely correct and produces slight brightness differences but those are much smaller than before.
Might make sense to do the opposite and make some of linear math use non-linear space when the same happens with non-linear TFs. That will keep existing setups the same way it's on main but will require extra math.
The solution is a bit hacky and there might be a better way to apply additional alpha with premultiplied colors that will look the same with linear and non-linear TFs.

Ready

- [x] clear color
- [x] shadow
- [x] opacity

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -3402,11 +3402,11 @@ using ColorConversionKey = std::tuple<float, float, float, float, uint64_t>;
 static std::map<ColorConversionKey, CHyprColor> colorConversionCache;
 
 CHyprColor                                      IHyprRenderer::getConvertedColor(const CHyprColor& color) {
-    const auto               targetId = m_renderData.currentFB->imageDescription();
-    const ColorConversionKey key      = {color.r, color.g, color.b, color.a, targetId};
+    const auto               DESCR = m_renderData.currentFB->imageDescription();
+    const ColorConversionKey key   = {color.r, color.g, color.b, color.a, DESCR->id()};
     if (colorConversionCache.contains(key))
         return colorConversionCache[key];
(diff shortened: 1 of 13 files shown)


vaxerski:
https://github.com/hyprwm/Hyprland/pull/14364

shadow rect and glow should be fixed there

src/render/Renderer.cpp
```diff
@@ -3397,3 +3397,16 @@ SP<ITexture> IHyprRenderer::renderSplash(const std::function<SP<ITexture>(const
...
+
+using ColorConversionKey = std::tuple<float, float, float, float, uint64_t>;
+static std::map<ColorConversionKey, CHyprColor> colorConversionCache;
+
+CHyprColor                                      IHyprRenderer::getConvertedColor(const CHyprColor& color) {
+    const auto               targetId = m_renderData.currentFB->imageDescription();
+    const ColorConversionKey key      = {color.r, color.g, color.b, color.a, targetId};
+    if (colorConversionCache.contains(key))
+        return colorConversionCache[key];
+    const auto converted      = convertColor(color, DEFAULT_SRGB_IMAGE_DESCRIPTION, workBufferImageDescription());
```
vaxerski: wth? use the current fb's description here this is breaking it again

umbrageodotus:
doesn't boot 😭

umbrageodotus:
I can't get logs rn so if someone can TYSM but I can get them in some time

vaxerski (approved):

PR #14402, opened 2026-05-10
Fix/stale keyboard enter

I-who-ant:
#### -bug
After  keyboard destroyed   seat 's pressedkey cache isn't rebuilt.

#### -The-fix
Rebuild the cache in the function : src/managers/input/InputManager.cpp > void CInputManager::destroyKeyboard

#### -what'files I added
1.The Helper: PressKeyCache and its test (accord with the "## Submitting new tests : must be gtest")

<details>
<summary>More notes</summary>

  (   For the helper tests we wrote, I can only find in "tests/config/lua/LuaConfigUtils.cpp" a case where file-inline fake objects are used to test methods.)
     While writing, there are things I need to pay attention to: for example, in tests, the "<gtest/gtest.h>" TEST method, I have these doubts:

   1. What does each parameter mean? Why does the first parameter not follow other tests and be written as helpers,
      but after compilation the tests can still pass? And it should be noted that the name does not need to add the corresponding test file name, just write its purpose.
      (sorry to say one more word here  that this is my first time to use gtest that I write down some of what I learn from this pr)
      
   2. And, this is my first time designing a function test: I need to think according to each function's features about what TESTs can be designed to verify it.

</details>

2.The hyprtester  :  hyprtester client + plugin + focus flow

<details>
<summary>More notes : (sorry for I write to many)</summary>

   About hyprtester/, I had never contacted this kind of technology before that can insert tests inside the process.

   So, when I added "rebuildSeatPressedCacheFromKeyboards" in "InputManager.cpp" for fixing the bug,
   I did not think of how to verify whether the seat level pressed-key cache is correctly rebuilt after the lifecycle change of "keyboard-like devices".

   After deeply analyzing the architecture and understanding how Hyprland "produces test scenarios in-process",
   I understood that this special hyprtest directory is used to test injecting into the real process.

   After analysis, I learned that "hyprtester/src/tests/clients/" is used to design for the client under "hyprtester/clients":
   
   1. how to start it
   2. how to communicate with it
   3. how to read its output
   4. how to turn output into assertions

   So, I was able to design hyprtester/clients/keyboard-enter.cpp:

   1 . when to start the client
   2 . when to call the plugin injection entry
   3 . when to switch focus
   4 . when to send key
   5 . when to destroy keyboard
   6 . when to read result
   7 . how to assert at the end
   
   Later I rethought that this is a bug triggered by me in emacs,and after multiple window focus it would all have such a problem :
   for example, my CX31988+97220 suddenly disconnected, then came back, after Hyprland received this matter of "m_events.destroy"
   the internal state was not cleaned up : like "XF86AudioPlay".
   (While in emacs ,it immediately reflect when I insert "Alt,ctrl,super.." :  s-<XF86AudioPlay> is undefine)
   
   So here it is necessary to design two surfaces / clients to switch back and forth, so as to stably trigger enter and observe the difference
   (for example, whether stale key exists / is cleared)
   So, when designing this "CKeyboardEnterClient" class, it is different from the output protocol designed by other clients in the same directory

   Then, first, I need to think about how to input the key, and can know that it is necessary to use plgin to simulate a tool for testing scenarios inside the compositor process

So need to design  
1. stale_enter_setup
    - create two test keyboards
2. stale_enter_key
    - send key to the specified keyboard
3. stale_enter_destroy
    - destroy the specified keyboard
4. stale_enter_cleanup
    - clean up residual state

( and then the bug can be tested from whether after destroy keyboard the seat pressed-key cache should be rebuilt )

 So, such a situation can be designed in the test  

   1. first register two clients as A and B respectively, and call plugin to create two keyboards
   2. then, by focusing to A (through the designed Client code, it can be ensured that it will output enter 0, ensuring that there are no residual keys in the initial state)
   3. press a key by using the 0th keyboard to set a seat shared-state key, then focus to B, ensuring that B can get this shared key
   4. when focusing B, destroy the 0th keyboard, then focus to A again  
      (it can be known that if the seat cache is not rebuilt, then the system will still incorrectly think that: key 1 is still in the pressed set, so when A receives enter again, it may still be: enter 1 1, and after verification, the code before the fix indeed has this situation)
      (so, after fixing it, when focusing to A again, it will trigger wl_keyboard.enter in the client,
       that is  the clientlog  when  there is no key)  
      (some design considerations on the client design side: because when a client is created, from example  by observing "/src/desktop/view/Window.cpp:2175", after a new window is mapped Hyprland will automatically give focus to it, causing enter to be triggered)
      (so it is necessary to design drain() and focus() well in "hyprtester/src/tests/clients/keyboard-enter.cpp")

</details>

### Is there anything you want to mention? 

emmm, I'm not sure it's fine design

1.The function "hyprtester/plugin/src/main.cpp > staleEnterSetup" :
     -I created using "g_staleEnterKeyboards" to let it use 2 keyboard to fit this hyprtest .
2. about src/devices/IKeyboard.hpp:
```
+    const std::vector<uint32_t>&      getPressedKeys() const;
+    bool                              shareStates() const;
```
  the two functions  "const"  for noting need to change the state of keyboard

3. I try my best to follow the code standards like the Same-level directory file, if there are some code break, I'll be sorry.

<details>
<summary>maybe ？</summary>

I don't know , because in my laptop,  it always has :
  - processSpawning
  - gestures
  - groups
  - keybinds
  - initialFloatSize
  - windows
  - childWindow
  - shortcutInhibitor
the eight problems (from ../build/hyprtester/hyprtester --plugin ./plugin/hyprtestplugin.so)
and with the message from start"unable to lock lockfile /run/user/1000/wayland-1.lock, maybe another compositor is running"
I'm confused, cause I'm using it, it generate the ".lock" : How can I avoid it ? 
The I reboot and use tty to advoid this, still had the eights, I don't know why. maybe It's my own problems...

(honestly, after confused with the eight, I even doubt that it is only my case..I hope it's not.)
</details>

<details>
<summary>myDevice</summary>

- `Distro`: Arch Linux
- `Kernel`: Linux 7.0.5-zen1-1-zen
- `Desktop`: hyprland
- `Session`: wayland
- `XDG_RUNTIME_DIR`: /run/user/1000
- `WAYLAND_DISPLAY`: wayland-1
- `DISPLAY`: :1
- `Wayland`: 1.25.0
- `Wayland protocols`: 1.48
- `Aquamarine`: 0.11.0-2
- `Hyprutils`: 0.13.1-1
- `Hyprland`: 0.54.3-4
- `xdg-desktop-portal-hyprland`: 1.3.12-2
- `OS`: Linux seeback 7.0.5-zen1-1-zen
- `CPU`: 12th Gen Intel(R) Core(TM) i5-12450H

</details>

needs work ，cause  local  hyprtester  failed.

[hyprtester1.log](https://github.com/user-attachments/files/27572153/hyprtester1.log)

### DIFF
--- a/src/helpers/PressedKeyCache.cpp
+++ b/src/helpers/PressedKeyCache.cpp
@@ -1,0 +1,40 @@
+#include "PressedKeyCache.hpp"
+
+#include <algorithm>
+#include <ranges>
+
+namespace NInputUtils {
+    bool isKeyboardEligibleForPressedKeyCache(const IKeyboard& keyboard) {
+        return keyboard.m_enabled && keyboard.m_allowed && keyboard.shareStates();
+    }
+
+    std::vector<uint32_t> collectUniquePressedKeys(const std::vector<std::vector<uint32_t>>& pressedKeySets) {
+        std::vector<uint32_t> rebuiltPressedKeys;
+
+        for (const auto& pressedKeys : pressedKeySets) {
+            for (const auto& key : pressedKeys) {
+                if (!std::ranges::contains(rebuiltPressedKeys, key))
+                    rebuiltPressedKeys.emplace_back(key);
+            }
+        }
+
+        return rebuiltPressedKeys;
+    }
+
+    std::vector<uint32_t> collectPressedKeysFromKeyboards(const std::vector<SP<IKeyboard>>& keyboards) {
+        std::vector<std::vector<uint32_t>> pressedKeySets;
+        pressedKeySets.reserve(keyboards.size());
+
+        for (const auto& kb : keyboards) {
+            if (!kb)
+                continue;
+
+            if (!isKeyboardEligibleForPressedKeyCache(*kb))
+                continue;
+
+            pressedKeySets.emplace_back(kb->getPressedKeys());
+        }
+
+        return collectUniquePressedKeys(pressedKeySets);
+    }
+}
--- a/hyprtester/CMakeLists.txt
+++ b/hyprtester/CMakeLists.txt
@@ -101,4 +101,5 @@ protocolnew("unstable/keyboard-shortcuts-inhibit" "keyboard-shortcuts-inhibit-un
 clientNew("pointer-warp" PROTOS "pointer-warp-v1" "xdg-shell")
 clientNew("pointer-scroll" PROTOS "xdg-shell")
 clientNew("child-window" PROTOS "xdg-shell")
+clientNew("keyboard-enter" PROTOS "xdg-shell")
 clientNew("shortcut-inhibitor" PROTOS "xdg-shell" "keyboard-shortcuts-inhibit-unstable-v1")
--- a/hyprtester/clients/keyboard-enter.cpp
+++ b/hyprtester/clients/keyboard-enter.cpp
@@ -1,0 +1,287 @@
+#include <cstring>
+#include <sys/poll.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <print>
+#include <format>
+#include <string>
+#include <vector>
+
+#include <wayland-client.h>
+#include <wayland.hpp>
+#include <xdg-shell.hpp>
+
+#include <hyprutils/memory/SharedPtr.hpp>
+#include <hyprutils/math/Vector2D.hpp>
+
+using Hyprutils::Math::Vector2D;
+using namespace Hyprutils::Memory;
+
+struct SWlState {
+    wl_display*                    display;
+    CSharedPointer<CCWlRegistry>   registry;
+
+    CSharedPointer<CCWlCompositor> wlCompositor;
+    CSharedPointer<CCWlSeat>       wlSeat;
+    CSharedPointer<CCWlShm>        wlShm;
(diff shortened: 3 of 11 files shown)


src/helpers/PressedKeyCache.cpp
```diff
@@ -0,0 +1,40 @@
...
+        std::vector<uint32_t> rebuiltPressedKeys;
+
+        for (const auto& pressedKeys : pressedKeySets) {
+            for (const auto& key : pressedKeys) {
+                if (!std::ranges::contains(rebuiltPressedKeys, key))
+                    rebuiltPressedKeys.emplace_back(key);
+            }
+        }
+
+        return rebuiltPressedKeys;
```
vaxerski, line 21: isn't there a ranges::unique or something?

I-who-ant: yea, not considered too much cause there are little input it.. I'll write  more C++ code these days for the graduate test.
Thank you for reminding me using a better stl algorithm. 🥲

I-who-ant: set is better?

vaxerski: sounds adequate if you don't need ordering yes

PR #14421, opened 2026-05-11
hyprtester: report failures at the end

vaxerski:
Prints failures

### DIFF
--- a/hyprtester/src/main.cpp
+++ b/hyprtester/src/main.cpp
@@ -34,8 +34,10 @@
 #include <chrono>
 #include <thread>
 #include <print>
+#include <string>
 #include <string_view>
 #include <span>
+#include <vector>
 
 #include "Log.hpp"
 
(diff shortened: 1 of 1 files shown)


hyprtester/src/main.cpp
```diff
@@ -169,17 +171,18 @@ static bool preTestCleanup() {
     return !failed;
 }
 
-static void runTests(std::map<const char*, CTestCase&>& testCases) {
+static void runTests(std::map<const char*, CTestCase&>& testCases, std::string_view suiteName, std::vector<std::string>& failedTestNames) {
```
kolayne, line 174: Oh no, out parameters...

I think it's an anti-pattern in C++, isn't it? Let's use return values instead. (I'll include this in a PR I'm about to submit)

vaxerski: it is but it's a tester oh well

PR #14425, opened 2026-05-11
monitor: fix top-layer bar visibility on workspace change with scrolling-layout fullscreen

zerbiniandrea:
When a workspace uses the scrolling layout and one of its columns is fullscreen, switching away from that workspace and then back to it leaves shell bars (waybar, wayle, etc.) rendered on top of the fullscreen window. 

Scrolling between columns inside the workspace works fine insteaad, the bar fades out / in correctly instead.

This PR factors the check into a small `workspaceHasFullscreen` helper and uses it at all five call sites. After this patch, the bar is correctly hidden, tested on my machine with the patch applied.

Should definitely not have any side effect, the change is quite small

Ready for merging, already ran formatting as described in the guidelines

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -61,6 +61,19 @@ using namespace NColorManagement;
 using namespace Render::GL;
 using namespace Monitor;
 
+// True if the workspace has compositor-level OR layout-managed (e.g. scrolling) fullscreen.
+// Mirrors the check in CMonitor::inFullscreenMode so layer-fade decisions on workspace
+// change stay consistent with rendering, focus, and DS eligibility.
+namespace {
+    static bool workspaceHasFullscreen(const PHLWORKSPACE& ws) {
+        if (!ws)
+            return false;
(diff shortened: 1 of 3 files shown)


src/helpers/Monitor.cpp
```diff
@@ -61,6 +61,19 @@ using namespace NColorManagement;
 using namespace Render::GL;
 using namespace Monitor;
 
+// True if the workspace has compositor-level OR layout-managed (e.g. scrolling) fullscreen.
+// Mirrors the check in CMonitor::inFullscreenMode so layer-fade decisions on workspace
+// change stay consistent with rendering, focus, and DS eligibility.
+namespace {
+    static bool workspaceHasFullscreen(const PHLWORKSPACE& ws) {
```
vaxerski: there is a helper in CWorkspace IIRC, but if there isn't, that is where this should go

zerbiniandrea: Just found the helper function for `hasFullscreenWindow`, moved the helper there and reused that function

vaxerski (approved):

PR #14447, opened 2026-05-12
config/lua: allow re-enabling monitors

bizmythy:

This PR configures the Lua configuration setting `hl.monitor.disabled = false` to enable the monitor.

This allows Lua scripts to enable monitors that were previously disabled. Without this change, the only way I could find to re-enable a disabled monitor was to reload the entire configuration.

This is not how the original `hyprlang` configuration worked, so I totally understand if this is not a change that we want to make. However, I think the Lua configuration needs to provide some way to allow a user script to re-enable disabled monitors. Otherwise, I'm not sure how one would write a keybind that toggled a monitor on/off for example.

Ready for merging, very small change.

### DIFF
--- a/src/config/lua/bindings/LuaBindingsConfigRules.cpp
+++ b/src/config/lua/bindings/LuaBindingsConfigRules.cpp
@@ -93,8 +93,7 @@ namespace {
          }},
         {"disabled", []() -> ILuaConfigValue* { return new CLuaConfigBool(false); },
          [](ILuaConfigValue* v, CMonitorRuleParser& p) {
-             if (*sc<const Config::BOOL*>(v->data()))
-                 p.setDisabled();
+             p.rule().m_disabled = *sc<const Config::BOOL*>(v->data());
              return true;
          }},
         {"transform", []() -> ILuaConfigValue* { return new CLuaConfigInt(0, 0, 7); },


vaxerski (approved):
thanks!

alba4k:
Doesn't seem to work, though?

Furthermore, I don't think reloading your config should always re-enable every single connected monitor (?)

drew-council:
> Doesn't seem to work, though?

It does seem to be working for my configuration. I'm curious what isn't working for you.

I got the AI to write a small test script to show the state switching in a basic Lua config: https://gist.github.com/bizmythy/94afea4178eaf2f181d738349ea240fc
Note that the wayland/debugging live session doesn't properly visually show the monitors being disabled, but the log output inspecting `hyprctl` shows they are disabled/enabled. On real monitors in my config, the actual monitor fully disconnects/reconnects properly.

> Furthermore, I don't think reloading your config should always re-enable every single connected monitor (?)

This is not what this does, to my knowledge. Reloading the config resets all configured monitors to their configured state. If a monitor specifies `disabled = true`, it will be disabled (was the case before). Now, if a monitor specifies `disabled = false`, it will be _enabled_.
This is important for things like toggling monitor states with a Lua function (run from a keybind or other event).
Before this change, once the config sets `disabled = true` for a monitor, there is no way to re-enable it without a full configuration reload.

alba4k:
I tested it again. While this works when just trying to re-enabled with hyprctl, things like:
```lua

    if layout == nil then
        hl.exec_cmd("hyprctl reload")
    end
end)
```

fail to re-enable my monitor without the `hyprctl reload`, same with
```lua
hl.bind("switch:off:Lid Switch", function()
        if layout == nil then -- not a known layout where eDP-1 should stay off
            hl.exec_cmd("hyprctl reload")
        end
end, {locked = true})
```

> This is not what this does, to my knowledge
I know it's not related to this mr, was just checking. What I'm saying is that currently, even if I specify `disabled = true`, all monitors get reset to `disabled = false` whenever the config is reloaded.

vaxerski:
@alba4k 

```lua
local enabledMon = true

hl.bind("ALT + O", function()
    if enabledMon == true then
        hl.monitor({ output = "WAYLAND-2", disabled = true })
        enabledMon = false
    else
        hl.monitor({ output = "WAYLAND-2", mode = "preferred", position = "auto", scale = 1, disabled = false})
        enabledMon = true
    end
end)
```

works on my end. I can't repro any problems :(

alba4k:
You're right, that works. The lid switch bind works too, actually. 

However, they both stop working if there is no active monitor, for example in this scenario:
- eDP-1 disabled, DP-5 and DP-7 enabled
- disconnect the dock with DP-5 and DP-7
- `hl.on("monitor.removed", ...)` doesn't turn eDP-1 back on. That only happens when adding `hl.exec_cmd("hyprctl reload")` to the lambda

I assumed the bind was also borked because I was trying to use it as a fallback (to turn back eDP-1 after `hl.on(...)` failed) and it was also failing.

Adding a reload in either of the two works

vaxerski:
thanks now I can repro. I'll get it fixd

PR #14478, opened 2026-05-12
groups: sync window monitor/workspace when added to a group

andresilva:
When a window was moved from a group on one monitor into a group on another monitor, the target group's active tab (i.e. the window that was just moved into it) became invisible. Tab-switching restored visibility, confirming the window was a member of the target group but rendered on the wrong monitor. `moveWindowIntoGroupHelper` pre-set `pWindow->m_monitor`/`m_workspace` to the destination, but `CGroup::remove` immediately undid that via `assignToSpace(oldGroup.space)`, and `CGroup::add` then only set a `setSpaceGhost`, it never re-synced the incoming window.

This PR updates `CGroup::add` so that when an incoming window's `m_monitor` doesn't match the group's space's workspace monitor, it updates it. This enforces the invariant "a window in a group lives on the group's monitor" at the `CGroup` layer.

### DIFF
--- a/src/config/shared/actions/ConfigActions.cpp
+++ b/src/config/shared/actions/ConfigActions.cpp
@@ -1259,11 +1259,6 @@ static void moveWindowIntoGroupHelper(PHLWINDOW pWindow, PHLWINDOW pWindowInDire
 
     updateRelativeCursorCoords();
 
-    if (pWindow->m_monitor != pWindowInDirection->m_monitor) {
-        pWindow->moveToWorkspace(pWindowInDirection->m_workspace);
-        pWindow->m_monitor = pWindowInDirection->m_monitor;
-    }
-
     if (pWindow->m_group)
         pWindow->m_group->remove(pWindow);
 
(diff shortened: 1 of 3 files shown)


src/config/shared/actions/ConfigActions.cpp
```diff
@@ -1259,11 +1259,6 @@ static void moveWindowIntoGroupHelper(PHLWINDOW pWindow, PHLWINDOW pWindowInDire
 
     updateRelativeCursorCoords();
 
-    if (pWindow->m_monitor != pWindowInDirection->m_monitor) {
-        pWindow->moveToWorkspace(pWindowInDirection->m_workspace);
-        pWindow->m_monitor = pWindowInDirection->m_monitor;
-    }
-
```
andresilva, line 1266: This is no longer needed and keeping this would cause unnecessary `movewindow` events.

andresilva:
Test failure seems unrelated, `followMouseShrink` is also failing on main.

vaxerski (approved):

PR #14526, opened 2026-05-13
lua: restrict package module loading to lua-only modules

x1h0:
Restricts lua module loading to lua-only modules.

nope

ready

### DIFF
--- a/src/config/lua/ConfigManager.cpp
+++ b/src/config/lua/ConfigManager.cpp
@@ -297,6 +297,21 @@ void CConfigManager::reinitLuaState() {
     lua_getglobal(m_lua, "package");
     lua_pushstring(m_lua, luaPath.c_str());
     lua_setfield(m_lua, -2, "path");
+
+    lua_pushstring(m_lua, "");
+    lua_setfield(m_lua, -2, "cpath");
+    lua_pushnil(m_lua);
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks!

darkwater:
Why?

ItsOhen:
I just don't understand why..
And don't give me that "BRUH SECURITY!" crap.

alerque:
This already broke my newly created Lua based config in the last patch release and for what? No reason given, no reason asked for in review, just merged? Hard to say with nobody saying anything at all, but this smells like an LLM generated "security" suggestion than nobody bothered to even review. There are lots of legitimate uses for loading C modules into Lua: `lpeg`, `luafilesystem`, `luasocket`, and many others rely on C modules to power their functions that might have perfectly valid use cases for some people. And it works, so so what are you even trying to fix with this?

Dregu:
Yeah fuck this.

CelDaemon:
No reason was given for this change, and no notes about breaking compatibility were given. There is no security benefit either, as configs can already execute any system commands. This is just plain silly, and I really think this should be reverted. Please consider #14533 .

SergeyKasmy:
Why was this made? At least some explanation would be nice because as of now this doesn't seem useful whatsoever and has already broken by setup :(

PR #14529, opened 2026-05-13
config: schedule monitor reloads in event loop

aruhier:

Context: #14230.

Fixes the situation when kanshi cannot re-enable a disabled monitor when applying a profile on disconnect.

The monitor config reloads are not applied when disconnecting a screen, in the context of docking a laptop and using wlr-output-management to enable/disable the internal screen when docked. Schedule the reload in the event loop forces the config reload to happen.

I don't think that using the event loop is the best place to schedule this event, but I haven't found a better solution.

Ready to merge (tested and confirmed by another user https://github.com/hyprwm/Hyprland/discussions/14230#discussioncomment-16887688).

### DIFF
--- a/src/config/shared/monitor/MonitorRuleManager.cpp
+++ b/src/config/shared/monitor/MonitorRuleManager.cpp
@@ -115,6 +115,13 @@ void CMonitorRuleManager::scheduleReload() {
         return;
 
     m_reloadScheduled = true;
+
+    g_pEventLoopManager->doLater([this] {
+        if (m_reloadScheduled) {
+            performMonitorReload();
+            m_reloadScheduled = false;
+        }
+    });
 }
 
 void CMonitorRuleManager::performMonitorReload() {


alba4k:
does this adress the situation where after a monitor is disabled with `hl.monitor({output = "XYZ", disabled = true})` it cannot be enabled via `hl.monitor({output = "XYZ", mode = "preferred", position = "auto"})`? even not considering kanshi

Furthermore, does it adress the fact that currently every connected monitor is re-enabled on every manual or automatic config reload?

edit: it doesn't seem either, although it sounds like it should fix at least the first one?

aruhier:
> does this adress the situation where after a monitor is disabled with hl.monitor({output = "XYZ", disabled = true}) it cannot be enabled via hl.monitor({output = "XYZ", mode = "preferred", position = "auto"})?

I just tested to confirm (but with hyprlang), and yes it works. Without this patch it wasn't working, I tried to workaround it this way by manually calling hyprctl in kanshi but the monitor stayed off.

> Furthermore, does it adress the fact that currently every connected monitor is re-enabled on every manual or automatic config reload?

No, it still continues to happen, at least on hyprlang. `hyprctl reload config-only` still wakes up monitors that have been disabled. This however doesn't happen with kanshi running, as it keeps my profile.

alba4k:
> I just tested to confirm (but with hyprlang), and yes it works

Huh, strange, I was still unable to turn on my laptop even after switching to your fork. Could you try to see what happens when using a lua config? Shikane also fails to turn the screen back on

> No, it still continues to happen, at least on hyprlang.

Yeah, that seems to be the same with lua. My screen also stays off when using shikane. The thing is, I noticed all this while trying to replace shikane with a lua script in my config

alba4k:
Sort of unrelated, but do you also experience https://github.com/hyprwm/Hyprland/discussions/12325 when no display is enabled?

aruhier:
> Sort of unrelated, but do you also experience #12325 when no display is enabled?

Yes, switching to another tty doesn't re-enable the monitor. But that is to be expected, no? I think it was always the case.

aruhier:
@alba4k for lua, are you sure you have b76a9e07229c85298d44423b77d8561c8786f32d?

alba4k:
I just built from your fork, and yeah, it has it. `hyprctl version` reports as follows:

```
Hyprland 0.55.0 built from branch pr-fix-kanshi-disabled-monitor at commit ff7143b5223872b9571641119c0bda190921b621 clean (Schedule monitor reloads in event loop).
Date: Wed May 13 23:27:18 2026
Tag: , commits: 7318

Libraries:
Aquamarine: built against 0.11.0, system has 0.11.0

Version ABI string: ff7143b5223872b9571641119c0bda190921b621_aq_0.11_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```

aruhier:
Indeed, I converted my laptop config to lua, and this patch doesn't fix it. I'll debug it further.

Edit: I think a change broke my patch since I tested it, I just rebased on main to do this PR. I wrote and tested this patch on 0.55.0, so I tried to go back to 0.55.0 and applied this PR with a lua config. It works, so I'm going to bisect.

alba4k:
Alright, thanks. 

BTW https://github.com/hyprwm/Hyprland/commit/b76a9e07229c85298d44423b77d8561c8786f32d works, actually, but just when trying to re-enable a monitor via hyprctl directly. Lua still fails to re-enable without `hyprctl reload` however

aruhier:
Can you provide an example? A config snippet, so I can understand exactly what you mean and test on my side.

alba4k:
```lua
hl.bind("switch:off:Lid Switch", function()
        if layout == nil then -- not a known layout where eDP-1 should stay off
            hl.exec_cmd("hyprctl reload")
        end
end, {locked = true})

    if layout == nil then
        hl.exec_cmd("hyprctl reload")
    end
end)
```

You can use `function DetectScreenLayout() return nil end` to test I guess

These both fail to re-enable the screen and need the `hyprctl reload` to do so

aruhier:
> Indeed, I converted my laptop config to lua, and this patch doesn't fix it. I'll debug it further.
> 
> Edit: I think a change broke my patch since I tested it, I just rebased on main to do this PR. I wrote and tested this patch on 0.55.0, so I tried to go back to 0.55.0 and applied this PR with a lua config. It works, so I'm going to bisect.

Ok, about this I know what happened: when I tested with the lua config, I mixed a bit my testing and that's what ended up not working: I disabled the monitor with `hl.monitor({…, disabled = false})`, so kanshi couldn't wake it up.

However, without doing this, this PR works on top of `main`, even on a lua config. I'll check with your snippet. Can you paste your `DetectScreenLayout()` function?

alba4k:
> However, without doing this, this PR works on top of main

What works? I thought the thing you mention not working was exactly the single case this pr was supposed to fix

> Can you paste your DetectScreenLayout() function?
As said, you can just test with `return nil`. But It looks like this:
<details>

```
function DetectScreenLayout() -- global, also needed in binds.lua
    local mons = hl.get_monitors()
    local screens = {}
    local layout = nil

    for _, mon in ipairs(mons) do
        -- Check for "home" setup (two SMBX2450)
        if mon.description:find("SMBX2450") then
            table.insert(screens, mon.name)
            if #screens == 2 then
                layout = "home"
                break
            end
        -- Check for "zurich" setup (one DELL P3425WE)
        elseif mon.description:find("DELL P3425WE") then
            screens[1] = mon.name
            layout = "zurich"
            break
        end
    end

    return layout, screens
end
```

</details>

btw I opened a discussion regarding monitors always turning on on reload: https://github.com/hyprwm/Hyprland/discussions/14541

edit: They both stop working if there is no active monitor, for example in this scenario:
- eDP-1 disabled, DP-1 enabled
- disconnect DP-1
- `hl.on("monitor.removed", ...)` doesn't turn eDP-1 back on. That only happens when adding `hl.exec_cmd("hyprctl reload")` to the lambda

I assumed the bind was also borked because I was trying to use it as a fallback (to turn back eDP-1 after `hl.on(...)` failed) and it was also failing.

Adding a reload after either of the two works

So the enable in the bind is actually fine. The problem is that it never gets executed when no monitor is connected

vaxerski:
I can't repro the issue see my comment on the previous mr.

aruhier:
> What works? I thought the thing you mention not working was exactly the single case this pr was supposed to fix

What now works and what 8726a7363eb9213235e6e23657668e0daac39a5b broke is to kanshi to re-enable my internal monitor (or anything using wlr-output-management) when undocking my laptop, as explained in my first comment of https://github.com/hyprwm/Hyprland/discussions/14230#discussion-9984932. In the context of only using kanshi to manage my screens.

However, that might be only a "laptop thing", only happening to the internal monitor (on Intel and Qualcomm GPU) when I disconnect the USB-C cable to my external screen. With the cable still connected, it works to do `kanshictl switch` to a profile that disables the external monitor and enables the internal one. The reason seems to be that the monitor config reload is never triggered when disconnecting the USB-C cable, even if kanshi sees the event and switches to the correct profile. Which this PR mitigates by forcing it to happen by scheduling it in the event loop.

Mixing `hl.monitor({…, disabled = true })` and wlr-output-management might be buggy, because kanshi in that case is not able to re-enable to screen. But this might be a different bug: enabling a screen in kanshi might not remove the disabled flag of the monitor config.

Edit: I reworded my PR comment to make the problem clearer.

vaxerski:
https://github.com/hyprwm/Hyprland/pull/14547 should fix this

aruhier:
I close this PR then, as it's not solving the real problem. I'll follow #14547, thanks a lot @vaxerski!

PR #14543, opened 2026-05-14
remove manual format handling and pixelread parsing over to hyprgraphics

gulafaran:
use newly added format helpers from hyprgraphics, and dont manually if else case the drm formats, stick to the format list we have in hyprgraphics to avoid conflicts or future missing formats.

requires: https://github.com/hyprwm/hyprgraphics/pull/46


gulafaran:
tested it as far as i can, various obs capturing. 8bit/10bit checked, hyprshot, https://mozilla.github.io/webrtc-landing/gum_test.html, and made claude hallucinate together a stupid tool. seems to work. and should make any future format shenanigans way less likely.

fxzzi:
also working fine on my end. tested in and out of DS, with / without HDR, different capture tools, everything seems to be in order

PR #14547, opened 2026-05-14
state/monitor: refactor monitor state, init fallback state

vaxerski:
Instead of a broken unsafe state, support proper fallback, with a real output, and pure headless runs

This should finally make Hyprland run just fine in headless environments, and fixes problems with monitor reloading like https://github.com/hyprwm/Hyprland/pull/14447#issuecomment-4451490727

@alba4k for testing please


vaxerski:
also IIRC @System64fumo might be interested in the headless part

alba4k:
When I first started the session (I just had my laptop), the `FALLBACK` output was enabled and claiming workspaces. It also claimed the space for the errors (due to plugins not being loaded), meaining that they covered part of the windows on my laptop

I also couldn't disable it with `hyprctl eval 'hl.monitor({output = "FALLBACK", disabled = true})'`

I then tried to dock my laptop and connected two external monitors. I now only have the two monitors enabled, while `FALLBACK` and `eDP-1` are correctly disabled ([monitors.lua](https://github.com/user-attachments/files/27767822/monitors.lua.txt))

alba4k:
disconnecting form the dock (which is then `hl.on("monitor.removed", ...)` tried to re-enable `eDP-1`, without reload) crashed hyprland

[hyprlandCrashReport279065.txt](https://github.com/user-attachments/files/27767124/hyprlandCrashReport279065.txt)

[coredump.txt](https://github.com/user-attachments/files/27767171/coredump.txt)

vaxerski:
:(

alba4k:
Any monitor update seems to make the `FALLBACK` output disappear though. Even just closing and re-opening the lid (which disables and re-enables `eDP-1`) removes it

However, it's still annoying that it appears in `hyprctl monitors all`

vaxerski:
must be wizard crime

vaxerski:
try now

aruhier:
I hit the same problem.

alba4k:
can confirm, the `FALLBACK` output is still enabled until something updates the monitor layout

however, hyprland didn't crash when I disconnected the dock, and `eDP-1` enabled just fine without hyprctl reload.

This said, after disconnecting the dock, firefox (which was open) crashed, which is probably unrelated to this mr as it happens on 0.55.0/1 too with certain reloads (see https://github.com/hyprwm/Hyprland/discussions/14541#discussioncomment-16919667)

vaxerski:
does it happen on default config or does it require some lua?

Cuz I cant repro

alba4k:
Same monitor lua linked above: ([monitors.lua](https://github.com/user-attachments/files/27767822/monitors.lua.txt))

If I comment out the `hl.on` parts, after undocking Im left with the laptop screen off (which is intended).

However, I can now use a keybind to `hyprctl reload`  or close and reopen the lid (which was _not_ the case before! which is the point of https://github.com/hyprwm/Hyprland/discussions/12325) and get back to a usable state as that re-enables my laptop's screen

vaxerski:
yeah no I mean the initial bug

alba4k:
Yes, tried with safe mode and `FALLBACK` is still active until I disable and re-enable `eDP-1`

vaxerski:
build commit and give log

alba4k:
[hyprland.log](https://github.com/user-attachments/files/27775403/hyprland.log)
looking good. display is now disabled when hyprland starts (and seems to stay disabled, 0x0 at -1x-1)

laptop screen turns on when undocking, or when reloading after all outputs have been disabled

`FALLBACK` is still visible in `hyprctl monitors all`, but can't be manually enabled (probably intended). If you plan on keep it visible there, maybe add a description to make it clear that it's intended to be there

alba4k:
Just got a crash: 
[hyprlandCrashReport58154.txt](https://github.com/user-attachments/files/27777383/hyprlandCrashReport58154.txt)
[coredump.txt](https://github.com/user-attachments/files/27777508/coredump.txt)

I wasn't doing anything special - had yt in the background, telegram, discord and whatsapp open and was trying to resize them by dragging the corners. Can't reproduce. Don't know if it's related

aruhier:
I tested with the last commit, it works for me!

Though one thing: FALLBACK is exposed in wlr-output-management and hyprctl monitors all, do you think it could be just kept as internal in a next iteration? Beside having to handle it in kanshee for the optional output match, for example I initially messed up by setting this output as disabled in kanshee, and I couldn't get the internal monitor enabled anymore, so I don't think the user should be able to mess up with the fallback virtual monitor (maybe it should be hidden from the list the monitors?).

It also make waybar to bug and be transparent without any log when changing output, but I can maybe dig into that later.

> Just got a crash: [hyprlandCrashReport58154.txt](https://github.com/user-attachments/files/27777383/hyprlandCrashReport58154.txt) [coredump.txt](https://github.com/user-attachments/files/27777508/coredump.txt)
> 
> I wasn't doing anything special - had yt in the background, telegram, discord and whatsapp open and was trying to resize them by dragging the corners. Can't reproduce

That might be unrelated, I had similar crashes on my desktop (in the behavior, I didn't take a crash report or coredump so I can't compare) without this MR while testing my animations changes and moving windows around a lot.

Edit: rectifying what I tested: kanshi is able to disable/enable the internal screen when I dock/undock my laptop.

vaxerski:
ah right. FALLBACK should be fully removed from protocols outside of fallback state

@alba4k can you please make your comments more explicit I swear to god I have NO fucking idea what you are saying and if it works fine or not

alba4k:
works fine other than that random crash (probably unrelared) and it appearing in `hyprctl monitors all`

aruhier:
> ah right. FALLBACK should be fully removed from protocols outside of fallback state

Hmm even in fallback state, if it's listed, it would still be a problem here: if I have kanshi doing `output eDP-1 enable` when I unplug the external monitor, but FALLBACK is listed, if I don't make kanshi aware of this (by having a stub `...output FALLBACK`) it won't match and stay in fallback state.

I don't know how other compositors handle this.

vaxerski:
I added a lot of shit, much more major refactor, but FALLBACK should now disappear as long as its not in fallback state.

> Hmm even in fallback state, if it's listed, it would still be a problem here: if I have kanshi doing output eDP-1 enable when I unplug the external monitor, but FALLBACK is listed, if I don't make kanshi aware of this (by having a stub ...output FALLBACK) it won't match and stay in fallback state.

Great! I don't give a shit about kanshi. God gave you lua, fucking use it.

> I don't know how other compositors handle this.

IIRC sway does exactly this.

alba4k:
> God gave you lua, fucking use it.

Not really a perfect fix, kanshi/shikane keeps monitors disabled on reload without weird things happening. off topic tho

---

After building https://github.com/hyprwm/Hyprland/pull/14547/commits/0e3d6b1eed79b9204467adadcde5ad52702b0a1b I see `eDP-1` _twice_ in `hyprctl monitors`  (and only a single eDP-1 in `hyprctl monitors all`, no FALLBACK ). I get the "overlapping outputs in your layout" warning notification

Running `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})' && sleep 1 && hyprctl reload` fixed that and I now only have a single eDP-1 in `hyprctl monitors (all)`

Running `hyprctl output create headless` also creates _two_ overlapping active outputs `HEADLESS-1` (which, again, only appears once in `hyprctl monitors all`)

Same with FALLBACK when entering the fallback state with `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})' && sleep 3 && hyprctl eval
 'hl.monitor({output = "eDP-1", disabled = false})'`

TLDR: whenever an output is added for the first time, it is activated with two overlapping placements

alba4k:
secondly, `FALLBACK` is hidden from `hyprctl monitors all` and from `shikanectl export`, but seen by wf-recorder after having entered fallback once:

```
Please select an output from the list to capture (enter output no.):
1. Name: FALLBACK Description: 
2. Name: eDP-1 Description: LG Display 0x06CE  (eDP-1)
Enter output no.:
```

edit: also appears in `hyprland-share-picker`. both use screencast

it's also claiming workspaces, I think, cause I see workspace 4 active in waybar but can't access it

alba4k:
Third, running `disable eDP-1; sleep 3; enable eDP-1` while in the state with 2 eDP-1 placements consistently crashes hyprland:

[hyprlandCrashReport53671.txt](https://github.com/user-attachments/files/27791790/hyprlandCrashReport53671.txt)
[coredump.txt](https://github.com/user-attachments/files/27791813/coredump.txt)

vaxerski:
try

alba4k:
I dont see monitors appearing twice anymore, however hyprland now crashes when going to fallback (crash report was with `disable eDP-1; sleep 3; enable eDP-1`)

[hyprlandCrashReport337684.txt](https://github.com/user-attachments/files/27804352/hyprlandCrashReport337684.txt)
[coredump.txt](https://github.com/user-attachments/files/27804358/coredump.txt)

vaxerski:
can you build in debug and provide a bt -full thanks

alba4k:
Here ya go

[hyprlandCrashReport7647.txt](https://github.com/user-attachments/files/27807400/hyprlandCrashReport7647.txt)
[coredump.txt](https://github.com/user-attachments/files/27807372/coredump.txt)
[bt_full.txt](https://github.com/user-attachments/files/27809108/bt_full.txt)
[hyprland.core](https://limewire.com/d/AoQ45#AR9WNqBG62) to help you replicate

vaxerski:
thanks. Crash fix pushed

alba4k:
Still crashes on `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})'`

[hyprlandCrashReport62241.txt](https://github.com/user-attachments/files/27810059/hyprlandCrashReport62241.txt)

How can I get a debug build that loads my config? previously used `cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_INSTALL_PREFIX:STRING=/usr -S . -B ./build`

vaxerski:
pass `--config /path.lua`

alba4k:
I mean, sure

vaxerski:
post bt full of crash

alba4k:
[hyprlandCrashReport132166.txt](https://github.com/user-attachments/files/27813253/hyprlandCrashReport132166.txt)
[bt_full.txt](https://github.com/user-attachments/files/27813301/bt_full.txt)

vaxerski:
check now

alba4k:
[hyprlandCrashReport184286.txt](https://github.com/user-attachments/files/27815263/hyprlandCrashReport184286.txt)
[bt_full.txt](https://github.com/user-attachments/files/27815334/bt_full.txt)

:)

tried running `hyprctl keyword monitor eDP-1,disabled; sleep 5; hyprctl reload`

vaxerski:
ah yeah

vaxerski:
fixed

alba4k:
Ok, yeah. Couple things I noticed now:
- `hyprctl keyword monitor eDP-1,disabled; sleep 5; hyprctl reload` now works, but when the screen turns back on I'm left on a new, empty workspace (which I'm assuming was created on the FALLBACK output and then got moved to eDP-1)
- I can get a crash when using my config (and thus [monitors.lua](https://github.com/user-attachments/files/27767822/monitors.lua.txt), using its `home` layout) when undocking:
        * I disconnect the USB-C cable
        * Laptop display turns on (as it should, because of that script) but is frozen
        * after a few seconds, hyprland crashes

[hyprlandCrashReport211032.txt](https://github.com/user-attachments/files/27815960/hyprlandCrashReport211032.txt)
[bt_full.txt](https://github.com/user-attachments/files/27815994/bt_full.txt)

alba4k:
If I comment out the monitor.removed part of the script and undock and try to `hyprctl reload` with a keybind I get a similar crash, except the screen doesn't even turn on. probably just the same thing as before but, just in case:

[hyprlandCrashReport23852.txt](https://github.com/user-attachments/files/27816201/hyprlandCrashReport23852.txt)
[bt_full.txt](https://github.com/user-attachments/files/27816220/bt_full.txt)

vaxerski:
thanks albania4k fixed

alba4k:
u sure?

[bt_full.txt](https://github.com/user-attachments/files/27819067/bt_full.txt)
[hyprlandCrashReport366855.txt](https://github.com/user-attachments/files/27819070/hyprlandCrashReport366855.txt)
[hyprlandCrashReport362291.txt](https://github.com/user-attachments/files/27819071/hyprlandCrashReport362291.txt)

same 2 scenarios as before as far as I can tell

vaxerski:
image

alba4k:
Trying out `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})' && sleep 3 && hyprctl reload`, I noticed that `hyprpaper` (almost) always crashes:
<details>

```
DEBUG ]: Welcome to hyprpaper!
built from commit  ()
DEBUG ]: Found 1 output(s)
ERR ]: targetChanged but  has no output?
wl_display#1: error 0: invalid object 22

==========================================================================================
ASSERTION FAILED! 

[core] Disconnected from pollfd id 0

at: line 367 in Backend.cpp[HT] Assertion failed!fish: Job 1, 'hyprpaper' terminated by signal SIGABRT (Abort)
```

</details>

[coredump.txt](https://github.com/user-attachments/files/27824227/coredump.txt) (stable hyprpaper, probably useless)

And, as before, I'm still left on a new, empty workspace when the screen comes back on (probably the one that had been assigned to FALLBACK)

Furthermore, still crashing on undock:

[hyprlandCrashReport32981.txt](https://github.com/user-attachments/files/27824590/hyprlandCrashReport32981.txt)
[hyprlandCrashReport13646.txt](https://github.com/user-attachments/files/27824592/hyprlandCrashReport13646.txt)
[bt_full.txt](https://github.com/user-attachments/files/27824602/bt_full.txt)

vaxerski:
> And, as before, I'm still left on a new, empty workspace when the screen comes back on (probably the one that had been assigned to FALLBACK)

yes, that likely wont change

crash fixed check

alba4k:
Undocking the first time was fine, second time crashed.
I then tried `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})' && sleep 3 && hyprctl reload` and again, first time worked fine but second time crashed.

[hyprlandCrashReport56346.txt](https://github.com/user-attachments/files/27856018/hyprlandCrashReport56346.txt)
[bt_full.txt](https://github.com/user-attachments/files/27856038/bt_full.txt)

vaxerski:
duh what would you expect? a working system? lolol

vaxerski:
should be fixed

alba4k:
Yeah can now reload fine, however:
- still get hyprpaper and others to often crash
- I only get placed on a new empty workspace after the first restore from fallback, on the following ones I get left on the same one I was before (not an issue, just worth pointing out ig)
- After undocking, `monitors.lua` fails to turn back on eDP-1 (which it should do as per `hl.on("monitor.removed", ...)`). reloading with a keybind turns it on successfully
- On my first session, after undocking, I got two ghost workspaces (the two that used to be on the external monitors), which appeared on monitor `?` in `hyprctl workspaces`. I can't replicate this anymore.
- Shortly after, while I was trying to report the thing above, hyprland crashed. might be unrelated to this pr but here you go:
[hyprlandCrashReport3434.txt](https://github.com/user-attachments/files/27859408/hyprlandCrashReport3434.txt)
[bt_full.txt](https://github.com/user-attachments/files/27859424/bt_full.txt)

alba4k:
regarding the last two point: happened again. I had two ghost workspaces after undocking (and reloading the config to turn the laptop back on). System was still usable. A couple of seconds after, hl freezed and then crashed

vaxerski:
fixed

aruhier:
No crash for me, but with kanshi (I know, lua yada yada yada), with this:

```
profile internal {
  output eDP-1 enable
  ...output FALLBACK
}
```

It gets in a weird state on monitor disconnect where eDP-1 stays off but is shown as `disabled: false` by hyprctl monitors. However, doing `kanshictl switch internal` when the external screen is connected works well.

Reconnecting the external screen after disconnecting it, I can see that all my apps (firefox, foot terminal, thunderbird) crashed.

alba4k:
I got a crash again after undocking (screen stayed frozen for a while in the "added monitor" animation, then hl died). I got multiple successful undocks before this one though

[bt_full.txt](https://github.com/user-attachments/files/27862937/bt_full.txt)

Regarding the previous undocks being fine: I get that crash with:
```lua
    if layout == nil then
    end
end)
```
but this works fine
```lua
    if layout == nil then
        hl.exec_cmd("hyprctl reload") --!!!!!!!!!
    end
end)
```
So having the reload after the re-enable (or even just the reload by itself) seems to mitigate it

vaxerski:
these are all because of how we queue those dam events aarggghhhh this needs a rethink there

vaxerski:
this is all awful because it's ancient retarded code I wrote in 2022 fucking hell I might do an even larger refactor here

aruhier:
Just for my curiosity: why did it work before without a fallback monitor? Did it work by luck because there was a small gap between when one monitor was destroyed and the other monitor was enabled?

alba4k:
It's always been a hit or miss for me tbh, my laptop often never turning back on after undocking was one of my most common issues with hyprland (even with kanshi), ever since the switch to aquamarine iirc

vaxerski:
it didnt work before

vaxerski:
hopefully this improves stuff

vaxerski:
lmk, it does work fine on my end

aruhier:
I think it's getting closer: there still seems to be a race condition in the events, with kanshi it gets in a state that both screen are set as `disabled: false`, but the external screen still stays off. However, the external screen receives dpms signals, and I see (in `hyprctl workspaces`) that a workspace is active on it. I guess it's the rendering part that is not synced.

Edit: to clarify, with kanshi while using 2 profiles supposed to either enable only the external screen when docked, and re-enable the internal screen when undocked. So the screens are not supposed to be both enabled here.

vaxerski:
how do I repro this

aruhier:
Use this kanshi config:

```
profile "default" {
  output eDP-1 mode "2944x1840@90.00Hz" scale 1.6 enable
  ...output FALLBACK
}

profile "external" {
  output DP-2 enable mode 3840x2160@120 position 0,0 scale 1.25
  output eDP-1 disable
  ...output FALLBACK
}
```

Plug / unplug an external screen (adapt the output for the external monitor in kanshi).

aruhier:
Pfff what a mess to reproduce this, sorry… I retested again, now both screens are on. Starting kanshi with hyprland seems to mess with the monitors state more often, that might help to reproduce the bug.

vaxerski:
I cant seem to repro this in a nest. 

```
profile "default" {
  output WAYLAND-1 enable
  ...output FALLBACK
}

profile "external" {
  output WAYLAND-2 enable
  output WAYLAND-1 disable
  ...output FALLBACK
}
```

I add WAYLAND-2, -2 is enabled, -1 is disabled. I remove -2, -1 is enabled.

aruhier:
I guess that my external screen might flicker when connecting to usb-c, because I see quick events of added/removed screen, maybe that's what's messing with the monitors states.

I'll add more debugging to understand what's going on.

alba4k:
Found an issue:
- run `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = false, mode = "preferred", position = "auto"})'`
- run `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})'`
- nothing happens. screen doesn't disable anymore

everything is fine if I just disable the screen and turn it back on with `hyprctl reload`. I can repeat that as often as I wish. But after setting `disabled = false` I can't disable it anymore™️ (see edit)

edit: if I `hyprctl reload` that state is reset, i.e. I can disable it again. But if I re-enable it manually, then again I can't disable it

alba4k:
I also got a crash after undocking, I don't know why other previous undocks worked fine but this crashed:
[bt_full.txt](https://github.com/user-attachments/files/27904650/bt_full.txt)

aruhier:
> Found an issue:
> 
>     * run `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = false, mode = "preferred", position = "auto"})'`
> 
>     * run `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})'`
> 
>     * nothing happens. screen doesn't disable anymore
> 

Nice, it feels like the same thing as what I'm hitting with kanshi (when both screens are stuck on being enabled).

aruhier:
I debugged it with an LLM (the change touches at too much internal logical that I'm lost if I try to follow up), and I managed to get something working. @alba4k, feel free to try.

I kept the change to the minimal to make it work, and I've split it by patch by what it fixes:
* [0001-Allow-to-enable-screen.patch](https://github.com/user-attachments/files/27937259/0001-Allow-to-enable-screen.patch): without this, the monitor stays disable.
* [0002-Hide-FALLBACK-from-wlr-output-management.patch](https://github.com/user-attachments/files/27937263/0002-Hide-FALLBACK-from-wlr-output-management.patch): optional, hide FALLBACK from wlr output management list
* [0003-Fix-crash-on-monitor-getting-destroyed.patch](https://github.com/user-attachments/files/27936815/0003-Fix-crash-on-monitor-getting-destroyed.patch): I wanted to debug and launched `hyprctl monitors all` after disconnecting my external screen, to understand what was happening. I realized it made hyprland to crash everytime while in fallback mode, this fixes it
* [0004-Fix-disable-monitor.patch](https://github.com/user-attachments/files/27936814/0004-Fix-disable-monitor.patch): allows to disable an enabled monitor
* [0005-Fix-crash-in-fallback.patch](https://github.com/user-attachments/files/27936813/0005-Fix-crash-in-fallback.patch): while in fallback, I wanted to understand what was happening from the wlr output management perspective, I started `sleep 10; wlr-randr` in a terminal before disconnecting my screen, realized it made Hyprland to crash. This fixes it. 

With all of this, it seems ok. One thing is that kanshi doesn't seem to receive the event *after* the monitor has been destroyed, so it still tries to use the profile "external-screen" after being undocked. But if I force it with a stub output just to make it match with the profile, it works and kanshi can re-enable the internal screen. So there might be something about the event sent too early.

@vaxerski: the patches are probably not a good way to fix things, I thought it would be better than nothing to help you understand what's going on.

Update: `0002-Hide-FALLBACK-from-wlr-output-management.patch` was including a fix for the previous problem, changed patches 0001 and 0002.

aruhier:
About the last part and kanshi not matching with the good profile, this might be linked with the crash of `hyprctl monitors all` that I fixed by ignoring bad monitor states. The external screen might be stuck to be destroyed, and still listed, and that's why kanshi would match.

The weird part is that wlr-randr shows only the internal screen, in that case.

aruhier:
I managed, with LLM, to fix the last issue, now docking and undocking works. So based on 4af1e1f0b760880d52178cd31892755ea122a3f0 (@vaxerski,  your last commit), here is what I got:

* [0001-Allow-to-enable-and-disable-displays.patch](https://github.com/user-attachments/files/27970965/0001-Allow-to-enable-and-disable-displays.patch): without this, I can't enable/disable the screens if they have been disabled or enabled before. I think moving the `m_fallbackActive = enabled;` a bit up is a hack that workarounds quick calls to this function when a m_fallbackStateUpdateDeferred is pending.
* [0002-Fix-random-crashes-and-segvs.patch](https://github.com/user-attachments/files/27970964/0002-Fix-random-crashes-and-segvs.patch): Fixes the different crashes I encountered. One is for example running `hyprctl monitors all` with a broken state on my external monitor. It shouldn't happen anymore, but it seems good to keep this safeguard.
* [0003-Resolve-ghost-monitors-and-wrong-scaling-exiting-fal.patch](https://github.com/user-attachments/files/27970963/0003-Resolve-ghost-monitors-and-wrong-scaling-exiting-fal.patch): Without this, I had a "ghost" display after unplugging the external monitor: while in fallback, my external monitor would still be detected by kanshi, therefore it never switched to the profile of my internal monitor. This destructs the monitor so that kanshi detects the right profile. 
  Also, after fixing all of this, I realized that the default scaling was all messed on my internal screen after undocking: the scaling was set at 1.5 automatically (that is expected) but the region was wrong and the screen cropped. This fixes it.

I've let the LLM description of each change in each patch, so you can understand a bit more the logic, though I removed a bunch of crap and corrected the directions many times. The `keepOutputAlive` in the last patch seems really dirty, and maybe will lead to leaks or unsafe access. Again, it's more for you to understand what could be wrong with this PR.

I tried undocking/docking multiple times, it works with kanshi and by disabling / reenabling the screen through hyprctl, and I had no crash so far.

Update: I pushed these 3 patches on https://github.com/aruhier/Hyprland/tree/pr-14547.

alba4k:
I'll try lua tomorrow

aruhier:
> I'll try lua tomorrow

I'm gonna try lua on my laptop with the config converted to lua in ~2 hours.

aruhier:
> Found an issue:
> 
>     * run `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = false, mode = "preferred", position = "auto"})'`
> 
>     * run `hyprctl eval 'hl.monitor({output = "eDP-1", disabled = true})'`
> 
>     * nothing happens. screen doesn't disable anymore
> 

I tested this, it works. It also works with hyprlang.

I pushed my patches on https://github.com/aruhier/Hyprland/tree/pr-14547.

aruhier:
One thing to note: hyprlock crashes with monitor hot swaping. I'll check what it is, but I guess it's something specific to hyprlock considering that none of my apps are crashing otherwise.

Update: ~~I updated hyprlock to the last commit on `main`, it seems to solve the problem.~~ Nope, it still happens, I'll debug it and see if there's something I can fix in hyprlock.

alba4k:
Not facing that. btw I checked out your fork and got the same random crash when undocking. Enabling and disabling was fine tho

aruhier:
> Not facing that. btw I checked out your fork and got the same random crash when undocking. Enabling and disabling was fine tho

Can you share the BT and crashlog?

aruhier:
> Not facing that. btw I checked out your fork and got the same random crash when undocking. Enabling and disabling was fine tho

On my side, I've been using this for 3 days on my work laptop that I dock/undock multiple times per day, I haven't encountered a crash, it has worked reliably.

vaxerski:
@alba4k that crash was inside aq, will need aq debug as well for that

vaxerski:
@aruhier some patches there were good, but the last one raised a lot of eyebrows and looked more like an awful band-aid instead of looking into the actual problems

vaxerski:
what are the current issues remaining?

aruhier:
Nice, I'm happy it helped! I guessed that the last patch was dirty as hell, more an indicator to you to point to the right direction of wtf was going on.

I'm leaving a comment in setFallbackActive because I thought about a potential (rare) case when an update would be missed.

src/state/FallbackState.cpp
aruhier, line 140: I thought about one case where the update here would be missed if this function is called at a rare bad timing. Because this is non atomic, I think there's a small frame between when the m_fallbackActive is changed by a call to setFallbackActive and when the event is processing the update.

What about having 2 flags for this: one if the update is queued in the g_pEventLoopManager, in which case you can just change `m_fallbackActive` like right now, and a second flag if this update is being processed?
If this update is being processed (meaning: if it's in this if/else block), if m_fallbackActive would be the same as for the previous update, you can just return early as the update is unnecessary, and otherwise you'll need to queue another update in the g_pEventLoopManager.

I think it would protect this against quick flickers, without queuing unnecessary updates.

Edit: english doesn't work for my brain today, after 6 edits I think this comment might finally be understandable…

aruhier:
> what are the current issues remaining?

I'll test in a few hours. From what I see from your change, we might still hit the scaling issue I had. I'll confirm later.

aruhier:
I tested, unplugging the external monitor didn't work, the internal monitor stayed off. Plugging it again right after triggered a crash.

[hyprlandCrashReport2688214.txt](https://github.com/user-attachments/files/28162067/hyprlandCrashReport2688214.txt)
[hyprlandd.log](https://github.com/user-attachments/files/28162068/hyprlandd.log)
[bt-full.txt](https://github.com/user-attachments/files/28162164/bt-full.txt)

ndavd:
Tested this branch for https://github.com/hyprwm/Hyprland/discussions/14715:
It made things slightly worse since after switching back to non-HDR it still doesn't update the refresh rate, but while before I was able to reset it by saving the config again or by doing two consecutive `hyprctl reload`, with these changes it is permanently stuck with the lower refresh rate and saving the config and reloading does nothing. The only way of getting back to the original state is by exiting hyprland and relaunching it again.

deviantfero:
just chiming in to say I'm running into the same issue, I've downgraded to 0.54 in the meantime and everything is working fine.

alba4k:
> I tested, unplugging the external monitor didn't work, the internal monitor stayed off. Plugging it again right after triggered a crash.

Can't reproduce this, it's working flawlessly on my machine. Maybe kanshi-related?

The only issue that still consistently remains on my end is the monitor not being disabled after having been explicitly enabled once (https://github.com/hyprwm/Hyprland/pull/14547#issuecomment-4471800998)

vaxerski:
@aruhier again the crash is inside aq, needs aquamarine in debug too.

alba4k:
@vaxerski got a crash after undock, hopefully this helps (debug aquamarine)
[bt_full.txt](https://github.com/user-attachments/files/28229742/bt_full.txt)
[hyprlandCrashReport223837.txt](https://github.com/user-attachments/files/28229757/hyprlandCrashReport223837.txt)

vaxerski:
looks like memory corruption lol can you do ASan?

alba4k:
do I need to rebuild aquamarine with `-DWITH_ASAN=ON` too? or is hl enough? as I don't see anything set up in `CMakeLists.txt`

alba4k:
I got the following, however `build/Hyprland` still built so I should be able to check?

<details>

```
[100%] Linking CXX executable Hyprland
[100%] Linking CXX executable hyprland_gtests
[100%] Built target Hyprland
==21016==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
CMake Error at /usr/share/cmake/Modules/GoogleTestAddTests.cmake:381 (message):
  Error running test executable.

    Path: '/home/alba4k/Hyprland/build/hyprland_gtests'
    Working directory: '/home/alba4k/Hyprland/build'
    Timeout: '5'
    Result: 1
    Output:
      

Call Stack (most recent call first):
  /usr/share/cmake/Modules/GoogleTestAddTests.cmake:421 (gtest_discover_tests_impl)


make[2]: *** [CMakeFiles/hyprland_gtests.dir/build.make:579: hyprland_gtests] Error 1
make[2]: *** Eliminazione del file «hyprland_gtests»
make[1]: *** [CMakeFiles/Makefile2:1287: CMakeFiles/hyprland_gtests.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

</details>

anyway, I tried replicating an undock by adding and removing a headless display and got none. Will try this evening with an actual monitor ig

vaxerski:
yes that error is harmless. Hyprland alone with ASan is OK.

alba4k:
...I can't get a `bt full`? I built after `make clear` and `cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Debug -DWITH_ASAN=ON -S . -B ./build` but gdb just says 
```
#0  0x00007a4a95c256e6 in ?? ()
No symbol table info available.
#1  0x0000000000000000 in ?? ()
No symbol table info available.
```
and I'm unsure why. Here's what I have:
[hyprlandd.log](https://github.com/user-attachments/files/28275099/hyprlandd.log)
[coredump.txt](https://github.com/user-attachments/files/28275135/coredump.txt)
[hyprlandCrashReport19824.txt](https://github.com/user-attachments/files/28275185/hyprlandCrashReport19824.txt)

vaxerski:
with asan it shouldnt crash it should quit and drop an asan log. check https://wiki.hypr.land/Crashes-and-Bugs/#building-the-wayland-stack-with-asan

alba4k:
[CMakeLists.txt.rej](https://github.com/user-attachments/files/28322170/CMakeLists.txt.rej.thankyougithub.txt)

..?

aruhier:
@alba4k the link doesn't work

vaxerski:
ah there might be some conflict, I need to adjust make asan. For now, you can just build normally but make it debug and add `-DWITH_ASAN:BOOL=True` should work

alba4k:
I get the behavior you described but no `asan.log.whatever` got created anywhere
immagine

vaxerski:
check the make asan recipe, there is an env there to make it produce a log instead of printing to srdout

alba4k:
it also didn't print anything to stdout so I doubt that's it? I'll try tho

vaxerski:
stdout could be garbled

alba4k:
I had a normal stdout, which cut cleanly the moment hyprland exited (no "interrupted by SEGV" or whatever either

I can check again this evening

vaxerski:
add the env and it will spit out files, trust me bro

vaxerski:
@alba4k ?

aruhier:
> @alba4k ?

I think you can just merge it, so you dont wait on that to merge other changes. It's anyway better than the current state, and the crashes in aquamarine can happen later

vaxerski:
well I cant merge it due to conflicts thats first and second memory corruption is quite bad, has to be looked into. If you can run with ASan, please do.

alba4k:
I can check this crash in a few mins

alba4k:
@vaxerski makefile has `-DWITH_ASAN:STRING=True -DUSE_TRACY:STRING=False -DUSE_TRACY_GPU:STRING=False`, I need to use `-DWITH_ASAN:BOOL=True -DUSE_TRACY:BOOL=True -DUSE_TRACY_GPU:Bool=False`, correct?

vaxerski:
you can disable tracy, that doesnt matter

alba4k:
nvm I get what you meant now. wtf is wrong my reading comprehension

alba4k:
[asan.log.23262](https://github.com/user-attachments/files/28420252/asan.log) sorry for taking this long

vaxerski:
what the fck

vaxerski:
this should now be fixed with aquamarine-git, can you check?

vaxerski:
rebased on main, a retest would be nice. If everything is good, this is probably gtg

ndavd:
Just tested it regarding https://github.com/hyprwm/Hyprland/discussions/14715, same behavior as described in my previous comment: 

> It made things slightly worse since after switching back to non-HDR it still doesn't update the refresh rate, but while before I was able to reset it by saving the config again or by doing two consecutive `hyprctl reload`, with these changes it is permanently stuck with the lower refresh rate and saving the config and reloading does nothing. The only way of getting back to the original state is by exiting hyprland and relaunching it again.

aruhier:
I'm gonna test.

@ndavd I wonder if it's not linked to the same scaling bug I had: the screen was enabled but the config was then changed without being correctly reapplied to the screen after unplugging (my understanding of what was happening). Because the internal state matched with the reload, it didn't reapply the config on reload.

Can you try if changing the resolution works?

ndavd:
> Can you try if changing the resolution works?

Tried changing the resolution and then changing back to the desired setting but had no effect.

alba4k:
`hyprctl eval "hl.monitor({output = 'eDP-1', disabled = true})"` while on laptop crashed hl
[bt_full.txt](https://github.com/user-attachments/files/28422000/bt_full.txt)
[hyprlandCrashReport18695.txt](https://github.com/user-attachments/files/28422007/hyprlandCrashReport18695.txt)
[coredump.txt](https://github.com/user-attachments/files/28422029/coredump.txt)

ndavd:
And apart from that, I've just got a crash.

[hyprlandCrashReport30012.txt](https://github.com/user-attachments/files/28422387/hyprlandCrashReport30012.txt)
[coredump.txt](https://github.com/user-attachments/files/28422379/coredump.txt)
[hyprland.log](https://github.com/user-attachments/files/28422385/hyprland.log)

vaxerski:
@alba4k fixed

@ndavd update aq to git, it's the same crash as alba had

ndavd:
> @ndavd update aq to git, it's the same crash as alba had

I'm on the `aquamarine-git` package from AUR.

alba4k:
are you sure it's up to date? the commit you need is from 2h ago or so. `pacman -Q aquamarine-git` should be `aquamarine-git 0.12.0.r1.g1699271-1`

vaxerski:
@ndavd regarding the switching issue, I can't repro it with headless. It works fine on my end.

ndavd:
> @ndavd regarding the switching issue, I can't repro it with headless. It works fine on my end.

I suspect it must be a hardware issue. But it's odd that on `main` I'm able to get back to original state but here I'm not :/

ndavd:
> are you sure it's up to date? the commit you need is from 2h ago or so. `pacman -Q aquamarine-git` should be `aquamarine-git 0.12.0.r1.g1699271-1`

That's what I have too:
```
pacman -Q aquamarine-git
aquamarine-git 0.12.0.r1.g1699271-1
```

vaxerski:
@ndavd yeah sorry tried on eDP-1 too on a real session and it all works.

alba4k:
I got a crash on undock too:
[hyprlandCrashReport150496.txt](https://github.com/user-attachments/files/28422820/hyprlandCrashReport150496.txt)
[bt_full.txt](https://github.com/user-attachments/files/28422836/bt_full.txt)

alba4k:
I also have an issue where after resuming from fallback (e.g. with a reload keybind) displays appear scaled weirdly:
immagine
^ a screenshot of what I see. it's as if the whole eDP-1 display is only showing the top-left quarter. Only happens when I comment out my monitor configuration.

Programs also now crash when quickly entering and leaving fallback, e.g. when you have `hl.on("monitor.removed", function() hl.exec_cmd("hyprctl reload") end)` and disabling eDP-1, see https://github.com/hyprwm/Hyprland/discussions/14485. Try having firefox open, it will have crashed when the screen comes back on

alba4k:
BTW this fallback state should be documented somewhere again as it can cause unexpected behavior, e.g. `"monitor.added"` being triggered whenever the last monitor is removed

aruhier:
> I also have an issue where after resuming from fallback (e.g. with a reload keybind) displays appear scaled weirdly: immagine ^ a screenshot of what I see. it's as if the whole eDP-1 display is only showing the top-left quarter. Only happens when I comment out my monitor configuration.
> 
> Programs also now crash when quickly entering and leaving fallback, e.g. when you have `hl.on("monitor.removed", function() hl.exec_cmd("hyprctl reload") end)` and disabling eDP-1, see #14485. Try having firefox open, it will have crashed when the screen comes back on

About the scaling, I had the same issue on my branch.

@vaxerski: you can look at the scaling related changes in https://github.com/aruhier/Hyprland/commit/3d76676f1676f84cc8ed5ee2174e0ab7db73a5e6, this fixed the scaling problem (probably not well, but it can give you a hint on what is going on).

aruhier:
If I remember correctly, and from what I understood, it's because the default scaling is calculated before the screen is enabled. So default scaling sets it to 1, because of:

```
float CMonitor::getDefaultScale() {
    if (!m_enabled)
        return 1;
```

The config calculates and set the automatic default scale but the rendering is still kept at 1. So you have a mismatch that makes the region looking cropped.

alba4k:
I had a similar issue on my ultrawide, where all the screen was "shifted" by half a width (the left half was empty, the right half showed the left half). Windows moved back to the right place after opening one, and layers also moved to the right place after killing waybar. this was with my config enabled.

ndavd:
About my issue I guess I'll completely remove the HDR desktop<>fullscreen toggle as I normally only use the latter. It's probably an hardware issue, but going to leave the discussion open in case someone finds a solution eventually.

So from my side the remaining issue is that the crash still happens (even with aquamarine-git).

vaxerski:
fixed scaling and crash

awsms:
Position-only monitor rule updates are classified as `COMPARISON_SOFT_MISMATCH`, but the PR skips applying those rules. This leaves `m_activeMonitorRule` / monitor layout state stale when switching profiles, eg. multi-monitor HDMI-A-1 at 1600x0 to external-only HDMI-A-1 at 0x0, causing overlap warnings and stale layer-shell geometry until a full Hyprland reload.
Had to remove this bloc to avoid this issue:
```
  if (cmp == COMPARISON_SOFT_MISMATCH)
      continue;
```

alba4k:
starting hyprland while docked, my monitor config was not applied (`zurich` layout, [monitors.lua](https://github.com/user-attachments/files/28429635/monitors.lua.txt)). It was only applied after a reload, after which eDP-1 correctly turned off but DP-3 still looked like this, what I described previously.
immagine

Opening a terminal opened it in the correct position, layers were still shifted. Changing workspace fixed it.

I'm guessing this is because prior to the switch this monitor was placed at 1920x0, so could be what is being described above.

alba4k:
the crash described at the bottom of https://github.com/hyprwm/Hyprland/pull/14547#issuecomment-4583098303 is still there (but that's unrelated to this pr, it's just a new way to repro it)

I also noticed that Hyprland takes strangely long to exit with `hl.dsp.exit()`? it stays frozen for 1-2 seconds on this branch, it's instant on main, fwiw: [hyprlandd.log](https://github.com/user-attachments/files/28429770/hyprlandd.log)

vaxerski:
this should help

aruhier:
It works great for me! No more crash happening.

wlr-output-management doesn't work well in fallback mode, I'll have to debug Shikane/kanshi to understand why. wlr-randr lists FALLBACK and eDP-1 properly, but shikane doesn't detect the profile. I don't know if it's because FALLBACK has no property (Make, Model, Serial).

I'll just set a hook in Lua to force enabling eDP-1 in that case, in the meantime.

awsms:
> this should help

it did, thx

alba4k:
Everything I described above is still happening (layers stay at 1920x0 even after the monitor is repositioned to 0x0, eDP-1 is not being disabled by the function ran on  `monitor.added`, 2-second delay on `exit()` and with no config my laptop only shows a quarter of the screen)


edit: looks like it's crashing on exit, that's why it's delayed. it also doesn't happen consistently, I'll provide a report when I have one

vaxerski:
@alba4k layers should be fixed. For the crash, give a bt full.

alba4k:
can't replicate anymore, sorry :(
issues I'm still having:
- When launched with no config, eDP-1 only shows the top-left quarter of the entire output (i.e. the whole screen only shows that quarted)
- when docking, eDP-1 stays on (should be turned off by  [monitors.lua](https://github.com/user-attachments/files/28429635/monitors.lua.txt) in both `zurich` and `home` layouts, neither works, both work on main), same when starting hl while docked


edit: regarding the second issue: eDP-1 turns off fine if I dock (eDP-1 stays on), reload (eDP-1 turns off), undock (eDP-1 turns on) and dock again (now, eDP-1 turns off).

If I undock and re-dock without first running `hyprctl reload` while docket, thus turning it off, it keeps staying on when docked

dmayle:
Just a note for @alba4k -> the issue with eDP-1 only showing the top-left quarter of the screen is most likely because you don't have a default entry in the config for that screen, and with the high DPI, you're getting a scaling ratio greater than 1 (on my laptop it defaults to 1.5).

alba4k:
I know - as I said, it indeed doesn't happen with a config. It's still unintended behavior though

vaxerski:
I can't repro the zoomed-in myself, I remember being able to without this MR but I cant here.

vaxerski:
@alba4k check now?

kRHYME7:
```
❯ hyprctl version
Hyprland 0.55.0 built from branch makepkg at commit e9f27d9d81fca61cbbceadab00648f8570585553 clean (fix).
Date: Thu Jun 4 21:13:28 2026
Tag: , commits: 7416

Libraries:
Aquamarine: built against 0.12.0, system has 0.12.0

Version ABI string: e9f27d9d81fca61cbbceadab00648f8570585553_aq_0.12_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```
@vaxerski  built on that branch and got scaling issue. But creating headless now will not crash it


 Here is my monitor.
image

what I see  
 
image


* typing this on my external monitor
* * Also compiling and testing took me aroun 30 mins I can't help much 


---

I also think this is isolated in lua as my sddm hyprland config (uses hyprlang is fine ).

kRHYME7:
Ah toggling `scale = "auto"` fixes the error. 
Also setting it to `2` fixes it, I think explicit scale is my workaround for now .


So the auto scaling imple is borke on this build.  

```
hl.monitor({
	output = "eDP-1",
	mode = "preferred",
	position = "auto",
	scale = "auto", -- toggling this fixes the monitor scaling issue
})

```

alba4k:
> @alba4k check now?

same as before on both issue (laptop not turning off, scaling/zoom)

vaxerski:
@alba4k now?

alba4k:
eDP-1 now turns off correctly when docked, scaling still fucked when no config

kRHYME7:
I also noticed this specific issue. (Might not be relevant to the PR though )


try to error

image

Removed the error

image

It fixes itself when a new layer is introduced like `slurp`, rofi etc as long as it is a layer. 


```
ok
❯ hyprctl layers
Monitor eDP-1:
        Layer level 0 (background):
                Layer 5640963c08b0: xywh: 0 0 1500 1000, namespace: awww-daemon, pid: 4786
        Layer level 1 (bottom):
        Layer level 2 (top):
                Layer 564096f6bd80: xywh: 0 87 1500 38, namespace: waybar, pid: 30143
        Layer level 3 (overlay):


Monitor DP-2:
        Layer level 0 (background):
                Layer 564096361b00: xywh: 1500 0 1366 768, namespace: awww-daemon, pid: 4786
        Layer level 1 (bottom):
        Layer level 2 (top):
                Layer 56409922bcf0: xywh: 1500 0 1366 38, namespace: waybar, pid: 30143
        Layer level 3 (overlay):


❯ hyprctl version
Hyprland 0.55.0 built from branch makepkg at commit 15f5078f93cb413ba54c88f85caf184bfa385e41 clean (piss).
Date: Fri Jun 5 03:04:55 2026
Tag: , commits: 7417

Libraries:
Aquamarine: built against 0.12.0, system has 0.12.0

Version ABI string: 15f5078f93cb413ba54c88f85caf184bfa385e41_aq_0.12_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```

alba4k:
noticed that too, a reload usually fixes it, or even changing workspace

getting an error and then fixing it should be enough to repro

vaxerski:
error reserving should be fixed please test

vaxerski:
fixed scale please check

aruhier:
The scale is fixed for me. Sorry I didn't test this in the last fixes, I tried to remove setting the scaling in my monitor config to try with the automatic default scale, it works. In your first commits, I had this issue.

Using Lua events, I can apply the different profiles and I can enable / disable monitors without an issue. Apps don't crash anymore when going through fallback mode.

wlr-output-management still doesn't detect the right profile to apply, but on this I'll do some debugging to understand what shikane/kanshi are seeing. But that's not blocking this PR, `main` is anyway already quite broken with kanshi/shikane, and now they're able to apply a profile it's just that they don't detect the right one.

vaxerski:
@alba4k hows it on your front

alba4k:
lgtm

- docking and undocking works fine
- scaling when no config is fixed
- persisting error reserved area is fixed

Can be merged IMO, I'd expect more issues / crashes to come from people with strange setups but we'll see

---

Apps still crash for me when having fast repeated monitor (dis)connects, e.g. in the following scenario:
- open firefox (good for testing as it _always_ seem to crash, not just sometimes)
- run `hyprctl eval "hl.monitor({output = 'eDP-1', disabled = true})"` with just my laptop
> eDP-1 turns off
> FALLBACK "turns on"
- my config tries to re-apply my layout due to `hl.on("monitor.removed", applyLayout)` and `hl.on("monitor.added", applyLayout)`
> DP-1 turns on
> FALLBACK "turns off"
- when everything comes back to life, firefox (and sometimes other apps like hyprpaper, waybar, swaync) has crashed and I see its session recovery window

**HOWEVER** this issue is not related to this pr, I've encountered it before with specific scenarios (see https://github.com/hyprwm/Hyprland/discussions/14485). it's just an easier way to reproduce it.

kRHYME7:
@vaxerski  


### Scaling

blank scaling defaults to 1, but it should be 2 given my high ppi monitor 

image



config looks like this ( will try to nuke it later ) 
```lua
hl.monitor({
	output = "eDP-1",
	mode = "preferred",
	position = "auto",
	-- scale = "2",
})

```

#### Scaling and set explicitly to "auto"

Works in this case, removing it resets to 1 

image

version
```
Hyprland 0.55.0 built from branch makepkg at commit 49e0190044102184df105fe7d23ee703ced96f8f clean (state/monitor: refactor monitor state, add proper fallback state).
Date: Fri Jun 5 20:20:35 2026
Tag: , commits: 7421

Libraries:
Aquamarine: built against 0.12.0, system has 0.12.0

Version ABI string: 49e0190044102184df105fe7d23ee703ced96f8f_aq_0.12_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```

kRHYME7:
and  I will also add that removing the whole monitor block sets a correct scaling. 

image

@alba4k  can you repro?

alba4k:
Uhm, it's always been like that for me when not explicitly setting scale?

What happens if you set no config for that monitor at all? It correctly defaults to `1.5` for me in that case

edit: saw your second post. yes exactly the same for me, but hasn't it always been like this?

kRHYME7:
> Uhm, it's always been like that for me when not explicitly setting scale?

so it only occurs in lua right as we can ommit other props.  

> What happens if you set no config for that monitor at all? It correctly defaults to 1.5 for me in that case

if the whole block, it sets to 2 (correct). But  if `scale` is not declared it defaults to 1 (wrong). Imo, it should be implied that missing `scale` should be auto.

alba4k:
Can you test if that behavior is the same on `main`? if so I'd just open a separate discussion and link your previous comment to that one

kRHYME7:
Ah yeah behavior  in main. should be a separate issue.

image

alba4k:
so TLDR this can probably be merged now

edit: see https://github.com/hyprwm/Hyprland/discussions/14989

kRHYME7:
btw is it normal that the created HEADLESS don't render anything?  


I am viewing this thru vnc ( I kinda remember I have tried using my phone as vnc viewer for my 2nd monitor)

image


---

Might be separate issue though.  I will try to use my hyprlang config and see if I can repro there. 


- Tried it on my phone and same visuals as this, tried with differen viewer the issue is consistent. 
- But I can interact, I can lock the screen, or launch  wlogout using binds (no graphics) and press enter and it locks my laptop (did it on my phone). therefore it accepts input. 
- Last Year as my last year of being a student, I always used this feature to have my second monitor on my phone. 

The workflow was 
- create headless 
- expose to vnc using wayvnc 
- Use rvnc to view it on my phone.

alba4k:
huh, you're right.

can repro as follows:
- `hyprctl output create headless TEST`
- `wf-recorder -o TEST -f test.mp4`

`wf-recorder` fails to get any frames from the headless output (works fine on `eDP-1`), the process stops responding

can't test on main as creating headless outputs still crashes hl there lol. worked last time I tried tho, on this branch too

@vaxerski should™️ be the last issue

vaxerski:
do they render if you open something on them?

kRHYME7:
> do they render if you open something on them?

nothing is rendered at all. 

While in VNC I can trigger `ctrl + alt + delete` (which in normal ouput triggers wlogout )and press enter it locks my session. 

wlogout defaults to the first item which is to lock my session 
image

- it can receive input
- a  layer can think it is a real ouput. 
- but no frames being sent 


---

in terms of using a screencapture only

image

* Edit: show no frames from the TEST output

vaxerski:
crazy how you got a rust app to segfault lmao

vaxerski:
@alba4k @kRHYME7 sharing fixed in aquamarine.

alba4k:
@vaxerski it's just cpp being so bad that it's contagious

alba4k:
anyway can confirm recording works now

kRHYME7:
image


@alba4k  @vaxerski  works now. Thank you! Mua! 


edit: used obs for reference, wf-recorder now receives frames. LGTM

kRHYME7:
@alba4k  What is the purpose of  `hyprctl output create wayland`. It just errors. Is this not relevant for testing? 

image

alba4k:
that's only avaiable if you're already running hyprland nested inside another wayland session, iirc?

alba4k:
wallahi

jokes aside, just here to say that everything still works on my end

vaxerski:
ye I'm trying to get the bitch ass tests passing

vaxerski:
at last

leiserfg:
So 0.55.4? :pray:

UniversePM:
my monitor seems to sleep at the DRM level, which crashes hyprland. will this update fix that, especially after this merged?

vaxerski:
only one way to know

vaxerski:
> So 0.55.4? 🙏

fuck no this is way too big for a patch release

alba4k:
early 0.56 🥺

###### /s

PR #14578, opened 2026-05-15
Input/KeybindManager : Don't process keybindings which are binded inside a submap

codenameone-akshat:
Issue:
- The master keybind for submap to disable and enable the keybinds doesnt work. It is defined in the docs here: (https://wiki.hypr.land/Configuring/Basics/Binds/#disabling-keybinds-with-one-master-keybind)

Changes:
- Ignore processing keybinds if the current keybind's submap is different that the submap at keypress event as the global state changes whilst processing the keybinds.

Context:
- After debugging the code a bit, `handleKeybinds` in the `KeybindManager` calls the lua dispatcher for all the matching keybindings in the list for the initial keybind that triggered the submap, and once more for the same keybind inside the keybindings list. This is seen in the `setSubmap` in `ConfigActions.cpp`
- TL;DR:
- Behavior before the change: `press combo -> setSubmap("gameMode") -> setSubmap("reset")`
- Behavior after:  `press combo -> setSubmap("gameMode")` | `press combo -> setSubmap("reset")`

Test:
- Ran CTests
- Tested with the following config in hyprlandd.lua which was not working before:-
```
hl.bind(mainMod .. " + SHIFT + G", hl.dsp.submap("gameMode"))
hl.define_submap("gameMode", function()
    hl.bind(mainMod .. " + SHIFT + G", hl.dsp.submap("reset"))
end)
```
Risks:
- I have ran the tests and tested with a few submaps and configurations I had, but I'm not fully aware of the advanced use cases of the submaps and not sure if this might break some other things.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -688,9 +688,8 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
         // If the keybind was found before and belongs to a different submap
         // Do not trigger it again since the same keybind might exist in the submap definition
         // with different action.
-        if (found && (key.submapAtPress.name != k->submap.name)) {
+        if (found && (key.submapAtPress.name != k->submap.name))
             continue;
-        }
 
         if (!pressed) {
             // Require mods to be matching when the key was first pressed.
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -601,7 +601,7 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
         if (!k->locked && g_pSessionLockManager->isSessionLocked())
             continue;
 
-        if (!IGNORECONDITIONS && ((modmask != k->modmask && !k->ignoreMods) || (k->submap.name != Config::Actions::state()->m_currentSubmap && !k->submapUniversal) || k->shadowed))
+        if (!IGNORECONDITIONS && ((modmask != k->modmask && !k->ignoreMods) || ((k->submap.name != key.submapAtPress.name) && !k->submapUniversal) || k->shadowed))
             continue;
 
         if (device) {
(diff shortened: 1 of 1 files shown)


wabarTM:
Fixes https://github.com/hyprwm/Hyprland/discussions/14211

src/managers/KeybindManager.cpp
```diff
@@ -685,6 +685,13 @@ SDispatchResult CKeybindManager::handleKeybinds(const uint32_t modmask, const SP
             continue;
         }
 
+        // If the keybind was found before and belongs to a different submap
+        // Do not trigger it again since the same keybind might exist in the submap definition
+        // with different action.
+        if (found && (key.submapAtPress.name != k->submap.name)) {
```
vaxerski: no {}

vaxerski (commented):
tests needed

src/managers/KeybindManager.cpp
codenameone-akshat: Done :)

src/managers/KeybindManager.cpp
kolayne: Maybe it should be handled here?

The bug seems to be that we compare against `Config::Actions::state()->m_currentSubmap`, which may change during execution of the key binding. Presumably, we should rather compare against `key.submapAtPress`.

codenameone-akshat: I'm pretty sure I checked that but some other thing in that if check was masking the issue, but regardless its a good point and I'll recheck tonight with some more tests.

wabarTM: also this can be kinda related https://github.com/hyprwm/Hyprland/discussions/14664

kolayne: Looks like another bug: #14664 does not have anything to do with submaps.

codenameone-akshat: @kolayne yeah you are correct. It does work when the check is here with `|| (k->submap.name != key.submapAtPress.name && !k->submapUniversal) ||` but I'm a bit concerned about the check for the same keybind being found before (the `if (found)` check). I don't know if this will behave well with nested submaps.

codenameone-akshat: Hmmm the FD from aquamarine comes for every key event, i guess it doesnt matter.
I'll change the code to live in that condition there.
Thanks for flagging the issue! :D

vaxerski:
superseded by #14856

PR #14582, opened 2026-05-15
config/lua: add get_loaded_plugins

vaxerski:
Adds `hl.get_loaded_plugins()` for querying plugin state.

Can't do via `hl.plugin.name...` because plugins define their namespaces.


alba4k:
Couldn't test with more than 0 plugins enabled, but seems like a good approach :)

PR #14584, opened 2026-05-15
Fix shaders

fxzzi:

i used gpt 5.3 for correcting the tonemapper but mr gippity did it right i think. Instead of producing a green / blue tinted image it produces one which is tinted gray / darkened slightly which is what we expect. 
image

no tonemapping VS fixed tonemapper VS broken tonemapper
image
image
image

the blur does actually work now, but it looks blown out. not looked into it but at least it actually functions now

@UjinT34 can take a look and give a LGTM if it's correct


umbrageodotus:
maybe the blur being blown out is due to the hard coded saturation now that tonemapping works?

src/render/shaders/glsl/tonemap.glsl
```diff
@@ -4,13 +4,17 @@
 #endif
 #include "constants.h"
 
-const mat3 BT2020toLMS = mat3(0.3592, 0.6976, -0.0358, -0.1922, 1.1004, 0.0755, 0.0070, 0.0749, 0.8434);
+const mat3 BT2020toLMS = transpose(mat3( //
```
UjinT34: I'd expect a shader compiler to transpose at compile time but it seems not to be the case. Better transpose it by hand.

src/render/shaders/glsl/tonemap.glsl
```diff
@@ -50,8 +54,8 @@ vec4 tonemap(vec4 color, mat3 dstXYZ, float maxLuminance, float dstMaxLuminance,
     float E         = pow(clamp(ICtCp[0], 0.0, 1.0), PQ_INV_M2);
     float luminance = pow((max(E - PQ_C1, 0.0)) / (PQ_C2 - PQ_C3 * E), PQ_INV_M1) * HDR_MAX_LUMINANCE;
 
-    float luminanceRatio = max(luminance / dstRefLuminance, 0.0);
-    float srcScale       = maxLuminance / dstRefLuminance;
+    float luminanceRatio = max(luminance / srcRefLuminance, 0.0);
```
UjinT34: not sure about that change, kwin uses dst for those if I understand their shader correctly

umbrageodotus: using dst twice seems broken, no?

umbrageodotus: basically, it makes it a generalized curve with different src and dst luminances compared to the singular one found in kwin

UjinT34: just need to check Reinhart tonemapping original formulas. the basic is just `color.rgb = color.rgb / (color.rgb + vec3(1.0))`. modified adds an extra white point var. `color.rgb = color.rgb * (vec3(1.0) + color.rgb / (l * l)) / (color.rgb + vec3(1.0))`
the full one requires calculating real average frame luminance or using a value from the proto or guessing it from other luminance settings.
the version we use differs both from those original formulas and kwin version. and that kind of stuff hard to judge by the eye because the simplest first formula can provide good enough results and the best formula will be very expensive to compute.

umbrageodotus: orrr hear me out
we use my version which is identical to kwin's other than some adaptations and we don't deal with this bs

fxzzi: > not sure about that change, kwin uses dst for those if I understand their shader correctly

couldn't see any change in visuals so reverted it to as previous

fxzzi:
> maybe the blur being blown out is due to the hard coded saturation now that tonemapping works?

not sure, but for more context the same blown out look is visible with `hl.config({ render = {use_shader_blur_blend = 1} })`

UjinT34:
lgtm

vaxerski (approved):

PR #14594, opened 2026-05-16
layout/scrolling:  Event Loop Scheduled Prop Refresh, and changing any workspace rules causing scrolling view to move when it shouldn't.

erstarr:

- Dispatches/events/actions that are to cause scrolling view shift should do so during their own execution, and not relegate the `setOffset()` call to be executed as a result of  `recalculate()`, `recalculateMonitor()` calls dispatched by Scheduled Prop Refresh.

- Workspace rules being changed enqueue a monitor monitor reload, which  moves scrolling view. This is indiscriminate and moves the viewport for any workspace rule, regardless of if the viewport moving to center/fit the currently focused column is necessary or desirable after that specific workspace rule change. Creating a recalculate reason for new CSpace's being created moving the viewport solves this issue.
   - Like prop refresh, this event is enqueued in the event loop manager and executes after the event responsible for setting the workspace rule so catching it requires hacks, and isn't reliable.

When an action is enqueued, why it was enqueued is not saved; therefore any action that enqueues a Scheduled Prop call causes the scrolling viewport to move.

Nothing broken or new bugs introduced.

- [x] add test


erstarr:
Adding a test to this is kinda problematic: I can add a test that triggers an action that schedules a Prop Refresh, but this is not so much a test of "does Scheduled Prop Refresh cause scrolling viewport mode", but more of a "Does this specific dispatch cause viewport move".

I'll add a test using setting workspaceRules, which at the time of writing this causes a Prop Refresh to be schedules, but this is not an ideal way of testing this.

vaxerski:
I don't understand. It's like saying "C function must not move scrolling". doLater is just a utility function.

erstarr:
Bad explanation, worse naming; sorry. I fixed the description.

`CPropRefresher::scheduleRefresh()` is the caller. The reasons for a refresh isn't saved so all refreshes move the scrolling viewport indiscriminately.

For example, all workspace rule changes (adding/dropping a workspace rule), all dispatches that enqueue a prop refresh, ... move the viewport, regardless of what they change.


Inhibiting scroll before and after the dispatch/workspace rule change doesn't work either: the scheduled event is fired via Event Loop Manager doLater() after the dispatch event ends. The only way to catch this rn is to watch for other events that might be emmited when a refresh is in progress; such as `hl.on("monitor.layout_changed")` for workspace rule changes(which both enqueue a prop refresh and a monitor refresh; prop refresh goes first, then monitor reload so uninhibiting on the 3rd monitor.layout_changed is the solution for this specific case; but isn't always reliable), which is a hack and not reliable: several prop refreshes can end up being enqueued and which action caused which is a royal pain/not always possible to differentiate.


Let individual dispatches decide if they need to move the viewport to fit/center. Besides, inhibition is a pain but it is trivial with lua config to call a fit col after a dispatch/rule change.


The other option would be to pass the reason as a bitmap as to why an event was enqueued, but it's overkill to mess with the event loop mngr for this I think

vaxerski (approved):
sure, this lgtm

PR #14621, opened 2026-05-17
Window Rule: Use original monitorstr if not silent

lichie567:
This is a fix for the regression in the discussion I made here: https://github.com/hyprwm/Hyprland/discussions/14620

There are no problems, but personally I think the original feature should be reverted and reimplemented with `silent` being a first-class lua property, similar to how `movetoworkspacesilent` became the `follow` property when specifying a workspace.

Ready to merge


vaxerski:
> There are no problems, but personally I think the original feature should be reverted and reimplemented with silent being a first-class lua property, similar to how movetoworkspacesilent became the follow property when specifying a workspace.

that can be done, the good thing about lua is that we can do that. Feel free to make a MR

vaxerski (approved):

PR #14629, opened 2026-05-17
renderer: make blurredBG optional

zacoons:
In most cases I would think that consumers of this method just want a transparent background right?


zacoons:
oh whoops, that's merging the whole 0.55 branch

PR #14630, opened 2026-05-17
renderer: make render data blurredBG optional

zacoons:
In most cases I would think that consumers of this method just want a transparent background right?

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -2003,33 +2003,35 @@ void CHyprOpenGLImpl::renderTextureWithBlurInternal(SP<ITexture> tex, const CBox
                 blurClipRegion.intersect(protocolBlur);
         }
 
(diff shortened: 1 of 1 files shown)


UjinT34:
Making blur optional in `renderTextureWithBlurInternal` doesn't make much sense. Just use `renderTextureInternal` directly.

zacoons:
well I'm actually using 
```
    g_pHyprOpenGL->renderTexture(m_tex_tl, box_tl,
                                 {.blur = shouldBlur(), .a = a});
```
and `shouldBlur()` is returning true and it's deciding to use `renderTextureWithBlurInternal` and throwing a nullptr error because I haven't set blurredBG.

source: https://codeberg.org/zacoons/imgborders/commit/05ec047b0b3e018257c132be3863512f4f7a8982

PR #14633, opened 2026-05-17
fix: Aggregate modifier states from all keyboards on focus enter

ahmedkall:
Problem:
When a client receives keyboard focus (wl_keyboard::enter), Hyprland sends the locked modifier state (Caps Lock, Num Lock) of only the single "active" keyboard (m_keyboard). 
Many modern keyboards (like the AULA F75 that I use) register as multiple input devices. If you toggle Caps Lock on device A, then type a character on device B, device B becomes the active keyboard. When a new client takes focus such as hyprlock appearing on screen, Hyprland sends device B's modifier state, incorrectly reporting Caps Lock as OFF to the client. This results in visual desyncs (e.g., hyprlock not drawing the Caps Lock warning border) and incorrect key translation (passwords typed in lowercase despite the physical LED being on).

Solution:
Locked modifiers are system-wide hardware toggles, not momentary per-device states. If any physical keyboard has Caps Lock engaged, the system has it engaged. 
This PR aggregates (bitwise OR) the `depressed`, `latched`, and `locked` modifiers across all enabled, state-sharing keyboards before sending them to the newly focused client. It follows the exact same isolation rules (checking shareStates() and shouldIgnoreVirtualKeyboard()) that InputManager already uses when merging depressed modifiers.


vaxerski (commented):
atp shouldn't this be done for everything?

ahmedkall:
> atp shouldn't this be done for everything?

Yeah, you're right. We already aggregate held keys across all keyboards anyway, so we should definitely do the same for depressed and latched modifiers to keep the client's state synced. I'll push an update doing all of them.

vaxerski (dismissed):

vaxerski:
clang-format and tests please

vaxerski:
tests still needed

ahmedkall:
> tests still needed

Added an E2E test that spawns a dummy Wayland client binding wl_keyboard, null-focuses it, sets CapsLock (locked=2) on kb0 and NumLock (locked=16) on kb1 via the test plugin, then refocuses. The client asserts it receives locked=18, proving multiple keyboards' modifier states are merged on focus enter.

vaxerski (dismissed):
overkill as hell but I respect this dedication

vaxerski:
ah yes mr github ai this is so critical whatever will we do.

vaxerski:
yrou'e test failed :-(

vaxerski (approved):

kolayne:
@ahmedkall, hey! Recently, we changed the order in which tests are run, and it resulted in a test failure, which, I believe, has to do with the keyboard-modifiers test added in this PR. Could you help by taking a look and/or providing some context?

I observe that after execution of test `keyboardModifiersMergedOnFocus`:
https://github.com/hyprwm/Hyprland/blob/e5c721f29ae035881aef1620b2d714da97067883/hyprtester/src/tests/clients/keyboard-modifiers.cpp#L133-L159

the test `keybinds`, subtest `shortcutLongPress` fails:
https://github.com/hyprwm/Hyprland/blob/e5c721f29ae035881aef1620b2d714da97067883/hyprtester/src/tests/main/keybinds.cpp#L330-L355

Line 351 expects that `yCount` is 1 or 2, but it is found to be 0.

<hr>

Do you know what the issue could be?

I suppose it has something to do with things not getting cleaned up properly after the client terminates. I wonder, though: is it a bug in _Hyprland_ (i.e., when the client terminated, Hyprland should have cleared the effects of that client), or in the test (i.e., Hyprland should maintain the effect even after the client terminates)?

hyprtester/src/tests/clients/keyboard-modifiers.cpp
```diff
@@ -0,0 +1,159 @@
...
+    NLog::log("{}Started keyboard-modifiers client", Colors::YELLOW);
+}
+
+CClient::~CClient() {
+    std::string cmd = "exit\n";
+    write(this->writeFd.get(), cmd.c_str(), cmd.length());
+
+    kill(this->proc->pid(), SIGKILL);
+    this->proc.reset();
+}
```
kolayne, line 107: Side note: the signal is sent as soon as the exit command is written. In practice this means that the client won't have enough time to process the exit command (e.g., flush the log) and will be mercilessly killed immediately. Maybe we should handle this better, e.g., by waiting for a bit while repeatedly checking if the process successfully terminated.

This is not the cause of the bug, though. I tried adding a sleep in between, and the bug is still there.

PR #14639, opened 2026-05-17
fix(background-effect): out of bounds blur region

aurelleb:

Follow up to https://github.com/hyprwm/Hyprland/pull/13211

Forgot to watch for out of bounds `wl_region` resulting in blur not getting applied correctly to some clients.
In this case it was because [foot](https://codeberg.org/dnkl/foot/src/commit/dea10e2e48162c6ab48ca564ceeca88dea886a04/wayland.c#L2537) sets the width and height to `INT_MAX`.

This makes sure we never exceed the box's region, so no overflow.

I tested this and can confirm blur now properly works on foot (with blur=true)


fxzzi:
thanks image

edit: tested and werks4me

fxzzi (approved):

vaxerski (approved):

PR #14658, opened 2026-05-17
make resizewindow param accessible to lua

SteinGaming:
Makes `hl.dsp.window.resize()` take either nothing/nil, a number or a table to accommodate the previous `resizeparam` option (see https://github.com/hyprwm/Hyprland/discussions/14480).

Nothing.

~~EDIT: I just noticed that this PR adds the `string` include because of my handling, imma see if i can improve that~~

Ready for merging

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -1263,8 +1263,48 @@ TEST_CASE(monitorrule) {
     OK(getFromSocket("/output remove HEADLESS-3"));
 }
 
+
+
 TEST_CASE(mouseResize) {
-    OK(getFromSocket("/dispatch hl.dsp.window.resize(nil, { mouse = true })"));
+#define RESET_WINDOW() \
+      OK(getFromSocket("r/reload")); \
+      OK(getFromSocket("r/eval hl.unbind('mouse:273')")); \
+      OK(getFromSocket("/dispatch hl.dsp.window.resize({ x = 640, y = 400, window = 'class:kitty' })")); \
+      OK(getFromSocket("/dispatch hl.dsp.window.move({ x = 0, y = 0, window = 'class:kitty' })")); \
+      OK(getFromSocket("/dispatch hl.dsp.cursor.move({ x = 640, y = 400 })")); \
+      EXPECT_CONTAINS(getFromSocket("/clients"), "size: 640,400"); \
+      EXPECT_CONTAINS(getFromSocket("/clients"), "at: 0,0");
+
+    OK(getFromSocket("/eval hl.window_rule({ match = { class = 'kitty' }, float = true })"));
+    Tests::spawnKitty();
+    ASSERT(Tests::windowCount(), 1);
+    OK(getFromSocket("/dispatch hl.dsp.focus({ window = 'class:kitty' })"));
+
+    RESET_WINDOW();
+    OK(getFromSocket("r/eval hl.bind('mouse:273', hl.dsp.window.resize(), { mouse = true })"));
+    OK(getFromSocket("/eval hl.plugin.test.click(273, 1)"));
+    OK(getFromSocket("/dispatch hl.dsp.cursor.move({ x = 700, y = 200 })"));
+    std::this_thread::sleep_for(std::chrono::milliseconds(2000));
+    OK(getFromSocket("/dispatch hl.dsp.cursor.move({ x = 700, y = 200 })")); // Socket needs a delayed second dispatch for the first time only, for some reason?
+    OK(getFromSocket("/eval hl.plugin.test.click(273, 0)"));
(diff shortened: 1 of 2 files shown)


SteinGaming:
Changed so it uses `lua_tostring` instead of `lua_tonumber`, because it's already checked in `hlWindowResize`

vaxerski:
no that sucks, add an option to pass a table `{}` for this, whatever 0/1 was doing.

vaxerski:
`{ whatever = true }` type beat.

SteinGaming:
> no that sucks, add an option to pass a table `{}` for this, whatever 0/1 was doing.

it was previously for 0 for ignore, 1 to force aspect ratio resizing, and 2 to force ignore aspect ratio resizing.

> `{ whatever = true }` type beat.

Should it be something like `{ x, y, relative?, window?, keep_aspect_ratio? }`? That wouldn't work because it would still require x and y, and that doesn't make sense for mouse resizing. Maybe a separate table then?

SteinGaming:
Or should another window dispatcher be made for that exact purpose, e. g. `hl.dsp.window.mouse_resize({ keep_aspect_ratio: bool? }?)`

vaxerski:
we've already solved this with other dispatchers, man. If you have `whatever`, you assume its { whatever }, otherwise the other thing.

vaxerski:
bitch ass github

image

SteinGaming:
> we've already solved this with other dispatchers, man. If you have whatever, you assume its { whatever }, otherwise the other thing.

My bad, didn't look at the others for reference 🥲, should be fine now. Dispatcher now either accepts nothing (use the `keep_aspect_ratio` prop), `{ x, y, relative?, window? }` or `{ keep_aspect_ratio }`

src/config/lua/bindings/LuaBindingsDispatchers.cpp
```diff
@@ -638,7 +638,9 @@ static int dsp_mouseResize(lua_State* L) {
     if (g_pKeybindManager->m_currentKeybind)
         g_pKeybindManager->m_currentKeybind->releasePending = true;
 
-    return Internal::checkResult(L, CA::mouse("resizewindow"));
+    auto keepAspectRatio = lua_tostring(L, lua_upvalueindex(1));
```
vaxerski: wait for https://github.com/hyprwm/Hyprland/pull/14695 to be merged then rebase and use Check::string

SteinGaming: Done, I added error checking for consistency, but it really isn't needed because the calling function guarantees a number

vaxerski (dismissed):
looks good. Just some tests, please.

SteinGaming:
> looks good. Just some tests, please.

Finally got hyprtester to work lmfao, this time with a 1050 running nouveau instead of nvidia-open (I might add a disclaimer to the wiki that NVIDIA official/propitiatory are prone to not work at all).

The tests are very barebone, but just verify that the syntax stays the same. I don't know how we could actually verify what kind of resize method is used, as it requires a bind and a button input to keep it active.

vaxerski:
yeah you can simulate buttons / mice with the plugin. Check the plugin source there are ways to move the mouse and issue keys etc.

SteinGaming:
> yeah you can simulate buttons / mice with the plugin. Check the plugin source there are ways to move the mouse and issue keys etc.

yeah i might've been lazy previously lol, done

hyprtester/src/tests/main/window.cpp
```diff
@@ -1262,3 +1262,49 @@ TEST_CASE(monitorrule) {
...
+    OK(getFromSocket("/eval hl.window_rule({ match = { class = 'kitty' }, float = true })"));
+    Tests::spawnKitty();
+    ASSERT(Tests::windowCount(), 1);
+    OK(getFromSocket("/dispatch hl.dsp.focus({ window = 'class:kitty' })"));
+
+    RESET_WINDOW();
+    OK(getFromSocket("r/eval hl.bind('mouse:273', hl.dsp.window.resize(), { mouse = true })"));
+    OK(getFromSocket("/eval hl.plugin.test.click(273, 1)"));
+    OK(getFromSocket("/dispatch hl.dsp.cursor.move({ x = 700, y = 200 })"));
+    std::this_thread::sleep_for(std::chrono::milliseconds(2000));
```
vaxerski: don't sleep in tests please. Consider syncing via setting a flag, check keyboard tests

SteinGaming: done!

vaxerski (dismissed):
tnx

vaxerski:
oops clang-format needed

SteinGaming:
Ah fuck, gonna take a bit till I'm home... No way I forgot that 😭😭😭

SteinGaming:
ok fixed now

vaxerski (approved):

PR #14672, opened 2026-05-18
screenshare: only copy fb for pending frames

imperishableSecret:
- This changes monitor copy-FB activation for screencopy sessions so the compositor only keeps/copies the unmodified framebuffer when a monitor or region screencopy frame is actually pending.
- Previously, `CMonitor::needsACopyFB()` treated any active monitor/region screenshare session as requiring a copy framebuffer. That could keep the extra copy path enabled even when no client frame was waiting to be copied. The new logic tracks output copy-FB state through `CScreenshareManager` and bases the decision on pending shared frames instead.

This reduces unnecessary render/copy work during idle screenshare sessions while preserving the copy path when a screencopy frame needs data.

Ready

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -2573,7 +2573,7 @@ void CMonitorState::applyCustomModeWithSwapchain(const SP<Aquamarine::SOutputMod
 }
 
 bool CMonitor::needsACopyFB() {
-    return !m_mirrors.empty() || Screenshare::mgr()->isOutputBeingSSd(m_self.lock());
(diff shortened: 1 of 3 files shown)


vaxerski (approved):

tannerellen:
This commit introduces a bug causing screen recordings and screenshare to flash black squares at higher refresh rates (144hz). I'll create a discussion but thought I would mention it here too since this commit is pretty new.

Link to discussion: https://github.com/hyprwm/Hyprland/discussions/14745

vaxerski:
looks pretty bad, will revert

PR #14693, opened 2026-05-19
config/lua: allow get_monitors with a table for specifying args

vaxerski:
adds { all } to get all monitor objects including disabled ones

Should get a wiki MR

fixes https://github.com/hyprwm/Hyprland/discussions/14645

@takofaru @lukasgeis please test this I am at an airport

### DIFF
--- a/meta/generateLuaStubs.py
+++ b/meta/generateLuaStubs.py
@@ -541,7 +541,7 @@ def generate_stub(root: Path) -> str:
         "hl.get_workspace": "fun(selector: HL.WorkspaceSelector): HL.Workspace|nil",
         "hl.get_active_workspace": "fun(monitor?: HL.MonitorSelector): HL.Workspace|nil",
         "hl.get_active_special_workspace": "fun(monitor?: HL.MonitorSelector): HL.Workspace|nil",
-        "hl.get_monitors": "fun(): HL.Monitor[]",
+        "hl.get_monitors": "fun(options?: HL.GetMonitorsOptions): HL.Monitor[]",
         "hl.get_monitor": "fun(selector: HL.MonitorSelector): HL.Monitor|nil",
         "hl.get_active_monitor": "fun(): HL.Monitor|nil",
         "hl.get_monitor_at": "fun(x: number|HL.Vec2, y?: number): HL.Monitor|nil",
@@ -685,6 +685,16 @@ def generate_stub(root: Path) -> str:
             [
                 ("timeout", "integer", False),
                 ("type", '"repeat"|"oneshot"', False),
+            ],
+        )
+    )
+    lines.append("")
+
+    lines.extend(
+        emit_class_block(
+            "HL.GetMonitorsOptions",
+            [
+                ("all", "boolean", True),
             ],
         )
     )
--- a/src/config/lua/bindings/LuaBindingsQuery.cpp
+++ b/src/config/lua/bindings/LuaBindingsQuery.cpp
@@ -12,6 +12,8 @@
 #include "../../../desktop/view/Window.hpp"
 #include "../../../managers/input/InputManager.hpp"
 
+#include <hyprutils/utils/ScopeGuard.hpp>
+
 using namespace Config;
 using namespace Config::Lua;
 using namespace Config::Lua::Bindings;
@@ -198,12 +200,32 @@ static int hlGetActiveSpecialWorkspace(lua_State* L) {
 }
 
 static int hlGetMonitors(lua_State* L) {
+
+    bool allMonitors = false;
+
+    if (!lua_isnoneornil(L, lua_upvalueindex(1))) {
+        if (!lua_istable(L, lua_upvalueindex(1)))
+            return Internal::configError(L, "get_monitors: expected nothing or a table { all }");
+
+        {
+            lua_getfield(L, lua_upvalueindex(1), "all");
+            Hyprutils::Utils::CScopeGuard x([L] { lua_pop(L, 1); });
+
+            if (lua_isboolean(L, -1))
+                allMonitors = lua_toboolean(L, -1);
+            else
+                return Internal::configError(L, "get_monitors: table field \"all\" must be a bool");
+        }
+    }
+
     lua_newtable(L);
     int i = 1;
-    for (const auto& mon : g_pCompositor->m_monitors) {
+
+    for (const auto& mon : allMonitors ? g_pCompositor->m_realMonitors : g_pCompositor->m_monitors) {
         Objects::CLuaMonitor::push(L, mon);
         lua_rawseti(L, -2, i++);
     }
+
     return 1;
 }
 
--- a/src/config/lua/bindings/LuaBindingsQuery.cpp
+++ b/src/config/lua/bindings/LuaBindingsQuery.cpp
@@ -16,6 +16,8 @@
 #include "../../../state/MonitorState.hpp"
 #include "../../../state/WorkspaceState.hpp"
 
+#include <hyprutils/utils/ScopeGuard.hpp>
+
 using namespace Config;
 using namespace Config::Lua;
 using namespace Config::Lua::Bindings;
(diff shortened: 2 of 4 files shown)


lukasgeis:
This does not seem to work for me unfortunately. I built hyprland from source (and rebooted):
```
Hyprland 0.55.0 built from branch main at commit 4deab89fab64f7fd1d92b77fe4f5ec5f57591b77 clean (config/lua: allow get_monitors with a table for specifying args).
Date: Tue May 19 19:12:34 2026
Tag: v0.55.0-53-g4deab89fa, commits: 7343

Libraries:
Aquamarine: built against 0.11.0, system has 0.11.0

Version ABI string: 4deab89fab64f7fd1d92b77fe4f5ec5f57591b77_aq_0.11_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```
But even, then the following only prints `#Monitors: 1` into the rollinglog despite 2 monitors being connected (that are *both* found by `hyprctl monitors all`; one is enabled, the other (internal display eDP-1) disabled):
```
hyprctl eval 'print("#Monitors: " .. tostring(#hl.get_monitors({ all = true, })))'
```

I hope, I am not misunderstanding the API here (or did something wrong in the building process...). 

Nonetheless, even if the monitors would get returned here, they still do not have a `disabled`-field? Although, one could check against `hl.get_monitors()` to filter out disabled ones.

takofaru:
Currently I'm using a headless virtual monitor called "test" to test this. For some reason, it only detects when the config is reloaded. This is the config I made for the test

```
hl.monitor({
	output = "eDP-1",
	mode = "1920x1080",
	position = "0x0",
	scale = 1
})

hl.config({
	input = {
		kb_layout = "us",
		follow_mouse = 1,
		sensitivity = 0,

		touchpad = {
			natural_scroll = true,
		}
	}
})

local mainMod = "SUPER"

hl.bind(mainMod .. " + escape", hl.dsp.exit())
hl.bind(mainMod .. " + Q", hl.dsp.exec_cmd("kitty"))
hl.bind(mainMod .. " + C", hl.dsp.window.close())
hl.bind(mainMod .. " + R", hl.dsp.exec_cmd("rofi -show drun"))
hl.bind(mainMod .. "+ SHIFT + S", hl.dsp.exec_cmd('grim -g "$(slurp -d)" - | wl-copy'))

local all_monitors = hl.get_monitors({ all = true })
local monitors = hl.get_monitors({ all = false })

hl.bind(mainMod .. " + T", function ()
    if all_monitors then
		for i, monitor in ipairs(all_monitors) do
			hl.notification.create({ text = '[Test All] Monitor ' .. i .. ' ' .. monitor.name .. ', is_mirror: ' .. tostring(monitor.is_mirror) or 'Unknown', timeout = 15000 })
    	end
	else
    	print("[Test] get_monitors returned nil or failed!")
	end
end)


hl.bind(mainMod .. "+ SHIFT + T", function ()
    if monitors then
		for i, monitor in ipairs(monitors) do
			hl.notification.create({ text = '[Test] Monitor ' .. i .. ' ' .. monitor.name .. ', is_mirror: ' .. tostring(monitor.is_mirror) or 'Unknown', timeout = 15000 })
    	end
	else
    	print("[Test] get_monitors returned nil or failed!")
	end
end)

hl.config({
	debug = {
		disable_logs = false
	}
})
```

I don't know if the config is hot-reloaded or not. But reloading it and press SUPER + T outputs the monitor when it's not in disabled or mirrored. (Sadly I can't give the proof with screenshot because I'm using the notification to check).

Using `hyprctl eval "hl.monitor({output = 'test', disabled = false, mirror = 'eDP-1'})"`, pressing SUPER + T outputs the "test" monitor that it's mirrored. But, pressing SUPER + SHIFT + T, which does not use the { all } still outputs the test monitor.

Using `hyprctl eval "hl.monitor({output = 'test', disabled = true})"`, the result are the same as when the monitor is mirrored.

For note, the test is continous without reloading every check, except for the first time when the config is reloaded but the virtual monitor is not has been created.

Here's the build information:
```
Hyprland 0.55.0 built from branch lua-monitors-all at commit 4c8413b43ab4f64c662cd9c7b687c404a5777b23 clean (config/lua: allow get_monitors with a table for specifying args).
Date: Tue May 19 22:33:41 2026
Tag: , commits: 7343

Libraries:
Aquamarine: built against 0.11.0, system has 0.11.0

Version ABI string: 4c8413b43ab4f64c662cd9c7b687c404a5777b23_aq_0.11_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
flags set:
debug
```
I'm probably wrong on some parts, as I'm not an expert on this. Feel free to correct me ;)

KobKobson:
Tried to pick commit 4c8413b43ab4f64c662cd9c7b687c404a5777b23 on top of main and couldn't compile
```
Hyprland/src/config/lua/bindings/LuaBindingsQuery.cpp: In function »int hlGetMonitors(lua_State*)«:
Hyprland/src/config/lua/bindings/LuaBindingsQuery.cpp:228:57: error: »class CCompositor« hat kein Element namens »m_realMonitors«
  228 |     for (const auto& mon : allMonitors ? g_pCompositor->m_realMonitors : g_pCompositor->m_monitors) {
      |                                                         ^~~~~~~~~~~~~~
Hyprland/src/config/lua/bindings/LuaBindingsQuery.cpp:228:89: error: »class CCompositor« hat kein Element namens »m_monitors«
  228 |     for (const auto& mon : allMonitors ? g_pCompositor->m_realMonitors : g_pCompositor->m_monitors) {
      |                                                                                         ^~~~~~~~~~
```

Thanks for the effort :)

KobKobson:
Sorry for the noise, fixed the compilation error by replacing the line with:
```cpp
for (const auto& mon : allMonitors ? State::monitorState()->allMonitors() : State::monitorState()->monitors()) {
```
I also only get monitors that are not disabled when setting `all`.

pbmarcano:
**This PR would fix a session-killing failure mode for MST monitors — motivation plus two review notes.**

Setup: LG UltraFine 5K over DisplayPort MST, Hyprland 0.56.2, Lua config. The panel presents as two monitors that are byte-identical in make/model/serial/description; only the mode list distinguishes them (the live tile offers 3840x2160, the dead one only 2560x2880), and port names follow the physical port. The dead tile is a real `wl_output` that displays nothing and takes a workspace, so my config detects it via `hl.get_monitors()` and folds it with `mirror = <primary>`.

The problem: on every `hyprctl reload`, monitor rules are cleared and the config re-runs — but `hl.get_monitors()` filters out mirrored/disabled monitors, so the config cannot see the folded tile and cannot re-register its mirror rule. `ensureMonitorStatus()` then finds the live mirror mismatching the catch-all rule, un-mirrors the tile (creating a `wl_output`), and the config's re-fold on `config.reloaded` destroys that output again. That destroy triggers the registry bind race from #12996: clients die with `wl_registry: global wl_output (N) is unavailable`, and XWayland dies and is never respawned. Since my distro's theme switcher runs `hyprctl reload`, every theme switch could kill the session.

The reload reconciler skips monitors whose live state already matches their rule (`COMPARISON_FULL_MATCH`), so being able to *see* the mirrored tile at load time and re-register the identical rule makes reload completely safe. I verified this with an out-of-VM workaround — persisting the fold decision to a file in `$XDG_RUNTIME_DIR` and re-registering it at every config load: multiple reloads including theme switches, the tile never left mirror state, zero client deaths. `get_monitors({ all = true })` would let configs do this properly, without the state file.

Two review notes on the current diff:

1. The argument is read from `lua_upvalueindex(1)` rather than stack index 1. Unless `setFn` passes caller arguments as upvalues, `hl.get_monitors({ all = true })` never sees the table (other query functions, e.g. `hlGetMonitor`, read from stack index 1) — which may be why testing stalled.
2. The branch predates the `State::monitorState()` refactor (`g_pCompositor->m_realMonitors` no longer exists on main), so it likely needs a rebase.

Happy to test a rebased build against the MST setup above.

vaxerski:
@lukasgeis can we get a recheck

lukasgeis:
Thanks! It seems to work fine now!

I played a bit around with `hl.get_monitors()` and found no mismatches compared to the output of `hyprctl monitors [all]`. Field values were consistent as well as the table of the output itself. Mirroring/Disabling posed no problems at all.

My setup was practically identical to the previous test, although some of the AUR `hypr*-git` packages were out of date, so I used the regular packages from `extra` this time.

Since the stubs and fields of `HL.Monitor` are already updated as well, I don't have any feature requests or missing pieces.

Thanks again!

alba4k:
fyi hypr*-git packages should never be out of date, even if the pkgver shown on the aur/in the pkgbuild seems to be old. they'll always just install the latest commit. it's up to you/your aur helper to manage updates

if you notice them being broken or something like that please report that at https://github.com/alba4k/hyprwm-pkgbuilds

lukasgeis:
Sorry for the confusion @alba4k. You're right! I had problems yesterday with `hyprutils-git` but apparently this was a quirk of my package manager. The machine I tested it on wasn't updated for a long time as I mainly use it for testing builds.

Since both updating packages as well as building Hyprland from source failed with errors from `hyprutils-git`, I assumed this to be the cause of it, as the metadata in the AUR still referred to version 0.13.1. Nonetheless, installing it now, correctly installs the newest version with no errors, so no broken PKGBUILD there.

PR #14705, opened 2026-05-20
fullscreen: Refactor fullscreen logic from the ground up; Fullscreen Controller and Fullscreen Handlers, better handle defaut and layout handled FS, add ability to default FS windows even on layout handled FS workspaces, fix many bugs related to fullscreen

erstarr:

This PR seeks to: 

- [x] Refactor the fullscreen logic from the ground up
   - [x] Add a FS controller for central management of FS-unFS actions. Singleton
   - [x] Add FullscreenHandler s - one default and one for each layout that wishes to own FS handling for their workspaces. Tied to each layout object -> one per workspace so workspaces can independently keep information about their windows and their fullscreen states.

Note: Layouts use their Fullscreen Handlers for their 'unique' fullscreen management. e.g. with scrolling you can scroll onto a FS window. When you do this, it doesn't call the controller but calls its Fullscreen Handler; the logic for managing multiple FS windows in a workspace is unique to scrolling and the logic/attributes/methods should be isolated to scrolling specific classes 

Note:  Handlers keep a list of FS windows in their 'area of responsibility (workspace)'.`CWindow` and `CWorkspace` no longer have their `m_fullscreenState` and `m_hasFullscreenWindow` attributes: these values returned to lua are obtained from the FS handlers that handle those windows/workspaces

<br>

- Renames/Removals/Misc. Behaviour Changes
  - [x] rename `m_createdOverFullscreen` --> `m_allowedOverFullscreen` to be more descriptive of its function.
  - [x] Change `m_allowedOverFullscreen` logic: all windows when they are not hidden "behind" a FS window have `m_allowedOverFullscreen = true`. Only when they are hidden 'behind' a FS window do they have this attr `= false` (this works for floating windows but is iffy for tiled - all but current FS window is always hidden anyway)
  - [x] Remove "effective mode" logic in fullscreen - remove `FSMODE_MAX` (and `mode = 3` along with it) as a user-settable option.

- Further Reading For Some Changes

- FSMODE_MAX is no longer a user-facing mode, but is now a pure internal implementation detail. If a FS is requested by the client to be FSMODE_FULLSCREEN-ed while it is FSMODE_MAXIMIZED, it is considered in a special state (I called it FSMODE_MAX cuz the name already exists but it can be called something else.). It no longer has a special mode assigned to it.

This removes the confusion about what FSMODE_MAX is; and since it had no user-facing documentation about how it worked and why it existed, as well as the fact that it was really just FSMODE_FULLSCREEN as far as the user was concerned: it's moved to be a pure implementation detail now.

Requested values are now clamped to hi: FSMODE_FULLSCREEN so if users used `mode = 3`, it'll silently fall back to using ` = 2`.

<br>

- Fix a lot of bugs with scrolling FS behaviour:
  - [x] Existing Bug: Fullscreen/Unfullscreening a floating window causes that window to be unfocusable, although it remains visible
  - [x] Existing Bug: When fullscreen/unfullscreening a floating window while there is already a fullscreened tiled window, the floating windows that are supposed to be hidden behind the floating window are revealed. Not focusable but visible.
  - [x] Existing Bug: Fullscreening/maximising a window already maximised/fullscreening causes the original size of the window to be lost (the colsize of the maximised/fullscreened window is saved and the original is lost)
  - [x] Scroll Fs doesn't work with all dispatches that query fullscreen state, doesn't work with workspace rules with `f[0/1/2]
     - Test with border_size
  - [x] Regression: https://github.com/hyprwm/Hyprland/discussions/15206
  - [x] FSing grouped windows - groupbar hiding when Fullscreen

- Fixes for default handled FS
  - [x] Existing Bug: Tiled FS window -> float it -> unFSes. inconsistent with Floating FS window -> tile it -> still FS
  - [x]  The hl.dsp.window.fullscreen_state({ internal = 2, client = -1, action = "toggle" }) dispatcher should be toggling but it behaves as if action = "set".
  - [x] Fix scrolling maximise hyprtest - fix window size setting -> fix the test also

- Misc improvements
  -  [x] When a window which is being FSed is expelled from a col when there's >1 windows in that col, make `expel` of that window always act the same as `promote` layoutmsg dispatch. It makes more sense for the expelled target to go behind its prev. col, not before it

- New Features
  - [x] Ability to choose to use default FS behaviour when FSing a window in layout owned FS behaviour workspace by passing the param in fullscreen dispatch from 
  - [x] Add a `fullscreenHandler`  to hyprctl and `fullscreen_handler` attr to lua window class - as string.
  - [x] Add `pinFullscreened` to hyprctl and lua window class 

<br><br>

### Framework

<details>

<summary> Legacy - before the decision to refactor FS logic complately </summary>

#### Definitions

FS -> A window that may be Either: `FSMODE_FULLSCREEN` or `FSMODE_MAXIMIZED`

fullscreen window -> A window that `FSMODE_FULLSCREEN`

maximised window -> A window that is `FSMODE_MAXIMIZED`

covering FS/fullscreen/maximised window -> A window that is both FS, and is considered by the layout as the "currently in complete view fullscreen window".
--> The default and scrolling's handling of these windows are the same: If a window completely covers the whole monitor (fullscreen) or the whole work area (maximise), but different layouts may define this in different ways.

Maximize <-> Maximise -- Although we all know which one is the correct way of saying it :wink: 

#### FS Windows

Expose the following:

- internal/client fullscreen modes: the fullscreen mode of the window.
   - This is set regardless of its handler (default or layout)
   - This is not enough to judge that a window is a covering FS.

- m_target of the window has a `bool m_layoutManagedFullscreen` flag.
   - This must be set `true` if window is layout managed, `false` otherwise. This is used outside the FS related functions to differentiate between default and layout handled fullscreens. 

- fullscreen handler enum.
   - This is to be used to differentiate between different fullscreen handlers when/if there are more layouts that handle their own fullscreens.
      - For example: in scrolling, changing focus to another window when there's a covering FS should not cause fullscreen to cycle (depending on the config value; switch to that window after unFS, unFs current and FS the focused window, or simply unFS the current window); let's say another layout's FS behaviour would like this behaviour. We cannot use the `m_layoutManagedFullscreen` then, and need to carve out scrolling handled FS windows as the sole exception.
   - This can also be used to create window rules targeting only specific layout handled FS windows: I want scrolling handled FS windows to keep their borders, but don't want the default handled ones to have borders. (without this, scrolling FS windows that don't cover still get their borders stripped with a window rule that only targets fullscreen windows - which is what someone may want, or may not want)

- `isFullscreen()` method.
   -  This returns true IFF the window is a covering FS!
      - Note that if 2 FS windows overlap (scrolling tiled and floating example), this returns true for both; as both are indeed covering FS windows!
         - In this case, use workspace's FS method to get the FS window ontop.

#### Workspace

Have 2 fullscreen related attributes/methods

- `m_hasFullscreenWindow` - This is true if there is a covering FS window in the workspace

- `getFullscreenWindow()` - Returns the covering FS window of the workspace. 
   - It is possible for 2 covering fullscreen windows to overlap. In scrolling: a tiled layout managed covering FS window and a floating (default handled) covering FS window may exist at the same time. In this case, this method returns the covering floating FS window.
   
   
   
#### Monitor

- `inFullscreenMode()` - returns `true` if there's a covering FS window whose internal mode is `FSMODE_FULLSCREEN` (i.e. there's a fullscreen window)
-  `getFullscreenWindow()` - calls workspace method of the same name; same return val.

#### What FS behaviour a layout may own

A layout may own:
- How to fullscreen a window (`requestFullscreen()`)
- The definition of a FS window (consequence of the layout owning `requestFullscreen()` and `layoutFullscreenTarget()`)
- How to handle window hiding upon FS (`setNoMembersAboveFullscreen()`)
- What defines a covering FS window (`layoutFullscreenTarget()`)

A layout must override both `requestFullscreen()` and `layoutFullscreenTarget()` if it wished to own fullscreen behaviour, but may choose to not override `setNoMembersAboveFullscreen()` if it doesn't need window hiding behaviour different than the default.

Needless to say, layouts can define whatever helper functions they need internally to handle the above. All of that is isolated to the layout's own source/header files and won't  be used (or ideally available) outside of it.

Layouts handle each of these with minimal mixing of default handled fullscreen related code. As it is, some code related to default handled FS windows is inevitable as floating windows are always default handled; and in a future PR I'll make it possible to have a default handled tiled FS window in workspaces that have layouts that own FS semantics.

#### Default Handled FS

- The old FS behaviour - unchanged. The size and position of the window are set, its FS related window properties are set, and that's it.

All layouts that don't define their own FS behaviour will use the default handled FS behaviour.

In a future PR, i'll add the ability to choose if the user wants to use default or layout FS behaviour in workspaces whose layout does define custom FS behaviour.

##### Note for default handled FS with master and monocle layouts

Both of these set a window's size and pos in recalculate() calls. therefore, there is code in both their recalculate() calls to correctly set or skip re-setting window pos-size if that window is fullscreen.

#### Layout handled FS

- `layoutFullscreenTarget()` must return the target that is FS and is covering. Even if there is a covering floating FS window ontop, it must still return the covering FS window.
   -  This is used in `isFullscreen()` of `CWindow` to judge if a window is fullscreen if that window is layout handled.
  

- `requestFullscreen()` defines how to FS/un-FS a window when the corresponding dispatch or internal call to set window's internal FS state is fired.

- `setNoMembersAboveFullscreen()` can be optionally overridden if a layout needs custom window hiding logic (e.g. floating windows when there's an underlying covering layout managed tiled FS window in scrolling layout).

- `recalculate()` - scrolling uses this to keep the shapes and sizes of non-covering (or focused) FS windows. I reckon other layouts that may own FS semantics in the future will also almost always make use of this. 

<br><br><br>

</details>

Complete refactor of FS and some related things. Please read the description for the full list of changes and what they imply.

#### Known Bugs

- All bugs from "Fix a lot of bugs with scrolling FS behaviour:" - duplicated here for ease of viewing
  - [x] Existing Bug: Fullscreen/Unfullscreening a floating window causes that window to be unfocusable, although it remains visible
  - [x] Existing Bug: When fullscreen/unfullscreening a floating window while there is already a fullscreened tiled window, the floating windows that are supposed to be hidden behind the floating window are revealed. Not focusable but visible.
  - [x] Existing Bug: Fullscreening/maximising a window already maximised/fullscreening causes the original size of the window to be lost (the colsize of the maximised/fullscreened window is saved and the original is lost)
  - [x] Scroll Fs doesn't work with all dispatches that query fullscreen state, doesn't work with workspace rules with `f[0/1/2]`
  - [x] Regression: https://github.com/hyprwm/Hyprland/discussions/15206
  - [x] FSing grouped windows - groupbar hiding when Fullscreen
  
<br><br>

MR Related Bug Tracking:

- [x] Scrolling - group FS doesn't work
- [x] new window creation (other than kitty) -> invisible until it's switch to or fullscreened
- [x] make a group with at least two windows -> fullscreen it -> change focus between group members -> unfullscreen it -> there are not borders until you change focus to another group member. (both for default and layout handling - default fails to resize too)
- [x] Default Handled fullscreen - grouped windows -> switch btw windows -> windows size is not kept at fullscreen
- [x] Window rules broken
- [x] Workspace Rules Broken (maybe?)
- [x] The hl.dsp.window.fullscreen_state({ internal = 2, client = -1, action = "toggle" }) dispatcher should be toggling but it behaves as if action = "set".
- [x] Default Handled FS -> Fullscreen window -> create floation -> fullscreen/unfullscreen floating -> floating is lost behind FS window
- [x] Floating grouped windwos -> FS -> Tile without unFSing -> unFS => window/workspace rules keep applying
- [x] in non-master layouts maximise position/size setting is broken
- [x] fade animations are broken
- [x] master -> `on_focus_under_fullscreen = 0` -> maximise master, open a new tiled window. broken.

<br><br>

#### TODO

- [x] Case: FS a video in a browser that's maximised and unFS the video - use vid player's own controls. Browser should probably remain maximised.

Misc:

<details>
- [ ] When moving windows around workspaces, windows should be properly unFSed  before move and re-FSed after move. This should be the current behaviour -> make sure that it is reliable

- [ ] Replace all bitmap compare operators with == / !=  (????)

- [ ] Don't forget null checks where appropriate

- [ ] Test legacy dispatchers

- [ ] Check that the viewport doesn't move when default FSed on scrolling

- [ ] Additional virtual layouts may implement: `syncFullscreen()` - if necessary. may not need this

- [ ] Go over the framework part and properly document the current layout fullscreen logic

- [ ] Clear up includes and forward decelerations

- [ ] Refactor the Tiled (no need iirc), Master, monocle recalculate to better handle not `updatePos()` - ing a window that's fullscreened.

- [ ] Clear comments made during PR.
   - [ ] Clear out all commented code, artifacts and notes taken during the PR

</details>

Wiki MR TODO

- [x] Users can choose to default handled a window even in a workspace with layout fullscreen behaviour (scrolling) {layout_aware = bool}
- [x] FSMODE_MAX doesn't exist as a mode (in the enum) anymore -- it is now a state that can only be requested by clients and only sets internal value if internal and client are synced. Neither client not internal can be manually set to FSMODE_MAX; but client can be considered as FSMODE_MAX if it meets the conditions (FSMODE_MAXIMISED -> FSMODE_FULLSREEN)
- [x] Window rules -> f[1] matches covering FS windows. fullscreen = true matches non-covering as well.


erstarr:
Will change how `m_createdOverFullscreen` works a little so it makes more sense.  Needed to make the floating window hiding with scrolling work properly

1 - All newly created windows start with `m_createdOverFullscreen = true`. If it is `false`, it is because it is currently hidden under a FS window; otherwise it is true.

erstarr:
In Scrolling Layout: when a window that is the last/only window in a column is floating and the layout scrolls to a FS window, that window is left ontop of the tiled FS window. I think this is a good idea, as it's unlikely that the user would want to lose that window they just floated.

erstarr:
It's almost done. Only some edge case bug fixes, polish, and cleanup remains unless I also add in this PR the ability to default FS windows in workspaces with layout managed FS's available.

Just about everything else works now, so please help test!

erstarr:
I'm marking this as ready for now. Mainly to hopefully get more ppl testing and have vax have a look over the code if he approves of the general changes.

Still need to do some comment cleanup, need to look into some hyprtests, and play with windowrules a bit to see if all is good; but the broad strokes should be gtg.

Hopefully more people help test if it's marked as ready too.

vaxerski:
I'll tag auto review, I'll need a shot of something strong to even review this, fs states are notoriously finnicky. A lot of fs tests would be useful...

vaxerski:
@codex review please

src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
```diff
@@ -521,12 +537,47 @@ void SScrollingData::recalculate(bool forceInstant) {
...
+
+    // Every time m_hasFullscreenWindow is true, that means that an FS window is currently taking up the entire monitor/work area.
+    // We use that as the flag for determining if Direct Scanout should be enablad or not
+
+    // send a scanout tranche if we are entering fullscreen, and send a regular one if we aren't.
+    // ignore if DS is disabled.
+    static auto PDIRECTSCANOUT = CConfigValue<Config::INTEGER>("render:direct_scanout");
+
+    if (!CURRENT_FS_TDATA || !CURRENT_FS_TDATA->target->window())
+        return;
```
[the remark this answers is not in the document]
erstarr: Neat trick. Thanks bot.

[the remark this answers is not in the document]
erstarr: Fixed. Hopefully this covers all possible cases: dealing with this on scrolling is much more of a pain because of the possibility of non-covering FS windows than in default handled.

erstarr:
I'll work on adding some more tests after I iron out some stuff (2 hyprtests, going over comments). Still, the broad strokes should be hopefully gtg.

I tried to be generous with the comments because some parts of the code are pretty convoluted thanks to scrolling being scrolling, edge cases such as overlapping fullscreens, handling default and layout managed fullscreens without mixing code from one in the handlers of another, etc...

vaxerski:
maybe just refactor the logic from the ground up to be more clean? Although I am unsure how we'd do this. FullscreenController? idk

vaxerski:
I think a controller would make more sense. Isolate the fulslcrenn logic into one place.

erstarr:
I'll update the framework writeup i have in description to describe how things work with this PR in detail

erstarr:
> I'll update the framework writeup i have in description to describe how things work with this PR in detail

@vaxerski done. Please read that if you want a quick overview of what the current (as of PR) state of FS logic is.
Externally (outside FS functions being refactored and rewritten to integrate layout FS-s), there has been minimal change.

The relatively complex logic in for example scrolling window hiding is inevitable because layering 2 FS windows is fated to get messy, but in any case these things are always isolated to the layout's own source file and doesn't leak to outside.


### As for a controller:

Agreed it's tidier. Can make one.

I'm thinking:

- Tie it to layouts (UP<>). One interface (actually abstract in this case) class which includes virtual FS related methods with their "default handled" implementations, and one derived class per layout that wishes to own FS behaviour. Or maybe one derived class for default too and the interface class can be pure virtual. I'd lean towards the first option but maybe 2nd is cleaner...


- The interface would at least have the 3 FS related virtual methods (in `IModeAlgorithm` ). I think it's best to handle getting FS states/windows from window, workspace, monitor's methods.

- Move the 3 FS functions in CCompositor into controller class, make them static, let them dispatch to layout specific functions like now.
---> their implementations would not change much if at all

- window, workspace, monitor (this just calls workspace methods anyway) will call the controller's methods to get info about the fullscreenState of windows.
-> There will be minimal change here: the core functions all of these use are `CWorkspace::getFullscreenWindow()` and `CWindow::getFullscreen()`.
--> In their implementations, one basically calls the other and `getFullscreen()` either simply checks window's attributes to see if it's FS or calls the layout owned `layoutFullscreenTarget()` in case of layout owned FS. All that would change is the enclosing scope of the method.

- in scrolling's controller for example would be the 3 virtual fn overrides + all the FS related helper methods in `CScrollingAlgorithm`.
--> This would be the main gain of making a controller. Fullscreen pipeline outside of layout managed fullscreens is remarkably clean already.


- Recalculate would still be the core of scrolling's FS code. Can try and separate FS related code from scrollingData's `recalculate()` but not 100% sure how successful that'd be; should be doable.

Misc. Notes:

- Exceptions in say focus direction still need to exist: say `CFullscreenController::isWindowFullscreen()` is layout specific; it'd still return true if a window is indeed fullscreen and we'd still need to write `if (window->layoutManaged)` exceptions if a certain layout must not follow that behaviour.

- only 2 methods that are used outside the fullscreen pipeline have layout specific exceptions/code: `CWorkspace::getFullscreenWindow()` and `CWindow::isFullscreen`. The benefits of the cleanup would be more or less limited to not having much FS code in `IModeAlgorithm`, `CCompositor`, `CScrollingAlgorithm`; with the most significant gain being `CScrollingAlgorithm`; and moving them more or less as they are to the controller class. "if window/workspace FS" checks all across the codebase inevitable unfortunately. At least unifying functions makes them easier to find.


<br><br>

any suggestions or do-don't s ontop of the above?

vaxerski:
sounds good honestly. This should also make it much more testable. Layout should be able to control the fullscreen state via a specific implementation (strategy pattern) here for a default one that applies to everything but scrolling and a scrolling specific one. 

I think the controller should still hold fs state, unfs / refs and stuff like that, the layout strategy could expose something like `isFullscreenTaken` so that we can replace fs if needed (e.g. fullscreening another window)

Avoid making static functions, singletons are generally how we do state. `UP<Controller>& controller()` and dispatch via that.

erstarr:
Rewrote the description. I'll get working on 'fullscreens rethonk' now; might take a while as time's tight nowadays :(

src/managers/fullscreen/FullscreenController.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+
+
+
+        private:
+
+        // fullscreenTarget() -> dispatch FS request to FS handler
+
+        // Tracks FSed windows (internal OR client). For custom layout targets, keep track of them in their own FS handlers
+        std::unordered_map<WP<Desktop::View::CWindow>, SWindowFullscreenState> m_fullscreenWindows;
```
vaxerski: could get inefficient at scale. Consider adding a helper like `CAttachable` which CWindow can inherit, then you can add attachments extending `IAttachment` and do e.g. `window->getAttachment<ATTACHMENT_TYPE_SOMETHING>()` like we do in aq with buffers.

erstarr: Would that not go counter to the goal of centralising all FS logic into controller and handlers?

If we are moving the FS state to CWindow, I think it'd be logical to do similar for CWorkspace, as many parts of FS state checks involve checking if workspace has covering fullscreen. Then; I'd argue that the original goal of unification would be too diluted.

I think the perf cost would be syncFullscreen() needed to clear expired weak pointers: O(n). Lookups and placements are all O(1) otherwise.  Memory cost should be negligible; merely pointer and a struct.


Realistically, how many fullscreen windows will be had? -> only one per all non-scrolling workspaces and possible multiple for scrolling.

The performance concern would mainly be scrolling I think; In that case, I can move FS window states into handlers: All queries for if a window is FS would be handled by FS handlers. In this case, controller would realistically be only useful for keeping `setWindowFullscreen...()` functions originally in CCompositor.
In this case, the perf cost of `syncFullscreen()` would be very small as it would only handle the windows of the workspace which had a FS state change or otherwise necessitated a sync. I think this is very tidy, and very performant assuming people don't create thousands of scrolling FS windows all in one workspace (in that case some other part of the code would probably give out first).

If need be, I'd also keep a list of all FS handlers in controller (= number of workspaces); but i can't think of a case I'd need this.

In any case, you know the project best and if you still think attachment pattern is better here; I'll defer to your judgement. Thanks again for the help.

vaxerski: if you can keep the lookups fast then no need for an attachment. My concern was mostly performance.

Also, I want to make a quite large refactor to windows some time soon once I get my pc back and finish moving, but it shouldnt hinder this that much.

erstarr:
Decoupling FS related code from `SScrollingData::recalculate()` ends up being extremely messy; need to pass over more than half a dozen parameters.

Leaving the layoutbox setting code as-is for now. Might be better addressed as a part of a scrolling layout refactor

erstarr:
@vaxerski Controller, Handlers, ability to default handle windows even in scrolling are done. What's left is to replace legacy calls to `isFullscreen()`, `getFullscreenWindow()`, ... with calls to controller.

Anything that you disagree with or would recommend architecturally?

vaxerski:
I'll take a look at this once my other refactors are (mostly) done if that's fine by you. See https://github.com/hyprwm/Hyprland/pull/15260

I am trying to essentially get CCompositor to be as empty as possible because currently it's a piece of shit god class that does too much

(in general from a quick glance looks ok tho)

vaxerski:
once https://github.com/hyprwm/Hyprland/pull/15260 is merged we can take a look at this as it will be logical to move this one next (I'll make one more MR for stuff but it wont impact this majorly)

erstarr:
It is preliminarily functional.

Help with testing would be appreciated; FS states can have weird edge cases so more known bugs earlier the better.
Especially since I mainly use scrolling; bug reports and repo steps for other layouts would be appreciated

Please consult `Known Bugs` part in the description to check that your bug isn't documented before reporting. Thanks in advance.

vaxerski:
okay after https://github.com/hyprwm/Hyprland/pull/15282 I'll pause my refactors of CCompositor so we can get your thing merged then continue

erstarr:
> okay after #15282 I'll pause my refactors of CCompositor so we can get your thing merged then continue

You can keep refactoring if you want - all i need to do to make it work with this for the most part is change the FS calls.

This still has some work left because FS states are extremely convoluted and I had the bright idea to add so many messy things at one.

I'm making separate MRs for some misc. fixes this MR had so more refactors will have limited effect on this one.

Aqa-Ib:
small regression. Steps to reproduce:
1. make a group with at least two windows.
2. fullscreen it.
3. change focus between group members.
4. unfullscreen it.
5. there are not borders until you change focus to another group member.

Cheers.

erstarr:
And there's still some stuff that can be added onto this ontop of what is already there.
Like: Theoretically it should be easy to add custom FS calls that users can define with lua like custom lua layouts with the new framework.

Also I plan to write quite a lot of test so it'll be a while still.

kolayne:
Dwindle layout.

1. If I open any window, it won't appear. I have to switch to another workspace and then come back for it to become visible.
2. The window rule
    ```
    hl.window_rule({
      name = "maximized-workspace",
      match = { workspace = "f[1]" },
    
      border_color = "rgba(ff80ccff)",
    })
    ````
    should make the window border pink if it's on a workspace where there is a fullscreen window. But it behaves incorrectly:
    - when I maximize the window, its border color stays the same
    - when I open a tiling window over a maximized window, it's pink (correct), then I close it and the maximized window also becomes pink (correct, should have already been pink), but then I unmaximize it, and other windows on the same workspace become pink (incorrect, should have become regular because there no longer is a maximized window on the workspace)
3. The `hl.dsp.window.fullscreen_state({ internal = 2, client = -1, action = "toggle" })` dispatcher should be toggling but it behaves as if `action = "set"`.

I don't know if any of these is config-specific. My hyprland config can be found here: https://github.com/kolayne/how-to-Arch/tree/master/_.config/hypr

vaxerski:
> I'm making separate MRs for some misc. fixes this MR had so more refactors will have limited effect on this one.

much welcome. I know this is a messy refactor but let's try to stick to general clean code practices and we'll at least get something a bit better

erstarr:
> small regression. Steps to reproduce:
> 
>     1. make a group with at least two windows.
> 
>     2. fullscreen it.
> 
>     3. change focus between group members.
> 
>     4. unfullscreen it.
> 
>     5. there are not borders until you change focus to another group member.
> 
> 
> Cheers.

Fixed


> If I open any window, it won't appear. I have to switch to another workspace and then come back for it to become visible.

Fixed

The other 2 are added to tracked bugs. Will be fixed when going over their respective sections.

Thank you.

erstarr:
> Dwindle layout.
> (...)
>     2. The window rule
>        ```
>        hl.window_rule({
>          name = "maximized-workspace",
>          match = { workspace = "f[1]" },
>        
>          border_color = "rgba(ff80ccff)",
>        })
>        ```
>          
>        should make the window border pink if it's on a workspace where there is a fullscreen window. But it behaves incorrectly:
>        
>        * when I maximize the window, its border color stays the same
>        * when I open a tiling window over a maximized window, it's pink (correct), then I close it and the maximized window also becomes pink (correct, should have already been pink), but then I unmaximize it, and other windows on the same workspace become pink (incorrect, should have become regular because there no longer is a maximized window on the workspace)
> (...)

Fixed.

If i understood the steps correctly: Max a window -> create floating window ontop -> close the floating window (focus falls back to tiled Max) -> close the max window.

Thank you.

<br>


As a sidenote: scrolling FS position setting shouldn't be handled in `SScrollingData::recalculate()` but decoupling that would break a lot of things so I think it should be done in another PR after this one establishes a working base to compare to.

kolayne:
Dwindle, chromium with a youtube page. If I go fullscreen, only the client goes to fullscreen, the window remains in its regular state. If I then do `hl.dsp.window.fullscreen_state({ internal = 0, client = 0, action = "set" })`, nothing happens (should have exited fullscreen). But if I focus any other window, client exits fullscreen.

vaxerski:
@codex review please

erstarr:
> Dwindle, chromium with a youtube page. If I go fullscreen, only the client goes to fullscreen, the window remains in its regular state. If I then do `hl.dsp.window.fullscreen_state({ internal = 0, client = 0, action = "set" })`, nothing happens (should have exited fullscreen). But if I focus any other window, client exits fullscreen.

I assume you used youtube's own fullscreen button/keybind for this.

Fixed.

p.s. I'm going to re-add the `MAXIMISE | FULLSCREEN` dual mode after all. content type turns out to be rather useless as a parameter :)

Thank you.

erstarr:
@Dregu Implemented the framework for asking for specific FS handling other than default/layout.

Can now specify which type of handler (`FULLSCREEN_HANDLER_DEFAULT_MONITOR`, `FULLSCREEN_HANDLER_SCROLLING_WORKSPACE` for example). They can be specified as a sub-handler to the algo's main layout handler or another handler to exist alongside the layout handler in an algorithm.

Controller will take care of figuring out which handler to use for the request and dispatching accordingly.

To add new FS behaviour, just add the new handler's name/type to `eFullscreenHandler`, create its handler, add logic to correctly correctly grab it when requested in the controller (after the first handler that's not layout-specific is added, this should be unnecessary for others).

<br>

it should now be unnecessary to add new types of FS modes: use `FSMODE_FULLSCREEN` and `FSMODE_MAXIMIZED` with the desired handler.

Handlers still need to manage `FSMODE_MAX` though, which is really undesirable but I still don't have a good solution for that.

Maybe keep a list of `MAX` windows in the controller since the `FSMODE_MAX` behaviour applies to all handlers? not sure...

erstarr:
@Haku-taku I can't repo your bug. Can you check if it's fixed?

Haku-taku:
> @Haku-taku I can't repo your bug. Can you check if it's fixed?

https://github.com/user-attachments/assets/035279f8-3e70-4814-8266-c58654259911

I've built and tested it. Offset can still be observed after exiting FS. the bug is not fixed yet.

erstarr:
> I've built and tested it. Offset can still be observed after exiting FS. the bug is not fixed yet.

Fixed.

Aqa-Ib:
bug:
1. make a floating group.
2. fullscreen it.
3. un-fullscreen no longer works.

edit: it really unfullscreens it but makes it have the monitor size instead of its size pre-fullscreen.

Aqa-Ib:
On another topic:
Fullscreening a group is incoherent both in main branch and in this branch.
When we fullscreen a group, all of its members are expected to have the same position and size reported by hyprctl clients. But this is not the case, producing resize glitches when changing focus between group members.

Also you told me on discord: what about the size when moving a member out of the group while at fullscreen? IMO, it is supposed to also be fullscreened when moved out, because groups members, visually in practice, are just one single window, so they are expected to preserve that same window size across every interaction with them.

Thoughts?

erstarr:
> bug:
> 
>     1. make a floating group.
> 
>     2. fullscreen it.
> 
>     3. un-fullscreen no longer works.
> 
> 
> edit: it really unfullscreens it but makes it have the monitor size instead of its size pre-fullscreen.

Fixed

Thank you

erstarr:
> On another topic: Fullscreening a group is incoherent both in main branch and in this branch. When we fullscreen a group, all of its members are expected to have the same position and size reported by hyprctl clients. But this is not the case, producing resize glitches when changing focus between group members.
> 
> Also you told me on discord: what about the size when moving a member out of the group while at fullscreen? IMO, it is supposed to also be fullscreened when moved out, because groups members, visually in practice, are just one single window, so they are expected to preserve that same window size across every interaction with them.
> 
> Thoughts?

imo the handling of targets that belong to a group could use some improvement. ITarget methods should auto-detect that the target is in a window group and dispatch the request to the groupTarget instead of incorrectly handling it themselves - gets rid of a good amount of footguns.

It's gotta be in another PR tho.

src/managers/fullscreen/FullscreenController.hpp
```diff
@@ -0,0 +1,199 @@
...
+        /* FSMODE_MAX */
+
+        // List of FSMODE_MAX windows
+        std::unordered_set<WP<Desktop::View::CWindow>> m_fsModeMaxWindows;
+    };
+
+
+  }
+  
+inline UP<Fullscreen::CFullscreenController> g_pfullscreenController = makeUnique<Fullscreen::CFullscreenController>();
```
vaxerski: can you move this to the new-styled no-global-ns `UP<CFullscreenController>& controller();`?

vaxerski:
also format the code as you go it makews it easier to reasd

erstarr:
Implemented some error detection and correction logic in controller. In testing state mishandling hasn't happened with any regularity but more redundancy doesn't hurt.

Ideally this should be implemented in each handler but that's significant complexity that's best handled in its own MR (any maybe not feesible due to that complexity). The only case that current error handling doesn't cover is calling the handler directly, which i'm on the fence about anyway

erstarr:
Removed FSMODE_MAX as a user-facing option altogether; it's now pure internal implementation detail.

There was no documentation about what it did; and what exited said "it's basically FSMODE_FULLSCREEN".

Modes now clamp to FSMODE_FULLSCREEN as upper bound.

I made it so that this is not a breaking change: 3 will silently act like 2 - the same as if you passed mode = 1000 => mode = 2.

For all intents and purposes I intend to make it not user-settable but I CAN make it 'viewable' (see if a window is in FSMODE_MAX state) - with its new implementation anyway (add back the enum var, FS mode getter will search if the target is in the list of 'FSMODE_MAX windows' and return 'FSMODE_MAX' which will show as '3' to the user) if people want.

vaxerski:
is this ready for review? I still see a bunch of TODOs in here

erstarr:
> is this ready for review? I still see a bunch of TODOs in here

Yeah I put a bunch of todos for features that are disabled for scrolling FS handling since they react very messily with scrolling FS behaviour and don't make sense to me (can use scrolling layout messages to achieve feature parity, etc...), but could be feasibly implemented if someone really wanted to. Should probably remove the todo there yeah.

Also one todo for adding fsed pinned window attr to lua and hyprctl output, which is a 2 line change I'll make soon.

Other than some comment messiness (cosmetic only) I didn't have time to resolve yesterday, it's good to review.

src/managers/fullscreen/FullscreenController.hpp
```diff
@@ -0,0 +1,206 @@
...
+      FSMODE_MAX : When a client request FSMODE_FULLSCREEN and the current window's internal mode is FSMODE_MAXIMISED, it becomes FSMODE_MAX.
+
+
+      This is not an official FS mode, but handling of a special case:
+        e.g. maximised browser window -> fullscreen and unfullscreen window -> browser window should still be maximised
+
+
+      Effectively, it is treated as a Fullscreen window in all aspects but one: If this window is un-FSed, it will return to being maximised instead of un-FSed
+      
+      Handlers don't track this as an official mode
```
vaxerski: comments like this make this unreadable. You put way too much comments. Most functions are self-describing. Only comment if a reasonable dev can't predict something.

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -0,0 +1,592 @@
...
+    if (!monitor || (!monitor->m_activeSpecialWorkspace && !monitor->m_activeWorkspace))
+        return nullptr;
+
+    PHLWORKSPACE activeWorkspace = nullptr;
+
+    // Check special workspace first since it renders on top of regular workspaces
+    if (const auto ACTIVE_SPECIAL_WORKSPACE = monitor->m_activeSpecialWorkspace; ACTIVE_SPECIAL_WORKSPACE) {
+        activeWorkspace = ACTIVE_SPECIAL_WORKSPACE;
+
+    } else if (const auto ACTIVE_WORKSPACE = monitor->m_activeWorkspace; ACTIVE_WORKSPACE) {
```
vaxerski: go through the MR and fixup those {}, also the newline above here is confusing

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -0,0 +1,592 @@
...
+        activeWorkspace = ACTIVE_WORKSPACE;
+    }
```
vaxerski: this thing repeats like 10 times, helper fn?

src/managers/fullscreen/handler/FullscreenHandler.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include "../../../output/Monitor.hpp"
+#include "../../../layout/target/WindowGroupTarget.hpp"
+#include <hyprutils/memory/Casts.hpp>
+
+using namespace Fullscreen;
+
+IFullscreenHandler::IFullscreenHandler(Layout::IModeAlgorithm* const algorithm) : m_algorithm(algorithm) {
+    if (!m_algorithm) {
+        Log::logger->log(Log::CRIT, "IFullscreenHandler failed during construction: Owning layout algorithm does not exist!");
+        throw std::runtime_error("CScrollingFullscreenHandler: bad algorithm type");
```
vaxerski: please don't use exceptions

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
```diff
@@ -0,0 +1,712 @@
...
+    else if (m_fullscreenWindowHidingState.lastTiledLayoutManagedFsWindow && !TARGET_WORKSPACE_HAS_FS) {
+        const auto LAST_FS_WINDOW = m_fullscreenWindowHidingState.lastTiledLayoutManagedFsWindow.lock();
+        LAST_FS_WINDOW->m_ruleApplicator->propertiesChanged(Desktop::Rule::RULE_PROP_FULLSCREEN | Desktop::Rule::RULE_PROP_FULLSCREENSTATE_CLIENT |
+                                                            Desktop::Rule::RULE_PROP_FULLSCREENSTATE_INTERNAL | Desktop::Rule::RULE_PROP_ON_WORKSPACE);
+        LAST_FS_WINDOW->updateDecorationValues();
+
+        Config::Supplementary::refresher()->scheduleRefresh(Config::Supplementary::REFRESH_WINDOW_STATES | Config::Supplementary::REFRESH_MONITOR_STATES);
+    }
+
+    /* Setting DSO and VRR */
```
vaxerski: what is DSO?

vaxerski:
@codex review too please

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
erstarr: direct scanout - Changed to DS

vaxerski:
req my review once ready (also I see a conflic)

erstarr:
@vaxerski any more problems or just waiting for a good time to merge?

vaxerski:
oh that didnt show up for some reason ok github

vaxerski:
@codex review please

vaxerski:
also can you rebase on main so we have a linear history, I wanna make sure we don't break shit and I think I pushed a thing that might need a 2 line refactor with this to the fullscreen:: stuff

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -0,0 +1,555 @@
...
+        return;
+
+    const bool WANT_SYNC = window->m_ruleApplicator->syncFullscreen().valueOrDefault();
+
+    bool       stateChanged = false;
+
+    if (internal.has_value())
+        internal = std::clamp(internal.value(), sc<eFullscreenMode>(0), FSMODE_FULLSCREEN);
+    if (client.has_value())
+        client = std::clamp(client.value(), sc<eFullscreenMode>(0), FSMODE_FULLSCREEN);
```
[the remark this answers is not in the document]
erstarr: It's explicitly noted that `#mode = 3` now falls back to 2. 3 is no longer a user settable option

Noted in the wiki (that it's no longer a user settable option, 3 removed from FS possible values). prev it was 'fullscreen as far as the user was concerned' per the wiki and now it's literally fullscreen as far as the user is concerned.

erstarr:
done

vaxerski:
@codex review please

erstarr:
done

vaxerski:
@codex review please

src/config/lua/bindings/LuaBindingsDispatchers.cpp
```diff
@@ -496,65 +497,68 @@ static int dsp_floatWindow(lua_State* L) {
...
+    const auto CURRENT        = Fullscreen::controller()->getFullscreenModes(target);
     const bool atDesiredState = CURRENT.internal == desiredInternal && CURRENT.client == desiredClient;
 
-    if (actionRaw == 0) {
-        return Internal::checkResult(L, CA::fullscreenWindow(desiredInternal, desiredClient, maybeW));
-    }
+    if (actionRaw == 0)
+        return Internal::checkResult(L,
+                                     CA::fullscreenWindow(CURRENT.internal == desiredInternal ? Fullscreen::FSMODE_NONE : desiredInternal,
+                                                          CURRENT.client == desiredClient ? Fullscreen::FSMODE_NONE : desiredClient, layoutAware, maybeW));
```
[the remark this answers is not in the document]
erstarr: this is intended. If you are separating modes, it's reasonable to assume that actions will apply to each individually.

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -0,0 +1,553 @@
...
+    const auto TO_BE_USED_FS_HANDLER = getFsHandler(window, layoutAware.value_or(WAS_LAYOUT_HANDLED));
+    if (!TO_BE_USED_FS_HANDLER) {
+        Log::logger->log(Log::ERR, "window {} doesn't have FS handler assinged. This should never happen", window->m_title);
+        return;
+    }
+
+    // handles FSMODE_MAX case. If nothing to handle on that front, saves the provided old values
+    const auto saveClientInternalValues = [&](const SFullscreenMode& OLD_FS_MODES) {
+        if (client.value_or(FSMODE_NONE) == FSMODE_FULLSCREEN && OLD_FS_MODES.internal == FSMODE_MAXIMIZED)
+            m_fsModeMaxWindows.emplace(window);
```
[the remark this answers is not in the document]
erstarr: FSMODE_MAX is not a mode anymore. It's a special case.

When window is in FSMODE_MAX state, it's because of the client's request. Client also is allowed to dispel this state.

It was tested with youtube and some other video services: unFSing with video controls also sends internal state but it may not be the case always. A client-caused state should be able to be dispelled by the client

erstarr:
Reviewed. ~~Didn't find any bugs this time.~~ Really useful tho, thanks for the runs.

Actually wait i have an idea for the first one let me try to handle that. The other two are not bugs

erstarr:
Done.

I really want to avoid slapping a `setGlobalPos()` in `recalculate()` if i can help it.

vaxerski:
@codex review please

vaxerski:
> really want to avoid slapping a setGlobalPos() in recalculate() if i can help it.

yeah, but you could've just added some rechec() but oh well this works ig

erstarr:
done

vaxerski:
@codex review please

vaxerski:
shame it cant do more at once

src/config/shared/actions/ConfigActions.cpp
```diff
@@ -267,35 +269,44 @@ ActionResult Actions::pinWindow(eTogglableAction action, std::optional<PHLWINDOW
...
+ActionResult Actions::fullscreenWindow(Fullscreen::eFullscreenMode internalMode, Fullscreen::eFullscreenMode clientMode, bool layoutAware, std::optional<PHLWINDOW> w) {
     auto window = xtract(w);
     if (!window)
         return {};
 
+    const bool KEEP_INTERNAL = (internalMode == -1);
+    const bool KEEP_CLIENT   = (clientMode == -1);
+
+    internalMode = std::clamp(internalMode, sc<Fullscreen::eFullscreenMode>(0), Fullscreen::FSMODE_FULLSCREEN);
+    clientMode   = std::clamp(clientMode, sc<Fullscreen::eFullscreenMode>(0), Fullscreen::FSMODE_FULLSCREEN);
```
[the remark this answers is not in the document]
erstarr: https://github.com/hyprwm/Hyprland/pull/14705#discussion_r3561223103

erstarr:
Repeated (https://github.com/hyprwm/Hyprland/pull/14705#discussion_r3560615872). Guess it's a good sign that it only gave this


I'm unsatisfied with pos setting with changing work area so I'll redo that. Otherwise, if the oracle gives its blessings, merge at your discretion.

edit: done.

p.s. `visible` is very finicky and it's hard to test the visibility of a window with it. Used `visible`, `hidden`, `acceptInput` in combination to compensate. Will look into it but it doesn't need to hold this MR up since the other 2 confirm proper behaviour.

vaxerski (approved):
Image

PR #14727, opened 2026-05-21
config/lua: allow any callable object for dispatch/actions

klapptnot:
## Allow callable values (`__call`) as dispatchers in `hl.bind` and `hl.dispatch`

Previously, `hl.bind` only accepted the following:
- Lua functions
- native dispatcher userdata (`hl.dsp.*`)

This changes dispatcher resolution to also accept any value with a `__call` metamethod. The reason behind is to make a declarative way to run commands, avoiding wrapping them in dispatchers or helper functions

```lua
-- From only
local function bash (s) return function () hl.exec_cmd (s) end end
hl.bind ("XF86AudioPrev", bash ("playerctl previous"))
hl.bind ("XF86AudioNext", hl.dsp.exec_cmd ("playerctl next"))

-- now also allow this
getmetatable ("").__call = hl.exec_cmd
hl.bind ("META+L", "dms ipc call lock lock")
```

This also works with `hyprctl`

```bash
hyprctl dispatch "'www-browser'" # somewhat pointless for my example
```

There's not much behavior change on this change; aside of maybe changing Lua stubs, this was not implemented on gestures, as gestures use string configurations. 

## [Breaking] Drop string actions in favor of runtime values

Previously `hl.gesture(spec)` action accepted:
- Lua functions
- Strings, named actions

This changes from strings to runtime values in a callable table `hl.gesture`

```lua
getmetatable ("").__call = hl.exec_cmd
local gestures = {
  {
    fingers = 3,
    direction = "horizontal",
    action = hl.gesture.WORKSPACE, -- previous "workspace"
  },
  {
    fingers = 3,
    direction = "vertical",
    action = "notify send 'Hi from Lua' 'I just swiped on my trackpad!'" -- since __call is set
  },
  {
    fingers = 2,
    direction = "vertical",
    action = function ()
      hl.notification.create ({ text = "Bruh", duration = 60000 })
    end
  }
}

for _, g in ipairs (gestures) then
  hl.gesture (g)
end
```

### Changes on generateLuaStubs.py

They only help for fields added from a loop, explicit fields

```cpp
lua_newtable(L);
//! @fields: [FOO:string, BAR:integer|string, BAZ:table]
```

Callable tables had the `---@operator call:<signature>`, added the signature after type, since LSP didn't pick it up correctly

PR is ready to merge, no known bugs

### DIFF
--- a/meta/generateLuaStubs.py
+++ b/meta/generateLuaStubs.py
@@ -13,6 +13,7 @@ from typing import Iterable
 @dataclass
 class ApiNode:
     methods: set[str] = field(default_factory=set)
+    fields: dict[str, str] = field(default_factory=dict)
     children: dict[str, "ApiNode"] = field(default_factory=dict)
 
 
@@ -64,7 +65,9 @@ def find_matching_brace(text: str, open_brace_idx: int) -> int:
     raise ValueError("Unbalanced braces while parsing C++ source")
 
 
-def extract_function_bodies(source: str, header_pattern: re.Pattern[str]) -> list[tuple[re.Match[str], str]]:
+def extract_function_bodies(
+    source: str, header_pattern: re.Pattern[str]
+) -> list[tuple[re.Match[str], str]]:
     out: list[tuple[re.Match[str], str]] = []
     for match in header_pattern.finditer(source):
         open_idx = source.find("{", match.end() - 1)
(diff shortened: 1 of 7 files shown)


vaxerski:
no breaking chnages

klapptnot:
Anything similar to `action = "workspace"`, will stop working until updated to something like `hl.gesture.WORKSPACE`.

Anyone can still inject custom values to the enum table

```lua
hl.gesture.WORKSPACE = 11 -- hmmm
hl.gesture.BRUH = 12 -- __newindex will block this
```

leiserfg:
@klapptnot  you can just do:
```lua
local c = hl.exec_cmd

hl.bind ("META+L", c"dms ipc call lock lock")
```

using `__call` is nice for doing OOP like apis, but overriding builtin's metatables is nasty.

Dregu:
> using `__call` is nice for doing OOP like apis, but overriding builtin's metatables is nasty.

Well the overriding part is strictly on the user side, but it also might be the only use case for this, so it's kinda splitting hairs if this is a useful addition or a pointless additon to golf one more character out of your config...

klapptnot:
The point of this is mostly convenience, after all

In my example of adding `hl.exec_cmd` as the `__call` handler for strings, it only adds a reference to the string metatable, and every string used afterward doesn't allocate any additional memory beyond the string itself

Whereas adding `hl.dsp.exec_cmd` or wrapping it in a function creates more GC objects in memory, either functions or dispatchers

A function to string would make `function: 0x5570c76f0730`, and a dispatcher `HL.Dispatcher`, so as now there's no API to get keybindings, you may want to get which command it is, for example, a "Key bindings" window (`{ desc = "what it does; not command" }`), for this it may be handy to use strings

Neither `hyprctl binds -j` shows what it executes aside of `"__lua"`, may be modified to have more helpful info for any shell, but it may defeat the point of even changing to Lua (to deprecate bash scripts scattered around)
By storing them as a string with call attribute, you have access to that without any more memory, or API

Using it is totally optional for the user. No string is being modified internally; instead of rejecting a value, it accepts a value that Lua already knows how to call

As for `hl.gesture.*`, that's already using `__call` through `hl.gesture()`

PR #14734, opened 2026-05-21
plugins: add api for registering and dispatching events

zacoons:
adds functions for this

### DIFF
--- a/src/config/lua/LuaEventHandler.cpp
+++ b/src/config/lua/LuaEventHandler.cpp
@@ -231,10 +231,10 @@ std::expected<void, std::string> CLuaEventHandler::addCustomEvent(SP<Event::CEve
                     case CEventBus::CCustomEvent::TYPE_INT: lua_pushinteger(L, std::get<int>(arg)); break;
                     case CEventBus::CCustomEvent::TYPE_DOUBLE: lua_pushnumber(L, std::get<double>(arg)); break;
                     case CEventBus::CCustomEvent::TYPE_STRING: lua_pushstring(L, std::get<std::string>(arg).c_str()); break;
-                    case CEventBus::CCustomEvent::TYPE_WINDOW: CLuaWindow::push(L, std::get<PHLWINDOW>(arg)); break;
-                    case CEventBus::CCustomEvent::TYPE_WORKSPACE: CLuaWorkspace::push(L, std::get<PHLWORKSPACE>(arg)); break;
-                    case CEventBus::CCustomEvent::TYPE_LAYER_SURFACE: CLuaLayerSurface::push(L, std::get<PHLLS>(arg)); break;
-                    case CEventBus::CCustomEvent::TYPE_MONITOR: CLuaMonitor::push(L, std::get<PHLMONITOR>(arg)); break;
+                    case CEventBus::CCustomEvent::TYPE_WINDOW: CLuaWindow::push(L, std::get<PHLWINDOWREF>(arg)); break;
+                    case CEventBus::CCustomEvent::TYPE_WORKSPACE: CLuaWorkspace::push(L, std::get<PHLWORKSPACEREF>(arg)); break;
+                    case CEventBus::CCustomEvent::TYPE_LAYER_SURFACE: CLuaLayerSurface::push(L, std::get<PHLLSREF>(arg)); break;
+                    case CEventBus::CCustomEvent::TYPE_MONITOR: CLuaMonitor::push(L, std::get<PHLMONITORREF>(arg)); break;
                 }
             }
         });
--- a/src/config/lua/LuaEventHandler.cpp
+++ b/src/config/lua/LuaEventHandler.cpp
@@ -8,8 +8,11 @@
 
 #include "../../event/EventBus.hpp"
 #include "../../desktop/state/FocusState.hpp"
+#include <expected>
+#include <vector>
 
 extern "C" {
+#include <lua.h>
 #include <lauxlib.h>
 }
 
@@ -18,7 +21,7 @@ extern "C" {
 using namespace Config::Lua;
 using namespace Config::Lua::Objects;
 
-void CLuaEventHandler::dispatch(const std::string& name, int nargs, const std::function<void()>& pushArgs) {
+void CLuaEventHandler::dispatch(const std::string& name, int nargs, const std::function<void(lua_State*)>& pushArgs) {
     auto it = m_callbacks.find(name);
     if (it == m_callbacks.end() || it->second.empty())
         return;
@@ -59,7 +62,7 @@ void CLuaEventHandler::dispatch(const std::string& name, int nargs, const std::f
         ++m_dispatchDepth;
 
         lua_rawgeti(m_lua, LUA_REGISTRYINDEX, sub->second.luaRef);
-        pushArgs();
+        pushArgs(m_lua);
 
         auto* mgr = CConfigManager::fromLuaState(m_lua);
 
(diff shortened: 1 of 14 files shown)


vaxerski (commented):
and why is this not going via the event bus?

zacoons:
do you mean I should add `m_pluginEvents` to the event bus or something? I'm not sure what you mean.

vaxerski:
I think you should do _something_ to not tie the events to the lua configmanager, because these events are not config-related.

Probably something on the bus, which lua config manager then uses

zacoons:
How's it look now?

I'll fix up the commit history when this is all done. Or u could just do a sqash commit or smth.

src/event/EventBus.hpp
```diff
@@ -32,6 +36,12 @@ namespace Event {
         template <typename... Args>
         using Cancellable = CSignalT<Args..., SCallbackInfo&>;
 
+        struct SPluginEvent {
+            int                                    nargs;
+            const std::function<void(lua_State*)>& pushArgs;
```
vaxerski: this needs to be lua-agnostic. You can for example pass something like vector<variant<...>> instead and have the lua listener translate that.

vaxerski (commented):
better!

src/event/EventBus.hpp
zacoons: oh yeah I forgot to delete that. not used.

src/plugins/PluginAPI.hpp
```diff
@@ -362,6 +363,28 @@ namespace HyprlandAPI {
     */
     APICALL bool removeLuaFunction(HANDLE handle, const std::string& namespace_, const std::string& name);
 
+    /*
+        Add a plugin-owned Lua event that can be used with hl.on("<name>", ...)
+
+        returns: true on success. False otherwise.
+    */
+    // APICALL bool addLuaEvent(HANDLE handle, const std::string& name);
+    APICALL bool addLuaEvent(HANDLE handle, const std::string& name, int nargs, const std::function<void(lua_State*)>& pushArgs);
```
vaxerski: this is a bad entry point because it relies on lua where it need not to. Instead:

- make a custom event class
- store a list of arg types expected (std::type_index)
- register on the eventBus
- have Lua pick it up and parse called invocations. You can pass arguments as a vector<variant<...>> and do `using` for the variant<> of allowed types - int, double, window, layer, whatever.
- eventBus can reject invocations with invalid args based on the typeinfo

```cpp
class CCustomEvent {
   public:
    using ValidVariant = std::variant<int, double, PHLWINDOWREF, ...>;
     CCustomEvent(std::vector<const std::type_index*> expected);

     void emit(std::vector<ValidVariant>&& data); // does type-checking
};
```

something like this. Then you can `APICALL bool addCustomEvent(HANDLE handle, SP<CCustomEvent> ev)`

src/plugins/PluginAPI.hpp
```diff
@@ -362,6 +363,28 @@ namespace HyprlandAPI {
...
+        returns: true on success. False otherwise.
+    */
+    APICALL bool removeLuaEvent(HANDLE handle, const std::string& name);
+
+    /*
+        Dispatch any Lua event.
+
+        returns: true on success. False otherwise.
+    */
+    APICALL bool emitLuaEvent(HANDLE handle, const std::string& name, void* data);
```
vaxerski: redundant with design above, you can `.emit()`

src/event/EventBus.hpp
```diff
@@ -32,6 +34,32 @@ namespace Event {
...
+                TYPE_INT           = 1,
+                TYPE_DOUBLE        = 2,
+                TYPE_STRING        = 3,
+                TYPE_WINDOW        = 4,
+                TYPE_WORKSPACE     = 5,
+                TYPE_LAYER_SURFACE = 6,
+                TYPE_MONITOR       = 7,
+            };
+
+            using ValidVariant = std::variant<bool, int, double, std::string, PHLWINDOW, PHLWORKSPACE, PHLLS, PHLMONITOR>;
```
vaxerski: `PHL*` need to be `PHL*REF` for non-owning refs

src/event/EventBus.cpp
```diff
@@ -1,8 +1,23 @@
...
     return p;
 }
+
+CEventBus::CCustomEvent::CCustomEvent(std::string name, std::vector<eType> argTypes) : m_name(name), m_argTypes(argTypes), m_event({}) {}
+CEventBus::CCustomEvent::~CCustomEvent() = default;
+
+std::expected<void, std::string> CEventBus::CCustomEvent::emit(const std::vector<ValidVariant>& args) {
+    for (size_t i = 0; i < args.size(); i++) {
+        const auto& arg = args[i];
+        if (arg.index() != m_argTypes[i]) {
```
vaxerski: no {}

src/plugins/PluginAPI.cpp
```diff
@@ -496,3 +497,62 @@ APICALL bool HyprlandAPI::removeLuaFunction(HANDLE handle, const std::string& na
...
+
+    using namespace Event;
+
+    if (!bus()->m_events.plugin.contains(name)) {
+        Log::logger->log(Log::ERR, "failed to unregister event {}: event not registered.", name);
+        return false;
+    }
+
+    // remove from LuaEventHandler
+    auto config_mgr = dynamicPointerCast<Config::Lua::CConfigManager>(WP<Config::IConfigManager>(Config::mgr()));
```
vaxerski: feature envy: have eventBus emit an event like customChanged and make LuaCfgMgr listen to it

zacoons: better now???

vaxerski (commented):
yup, looks nice! Just add a test for this please. Should be simple

vaxerski:
@codex review please

src/plugins/PluginAPI.cpp
```diff
@@ -496,3 +497,38 @@ APICALL bool HyprlandAPI::removeLuaFunction(HANDLE handle, const std::string& na
...
+
+    using namespace Event;
+
+    // add to EventBus
+    if (!bus()->m_events.plugin.try_emplace(event->m_name, event).second) {
+        Log::logger->log(Log::ERR, "failed to register event {}: event already registered.", event->m_name);
+        return false;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
zacoons: oh yeah I never actually emit those events lmao

src/config/lua/LuaEventHandler.cpp
```diff
@@ -200,8 +214,39 @@ void CLuaEventHandler::clearEvents() {
     m_callbacks.clear();
 }
 
-const std::unordered_set<std::string>& CLuaEventHandler::knownEvents() {
-    static const std::unordered_set<std::string> EVENTS = {
+std::expected<void, std::string> CLuaEventHandler::addCustomEvent(SP<Event::CEventBus::CCustomEvent> event) {
+    using namespace Event;
+
+    auto listener = event->m_event.listen([&](const std::vector<CEventBus::CCustomEvent::ValidVariant>& args) {
+        dispatch(event->m_name, event->m_argTypes.size(), [&](lua_State* L) {
```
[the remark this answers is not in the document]
zacoons: what?

zacoons:
do you reckon the destructor of CCustomEvent should remove it from the event bus?

zacoons:
I got AI to write the tests, but I read them over and rewrote some of them. I reckon they're lookin' pretty good.

vaxerski:
> do you reckon the destructor of CCustomEvent should remove it from the event bus?

No? THe event bus has a strong ref lmao

src/config/lua/LuaEventHandler.cpp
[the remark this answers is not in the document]
vaxerski: it is correct. You don't use any captures though so replace `[&]` with `[]`

vaxerski:
also a rebase pls

src/config/lua/LuaEventHandler.cpp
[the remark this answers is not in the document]
zacoons: I think I fixed this. Can you confirm?

zacoons:
> > do you reckon the destructor of CCustomEvent should remove it from the event bus?
> 
> No? THe event bus has a strong ref lmao

oh yeah, ofc

src/config/lua/LuaEventHandler.cpp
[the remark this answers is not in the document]
vaxerski: nah bro just remove the `[&]`s and replace with `[]`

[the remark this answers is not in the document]
zacoons: You might be looking at this outdated snippet. I can't replace with `[]` because those functions need to capture some variables. However, I've manually specified them. See line 226 and 227 https://github.com/hyprwm/Hyprland/pull/14734/changes#diff-086c520da411e00a46d2569d708007dd6e9bc36817ca018384d5b5cb0c0b7fc0R226-R227

`event->m_event.listen` needs to capture `this` in order to call `dispatch` and `event` in order to get the name. Then `[args](lua_State* L) {` needs to capture `args` for obvious reasons.

Or maybe I'm being stupid somehow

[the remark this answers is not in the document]
vaxerski: ah yeah mb

vaxerski (dismissed):
tnx

vaxerski:
oops build fails

zacoons:
> oops build fails

works on my machine.

looking at the error messages it might just need a re-run:
`error: unable to download 'https://cache.nixos.org/nar/07wvj3iw5nc3gilwqc27yiakifv9ri9irz1a20ywykvc17bikrjx.nar.xz': HTTP error 416`

vaxerski (approved):
no bone movies

PR #14743, opened 2026-05-21
keybinds: store hit binds first, then execute callbacks

vaxerski:
lua can invalidate the m_keybinds vec in callbacks

ref #14612


Abdisto:
Was able to test. Yep, this fixes it.

PR #14746, opened 2026-05-21
config/monitor: allow per-monitor vrr to inherit misc:vrr

BlueManCZ:
#### Background

Per-monitor `vrr` (set via `hl.monitor()` or a `monitor=` rule) cannot be reset back to following the global `misc:vrr`. `CMonitorRule::m_vrr` is a `std::optional<int>` whose `std::nullopt` already means "inherit `misc:vrr`", but the live `hl.monitor()` API is additive — omitting a field keeps its previous value — and the `vrr` field only accepts `0..3`. So once a per-monitor override is set, there is no value you can pass to return the monitor to the global setting. The override is sticky until the next full config reload.

#### Change

Widen the accepted input range for `vrr` to `[-1, 3]` and map an incoming `-1` to `std::nullopt` at the three input boundaries:

* the Lua `hl.monitor()` binding,
* the `monitor=NAME,...,vrr,N` parser, and
* the legacy `monitorv2` config path.

So `hl.monitor({output = "...", vrr = -1})` or `monitor=NAME,...,vrr,-1` clears the per-monitor override and the monitor follows `misc:vrr` again. `CMonitorRule::m_vrr` stays a `std::optional<int>`.

Ready.


vaxerski:
I think the approach kinda sucks? -1 is just std::nullopt, why rid of the optional<>?

BlueManCZ:
Fair. Reverted to `std::optional`, the clear is now just `nullopt` at the input boundary.

vaxerski (approved):
yep tnx

PR #14748, opened 2026-05-21
hyprctl: add config full-reload for performing a ground-up reload

vaxerski:
COPS CANT FUCKING CATCH ME


alba4k:
is `hyprctl reload full-reset` supposed to just crash hyprland?
[hyprlandCrashReport3375.txt](https://github.com/user-attachments/files/28459714/hyprlandCrashReport3375.txt)

vaxerski:
every time something crashes its you

it works for me...

alba4k:
it works the first 3-4 times or so, but crashes hl after a bit. I noticed that the reload which precedes the one that crashes seems to take a bit longer. Spamming it with a keybind also immediately crashes hyprland

in case it helps, [bt_full.txt](https://github.com/user-attachments/files/28476789/bt_full.txt)

PR #14749, opened 2026-05-21
Fix inverted scrollwheel direction in keybinds

zjeffer:
```
hl.bind("mouse_down", function ()
    hl.notification.create({text = "Mouse down", timeout = 1000})
end)
hl.bind("mouse_up", function ()
    hl.notification.create({text = "Mouse up", timeout = 1000})
end)
```
The above binds are an easy way to show the issue: when scrolling down, you'll see the "Mouse up" notification and vice versa. The if statement needs to be inverted.

Checking with `wev`, scrolling down yields: 
```
[        15:      wl_pointer] axis: time: 14953488; axis: 0 (vertical), value: 15.000000
```
so a positive delta.

This has been broken for 4 years (from its initial implementation in https://github.com/hyprwm/Hyprland/pull/265) so this might break some people's configs.

Ready. I didn't test it though but this seems pretty trivial.

### DIFF
--- a/src/managers/KeybindManager.cpp
+++ b/src/managers/KeybindManager.cpp
@@ -425,7 +425,7 @@ bool CKeybindManager::onAxisEvent(const IPointer::SAxisEvent& e, SP<IPointer> po
 
     bool found = false;
     if (e.source == WL_POINTER_AXIS_SOURCE_WHEEL && e.axis == WL_POINTER_AXIS_VERTICAL_SCROLL) {
-        if (e.delta < 0)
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
one of the moments of all time

wabarTM:
Probably worth mentioning in breaking changes tracker since this will break literally every config

vaxerski:
well no it will fix every cofig that used mosue_ to bind anything, I doubt many do tbh

rakibdev:
it broke everything. lost my job, family

PR #14755, opened 2026-05-22
hyprpm: allow distro packagers to provide extra cflags for building plugins

GaugeAndGravity:

This PR allows distro packagers to inject extra CFLAGS when `hyprpm` is building plugins.

I make this because `hyprpm` cannot build many plugins on Gentoo. For example, `hyprbars` uses `#include <lua.h>`, but does not use `pkg-config` to search `lua` properly. On many systems, like Gentoo and Debian, multiple versions installing  of Lua is allowed, and `/usr/include/lua.{c,cpp}` does not exist. The headers are inside `/usr/include/luax.y/`, so they are not visible globally.

Since versioned include directories can happen on many distros, I think it's good to add this optional compile-time macro for packagers to append extra `CFLAGS` and `CXXFLAGS` to `hyprpm` plugin build steps.

Tested on my Gentoo.

It should be ready.

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -313,8 +313,10 @@ bool CPluginManager::addNewPluginRepo(const std::string& url, const std::string&
 
         progress.printMessageAbove(infoString("Building {}", p.name));
 
+        const auto env = getPluginBuildEnv();
+
         for (auto const& bs : p.buildSteps) {
-            const auto CMD_RAW = nixDevelopIfNeeded(std::format("cd {} && PKG_CONFIG_PATH=\"{}\" {}", m_szWorkingPluginDirectory, getPkgConfigPath(), bs), HLVER);
+            const auto CMD_RAW = nixDevelopIfNeeded(std::format("cd {} && {} {}", m_szWorkingPluginDirectory, env, bs), HLVER);
 
             if (!CMD_RAW) {
                 progress.printMessageAbove(failureString("Failed to build {}: {}", p.name, CMD_RAW.error()));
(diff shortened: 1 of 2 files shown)


hyprpm/src/core/PluginManager.cpp
```diff
@@ -1035,6 +1039,18 @@ bool CPluginManager::hasDeps() {
     return hasAllDeps;
 }
 
+std::string CPluginManager::getPluginBuildEnv() {
+    std::string env = std::format("PKG_CONFIG_PATH=\"{}\"", getPkgConfigPath());
+
+#if defined(HYPRPM_EXTRA_CFLAGS)
+    if (std::string_view{HYPRPM_EXTRA_CFLAGS}.size() > 0) {
```
vaxerski: no {}

hyprpm/src/core/PluginManager.cpp
```diff
@@ -1035,6 +1039,18 @@ bool CPluginManager::hasDeps() {
...
+        env += std::format(" CFLAGS=\"{} $CFLAGS\" CXXFLAGS=\"{} $CXXFLAGS\"", HYPRPM_EXTRA_CFLAGS, HYPRPM_EXTRA_CFLAGS);
```
vaxerski: shouldn't the other be EXTRA_CXXFLAGS?

hyprpm/src/core/PluginManager.cpp
GaugeAndGravity: Thanks, this format is fixed.

hyprpm/src/core/PluginManager.cpp
GaugeAndGravity: OK, I make them separate as two macros now.

szczot3k:
This change originated inside of hyproverlay. While we could maintain it there, I believe it would be best suited upstream.

vaxerski (approved):
thanks!

PR #14756, opened 2026-05-22
algo/master: guard target in remove

vaxerski:
Fixes #14613

needs a test because I couldnt repro


Szwagi:
It was inconsistent to crash it before so not 100%, but couldn't get it to crash after this patch 👍

PR #4406, opened 2024-01-10
Set preferred scale on monitor config reload

1over137:
#### Describe your PR, what does it fix/add?
Fix #4363

Send preferred scale to surfaces on monitor config change. This is needed for
some applications to adjust their rendered scale properly, including
KeepassXC, alacritty, and mpv, otherwise the rendered scale is not updated for
the applications, resulting in blurry windows due to nearest neighbor or
linear scaling.  

Fix is confirmed to work as a patch on v0.34.0

PR #14771, opened 2026-05-23
layersurface: inform layer surfaces of scale changes

awused:
No AI.

Informs layer surfaces of monitor scale/transform changes at the same time window surfaces are informed. This has the same logic as `CWindow::updateSurfaceScaleTransformDetails`.

Also fixes https://github.com/hyprwm/Hyprland/discussions/13085 - layer subsurfaces not receiving scales on map/commit.

I am not sure if the code in onNewMonitor is actually reachable or if m_layerSurfaceLayers will always be empty at that point.

ctest passes, but I wasn't able to complete a hyprtester run before or after making the change. It didn't regress any tests that it was able to complete. 

Ready.

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -58,6 +58,7 @@ using namespace Hyprutils::Utils;
 using namespace Hyprutils::OS;
 using enum NContentType::eContentType;
 using namespace NColorManagement;
+using namespace Desktop::View;
 using namespace Render::GL;
 using namespace Monitor;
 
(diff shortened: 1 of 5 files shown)


vaxerski (commented):
I don't like this. It's flaky. Maybe we should add one method that does both updateSurfaceScaleTransform for windows and updates layers on a given monitor, then use it when mode changes / on connect?

awused:
> I don't like this. It's flaky. Maybe we should add one method that does both updateSurfaceScaleTransform for windows and updates layers on a given monitor, then use it when mode changes / on connect?

Makes sense, moved it into CMonitor::updateSurfaceScaleTransformDetails. The loop in applyMonitorRule was running for windows on all monitors, not just the changed one, but as far as I can tell that was unintentional in https://github.com/hyprwm/Hyprland/pull/4406.

src/helpers/Monitor.cpp
```diff
@@ -1868,6 +1864,22 @@ uint8_t CMonitor::isTearingBlocked(bool full) {
...
+        return;
+
+    for (auto const& w : g_pCompositor->m_windows) {
+        if (w->m_monitor == m_self)
+            w->updateSurfaceScaleTransformDetails();
+    }
+
+    for (auto const& lsl : m_layerSurfaceLayers) {
+        for (auto const& ls : lsl) {
+            ls->updateSurfaceScaleTransformDetails();
```
vaxerski, line 1878: forgor to check monitor?

awused: Is that necessary? I thought m_layerSurfaceLayers was only the layers on this monitor.

vaxerski: oh woops right. My bad, forgot we do that since 2024.

vaxerski (approved):

PR #14774, opened 2026-05-23
tests: some code improvements 

cebem1nt:
Makes hyprtester code a bit more readable, adds NLog::info and NLog::error :)

nope

ready


vaxerski (approved):

hyprtester/src/tests/shared.cpp
```diff
@@ -13,7 +13,6 @@ using namespace Hyprutils::Memory;
 
 // Almost everywhere `Tests::spawnKitty` is used, the return value is immediately tested against `nullptr`
 // and the test fails if it is.
-// TODO: add a test macro for that.
```
kolayne, line 16: This is not correct :((

1. What this "TODO" was for is to create a macro to be used _instead_ of `ASSERT(!!Tests::spawnKitty("kitty"), true);`. This expression is just a lazy hack. Moreover, in some places in the tests a more readable, though, even more verbose version is used.
2. Now the rest of the comment (the previous two lines) does not make much sense.

hyprtester/src/Log.hpp
```diff
@@ -16,4 +16,27 @@ namespace NLog {
...
+    //NOLINTNEXTLINE
+    void error(std::format_string<Args...> fmt, Args&&... args) {
+        std::string logMsg = "";
+
+        logMsg += std::vformat(fmt.get(), std::make_format_args(args...));
+
+        std::println("{}{}{}", Colors::RED, logMsg, Colors::RESET);
+        std::fflush(stdout);
+    }
 }
```
kolayne, line 42: This change is a nice idea!

It would also be nice if:
- the `info` and `error` methods reused the previous code (by calling `log`), rather than duplicating it;
- the existing uses of `NLog::log` in the existing tests were replaced with calls to these methods.

hyprtester/src/main.cpp
```diff
@@ -47,20 +46,33 @@ using Path = std::filesystem::path;
...
+    struct SSettings {
+        Path configPath;
+        Path binaryPath;
+        Path pluginPath;
+    };
+
+    struct STestsInfo {
+        unsigned long long       failed, total;
+        std::vector<std::string> failedNames;
+    };
```
kolayne, line 59: This structure definition contains redundancy: the size of `.failedNames` and the value of `.failed` are the same thing.

This is problematic because it's unclear how to interpret a value of `STestsInfo` with redundancy: when you get a value of `STestsInfo`, should you check its `.failed` or its `.failedNames.size()`? What should you do if they don't match?

hyprtester/src/main.cpp
```diff
@@ -146,149 +150,121 @@ static bool preTestCleanup() {
...
-        NLog::log("{}Internal failure: failed to reset cursor position", Colors::RED);
+        NLog::error("Internal failure: failed to reset cursor position");
         failed = true;
     }
 
     return !failed;
 }
 
-static void runTests(std::map<const char*, CTestCase&>& testCases, std::string_view suiteName, std::vector<std::string>& failedTestNames) {
+static void runTests(std::map<const char*, CTestCase&>& testCases, std::string suiteName, struct STestsInfo& testsInfo) {
```
kolayne, line 178: Why `std::string` instead of `std::string_view`?

hyprtester/src/tests/misc/plugin.cpp
```diff
@@ -0,0 +1,23 @@
...
+}
+
+TEST_CASE(pluginVkb) {
+    const auto RESPONSE = getFromSocket("/eval hl.plugin.test.vkb()");
+
+    if (RESPONSE != "ok")
+        NLog::log("{}Vkb tests failed, tests returned:\n{}{}", Colors::RED, Colors::RESET, RESPONSE);
+    else
+        NLog::log("{}Test passed: vkb test from plugin", Colors::GREEN);
+}
```
kolayne, line 23: That's a great improvement, thank you.

kolayne (commented):
Hello. Thank you for this PR! It was useful for another patch that I was making today.

I left some comments on the code changes that raised concerns when I was reading it.

I included fixes for some of the issues (the todo, `STestsInfo`, and `runTests`) in #15025, you are welcome to take a look. If I misunderstood some of your changes or missed something, please let me know!

The rest I did not do anything with. If you're interested, you are welcome to submit another patch!

hyprtester/src/main.cpp
cebem1nt: It was later converted to std::string anyway, and I don't thing using string_view here gives any advantage

Issue #9126, opened 2025-01-21
Hyprland sends an extra fractional scale event


kovidgoyal:
### Already reported ? *

- [x] I have searched the existing open and closed issues.

### Regression?

Not a regression - it's an old bug

### System Info and Hyprland Version

Hyprland 0.46.2 built from branch  at commit 0bd541f2fd902dbfa04c3ea2ccf679395e316887  (version: bump to 0.46.2).
Date: Thu Dec 19 19:26:47 2024
Tag: v0.46.2, commits: 5566
built against:
 aquamarine 0.7.1
 hyprlang 0.6.0
 hyprutils 0.3.3
 hyprcursor 0.1.11
 hyprgraphics 0.1.1

flags set:
debug

System Information:
System name: Linux
Node name: getafix
Release: 6.12.8-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Thu, 02 Jan 2025 22:52:26 +0000

GPU information: 
07:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Cezanne [Radeon Vega Series / Radeon Vega Mobile Series] [1002:1638] (rev d1) (prog-if 00 [VGA controller])

os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo

plugins:
	unknown: not runtime

### Description

Hyprland sends two fractional scale events, one with scale 1 before the surface configure and one with the actual scale after the surface configure. This is inefficient as it leads to clients performing unnecessary resizes. Ideally, Hyprland should send just one fractional scale event, before the surface configure, but at the least it should not send two fractional scale events.

For more discussion of the problems caused by this, see:
https://github.com/kovidgoyal/kitty/issues/8236#issuecomment-2604024714

### How to reproduce

Run 

kitty --debug-rendering

Look for the Fractional scale lines in the resulting log.

### Attach not paste

[...]

PR #14798, opened 2026-05-24
fractionalscale: track if scale is known and send scales eagerly when known

awused:
No AI.

This tracks whether a fractional scale has been explicitly set for a surface or not, and if one has been set it sends it when requested in onGetFractionalScale even if the surface is currently unmapped. That shouldn't ever be true for top level window surfaces unless they've been unmapped, but it means subsurfaces of unmapped surfaces will get the same scale as their parent.

Layers spawned on a specific monitor have the scale/transform immediately set from that monitor. This doesn't change the behaviour if the monitor's scale changes before mapping the layers.

Subsurfaces inherit the scales and transforms of their parents on creation, if known. This seems reasonable to me but it's the part I'm least confident about.

I looked into the bug report around double/redundant scale events (https://github.com/hyprwm/Hyprland/issues/9126) and confirmed this this does not regress it.

`kitty --debug-rendering` at 0.55.2 and this commit:
[kitty-55.2.log](https://github.com/user-attachments/files/28198366/kitty-55.2.log)
[kitty-debug.log](https://github.com/user-attachments/files/28198367/kitty-debug.log)

Worth reviewing closely but I think it's ready.

### DIFF
--- a/src/protocols/FractionalScale.cpp
+++ b/src/protocols/FractionalScale.cpp
@@ -1,5 +1,6 @@
 #include "FractionalScale.hpp"
 #include <algorithm>
+#include "Compositor.hpp"
 #include "core/Compositor.hpp"
 
 CFractionalScaleProtocol::CFractionalScaleProtocol(const wl_interface* iface, const int& ver, const std::string& name) : IWaylandProtocol(iface, ver, name) {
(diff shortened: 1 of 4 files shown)


vaxerski (approved):
this is fine

awused:
Ran the keybinds tests locally after patching in https://github.com/hyprwm/Hyprland/pull/14846, seems like the failure was a flake.

vaxerski:
yes some tests do a funny sometimes

PR #14802, opened 2026-05-24
Fire off 'layoutchanged' event on layout change

MicaelLoberg:
Currently there's no IPC event when a workspace's tiling layout changes (e.g. dwindle ->master), 
so external tools like bar widgets have no way to detect it without polling. 
This adds a layoutchanged event emitted on socket2 with the format

layoutchanged>>workspacename,layoutname, 

fired from updateWorkspaceLayouts() after the algorithm switch.

Ready for merging!

### DIFF
--- a/hyprtester/src/tests/main/layout.cpp
+++ b/hyprtester/src/tests/main/layout.cpp
@@ -1,3 +1,6 @@
+#include <filesystem>
+#include <fstream>
+#include <thread>
 #include "../shared.hpp"
 #include "../../shared.hpp"
 #include "../../hyprctlCompat.hpp"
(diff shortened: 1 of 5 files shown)


vaxerski (commented):
I am against any changes to the IPC at this point, rather focus on lua. This would be better as an event on the bus and forwarded to lua for access

MicaelLoberg:
Removed the IPC and added bus events instead.
I also added events for special workspace opened/closed. Though the changes were similar enough to be in the same PR.

src/config/lua/LuaEventHandler.cpp
```diff
@@ -137,6 +137,24 @@ CLuaEventHandler::CLuaEventHandler(lua_State* L) : m_lua(L) {
...
+        });
+    }));
+    m_listeners.push_back(bus()->m_events.workspace.specialOpened.listen([this](PHLWORKSPACE ws, PHLMONITOR mon) {
+        dispatch("workspace.special_opened", 2, [&] {
+            CLuaWorkspace::push(m_lua, ws);
+            CLuaMonitor::push(m_lua, mon);
+        });
+    }));
+    m_listeners.push_back(bus()->m_events.workspace.specialClosed.listen([this](PHLWORKSPACE ws, PHLMONITOR mon) {
+        dispatch("workspace.special_closed", 2, [&] {
```
vaxerski: these are redundant, there's `actgive_special` if you rebase on main

tests/config/lua/LuaEventHandler.cpp
```diff
@@ -113,3 +113,70 @@ TEST(ConfigLuaEventHandler, callbackLuaErrorDoesNotCrashDispatch) {
...
+            lastLayout = layout
+        end
+    )"),
+              LUA_OK);
+
+    const int  ref    = refGlobalFunction(L, "onLayoutChanged");
+    const auto handle = handler.registerEvent("workspace.layout_changed", ref);
+    ASSERT_TRUE(handle.has_value());
+
+    Event::bus()->m_events.workspace.layoutChanged.emit(nullptr, "master");
```
vaxerski: these don't test much. Leave them be, but add integration tests in hyprtester.

MicaelLoberg:
Dropped the special workspace event commit and added an integration test for the layout change event

hyprtester/src/tests/main/layout.cpp
```diff
@@ -141,3 +144,35 @@ TEST_CASE(focusPreservedLayoutChange) {
...
+
+    OK(getFromSocket("/eval hl.on('workspace.layout_changed', function(ws, layout) "
+                     "local f = io.open('/tmp/hyprtester-layout-changed.txt', 'w') "
+                     "f:write(layout) f:close() end)"));
+
+    ASSERT(!!Tests::spawnKitty(), true);
+
+    OK(getFromSocket("r/eval hl.config({ general = { layout = 'master' } })"));
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
vaxerski, line 172: don't sleep, sync with the compositor. (dispatch any hypctl call)

PR #14804, opened 2026-05-24
config/workspace_rules: add ability to pass workspace as object, fix passing workspace by ID

erstarr:

- Passing workspace object as selectors to workspace_rules do not work: add that ability to match the claims of the wiki

- Passing special workspace's ID doesn't work as the selector string parser interprets it as a relative workspace selector instead of an ID.

```lua
        hl.workspace_rule({
        workspace = currentWorkspace.id,
        gaps_in = 0,
        gaps_out = 0
    })
```

When on special workspace; it passes `-INT`, which is interpreted as a relative workspace.

No.

This should work with all dispatches that also accept workspaces as arguments as this checks IDs as a part of the selector string, and the object change was only necessary for workspace_rules

TODO:

- [x] Wiki PR for the ID change

### DIFF
--- a/hyprtester/src/tests/main/workspaces.cpp
+++ b/hyprtester/src/tests/main/workspaces.cpp
@@ -752,6 +752,129 @@ TEST_CASE(workspacesCombined) {
     ASSERT(Tests::windowCount(), 0);
 }
 
+
+TEST_CASE() {
+
+
+    Tests::spawnKitty("kittycat");
+
+    // normal Workspace
+
+    // test both moving a window and focusing a workspace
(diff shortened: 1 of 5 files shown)


erstarr:
any problems with this @vaxerski?

vaxerski:
yes, there arent any tests

erstarr:
> yes, there arent any tests

Instead of adding a test for just this one case, all possible workspace_rule selector options should be tested instead.

If you concur, it's outside this PRs scope; should be done in a new PR.

vaxerski:
make the PR then and link it here.

PR #14820, opened 2026-05-26
popup: fix reposition artifacts

gulafaran:
visible() m_alpha check was reversed, m_lastPos cant be set in onReposition() and then later used in onCommit() it means it can be ahead of time and use wrong cords. also expand the damage box by 4 like done elsewhere.

fixes: https://github.com/hyprwm/Hyprland/discussions/14738

### DIFF
--- a/src/desktop/view/Popup.cpp
+++ b/src/desktop/view/Popup.cpp
@@ -69,7 +69,10 @@ eViewType CPopup::type() const {
 }
 
 bool CPopup::visible() const {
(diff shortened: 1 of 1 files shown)


tekstryder:
This fixes the tooltip issue on `pcmanfm-qt` [pour moi](https://github.com/hyprwm/Hyprland/discussions/14738#discussioncomment-17061122).

vaxerski (approved):

PR #14821, opened 2026-05-26
desktop/window: allow focus while held to non-OR X11 windows

vaxerski:
the drag bug only happens on OR windows, so we should just skip those. Regular toplevels should be focusable.

should fix #14782

tagging @nynxz and @enzi for testing


Nynxz:
yep this fixes https://github.com/hyprwm/Hyprland/discussions/14782

thank you

https://github.com/user-attachments/assets/4af31646-df1e-462e-a756-bb9c2b4870a7

PR #14841, opened 2026-05-27
added occupied_workspace blur

gusjengis:
A new setting that does what special workspace blur does anytime there is a window open in the current workspace. With this on, the only time the wallpaper is in focus is when there is nothing open on screen. Did some fiddling to make sure that there is only ever one blur applied at time. Opening a special workspace on top of an already occupied workspace reapplies the blur.

This is something I want, and it's good enough for me. I'll keep cherry picking this commit into new builds for myself. Just making this MR to share the idea in case someone else likes it.


vaxerski (commented):
this should definitely be a workspace rule instead. Also should probably be tracked differently, this is just clutter.

Consider keeping an alpha in CMonitor so its not abrupt

gusjengis:
Switched to a workspace rule.

Did a [wiki PR](https://github.com/hyprwm/hyprland-wiki/pull/1553)

vaxerski:
we already update workspace rules on window number changes, why are you adding 10 different calls for update workspace blur?

gusjengis:
[this](https://github.com/hyprwm/Hyprland/pull/14841/commits/b3d33edd15aa696512333e81fc8c7b66686969cd) good?

vaxerski:
no, you are still calling updateWorkspaceBlur in like 5 places, which is bugprone. What is stopping you from calling it once somewhere like `CWorkspace::updateWindowData` and once in CMonitor::changeWorkspace?

PR #14846, opened 2026-05-28
hyprtester: add -t option for running specific tests

ArchSav:
Lets devs run a specific set of tests without waiting for every test to run.

test suite is referenced as 'custom', not the section the test is actually from

yes

### DIFF
--- a/hyprtester/src/main.cpp
+++ b/hyprtester/src/main.cpp
@@ -48,9 +48,10 @@ using Path = std::filesystem::path;
 
 namespace {
     struct SSettings {
-        Path configPath;
-        Path binaryPath;
-        Path pluginPath;
+        Path                     configPath;
+        Path                     binaryPath;
+        Path                     pluginPath;
+        std::vector<std::string> requestedTests;
     };
 
     struct STestsInfo {
@@ -79,10 +80,11 @@ static bool hyprlandAlive() {
 [[noreturn]] static void helpAndDie(int exit_code) {
     NLog::log("usage: hyprtester [arg [...]].\n");
     NLog::log(R"(Arguments:
-    --help              -h       - Show this message again
-    --config FILE       -c FILE  - Specify config file to use (default: './test.lua')
-    --binary FILE       -b FILE  - Specify Hyprland binary to use (default: '../build/Hyprland')
-    --plugin FILE       -p FILE  - Specify the location of the test plugin (default: './'))");
+    --help              -h         - Show this message again
+    --config FILE       -c FILE    - Specify config file to use (default: './test.lua')
+    --binary FILE       -b FILE    - Specify Hyprland binary to use (default: '../build/Hyprland')
+    --plugin FILE       -p FILE    - Specify the location of the test plugin (default: './')
+    --test [tests]      -t [tests] - Specify list of tests to run (separated by spaces))");
 
     std::exit(exit_code);
 }
@@ -102,9 +104,16 @@ static Path validatePathOrDie(Path path) {
 static SSettings parseSettings(const std::span<const char*> args) {
     static const auto cwd = std::filesystem::current_path();
     SSettings         settings{};
+    bool              testSetOn = false;
 
     for (auto it = args.begin(); it < args.end(); it++) {
         std::string_view value = *it;
+
+        if (testSetOn && !value.starts_with("-")) {
+            settings.requestedTests.emplace_back(value);
+            continue;
+        } else
+            testSetOn = false;
 
         if (value == "--config" || value == "-c") {
             if (std::next(it) == args.end()) {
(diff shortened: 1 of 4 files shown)


vaxerski (approved):

hyprtester/src/main.cpp
```diff
@@ -197,10 +211,43 @@ static void runTests(std::map<const char*, CTestCase&>& testCases, std::string s
...
+        for (const auto& name : tInfo.failedNames) {
+            NLog::log("{}\t- {}", Colors::RED, name);
+        }
+    }
+
+    kill(hyprlandProc->pid(), SIGKILL);
+    hyprlandProc.reset();
+
+    return tInfo.failed > 0;
+}
```
kolayne, line 231: I'd argue this function takes on too many responsibilities. It:
- Terminates Hyprland and cleans it up,
- Report test results,
- Indicates to the caller whether there were failing tests.

Also, it copies its argument, which contains a vector of strings - a potentially expensive implicit copy.

hyprtester/src/main.cpp
```diff
@@ -197,10 +211,43 @@ static void runTests(std::map<const char*, CTestCase&>& testCases, std::string s
...
+        if (mainTestCases.contains(test)) {
+            tempTestList.emplace(test, mainTestCases.at(test));
+            continue;
+        } else if (clientTestCases.contains(test)) {
+            tempTestList.emplace(test, clientTestCases.at(test));
+            continue;
+        } else if (miscTestCases.contains(test)) {
+            tempTestList.emplace(test, miscTestCases.at(test));
+        } else
+            NLog::log("{}ERROR: '{}' Invalid test name", Colors::RED, test);
```
kolayne, line 249: I think this should be a hard failure: if the user provided an invalid name, hyprtester should fail, rather than run the rest of the tests.

hyprtester/src/main.cpp
```diff
@@ -241,6 +288,12 @@ int main(int argc, char** argv, char** envp) {
 
     struct STestsInfo tInfo = {0};
 
+    if (!tempTestList.empty()) {
+        runTests(tempTestList, "custom", tInfo);
+
+        std::exit(quitTests(tInfo));
+    }
```
kolayne, line 295: It's a bit awkward how the code takes a completely different path here, in case the user requested tests. Also, it terminates with a `std::exit` instead of a `return`, which is surprising.

<hr>

Also, when you provide the value `"custom"` as an argument to `runTests`, its output will no longer make sense: `runTests` uses that parameter as a namespace name: to tell the user what group this tests belongs to. But with this argument, `runTest` will list all the tests as if they belong to the test namespace called `"custom"` (which does not really exist).

Though, this second problem is not trivial to resolve because there is a design issue: test cases do not carry the knowledge about which namespace they belong to, so `runTests` tries to get this information from an external source.

hyprtester/src/main.cpp
```diff
@@ -102,10 +104,17 @@ static Path validatePathOrDie(Path path) {
...
+    bool              testSetOn = false;
 
     for (auto it = args.begin(); it < args.end(); it++) {
         std::string_view value = *it;
 
+        if (testSetOn && !value.starts_with("-")) {
+            settings.requestedTests.emplace_back(value);
+            continue;
+        } else
+            testSetOn = false;
```
kolayne, line 116: An option (`-t`) that is specified once and then accepts an arbitrary, possibly empty set of values, is a very non-standard/uncommon way to accept command-line arguments. Following the principle of least surprise, I would instead implement one of the following:
- Like all other flags, `-t test` is an option that accepts one value (a test name) at a time, but can be mentioned multiple times, e.g., `hyprtester ... -t test_a -t test_b`;
- Like all other flags, `-t tests` is an option that accepts one value at a time, but the value is the set of all requested tests with a separator (such as a space or a comma), e.g., `hyprtester ... -t test_a,test_b` or `hyprtester ... -t "test_a test_b"`
- Test names are positional arguments to `hyprtester` that don't require an option, e.g., `hyprtester test_a test_b ...` or `hyprtester ... -- test_a test_b`. That's how, e.g., gcc, clang, grep, etc accept their sets of files to work with.

kolayne (commented):
Hello. Thank you for adding this option to hyprtester, it'll definitely be useful for me!

I left here some notes on your implementation where I disagree with your approach or see potential issues (though I'm a bit late to the party 😅).

I include fixes for the listed issues in #15025, you are welcome to take a look. If I misunderstood some of your changes or missed something, please let me know!

Btw, there is also a [wiki page for tests](https://wiki.hypr.land/Contributing-and-Debugging/Tests/), which documents hyprtester. Tests selection is a useful feature, and it's part of the hyprtester UI, so I will document it there. If you add more features to hyprtester in the future, don't forget to [add documentation](https://github.com/hyprwm/hyprland-wiki), too!

PR #14848, opened 2026-05-28
fix(hyprctl): dispatch v2 fallback

sim590:

Since 0.55, hyprctl IPC calls are routed to lua evaluator when the user has Lua config active and just errors if the call failed either if it was a true lua call or a old-style hyprlang call. I judge that this behaviour is extreme and unnecessary. It's perfectly possible to reroute the calls to the hyprlang evaluator if Lua evaluator failed AND the input expression didn't have character `(` in it. It's in fact an heuristic that is used right now, but only to return blank evaluator return output. I think it would be fair to fall back to the direct m_dispatchers lookup before reporting a failure.

Why? Well, first because this program is an IPC program. It shouldn't care what config the user is using to handle IPC calls. This has **significant** impact on the ecosystem. More precisely, I maintain [hypr-layout](https://github.com/sim590/hypr-layout) which relies on hy3 dispatchers to function. This situation can seem very confusing for users of my program as hypr-layout will not fail at all when the user is running on 0.55 with hyprlang config, but it will fail to work properly when the user has Lua config.

I don't see any reason why the this backward compatibility fix could not be applied while 0.55 Hyprland still supports Hyprlang config. But again, I think that IPC calls shouldn't be impacted by users' configuration. I think that hyprlang based IPC should still continue to live even though users have passed on to lua config. At least for a while to give other hyprctl based tools developers the time to migrate their IPC calls.

No.

It's ready. I tested it. I'm running it right now and it fixed hypr-layout.

### DIFF
--- a/src/debug/HyprCtl.cpp
+++ b/src/debug/HyprCtl.cpp
@@ -1103,17 +1103,14 @@ static std::string dispatchRequest(eHyprCtlOutputFormat format, std::string in) 
     // get rid of the dispatch keyword
     in = in.substr(in.find_first_of(' ') + 1);
 
-    if (Config::mgr()->type() == Config::CONFIG_LUA) {
-        // For lua, this is just a wrapper for `eval("hl.dispatch(in)")
+    if (Config::mgr()->type() == Config::CONFIG_LUA && in.contains("(")) {
+        // Input contains parentheses, so it is a Lua expression (e.g.
+        // hl.plugin.foo.bar("arg")).  Evaluate it through the Lua config
+        // manager.  Dispatcher names registered via addDispatcherV2 never
+        // contain '(', so the direct lookup below will handle them.
         std::string evalStr = std::format("return hl.dispatch({})", in);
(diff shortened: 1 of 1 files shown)


sim590:
Just mentioning here that I have an alternative fix in 82afc8dc58f450bf43584ba9f4a0d8c5658dca82. It's just another approach for doing the same thing. Either we decide to detect the absence of Lua preemptively (according to the heuristic) and avoid calling the Lua evaluator (which is what 82afc8dc58f450bf43584ba9f4a0d8c5658dca82 would do) or we try Lua first. The latter is my initial proposal.

vaxerski:
No, because that dispatch table will be removed once hyprlang is gone. It's now undocumented and not meant to be used

sim590:
@vaxerski 

> No, because that dispatch table will be removed once hyprlang is gone. It's now undocumented and not meant to be used

Couldn't that backward compatibility change be active until hyprlang is gone? This is clearly for sake of consistency. Still supporting hyprlang while breaking the IPC for the direct m_dispatchers calls seems wrong.

This is a simple fix that ease the transition for the whole ecosystem. Why discard it?

vaxerski:
you're moving the problem to happen just the same, later.

PR #14849, opened 2026-05-28
renderer: send frame callbacks on presented if no change

gulafaran:
some very picky clients like firefox expects a frame callback to continue rendering and send buffers, if the frame is returning early because no change its not going to recieve it, this becomes very noticeable on a single firefox window and using new_render_scheduling where it renders the monitor one extra time where occasionally nothing had really changed.


vaxerski (approved):
thanks!

PR #14856, opened 2026-05-28
fix submap enter bind == bind inside submap

bendi-github:
This adds the test vaxry wanted for #14578. As per the docs, the test case should fail without the fix, which it does. Adding the bind to `test.lua` without @codenameone-akshat's commit, causes the submap test to fail. With the change, the test passes.
  
no

Ready, includes the test and the fix(in squashed form), but as this rudely supersedes @codenameone-akshat's PR, perhaps it would be better if they incorporated the test into their PR.

### DIFF
--- a/hyprtester/src/tests/main/keybinds.cpp
+++ b/hyprtester/src/tests/main/keybinds.cpp
@@ -469,6 +469,14 @@ SUBTEST(submap) {
     press(KEY_P);
     EXPECT_CONTAINS(getFromSocket("/submap"), "default");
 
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
tnx

codenameone-akshat:
Thanks for this! :D
No worries at all. This is not rude xD
I didn't get time to add the tests soon enough.

Shall I close my original PR vaxry?

vaxerski:
I'll close it

PR #14868, opened 2026-05-29
renderer: dont send frame on discarded elements

gulafaran:
as wl_surface::frame says.
A server should avoid signaling the frame callbacks if the surface is not visible in any way, e.g. the surface is off-screen, or completely obscured by other opaque surfaces.

also discard if the element renderer early outs if its not visible.

to not regress firefox further it requires: https://github.com/hyprwm/Hyprland/pull/14849

lowers gpu usage drasticly here.

### DIFF
--- a/src/protocols/core/Compositor.cpp
+++ b/src/protocols/core/Compositor.cpp
@@ -725,7 +725,8 @@ void CWLSurfaceResource::updateCursorShm(CRegion damage) {
 }
 
 void CWLSurfaceResource::presentFeedback(const Time::steady_tp& when, PHLMONITOR pMonitor, bool discarded) {
-    frame(when);
+    if (!discarded)
+        frame(when);
 
     auto FEEDBACK = makeUnique<CQueuedPresentationData>(m_self.lock());
     FEEDBACK->attachMonitor(pMonitor);
--- a/src/render/ElementRenderer.cpp
+++ b/src/render/ElementRenderer.cpp
@@ -265,8 +265,10 @@ void IElementRenderer::drawSurface(WP<CSurfacePassElement> element, const CRegio
 
     auto cancelRender = false;
     auto clipRegion   = element->visibleRegion(cancelRender);
-    if (cancelRender)
+    if (cancelRender) {
+        element->discard();
         return;
+    }
 
     // check for fractional scale surfaces misaligning the buffer size
     // in those cases it's better to just force nearest neighbor
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -195,7 +195,12 @@ IHyprRenderer::IHyprRenderer() {
                 w->wlSurface()->resource()->breadthfirst(
                     [](SP<CWLSurfaceResource> surf, const Vector2D& offset, void* data) {
                         surf->m_stateQueue.unlockFirst(LOCK_REASON_FENCE | LOCK_REASON_FIFO | LOCK_REASON_TIMER);
-                        surf->presentFeedback(Time::steadyNow(), Desktop::focusState()->monitor(), true);
+
+                        //#TODO: figure out if renderunfocused should send discarded or not here.
+                        // if we dont discard it presentFeedback will send a frame callback.
+                        auto NOW = Time::steadyNow();
+                        surf->presentFeedback(NOW, Desktop::focusState()->monitor(), true);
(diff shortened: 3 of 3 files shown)


UjinT34:
Render unfocused sends discarded presentations and should keep sending frame events

gulafaran:
> Render unfocused sends discarded presentations and should keep sending frame events

should render unfocused really send discarded then?

UjinT34:
Not sure. It's not clear whether the 'discarded' event should be sent every time the user doesn't see the content update or only when this happens due to mentioned reasons. With reduced render unfocused fps we can assume that there is always an extra pending frame update that supersedes the current one but that's not guaranteed.
> There are two possible outcomes: the content update is presented to the user, and a presentation timestamp delivered; or, the user did not see the content update because it was superseded or its surface destroyed, and the content update is discarded.

https://wayland.app/protocols/presentation-time#wp_presentation_feedback

gulafaran:
> Not sure. It's not clear whether the 'discarded' event should be sent every time the user doesn't see the content update or only when this happens due to mentioned reasons. With reduced render unfocused fps we can assume that there is always an extra pending frame update that supersedes the current one but that's not guaranteed.
> 
> > There are two possible outcomes: the content update is presented to the user, and a presentation timestamp delivered; or, the user did not see the content update because it was superseded or its surface destroyed, and the content update is discarded.
> 
> https://wayland.app/protocols/presentation-time#wp_presentation_feedback

hrml oh well this restores "old" logic to renderunfocused, il leave that for future "render unfocus" work to figure out

Skidamek:
For me this PR makes firefox/zen very laggy, you can repro it by selecting lines of text on some website until renderer stops updating then you can switch between tabs and while the switch happens internally (i can see different tab name in the browser title on my bar) the new tab doesn't render

gulafaran:
> For me this PR makes firefox/zen very laggy, you can repro it by selecting lines of text on some website until renderer stops updating then you can switch between tabs and while the switch happens internally (i can see different tab name in the browser title on my bar) the new tab doesn't render

yeah thanks, put in some breakpoints found a big oh boy issue thats gonna have to be solved before this can be considered. so far it looks like if a surface pushes commits faster then the frame rate of your monitor its going to end up rendered twice/or more. all those commits/buffers/ gets a presentfeedback and all of those gets sent as presented at once on next monitor pageflip, even tho its technically was only the last commit/buffer that was "presented".

gonna have to dig into the elementrenderer etc. il mark this as a draft for now.

umbrageodotus:
> > For me this PR makes firefox/zen very laggy, you can repro it by selecting lines of text on some website until renderer stops updating then you can switch between tabs and while the switch happens internally (i can see different tab name in the browser title on my bar) the new tab doesn't render
> 
> yeah thanks, put in some breakpoints found a big oh boy issue thats gonna have to be solved before this can be considered. so far it looks like if a surface pushes commits faster then the frame rate of your monitor its going to end up rendered twice/or more. all those commits/buffers/ gets a presentfeedback and all of those gets sent as presented at once on next monitor pageflip, even tho its technically was only the last commit/buffer that was "presented".
> 
> gonna have to dig into the elementrenderer etc. il mark this as a draft for now.

@gulafaran you sure that isn't related to that aq pr you sent to "not render the same frame twice"?

gulafaran:
> > > For me this PR makes firefox/zen very laggy, you can repro it by selecting lines of text on some website until renderer stops updating then you can switch between tabs and while the switch happens internally (i can see different tab name in the browser title on my bar) the new tab doesn't render
> > 
> > 
> > yeah thanks, put in some breakpoints found a big oh boy issue thats gonna have to be solved before this can be considered. so far it looks like if a surface pushes commits faster then the frame rate of your monitor its going to end up rendered twice/or more. all those commits/buffers/ gets a presentfeedback and all of those gets sent as presented at once on next monitor pageflip, even tho its technically was only the last commit/buffer that was "presented".
> > gonna have to dig into the elementrenderer etc. il mark this as a draft for now.
> 
> @gulafaran you sure that isn't related to that aq pr you sent to "not render the same frame twice"?

yes the entire presentation protocol is done completly wrong in hl -git, https://github.com/hyprwm/Hyprland/pull/15031 mostly solves it.

also what "not render the same frame twice" pr?

umbrageodotus:
it was in aq and related to new render mode

PR #14884, opened 2026-05-30
example/hyprland.lua: Use keycodes for switching workspaces

MayeulC:
This should improve compatibility with some layouts (example: azerty), at no cost to existing qwerty users.

The problem is [stated here](https://wiki.hypr.land/Configuring/Basics/Binds/#workspace-bindings-on-non-qwerty-layouts): azerty layouts use SHFIT + <key> for digits (I  agree that it is annoying; in practice most users use a numpad).

I am proposing to change the current workspace bindings to use keycodes. That should work transparently across more layouts. This also shows an example of keycode bindings in the default config.

Should I change the default bindings too? I am unsure if this example config is the one being used when no config file is present.

Ready to merge, though I can change the default config as well if it is located somewhere else.

### DIFF
--- a/example/hyprland.lua
+++ b/example/hyprland.lua
@@ -272,10 +272,11 @@ hl.bind(mainMod .. " + down",  hl.dsp.focus({ direction = "down" }))
 
 -- Switch workspaces with mainMod + [0-9]
 -- Move active window to a workspace with mainMod + SHIFT + [0-9]
+-- Using keycodes for compatibility with some non-qwerty keyboard layouts
 for i = 1, 10 do
-    local key = i % 10 -- 10 maps to key 0
-    hl.bind(mainMod .. " + " .. key,             hl.dsp.focus({ workspace = i}))
-    hl.bind(mainMod .. " + SHIFT + " .. key,     hl.dsp.window.move({ workspace = i }))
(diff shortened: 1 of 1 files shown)


vaxerski:
hm, I don't know. What do other envs do?

MayeulC:
i3 and sway seem to bind keynames by default as well.
* [sway](https://github.com/swaywm/sway/blob/3302f1ce5c1a84229673c96451576ab25674f125/config.in#L116) binds key names.
* i3 [as well](https://github.com/i3/i3/blob/next/etc/config#L141), but the `i3-config-wizard` tool reads a [keycode config](https://github.com/i3/i3/blob/next/etc/config.keycodes#L128) to generate a locale-appropriate config, as written [here](https://github.com/i3/i3/blob/8f09b7f04632a9b99f2a5f584401659589dd008e/etc/config#L8-L9).

I agree that key names are clearer, and key codes are more portable, so the better choice is not obvious (i3 kind of picked both).

Dregu:
Well it's an *example* config and the old one was missing code: examples, so this is a better example that also works better, since the purpose is also to provide some useful features for newbies. I'd leave the old thing commented out as an alternative though.

```lua
--local key = i % 10
local key = "code:" .. (i + 9)
```

PR #14891, opened 2026-05-30
elementrenderer: allow smaller viewports near the top/left edges of the buffer

awused:

Allows smaller viewport sources as long as they're at least 1 pixel in size and uvBR isn't excessively close to 0.

I was using a full output capture and a viewport to make a screen magnifier, but noticed it wasn't working near the left and top edges of the screen. The actual size of the viewport didn't matter, just that the bottom right corner was at least 1% from the left/top edge, so it seems safe to relax these checks.

Ready.

### DIFF
--- a/src/render/ElementRenderer.cpp
+++ b/src/render/ElementRenderer.cpp
@@ -68,7 +68,7 @@ void IElementRenderer::calculateUVForSurface(PHLWINDOW pWindow, SP<CWLSurfaceRes
             uvTL = Vector2D(bufferSource.x / bufferSize.x, bufferSource.y / bufferSize.y);
             uvBR = Vector2D((bufferSource.x + bufferSource.width) / bufferSize.x, (bufferSource.y + bufferSource.height) / bufferSize.y);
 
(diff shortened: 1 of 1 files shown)


src/render/ElementRenderer.cpp
```diff
@@ -68,7 +68,7 @@ void IElementRenderer::calculateUVForSurface(PHLWINDOW pWindow, SP<CWLSurfaceRes
             uvTL = Vector2D(bufferSource.x / bufferSize.x, bufferSource.y / bufferSize.y);
             uvBR = Vector2D((bufferSource.x + bufferSource.width) / bufferSize.x, (bufferSource.y + bufferSource.height) / bufferSize.y);
 
-            if (uvBR.x < 0.01f || uvBR.y < 0.01f) {
+            if (bufferSource.width < 1.0 || bufferSource.height < 1.0 || uvBR.x < 0.00001f || uvBR.y < 0.00001f) {
```
vaxerski: relax I get, but what's the other w/h check for?

awused: I wanted to remove the uvBR checks entirely and just have the w/h checks, but if you're bringing them up it was probably the wrong approach. Removed them.

vaxerski (approved):
lgtv

PR #14893, opened 2026-05-30
layout: harden workspace layout refresh during monitor changes

awsms:
`CWorkspaceAlgoMatcher::updateWorkspaceLayouts()` iterated `g_pCompositor->getWorkspaces()`, a live filtered view over weak workspace refs. During monitor teardown, that underlying workspace vector can be mutated, and surviving workspace objects may already be inert or missing monitor, space, algorithm, or tiled-layout state.

Take a strong workspace snapshot with `getWorkspacesCopy()` before iterating, and skip workspaces that are not safe to refresh. 

I've been testing this patch for a while now (since lua conf was first implemented in Hyprland basically), as I came across this bug while scripting my monitors layout (Hyprland crashed when I switched from integrated-only to multi-screen quickly). Never had an issue with it so far afterwards.

Note that this patch was entirely written by codex.

The patch should be ready to be merged but feedback welcome + I guess I need to add some tests!

### DIFF
--- a/hyprtester/src/tests/main/layout.cpp
+++ b/hyprtester/src/tests/main/layout.cpp
@@ -3,6 +3,25 @@
 #include "../../hyprctlCompat.hpp"
 #include "tests.hpp"
 
+#include <chrono>
+#include <format>
+#include <thread>
+
+#include <hyprutils/utils/ScopeGuard.hpp>
+
+using namespace Hyprutils::Utils;
+
+static bool waitForMonitor(const char* name) {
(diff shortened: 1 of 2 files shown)


vaxerski:
yes some tests would be nice

vaxerski (dismissed):
tnx

vaxerski:
tests failing

vaxerski:
rebase pls

vaxerski (approved):
tnx

PR #14903, opened 2026-05-31
Tonemap settings

UjinT34:
Adds tonemap window rule to disable or modify tonemapping behavior. on (default) - current behavior, off - disable tonemapping, clamp - clamp source luminance to target luminance, limited - use compromised tonemapping suggested by @LionHeartP 

Might need an additional dispatcher for easier toggling.

Ready


LionHeartP:
Tested in chromium, all modes work flawlessly.

src/render/shaders/glsl/tonemap.glsl
```diff
@@ -54,11 +54,24 @@ vec4 tonemap(vec4 color, mat3 dstXYZ, float maxLuminance, float dstMaxLuminance,
     float E         = pow(clamp(ICtCp[0], 0.0, 1.0), PQ_INV_M2);
     float luminance = pow((max(E - PQ_C1, 0.0)) / (PQ_C2 - PQ_C3 * E), PQ_INV_M1) * HDR_MAX_LUMINANCE;
 
-    float luminanceRatio = max(luminance / dstRefLuminance, 0.0);
-    float srcScale       = maxLuminance / dstRefLuminance;
-    float dstScale       = dstMaxLuminance / dstRefLuminance;
-    float v              = (dstScale * (1.0 + srcScale) - srcScale) / pow(srcScale, 2.0);
-    float newLuminance   = (luminanceRatio * (1.0 + luminanceRatio * v) / (1.0 + luminanceRatio)) * dstRefLuminance;
+    float newLuminance;
+    if (tonemapMode == 3) {
```
vaxerski: not better as a shader prop to avoid a branch?

UjinT34: this one is based on a uniform and shouldn't cause any issues since all units execute with the same value. might be an issue with some older hw with stupid shader compilers/executors bit those are unlikely to use hdr anyway.  `luminance > knee` is much worse because it's per pixel.
it can be converted to a prop if it causes a measurable slowdown for somebody. I just don't see a point in doing so rn without benchmarks

vaxerski: even if its a uniform it causes a gpu branch. That was the entire point of the SH_FEAT_* stuff in the first place, it was if() guards with uniforms.

vaxerski (approved):
neato. Code is fine, but I can't speak for functionality thats down to whatever @LionHeartP wanted.

wiki mr needed

fxzzi:
Werks 4 me thanks

PR #14909, opened 2026-05-31
replace O_CLOEXEC with FD_CLOEXEC for file descriptor flags

ghost:
small fixes, replaced O_CLOEXEC with FD_CLOEXEC in the codebase

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+CompileFlags:
+  CompilationDatabase: build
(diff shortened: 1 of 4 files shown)


ItsOhen:
Fair enough.

.clangd
```diff
@@ -0,0 +1,2 @@
+CompileFlags:
```
vaxerski: what is this file doing

ghost: its for my clang config its useless u can delete i forgot it mb

vaxerski (approved):

PR #14914, opened 2026-05-31
config/actions: fix null derefs in pin dispatcher

DxCx:

 `layoutTarget()` call. The two new derefs aren't covered by the existing `PMONITOR` check — any of
  `PMONITOR->m_activeWorkspace`, `m_activeWorkspace->m_space`, or `window->layoutTarget()` being null crashes the
  compositor on `hyprctl dispatch pin`.

  I hit this with pyprland on a laptop: lid close disables the internal monitor, the pinned floating window's `m_monitor`
  weak_ptr still resolves (the `CMonitor` stays alive in `m_realMonitors`) but `m_activeWorkspace` got reset to null in
  `onDisconnect`. pyprland's next pin toggle then dereferences null and aborts the compositor.

from stack trace:
```
  Layout::CWindowTarget::assignToSpace   <- segv
  Config::Actions::pinWindow
  pin
  CDispatcherTranslator::run
  CHyprCtl::getReply
```

I've run with the patched  version locally and i do not observe crashes anymore.
in addition i've added a test which trigger the crash before the fix. and the the whole test suite runs succesfully after it (including the new test added for this)

from what i understand it should be, just require review...
as mentioned above, it was verified locally via `nix build .#checks.x86_64-linux.tests`: `issue14134` passes;

Fixes #14134

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -588,6 +588,27 @@ TEST_CASE(issue14038) {
     // this should not crash hyprland. If we are alive, we good.
 }
 
+TEST_CASE(issue14134) {
+    OK(getFromSocket("/output create headless HEADLESS-4"));
+    OK(getFromSocket("/dispatch hl.dsp.focus({ monitor = 'HEADLESS-4' })"));
+
+    if (!spawnKitty("kitty_14134"))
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
thanks!

PR #14918, opened 2026-06-01
Fix screenshader with fp16

### LINKED ISSUE #10558: FP16 support

UjinT34:
Fixes `decoration:screen_shader` with fp16. Closes #10558

Ready

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -813,12 +813,23 @@ void CHyprOpenGLImpl::end() {
         g_pHyprRenderer->bindFB(g_pHyprRenderer->m_renderData.outFB);
         blend(false);
 
-        const auto PRIMITIVE_BLOCKED = m_finalScreenShader->program() >= 1 || g_pHyprRenderer->m_crashingInProgress ||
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks!

PR #14923, opened 2026-06-01
renderer: cache workspace fullscreen windows

ghost:
cache each workspace's fullscreen window as a weak reference and keep it synchronized across fullscreen transitions, workspace moves, unmaps, and recovery paths
make fullscreen window lookup O(1) while preserving layout-managed fullscreen handling
simplify the fullscreen renderer path: avoid temporary smart-pointer vectors, use the cached fullscreen window in the common path, and keep the broader scan only while other workspaces animate avoid smart-pointer copies across hot renderer helpers and blur predicates by passing handles as `const&`
cache repeated monitor workspace IDs locally in `shouldRenderWindow`


vaxerski:
@codex review please

vaxerski:
this is iffy fullscreen state is quite flaky I dont like it in general but ugh

PR #14927, opened 2026-06-01
monitor: retry transient mode selection failures

ghost:
## Summary
- retry monitor rule application up to three times when every tested output mode is rejected
- delay retries by one second so transient driver failures can recover without a tight loop
- cancel pending retries as soon as mode selection succeeds or the monitor disconnects

## Why
`f7526d6b` added the first part of #10678 by refusing to render monitors with an invalid pixel size. A transient driver rejection can still leave the monitor at `0x0` for the rest of the session because mode selection is never attempted again.

This adds bounded recovery while preserving the existing renderer guard.

Addresses #10678.

## Validation
- `git diff --check upstream/main...HEAD`
- `cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON`
- `cmake --build build --target Hyprland hyprland_gtests -j2`
- `ctest --test-dir build --output-on-failure` (`205/205` passed)

### DIFF
--- a/src/helpers/Monitor.cpp
+++ b/src/helpers/Monitor.cpp
@@ -395,6 +395,7 @@ void CMonitor::onDisconnect(bool destroy) {
     }};
 
     m_frameScheduler.reset();
+    clearModeRetry();
 
     if (!m_enabled || g_pCompositor->m_isShuttingDown)
         return;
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
thanks

PR #14932, opened 2026-06-01
renderer: reduce per-frame heap allocations

ghost:
## Description

Reduce renderer allocations identified with `heaptrack` against disposable nested Hyprland instances:

- Keep shader VBO defaults resident and only orphan/upload vertices for custom UV draws or the transition back to defaults.
- Store render-pass bookkeeping records inline so `CRenderPass::clear()` can reuse vector capacity without allocating one wrapper per element.
- Use pixman rectangle counts and views instead of materializing temporary `CRegion::getRects()` vectors in render-pass simplification, damage-ring coalescing, and cursor SHM updates.
- Defer copying pass damage until after rejecting elements outside the damaged area.
- Use stack-backed roots for surface and popup breadth-first traversal, reserve descendant vectors only when descendants exist, and pre-size the popup child recheck copy.
- Bind read-only render-element payloads by reference instead of copying regions and vectors during draws.
- Keep the common window-decoration update snapshot inline while retaining a vector fallback for plugin-provided decoration sets larger than the built-in set.
- Compare cached shader vectors by reference instead of copying both vectors before every uniform equality check.

## Heaptrack results

The same nested compositor workload was replayed before and after the initial patch with XWayland disabled, one nested `kitty` surface, and 150 identical `hyprctl --batch` iterations.

| Metric | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Allocation calls | 519,055 | 420,045 | -99,010 (-19.08%) |
| Temporary allocations | 112,499 | 68,527 | -43,972 (-39.09%) |

A second matched replay against that result measured the damage-ring rectangle-view change:

| Metric | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Allocation calls | 420,186 | 416,906 | -3,280 |
| Temporary allocations | 68,374 | 66,657 | -1,717 |

Targeted stack comparisons also showed:

- `SPassElementData` wrapper allocation: -14,868 calls.
- `CRegion::getRects()` from the pass simplifier: -2,830 temporary allocations.
- Deferred damage copy for rejected elements: -4,620 `pixman_region32_copy` calls in a matched follow-up capture.
- `CRegion::getRects()` from `CDamageRing::getBufferDamage()`: -1,355 calls.
- Popup breadth-first helper allocator stack after switching to stack roots and lazy reserve: -1,303 temporary allocations.
- Surface breadth-first helper allocator stack after switching to stack roots and lazy reserve: -1,617 calls.

I then replayed a richer nested workload with five long-lived `kitty` windows, a child toplevel, a `zenity` dialog, periodic short-lived windows, and 240 iterations of focus, floating, resize, move, workspace, fullscreen, swap, and special-workspace churn. Focused comparisons showed:

- Read-only render-element payload references: -172,036 allocation calls and -29,484 raw temporary allocations.
- Inline common-path window-decoration snapshots: -94,796 allocation calls in the `CWindow::updateWindowDecos()` snapshot stack.
- Cached shader-vector reference: -36,006 allocation calls in `CShader::setUniformfv()`.
- Reference parameters in `compareFloat()`: the remaining `setUniformfv()` vector-copy temporary stacks disappeared from the follow-up captures.

The final extended replay ran the richer workload for 600 iterations. It recorded 3,509,128 allocation calls and 730,967 temporary allocations over 17.95 seconds, with no `CShader::setUniformfv()` or `compareFloat()` allocation stack remaining. The remaining dominant temporary sites were signal emission in `hyprutils`, workload-induced RE2 parsing from repeated `hyprctl focuswindow`, and region copies whose clipping behavior is intentionally preserved.

## Verification

- `clang-format -i src/render/OpenGL.cpp src/render/Shader.cpp src/render/Shader.hpp src/render/pass/Pass.cpp src/render/pass/Pass.hpp src/helpers/DamageRing.cpp src/protocols/core/Compositor.cpp src/protocols/core/Compositor.hpp src/desktop/view/Popup.cpp src/desktop/view/Popup.hpp src/desktop/view/Window.cpp src/render/ElementRenderer.cpp src/render/gl/GLElementRenderer.cpp`
- `git diff --check`
- `cmake --build build -j2`
- `ctest --test-dir build --output-on-failure` (205 tests passed)
- Nested Hyprland smoke/profile runs under `heaptrack`, including the richer 600-iteration capture

### DIFF
--- a/src/config/shared/actions/ConfigActions.cpp
+++ b/src/config/shared/actions/ConfigActions.cpp
@@ -976,7 +976,10 @@ static PHLWORKSPACE resolveWorkspaceForChange(const std::string& args) {
         return nullptr;
 
     const auto PCURRENTWORKSPACE = PMONITOR->m_activeWorkspace;
-    const bool EXPLICITPREVIOUS  = args.contains("previous");
+    if (!PCURRENTWORKSPACE)
+        return nullptr;
+
+    const bool EXPLICITPREVIOUS = args.contains("previous");
 
     // handle "previous" workspace
     if (args.starts_with("previous")) {
--- a/src/config/lua/ConfigManager.cpp
+++ b/src/config/lua/ConfigManager.cpp
@@ -195,6 +195,17 @@ CConfigManager::CConfigManager() : m_mainConfigPath(Supplementary::Jeremy::getMa
     });
 }
 
+CConfigManager::~CConfigManager() {
+    m_eventHandler.reset();
+
+    cleanTimers();
+    clearLuaLayoutProviders();
+    clearHeldLuaRefs();
+
+    if (m_lua && m_ownsLuaState)
+        lua_close(m_lua);
+}
+
 CConfigManager* CConfigManager::fromLuaState(lua_State* L) {
     if (!L)
         return nullptr;
@@ -275,12 +286,13 @@ void CConfigManager::reinitLuaState() {
     cleanTimers();
     clearLuaLayoutProviders();
 
-    if (m_lua) {
+    if (m_lua && m_ownsLuaState) {
         lua_close(m_lua);
-        m_lua = nullptr;
     }
+    m_lua = nullptr;
 
-    m_lua = luaL_newstate();
+    m_lua          = luaL_newstate();
+    m_ownsLuaState = true;
     luaL_openlibs(m_lua);
 
     lua_getglobal(m_lua, "debug");
--- a/src/config/lua/ConfigManager.hpp
+++ b/src/config/lua/ConfigManager.hpp
@@ -52,6 +52,7 @@ namespace Config::Lua {
     class CConfigManager : public Config::IConfigManager {
       public:
         CConfigManager();
+        virtual ~CConfigManager() override;
 
         virtual eConfigManagerType               type() override;
 
@@ -157,7 +158,8 @@ namespace Config::Lua {
 
         static void                                  watchdogHook(lua_State* L, lua_Debug* ar);
 
-        lua_State*                                   m_lua = nullptr;
+        lua_State*                                   m_lua          = nullptr;
+        bool                                         m_ownsLuaState = false;
 
         bool                                         m_lastConfigVerificationWasSuccessful = true;
         bool                                         m_isFirstLaunch                       = true;
--- a/src/desktop/view/Popup.cpp
+++ b/src/desktop/view/Popup.cpp
@@ -12,6 +12,7 @@
 #include "../../managers/eventLoop/EventLoopManager.hpp"
 #include "../../render/Renderer.hpp"
 #include "../../render/OpenGL.hpp"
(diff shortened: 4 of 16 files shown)


ghost:
Il see if I can continue to reduce the allocations to try to improve hyprland performance overall

vaxerski:
so you want me to review this or wait?

ghost:
> so you want me to review this or wait?

Review Il make it a new Pr, it’s better bcs it keeps them short

vaxerski (approved):

aruhier:
Thanks! Spring animations for workspace changes for example were quite laggy on my laptop when plugged to a 4K screen, now things are way smoother!

ghost:
> Thanks! Spring animations for workspace changes for example were quite laggy on my laptop when plugged to a 4K screen, now things are way smoother!

Proud to help, im gonna continue this types of changes to see if I can fix my performance issues to

gulafaran:
> > Thanks! Spring animations for workspace changes for example were quite laggy on my laptop when plugged to a 4K screen, now things are way smoother!
> 
> Proud to help, im gonna continue this types of changes to see if I can fix my performance issues to

depends on what perf issues you have, but we have a presentation issue and a general frame callback issue. if you want to look into it. currently at something like this https://github.com/hyprwm/Hyprland/commit/0994aa881c976b9f22e5051a972cfa17e4ae9ba9 should make presentation callbacks more in line of the protocol but im just overloaded with 2 kids and summer prep boat etc. 

the frame callback issue is both being sent a bit to aggressive and specificly this part "A server should avoid signaling the frame callbacks if the surface is not visible in any way, e.g. the surface is off-screen, or completely obscured by other opaque surfaces." but it easily breaks firefox rendering that expects a frame callback for vsync buisness.

gulafaran:
there is also the buffer fence deadline that gnome/kwin currently does that i have some prototype for 
[drm2.patch](https://github.com/user-attachments/files/28661203/drm2.patch)

essentially a hint for the gpu driver to schedule and upclock it to reach the targeted deadline.

src/config/shared/actions/ConfigActions.cpp
```diff
@@ -976,7 +976,10 @@ static PHLWORKSPACE resolveWorkspaceForChange(const std::string& args) {
         return nullptr;
 
     const auto PCURRENTWORKSPACE = PMONITOR->m_activeWorkspace;
-    const bool EXPLICITPREVIOUS  = args.contains("previous");
+    if (!PCURRENTWORKSPACE)
```
tomben13, line 979: Hey @NotPppp1116 ,
I'm trying to rebase my PR and I have a conflict with this. Could you explain what led you to make this change ?

ghost: its a null check if a focused monitor exists but has no active workspace we could crash.

PR #14944, opened 2026-06-02
jeremy: fix auto-generating hyprlang instead of lua config file by default

LionHeartP:
Makes it so on empty config start, hyprland.lua is generated instead of hyprland.conf

It's a single word. Probably an oversight from the massive lua PR.

Ready

### DIFF
--- a/src/config/supplementary/jeremy/Jeremy.cpp
+++ b/src/config/supplementary/jeremy/Jeremy.cpp
@@ -40,7 +40,7 @@ std::expected<SConfigStateReply, std::string> Jeremy::getMainConfigPath() {
         else if (CONF_PATHS.first.has_value())
             return SConfigStateReply{.path = CONF_PATHS.first.value(), .type = CONFIG_TYPE_REGULAR};
         else if (LUA_PATHS.second.has_value()) {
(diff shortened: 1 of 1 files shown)


Dregu:
Well at least `rm -rf .config/hypr && Hyprland` generates a lua config now. However:

- `hyprland-welcome` is completely broken and confusing since it hasn't been updated for lua at all
- `--safe-mode` generates a .conf, hitting "load config" then loads the lua file as hyprlang, which of course leaves the compost completely unusable

LionHeartP:
> Well at least `rm -rf .config/hypr && Hyprland` generates a lua config now. However:
> 
> * `hyprland-welcome` is completely broken and confusing since it hasn't been updated for lua at all
> * `--safe-mode` generates a .conf, hitting "load config" then loads the lua file as hyprlang, which of course leaves the compost completely unusable

No 1. needs to be fixed in guiutils, but no2 should be good now. Another 3 letter change.

Dregu:
Yeah safe-mode, real crash and everything seems to use lua now.

At the very least welcomemanager should also be disabled if not fixed before autogenerated lua can be *released*, but vaxry can decide that.

vaxerski:
welcome support is simple, I just forgot to do it, if anyone wants to do it before I finish moving (probably late june) then be my guest

vaxerski (approved):

PR #14945, opened 2026-06-02
Optimize mapping rule properties to strings and engines

danielkrupinski:
### Before
- `MATCH_PROP_STRINGS` allocated heap memory
- `RULE_ENGINES` allocated heap memory
- `Rule::allMatchPropStrings()` allocated heap memory
- 3 copies of prop strings existed in memory - string literals in the binary file and 2 copies in the heap memory
- Linear search in `Rule::matchPropFromString()`

### After
- No heap memory allocations
- Only 1 instance of prop strings in memory
- Binary search in `Rule::matchPropFromString()`


vaxerski (approved):
thanks!

PR #14952, opened 2026-06-03
gestures/scrolling: add move_natural to invert scroll_move direction

cmihail:
Adds a new boolean option gestures:scrolling:move_natural (default false) that inverts the scroll_move trackpad gesture so the tape/content tracks the fingers, matching natural-scrolling behavior found elsewhere.

I tried using negative scale but did not work on my laptop.

It fixes the following problem: https://github.com/hyprwm/Hyprland/discussions/14723

 No automated test added. The hyprtester gesture simulator only sends modifier-less swipes, and every
unmodified finger-count/axis combination in the shared test.lua is already bound to non-scroll actions
 that the existing gestures test depends on.

### DIFF
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -410,6 +410,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
         MS<Int>("gestures:close_max_timeout", "Timeout for closing windows with the close gesture, in ms.", 1000, {.min = 10, .max = 2000}),
         MS<Bool>("gestures:scrolling:move_snap_to_grid", "When releasing the scroll move gesture, whether it shoud try to snap to the grid.", true),
         MS<Bool>("gestures:scrolling:move_snap_cursor", "When releasing the scroll move gesture, whether it shoud snap the cursor to the newly focused window.", true),
+        MS<Bool>("gestures:scrolling:move_natural", "When enabled, inverts the scroll move gesture direction so the content tracks the fingers (natural scrolling).", false),
         /*
          * group:
          */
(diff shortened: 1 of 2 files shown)


src/config/values/ConfigValues.cpp
```diff
@@ -410,6 +410,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
         MS<Int>("gestures:close_max_timeout", "Timeout for closing windows with the close gesture, in ms.", 1000, {.min = 10, .max = 2000}),
         MS<Bool>("gestures:scrolling:move_snap_to_grid", "When releasing the scroll move gesture, whether it shoud try to snap to the grid.", true),
         MS<Bool>("gestures:scrolling:move_snap_cursor", "When releasing the scroll move gesture, whether it shoud snap the cursor to the newly focused window.", true),
+        MS<Bool>("gestures:scrolling:move_natural", "When enabled, inverts the scroll move gesture direction so the content tracks the fingers (natural scrolling).", false),
```
cmihail, line 413: If this looks good I will create a PR to add the new config value to the hyprland docs.

vaxerski:
why not use a -1 delta scale

cmihail:
> why not use a -1 delta scale

I tried that as mentioned at https://github.com/hyprwm/Hyprland/discussions/14723 and it was a no-op. Is it working for others? I could not get it working for natural scrolling.

vaxerski:
maybe it's clamped somewhere, I think delta -1 is just more universal. Or an invert: flag

rogervn:
@cmihail do you intend to make the changes for delta -1 or invert flag?

cmihail:
> @cmihail do you intend to make the changes for delta -1 or invert flag?

I am low on time and the fix here works for me in my local build. So even if not merged I do not plan to focus on this at the time.

rogervn:
> > @cmihail do you intend to make the changes for delta -1 or invert flag?
> 
> I am low on time and the fix here works for me in my local build. So even if not merged I do not plan to focus on this at the time.

No worry, I've published another PR as this is important for me as well, it's quite similar, but it applies the flag name change suggested. I've also tested so it should be the same. Hopefully we can land either.

PR #14853, opened 2026-05-28
Fix Keyboard focus lost when cursor crosses layer-shell panels

ar21-jh:
#### Describe your PR, what does it fix/add?
Keyboard wont refocus when cursor crosses layer-shell panels with follow_mouse != 1.
it fixes https://github.com/hyprwm/Hyprland/discussions/14285, https://github.com/hyprwm/Hyprland/discussions/14730
and i can reproduce using DankMaterialShell when bar pop up opened then closed, the keyboard wont refocus

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)
Already tested and i use AI to fix

#### Is it ready for merging, or does it need work?

ready

PR #14190, opened 2026-04-27
layout/scrolling: handle fullscreen manually

vaxerski:
Makes scroll handle fullscreen by itself, allowing for fs scrolls.

TODOs:
- [x] maximize should just expel and set width to 1
- [x] testing

please test this I am not a scroll guy so I did some basic testing only

Fixes https://github.com/hyprwm/Hyprland/issues/13497

PR #14953, opened 2026-06-03
focus: don't skip window refocus when surface focus is cleared

chenxiex:

This pr is inspired by #14853 and should fix the same things:
> Keyboard wont refocus when cursor crosses layer-shell panels with follow_mouse != 1.
it fixes https://github.com/hyprwm/Hyprland/discussions/14285, https://github.com/hyprwm/Hyprland/discussions/14730
and i can reproduce using DankMaterialShell when bar pop up opened then closed, the keyboard wont refocus

But I implement the fix in a different way. I try to remove the breaking codes instead of writing new fixing code. More specifically, this pr removes the same-window/same-surface early return added in #14190 from `rawWindowFocus()`.

The guard conflates `surface == nullptr` as an optional argument meaning "use the window's main surface" with `m_focusSurface == nullptr` meaning "no focused surface". As a result, a window refocus can return before restoring seat keyboard focus.

I also add a new test to cover this scene. It basically simulates the following steps:
1. Focus a normal window.
2. Clear surface focus while leaving `m_focusWindow` intact, e.g. layer-shell/unmap path.
3. Soft-focus the same window again.
4. `rawWindowFocus()` returns early and seat keyboard focus remains null.

I use AI to fix, mainly for writing tests. This pr passed all the ci tests. However, this may break the fix in #14190.

It needs work. #14190 also fixed/mitigated scrolling fullscreen focus/cursor-lock issues, so I’d like confirmation that this guard was not required for that behavior.

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -415,6 +415,11 @@ static SDispatchResult nullfocus(std::string in) {
     return {};
 }
 
+static SDispatchResult clearSurfaceFocus(std::string in) {
+    Desktop::focusState()->rawSurfaceFocus(nullptr);
+    return {};
+}
+
 static Desktop::Rule::CWindowRuleEffectContainer::storageType windowRuleIDX = 0;
 
 //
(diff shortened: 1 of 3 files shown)


anayoA:
This is still broken for me :(

chenxiex:
Fixed on dms side. AvengeMedia/DankMaterialShell#2655

qappell:
@chenxiex Can you keep this open, since I use Noctalia v4 Legacy and occasionally other QS shells that this PR fixes focus issue for? https://github.com/hyprwm/Hyprland/discussions/14285

myamusashi:
I think we need to reopen this, I have the same issue with my own shell

PR #14971, opened 2026-06-04
event/bus: add specialActive event to workspace

vaxerski:
allows for special workspace state tracking


MicaelLoberg:
I have a PR that does something very similar
https://github.com/hyprwm/Hyprland/pull/14802

Difference is that I have 2 events, one for special opened and one for closed so I don't have to do null checks in lua to figure out what is happening

vaxerski:
I believe this one makes more sense, it relays precisely what happened.

I don't see the problem in doing `if ws ~= nil`

Dregu:
At least this is more versatile than simply extending workspace.active to cover specials and that wouldn't have even made sense if get_active_workspace doesn't even return specials.

Doesn't really answer which special workspace was closed though. I don't know if that is useful information.

vaxerski:
it does, because it still has the monitor param. If you keep track of what is open where, you can know what closed.

Dregu:
Yeah well having to keep track of it yourself is like the opposite of what I meant. But at least it's not hard and like I said might not even be useful.

PR #14972, opened 2026-06-04
Fix filename string not being computed at compile time in RASSERT and GLCALL macros

danielkrupinski:
`constexpr` on functions and lambdas means the function can be evaluated at compile time but it does not have to. `consteval` forces compile time evaluation.


vaxerski (approved):
thanks!

PR #14979, opened 2026-06-05
Config/Prop Refresh: Add a prop refresh event, add granular reasons for prop refresh, remove old hack for stopping prop refresh from moving scrolling viewport

erstarr:

1. It reverts https://github.com/hyprwm/Hyprland/pull/14594:

In that PR, I sought to address the issue of prop refresh indiscriminately moving the scrolling view regardless of why it was fired. However, some causes of prop refresh should indeed move the scrolling viewport and to add scrolling specific code as exceptions to lua dispatch/config/workspace rule handlers is untidy and not sustainable.

2. To this end, I wish to take a different approach: The existence of the scroll inhibitor allows the prevention of viewport moves; the problem is the fact that the prop refresh is fired after the event ends and there is no good way to catch it and uninhibit scrolling. Even if we catch it using another event that it incidentally causes the firing of; the reason why that prop refresh is fired isn't available to us so we don't know if this prop refresh is the one we want to uninhibit on; or if the event was fired not as a result of a prop refresh at all.

For this, I will:

- Add a new event which will fire when a prop refresh happens
- Add a new enum member to `ePropRefreshProp`

Using these, save the fact that a prop refresh was enqueued because of a workspace rule change, and the reason for why a prop refresh is executed will be passed to the user as an `int` which they can capture using the lua event from the lua config.

The user can now tell exactly why the prop refresh was executed using that integer and uninhibit scrolling.

Note: In this PR, I will only introduce 1 new enum member. It is what I personally need to make my scripts work. Users can add new ones for their needs; and I can add them here if people ask me to in the 'comments'.

WIP - but after it is done, it will not break anything.

TODO:

- [ ] Add new event

- [ ] Add new prop refresh reason

### DIFF
--- a/hyprtester/src/tests/main/scroll.cpp
+++ b/hyprtester/src/tests/main/scroll.cpp
@@ -1023,55 +1023,6 @@ TEST_CASE(testScrollingViewBehaviourMoveFocusInGroupFollowFocusTrue) {
     }
 }
 
-TEST_CASE(testScrollingViewBehaviourScheduledPropRefresh) {
-
-    /*
-     Scheduled prop refresh must not move scrolling viewport.
-     The reason a prop refresh was queued is not saved, therefore it is not possible to clearly tell when and when not to move scrolling viewport
-     In this test, we test this by setting a workspace rule, which schedules a prop refresh
-     --------------------------------------------------------------------------------------------------------------------------------------
-    */
-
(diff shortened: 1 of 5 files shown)


erstarr:
https://github.com/hyprwm/Hyprland/pull/14990

PR #14981, opened 2026-06-05
gestures: configure client to final size on fullscreen/float gesture completion

### LINKED ISSUE #14978: gestures: fullscreen/float gesture actions resize the window but never reconfigure the client (client keeps rendering at the old size)
### Regression?

No (the trackpad gesture actions feature is relatively new)

### System Info

```
Hyprland 0.55.2 built from branch at commit 39d7e209c79d451efab1b21151d5938289da838d clean (version: bump to 0.55.2).
Date: Sat May 16 14:59:19 2026
Tag: v0.55.2, commits: 7319
```

- Hardware: Apple MacBook Air **M1 (aarch64/ARM)**
- Distro: Fedora Asahi Remix
- Built entirely from source on the machine itself (no prebuilt packages)

### Description

When a trackpad gesture is bound to the `fullscreen` (or `maximize`) **or `float`** action, e.g.:

```lua
hl.gesture({ fingers = 3, direction = "up", mods = "SUPER", scale = 1.5, action = "fullscreen" })
```

…Hyprland resizes the window *box* to fullscreen, but **the client is never reconfigured to the new size**. The client keeps rendering its old (windowed) buffer, which Hyprland then stretches to fill the fullscreen box. `kitty`, for example, stays at its original cell size while the surface is scaled up.

Crucially, the **fullscreen state itself is correct** — `hyprctl clients` reports both `fullscreen: 2` and `fullscreenClient: 2`. So this is **not** a state-sync / `sync_fullscreen` issue. It is purely a *window-size configure* problem.

Doing the same thing via a keybind/dispatcher (`fullscreen`, `togglefloating`, …) works perfectly. Only the **gesture** path is affected. The same symptom happens with `action = "float"`. Removing `scale = 1.5` does not help.

### Root cause (analysis)

The client only receives a new size through `CWindow::sendWindowSize()`, which reports `m_realSize->goal()`. That call is queued **deferred** (`g_pEventLoopManager->doLater`) from the `m_realSize` *begin* callback, which only fires when a size **animation starts** (`src/desktop/view/Window.cpp`, `setCallbackOnBegin`). The per-frame `setUpdateCallback` does **not** send size to the client.

The gesture fights this flow:

1. `begin()` toggles fullscreen via `setWindowFullscreenInternal(...)`. The layout starts a size animation to the fullscreen geometry → the begin-callback queues a **deferred** `sendWindowSize()`. `m_realSize->goal()` == fullscreen at this point. (`src/managers/input/trackpad/gestures/FullscreenGesture.cpp` `begin()`)
2. `update()` calls `m_realSize->setValueAndWarp(lerp(...))`. This is a *warp* (no animation begin → no callback), but it **overwrites `m_realSize->goal()`** with an intermediate value.
3. The deferred `sendWindowSize()` from step 1 runs *after* an `update()` has already warped the goal → the client gets configured to an **intermediate** size.
4. `end()` does `*m_realSize = m_sizeTo`. If the last `update()` already warped the goal to ≈ `m_sizeTo`, this assignment is a no-op → **no animation, no begin-callback, no new `sendWindowSize`** → the client is never corrected and stays at the wrong size.

`scale = 1.5` makes `FADEPERCENT` reach `1.0` faster, so `update()` is more likely to warp the goal to the final size before the deferred send runs — making the race lose more reliably. On aarch64/Asahi the `doLater` vs. input-event ordering differs from x86, which is likely why this reproduces so consistently there.

The same pattern (warping `m_realSize->goal()` in `update()`, no forced configure in `end()`) exists in:
- `src/managers/input/trackpad/gestures/FullscreenGesture.cpp`
- `src/managers/input/trackpad/gestures/FloatGesture.cpp`

A fix is to force `m_window->sendWindowSize(true)` once the gesture commits in `end()`, so the client is always reconfigured to the final size regardless of the animation/race.

### How to reproduce

1. On aarch64 (e.g. Asahi/M1), add to the config:
   ```lua
   hl.gesture({ fingers = 3, direction = "up", mods = "SUPER", scale = 1.5, action = "fullscreen" })
   ```
2. Open `kitty` (or any client that draws to its reported size).
3. Perform the 3-finger up gesture to fullscreen it.
4. Observe: the window box is fullscreen but the client buffer is stretched/blurry; `hyprctl clients` shows `fullscreen: 2` / `fullscreenClient: 2`.
5. Fullscreen via keybind instead → renders correctly. `action = "float"` shows the same bug.

### Additional context

Original forum report: https://forum.hypr.land/t/gestures-actions-are-not-functioning-as-expected-on-latest-v0-55-2/1668?u=tim

_n/a_

TimaDub:
## What

The `fullscreen` and `float` trackpad gesture actions resize the window box but never reconfigure the client to the final size, so the client keeps rendering at its old size and Hyprland stretches the stale buffer. The fullscreen state itself is correct (`hyprctl clients` shows `fullscreen: 2` / `fullscreenClient: 2`); only the size configure is missing.

## Why

The client only learns its size via `CWindow::sendWindowSize()`, which is queued deferred from the `m_realSize` begin-callback when a size animation starts. The gestures repeatedly `setValueAndWarp()` `m_realSize->goal()` in `update()` — a warp, so it overwrites the goal without starting an animation or scheduling a new send. The single deferred send queued in `begin()` then runs after the goal was already warped to an intermediate value, and on completion the assignment in `end()` is often a no-op (the goal was already warped there by the last `update()`), so no corrected size is ever sent.

Forcing `sendWindowSize(true)` once the gesture commits fixes it, mirroring what `CCloseTrackpadGesture` already does and what the normal layout path (`CWindowTarget::setPositionGlobal`) does.

Affected actions are only `fullscreen` and `float`; `resize` re-sends each frame, `move` only changes position, `close` already forces a send.

Closes #14978

## Test plan

Tested on aarch64 (Fedora Asahi Remix, Apple M1) only:

- [x] `action = "fullscreen"` gesture on kitty → crisp fullscreen, client at full size
- [x] `action = "float"` gesture → client redraws at the new size, no stretching
- [x] keybind `fullscreen` / `togglefloating` unaffected
- [x] `resize` / `move` / `close` gestures unaffected

Not yet tested on x86_64.

---
Implemented with AI assistance (Claude Code); changes reviewed and tested by me.

### DIFF
--- a/src/managers/input/trackpad/gestures/FloatGesture.cpp
+++ b/src/managers/input/trackpad/gestures/FloatGesture.cpp
@@ -85,4 +85,9 @@ void CFloatTrackpadGesture::end(const ITrackpadGesture::STrackpadGestureEnd& e) 
 
     *m_window->m_realPosition = m_posTo;
     *m_window->m_realSize     = m_sizeTo;
+
+    // the gesture warps m_realSize->goal() around during update(), which races the deferred
+    // sendWindowSize() queued when the size animation began and can leave the client configured to
+    // an intermediate size. force a configure to the final size so the client matches its box.
+    m_window->sendWindowSize(true);
 }
--- a/src/managers/input/trackpad/gestures/FullscreenGesture.cpp
+++ b/src/managers/input/trackpad/gestures/FullscreenGesture.cpp
@@ -84,6 +84,12 @@ void CFullscreenTrackpadGesture::end(const ITrackpadGesture::STrackpadGestureEnd
 
     *m_window->m_realPosition = m_posTo;
     *m_window->m_realSize     = m_sizeTo;
+
+    // the gesture warps m_realSize->goal() around during update(), which races the deferred
+    // sendWindowSize() queued when fullscreen began and can leave the client configured to an
+    // intermediate size. force a configure to the final size so the client matches its box.
+    m_window->sendWindowSize(true);
+
     g_pDesktopAnimationManager->overrideFullscreenFadeAmount(m_window->m_workspace, m_originalMode == FSMODE_NONE ? 0.F : 1.F);
 }
 


vaxerski (approved):
tnx

alba4k:
finally :) this has been broken for almost a year

fwis, I think it would be better to resize the app __before__ the gesture, i.e. at the start of the resize (at least for fullscreen, where the size is already known) and not at the end

TimaDub:
Ok, I will try to improve it a bit. It was my first contribution btw, so I wasn't sure how it works and just did it the way it works.

alba4k:
I think it's still great to finally have this fixed! it's just a cosmetic change for situations like this:

immagine

PR #14984, opened 2026-06-05
Add `stableid:` window selector

LRitzdorf:
A natively-supported selector for grabbing windows by their "stable ID" would be nice for interoperability with other programs, especially ones that speak the `ext-foreign-toplevel-list` protocol (which [exposes these IDs](https://wayland.app/protocols/ext-foreign-toplevel-list-v1#ext_foreign_toplevel_handle_v1:event:identifier)).

With the introduction of Lua, it was already possible to act on windows by stable ID with an invocation like:
```sh
hyprctl eval 'for _,w in ipairs(hl.get_windows()) do if w.stable_id == WHATEVER then hl.dispatch(hl.dsp.window.move({ window = w, workspace = "+0" })) return end end'
```
...but that's pretty clunky. Instead, we could have:
```sh
hyprctl dispatch 'hl.dsp.window.move({ window = "stableid:WHATEVER", workspace = "+0" })'
```

Doing this matching inside the compositor itself is straightforward enough, and would let clients that speak the ext toplevel protocol more directly do useful stuff with the stable IDs they're given. Also, maintenance cost should be basically zero.

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -631,6 +631,34 @@ TEST_CASE(specialFloatRecenters) {
     ASSERT(Tests::windowCount(), 0);
 }
 
+TEST_CASE(exactWindowSelectors) {
+    if (!spawnKitty("kitty_A"))
+        FAIL_TEST("Could not spawn kitty");
+    if (!spawnKitty("kitty_B"))
+        FAIL_TEST("Could not spawn kitty");
+
+    getFromSocket("/dispatch hl.dsp.focus({ window = 'class:kitty_B' })");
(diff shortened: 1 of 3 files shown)


vaxerski (dismissed):
integration tests needed :)

LRitzdorf:
Ah okay, was wondering about that. I can test focusing via all three of the "exact" selectors (address, PID, stable ID); hopefully that makes sense? Address and PID selectors are used in a bunch of other tests, but not explicitly tested for on their own yet.

vaxerski (approved):
lgtm, just a wiki mr needed

LRitzdorf:
(docs ready, this should be good to go whenever)

PR #14986, opened 2026-06-05
render: handle explicit sync fallback safely

ghost:
## Summary
- Flush submitted GL work before releasing async buffer refs when explicit sync manager creation fails, without blocking the compositor loop.
- Clear stale normal-render explicit fence state in that fallback path.
- Guard direct scanout and the frame scheduler against null or invalid sync managers.

## Review follow-up
- Replaced the blocking `glFinish()` fallback with `glFlush()` as requested.

## Testing
- `clang-format -i src/helpers/Monitor.cpp src/helpers/MonitorFrameScheduler.cpp src/render/GLRenderer.cpp src/render/OpenGL.cpp`
- `git diff --check --cached`
- `cmake --build build -j$(nproc)`
- `ctest --test-dir build --output-on-failure` (205/205 passed)


src/render/GLRenderer.cpp
```diff
@@ -148,9 +148,18 @@ void CHyprGLRenderer::endRender(const std::function<void()>& renderingDoneCallba
             PMONITOR->m_output->state->setExplicitInFence(PMONITOR->m_inFence.get());
         }
     } else {
-        Log::logger->log(Log::ERR, "renderer: Explicit sync failed, releasing resources");
+        Log::logger->log(Log::ERR, "renderer: Explicit sync failed, waiting before releasing resources");
 
-        m_usedAsyncBuffers.clear(); // release all buffer refs and hope implicit sync works
+        // Fence export failed after rendering. Without a fence, wait before
+        // dropping buffer refs that may still be used by the GPU.
+        glFinish();
```
vaxerski: this will cause massive lag, are we sure glFlush() is not enough to just signal an implicit sync point?

src/render/GLRenderer.cpp
```diff
@@ -148,9 +148,21 @@ void CHyprGLRenderer::endRender(const std::function<void()>& renderingDoneCallba
...
         }
     } else {
-        Log::logger->log(Log::ERR, "renderer: Explicit sync failed, releasing resources");
+        Log::logger->log(Log::ERR, "renderer: Explicit sync failed, falling back to implicit sync");
 
-        m_usedAsyncBuffers.clear(); // release all buffer refs and hope implicit sync works
+        // Match the implicit-sync fallback above: only drivers without usable
+        // implicit synchronization need the blocking wait.
+        if ((isNvidia() && *PNVIDIAANTIFLICKER) || isSoftware())
+            glFinish();
```
vaxerski: I don't think we should do it in 2026

vaxerski (approved):

PR #14912, opened 2026-05-31
seat: preserve pointer button state across pass focus changes

ghost:
#### Describe your PR, what does it fix/add?

Preserves pointer button state when pointer focus is temporarily redirected by pass-style focus handling, instead of forcing synthetic button releases on leave.

This keeps pointer button state consistent across the focus transition and updates the pointer-warp test expectations accordingly.

PR #14988, opened 2026-06-05
cursor: transform hardware cursors within content bounds

ghost:
### Description

Hardware cursor buffers can be larger than the visible cursor image, especially on NVIDIA where the CPU-backed cursor path uses the cursor plane maximum size. For transformed outputs, the CPU cursor renderer and hotspot transform were using the full backing buffer dimensions, which can place the visible cursor and hotspot at the far edge of a much larger transparent buffer.

This keeps the backing buffer size unchanged, but applies cursor transforms using the scaled visible cursor content bounds. That keeps the transformed image and hotspot local to the actual cursor content instead of the whole cursor plane.

Potentially addresses hyprwm/aquamarine#302.

### Testing

- `clang-format --dry-run --Werror src/managers/PointerManager.cpp`
- `git diff --check`
- `cmake --build ./build --target Hyprland -j$(nproc 2>/dev/null || getconf NPROCESSORS_CONF)`
- `ctest --test-dir build --output-on-failure` (205/205 passed)

I do not have a rotated NVIDIA output here, so runtime validation for the reported hardware setup is still needed.


vaxerski:
can someone affected test?

lichie567:
This *almost* works... In my case (monitor is rotated 270 dregrees) normally I cannot reach the bottom of the screen. This change let my cursor reach the bottom of the screen but the cursor stops when the bottom of the cursor image hits the bottom of the screen. Normally the pointer hotspot should be able to reach the very bottom pixel, which means most of the cursor should be able to go off screen by `cursorHeight - 1` pixels.

umbrageodotus:
try using this with @NotPppp1116's seat pr

lichie567:
@umbrageodotus i assume you mean this one: https://github.com/hyprwm/Hyprland/pull/14912? I tried it but no change.

umbrageodotus:
yeah then idfk NGL

vaxerski (approved):

vaxerski:
I should not have merged this

PR #14990, opened 2026-06-05
config/propRefresher: add capability to execute scheduled prop refresh immediately with lua helper function, add event for prop refresh

erstarr:

- Implement a lua helper function to execute a scheduled prop refresher immediately. If it's not executed immediately, it goes on to execute as scheduled

- Add an event for prop refresh execution. every time it is executed, this event is emitted. The information about if the event was executed as a result of "executed before schedule)" or "executed as scheduled" is passed to the used in the lua event.

This fixes the problem of not refreshing states when an action that requires it (workspace rule) but instead waits until the end of that function (i.e. all events of that function have already executed) to refresh states.

Example:

```lua

hl.bind(mainMod .. " + A", function ()

    local currentWorkspace = hl.get_active_workspace()

    hl.workspace_rule({
    workspace = currentWorkspace.name,
    gaps_in = 5,
    gaps_out = 5
    })


    hl.dispatch(hl.dsp.layout("focus right"))

end)
```

In this function, it is expected that the workspace rule will apply first, then we will focus on the column on the right (scrolling layout).

In reality, what happens is that the refresh event is queued, then we focus on the column on the right, and only then does the refresh event execute and apply the workspace rule. i.e: instead of workspaceRule -> focus right, what happened was focus right -> workspaceRule

practically: this makes the viewport move happen using the old gaps values, and with the updates gaps values the viewport move is not done properly (there's empty space on the right if the old gaps were < 5, and the column on the right isn't fully in view if gaps were > 5)

- This also makes https://github.com/hyprwm/Hyprland/pull/14594 redundant and undesirable, since that PR was a patch to try and fix unwanted scrolling viewport changes. With this change, it's trivial to inhibit scrolling correctly

No

WIP

TODO

- [x] some non-`return 0` condition for helper function.

- [x] Add event. Save info on if it was executed immediately or as scheduled.

- [x] Wiki MR for the helper func and event. Also explain how prop refresh scheduling works and possible consequences of it

### DIFF
--- a/hyprtester/src/tests/main/scroll.cpp
+++ b/hyprtester/src/tests/main/scroll.cpp
@@ -1027,55 +1027,6 @@ TEST_CASE(testScrollingViewBehaviourMoveFocusInGroupFollowFocusTrue) {
     }
 }
 
-TEST_CASE(testScrollingViewBehaviourScheduledPropRefresh) {
-
-    /*
-     Scheduled prop refresh must not move scrolling viewport.
-     The reason a prop refresh was queued is not saved, therefore it is not possible to clearly tell when and when not to move scrolling viewport
-     In this test, we test this by setting a workspace rule, which schedules a prop refresh
-     --------------------------------------------------------------------------------------------------------------------------------------
-    */
-
-    OK(getFromSocket("r/eval hl.config({ general = { layout = 'scrolling' } })"));
-
-    // ensure variables are correctly set for the test
-
-    OK(getFromSocket("/eval hl.config({scrolling = {follow_focus = false}})"));
-
-    if (!Tests::spawnKitty("a")) {
-        FAIL_TEST("Could not spawn kitty with win class `a`");
-        return;
-    }
-
-    OK(getFromSocket("/dispatch hl.dsp.layout('colresize 0.8')"));
-
-    if (!Tests::spawnKitty("b")) {
-        FAIL_TEST("Could not spawn kitty with win class `b`");
-        return;
-    }
-
-    // since follow_focus = false, viewport does not move
-    OK(getFromSocket("/dispatch hl.dsp.focus({ window = 'class:a' })"));
-
-    // setting a workspace rule queues a doLater() call in the Event Loop Manager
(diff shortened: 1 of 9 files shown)


vaxerski:
I am unsure about this. It's really inefficient for many cases...

erstarr:
> I am unsure about this. It's really inefficient for many cases...

Other than this, I can try to make a "execute pending prop refresh immediately" lua "convenience function":
-> if there's a pending refresh event in the event loop, execute that during the current event immediately and remove that entry from the event loop mgr

<br>


But i think after a workspace rule that updates gaps values (I use that as the example cuz I'm more familiar with it, but i'm sure this applies to other calls that causes propRefresh), every other line of code afterwards should now have the updated gaps values instead of these values only being updated after the whole lua function finishes executing. I would think it's a bug if this doesn't happen.

erstarr:
Work arounds i can think of rn is to split any function in two: one part until anything that procs the prop refresh, and the rest to happen after the prop refresh - after the updated values are applied. Catch the prop refresh with a new event, return the reason as int and use bitwise comparison in lua to detect if that's the one we want, and execute the rest of the function then.

Or to split the bind into press and release and put each half of the function there. 

These are exceptionally hacky

vaxerski:
I mean I understand where you are coming from but I do like the scheduled refreshing.

erstarr:
> I mean I understand where you are coming from but I do like the scheduled refreshing.

well if you insist. I'll add a helper function to immediately execute the pending refresh and a wiki MR to note how workspace rule application works (is scheduled, if you wanna apply now use this function). Does that work?

vaxerski:
that would be fine

erstarr:
Edited description for new goal.
Also adding the event.

vaxerski:
why are we removing a test

erstarr:
> why are we removing a test

was a test for https://github.com/hyprwm/Hyprland/pull/14594 which was reverted.

I adapted the test to test for the new helper function, but it really is not a good test to test the helper function; it's a good test to see if it's possible to inhibit workspace rule creation moving viewport, but not a good test of the helper function since the test is using scrolling specific behaviour.

You can decide if you want the new test to stay.

vaxerski (approved):
this is good

PR #14995, opened 2026-06-05
renderer: fix `cursor:zoom_rigid` being ignored with detached camera

barrettruth:

`cursor:zoom_rigid` is only honored in the non-detached zoom path. Basically, even with zoom rigid set, if I put my cursor near the edge of the screen, it would auto-scroll :/

The default detached camera now locks the camera instead of edge-panning

Unsure of exactly how to test it besides manually observing the behavior being fixed.

r4r.


vaxerski (approved):
tnx

PR #15002, opened 2026-06-06
config/windowrules: allow requiring multiple tags in window rules

### LINKED ISSUE #14970: Make tags more flexible for window rules
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/12076#discussioncomment-17142672

## Description

Allow requiring multiple tags for window rules. E.g. tag1 + tag2

marvelpokemaster:

Allow requiring multiple tags for window rules (e.g. `tag = tag1 + tag2`).
This is done by splitting the tag rule match string by `+`, trimming whitespace around each tag, and checking that all required tags are present. Also added Unit tests to verify multiple tag matching behavior and integration with negative matching.

No.

Ready.

Fixes #14970

### DIFF
--- a/src/config/lua/bindings/LuaBindingsConfigRules.cpp
+++ b/src/config/lua/bindings/LuaBindingsConfigRules.cpp
@@ -1112,8 +1112,21 @@ static int hlWindowRule(lua_State* L) {
                     matchVal = std::to_string(lua_tointeger(L, -1));
                 else if (lua_isstring(L, -1))
                     matchVal = lua_tostring(L, -1);
-                else {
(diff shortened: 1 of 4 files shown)


ItsOhen:
I'd rather have multiple `tag =` but maybe that's just me.

vaxerski:
considering we do lua, I think a better approach would be to use a lua table.

marvelpokemaster:
> considering we do lua, I think a better approach would be to use a lua table.

Yeah, I changed it to support a Lua table format. Tags can now be specified as:
`tag = {"tag1", "tag2"}`
which is converted to the existing multi-tag matcher format internally.

vaxerski:
you kinda silently break tags with a + in them. Pass a vector.

PR #15003, opened 2026-06-06
input: fix missing a pointer frame after absolute pointer motion

seven-mile:
Send a pointer frame after absolute pointer motion.

`onMouseMoved()` already sends a pointer frame after *relative* motion. `onMouseWarp()` can also send pointer motion to clients via `mouseMoveUnified()`, but did not flush it with `wl_pointer.frame`. This can leave Wayland clients waiting to apply hover state until another framed pointer event, such as button or axis input.

This affects virtual-pointer absolute motion users, such as Sunshine remote desktop cursor-position updates.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -165,6 +165,8 @@ void CInputManager::onMouseWarp(IPointer::SMotionAbsoluteEvent e) {
 
     m_lastInputTouch  = false;
     m_lastInputTablet = false;
+
+    g_pSeatManager->sendPointerFrame();
 }
 
 void CInputManager::simulateMouseMovement() {


UjinT34:
Needs testing with `cursor:no_break_fs_vrr = 1`. Some pointer frames might be missing, some might be unwanted when cursor events are skipped with this setting.

seven-mile:
@UjinT34 Hi~ Thanks for the quick context!

I took a look at `cursor:no_break_fs_vrr` and found that it controls the firing of `pageFlip->connector->output->events.frame`. I believe this is related to compositor rendering frames, which are asynchronous to pointer frames.

This patch does not include additional testing because it simply mirrors what the relative-motion code path has already been doing.

Please let me know if I'm missing something ; )

UjinT34:
Those pointer frames can trigger a content "update" that'll trigger a monitor frame. Not so obvious sequence of events that happens under specific conditions. Locked cursors + xwayland iirc. Or some other combo that's usually used with shooter games. Might not be relevant here if those never use `CInputManager::onMouseWarp`, not really sure what they do to keep the cursor at the center.

umbrageodotus:
Should be GTG because of what @seven-mile said.

Current behavior for relative positioning seems alright.

vaxerski (approved):

PR #15006, opened 2026-06-06
render: set main buffer fence deadlines

ghost:
## Summary
- add `render:main_buffer_deadline`, default off
- set `SYNC_IOC_SET_DEADLINE` on the main render buffer fence before commit
- track the estimated next vblank as monotonic nanoseconds only when the option is enabled
- avoid stale explicit fences and skip deadline computation until a usable fence exists

## Testing
- `clang-format --dry-run --Werror src/config/values/ConfigValues.cpp src/helpers/Drm.cpp src/helpers/Drm.hpp src/helpers/Monitor.cpp src/helpers/Monitor.hpp src/render/Renderer.cpp tests/helpers/Drm.cpp`
- `git diff --check`
- `cmake --build build -j2`
- `ctest --test-dir build --output-on-failure` (206/206)
- nested Hyprland workload with `kitty`, `zenity`, and `vkcube`
- `strace`: enabled run produced 1,734 successful `SYNC_IOC_SET_DEADLINE` calls and 0 failures; disabled run produced 0 deadline ioctls
- `perf`: no samples attributed to the deadline helpers
- `heaptrack`: no allocation stacks attributed to the deadline helpers
- `hotspot`: opened the captured `perf.data` without parser errors

## Docs
- Wiki PR: hyprwm/hyprland-wiki#1546

### DIFF
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -559,6 +559,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          * render:
          */
 
+        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer.", false),
         MS<Int>("render:direct_scanout", "Enables direct scanout.", 0, {.min = 0, .max = 2, .map = OptionMap{{"disable", 0}, {"enable", 1}, {"auto", 2}}}),
         MS<Bool>("render:expand_undersized_textures", "Whether to expand textures that have not yet resized to be larger.", true),
         MS<Bool>("render:xp_mode", "Disable back buffer and bottom layer rendering.", false),
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -559,7 +559,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          * render:
          */
 
-        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer.", true),
+        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer when supported.", false),
         MS<Int>("render:direct_scanout", "Enables direct scanout.", 0, {.min = 0, .max = 2, .map = OptionMap{{"disable", 0}, {"enable", 1}, {"auto", 2}}}),
         MS<Bool>("render:expand_undersized_textures", "Whether to expand textures that have not yet resized to be larger.", true),
         MS<Bool>("render:xp_mode", "Disable back buffer and bottom layer rendering.", false),
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -559,7 +559,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          * render:
          */
 
-        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer.", false),
+        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer.", true),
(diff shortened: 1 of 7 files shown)


gulafaran:
just a sidenote as to why i had this in the prototype when someone reviews this. 

`pMonitor->isMultiGPU() || pMonitor->m_vrrActive || pMonitor->m_tearingState.activelyTearing || !pMonitor->m_hasEstimatedNextVblank` 

multigpu we kinda need it asap because of blitting that wants to copy it as fast as possible and the blitted buffer can potentially deadline in AQ to the next vblank.
vrr/tearing it kinda fluctuates that im not sure one can reliably calculate next vblank.

as for `Time::steadyNow() + std::chrono::microseconds(100)` im not so sure all drivers gracefully handles the case if the the time is in the past and potentially just clocks up to max pstate for no reason so 100 microseconds added seemed like a minimum fair estimate to always be in the future or as close as possible to "now".

and im not so sure nvidia itself has any support for this deadline, so it mostly affects intel/amd.

otherwise ive locally had it running for quite some time.

src/config/values/ConfigValues.cpp
```diff
@@ -553,6 +553,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          * render:
          */
 
+        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer.", false),
```
vaxerski: why off by default?

gulafaran:
this needs a rebase, and yeah probably set the config option default true

src/config/values/ConfigValues.cpp
ghost: i wanted to be more conserveti so i had opt in but il change to default on

src/helpers/Monitor.cpp
```diff
@@ -130,13 +148,25 @@ void CMonitor::onConnect(bool noRule) {
...
+
+        if (m_pendingFrame && *PMAINBUFFERDEADLINE) {
+            const auto REFRESHINTERVAL = refreshIntervalNsFromPresentEvent(event, m_refreshRate);
+            if (REFRESHINTERVAL) {
+                m_estimatedNextVblankNs  = monotonicNs(ts ? *ts : mono) + *REFRESHINTERVAL;
+                m_hasEstimatedNextVblank = true;
+            } else
+                m_hasEstimatedNextVblank = false;
+        } else
+            m_hasEstimatedNextVblank = false;
```
gulafaran: doing a cleanup PR of scheduling frames i noticed m_pendingFrame is really wrong here, this needs a bit of rework. m_pendingFrame is only set if something scheduled a frame while one was already running. so this almost never was true.

in aq one more .frame() will emit after the presented signal, either the idleFrame or from the pf handler itself.
m_estimatedNextVblankNs probably should just be calculated here, and reset in endRender after commit or early out.

and on next .frame() if coming from the idleFrame callback no pageflip and present() has occured its going to be in a reset state and wont use the old calculated value and just deadline it "now"

gulafaran:
generally looks good to me, just needs some amd/intel testing i guess, so the power usage / gpu clocks isnt going vr00m.

gulafaran (approved):

vaxerski (dismissed):
appr from me

vaxerski:
@codex review please

src/config/values/ConfigValues.cpp
```diff
@@ -554,6 +554,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          * render:
          */
 
+        MS<Bool>("render:main_buffer_deadline", "Set a GPU fence deadline hint for the main render buffer.", true),
```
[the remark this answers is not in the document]
vaxerski: right, do we want this on by default?

gulafaran:
worth noting, seemingly only msm supports this anyways https://github.com/search?q=repo%3Atorvalds%2Flinux%20set_deadline&type=code

vaxerski:
I dont htink there's a point hten

gulafaran:
> I dont htink there's a point hten

probably not, unless some Snapdragon users wants it

PR #15009, opened 2026-06-07
groups: add groupbar disable_when_only option

feelamee:
#### group:groupbar:disable_when_only

Option allow to hide groupbar if it contains only single window.
This is especially useful with rule:
```lua
-- any window will be opened in group
hl.window_rule({ match = { title = ".*" }, group = "set" })
```

#### Possible issues

- option was added looking at `group:groupbar:enable`, but in `refreshGroupBarGradients` there is no availible active group to properly check option `disable_when_only`
  I think this check unnecessary here because same scenario already exist when there is no windows with groupbar, but option `enabled == true`. But unsure - require review.
- new hyprtest is unstable for some reason. Sometimes fail, and sometimes succeed. Maybe this is related to error "(EE) failed to read Wayland events: Broken pipe" which printed in stdout when I launched hyprtest. Anyway, I tested it in practice - launching Hyprland with this patch as single window manager - and feature works well.

First of all it is need an approve - do we need such feature at all. Then - I tried my best to do this PR as possible as ready for merge. But I'm open for critique and suggestions.

---

wiki PR - hyprwm/hyprland-wiki#1551

### DIFF
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -434,6 +434,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          */
 
         MS<Bool>("group:groupbar:enabled", "enables groupbars", true),
+        MS<Bool>("group:groupbar:disable_when_only", "disable if contains single window. Considered only if enabled == true", false),
         MS<String>("group:groupbar:font_family", "font used to display groupbar titles", "[[EMPTY]]"),
         MS<FontWeight>("group:groupbar:font_weight_active", "weight of the font used to display active groupbar titles"),
         MS<FontWeight>("group:groupbar:font_weight_inactive", "weight of the font used to display inactive groupbar titles"),
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -433,8 +433,9 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          * group:groupbar:
          */
 
-        MS<Bool>("group:groupbar:enabled", "enables groupbars", true),
-        MS<Bool>("group:groupbar:disable_when_only", "disable if contains single window. Considered only if enabled == true", false),
+        MS<Bool>("group:groupbar:enabled", "enables groupbars", true, {.refresh = Supplementary::REFRESH_WINDOW_STATES}),
+        MS<Bool>("group:groupbar:disable_when_only", "disable if contains single window. Considered only if enabled == true", false,
+                 {.refresh = Supplementary::REFRESH_WINDOW_STATES}),
         MS<String>("group:groupbar:font_family", "font used to display groupbar titles", "[[EMPTY]]"),
         MS<FontWeight>("group:groupbar:font_weight_active", "weight of the font used to display active groupbar titles"),
         MS<FontWeight>("group:groupbar:font_weight_inactive", "weight of the font used to display inactive groupbar titles"),
--- a/src/render/decorations/CHyprGroupBarDecoration.cpp
+++ b/src/render/decorations/CHyprGroupBarDecoration.cpp
@@ -101,8 +101,10 @@ void CHyprGroupBarDecoration::draw(PHLMONITOR pMonitor, float const& a) {
 
     const bool VISIBLE = visible();
 
-    if (VISIBLE != m_bLastVisibilityStatus)
+    if (!m_bLastVisibilityStatus.has_value() || VISIBLE != *m_bLastVisibilityStatus) {
         g_pDecorationPositioner->repositionDeco(this);
+        m_bLastVisibilityStatus = VISIBLE;
+    }
 
     if (!VISIBLE)
         return;
--- a/src/render/decorations/CHyprGroupBarDecoration.cpp
+++ b/src/render/decorations/CHyprGroupBarDecoration.cpp
@@ -25,8 +25,8 @@ static SP<ITexture> m_tGradientLockedInactive;
 constexpr int       BAR_TEXT_PAD = 2;
 
 CHyprGroupBarDecoration::CHyprGroupBarDecoration(PHLWINDOW pWindow) : IHyprWindowDecoration(pWindow), m_window(pWindow) {
-    static auto PGRADIENTS = CConfigValue<Config::INTEGER>("group:groupbar:enabled");
-    static auto PENABLED   = CConfigValue<Config::INTEGER>("group:groupbar:gradients");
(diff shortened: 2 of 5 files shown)


src/config/values/ConfigValues.cpp
```diff
@@ -433,6 +433,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          */
 
         MS<Bool>("group:groupbar:enabled", "enables groupbars", true),
+        MS<Bool>("group:groupbar:disable_if_contains_single_window", "disable if contains single window. Considered only if enabled == true", false),
```
vaxerski: `disable_when_only` would be way less of a mouthful

vaxerski (dismissed):
tnx

vaxerski:
your test fails :(

feelamee:
@vaxerski
look, now I broke `groups` test too :D 
Detailed explanation in commit message, pasting it here to:

> `groups_disable_when_only` test fails because it checks windows sizes after
> `disable_when_only` option changed. Changing option currently don't
> trigger window size update.
> 
> Same problem with `groups` test, but it worked because it or:
> - didn't checked window size after changing option.
> - called `hl.dsp.group.toggle()`, which trigger size update
> - at the begining set `enable = 1` while it `== 1` by default
> Now it fails too :D
> 
> Seems it works in usual Hyprland usage because:
> - `CHyprGroupBarDecoration::draw` calls
> - `g_pDecorationPositioner->repositionDeco(this)`,
>   which will invalidate cache and update sizes.

So, at first sight we just need to add `REFRESH_WINDOW_STATE | REFRESH_LAYOUTS` to `disable_when_only` config option. But it didn't work because in `CDecorationPositioner::onWindowUpdate` we will go in fast path and return without updating `WINDOWDATA->reserved`, which will be used in `CWindowTarget::updatePos` to finally update window size.
Therefore we need to `uncacheWindowDecos` first. I made ugly fix to prove this [here](https://github.com/feelamee/Hyprland/commit/91c8fd34f95ef6af51e54953d83a66d94214ec86), and this works. But we need something better.

My question:
- maybe we need `uncacheWindowDecos` when `REFRESH_WINDOW_STATE` triggered? If no, where we can insert it?
- then we need remove
```cpp
    if (VISIBLE != m_bLastVisibilityStatus)
        g_pDecorationPositioner->repositionDeco(this);
```
  from `CHyprGroupBarDecoration::draw`?
- and also add `REFRESH_WINDOW_STATE | REFRESH_LAYOUTS` to `group:groupbar:enabled` config option?

vaxerski:
~~you can call `hyprctl reload` probably then sync. Should work.~~

doh, you mean live. Yes, uncaching when window state changes is ok.

src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -101,7 +101,10 @@ void CHyprGroupBarDecoration::draw(PHLMONITOR pMonitor, float const& a) {
     const bool VISIBLE = visible();
 
     if (VISIBLE != m_bLastVisibilityStatus)
+    {
```
feelamee: this changes not actually required, I can't find reproducible bug.
But logic seems strange - we always read from `m_bLastVisibilityStatus`, which is `true` by default, but never write to it. Now fixed.

vaxerski:
clang-format please

feelamee:
@vaxerski, I fixed formatting. Any other suggestions? Or we can merge.
P.S. sorry for mentioning, is it ok to remind about PR here?

src/render/decorations/CHyprGroupBarDecoration.cpp
```diff
@@ -100,8 +100,10 @@ void CHyprGroupBarDecoration::draw(PHLMONITOR pMonitor, float const& a) {
 
     const bool VISIBLE = visible();
 
-    if (VISIBLE != m_bLastVisibilityStatus)
+    if (VISIBLE != m_bLastVisibilityStatus) {
```
vaxerski: this is now an optional. Are you sure about this comparison? this is basically `if (VISIBLE != m_bLastVisibilityStatus.has_value())`

vaxerski:
> I fixed formatting. Any other suggestions? Or we can merge.
P.S. sorry for mentioning, is it ok to remind about PR here?

you can ping me all you want, still shows up as a regular notification

src/render/decorations/CHyprGroupBarDecoration.cpp
feelamee: no, optional has comprasion `operator==(std::optional<bool>, bool)` ([here](https://en.cppreference.com/cpp/utility/optional/operator_cmp)).
So, in function resolution it will take precedence because `operator==(bool, bool)` also require implicit conversion.
See example in godbolt - https://godbolt.org/z/7r7sjovdq

For me that was obvious, but if for you or others not, I can rewrite it to more explicit comprasion like `!m_bLastVisibilityStatus.has_value() || *m_bLastVisibilityStatus != VISIBLE`

vaxerski: yeeeah I'd rather that. This is quite confusing as we've just seen

feelamee: done

vaxerski:
there's a conflict

feelamee:
Rebased on main, fixed conflict.
No new changes to PR added.

vaxerski (approved):

tekstryder:
This works. Thanks!

Missing a wiki entry tho.

feelamee:
@tekstryder, wiki PR will be merged soon (see https://github.com/hyprwm/hyprland-wiki/pull/1551)

tekstryder:
After using this since yesterday (same session) the ability to toggle a group eventually stopped working... any workspace for any window. The dispatcher simply does nothing. 

I'll open a discussion and link it here.

feelamee:
> ... the ability to toggle a group eventually stopped working... any workspace for any window. The dispatcher simply does nothing.

I can't reproduce, because my Hyprland build currently crashing on main just when first window opened. But, for now, can you describe step by step:
- what you doing to reproduce
- what you expect
- what you get

I'm not sure why this can happens because patch in this PR really influence only *drawing* of groupbar. And doesn't influence logic of *toggling* groupbar

tekstryder:
>my Hyprland build currently crashing on main

Are you also using Ubuntu?

- https://github.com/hyprwm/Hyprland/discussions/15426

Either way, you need to rebuild the entire stack.

>this PR really influence only drawing of groupbar.

Well yeah. It's not drawing.

See:

- https://github.com/hyprwm/Hyprland/discussions/15430

PR #15010, opened 2026-06-07
feat: implement vicinae-hotkey-v1 (client-managed global shortcuts)

aurelleb:

This PR implements the [vicinae-hotkey-v1](https://github.com/vicinaehq/vicinae-wayland-protocols/tree/main/staging/vicinae-hotkey) protocol.

This is a very simple protocol allowing a client to request global keybinds to the compositor, which can deny and later revoke them according to compositor policy.

This is how apps do global shortcuts on Windows, macOS and X11, making it trivial for existing apps to support global shortcuts on wayland.

The protocol can be tested with [this example client](https://github.com/vicinaehq/vicinae-wayland-protocols/tree/main/examples) or if you use [vicinae](https://github.com/vicinaehq/vicinae).

There is no freedesktop proposal for this protocol just yet. My goal is to get this upstreamed in major standalone compositors and then make an official proposal to try generalizing it further.

I submitted an implementation for [niri](https://github.com/niri-wm/niri/pull/4145) as well.

Ready. It is of limited utility as for now [vicinae](https://github.com/vicinaehq/vicinae/blob/d9fd7a17544c01b504fb166f9e8afd1d69858844/src/server/src/services/global-shortcuts/ext-hotkey-global-shortcut-backend.cpp) is the only client implementing it, but my goal is to start making PRs to popular clients if it gets upstreamed by major compositors.


vaxerski:
is this a vicinae-vendored protocol or a wp proposal?

aurelleb:
It is a vicinae vendored protocol.
My goal would be to get it into major standalone compositors (mostly hypr+niri) then turn it into an official wp one. If it gets upstreamed I'm going to PR client-side implementations for popular software that have broken global shortcut support.

Not starting with wp because I don't want it to stall forever

leiserfg:
I would be nice to have an api in lua so one can implement extra policies there, as simple as a function that gets called with every binding attempt and returns a `bool` (if `false` then it's a  `deny_reason.not_permitted`). Potential uses: acl for well know good/bad actors, showing a popup to learn about any app that created a binding...

aurelleb:
Yeah a keybind acceptance filter in lua can definitely be a good way to do it, and could supersede the default compositor policy maybe. The protocol also requires that the client sends an app id so it's kind of possible to do ACLs even though it's spoofable, so you don't get a lot of actual security out of that

vaxerski:
I would advise against making a vendored protocol under `ext-` as that's reserved for w-p ones. Why not just `vicinae-whatever-blah-v1`?

aurelleb:
True, my bad. I renamed it everywhere, hopefully I didn't miss anything.

protocols/vicinae-hotkey-v1.xml
```diff
@@ -0,0 +1,305 @@
...
+        can correlate the two. It is advisory and may be spoofed; a compositor
+        that needs a trustworthy identity SHOULD obtain it through the
+        security-context mechanism rather than relying on this string.
+
+        description is a human-readable, localized description of what the hotkey
+		  does (e.g. "Toggle the launcher"), for display in compositor UI or even in this protocol error messages.
+      </description>
+      <arg name="id" type="new_id" interface="vicinae_hotkey_v1"
+           summary="the new hotkey object"/>
+      <arg name="keysym" type="uint" summary="keysym of the trigger key (see description)"/>
```
vaxerski, line 156: what about binds like SUPER + H + K? I assume we dont allow those?

src/managers/KeybindManager.cpp
```diff
@@ -257,6 +258,32 @@ uint32_t CKeybindManager::keycodeToModifier(xkb_keycode_t keycode) {
     }
 }
 
+SP<SKeybind> CKeybindManager::findConflictingKeybind(xkb_keysym_t keysym, uint32_t modmask) {
+    if (keysym == XKB_KEY_NoSymbol)
+        return nullptr;
+
+    for (const auto& k : m_keybinds) {
+        if (!k->enabled || k->shadowed || k->mouse)
```
vaxerski: checking shadowed here makes no sense

src/protocols/Hotkey.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "../defines.hpp"
+#include "vicinae-hotkey-v1.hpp"
+#include "./WaylandProtocol.hpp"
+#include "../helpers/signal/Signal.hpp"
+
+#include <vector>
+#include <string>
+#include <xkbcommon/xkbcommon.h>
+
+struct SBoundHotkey {
```
vaxerski: move this struct inside CHotkeyProtocol

protocols/vicinae-hotkey-v1.xml
aurelleb: I didn't know we could do these.  Existing cross platform apps would never need this as this is not allowed on Windows or macOS (or it requires a third party keybind tool).

I'm assuming a lot of compositors also don't support it so I guess the protocol shouldn't either. Wdyt?

vaxerski: fair enough

vaxerski:
> There is no freedesktop proposal for this protocol just yet. My goal is to get this upstreamed in major standalone compositors and then make an official proposal to try generalizing it further.

poodesktop

vaxerski (approved):
this lgtm lmk if you want it merged now or wait for something

aurelleb:
I think we're good!

If you guys have any suggestions on what clients I should start sending patches to, I'm taking them. Of course, Hyprland support isn't enough to upstream to very big projects or toolkits, but smaller standalone ones, yes. I was thinking ghostty, OBS and maybe mumble/teamspeak to begin with.

I guess the various shells could make use of that too, so maybe quickshell

And thank you @vaxerski  for upstreaming this :saluting_face:

vaxerski:
you are forgetting hyprland is by far the most important desktop.

(*survey done amongst hyprland users)

aurelleb:
If any of you is involved with FDO, I submitted an official proposal: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/525

vaxerski:
we're all banned

aurelleb:
well I'm not (yet)

PR #15012, opened 2026-06-07
Avoid multiple instances of static variables being created in CLogger::log()

danielkrupinski:
Multiple instances of `static bool TRACE` existed because static variables inside template functions are created per template instantiation.


vaxerski (approved):
tnx!

PR #15013, opened 2026-06-07
Extract rarely executed code from RASSERT macro to a non-inline function

danielkrupinski:
Code that is unlikely to be executed (such as assertion failure handling) should not be inlined to not pollute the CPU instruction cache.

### DIFF
--- a/src/macros.hpp
+++ b/src/macros.hpp
@@ -44,13 +44,18 @@
 
 #define HYPRATOM(name) {name, 0}
 
+template <typename... Args>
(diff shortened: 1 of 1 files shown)


vaxerski (dismissed):
tnx!

vaxerski:
actually while we're at it, why not mark it as `[[unlikely]]`?

vaxerski:
clang-format please

danielkrupinski:
Done, added `[[unlikely]]` and `[[gnu::cold]]`

vaxerski (approved):
thanks!

PR #15035, opened 2026-06-09
renderer: harden damage handling and workspace color cache

ghost:
fixes a crash, getViewsForWorkspace returns empty if not found  and optimizes color conversion and hardens several renderer, damage-tracking, input-region, and compositor-cleanup paths that could result in sanitizer errors, invalid Pixman rectangles, or undefined behaviour.

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -3465,8 +3465,11 @@ using ColorConversionKey = std::tuple<float, float, float, float, uint64_t>;
 
 struct SColorConversionKeyHash {
     size_t operator()(const ColorConversionKey& key) const {
-        size_t     hash = 0;
+        size_t hash = 0;
 
+        // fold each tuple element to a order sensitive hash the constant is
+        // the 64-bit golden-ratio value used to
+        // distribute bits and reduce collisions between adjacent fields.
         const auto hashCombine = [&hash](const auto& value) { hash ^= std::hash<std::decay_t<decltype(value)>>{}(value) + 0x9e3779b97f4a7c15ULL + (hash << 6) + (hash >> 2); };
 
         hashCombine(std::get<0>(key));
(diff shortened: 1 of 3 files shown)


ghost:
btw i would recommend adding       -ftrivial-auto-var-init=pattern  to debug builds

vaxerski:
what's that do?

vaxerski:
can we break these MRs into smaller ones? you do 10 different things here each could be a separate MR or at least commit.

ghost:
> what's that do?

It initializes uninitialized automatic variables to a fixed pattern (generally 0xAA) and
Ptrs become values like (void*)0xAAAAAAAAAAAAAAAA which helps to spot usage of unitialized vars when debugging, and btw fno-omit-frame-pointer is in cmake but it might not be applied to hyprland bcs u first create lib and executable then u apply the flags so it might not be applied to targets that come before like hyprland

src/render/Renderer.cpp
```diff
@@ -3472,8 +3474,30 @@ SP<ITexture> IHyprRenderer::renderSplash(const std::function<SP<ITexture>(const
...
 
 using ColorConversionKey = std::tuple<float, float, float, float, uint64_t>;
-static std::map<ColorConversionKey, CHyprColor> colorConversionCache;
-constexpr const size_t                          MAX_COLOR_CONVERSION_CACHE_SIZE = 4096;
+
+struct SColorConversionKeyHash {
+    size_t operator()(const ColorConversionKey& key) const {
+        size_t     hash = 0;
+
+        const auto hashCombine = [&hash](const auto& value) { hash ^= std::hash<std::decay_t<decltype(value)>>{}(value) + 0x9e3779b97f4a7c15ULL + (hash << 6) + (hash >> 2); };
```
vaxerski, line 3473: this needs a comment, what the fuck is this

vaxerski (commented):
rest of the commits are ok

vaxerski:
rebase your commits, I'll rebase them onto main. Also clang-format.

vaxerski:
conflicts

vaxerski (approved):

PR #15036, opened 2026-06-09
Reduce the size of the machine code generated from CI18nEngine constructor

danielkrupinski:
Hyprland binary file size reduction: `18 639 KB` -> `18 286 KB`


vaxerski (approved):
harmless microbenchmark I guess xD

PR #15025, opened 2026-06-08
Hyprtester misc bug fixes & refactorings

kolayne:
#### Describe your PR, what does it fix/add?

A few improvements to different parts of hyprtester. Refer to commit messages and descriptions for details.

#### Is there anything you want to mention? (unchecked code, possible bugs, found problems, breaking compatibility, etc.)

No.

#### Is it ready for merging, or does it need work?

Invited to previous hyprtester contributors to take a look at this (I don't know if they are coming). From my perspective, ready for merge.

<hr>

@vaxerski, I'd appreciate it if you pinged/invited me to review future pull requests concerning hyprtester. Not the ones that change tests themselves, of course, but the ones that affect the hyprtester framework or the executable.

PR #15038, opened 2026-06-09
hyprtester: improvements, remove some redundancy and other things.

cebem1nt:
fixes some things commented at #14774, Adds `SPAWN_KITTY` and `SPAWN_LAYER_KITTY` macros that automatically check if kitty was spawned. Solve some TODOs, cleanup. Renames `NLog::info` to `NLog::green`, `NLog::error` to `NLog::red`. 

nope

ready


cebem1nt:
@kolayne, you might be interested

hyprtester/src/shared.hpp
kolayne: In a test such as
```cpp
TEST_CASE(hello_world) {
    EXPECT(true, false);  // Marks test as failed without terminating it
    PASS_TEST("Done");
}
```

should `PASS_TEST` result in test failure or test success?

If it should override the previous failure, then it's not implemented correctly. If it should not, then the name is confusing: it does not actually pass the test, it just finishes it early.

I suppose, you intended the latter. The reason why I did not initially add such a macro is that it essentially just makes an early return, and the test flow is more clear when there is an explicit return keyword in the code. Or, maybe we could use the word "return" in the macro name, to make its purpose more clear.

kolayne (changes requested):
I'll have time to review it more carefully today. The first thing that caught my eye

hyprtester/src/shared.hpp
kolayne: Notably, the old messages were questionable, too. They print "passed", when the test has not actually passed yet, just a check evaluated correctly. So maybe we could just keep the old messages and remove the "Passed:" prefix.

But you also additionally `return` from the tests, which is not what happened there before. This is incorrect.

hyprtester/src/tests/main/scroll.cpp
```diff
@@ -787,55 +699,48 @@ TEST_CASE(testScrollingViewBehaviourNewLayer) {
...
-    if (std::stoi(currentWindowPosX) < 0) {
-        NLog ::log("{}Passed: {}window of class 'a' has negative x coordinates for its position: {}", Colors ::GREEN, Colors::RESET, currentWindowPosX);
-    }
-    // test fail
-    else {
-        FAIL_TEST("{}Failed: {}window of class 'a' does not have negative x coordinates for its position: {}", Colors::RED, Colors::RESET, currentWindowPosX);
-    }
 
     // TEST_CASE's own cleanup functions fail to kill all layers with this test. Manually do it
+    // TODO Why it fails?
```
kolayne: Outdated comment?

hyprtester/src/tests/misc/shared.cpp
```diff
@@ -5,5 +5,6 @@
 
 // Test if Tests::spawnKitty works
 TEST_CASE(spawnKitty) {
-    ASSERT(!!Tests::spawnKitty("kitty"), true);
+    if (!Tests::spawnKitty("A"))
+        FAIL_TEST("Could not spawn kitty!");
```
kolayne, line 9: Should it become `SPAWN_KITTY` now?

hyprtester/src/Log.hpp
```diff
@@ -17,25 +17,27 @@ namespace NLog {
...
 
+    // Logs message with yellow
     template <typename... Args>
-    //NOLINTNEXTLINE
-    void info(std::format_string<Args...> fmt, Args&&... args) {
-        std::string logMsg = "";
-
-        logMsg += std::vformat(fmt.get(), std::make_format_args(args...));
+    // NOLINTNEXTLINE
+    void alert(std::format_string<Args...> fmt, Args&&... args) {
```
kolayne: 1. If these comments begin with three slashes (`///`), some IDEs will recognize them as doxygen documentation and display this message when describing this function (e.g., on hover, etc).
2. To me, the new correspondence between method names and colors feels a bit counter-intuitive. Green feels like a color of success, so maybe `info` should be some more neutral color (like yellow or no color at all?); and I'd expect a method named `alert` print a message in either red or other eye-catching color (e.g., purple?). <br> Others are welcome to share their opinion. Personally, I usually look at hyprtester output with colors removed.

hyprtester/src/shared.hpp
```diff
@@ -248,4 +254,16 @@ class CTestCase {
...
+        if (!Tests::spawnKitty(class_))                                                                                                                                            \
+            FAIL_TEST("Could not spawn kitty with class: {}", class_);                                                                                                             \
+    } while (0)
+
+#define SPAWN_LAYER_KITTY(class_)                                                                                                                                                  \
+    do {                                                                                                                                                                           \
+        if (!Tests::spawnLayerKitty(class_))                                                                                                                                       \
+            FAIL_TEST("Could not spawn layer kitty with class: {}", class_);                                                                                                       \
+    } while (0)
```
kolayne: `Tests::spawnKitty` and `Tests::spawnLayerKitty` also accept optional arguments. Could we support the same in the macros?

hyprtester/src/shared.hpp
cebem1nt: Yeah, my bad, forgot to explicitly mark it as passed. 

> The reason why I did not initially add such a macro is that it essentially just makes an early return

Exactly, and prints a message. It is really useful in such test cases as in scrolling. You check value, explicitly print if it's all fine and return.

hyprtester/src/tests/misc/shared.cpp
cebem1nt: I think the explicit one looks better

hyprtester/src/Log.hpp
cebem1nt: > If these comments begin with three slashes (///)

~~well, they don't~~ Ah, I got you

> To me, the new correspondence between method names and colors feels a bit counter-intuitive.

Sure, I think the best way would to just create `NLog::red()`, `NLog::green()` etc functions but though, it will look dumb, and create a rainbow disaster. 

As it is right now, seems to be second best way, green color was used to informate, yellow to informate about "a bit extraordinary things", and red to report errors.

hyprtester/src/shared.hpp
kolayne: But why do you want to return early? Today these lines are at the end of functions, so there is seemingly no difference. But if in the future someone wants to add something extra at the end - even just another assertion - they would need to modify this previous call that terminates the test early.

As far as I can see, it was incorrect that the tests in question printed "Passed:" in their body, allegedly indicating that the whole test has passed. In fact, it's just one check that is passing (even if it's the final check), and it's up to hyprtester itself (the framework/engine) to indicate that the test was successful.

hyprtester/src/Log.hpp
kolayne: IMO, `NLog::red`, `NLog::yellow`, `NLog::green` seem not that bad!

In the context of hyprtester, I think it's pretty unlikely that we will, e.g., suppress logs of certain levels or do some other manipulations based on their "severity" (alert/error/info). The only difference seems to be the color. So, I'd say we should pick some names that make it clear what color the message will be printed on. Color names work, too :)

kolayne: Wait, it actually sounds like we need something like
```cpp
NLog::log(Colors::RED, "The {}", "message");
```
which would expand to
```cpp
Nlog::log("{}The {}{}", Colors::RED, message, Colors::RESET);
```

cebem1nt: > Wait, it actually sounds like we need something like
> 
> ```c++
> NLog::log(Colors::RED, "The {}", "message");
> ```
> 
> which would expand to
> 
> ```c++
> Nlog::log("{}The {}{}", Colors::RED, message, Colors::RESET);
> ```

I think `NLog::red`, `NLog::yellow`, `NLog::green` is cleaner

kolayne:
LGTM except for the `PASS_TEST` macro and its use, i.e., forgetting previous failures and terminating a test early. I don't think that's the right pattern.

vaxerski:
yeah would be nice if you actually made it pass

cebem1nt:
> yeah would be nice if you actually made it pass

yup but no clue why it fails, seems to be all red from #15025 :(

kolayne:
@vaxerski, the test is failing due to this issue: https://github.com/hyprwm/Hyprland/pull/14633#issuecomment-4706749995

It is a problem related to keybinds overriding logic, and I suspect it might be a Hyprland bug rather than a tests bug (haven't look into details yet). I tried asking @ahmedkall, the original author of that logic, but he seems unresponsive, so, I think I will have time to look into it tomorrow or on the weekend (unless someone else fixes it before that). If you're familiar with the logic in question, you may take a look and leave a comment to provide more context / explain the expect behavior in #14633.

In any case, IMO, that fix is conceptually unrelated to this PR, so I think they should be worked on separately.

kolayne (approved):

kolayne:
@cebem1nt, please rebase on master. Tests should pass now.

kolayne:
@vaxerski, ready for merge

vaxerski (approved):
tnx

PR #15043, opened 2026-06-09
hyprctl: add interactive Lua REPL mode

Dregu:

Adds interactive Lua REPL mode and better eval to hyprctl. Old eval only prints errors or "ok", this is not very useful.

- `hyprctl repl` with no arguments starts a readline loop, end with ^D
- `hyprctl repl 'lua code'` works like eval but prints anything returned through tostring
- `print()` is also hooked so anything printed during repl prints to hyprctl
- added exit status if output is an error

```sh
~  hyprctl eval 'hl.version()'
ok
~  hyprctl repl 'hl.version()'
0.55.0
~  hyprctl repl 'balls'
nil
~  hyprctl repl 'balls()'
error: [string "return balls();"]:1: attempt to call a nil value (global 'balls')
~  hyprctl repl                                                                                                                                          ✘ 7 
> hl.version()
0.55.0
> print("is hooked")
is hooked
> 1,2,3
1	2	3
> inspect(hl.dsp.workspace)
{
  move = <function 1>,
  rename = <function 2>,
  swap_monitors = <function 3>,
  toggle_special = <function 4>
}
```

- Didn't want to break "ok" support since all tests and probably other scripts depend on it
- Similar but not as sophisticated as Luas own repl
  - Tries to put "return" in front of everything and tests if that's valid. This is how `lua` does it too.
  - Tabulates multiple return values
  - Doesn't support incomplete multiline input though, maybe later
- Not a Lua stack expert yet, hopefully it's right

Ready. https://github.com/hyprwm/hyprland-wiki/pull/1557

### DIFF
--- a/hyprctl/src/main.cpp
+++ b/hyprctl/src/main.cpp
@@ -33,6 +33,9 @@ using namespace Hyprutils::Memory;
 #include "Strings.hpp"
 #include "hyprpaper/Hyprpaper.hpp"
 
+#include <readline/readline.h>
+#include <readline/history.h>
+
 std::string instanceSignature;
 bool        quiet = false;
 
@@ -532,7 +535,17 @@ int main(int argc, char** argv) {
         std::println("{}", USAGE);
     else if (fullRequest.contains("/rollinglog") && needRoll)
         exitStatus = request(fullRequest, 0, true);
-    else {
+    else if (fullRequest.contains("/repl")) {
+        char* input = nullptr;
+        while ((input = readline("> ")) != nullptr) {
+            std::string line(input);
+            if (!line.empty()) {
+                exitStatus = request("/eval " + line);
+                add_history(input);
+            }
+            free(input);
+        }
+    } else {
         exitStatus = request(fullRequest);
     }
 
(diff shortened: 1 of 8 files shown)


hyprctl/src/main.cpp
```diff
@@ -532,7 +538,23 @@ int main(int argc, char** argv) {
...
         exitStatus = request(fullRequest, 0, true);
-    else {
+    else if (fullRequest.contains("/repl")) {
+        if (ARGS.size() > 1) {
+            // single command with output
+            exitStatus = request(fullRequest, 1);
+        } else {
+            // interactive REPL mode
+            char* input = nullptr;
+            while ((input = readline("> ")) != nullptr) {
```
vaxerski, line 548: ew is there no STL equiv?

Dregu: I don't think with this kind of history and cursor capabilities ootb? Readline is a dep of lua anyway, that's one reason this didn't feel ew to me...

vaxerski: fair enough

vaxerski (commented):
docs would be nice

Dregu:
Wiki done

vaxerski (approved):

PR #15048, opened 2026-06-10
renderer: fix SEGV on monitor disconnect

neonetizen:

adds null check in `View::getViewsFor Workspace` and guards the `m_activeWorkspace` frame events call, preventing a SEGV when a monitor disconnects hwile a frame event is pending. when a DP monitor drops its link (dpms off, suspend/resume), `m_activeWorkspace` can be null while a queued drm frame event still fires and `renderMonitor` passes it unconditionally to `sendFrameEventsToWorkspace`, handing it to `getViewsForWorkspace` where `ws->m_monitor` derefs null.

reproduces reliably on a 3x DP setup with `hyprctl dispatch dpms off`

backtrace is from the release package build (caught in normal use); i can produce a debug-build trace if needed, but the deref is visible from the trace + code. null workspace tick becomes a no-op which seems right.

ready; tested through dpms off/on cycles and suspend/resume on the affected setup and works fine now

[hyprlandCrashReport75844.txt](https://github.com/user-attachments/files/28777227/hyprlandCrashReport75844.txt)

### DIFF
--- a/src/desktop/view/GlobalViewMethods.cpp
+++ b/src/desktop/view/GlobalViewMethods.cpp
@@ -17,6 +17,9 @@ using namespace Desktop::View;
 std::vector<SP<IView>> View::getViewsForWorkspace(PHLWORKSPACE ws) {
     std::vector<SP<IView>> views;
 
+    if (!ws)
+        return views;
+
     for (const auto& w : g_pCompositor->m_windows) {
         if (!w->aliveAndVisible() || w->m_workspace != ws)
             continue;
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -2192,7 +2192,8 @@ void IHyprRenderer::renderMonitor(PHLMONITOR pMonitor, bool commit) {
             }
         }
     } else if (!pMonitor->isMirror()) {
-        sendFrameEventsToWorkspace(pMonitor, pMonitor->m_activeWorkspace, NOW);
+        if (pMonitor->m_activeWorkspace)
+            sendFrameEventsToWorkspace(pMonitor, pMonitor->m_activeWorkspace, NOW);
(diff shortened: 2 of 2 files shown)


vaxerski (approved):

derekszen:
Adding a downstream confirmation that this PR appears to match a real-world Omarchy-triggered crash on an HKC GS49UK DisplayPort monitor.

Local crash reports from 2026-06-18, before updating to `0.55.4`, show:

```text
Hyprland 0.55.3
Aquamarine 0.12.0
GPU: NVIDIA GB202 [GeForce RTX 5090 D]
Monitor: HKC GS49UK, DP-1, 5120x1440@240

Desktop::View::getViewsForWorkspace(...)
Render::IHyprRenderer::sendFrameEventsToWorkspace(...)
Aquamarine::CHeadlessOutput::commit()
CCompositor::enterUnsafeState()
CMonitor::onDisconnect(bool)
...
DEBUG from aquamarine ]: drm: Connector DP-1 disconnected
```

This happened when Omarchy's lock/idle path turned the display off via `hyprctl dispatch dpms off`. The monitor appears to drop/retrain the DP link, which lines up with this PR's description: a DP monitor drops link during `dpms off`, a pending frame event fires, and `getViewsForWorkspace` sees invalid/null workspace state.

Timeline:

```text
2026-06-11: hyprland 0.55.2 -> 0.55.3
2026-06-18: two matching Hyprland crashes under 0.55.3
2026-06-19: hyprland 0.55.3 -> 0.55.4
```

After updating to `0.55.4`, I do not see any new Hyprland coredumps or `Hyprland has crashed` journal entries, though I have intentionally not re-run the risky `dpms off` repro because the earlier failures forced a reboot.

Related downstream Omarchy report: https://github.com/basecamp/omarchy/issues/6111

PR #15070, opened 2026-06-11
state/monitorQuery: implement monitor query, move logic off of Compositor

vaxerski:
Reduce god-classiness of CCompositor, part 1: monitor lookups


ItsOhen:
Does it have to be so ugly though?
`query()->thing(thing).run()`?

vaxerski:
ugly is subjective.

PR #15074, opened 2026-06-11
config: fix description matching for default workspaces

vaxerski:
Ref https://github.com/hyprwm/Hyprland/discussions/15047

### DIFF
--- a/src/config/shared/workspace/WorkspaceRuleManager.cpp
+++ b/src/config/shared/workspace/WorkspaceRuleManager.cpp
@@ -51,19 +51,14 @@ std::optional<CWorkspaceRule> CWorkspaceRuleManager::getWorkspaceRuleFor(PHLWORK
     return mergedRule;
 }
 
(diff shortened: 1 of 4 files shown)


lichie567:
Seems like it works

Dinozor:
Not sure if I should write it here, but found out that this one probably breaks old behavior about monitor mirroring.
I have eDP-1 monitor mirrored on DP-4. All the workspaces behaved like one monitor. Workspace 1 and workspace 2 were visible same on both monitors.
Now it is being given the defaults per monitor ignoring mirroring. Meaning that whatever I do - one of the workspaces is somehow "lost". I see it a list, it takes space in the workspaces bar, consumes the shortcut, but does not work.
Whatever I do - I don't know how to get previous "normal" behavior to work.
No I just did 
`hl.workspace_rule({ workspace = "99", monitor = "DP-4", default = true })`
Not a fix as now I permanently have this "99" on my bar and need some filtering and other hoops to not show it.
Is there a way to account for mirrors?

PR #15077, opened 2026-06-11
dnd: fix touch-driven drag-and-drop

hjanuschka:

Fixes drag-and-drop initiated from a touchscreen (long-press drag). Touch could *start* a drag, but not *drive* it:

- `wl_data_device.motion` was sent with `ITouch::SMotionEvent.pos` as-is. That value is normalized `[0, 1]` monitor-relative, not surface-local, so drop targets saw the drag permanently stuck in their top-left corner.
- `dndPointerFocus` is only updated by the pointer focus pipeline, so `enter`/`leave` never followed the finger.
- `renderDND()` draws the drag icon at the mouse cursor, which is unrelated to where the finger is.

Fix:

- translate the touch position to global layout coords (same mapping `CInputManager::onTouchMove` uses)
- call `refocus()` on touch motion during a drag, so the existing dnd focus path keeps `enter`/`leave`/offers correct, exactly like mouse drags
- send surface-local motion coords
- anchor the drag icon to the touch point, from the moment the drag starts

Tablets got the same treatment recently (#11270, #11390); touch was the last input type left behind.

Tested on a touchscreen with Chromium (working on bringin touchdrag to chromium -> https://crbug.com/451651623)

Ready for merging.

### DIFF
--- a/src/protocols/core/DataDevice.cpp
+++ b/src/protocols/core/DataDevice.cpp
@@ -564,6 +564,12 @@ void CWLDataDeviceProtocol::initiateDrag(WP<CWLDataSourceResource> currentSource
     Cursor::overrideController->setOverride("grabbing", Cursor::CURSOR_OVERRIDE_DND);
     m_dnd.overriddenCursor = true;
 
+    // For touch-initiated drags, anchor the drag icon to the touch point
+    // right away. Until the first touch motion arrives it would otherwise
+    // be rendered at the (possibly far away) mouse cursor position.
+    if (g_pInputManager->m_lastInputTouch)
+        m_dnd.touchPos = g_pInputManager->m_touchData.lastTouchPos;
+
     LOGM(Log::DEBUG, "initiateDrag: source {:x}, surface: {:x}, origin: {:x}", (uintptr_t)currentSource.get(), (uintptr_t)dragSurface, (uintptr_t)origin);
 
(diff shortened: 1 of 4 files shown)


src/protocols/core/DataDevice.cpp
```diff
@@ -623,6 +630,22 @@ void CWLDataDeviceProtocol::initiateDrag(WP<CWLDataSourceResource> currentSource
...
+
+        if (!PMONITOR)
+            return;
+
+        const auto COORDS = PMONITOR->m_position + (e.pos * PMONITOR->m_size);
+
+        m_dnd.touchPos                            = COORDS;
+        g_pInputManager->m_touchData.lastTouchPos = COORDS;
+
+        g_pInputManager->refocus(COORDS);
```
vaxerski: mouse does not refocus, is this needed? Smeels wrong

hjanuschka: thanks for pushing back, that was uglier than it needed to be. Moved the refocus into `CInputManager::onTouchMove`, so the input layer drives the focus update during a touch drag the same way `mouseMoveUnified` does for cursor motion.

The dnd listener in `DataDevice.cpp` is now coords + sendMotion only, structurally identical to the mouseMove one.

manually confirmed on hardware touch display

vaxerski (approved):
tnx

PR #15079, opened 2026-06-11
Dwindle: fix moving windows across a split

LRitzdorf:
This PR fixes two related window-move bugs in dwindle:
- Attempting to move a window that fills less than 50% of its split to the opposite side of said split, will just place the window back on the same side.
- A split's direction is forgotten during a move when it probably shouldn't be. e.g. you have a vertical-split node that you've explicitly made that way (via the togglesplit layout message), attempting to move a child of that node up or down makes the node revert to its original horizontal layout.

Both of these behaviors arise because the window-move logic works by choosing a "focus point" for new-window placement. That focus point works fine most of the time, except when we have two windows that are direct siblings of a layout node, and we try to move one of them toward the other.
In the small-window case, the focus point is on the same side of the split that the window-to-move was already on, so the window simply doesn't get moved.
In the togglesplit-forgetting case, the new layout node just falls back to whatever split orientation dwindle prefers, based on aspect ratio. This can result in you hitting, say, your move-up bind and having the window go right instead.

Both of these are resolved by identifying move operations where a) a window is pushed toward its most immediate split divider and b) its partner across that divider is not another subdivided node, and overriding the split direction to ensure the new window is placed on the opposite side of that partner. e.g. if you're on the left side of a horizontal split, moving right triggers the new logic.

I've also included a single test that covers both of the issues this change fixes. Let me know if two separate tests would be preferable, or if maybe some of this behavior isn't specified well enough to be worth testing(?).

### DIFF
--- a/hyprtester/src/tests/main/dwindle.cpp
+++ b/hyprtester/src/tests/main/dwindle.cpp
@@ -214,3 +214,52 @@ TEST_CASE(dwindleForceSplitOnMoveToWorkspace) {
     std::string activeWindow = getFromSocket("/activewindow");
     EXPECT(activeWindow.contains(posBefore), false);
 }
+
+TEST_CASE(dwindleMoveAcrossToggledSplit) {
+    // If we have a split whose orientation has been manually toggled (e.g.
+    // vertically stacked, when the split's aspect ratio is such that it would
+    // prefer to be horizontally stacked by default), moving a window across
+    // the split should NOT revert back to the preferred split orientation
+
+    OK(getFromSocket("/eval hl.config({ dwindle = { force_split = 2 } })"));
+    for (auto const& win : {"a", "b"}) {
+        if (!Tests::spawnKitty(win)) {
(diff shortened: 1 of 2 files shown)


LRitzdorf:
Dang it, the one time I'm sure I have code style right on the first try, the tests get me

vaxerski:
considering you're fixing two cases, two tests would be nice :)

LRitzdorf:
Tests have been split! Also rebased onto latest main, because why not

LRitzdorf:
Looks like a random failure of the keybinds test? Is that a thing that just happens occasionally?

Dwindle ones run fine, starting at https://github.com/hyprwm/Hyprland/actions/runs/27473630275/job/81208862739?pr=15079#step:3:7403

vaxerski:
yes dont worry about those

vaxerski (approved):

vaxerski:
we run a muammar gaddafi style work environment when it comes to test regressions, and the person who committed code that fails tests has been scheduled for execution, don't worry

PR #15088, opened 2026-06-12
buffer: wait on seperate fences instead of merging

gulafaran:
most likely a lot of these fences will signal at the same time and merging them creates unneeded syscalls, just wait independently on each one of them and drain them as they become readable.


src/protocols/core/Compositor.cpp
```diff
@@ -548,15 +548,36 @@ void CWLSurfaceResource::scheduleState(WP<SSurfaceState> state) {
...
         // state commit without a buffer.
         m_stateQueue.tryProcess();
     }
 }
 
+void CWLSurfaceResource::drainSyncFds(WP<SSurfaceState> state, eLockReason reason) {
+    auto& fds = state->buffer->m_syncFds;
+
+    while (!fds.empty() && fds.front().isReadable())
+        fds.erase(fds.begin());
```
vaxerski: should you really stop? why not just erase_if on all

gulafaran: was basing it off how kwin does, sort of fifo. but it seems they sort of only half does it. so yeah erase_if could be used, will change.

vaxerski (approved):

PR #15094, opened 2026-06-12
hyprpm: sanitize commit pin targets

vaxerski:
Fixes missing sanitization around commit pins


umbrageodotus:
there's still more bs like this one, line 210 (you need to replace `'` with it's url-encoded form)

umbrageodotus:
also, no need to check for ' AND for hash

vaxerski:
> there's still more bs like this one, line 210 (you need to replace ' with it's url-encoded form)

I mean it will reject but not break out

umbrageodotus:
in line 210 it would run the command and then error...

vaxerski:
no, url is in single quotes and is checked via `validArg()`

umbrageodotus:
> no, url is in single quotes and is checked via `validArg()`

oh ye mb

PR #15095, opened 2026-06-12
helpers: remove old header

vaxerski:
Oops.


numkem:
This broke hy3 and hyprsplit at minimum. Is there a reason why the file was considered old and needed removal?

vaxerski:
it was refactored and I forgot to remove this, it wasnt used

PR #15103, opened 2026-06-13
fix several safety issues and clean up unused state

ghost:
- **cleanup unused test and protocol state**
- **fix subsurface damage and notification fallback color**
- **render: initialize current GL program**
- **seat: handle empty and failed key array allocation**
- **animation: harden parsing and weak pointer handling**
- **window: harden workspace and state handling**
- **layout: guard against invalid calculated window sizes**
- **motion blur: validate state before calculations**
- **cmake: improve compiler flag and debug configuration handling**

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -461,55 +461,62 @@ void CWindow::updateSurfaceScaleTransformDetails(bool force) {
     if (!m_isMapped || m_hidden)
         return;
 
-    const auto PLASTMONITOR = State::monitorState()->query().id(m_lastSurfaceMonitorID).run();
+    if (!m_wlSurface)
+        return;
 
-    m_lastSurfaceMonitorID = monitorID();
+    const auto RESOURCE = m_wlSurface->resource();
+    if (!RESOURCE)
+        return;
 
     const auto PNEWMONITOR = m_monitor.lock();
-
     if (!PNEWMONITOR)
         return;
 
-    if (PNEWMONITOR != PLASTMONITOR || force) {
-        if (PLASTMONITOR && PLASTMONITOR->m_enabled && PNEWMONITOR != PLASTMONITOR)
-            m_wlSurface->resource()->breadthfirst([PLASTMONITOR](SP<CWLSurfaceResource> s, const Vector2D& offset, void* d) { s->leave(PLASTMONITOR->m_self.lock()); }, nullptr);
+    const auto PLASTMONITOR = State::monitorState()->query().id(m_lastSurfaceMonitorID).run();
 
-        m_wlSurface->resource()->breadthfirst([PNEWMONITOR](SP<CWLSurfaceResource> s, const Vector2D& offset, void* d) { s->enter(PNEWMONITOR->m_self.lock()); }, nullptr);
+    m_lastSurfaceMonitorID = PNEWMONITOR->m_id;
+
+    if (PNEWMONITOR != PLASTMONITOR || force) {
+        if (PLASTMONITOR && PLASTMONITOR->m_enabled && PNEWMONITOR != PLASTMONITOR) {
+            RESOURCE->breadthfirst([PLASTMONITOR](SP<CWLSurfaceResource> surface, const Vector2D&, void*) { surface->leave(PLASTMONITOR->m_self.lock()); }, nullptr);
+        }
+
+        RESOURCE->breadthfirst([PNEWMONITOR](SP<CWLSurfaceResource> surface, const Vector2D&, void*) { surface->enter(PNEWMONITOR->m_self.lock()); }, nullptr);
     }
 
-    const auto PMONITOR = m_monitor.lock();
+    RESOURCE->breadthfirst(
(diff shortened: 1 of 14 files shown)


src/protocols/Screencopy.cpp
```diff
@@ -54,14 +54,16 @@ bool CScreencopyClient::good() {
...
 }
 
-CScreencopyFrame::CScreencopyFrame(SP<CZwlrScreencopyFrameV1> resource_, WP<CScreenshareSession> session, bool overlayCursor) :
-    m_resource(resource_), m_session(session), m_overlayCursor(overlayCursor) {
+CScreencopyFrame::CScreencopyFrame(SP<CZwlrScreencopyFrameV1> resource_, WP<CScreenshareSession> session, bool overlayCursor) : m_resource(resource_), m_session(session) {
     if UNLIKELY (!good())
         return;
 
     m_resource->setOnDestroy([this](CZwlrScreencopyFrameV1* pMgr) { PROTO::screencopy->destroyResource(this); });
+
```
vaxerski, line 62: what are these for

src/managers/SeatManager.cpp
```diff
@@ -147,10 +147,17 @@
...
 
+    if (PRESSEDARRSIZE != 0) {
+        const auto PKEYS = wl_array_add(&keys, PRESSEDARRSIZE);
+
+        if (!PKEYS) {
+            Log::logger->log(Log::ERR, "Failed to allocate pressed keyboard key array");
+            return;
+        }
+
+        memcpy(PKEYS, PRESSED.data(), PRESSEDARRSIZE);
```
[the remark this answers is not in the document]
vaxerski: I think this is partially correct, we mioght be copying garbage because pressed airr size is + 1

src/managers/animation/AnimationManager.cpp
```diff
@@ -401,18 +414,23 @@ std::string CHyprAnimationManager::styleValidInConfigVar(const std::string& conf
...
-
+                const auto pos = style.find_last_of(' ');
+                if (pos == std::string::npos)
+                    return "invalid movePerc";
+                const std::string percstr = style.substr(pos + 1);
+                if (percstr.empty() || percstr.back() != '%')
+                    return "invalid movePerc";
+                const std::string numstr = percstr.substr(0, percstr.length() - 1);
+                int               val    = 0;
+                const auto        res    = std::from_chars(numstr.data(), numstr.data() + numstr.size(), val);
```
vaxerski: use strToNumber from hyprutils' numeric

src/desktop/view/Window.cpp
```diff
@@ -518,29 +525,38 @@ void CWindow::moveToWorkspace(PHLWORKSPACE pWorkspace) {
...
     setAnimationsToMove();
 
     g_pCompositor->updateAllWindowsAnimatedDecorationValues();
 
-    if (valid(pWorkspace)) {
-        g_pEventManager->postEvent(SHyprIPCEvent{.event = "movewindow", .data = std::format("{:x},{}", rc<uintptr_t>(this), pWorkspace->m_name)});
-        g_pEventManager->postEvent(SHyprIPCEvent{.event = "movewindowv2", .data = std::format("{:x},{},{}", rc<uintptr_t>(this), pWorkspace->m_id, pWorkspace->m_name)});
-        Event::bus()->m_events.window.moveToWorkspace.emit(m_self.lock(), pWorkspace);
-    }
+    g_pEventManager->postEvent(SHyprIPCEvent{
```
vaxerski, line 549: removing a null guard??

src/render/transformer/MotionBlurTransformer.cpp
```diff
@@ -54,13 +69,25 @@ void CMotionBlurTransformer::amendTransformedRenderData(const CBox& currentBox,
...
+    motionBlur.current = currentBox;
+
+    assert(boxIsFinite(motionBlur.previous));
+    assert(boxIsFinite(motionBlur.current));
 
     *pMotionBlurData = motionBlur;
 }
 
 void CMotionBlurTransformer::record(const CBox& previous, const CBox& current) {
+    assert(boxIsFinite(previous));
```
vaxerski, line 88: `ASSERT`

vaxerski (commented):
please for the love of god make smaller, more focused MRs...

src/desktop/view/Window.cpp
ghost: Doesn't protect against anything, a new null guard was added earlier and it's only read since then, it isn't modified so it can't be null

src/managers/SeatManager.cpp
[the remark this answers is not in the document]
ghost: It isn't +1 if I'm not wrong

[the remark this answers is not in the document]
vaxerski:
```
const auto PRESSEDARRSIZE = PRESSED.size() * sizeof(uint32_t);
```

it is + 1 (uint32_t)

and then you memcpy from PRESSED.data() with size() + 1

PR #15109, opened 2026-06-14
MonitorQuery: fix monitor desc selector being treated as "down" selector

lichie567:
Monitor query by monitor description was broken because `isDirection(sv[0])` was called first and returns true when the first character is a 'd'. This PR adds a simple guard to prevent that happening when the selector starts with "desc:". Also removed a redundant `sv.empty()` check as it is already checked above. This section of the code seems like it could use some significant refactoring.

No

Ready

### DIFF
--- a/src/helpers/MiscFunctions.cpp
+++ b/src/helpers/MiscFunctions.cpp
@@ -113,12 +113,11 @@ std::optional<float> getPlusMinusKeywordResult(std::string source, float relativ
     }
 }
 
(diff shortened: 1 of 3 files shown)


vaxerski:
a better solution is to simply make `isDirection` accept a string_view and check for `lrudtb` or `left` `right` etc

lichie567:
Done

vaxerski (approved):

PR #15130, opened 2026-06-14
fix: remove the seam/line between two adjacent layers surfaces with blur effect

Ahmed5Emad:
fixes a weird seam/line when two panels using blur-effect(ext-background-effect) are positioned adjacent to each other.
**Example**
image

no breaking changes really : All 255 GTests is passed
but there is performance tradeoff , instead of damaging or computing blur per area , it compute full monitor so it just compute one time instead of two different surfaces (if there is better idea please tell me tried many things and this the best I got) , it should have effect on performance but to be honest in my test on my laptop I didn't see any change other than 50-150mb in gpu when I open the panel.

yes it is ready and here is the live result:
image
image

#### Why did I even do this 
first I thought it is noctalia error but it damn worked on niri fine and I kinda didn't like that 🙂


vaxerski:
what?

Instead of a very weird band-aid, maybe find out the root cause first?

Ahmed5Emad:
> what?
> 
> Instead of a very weird band-aid, maybe find out the root cause first?

I kinda know the root which is background-effect surfaces each blur independently, creating the weird edge mismatches 
I thought of other fix but couldn't implement(complex)    , which was changing the order to check first if they near each other , so instead of damaging or computing N=2(the problem) it could even be N=1 which is better performance instead , of  my whole monitor which is N=N , but it turned complex for me , I will try it later 

thanks for replying , I will try again before updating the pull

Ahmed5Emad:
## remade fix from the ground up 
instead of just computing the whole monitor like the old fix (band aid) , it pre-pass them into group/cluster so only adjacent or overlapped are grouped and computed one time  

it is now only computing the needed area so there is no performance issues even after test , did not notice any changes in my gpu usage , eliminating the sudden **50-150mb Vram** 

yes tested it and I am even using the version I built it right now 

image

vaxerski (commented):
how will this interact if the layers overlap? Won't (essentially) x-ray then look wrong?

Ahmed5Emad:
> how will this interact if the layers overlap? Won't (essentially) x-ray then look wrong?

no it does not first sorry for late response , second after they are clustered they are treated as one surface , they get alll effects as original (before grouping) , the only idea that they do group when the touching or overlapping so they processed as same surface , so the seam problem which happens because the edges which gets double the effect from two rendered surface stop to happening 

third I just tested it  

with xray
image

no xray 
image

I tested many things like adding many surface next to each other too and there was no problem

with xray : 
image

with no 
image


all of them with the blur

Ahmed5Emad:
## Summary after testing many things to make sure:
clustering or grouping doesn't actually change the visual behavior for overlapping surfaces because the blur source is always the main framebuffer (same as before).

no new xray visual breaking(tested). The orignal behavior (blurring wallpaper only) is preserved.
I tested some effects under decoration that I was not using to make sure nothing else will break .

UjinT34:
The root cause is that blur samples slightly outside of the cleared/rendered area and gets some random noise on the edge. Usually it's unnoticeable because of the very low weight for those random pixels. Seamless adjacent blur requires correct damage and scissoring of the blur area. If this is fixed then some invalidation and clearing shenanigans can be removed

Ahmed5Emad:
> The root cause is that blur samples slightly outside of the cleared/rendered area and gets some random noise on the edge. Usually it's unnoticeable because of the very low weight for those random pixels. Seamless adjacent blur requires correct damage and scissoring of the blur area. If this is fixed then some invalidation and clearing shenanigans can be removed

I don't know if it fixed this my idea was instead of processing every surface alone which end with the outside pixels processed twice which cause the random noise , why don't we just look for overlapped or touched surface and group them pre-pass them as one , so the noise from them reprocessing the edge disappear

Ahmed5Emad:
I did not really fix how the blur sample , i just told it those two surfaces is one there is no reason to do them separate just give them the same blur and because in noctalia they are identical material , opacity and color the line disappear

UjinT34:
The root cause isn't fixed with this PR. This one is a resource heavy band-and because every blur group will grab an extra fullscreen framebuffer and will hit `MAX_WORK_BUFFERS` if there are more than 5 regions (regular rendering with blur grabs at least 3 buffers)

Ahmed5Emad:
> The root cause isn't fixed with this PR. This one is a resource heavy band-and because every blur group will grab an extra fullscreen framebuffer and will hit `MAX_WORK_BUFFERS` if there are more than 5 regions (regular rendering with blur grabs at least 3 buffers)

but isn't MAX_WORK_BUFFERS   = 8 , and I will need 5-6 non-overlapping bars like the one I am using to fill it like if cluster take  1 buffer .

I didn't really think about that because , I really didn't think some one will open 5 separate regions just to fill it , but your concern is valid , I will try something in mind to fix the buffer problem if not working I will give up on this problem, I will update it  if  fixed.

anyway thanks , sorry

Ahmed5Emad:
## First : 
I looked at root cause I thought it was processing but it is like @UjinT34 said it is the samples slightly outside which cause the doable processing  and no I really can't fix this , it is deep for me (sorry)

## Second:
 I tried different approach  by just comping every new item into the same buffer which fix the ```MAX_WORK_BUFFERS```  now it will use only 1 buffer no buffer per region so it doesn't waste resource
I know it used buffer in last fix but I thought it used only one(I have one bar) didn't think about having different groups

## third:
yes tested it with xray and most of rules



### last : 
I am closing this PR anyway if this fix not working , I am kinda not even working on the main cause(as I thought I am) it just better band aid from the first one  (sorry for wasting your time) 😅 , I will just use it for personal use

PR #15135, opened 2026-06-15
desktop/groups: fix crash on dissolution

vaxerski:
Fixes a crash when a group is simply toggled lole


tekstryder:
Confirmed fixes:

- https://github.com/hyprwm/Hyprland/discussions/15205

PR #15137, opened 2026-06-15
fix: hide the top layer in workspace swipe to scrolling layout fullsc…

### LINKED ISSUE #15053: Top layer not covered after workspace gesture ends on scrolling fs
### Discussion

https://github.com/hyprwm/Hyprland/discussions/14721#discussioncomment-17224424

## Description

Scrolling
Fs
Swipe to another ws
Swipe to come back
Bar not hidden

Vic-41148:

Fixing bar visibility when swipe to a workspace containing a scrolling layout fullscreen 

Issue was caused by fullscreen checks relying on m_fullscreenMode and m_hasFullscreenWindow the code now workd out fullscreen state through getFullscreenWindow()

added isDirection(char) overload and sv.empty() 
Fixes #15053 and closes #14721

No

yes can review

### DIFF
--- a/src/managers/animation/DesktopAnimationManager.cpp
+++ b/src/managers/animation/DesktopAnimationManager.cpp
@@ -7,6 +7,7 @@
 #include "../../desktop/view/Window.hpp"
 #include "../../desktop/view/Group.hpp"
 #include "../../desktop/Workspace.hpp"
+#include "../../layout/target/Target.hpp"
 
 #include "../../config/shared/animation/AnimationTree.hpp"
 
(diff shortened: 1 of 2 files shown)


src/helpers/MiscFunctions.cpp
```diff
@@ -114,12 +114,19 @@ std::optional<float> getPlusMinusKeywordResult(std::string source, float relativ
...
+    if (sv.empty())
+        return false;
+
     if (sv[0] == 'l' || sv[0] == 'r' || sv[0] == 'u' || sv[0] == 'd' || sv[0] == 't' || sv[0] == 'b')
         return sv.length() == 1 || sv == "left" || sv == "right" || sv == "up" || sv == "down" || sv == "top" || sv == "bottom";
 
     return false;
 }
 
+bool isDirection(const char c) {
```
vaxerski: why

vaxerski (dismissed):

src/helpers/MiscFunctions.cpp
Vic-41148: passing char literals over to it can't convert to string_view implicitly so overloaded to make the tests to compile  fails otherwise :[  , only used by tests tho real calls uh here DispatcherTranslator.cpp line 269 and MonitorQueryCore.cpp line 202 do pass, should i wrap it at call site instead? will drop the overload ig

vaxerski: dont pass chars then, pass a string

Vic-41148: T - T that makes sense, rebased rom main, someone removed the overload upstream fixing isDirectionString left test case, but I added EXPECT_FALSE for "0" and " " we lose that from old char test

vaxerski (approved):

PR #15144, opened 2026-06-16
output: avoid unnecessary VCGT resampling

ghost:
Add a fast path to resampleInterleavedToKms() for cases where the ICC VCGT table already has the same number of entries as the KMS gamma LUT.

When both sizes match, every destination entry maps directly to the corresponding source entry, so interpolation, clamping and rounding are unnecessary. The new path directly interleaves the red, green and blue channels into the KMS output format.

The existing resampling path is kept for differing table sizes.


vaxerski (approved):
tnx

PR #15147, opened 2026-06-16
perf: avoid unnecessary copies and allocation

ghost:
Avoid several unnecessary copies and allocations.

This change:

allows CMonitorRule to use implicitly generated move operations by removing its explicitly defaulted destructor;
passes read-only PHLWINDOW, PHLWINDOWREF, PHLLS, and PHLLSREF parameters by const reference in small validation helpers;
reserves the required capacity before constructing the call stack vector.

The pointer-wrapper changes avoid redundant reference-count updates in frequently used helpers, while the CMonitorRule change allows its string members to be moved instead of copied.

### DIFF
--- a/src/config/shared/monitor/MonitorRule.hpp
+++ b/src/config/shared/monitor/MonitorRule.hpp
@@ -33,7 +33,6 @@ namespace Config {
     class CMonitorRule {
       public:
         CMonitorRule()  = default;
-        ~CMonitorRule() = default;
 
         eMonitorRuleComparisonResult compare(const CMonitorRule& other) const;
 
--- a/src/config/shared/monitor/MonitorRule.hpp
+++ b/src/config/shared/monitor/MonitorRule.hpp
@@ -32,7 +32,8 @@ namespace Config {
 
     class CMonitorRule {
       public:
-        CMonitorRule() = default;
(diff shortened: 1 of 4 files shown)


ItsOhen:
What? They are pointers are they not?

vaxerski:
@ItsOhen managed pointers, they still have to keep track of the control block.

vaxerski:
clang-foramt

ItsOhen:
Yeah. So what are the const and constref for?

ItsOhen:
Just to be clear, i'm not against this. Just curious about the thought process.

vaxerski:
copying a weak pointer goes to the control block to increase `wref` count, then destroying it decreases the count. That's not free. Refs don't do that.

src/config/shared/monitor/MonitorRule.hpp
```diff
@@ -32,8 +32,7 @@ namespace Config {
 
     class CMonitorRule {
       public:
-        CMonitorRule()  = default;
-        ~CMonitorRule() = default;
+        CMonitorRule() = default;
```
vaxerski: why this change?

vaxerski (approved):

PR #15165, opened 2026-06-17
input/window: fix dragging pointer-locked and fullscreen game windows (gamescope)

ZStud:
Three small focused fixes around dragging game windows, especially gamescope-wrapped ones. Kept as separate commits so they can be taken or dropped independently.

1. Don't re-fullscreen a window mid-drag. A client re-asserting fullscreen every frame (gamescope) snaps the window back the instant the drag pulls it out, so it oscillates and the move crawls. Ignored while it's the active drag target.

2. Ignore a pointer lock on a dragged window. A pointer-locked game couldn't be moved/resized; the lock warped the cursor back to the hint every frame. The active drag target is treated as unconstrained so the cursor drives the drag, and the lock resumes when the drag ends.

3. Don't send relative motion to a dragged window. An interactive grab should be exclusive; without this a locked game pans its camera from the drag motion.

Tested with gamescope games: dragging, resizing and tiling now behave like any normal window.

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -1350,7 +1350,12 @@ void CWindow::onUpdateState() {
     std::optional<MONITORID> requestsID = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsFullscreenMonitor : MONITOR_INVALID;
     std::optional<bool>      requestsMX = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsMaximize : m_xwaylandSurface->m_state.requestsMaximize;
 
-    if (requestsFS.has_value() && !(m_suppressedEvents & SUPPRESS_FULLSCREEN)) {
+    // a client re-asserting fullscreen every frame (notably gamescope) would snap the window back the instant
+    // the drag handler pulls it out of fullscreen, so ignore the request while this window is being dragged.
+    const auto DRAGTARGET         = g_layoutManager->dragController()->target();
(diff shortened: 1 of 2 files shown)


src/desktop/view/Window.cpp
```diff
@@ -1350,7 +1350,9 @@ void CWindow::onUpdateState() {
     std::optional<MONITORID> requestsID = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsFullscreenMonitor : MONITOR_INVALID;
     std::optional<bool>      requestsMX = m_xdgSurface ? m_xdgSurface->m_toplevel->m_state.requestsMaximize : m_xwaylandSurface->m_state.requestsMaximize;
 
-    if (requestsFS.has_value() && !(m_suppressedEvents & SUPPRESS_FULLSCREEN)) {
+    // a client re-asserting fullscreen every frame (notably gamescope) would snap the window back the instant
+    // the drag handler pulls it out of fullscreen, so ignore the request while it's the active drag target.
+    if (requestsFS.has_value() && !(m_suppressedEvents & SUPPRESS_FULLSCREEN) && g_layoutManager->dragController()->target() != m_self) {
```
vaxerski: target() != m_self is comparing layout target to a window, this is wrong

ZStud: Oh my bad, I wrote each of these up first and then tested, and it all worked, so I didn't even notice this one wasn't actually doing anything. Should be fixed now, it compares the drag target's window to this one (DRAGTARGET->window() == m_self.lock()), same as the check in the pointer-lock commit.

vaxerski (approved):

PR #15193, opened 2026-06-20
config: monitors default to "preferred" res, "auto" pos and scale

davc0n:
It is now possible to omit the keys entirely.

This way common monitors config entries can be more concise.


vaxerski (approved):
tnx

PR #15197, opened 2026-06-21
Do not wrap std::function in std::optional

danielkrupinski:


vaxerski (approved):

zjeffer:
Doesn't compile anymore on Arch after this PR

https://github.com/hyprwm/Hyprland/actions/runs/27901687640/job/82563006635 

EDIT: Actually, looks like it's caused by https://github.com/hyprwm/hyprutils/commit/9d8bf6e810597152eef8906c670b96679af2faec

PR #15213, opened 2026-06-22
Fix: nested submaps and resets

ItsOhen:
After lua and the changes to handlers, we can no longer match only k->handlers to break out if the submap is changed / reset. So fixes that by storing what submap we were in before the handler was called.

Could impact performace, but shouldn't be a lot. Still need a rewrite of the entire manager imho.

Yes.


vaxerski (approved):
rewrite would be great but the big thing in keybindmgr is just that it needs to do so many weird things all at once

PR #15217, opened 2026-06-23
misc: modernize truthy(fix possible UB)

b1tflyyyy:
Modernizes the truthy function by switching to std::views::transform to cut out heap allocations, and fixes potential UB by casting characters to unsigned char before std::tolower.

I've wrapped the main logic in clang-format off/on comments. The default formatting was messing up the lambda and fold expression layout, so this was done purely to keep the code clean and readable.

Ready for merging.


b1tflyyyy:
It looks like the CI build failed because std::ranges::starts_with requires GCC 16+, while the current CI is using GCC 15.2.0.

Should I rewrite this in the other way, or is there a plan to bump the compiler version?

feelamee:
> I've wrapped the main logic in clang-format off/on comments. The default formatting was messing up the lambda and fold expression layout, so this was done purely to keep the code clean and readable.

IMHO, it was much more readable without lambda and variadics..
This looks like unnecessary complexity here

b1tflyyyy:
> > I've wrapped the main logic in clang-format off/on comments. The default formatting was messing up the lambda and fold expression layout, so this was done purely to keep the code clean and readable.
> 
> IMHO, it was much more readable without lambda and variadics.. This looks like unnecessary complexity here

IMHO, I find it easier to have a small generic lambda and avoid repeating the same call multiple times with different arguments by hand. It also makes adding new truthy prefixes later a one-line change.

src/helpers/MiscFunctions.cpp
```diff
@@ -819,11 +819,18 @@ std::string getBuiltSystemLibraryNames() {
...
+    using std::operator""sv;
+
+    if (str == "1"sv)
         return true;
 
-    std::string cpy = str;
-    std::ranges::transform(cpy, cpy.begin(), ::tolower);
+    // clang-format off
+    auto str_view = str | std::views::transform([](unsigned char ch) -> char {
+        return static_cast<char>(std::tolower(ch));
```
vaxerski: sc<> from hu

vaxerski:
@fufexan when gcc 16 in ci?

fufexan:
In a few minutes if it builds fine.

src/helpers/MiscFunctions.cpp
b1tflyyyy: > sc<> from hu

Could you please explain what this means?

fufexan:
Pushed, please rebase on main.

src/helpers/MiscFunctions.cpp
fufexan: static cast from hyprutils

b1tflyyyy: > static cast from hyprutils

so, 
```cpp
...
return sc<char>(std::tolower(ch));
```
yes?

fufexan: Yes, should be fine

vaxerski (approved):

PR #15229, opened 2026-06-25
Handle color state for direct scanout

ghost:
## Description

Part of #9064.

This updates direct scanout so fullscreen color state is applied before attempting scanout, and so the output state is rolled back if scanout cannot be committed. When the scanout buffer is unchanged but the color state changed, the direct scanout path now commits the same buffer update instead of presenting feedback without applying the new output state.

It also blocks SDR direct scanout when luminance conversion would be required. CTM can cover compatible primaries, but direct scanout bypasses shader luminance conversion, so luminance mismatches need to fall back to the composited path.

## User impact

- HDR fullscreen clients can direct-scanout with their HDR metadata applied to the output state.
- SDR/HDR transitions around direct scanout keep the output color state in sync.
- Direct scanout no longer silently skips required SDR luminance conversion.

## Validation

- `cmake --fresh -S . -B build`
- `cmake --build ./build --target Hyprland -j2`
- `git diff --check`

Note: the fresh CMake configure was needed because the local build cache still referenced an older Abseil library name (`absl_borrowed_fixup_buffer`) after the system packages had already moved on.

### DIFF
--- a/src/managers/SeatManager.cpp
+++ b/src/managers/SeatManager.cpp
@@ -215,11 +215,12 @@ void CSeatManager::setKeyboardFocus(SP<CWLSurfaceResource> surf) {
     const auto& PRESSED = g_pInputManager->getKeysFromAllKBs();
     static_assert(std::is_same_v<std::decay_t<decltype(PRESSED)>::value_type, uint32_t>, "Element type different from keycode type uint32_t");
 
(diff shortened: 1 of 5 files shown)


UjinT34:
DS SDR luminance differences are negligible unless DS is toggled back and forth. At least keep a way to ignore this diff with a config setting.

PR #15230, opened 2026-06-25
treewide: clean up  and performance 

ghost:
## Summary

- replace scanner-flagged raw wl_array memcpy paths with typed std::ranges::copy after a single wl_array_add
- clean up low-risk hyprpm copies by passing repo identifiers by const reference and trimming strings in place
- initialize small state fields more defensively and remove logging/formatting from the XCB connection destructor

### DIFF
--- a/hyprpm/src/core/PluginManager.cpp
+++ b/hyprpm/src/core/PluginManager.cpp
@@ -16,6 +16,7 @@
 #include <fstream>
 #include <algorithm>
 #include <format>
+#include <string_view>
 
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -388,7 +389,7 @@ bool CPluginManager::addNewPluginRepo(const std::string& url, const std::string&
     return true;
 }
 
-bool CPluginManager::removePluginRepo(const SPluginRepoIdentifier identifier) {
+bool CPluginManager::removePluginRepo(const SPluginRepoIdentifier& identifier) {
     if (!DataState::pluginRepoExists(identifier)) {
         std::println(stderr, "\n{}", failureString("Could not remove the repository. Repository is not installed."));
         return false;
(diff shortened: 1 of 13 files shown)


src/xwayland/XWM.hpp
```diff
@@ -71,12 +71,8 @@ class CXCBConnection {
     }
 
     ~CXCBConnection() {
-        if (m_connection) {
-            Log::logger->log(Log::DEBUG, "Disconnecting XCB connection {:x}", rc<uintptr_t>(m_connection));
+        if (m_connection)
```
vaxerski: why are you removing the log?

ghost: I removed it to avoid logging from a destructor during teardown and because setting the pointer to null in the destructor is redundant, the object will be destroid immediately afterwards. But I can keep the debug/error logs if they are useful for diagnostics.

hyprpm/src/core/PluginManager.cpp
```diff
@@ -459,9 +459,12 @@ eHeadersErrors CPluginManager::headersValid() {
...
     std::string hash = verHeaderContent.substr(HASHPOS + 23);
-    hash             = hash.substr(0, hash.find_first_of('\n'));
-    hash             = hash.substr(hash.find_first_of('"') + 1);
-    hash             = hash.substr(0, hash.find_first_of('"'));
+    if (const auto POS = hash.find_first_of('\n'); POS != std::string::npos)
+        hash.erase(POS);
+    if (const auto POS = hash.find_first_of('"'); POS != std::string::npos)
+        hash.erase(0, POS + 1);
+    if (const auto POS = hash.find_first_of('"'); POS != std::string::npos)
+        hash.erase(POS);
```
danielkrupinski: Just use `std::string_view`  - its .substr() does not allocate memory.

hyprpm/src/core/PluginManager.cpp
```diff
@@ -587,9 +599,12 @@ bool CPluginManager::updateHeaders(bool force) {
     if (ret.contains("CMake Error at")) {
         // missing deps, let the user know.
         std::string missing = ret.substr(ret.find("CMake Error at"));
-        missing             = ret.substr(ret.find_first_of('\n') + 1);
-        missing             = missing.substr(0, missing.find("-- Configuring incomplete"));
-        missing             = missing.substr(0, missing.find_last_of('\n'));
+        if (const auto POS = missing.find_first_of('\n'); POS != std::string::npos)
```
vaxerski: .contains()

vaxerski:
@codex review please

vaxerski (approved):

PR #15243, opened 2026-06-26
presentation: associate feedback with committed surface state

ghost:
## What

Tie wp_presentation feedback objects to committed surface state instead of only to the surface. The change also discards feedback for superseded commits, rejected/queued state cleanup, surface destroy, and monitor removal.

## Why

When multiple commits are coalesced or destroyed before presentation, feedback from older or discarded state can otherwise be reported as presented or remain pending.

## Validation

- `git diff --check upstream/main..pr/presentation-feedback-committed-state`
- `cmake --build build --target Hyprland -j$(nproc)`
- `build/hyprtester/hyprtester -c hyprtester/test.lua -b ./build/Hyprland -p hyprtester/plugin/hyprtestplugin.so fullscreenScanoutStressTransitions` passed when run with the dependent stress-test branch and damage fix applied.

## Dependency

Independent production fix, but the separate fullscreen scanout stress-test PR is expected to exercise and protect this behavior.


src/protocols/types/SurfaceStateQueue.cpp
```diff
@@ -1,10 +1,16 @@
 #include "SurfaceStateQueue.hpp"
 #include "../core/Compositor.hpp"
+#include "../PresentationTime.hpp"
 #include "SurfaceState.hpp"
 
 CSurfaceStateQueue::CSurfaceStateQueue(WP<CWLSurfaceResource> surf) : m_surface(std::move(surf)) {}
 
 void CSurfaceStateQueue::clear() {
+    if (PROTO::presentation) {
```
vaxerski: can presentation ever be null? It's an unconditional protocol.

src/protocols/PresentationTime.hpp
```diff
@@ -64,7 +69,7 @@ class CPresentationProtocol : public IWaylandProtocol {
 
     //
     std::vector<UP<CWpPresentation>>         m_managers;
-    std::vector<UP<CPresentationFeedback>>   m_feedbacks;
+    std::vector<SP<CPresentationFeedback>>   m_feedbacks;
```
vaxerski: not a good idea - compositor state should not hold resources hostage

umbrageodotus:
@NotPppp1116 should this be merged or the other one be merged?

ghost:
> @NotPppp1116 should this be merged or the other one be merged?

merge this one

umbrageodotus:
okay, btw do u still have discord or did u fully delete everything

src/protocols/types/SurfaceState.hpp
```diff
@@ -77,6 +78,9 @@ struct SSurfaceState {
     // for wl_surface::frame callbacks.
     std::vector<SP<CWLCallbackResource>> callbacks;
 
+    // for wp_presentation feedbacks, tied to this commit.
+    std::vector<CPresentationFeedback*> presentationFeedbacks;
```
vaxerski: no raw ptrs allowed

ghost:
> okay, btw do u still have discord or did u fully delete everything

I have,u can dm IVE been cleaning up and leaving a lot of servers etc

umbrageodotus:
> I have,u can dm IVE been cleaning up and leaving a lot of servers etc

iirc I got blocked by u after u sent that file

src/protocols/PresentationTime.cpp
```diff
@@ -95,41 +110,48 @@ void CPresentationProtocol::onManagerResourceDestroy(wl_resource* res) {
 }
 
 void CPresentationProtocol::destroyResource(CPresentationFeedback* feedback) {
-    std::erase_if(m_feedbacks, [&](const auto& other) { return other.get() == feedback; });
+    feedback->m_done = true;
```
vaxerski, line 113: why are we keeping this around? we're potentially hogging a resource.

src/protocols/PresentationTime.hpp
vaxerski: I dont see any place where a SP<> is needed? why cant be up?

ghost: SP provides the backing ownership for the WP references stored in SSurfaceState and CQueuedPresentationData. Those weak references ensure compositor state does not keep the protocol resource alive, UP here would require raw non-owning pointers which the previous review rejected.

gulafaran: > SP provides the backing ownership for the WP references stored in SSurfaceState and CQueuedPresentationData. Those weak references ensure compositor state does not keep the protocol resource alive, UP here would require raw non-owning pointers which the previous review rejected.

UP from hyprutils can use WP from hyprutils, its not std::

erstarr: yeah hyprland's WP<> is great.

Just remember that you can't .lock() them since they are a WP of a UP after all

vaxerski: as above

vaxerski (approved):

tannerellen:
This breaks dead space in steam for me. Error on launch and quits. Crashes on this pull request but all commits before this one are good. 
image

vaxerski:
reverted

tannerellen:
> reverted

Tom is looking into this we are chatting in hyprland-dev. Might just be something weird in the game. Looks like it only has this issue when launched in Wayland. My concern was if this was a bigger issue and affected other directx 12 games. If it's not a more widespread issue or doesn't indicate some other underlying problem then the work around on this specific game is to launch the game in xwayland.

vaxerski:
I'd rather not break things though. We can investigate further and then decide what to do.

PR #15242, opened 2026-06-26
renderer: reject invalid damage before region handling

ghost:
## What

Reject zero/negative damage and clamp extreme damage coordinates before adding wl_surface damage to `CRegion`/Pixman-backed paths.

## Why

Clients can submit invalid or very large damage rectangles. Validating before region construction avoids feeding invalid rectangles into region handling.

## Validation

- `git diff --check upstream/main..pr/damage-clip-invalid-regions`
- `cmake --build build --target Hyprland -j$(nproc)`

## Notes

This is independent of the fullscreen scanout stress test; the stress test is proposed separately and covers this path under churn.

PR #15244, opened 2026-06-26
hyprtester: add fullscreen scanout transition stress test

ghost:
## What

Add a new hyprtester client and scenario, `fullscreenScanoutStressTransitions`, plus plugin assertions for render and presentation state.

The test drives fullscreen/game-like load while switching workspaces, destroying surfaces, toggling monitor state, submitting invalid damage, and checking presentation feedback cleanup.

## Why

This protects the production fixes proposed in the related PRs:

- invalid damage must be rejected before region handling
- presentation feedback must follow committed state and be discarded when commits/surfaces/queued state are superseded or destroyed
- inactive workspaces must not be rendered unnecessarily during fullscreen/game-like load
- Hyprland must still recover cleanly when leaving fullscreen, switching workspaces, destroying the surface, or changing monitor state

## Limitations

Hyprtester runs headless and this client is SHM-backed, so it does not prove hardware DMABUF/direct scanout on a real output. It still exercises fullscreen/direct-scanout transition logic, presentation feedback, damage handling, workspace switching, monitor toggling, and surface destruction under heavy rendering churn.

## Validation

- `git diff --check upstream/main..pr/fullscreen-scanout-stress-test`
- `cmake --build build/hyprtester -j$(nproc)`
- `build/hyprtester/hyprtester -c hyprtester/test.lua -b ./build/Hyprland -p hyprtester/plugin/hyprtestplugin.so fullscreenScanoutStressTransitions` passed when run with the dependent production fixes applied.

## Dependencies

This PR should be reviewed/merged after:

1. `hyprtester: handle empty Lua include path in plugin build`
2. `renderer: reject invalid damage before region handling`
3. `presentation: associate feedback with committed surface state`

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -11,6 +11,7 @@
 #include <src/managers/SeatManager.hpp>
 #include <src/managers/input/trackpad/TrackpadGestures.hpp>
 #include <src/output/Monitor.hpp>
+#include <src/render/Renderer.hpp>
 #include <src/desktop/rule/windowRule/WindowRuleEffectContainer.hpp>
 #include <src/desktop/rule/layerRule/LayerRuleEffectContainer.hpp>
 #include <src/desktop/rule/windowRule/WindowRuleApplicator.hpp>
@@ -18,6 +19,8 @@
 #include <src/desktop/state/WindowState.hpp>
 #include <src/Compositor.hpp>
 #include <src/desktop/state/FocusState.hpp>
+#include <src/event/EventBus.hpp>
+#include <src/protocols/PresentationTime.hpp>
 #include <src/state/MonitorState.hpp>
 #include <src/layout/LayoutManager.hpp>
 #undef private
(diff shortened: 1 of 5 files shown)


hyprtester/plugin/src/main.cpp
```diff
@@ -584,6 +594,53 @@ static SDispatchResult floatingFocusOnFullscreen(std::string in) {
...
+    } catch (...) { return {.success = false, .error = "invalid input"}; }
+
+    const uint64_t count = g_renderStats.windowRenders.contains(cls) ? g_renderStats.windowRenders.at(cls) : 0;
+    if (count > max)
+        return {.success = false, .error = std::format("Expected at most {} renders for '{}', got {}", max, cls, count)};
+
+    return {};
+}
+
+static SDispatchResult expectPresentationPendingMax(std::string in) {
```
vaxerski, line 624: this way of doing things is quite pointless? Why not simply return pending presentations and do the expect in the tester?

ghost: It's not pointless u need to have a look at #15242 #15243 #15241

vaxerski: ah, because you call from lua. You could make it an exposed fn instead then, but it's fine.

PR #15260, opened 2026-06-28
desktop/fadeout: extract bad logic, simplify fadeouts

vaxerski:
This is a touchy, but much needed refactor:

- unifies fadeouts into one single type
- removes spaghetti logic keeping objects alive when they werent
- cleanups in general

Needs testing of various animated shit, fadeouts, dims, etc.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -261,11 +261,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1782035033,
-        "narHash": "sha256-pUtCphVzH1iNUTMGdJr4+e/yzUXA6/DZwsn+cyfIVJU=",
+        "lastModified": 1782639496,
+        "narHash": "sha256-mjt47Xun3jPcGrXpGpYU85lzUnqvJ2rXLbQjVI1yvLo=",
         "owner": "hyprwm",
         "repo": "hyprutils",
-        "rev": "9d8bf6e810597152eef8906c670b96679af2faec",
+        "rev": "d2b40ffe7bfcb001bbf5888bb56ff646de53e7db",
         "type": "github"
       },
       "original": {
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -1186,7 +1186,7 @@ void CCompositor::setWindowFullscreenState(const PHLWINDOW PWINDOW, Desktop::Vie
     // make all windows and layers on the same workspace under the fullscreen window
     for (auto const& w : Desktop::windowState()->windows()) {
         if (w->m_workspace == PWORKSPACE) {
-            if (!w->isFullscreen() && !w->m_fadingOut && !w->m_pinned)
+            if (!w->isFullscreen() && !w->m_pinned)
                 w->m_createdOverFullscreen = false;
 
             w->updateFullscreenInputState();
--- a/src/desktop/DesktopTypes.hpp
+++ b/src/desktop/DesktopTypes.hpp
@@ -10,6 +10,7 @@ namespace Desktop::View {
     class IView;
     class CWindow;
     class CLayerSurface;
+    class CPopup;
 }
 
 namespace Monitor {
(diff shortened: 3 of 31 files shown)


kolayne (commented):
`rofi` won't launch:
```
wl_surface#3: error -1: layerSurface was not configured, but a buffer was attached
```

The `waybar` process launches and the output seems fine but the bar won't appear.

vaxerski:
layers should be fixed. I also bumped hyprutils as one bug in the animations there was a bug in hu

kolayne:
When rofi closes, it jumps to top left corner and slides to the center while fading :D

vaxerski:
yes, update hyprutils

kolayne:
Updated all dependencies to their git versions, clean-built this PR. If I launch hyprlock, hyprlock crashes.

Btw, on the topic of lock app crash (let me know if I should open a separate discussion):
photo_2026-06-29_00-01-09
1. The number in the top left corner is incorrect: I'm on tty1, not tty0. Launch command was `LANG=ko_KR.UTF-8 uwsm start Docs/Hyprland/build/Hyprland`
2. These commands only work for the legacy config. Since I'm running with a lua config, I should probably see commands for lua.
3. Is there a reason why we have to put the user through these complex steps, rather than giving a single command to unlock and go back to Hyprland? Like, in my case even if I relaunch hyprlock, it crashes again. I wish there was a dedicated action or dispatcher to exit this mode.

vaxerski:
1. oops maybe we should +1
2. I think we updated them recently
3. well no not really because we dont know what lockscreen they are using

vaxerski:
I can't repro the crash. Is this a regression from this MR?

kolayne:
> well no not really because we dont know what lockscreen they are using

Why does it matter? The screen above is drawn by Hyprland itself, right? So there could be a command to unlock it?


> I can't repro the crash. Is this a regression from this MR?

Oops, my bad! Just forgot to rebuild hyprlock. Works now.

kolayne:
Rofi, notifications and similar elements now closes fine, but I found that this pop up from discord (appears when I hover a link), when disappearing, jumps to the top left corner and slides to the center of the screen.

vaxerski:
uhh what popup? Any other popup I can verify? I've never seen any popups from discord, prolly cuz I use vencord

kolayne:
So far, haven't encountered it anywhere else. I'll let you know if I do

vaxerski:
I'll merge this because I want to continue the refactors, if that is widely reproducible I'll fix it in a followup.

fxzzi:
cooked my animations - before and after:


https://github.com/user-attachments/assets/0a8dd08b-4dee-4061-9a2d-772644cb06d3


https://github.com/user-attachments/assets/cc87da3c-5d33-4558-be41-b67faec2d096

fxzzi:
i think it's this PR that also broke no_anim on fadeout

```lua
hl.layer_rule({ match = { namespace = "selection" }, no_anim = true, order = -2 })
```

no anim on launch, but there is still an animation on exit.

and below is a slow mo of the fade

https://github.com/user-attachments/assets/f90addd7-5014-4e46-ae91-837c27e61710

vaxerski:
I don't know was I not fucking clear?

image

fxzzi:
im on nix and my hyprutils was updated by your PR kappachungus https://github.com/hyprwm/Hyprland/blob/254d6bae775258b554e1d3a4a853f3bc244190d9/flake.lock

vaxerski:
literally propaganda

vaxerski:
I pushed a speculative fix to hyprutils

fxzzi:
fixed, fade out looks right. no_anim still borken :)

actually no, half my animations are now borken also. moving window has no animations. a window becoming smaller due to a new window opening is not smooth and it snaps to its smaller size

https://github.com/user-attachments/assets/a4895548-0461-4ae5-bcf1-f63df2253f42

vaxerski:
I cant repro any of this fazzi.

fxzzi:
AAAAAAAAAAAAAA ok I'll find repro at some point

For now these are my curves and anima maybe try? https://gitlab.com/fazzi/nixohess/-/blob/main/modules/programs/hyprland/settings.nix?ref_type=heads#L277

vaxerski:
check https://github.com/hyprwm/Hyprland/pull/15302

PR #15261, opened 2026-06-28
Add code owners via codeowners-plus

kolayne:

Adds the code owners reviewer assignment logic via the [codeowners-plus](https://github.com/marketplace/actions/codeowners-plus) action. Unlike GitHub's built-in CODEOWNERS, codeowners-plus allows to specify users without write access to the repository.

Codeowners-plus recommends that each subdirectory includes its own `.codeowners` file. But I think it's more clear when everything is in one place, at the root of the repository.

Ready for merge. CC @fufexan

### DIFF
--- a/.codeowners
+++ b/.codeowners
@@ -1,0 +1,10 @@
+# Reference: https://github.com/marketplace/actions/codeowners-plus#codeowners-file-spec
+
+# Fallback: request Vaxry's review
+* @vaxerski
+
+# CI: request @fufexan's review
+.github/workflows/** @fufexan
+
+# Hyprtester (framework): CC @kolayne
+? hyprtester/src/* @kolayne
--- a/.github/workflows/codeownersplus.yml
+++ b/.github/workflows/codeownersplus.yml
@@ -1,0 +1,33 @@
+name: 'Code Owners'
+
+concurrency:
+  group: codeowners-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    # Note: to support more features of codeowners-plus, more event triggers
+    # may be required. See https://github.com/marketplace/actions/codeowners-plus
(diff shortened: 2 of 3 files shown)


vaxerski:
will this basically assign the person automatically?

kolayne:
For lines that begin with `?` (me and hyprtester), it will just leave a comment like this: https://github.com/kolayne/test/pull/3#issuecomment-4750326498

For default lines (fufexan and CI; you and everything else), it will request the user's review and leave a comment like this: https://github.com/kolayne/test/pull/5#issuecomment-4825983385

I just realized that this second comment may cause confusion, since it says that an approval from fufexan is _required_, which is not currently the case, is it?

Anyway, there are three ways in which we can configure it (with fufexan&CI as an example):

1. Codeowner requires an approval from fufexan on PRs that touch CI. If there's no approval from him, the codeowners workflow will fail (you may still merge the PR).
2. Coderowner won't enforce its rules, but will leave a comment like above, telling that fufexan's approval is required (but the check always succeeds).
3. We convert fufexan into an optional reviewer (same logic as me and hyprtester), and then the bot will just CC him in a comment.

<hr>

Also, how do you want it to behave for default PRs? Options:

1. Request your review and leave a comment that requires your approval
2. CC you
3. Do nothing at all

fufexan:
Can it simply request my review? I don't wanna get two emails for the same thing.

kolayne:
If it CCs, it only leaves a comment and that's it. If it assigns a reviewer, it also leaves a comment :(

Though, maybe upstream will accept a patch to support that.

vaxerski:
yeah I agree a mention is redundant

kolayne:
Fixed. Required reviewer (the one without a question mark) is now just requested to review, without a comment, like in https://github.com/kolayne/test/pull/7

vaxerski (approved):
tnx

PR #15262, opened 2026-06-28
eventmanager: handle partial IPC writes

ghost:
## What changed
- Track a write offset for each socket2 IPC client.
- Flush queued events through one helper that handles short writes, EAGAIN/EWOULDBLOCK, and EINTR.
- Keep WL_EVENT_WRITABLE enabled only while unsent event data remains.

## Why
The old socket2 path treated any non-negative write as a complete event write. A short socket write could silently drop the tail of an IPC event.

## Validation
- git diff --check
- cmake --build build -j2


vaxerski (approved):

vaxerski:
@codex review please

PR #15263, opened 2026-06-28
hyprctl: avoid blocking compositor loop

ghost:
## What changed
- Make the hyprctl server socket and accepted clients nonblocking.
- Move accepted clients into wl_event_loop-managed state objects.
- Read requests and write replies without poll/sleeping in the accept callback.
- Preserve pending promise replies and rolling-log follow handoff.
- Add an explicit request terminator for in-tree clients and a bounded 10 ms quiet-period fallback for legacy unframed clients.
- Complete partial client writes instead of assuming one `write()` sends the whole request.

## Why
The old hyprctl callback accepted a client, waited up to 5 seconds in poll(), then synchronously read and wrote the reply from the compositor event loop. A slow or stalled client could block compositor progress.

Treating `EAGAIN` as the end of a request could also process a truncated command while a client was still writing. Framed in-tree requests now wait for their terminator; older clients retain compatibility through the short quiet-period fallback.

## Validation
- `git diff --check`
- changed-line `clang-format --diff`
- prior branch validation: `cmake --build build -j2`
- fresh GitHub CI requested for the review follow-up


vaxerski:
@codex review please

vaxerski:
@codex review please

vaxerski:
comflic

vaxerski (approved):
nice. Tests please.

PR #15264, opened 2026-06-28
window: preserve final chunk when reading process environment

ghost:
## What changed
- Preserve the final short read from `/proc/<pid>/environ` via `gcount()`.
- Trim the parsing buffer to the bytes actually read before splitting entries.

## Why
The previous loop discarded the final short read. Environment data remains read on every call, so runtime changes are not hidden by a cache.

## Validation
- Merged current `main` without unresolved conflicts.
- `git diff --check` passes for the changed files.


vaxerski (commented):
what if the env changes? This can happen.

vaxerski (approved):

PR #15275, opened 2026-06-29
[codex] Fix protocol null dereference paths

ghost:
## What changed

Fixes two protocol paths that could dereference missing wrapper objects while handling client requests:

- `wp_color_manager_v1.get_image_description`: when the referenced image description lookup fails, create the requested image-description object and send the failure on it instead of dereferencing the missing old reference.
- `zxdg_output_manager_v1.get_xdg_output`: validate `CWLOutputResource::fromResource(outputResource)` before using the returned output wrapper.


src/protocols/ColorManagement.cpp
```diff
@@ -185,7 +185,17 @@ CColorManager::CColorManager(SP<CWpColorManagerV1> resource) : m_resource(resour
 
         const auto OLD_RES = CColorManagementImageDescription::fromReference(ref);
         if (!OLD_RES) {
-            OLD_RES->resource()->sendFailed(WP_IMAGE_DESCRIPTION_V1_CAUSE_UNSUPPORTED, "Not found");
+            const auto RESOURCE = PROTO::colorManagement->m_imageDescriptions.emplace_back(
```
vaxerski: excuse me what?

vaxerski (approved):

PR #15311, opened 2026-07-03
Global  window counting

vacchus:

This PR adds a new selector flag `a` (ALL) to the `w[...]` workspace selector.

When `a` is present, the window/group count evaluation is performed across all workspaces instead of only the current workspace.

- With `a`: counts are aggregated globally across all workspaces in `WorkspaceState`.
- Without `a`: behavior remains unchanged and workspace-local.

This applies to both: `getWindowCount(...)` and `getGroups(...)`

---

- The implementation iterates over all workspaces when `a` is present, which may introduce a performance cost in large workspace setups.
- Behavior of existing selectors is fully preserved when `a` is not used.

---

Ready for review.

### DIFF
--- a/src/desktop/Workspace.cpp
+++ b/src/desktop/Workspace.cpp
@@ -257,7 +257,7 @@ bool CWorkspace::matchesStaticSelector(const std::string& selector_) {
 
                 prop = prop.substr(2, prop.length() - 3);
 
-                int  wantsOnlyTiled    = -1;z
+                int  wantsOnlyTiled    = -1;
                 int  wantsOnlyPinned   = false;
                 bool wantsCountGroup   = false;
                 bool wantsCountVisible = false;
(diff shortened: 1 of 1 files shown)


vaxerski:
...what's the point?

vacchus:
> ...what's the point?

 It is meant to allow window/group counts to be evaluated across all workspaces, rather than only the currently active workspace.

vacchus:
Sorry, is there already a solution  for this which im not aware of?@vaxerski

erstarr:
Get all workspaces with lua, then get all windows within each with lua.

You can tie this lua funct to workspace change event

vacchus:
> Get all workspaces with lua, then get all windows within each with lua.
> 
> You can tie this lua funct to workspace change event

Yeah, that was my initial approach with hl.on, but it ended up causing lag when I had multiple workspaces open, each with several active windows. And don't even get me started on lua's io.popen

PR #15319, opened 2026-07-04
input: fix wheel scroll events from virtual pointers being dropped

Nicholas-Arthur-Cook:
Fixes https://github.com/hyprwm/Hyprland/discussions/15318, which is 2 inter-connected issue. The first being the `timeMs == 0` events getting dropped silently, and the second being wheel-source events with no discrete value getting dropped.

- Tested it locally. 
    - At first I tried just skipping the emulation branch in `InputManager.cpp` for smooth-only events (`deltaDiscrete == 0 && delta != 0`). My reasoning being that there's no discrete data to try and emulate. And that mostly worked with my apps, except for kitty, which made me realize Hyprland always passes `axis_value120` even if the value is zero, and kitty was using that value and ignoring delta. So I adjusted to account for that by synthesizing the missing discrete value
- Made sure `clang-format` passes

I think it's ready! I haven't written C++ code in a while so apologies if there are any obvious bugs.


vaxerski:
@codex review please

vaxerski (approved):

PR #15332, opened 2026-07-05
monitor: skip fallback monitor on recovery flag evaluation

oclaw:

In cases when one of the monitors attached to machine rapidly connects and disconnects - fallback mode might be activated and fallback monitor gets emplaced into compositor monitor list. Though it is incorrect to treat this fallback monitor as a usual display while evaluating if the orphaned workpsaces need to be recovered. This patch fixes that

Tested on my machine and it seems to be helpful. Did not find unit tests for `RECOVERY` monitor mode in general though


vaxerski (dismissed):
tnx

oclaw:
@vaxerski thanks for quick review! Noticed that there was some refactoring after my patch (tested against released 0.55.4). Pushed fix, could you re-approve pls?

vaxerski (dismissed):

vaxerski:
clang-format pls

oclaw:
done

vaxerski (approved):

oclaw:
@vaxerski the test failure does not seem to be related to my changes :/
Could we re-trigger the build?

PR #15334, opened 2026-07-06
fix: display lock died screen on restart after crash

### LINKED ISSUE #15314: Automatically drop back to a lock died screen on restart after a crash
### Discussion link

https://github.com/hyprwm/Hyprland/discussions/15308

## Description

When hyprland crashes and start restarts it, unlocked.

neal98:

Fixes a case where Hyprland started with start-hyprland restarts unlocked after crashing with a locked session.

This expands the communication from Hyprland to start-hyprland and introduces a new Hyprland CLI flag to allow the restarted instance to preserve the locked state. The displayed screen is the existing `lockdead.png` crash screen.

Fixes [#15314](https://github.com/hyprwm/Hyprland/issues/15314)

No.

It's ready, please review.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -38,7 +38,7 @@ static void help() {
     --verify-config              - Do not run Hyprland, only print if the config has any errors
     --version           -v       - Print this binary's version
     --version-json               - Print this binary's version as json
-    --locked-crash               - Used by start-hyprland)#");
+    --locked [COMMAND]           - Launches locker on startup via the provided command)#");
 }
 
 static void reapZombieChildrenAutomatically() {
@@ -71,8 +71,9 @@ int main(int argc, char** argv) {
     // parse some args
     std::string configPath;
     std::string socketName;
+    std::string startLockedCommand;
     int         socketFd   = -1;
-    bool        ignoreSudo = false, verifyConfig = false, safeMode = false, lockedCrash = false;
+    bool        ignoreSudo = false, verifyConfig = false, safeMode = false, startLocked = false;
     int         watchdogFd = -1;
 
     if (argc > 1) {
@@ -177,8 +178,17 @@ int main(int argc, char** argv) {
                     help();
                     return 1;
                 }
-            } else if (value == "--locked-crash") {
-                lockedCrash = true;
+            } else if (value == "--locked") {
+                startLocked = true;
+
+                if (std::next(it) != args.end()) {
+                    startLockedCommand = *std::next(it);
+                    if (!startLockedCommand.starts_with("-"))
+                        it++;
+                    else
+                        startLockedCommand.clear();
+                }
+                std::println("startLockedCommand: {}", startLockedCommand);
                 continue;
             } else {
                 std::println(stderr, "[ ERROR ] Unknown option '{}' !", value);
@@ -264,8 +274,10 @@ int main(int argc, char** argv) {
     if (safeMode)
         g_pCompositor->m_safeMode = true;
 
-    if (lockedCrash)
-        g_pCompositor->m_lockedCrash = true;
+    if (startLocked) {
+        g_pCompositor->m_startLocked        = true;
+        g_pCompositor->m_startLockedCommand = startLockedCommand;
+    }
 
     if (!watchdogOk && !verifyConfig)
         Log::logger->log(Log::WARN, "WARNING: Hyprland is being launched without start-hyprland. This is highly advised against.");
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -38,7 +38,7 @@ static void help() {
     --verify-config              - Do not run Hyprland, only print if the config has any errors
     --version           -v       - Print this binary's version
     --version-json               - Print this binary's version as json
-    --locked-crash               - Hyprland crashed while locked)#");
+    --locked-crash               - Used by start-hyprland)#");
 }
 
 static void reapZombieChildrenAutomatically() {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -38,7 +38,7 @@ static void help() {
     --verify-config              - Do not run Hyprland, only print if the config has any errors
     --version           -v       - Print this binary's version
     --version-json               - Print this binary's version as json
-    --locked [COMMAND]           - Launches locker on startup via the provided command)#");
+    --locked-cmd [COMMAND]       - Launches locker on startup via the provided command)#");
 }
 
 static void reapZombieChildrenAutomatically() {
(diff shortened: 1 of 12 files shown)


src/managers/SessionLockManager.cpp
```diff
@@ -78,6 +78,9 @@ void CSessionLockManager::onNewSessionLock(SP<CSessionLock> pLock) {
     });
 
     m_sessionLock->listeners.unlock = pLock->m_events.unlockAndDestroy.listen([this] {
+        g_pCompositor->m_lockedCrash = false;
+        g_pCompositor->writeWatchdogFd("unlock");
```
vaxerski: woudlnt it be better if m_sessionLock controlled this? Would reduce dup and bugprone. it can do it in the ctor and dtor

vaxerski: or actually even better, expose events for lock and  unlock and have compositor listen

neal98: Thanks, I'll add events to the lock manager and listeners to the compositor

src/managers/SessionLockManager.cpp
```diff
@@ -227,6 +232,11 @@ void CSessionLockManager::forceUnlock() {
     g_pInputManager->refocus();
 }
 
+void CSessionLockManager::forceLock() {
```
vaxerski, line 239: what is this for?

src/managers/SessionLockManager.cpp
```diff
@@ -218,6 +222,7 @@ bool CSessionLockManager::clientDenied() {
 
 void CSessionLockManager::forceUnlock() {
     PROTO::sessionLock->forceUnlock();
+    m_events.unlock.emit();
     m_sessionLock = {};
```
vaxerski, line 225: put unlock and m_sessionLock = {}; into one fn something like ap rivate clearSessionlock()

Dregu:
There's also been a lot of requests to simply start hyprland in a locked state, e.g. to use hyprlock as a simple login screen. Of course this can be done by putting hyprlock in autostart, but that's not very secure and there's a lot of time to mess with binds before hyprlock opens.

I imagine this pr could also be used as a secure way to start locked, but it would display a lockdead screen before a session lock app has time to start and would also need `allow_session_lock_restore`. Could we make this more intentional functionality by adding another command line flag (or lua api) that forces a session lock from the start, **doesn't show a silly image** but simply assumes a session lock app will eventually pop up, and in the meantime system is unusable i.e. secure.

Edit: Your thing doesn't even seem to work with `allow_session_lock_restore`, it keeps drawing lockdead on top of hyprlock.

src/managers/SessionLockManager.cpp
neal98: Sure

src/managers/SessionLockManager.cpp
neal98: forceLock() is used to force the session into a locked state immediately during compositor init if Hyprland is started with the locked crash CLI flag. Following the instructions on the lockdead screen and evaluating 'hl.clear_crashed_lockscreen()' calls the existing forceUnlock() and takes the session back out of the locked state.

neal98:
> I imagine this pr could also be used as a secure way to start locked, but it would display a lockdead screen before a session lock app has time to start and would also need `allow_session_lock_restore`. Could we make this more intentional functionality by adding another command line flag (or lua api) that forces a session lock from the start, **doesn't show a silly image** but simply assumes a session lock app will eventually pop up, and in the meantime system is unusable i.e. secure.

This PR is kind of capable of doing that if you turn on `allow_session_lock_restore` and set up Hyprland to start a session locker immediately, but the flag is mainly intended for start-hyprland use. Maybe I could generalize this PR to a start locked feature, but that requires the user to specify their lock manager somewhere (maybe in the CLI with a flag). I wrote this with the goal of just fixing #15314. It's all up to @vaxerski .

> Edit: Your thing doesn't even seem to work with `allow_session_lock_restore`, it keeps drawing lockdead on top of hyprlock.

start-hyprland restarts Hyprland in safe mode wiping out any config changes, so I didn't think I'd need to heed 'allow_session_lock_restore'. Good catch though, it's good to respect the setting for correctness. I'll fix that.

Dregu:
> Maybe I could generalize this PR to a start locked feature

I did also test that the idea works great by removing all the lockdead rendering parts, just needs another cmd line switch, e.g. `Hyprland --locked` would simply run forceLock, not render anything. And obviously user would also need to make sure the actual lockscreen is autostarted somehow, but that doesn't need to have anything to do with this feature. Well I guess hyprland.start event could receive an argument to tell it was launched locked so config knows a lockscreen has to be started too.

vaxerski:
what dregu suggests is okay, but I'd expand it via `--locked 'cmd'` so hyprland can launch the command immediately when ready instead of you

Dregu:
I suppose that would be the clearest way to tell the user that the feature needs a lockscreen command somewhere and would naturally only run it when `--locked` is used, no logic needed in config. However the "immediately when ready" point is probably on `hyprland.start` anyway?

neal98:
Alright, to try to keep the user interface as clear as possible I swapped out the `--locked-crash` flag for a more general `--locked [COMMAND]` flag. The command is optional. Just `--locked` forces a lock and displays the lockdead screen, so it's perfect for use by start-hyprland. The command variant uses the inbuilt lua manager to eval the lockscreen command and start it securely. What do you guys think?

Dregu:
Functionally seems great to me, but probably didn't need to involve lua, I'd just use `Config::Supplementary::executor()->spawn(m_startLockedCommand);` which exec_cmd runs anyway.

neal98:
> Functionally seems great to me, but probably didn't need to involve lua

Agreed, that's been changed now.

src/main.cpp
```diff
@@ -176,6 +178,18 @@ int main(int argc, char** argv) {
                     help();
                     return 1;
                 }
+            } else if (value == "--locked") {
+                startLocked = true;
+
+                if (std::next(it) != args.end()) {
```
vaxerski: dicey, maybe do `--locked-cmd` for command, and `--locked` for lockdead?

vaxerski (commented):
good. Needs an entry in --help as well for `--locked-cmd` (only the user facing one)

src/main.cpp
neal98: Alright

neal98:
I've separated `--locked [COMMAND]` into `--locked` and `--locked-cmd COMMAND`. Only `--locked-cmd` usage is documented in help.

vaxerski (dismissed):
tnx!

vaxerski:
clang-format please

neal98:
Fixed.

vaxerski (approved):

PR #15341, opened 2026-07-07
Refactor generateShaderIncludes.sh for readability

JactusTheCactus:
All the `echo`s are now in a block that is written to `./src/render/shaders/Shaders.hpp`, instead of writing each `echo` to the file individually.

The refactoring shouldn't cause issues, but I am much more familiar with bash than I am with the posix shell.

The PR should be ready to merge.


vaxerski (approved):

vaxerski:
doesn't seem to be working. Is that posix sh syntax?

PR #15347, opened 2026-07-07
hyprctl/hyprpaper: support custom transition shader argument

markg85:
Companion to: https://github.com/hyprwm/hyprpaper/pull/371

**Note:** This is written against tag **v0.55.4**. ~~Consider this a draft!~~

### DIFF
--- a/hyprctl/hw-protocols/hyprpaper_core.xml
+++ b/hyprctl/hw-protocols/hyprpaper_core.xml
@@ -88,7 +88,7 @@
     <value idx="2" name="unknown_error" description="unknown error"/>
   </enum>
 
-  <object name="hyprpaper_wallpaper" version="1">
+  <object name="hyprpaper_wallpaper" version="2">
     <description summary="wallpaper object">
       This is an object describing a wallpaper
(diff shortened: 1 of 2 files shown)


markg85:
This: `changed the base branch from v0.55.4-b to v0.56.2-b` was damn difficult! Apparently github has an option to do that in the GUI. Also apparently, that's not available to me here in this PR... (i eventually got it working by using the github API and curl, ouch).

Anyhow, there is one rather fundamental (though tiny) change here. The `shader_path` had a `since` version in the xml. Because that seemed neat. However, that version property level isn't wired up in hyprwire so i dropped the property version completely. I can add it back but then i'd also have to patch hyprwire to add support for versions there. It's a TODO item in the hyprwire code and based on the comment in there i get the feeling it's not really important hence going for the easy route and just removing it here. (and in the hyprpaper patch too https://github.com/hyprwm/hyprpaper/pull/371).

Review this please.

markg85:
It's becoming hard to keep this working, please review! I want to merge this for the next release if possible. @vaxerski

PR #15351, opened 2026-07-08
core/output: don't destroy bound wl_output resources on same-name global replace

titosymanankasina:
# Describe your PR, what does it fix/add?

Fixes a mass client disconnect on fast monitor unplug/replug (e.g. an external HDMI screen losing power for a second, DPMS-adjacent output transitions).

When a monitor comes back with the same name while clients still hold `wl_output` resources of the old global, `CProtocolManager`'s `monitor.added` listener (and the same pattern in `onMonitorModeChange`) does `PROTO::outputs.erase(name)` before emplacing the new global. Erasing the map entry destructs the old `CWLOutputProtocol`, which destroys **server-side** every `wl_output` resource clients still hold. Each client then processes the `global_remove` it was just sent and answers with `wl_output.release` — but the id is already dead, so libwayland-server posts `invalid object N` on `wl_display` and fatally disconnects the client. On a real power blip this killed nearly every running app at once (Chrome, fcitx5, quickshell, xdg-desktop-portal-hyprland, …) with the compositor itself surviving.

Repro on v0.55.4 and current main, no hardware needed:

```
hyprctl output create headless test
# let any client bind it (any running app that binds wl_output), then:
hyprctl --batch "output remove test ; output create headless test"
# -> every client that bound "test" is disconnected with:
#    wl_display#1: error 0: invalid object NN
```

The fix retires the old global instead of destroying it:

- `retireOutputGlobal()` removes it from `PROTO::outputs` (and from the registry via `remove()` if not already defunct), but if clients still hold resources it is parked in `PROTO::defunctOutputs` and stays alive until they all release.
- `CWLOutputResource`'s destroy/release handlers route through `m_owner` instead of a `PROTO::outputs.at(name)` lookup, so releases reach the defunct global they actually belong to (the name now points at the new global).
- The self-erase in `destroyResource()` is identity-checked, so a draining defunct global can no longer erase the healthy same-name global that replaced it (latent variant of the same bug).

No new fields on any class (ABI-stable for plugins); one new inline vector `PROTO::defunctOutputs`, cleared in `~CProtocolManager`.

# Is it ready for merging, or does it need work?

Ready for review. Tested on v0.55.4 (patched via nixpkgs) and rebased on main: with the patch, the repro above leaves all clients connected; the defunct global drains once the last `release` arrives.

### DIFF
--- a/src/protocols/core/Output.hpp
+++ b/src/protocols/core/Output.hpp
@@ -67,9 +67,9 @@ class CWLOutputProtocol : public IWaylandProtocol {
 
 namespace PROTO {
     inline std::unordered_map<std::string, SP<CWLOutputProtocol>> outputs;
-    // defunct globals that were replaced by a new global with the same name but still have
-    // client-bound resources. Destroying those resources server-side would fatally error the
-    // clients ("invalid object" on their in-flight wl_output.release); they are kept here
-    // until every client has released, then drop out via destroyResource.
+    // defunct globals that are gone but not yet unref'd by all clients. Destroying their
+    // still-bound resources server-side would fatally error the clients ("invalid object"
+    // on their in-flight wl_output.release); they are kept here until every client has
+    // released, then drop out via destroyResource.
     inline std::vector<SP<CWLOutputProtocol>> defunctOutputs;
 };
--- a/src/protocols/core/Output.hpp
+++ b/src/protocols/core/Output.hpp
@@ -43,7 +43,8 @@ class CWLOutputProtocol : public IWaylandProtocol {
 
     // will mark the protocol for removal, will be removed when no. of bound outputs is 0 (or when overwritten by a new global)
     void remove();
-    bool isDefunct(); // true if above was called
+    bool isDefunct();        // true if above was called
+    bool hasBoundResources(); // true if any client still holds a wl_output resource of ours
 
     struct {
         CSignalT<SP<CWLOutputResource>> outputBound;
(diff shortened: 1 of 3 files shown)


src/protocols/core/Output.hpp
```diff
@@ -66,4 +67,9 @@ class CWLOutputProtocol : public IWaylandProtocol {
 
 namespace PROTO {
     inline std::unordered_map<std::string, SP<CWLOutputProtocol>> outputs;
+    // defunct globals that were replaced by a new global with the same name but still have
+    // client-bound resources. Destroying those resources server-side would fatally error the
+    // clients ("invalid object" on their in-flight wl_output.release); they are kept here
+    // until every client has released, then drop out via destroyResource.
```
vaxerski: I think this comment is wrong. They don't necessarily need to be replaced - they might simply be gone-but-not-yet-unref'd-by-all-clients

titosymanankasina: Good catch, fixed. Thanks!

vaxerski:
@codex review please

vaxerski:
clang-foirmat pls

vaxerski:
clang-format still fails

Dregu:
I've been running this for 3 days and nights, especially the nights, cause when I dpms off my monitors to go to sleep all the windows on one of them used to crash a lot and I think this finally fixed that?

vaxerski:
yes we just need a clang format

vaxerski:
thanks!

PR #15366, opened 2026-07-11
crop popup snapshots to popup size

gulafaran:
crop the snapshot to the size of the popup instead of full monitor snapshots. reduces a lot of overhead in fading out popups. 

math is not my strong side, so the transform calculations might be off, but tried mimic it from the elementrenderer where transform/scale is calculated. locally tested rotating and scaling and seems to work.

### DIFF
--- a/src/desktop/state/PopupFadeout.cpp
+++ b/src/desktop/state/PopupFadeout.cpp
@@ -44,6 +44,9 @@ SP<CPopupFadeout> CPopupFadeout::create(SP<CPopup> popup, SP<Render::IFramebuffe
     fadeout->m_monitor     = MONITOR;
     fadeout->m_framebuffer = snapshot;
 
+    const CBox EXTENTS   = popup->wlSurface()->resource()->extends();
+    fadeout->m_renderBox = CBox{popup->coordsGlobal() + EXTENTS.pos() - MONITOR->m_position, EXTENTS.size()}.scale(MONITOR->m_scale).round();
+
     static CConfigValue PBLURIGNOREA = CConfigValue<Config::FLOAT>("decoration:blur:popups_ignorealpha");
     if (shouldBlurPopup()) {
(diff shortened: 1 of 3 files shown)


src/desktop/view/Popup.cpp
```diff
@@ -267,15 +267,17 @@ void CPopup::onUnmap() {
...
-
-    const auto SNAPSHOT    = g_pHyprRenderer->makeSnapshotFB(m_self);
-    const auto SOURCEALPHA = m_alpha.get(POPUP_ALPHA_FADE)->value();
-
-    m_alpha.get(POPUP_ALPHA_FADE)->setConfig(Config::animationTree()->getAnimationPropertyConfig("fadePopupsOut"));
-    m_alpha.get(POPUP_ALPHA_FADE)->setValueAndWarp(1.F);
-    *m_alpha.get(POPUP_ALPHA_FADE) = 0.F;
-
-    Desktop::fadingOutState()->add(CPopupFadeout::create(m_self.lock(), SNAPSHOT, SOURCEALPHA));
+    if (m_alpha.get(POPUP_ALPHA_FADE)->enabled()) {
```
vaxerski: I think this belongs in ::create for hte fadeout? Also the same could be added to layers and windows.

gulafaran: you mean something like this? and move the makeSnapshotFB and animation var setting to ::create

```
    if (m_alpha.get(POPUP_ALPHA_FADE)->enabled())
            Desktop::fadingOutState()->add(CPopupFadeout::create(m_self.lock(), m_alpha));
```

yeah layers and windows can be done, but one thing at a time :D, popups is especially nasty. have 8+ tabs in firefox and drag the mouse over the tabs to get "previews"  and poof its 8 full screen snapshots in a short time.

gulafaran:
to much nested things il just close this for now, and put it on the low prio of todo list.

PR #15371, opened 2026-07-11
renderer/cm: fix blur clipping with HDR

ryanpholt:

Problem: The blurprepare intermediate used an 80-nit range but HDR pixels can be much brighter. A 200-nit pixel was encoded as:

    (200/80)^(1/2.2) = 1.52

which would get clamped to 1 causing clipping.

Solution: Use the workbuffer's configured max luminance (1000, for example) for the range so that a 200-nit pixel becomes:

    (200/1000)^(1/2.2) = 0.48

We also must consistently use the EXT_LINEAR transfer function range in blurprepare/blurFinish.

Note: KDE somewhat similarly uses gamma 2.2 encoding where 1.0 == max luminance of monitor:
https://zamundaaa.github.io/wayland/2024/11/02/hdr-and-color-management-in-kwin-part-4.html

Before Picture (note bad clipping of the highlights):

before

After Picture (no clipping):

after

No issues that I am aware of.

Ready

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -1684,6 +1684,16 @@ void CHyprOpenGLImpl::renderTextureMatte(SP<ITexture> tex, const CBox& box, SP<I
     tex->unbind();
 }
 
+static SCMSettings blurIntermediateCMSettings(bool toIntermediate) {
+    const auto WORKBUFFER   = g_pHyprRenderer->workBufferImageDescription();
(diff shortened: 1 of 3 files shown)


umbrageodotus:
@vaxerski good for 0.56 I assume

vaxerski:
I don't hate this, @UjinT34 any comments?

vaxerski:
@codex review please

vaxerski (approved):

PR #15374, opened 2026-07-11
config/actions: unfocus last window on changeWorkspace when moving monitors

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/15076


fxzzi:
werks4me much love

PR #15398, opened 2026-07-13
layout/scroll: fix `focus_on_active` not scrolling to the focused window after switching to its workspace

erstarr:

It switches to the workspace but doesn't scroll to the window in question.

This MR fixes that

The fix causes double fullWindowFocus() calls but changing the reason of the call inside changeworkspace would regress https://github.com/hyprwm/Hyprland/pull/13974 (viewport move when you're focused on a column when changing to another workspace)

No

WIP
- [ ] Test


Dregu:
Works, but seems dumb we have to fix a scrolling issue in non-scrolling code, why is https://github.com/hyprwm/Hyprland/blob/main/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp#L631 unable to do this?

erstarr:
> Works, but seems dumb we have to fix a scrolling issue in non-scrolling code, why is https://github.com/hyprwm/Hyprland/blob/main/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp#L631 unable to do this?

Would need to register the listener with the scroll algo object. Possible, but i'm unsure about always saving a listener for this one event.

PR #15401, opened 2026-07-14
bindings/dispatchers: Don't inherit key state modmask during key sym lookup

masnax:

Discussion: https://github.com/hyprwm/Hyprland/discussions/14772

In some cases, such as with bind -> sendshortcut, the key sym lookup will fail resulting in a 'key not found' error when the cached xkb state is used for lookup.

This can happen if, for example, the SHIFT key is bound in `hl.bind`. Consider the following case, where SHIFT+1 is mapped to 2:

```lua
hl.bind("SHIFT + 1", hl.dsp.send_shortcut({mods = "", key = "2"}))
````

`resolveKeycode` will fail to lookup 2 because SHIFT is depressed in the modmask.

Here, I've modified the lookup to use a fresh key state with no modmask.

After lookup, `ConfigActions::pass` will clear the modmask before sending keys anyway, so an empty modmask can be used for lookup in `resolveKeycode` safely.

Good to go :)

### DIFF
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -125,16 +125,7 @@ class CTestKeyboard : public IKeyboard {
     }
 
     void setMods(uint32_t depressed, uint32_t latched, uint32_t locked, uint32_t group) {
-        m_modifiersState.depressed = depressed;
-        m_modifiersState.latched   = latched;
-        m_modifiersState.locked    = locked;
-        m_modifiersState.group     = group;
-        m_keyboardEvents.modifiers.emit(IKeyboard::SModifiersEvent{
-            .depressed = depressed,
-            .latched   = latched,
-            .locked    = locked,
-            .group     = group,
-        });
+        updateModifiers(depressed, latched, locked, group);
     }
 
     void destroy() {
--- a/hyprtester/plugin/src/main.cpp
+++ b/hyprtester/plugin/src/main.cpp
@@ -422,6 +422,28 @@ static SDispatchResult keybind2(std::string in) {
     return {};
 }
 
+static SDispatchResult keybindModmask(std::string in) {
+    CVarList2 data(std::move(in));
+    // 0 = release, 1 = press
+    bool press;
+    // See src/devices/IKeyboard.hpp : eKeyboardModifiers for modifier bitmasks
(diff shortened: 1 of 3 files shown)


vaxerski (dismissed):
just add tests for this please

masnax:
> just add tests for this please

Done. The existing `keybind` plugin command didn't actually set the underlying xkb state and so wasn't properly capturing the logic in `resolveKeycode`. I added a new `hyprtester` plugin command `keybind_modmask` that properly applies modmasks to the keyboard. I didn't change the behaviour of the existing `keybind` and `keybind2` which just set `m_lastMods` without changing the key state, because I couldn't get all the keybind tests to consistently pass (irrespective of the changes here). 

The `set_mods` command will now also update the xkb state with the modmask, as it uses the same underlying implementation. I only see that used [here](https://github.com/hyprwm/Hyprland/blob/main/hyprtester/src/tests/clients/keyboard-modifiers.cpp#L102), and that test also seems to occasionally fail for me both before & after the changes introduced here. Not sure if the tests are flaky in general or if it's a quirk of my setup.

Lastly, I noticed the same `send_shortcut` for key `q` was used repeatedly, so it was getting cached and skipping the lookup after the first call [(occurs here)](https://github.com/hyprwm/Hyprland/blob/main/src/config/lua/bindings/LuaBindingsDispatchers.cpp#L375). So I've updated the relevant calls to `send_shortcut` to send unique keys.

vaxerski:
tnx. clang-format pls?

masnax:
> tnx. clang-format pls?

Oops, done. Sorry for the runaround.

vaxerski (approved):

vaxerski:
test regressions?

masnax:
> test regressions?

The `shortcutInhibitor` is failing, but looking at the code, it doesn't seem related to the changes in this PR. I haven't modified any of the logic for `hl.bind`, or `hl.plugin.test.keybind`. 

I've tried following the steps in the `nix-test.yaml` workflow in a NixOS VM, I can get the `shortcutInhibitor` test (as well as a few other keybind tests, particularly longpress) to occasionally fail, both with/without the changes here. Might just be timeout related flakiness.

vaxerski:
I'll re-run it again but this is the 4th time

vaxerski:
well I guess 4th time's the charnb

PR #15408, opened 2026-07-14
ipc: handle partial client transfers

ghost:
## What

- retry interrupted and partial writes from `hyprctl` and `hyprpm`
- read `hyprpm` replies until EOF instead of treating a short stream read as completion
- construct IPC replies from the exact byte count returned by `read`

## Why

Unix stream sockets may complete a write or read with fewer bytes than requested. The previous clients could send a truncated command, truncate a reply after a short read, or read past the initialized response bytes when an 8192-byte reply filled the buffer.

## Impact

IPC clients behave reliably under short transfers and `EINTR`, with no protocol changes.

## Checks

- `cmake --build build --target hyprctl hyprpm -j2`
- focused ReturnGuard scan of both translation units (no remaining short-transfer diagnostics)

### DIFF
--- a/hyprctl/src/main.cpp
+++ b/hyprctl/src/main.cpp
@@ -1,6 +1,7 @@
 #include <re2/re2.h>
 
 #include <cctype>
+#include <cerrno>
 #include <netdb.h>
 #include <netinet/in.h>
 #include <cstdio>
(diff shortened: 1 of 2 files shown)


vaxerski (approved):

vaxerski:
@codex review please

PR #15411, opened 2026-07-14
init: gain real-time through rtkit

andresilva:
Hyprland tries to put its main thread on `SCHED_RR` at startup, but the direct `pthread_setschedparam` call only succeeds when the process is already privileged (`CAP_SYS_NICE` or a nonzero `RLIMIT_RTPRIO`), so on a stock install it silently fails and the compositor runs without realtime. This adds the standard desktop path as a fallback: asking rtkit (`org.freedesktop.RealtimeKit1`) over the system bus.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -361,9 +361,12 @@ if(NOT HAS_INOTIFY AND inotify_FOUND)
 endif()
 
 # sdbus-c++ is used to ask rtkit for realtime scheduling which is Linux-only
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if(NO_RTKIT OR NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  message(STATUS "rtkit realtime scheduling is disabled!")
+else()
   pkg_check_modules(sdbus_dep REQUIRED IMPORTED_TARGET sdbus-c++>=2.0.0)
   target_link_libraries(hyprland_lib PUBLIC PkgConfig::sdbus_dep)
+  add_compile_definitions(HAS_RTKIT)
 endif()
 
 if(NO_XWAYLAND)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,6 +360,12 @@ if(NOT HAS_INOTIFY AND inotify_FOUND)
   target_link_libraries(hyprland_lib PUBLIC PkgConfig::inotify)
 endif()
 
+# sdbus-c++ is used to ask rtkit for realtime scheduling which is Linux-only
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  pkg_check_modules(sdbus_dep REQUIRED IMPORTED_TARGET sdbus-c++>=2.0.0)
+  target_link_libraries(hyprland_lib PUBLIC PkgConfig::sdbus_dep)
+endif()
+
 if(NO_XWAYLAND)
   message(STATUS "Using the NO_XWAYLAND flag, disabling XWayland!")
   add_compile_definitions(NO_XWAYLAND)
--- a/src/init/initHelpers.cpp
+++ b/src/init/initHelpers.cpp
@@ -8,10 +8,11 @@
 #include <sys/syscall.h>
 #include <algorithm>
 #include <cerrno>
+#include <chrono>
 #include <cstring>
 #include <optional>
 
-#include <gio/gio.h>
+#include <sdbus-c++/sdbus-c++.h>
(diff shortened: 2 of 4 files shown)


andresilva:
Tested locally and it worked, but would be nice if someone else could test.

src/init/initHelpers.cpp
```diff
@@ -3,34 +3,359 @@
...
 
+#if defined(__linux__)
+#include <sys/resource.h>
+#include <sys/syscall.h>
+#include <algorithm>
+#include <cerrno>
+#include <cstring>
+#include <optional>
+
+#include <gio/gio.h>
```
vaxerski: including any form of GLib is banned in hyprland

andresilva: I wanted to avoid introducing a new dependency. Is adding sdbus okay with you? It's what's used in xdph iirc

andresilva: Updated to use sdbus.

vaxerski: sdbus is ok

CMakeLists.txt
```diff
@@ -360,6 +360,12 @@ if(NOT HAS_INOTIFY AND inotify_FOUND)
   target_link_libraries(hyprland_lib PUBLIC PkgConfig::inotify)
 endif()
 
+# sdbus-c++ is used to ask rtkit for realtime scheduling which is Linux-only
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
vaxerski: define some variable here so you can #ifdef the handling in the source.

andresilva: Updated.

vaxerski (approved):

vaxerski:
reverting cuz breaks @Dregu

Dregu:
I think it launched fine embedded, but from tty I could not run Hyprland at all, it just quits after loading config.

andresilva:
Is there an issue report with any details? Do you use uwsm? Or should I just launch Hyprland from a bare tty to test?

Dregu:
I just run Hyprland from tty1. I have no more details other than the Hyprland process prints a few normal things to stdout, gets killed before reaching desktop and exits with 137. In safe-mode, debug build, doesn't matter. Reverting this pr makes it work.

andresilva:
@Dregu Did the process write "Realtime budget exceeded" to stderr? Also can you test if the issue happens while rtkit is disabled?

```bash
prlimit --pid $$ --rtprio --rttime

busctl --system get-property \
    org.freedesktop.RealtimeKit1 \
    /org/freedesktop/RealtimeKit1 \
    org.freedesktop.RealtimeKit1 \
    RTTimeUSecMax

Hyprland >/tmp/hyprland-rt.log 2>&1
hypr_status=$?

echo "exit status: $hypr_status"

grep -inF "Realtime budget exceeded" /tmp/hyprland-rt.log ||
    echo "No realtime-budget message found"

tail -n 50 /tmp/hyprland-rt.log

journalctl -k -b --since "5 minutes ago" --no-pager |
    grep -Ei 'hyprland|oom|killed process|watchdog|rttime'

journalctl -b -u rtkit-daemon.service --since "5 minutes ago" --no-pager |
    tail -n 50
```

If you could run this and paste the output here would help with debugging (I can't replicate this myself).

Dregu:
So, I tested a bunch of random stuff and here's some results

- I can make realtime work by actually adding myself to the realtime group, no issue there, never really been an aspiration to run things realtime, shit worked fine
- OR I can make the pr run again (without rtprio) by masking rtkit-daemon
- Also works with `HYPRLAND_NO_RT=1 Hyprland` of course

Here's your log from when it definitely didn't work:

```
RESOURCE DESCRIPTION                      SOFT      HARD UNITS
RTPRIO   max real-time priority              0         0
RTTIME   timeout for real-time tasks unlimited unlimited microsecs
x 200000
exit status: 137
No realtime-budget message found
(unrelated libinput crap here)
Jul 17 22:48:52 onix rtkit-daemon[75921]: Successfully made thread 77072 of process 77072 owned by '1000' RT at priority 1.
```

fxzzi:
sooooo i need rtkit now which i didn't need before to maintain SCHED_RR?

andresilva:
@fxzzi No you don't need rtkit, if you look at the diff you can see that the existing direct path is still there. So if SCHED_RR was working for you it will keep working.

fxzzi:
stopped working for me weirdly enough. enabling rtkit solved it again

yasink21:
> stopped working for me weirdly enough. enabling rtkit solved it again

same for me, without rtkit commit I can't use rt

PR #15421, opened 2026-07-15
desktop/view: fix IView virtual inheritance and erase_if ptr

gulafaran:
CSubsurface and CSessionLock inherited IView non-virtually while CWindow/CLayerSurface/CPopup inherited it virtually.

erase_if compared event.address with the derived complete object pointer. under virtual inheritance those addresses differ.

### DIFF
--- a/src/desktop/state/LayerState.cpp
+++ b/src/desktop/state/LayerState.cpp
@@ -19,7 +19,7 @@ CLayerState::CLayerState() {
         if (event.type != View::VIEW_TYPE_LAYER_SURFACE)
             return;
 
-        std::erase_if(m_layers, [&](auto& x) { return !x || rc<uintptr_t>(sc<View::IView*>(x.get())) == event.address; });
+        std::erase_if(m_layers, [&](auto& x) { return !x || rc<uintptr_t>(dc<View::IView*>(x.get())) == event.address; });
     });
 }
 
--- a/src/desktop/state/LayerState.cpp
+++ b/src/desktop/state/LayerState.cpp
@@ -19,7 +19,7 @@ CLayerState::CLayerState() {
         if (event.type != View::VIEW_TYPE_LAYER_SURFACE)
             return;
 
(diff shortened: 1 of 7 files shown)


src/desktop/state/LayerState.cpp
```diff
@@ -19,7 +19,7 @@ CLayerState::CLayerState() {
         if (event.type != View::VIEW_TYPE_LAYER_SURFACE)
             return;
 
-        std::erase_if(m_layers, [&](auto& x) { return !x || rc<uintptr_t>(x.get()) == event.address; });
+        std::erase_if(m_layers, [&](auto& x) { return !x || rc<uintptr_t>(sc<View::IView*>(x.get())) == event.address; });
```
vaxerski: you cant use sc<> with multi / virtual inheritance, gotta do dc<>

gulafaran: changed

vaxerski:
this should be better. Can you check if it still works? I can't rpeor the crahs

gulafaran:
> this should be better. Can you check if it still works? I can't rpeor the crahs

then it probably needs to be `event.view == dynamicPointerCast<View::IView>(x->m_self);` just dynamic casting is always true because well the CWindow etc is always an IView

src/desktop/state/WindowState.cpp
```diff
@@ -21,7 +21,7 @@ CWindowState::CWindowState() {
         if (event.type != View::VIEW_TYPE_WINDOW)
             return;
 
-        std::erase_if(m_windows, [&](auto& x) { return !x || rc<uintptr_t>(dc<View::IView*>(x.get())) == event.address; });
+        std::erase_if(m_windows, [&](auto& x) { return !x || dynamicPointerCast<View::IView>(x->m_self); });
```
gulafaran: opsie :D

vaxerski (approved):
lol my bad

gulafaran:
> this should be better. Can you check if it still works? I can't rpeor the crahs

seems fine for me

PR #15428, opened 2026-07-15
renderer: block solitary on fadeouts

gulafaran:
after 254d6bae775258b554e1d3a4a853f3bc244190d9 fadeouts now live in fadingOutState() instead of the normal containers, so add SC_FADEOUT and iterate the new fadingOutState() and block solitary for fadeouts.

per request from @fxzzi


fxzzi:
can confirm works4me, thanks!!!

vaxerski (approved):

PR #15429, opened 2026-07-15
Screenshare fence and copy fixes

gulafaran:
erase done frames from m_pendingFrames to avoid growing to huge
numbers. a finished frame copied/failed/session gone is skipped by the
copy but was only erased once it expired. a slow client could let the
vector grow.

don recopy a dmabuf while fence is pending
copy is called for every pending frame on the output commit. and output
commit fires on any output commit not just buffer ones. so the fence can
easily be not signaled yet.

### DIFF
--- a/src/managers/screenshare/ScreenshareFrame.cpp
+++ b/src/managers/screenshare/ScreenshareFrame.cpp
@@ -143,7 +143,7 @@ eScreenshareError CScreenshareFrame::share(SP<IHLBuffer> buffer, const CRegion& 
 }
 
 void CScreenshareFrame::copy() {
-    if (done())
+    if (done() || m_copyInFlight)
         return;
 
     // tell client to send presented timestamp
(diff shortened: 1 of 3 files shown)


src/managers/screenshare/ScreenshareFrame.cpp
```diff
@@ -114,18 +114,15 @@ eScreenshareError CScreenshareFrame::share(SP<IHLBuffer> buffer, const CRegion&
     m_callback = callback;
     m_shared   = true;
 
-    if (m_session->m_type == SHARE_MONITOR || m_session->m_type == SHARE_REGION) {
-        const auto PMONITOR = m_session->monitor();
-        if (PMONITOR)
+    const auto PMONITOR = m_session->monitor();
+
+    if (PMONITOR) {
+        if (m_isFirst)
```
vaxerski: what if a client does a copy, waits a bit, then does another one? It will get partial data.

gulafaran: uh where exactly? m_isFirst only forces a full monitor rerender so the mirror texture is fully repainted before copying. the others use the pendingMirrorFBDamage so the only changed region is repainted

vaxerski: copy screenshot

sleep(2s)

copy screenshot

will that re-trigger isFirst?

gulafaran: yeah fair point, by the looks of it when you mention it currently it works by luck because above TODO so both paths are pretty much same right now. full monitor damages.

gulafaran: il just not touch this, but worth remembering this will most likely break as you say if damage tracking is introduced.

vaxerski (approved):

PR #15432, opened 2026-07-16
Change default new_status to "slave" for master layout

sinder38:

I noticed that [wiki](https://wiki.hypr.land/Configuring/Layouts/Master-Layout/) lists `new_status` default as `slave`, but the actual default is `master`!
this also is evident in the video demonstration.

Nah 

yep

### DIFF
--- a/example/hyprland.lua
+++ b/example/hyprland.lua
@@ -42,7 +42,7 @@ local menu        = "hyprlauncher"
 -- Autostart necessary processes (like notifications daemons, status bars, etc.)
 -- Or execute your favorite apps at launch like this:
 --
--- hl.on("hyprland.start", function () 
+-- hl.on("hyprland.start", function ()
 --   hl.exec_cmd(terminal)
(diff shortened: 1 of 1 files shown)


vaxerski:
no, the default is slave. The example config overrides it to set to master

sinder38:
@vaxerski, got it  thanks for clarifying, that makes sense

The issue I actually ran into: 
a fresh install uses the example config, which overrides `new_status` to `master`. 
So new users looking at the wiki's video showcase get different behavior than what they see out of the box.

Mismatch between wiki demonstration and example-config behavior is what tripped me up. Would a short comment in the example config (or a maybe wiki note) pointing out the override be worth adding?

vaxerski:
I don't see the point? Wiki just demonstrates some random settings, why would you expect that's the default? It's just to demonstrate what master is

sinder38:
The example config functions as "the default" to new users only.
New users trying out Hyprland may evaluate it lower because master layout behaves weird, without looking into why

PR #15433, opened 2026-07-16
fix tokens for xdg-desktop-portals

gulafaran:
tokens are only removed once a window maps, so things like xdg-desktop-portal got the 1337 month duration token added and always first launch in the wrong workspace. check PINITIALWSTRACKING and if mode 1 just give them a 1 second token. like other code paths does.

### DIFF
--- a/src/config/supplementary/executor/Executor.cpp
+++ b/src/config/supplementary/executor/Executor.cpp
@@ -149,7 +149,8 @@ std::optional<uint64_t> CExecutor::spawnWithRules(std::string args, PHLWORKSPACE
 }
 
 std::vector<std::pair<std::string, std::string>> CExecutor::getHyprlandLaunchEnv(PHLWORKSPACE pInitialWorkspace) {
-    static auto PINITIALWSTRACKING = CConfigValue<Config::INTEGER>("misc:initial_workspace_tracking");
+    static auto PINITIALWSTRACKING        = CConfigValue<Config::INTEGER>("misc:initial_workspace_tracking");
+    static auto PINITIALWSTRACKINGTIMEOUT = CConfigValue<Config::INTEGER>("misc:initial_workspace_token_timeout");
 
     if (!*PINITIALWSTRACKING || m_isLaunchingExecOnce)
         return {};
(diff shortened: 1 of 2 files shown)


vaxerski:
question is does 1 second make sense? I think the entire point here is to launch some (slow) app, switch away, and have it launch where we were. Maybe like 10 seconds?

gulafaran:
> question is does 1 second make sense? I think the entire point here is to launch some (slow) app, switch away, and have it launch where we were. Maybe like 10 seconds?

fair point, 10 seconds seems reasonable for silly things like `krita` that takes eons, but is an eternity for like a terminal. but i guess it can always be tweaked in the future if required. il change it

wabarTM:
maybe expose it as a variable under misc or quirks?

gulafaran:
> maybe expose it as a variable under misc or quirks?

yeah true, done.

gulafaran:
wiki PR https://github.com/hyprwm/hyprland-wiki/pull/1596

src/config/supplementary/executor/Executor.cpp
```diff
@@ -167,8 +168,8 @@ std::vector<std::pair<std::string, std::string>> CExecutor::getHyprlandLaunchEnv
             pInitialWorkspace = PMONITOR->m_activeWorkspace;
     }
 
-    result.emplace_back("HL_INITIAL_WORKSPACE_TOKEN",
-                        g_pTokenManager->registerNewToken(Desktop::View::SInitialWorkspaceToken{{}, pInitialWorkspace->getConfigName()}, std::chrono::months(1337)));
+    const auto TOKEN = (*PINITIALWSTRACKING == 2) ? std::chrono::seconds(std::chrono::months(1337)) : std::chrono::seconds(*PINITIALWSTRACKINGTIMEOUT);
```
vaxerski: why is a duration called TOKEN 💀

gulafaran: porażenie słoneczne

gulafaran: [992c1e2](https://github.com/hyprwm/Hyprland/pull/15433/commits/992c1e20b21e9436bd4ffb44feec328ad7987d38)

vaxerski (approved):
wiki mr please

PR #15461, opened 2026-07-17
config/lua: allow wildcards or absolute paths in require

vaxerski:
Allows absolute paths and wildcards in require()

needs wiki


BeyondMagic:
Nice, before this we needed to:
```lua
local function require_lua_tree(root)
    local handle = io.popen('/usr/bin/find -L ' .. root .. ' -maxdepth 2 -type f -name "*.lua"')
    if not handle then
        return
    end

    for path in handle:lines() do
        dofile(path)
    end

    handle:close()
end

require_lua_tree("~/.config/hypr/")
```

PR #15465, opened 2026-07-17
init: gain real-time through rtkit (again)

andresilva:
Second take of #15411.

### DIFF
--- a/src/init/initHelpers.cpp
+++ b/src/init/initHelpers.cpp
@@ -19,6 +19,12 @@ bool NInit::isSudo() {
     return getuid() != geteuid() || !geteuid();
 }
 
+// NixOS-specific fix to prevent all children from inheriting
+// CAP_SYS_NICE due to how the security wrapper works.
+void NInit::lowerAmbientCaps() {
+    prctl(PR_CAP_AMBIENT, PR_CAP_AMBIENT_LOWER, CAP_SYS_NICE, 0, 0);
+}
+
 // Asks the kernel directly to put the calling thread on SCHED_RR. Only succeeds if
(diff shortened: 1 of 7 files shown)


andresilva:
@Dregu can you test?

Dregu:
Hey now it doesn't crash and burn when rtkit is on and realtime group is off. I still don't actually see any sign of rtprio after it launches succesfully and claims "Gained realtime scheduling via rtkit"... But I don't know or care about that anyway, at least it runs without special attention.

andresilva:
You can check with `chrt -p <hyprland pid>`. Also if you're running something like ananicy it might bring it back to SCHED_OTHER depending on the rules you have.

src/init/initHelpers.cpp
```diff
@@ -19,6 +19,12 @@ bool NInit::isSudo() {
     return getuid() != geteuid() || !geteuid();
 }
 
+// NixOS-specific fix to prevent all children from inheriting
+// CAP_SYS_NICE due to how the security wrapper works.
```
JohnRTitor, line 23: We will remove this from NixOS side.

andresilva: I suggest we still keep it for now.

vaxerski (approved):

tekstryder:
```
pid 699920's current scheduling policy: SCHED_RR|SCHED_RESET_ON_FORK
pid 699920's current scheduling priority: 1
```

Very cool. Thx.

fxzzi:
yep thanks works for me as well after removing the ananicy-cachyos rule

PR #15471, opened 2026-07-18
windowstate/layerstate: don't dynamic_cast a view's self-ptr during destruction

gulafaran:
The view.destroy listeners erase_if'd their containers by evaluating dynamicPointerCast<IView>(x->m_self). Since the event is emitted from ~IView, the matching view's derived member m_self is already destroyed by then, so reading it is UB. Compare control-block identity via PHLVIEWREF{x} instead, which never dereferences the dying object.

see: https://github.com/hyprwm/hyprutils/pull/118

im a bit out of my brainpower here so this is all robot code.

### DIFF
--- a/src/desktop/state/LayerState.cpp
+++ b/src/desktop/state/LayerState.cpp
@@ -19,7 +19,10 @@ CLayerState::CLayerState() {
         if (event.type != View::VIEW_TYPE_LAYER_SURFACE)
             return;
 
-        std::erase_if(m_layers, [&](auto& x) { return !x || event.view == dynamicPointerCast<View::IView>(x->m_self); });
+        // A CLayerSurface is always an IView via a static upcast, so compare control-block identity directly.
+        // Must NOT dereference x->m_self here: this event is emitted from ~IView, by which point CLayerSurface's
+        // members (incl. m_self) are already destroyed. PHLVIEWREF{x} only reads x's own impl_/data pointer.
+        std::erase_if(m_layers, [&](auto& x) { return !x || event.view == PHLVIEWREF{x}; });
     });
 }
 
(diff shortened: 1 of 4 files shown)


src/desktop/state/LayerState.cpp
```diff
@@ -19,7 +19,10 @@ CLayerState::CLayerState() {
         if (event.type != View::VIEW_TYPE_LAYER_SURFACE)
             return;
 
-        std::erase_if(m_layers, [&](auto& x) { return !x || event.view == dynamicPointerCast<View::IView>(x->m_self); });
+        // A CLayerSurface is always an IView via a static upcast, so compare control-block identity directly.
+        // Must NOT dereference x->m_self here: this event is emitted from ~IView, by which point CLayerSurface's
+        // members (incl. m_self) are already destroyed. PHLVIEWREF{x} only reads x's own impl_/data pointer.
+        std::erase_if(m_layers, [&](auto& x) { return !x || event.view == PHLVIEWREF{x}; });
```
vaxerski: well this is wrong, it's virtually inherited. I think a simpler solution would just be to do `return x.expired();`?

gulafaran: well x is a SP so it has no expired and x->m_self would just hit the issue aswell i think

gulafaran: if i understand it right its because derived m_self is destructed before base x. which meant this `event.view == dynamicPointerCast<View::IView>(x->m_self);` tried to upcast that m_self that is expired. and what its trying to do here is creating a temp WP from X (that isnt destroyed yet) and use the == operator in hyprutils that compares the .impl to eachother.

vaxerski: and what's the problem with checking `x.expired()`? Expired is true when we're already in the dtor.

gulafaran: a SP doesnt have .expired()

vaxerski: ahh I see. Well then that entire listener is pointless.

vaxerski (approved):

PR #15474, opened 2026-07-18
a few protocol error touchups

gulafaran:
reduce bogus -1 errors, and try to honor spec as much as we can. 
there is more to go over in `protocols/*` and general logical flaws. but this should get most hard since version violations right.


vaxerski (approved):
nice

PR #15482, opened 2026-07-19
config/lua: support string[] in hl.bind

Shiphan:

Add support for `string[]` for the keys parameter in hl.bind

I think use an actual array will be a better api than a string that acts like an array

and string[] is also much better if you want to use any variable in keys:
```lua
hl.bind(mainMod .. " + 1", hl.dsp.focus({ workspace = 1, on_current_monitor = true }))
-- to
hl.bind({ mainMod, "1" }, hl.dsp.focus({ workspace = 1, on_current_monitor = true }))

for i = 1, 10 do
	local key = tostring(i % 10)

	hl.bind(mainMod .. " + " .. alternative .. " + " .. key, hl.dsp.window.move({ workspace = i }))
	-- to
	hl.bind({ mainMod, alternative, key }, hl.dsp.window.move({ workspace = i }))
end
```

Alternative: use this lua function if user want to use string[]
```lua
---@param keys string | string[]
---@param dispatcher HL.Dispatcher | function
---@param opts? HL.BindOptions
local function bind(keys, dispatcher, opts)
	if type(keys) == "table" then
		keys = table.concat(keys, " + ")
	end
	hl.bind(keys, dispatcher, opts)
end
```

### DIFF
--- a/meta/generateLuaStubs.py
+++ b/meta/generateLuaStubs.py
@@ -520,7 +520,7 @@ def generate_stub(root: Path) -> str:
 
     api_signatures: dict[str, str] = {
         "hl.on": "fun(event: HL.EventName, cb: fun(...)): HL.EventSubscription",
-        "hl.bind": "fun(keys: string, dispatcher: HL.Dispatcher|function, opts?: HL.BindOptions): HL.Keybind",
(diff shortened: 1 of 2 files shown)


vaxerski:
I don't think there's much point. The thing about lua is that you can write that convenience function if you wish

PR #15485, opened 2026-07-19
output: add support for setting color range

gulafaran:
add support for setting color range to outputs from a monitor rule. can be "auto, full, limited". and output it in hyprctl monitors.

requires: https://github.com/hyprwm/aquamarine/pull/335
wiki mr: https://github.com/hyprwm/hyprland-wiki/pull/1598


gulafaran:
only thing i dont know if this requires some error handling src/output/Monitor.cpp line 1053

fxzzi:
ah yeah tested this one too, but doesn't have any affect on my system. i am assuming both of my monitors are already using full colour range by default in some way

PR #15486, opened 2026-07-19
CI: Fetch PR branch before running codeowners

LRitzdorf:

Fixes codeowners-plus not finding the tip of the PR it's supposed to analyze. ~~This also matches their own docs.~~

**Edit:** Switching back to `pull_request` doesn't always work, because GH refuses to let PRs from *forks* write to PRs (even their own). We should be able to just manually fetch the PR branch instead.

~~@vaxerski, it looks like you changed from `pull_request` to `pull_request_target` in #15383, but it's not clear why. Hopefully I'm not stepping on any toes by reverting that.~~

~~For verification that the above change was what broke it, notice that the [last successful run of this workflow](https://github.com/hyprwm/Hyprland/actions/runs/29197931363) took place on July 12 at 9:18AM, just before #15383 was merged, and every run since has failed with the same error.~~

~~Also, see LRitzdorf#1 for proof that this fix works.~~

Tested and ready to merge.

### DIFF
--- a/.github/workflows/codeownersplus.yml
+++ b/.github/workflows/codeownersplus.yml
@@ -24,6 +24,9 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: 'Fetch PR Branch'
+        run: git fetch origin ${{ github.event.pull_request.head.sha }}
+
       - name: 'Codeowners Plus'
         uses: multimediallc/codeowners-plus@v1.9.1
         with:


LRitzdorf:
Uhhhh hold on, we don't have permission to request reviews now. Welp, presumably that's what vax was trying to fix with the original change...

LRitzdorf:
Okay, so [apparently](https://github.blog/news-insights/product-news/github-actions-improvements-for-fork-and-pull-request-workflows/) GitHub doesn't hand a token with `pull-requests: write` to PRs from forks, even if the workflow file requests that. There's a repo (and also org) setting to enable that, but I think the easier solution is to keep using `pull_request_target`, and just manually fetch the PR branch before running codeowners. I was able to [verify on my fork](https://github.com/LRitzdorf/Hyprland/actions/runs/29696408298/job/88217818423?pr=1#step:5:13) that this at least gets us far enough to successfully diff against the PR branch, which should™ be all we need at this point

LRitzdorf:
Fetching by SHA is safer than by "ref" (which for all I know, could just be the branch name on the fork, and we can't fetch that from the original repo)

vaxerski:
still fails or is that the main ci running

LRitzdorf:
That's main CI, you can see in the workflow it doesn't have the "Fetch PR Branch" step that fixes this

vaxerski (approved):
tnx

PR #15492, opened 2026-07-20
Allow constraining floating windows to be entirely or partially on-screen

LRitzdorf:

Two new config keys: `misc:float_force_onscreen`, and `misc:new_float_force_onscreen`. These control whether existing and new (respectively) floating windows are allowed to be positioned outside of the visible screen area, and if so, how.

Each config key accepts the values 0, 1, or 2. 0 means no constraints are applied to placement; 1 means that the window may be partially (but not fully) offscreen, and 2 means that the window must remain fully onscreen.

These constraints are applied as part of the "move window" logic, which should be fine. The only slightly weird interaction I can think of is that new windows can have a more permissive setting than existing ones, which means a window might spawn offscreen but be impossible to put offscreen again once moved. I think that's fine though, because if the user configures this it should be pretty self-explanatory. Also, this would just... make sense if, say, they want some specific window to spawn offscreen and never plan to interact with it, but want all other floating windows to stay onscreen.

To match the existing behavior, `float_force_onscreen` defaults to 0, and `new_float_force_onscreen` defaults to 2.

None that I'm aware of. ~~Waiting to mark ready for review until I get the hyprutils version bump and can test this locally without crashing, but I roped in Wabar to test and he reported that the settings generally work.~~

~~Waiting until I can personally test this on main, but should be good. See above.~~

Tested now, and this behaves in a manner I'm happy with!

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -1943,3 +1943,71 @@ TEST_CASE(sendFsWindowToAnotherWorkspace) {
         ASSERT_CONTAINS(str, "floating: 1");
     }
 }
+
+TEST_CASE(floatingForceOnscreen) {
+    // Test: move a floating window around
+    NLog::log("{}Testing force-onscreen for existing floating windows", Colors::GREEN);
+    Tests::spawnKitty();
+    OK(getFromSocket("/dispatch hl.dsp.window.float({ action = 'on' })"));
+    OK(getFromSocket("/dispatch hl.dsp.window.resize({ x = 100, y = 100, relative = false })"));
+
+    // No constraints
+    OK(getFromSocket("/eval hl.config({ misc = { float_force_onscreen = 0 }})"));
+    OK(getFromSocket("/dispatch hl.dsp.window.move({ x = -110, y = 0, relative = false })"));
+    ASSERT_CONTAINS(getFromSocket("/activewindow"), "at: -110,0");
+
+    // Partially onscreen
+    OK(getFromSocket("/eval hl.config({ misc = { float_force_onscreen = 1 }})"));
+    OK(getFromSocket("/dispatch hl.dsp.window.move({ x = -110, y = 0, relative = false })"));
+    ASSERT_CONTAINS(getFromSocket("/activewindow"), "at: -50,0");
+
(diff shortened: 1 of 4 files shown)


tekstryder:
>Waiting until I can personally test this on main

Have you? If so, I will too.

This could finally solve (the most annoying aspect of) https://github.com/hyprwm/Hyprland/discussions/12398#discussioncomment-15978510.

LRitzdorf:
I have, yes. Should probably update the initial message with that.

fwiw, the "fully onscreen" logic in here is just more of what we already had for ensuring floats spawn initially onscreen. It effectively says:
- Move window such that top left corner is onscreen
- Move window such that bottom right corner is onscreen

No resizing happens, so if the window is too large, you just get the bottom and right edges in view, with the top and left still overflowing offscreen.

That could be changed, though... thoughts?

tekstryder:
Built against main @ https://github.com/hyprwm/Hyprland/commit/62d890eaccc3fcbdc00f85fe27d611d0a8681e7a. This is greatly improved from the state of issue described in https://github.com/hyprwm/Hyprland/discussions/12398 which has been a problem since way back when commit https://github.com/hyprwm/Hyprland/commit/c2670e9ab90bd657e87a1da2c5322e9007dce01f landed.

I've never had an issue with newly created floating windows... so not much for me to test there as they are always sized and centered by rules.

When toggling tiled windows to float with this patch, the problem of floated windows being oversized and/or outside the bounds of a display is almost entirely mitigated with `float_force_onscreen = 2` 👍🏻

Not really worth a screen recording... but when I open 3+ tiled windows and toggle floating on them in random order lots of time to test this behavior, I can sometimes get a window to exceed the edge of a multimon setup by a few, but no more than say ~10ish pixels. Regardless, a huge improvement from windows' title bars disappearing up offscreen, and in every case I can still resize by dragging a corner.

Dragging a floated window offscreen on any display's edge snaps it back within the display bounds as expected.

Longwinded feedback. Thanks for addressing one of my last niggles with hypr vs sway behavior.

LRitzdorf:
Rebased. Will probably continue doing so if enough changes accumulate before this gets merged

vaxerski (approved):

PR #15499, opened 2026-07-20
example: update hyprland.lua default springs

cebem1nt:

Since v0.56 springs changed a bit, so default windows animations are now slow. This PR fixes it

nope

ready

### DIFF
--- a/example/hyprland.lua
+++ b/example/hyprland.lua
@@ -140,7 +140,7 @@ hl.curve("almostLinear",   { type = "bezier", points = { {0.5, 0.5},   {0.75, 1}
 hl.curve("quick",          { type = "bezier", points = { {0.15, 0},    {0.1, 1}     } })
 
 -- Default springs
-hl.curve("easy",           { type = "spring", mass = 0.55, stiffness = 229.2633, dampening = 18.8273644 })
+hl.curve("easy",           { type = "spring", mass = 1, stiffness = 218.2633, dampening = 18.8273644 })
 
 hl.animation({ leaf = "global",        enabled = true,  speed = 10,   bezier = "default" })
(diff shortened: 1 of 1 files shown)


erstarr:
I'm against having the default `window` and `windowIn` animations being "bouncy" (i.e. using spring "in the way it was intended" as some would say). Default should be the most generic to be applicable to the largest crowd.

The current default dampening is the critical dampening of the default mass and stiffness (i.e. no bounce) which practically makes it little different than a bezier curve anyway which, while it does do some extra steps, was a good default before 0.56 made it slower.

Your MR sets the dampening below the critical value thus making it bouncy. Not a good default imo

Simply making it faster to match the prev value is preferable.

Dregu:
I think it's good that the **example** config has **examples** of many different features, like springs behaving like springs. It's not supposed to be used as is anyway, it's supposed to be kinda usable to open a terminal.

example/hyprland.lua
```diff
@@ -140,7 +140,7 @@ hl.curve("almostLinear",   { type = "bezier", points = { {0.5, 0.5},   {0.75, 1}
 hl.curve("quick",          { type = "bezier", points = { {0.15, 0},    {0.1, 1}     } })
 
 -- Default springs
-hl.curve("easy",           { type = "spring", mass = 1, stiffness = 71.2633, dampening = 15.8273644 })
+hl.curve("easy",           { type = "spring", mass = 0.55, stiffness = 229.2633, dampening = 18.8273644 })
```
vaxerski: please keep mass at 1

erstarr:
> It's not supposed to be used as is anyway

I myself used the animation defaults all this time, and i can guarantee you that most people do too. Something not so bouncy will be best I think even if it must be bouncy to showcase the feature.

If you think of it as the _example config_ as and **example** config  and not a **default** config; you're right of course.

moizekan:
W

PR #15515, opened 2026-07-21
Fix DPMS wake triggers lockdead screen

CaffeinatedTech:
## Disregard
Sorry, I was working on old code and fixed things that have already been resolved.  Don't waste your time on this.

While the screen is locked, a DPMS off/on cycle can cause the lockdead screen to appear (Oopsie daisy).  This has been reported several times in issues, discussions, reddit, other repos for distros that use hyprland.  It seems that not everyone can reproduce it, probably hardware dependent.  My system has suffered from it intermittently since switching to Omarchy, I think because my monitor enters sleep mode and disconnects from the host system and the kernel reports a hotplug removal.  I had AI help diagnose the issue and isolate the cause.

The DPMS off/on cycle causes the lock client to be disconnected from the Wayland compositor.  After `misc:lockdead_screen_delay` ms (defaults to 1000ms), Hyprland shows the lockdead (Oopsie daisy) screen requiring manually  allowing relock and restarting the lock client from a different TTY.  On wake the kernel reports hotplug add with the same monitor name, but hyprland drops the shared pointer to the old output, and the destructor destoys all outputs for all clients.  The client now holds a stale wl_output id, any in-flight requests use the invalid object, wayland disconnects the client (lock screen) so we get the lockdead screen.

**fix:** defer wl_global_destroy via zombie resource list.
When a monitor is removed, the per-client wl_output resources are moved from m_outputs to a new m_zombieOutputs list.  We also override onDisplayDestroy to skip wl_global_destroy while zombie resources exist.

We later clean up the zombie list when the client eventually destroys the stale wl_output after processing the global_remove event.  When the last zombie is gone the destructor runs and wl_global_destroy fires normally at a safe time when no client references remain.

**testing:** Many `Lock -> DPMS off -> DPMS on -> unlock` cycles can no longer reproduce the crash event after applying the changes.
Manually killing the lock client with pkill for example (simulating the lock crashing) still shows the lockdead screen as expected.

comments in code explain reasoning for additional guards etc.  Its all around the monitor disconnect/reconnect flow while locked.

Ready for merging, would appreciate a more experienced eye looking over the code changes.


CaffeinatedTech:
Sorry, I'm a drongo.  I was working on old code, and fixed things that were already fixed.  There remains a slight improvement guard which I will open a new PR for.  don't waste your time on this one.

PR #15516, opened 2026-07-21
sessionlock: suppress lockdead when client is still alive

CaffeinatedTech:
During DPMS output recreation when a monitor disconnects and reconnects or sleep mode disconnects the screen from the host, the lock client may briefly have zero surfaces even though the connection is healthy.  This guard prevents `shallConsiderLockMissing` from showing the lockdead screen when the lock is still good.

Others have fixed other DPMS issues, this should be a defense in-depth approach to help prevent the annoying lockdead screen when your system goes to sleep.

ready to merge.

### DIFF
--- a/src/managers/SessionLockManager.cpp
+++ b/src/managers/SessionLockManager.cpp
@@ -245,7 +245,11 @@ bool CSessionLockManager::shallConsiderLockMissing() {
     if (!m_sessionLock)
         return true;
 
-    static auto LOCKDEAD_SCREEN_DELAY = CConfigValue<Config::INTEGER>("misc:lockdead_screen_delay");
+    // Lock client is still connected — surfaces may be temporarily gone
+    // during DPMS output recreation, not a real crash. Don't show lockdead.
+    if (m_sessionLock->lock && m_sessionLock->lock->good())
+        return false;
 
+    static auto LOCKDEAD_SCREEN_DELAY = CConfigValue<Config::INTEGER>("misc:lockdead_screen_delay");
     return m_sessionLock->lockTimer.getMillis() > *LOCKDEAD_SCREEN_DELAY;
 }


vaxerski (commented):
wouldn't this show the no-instruction one? What does it show instead, I assume black?

CaffeinatedTech:
It should prevent the lockdead screen from appearing when the lock is still good, but the client hasn't processed the new display yet (when the monitor re-connects on wake), it might have a queue or in-flight commands to an invalid output id.  Once lockdead shows, its game over because default settings prevent re-lock.  In my testing, the monitor is still initialising and there is nothing on screen until it fully wakes.  By then, the lock client should be drawing to the correct surface, or if its crashed, lockdead will be there.

That's my understanding of it anyway.  I'm happy to be wrong.

vaxerski:
there are two lockdeads. One is if the lockscreen is alive but doesnt have a surface for a given monitor, and one when the app dies. Just because the first one is showed shouldnt stop the app from opening a surface.

If it does, that is the problem and not the screen

CaffeinatedTech:
Ahh, I see what you mean.  Following it back up the chain `renderLockscreen` would call `renderSessionLockMissing` if `shallConsiderLockMissing` returned true.  `renderSessionLockMissing` checks if there are `anySessionLockSurfacesPresent` and shows image2 (without instructions).

The reason I'm here is because I've been experiencing a problem where when I return to my locked computer, my monitor has gone to sleep and sometime when I wake it up, I end up with the full lockdead screen with instructions and the TTY routine to recover it.  I noticed plenty of people complaining of this, but not everyone can reproduce it.  Found out that some monitors do more of a full shutdown in sleep mode and fully disconnect from the host causing hotplug events.  

In 0.55 there seemed to be a race condition where the output is dropped, then when it wakes it is assigned a new id, but the lock client still has a queue or in-flight commands for the old output id so it was getting killed and we got lockdead.  It seems that a different bug fix has solved this part - the DPMS disconnect/reconnect bug (fix monitor re-enabling on hotplug when dpms is off #14818 ).  I think that was the root cause.  But I also found that sometimes during the monitor re-initialisation the `LOCKDEAD_SCREEN_DELAY` of 1000ms was still catching it out.  Do you think that `anySessionLockSurfacesPresent` would return true or false if the lock client still didn't have the new monitor output while it wakes up with the DPMS on trigger?

vaxerski:
if you end up with actual dead (instructions) then either hyprlock actually died or hyprland fucked up.

I'd check `WAYLAND_DEBUG` of hyprlock first, see if it isnt crashing, and if it is, why

CaffeinatedTech:
Yeah thank you.  After looking at it more closely it looks like this is just a cosmetic fix anyway and the problem is somewhere else.  Even with the latest build including the hotplug fixes I still got lockdead yesterday.  I'll ratchet up the logging and find out more.

I'm grateful for your time, and I'm sorry for wasting it.  I'll close this PR now, it doesn't solve the real problem.

PR #15538, opened 2026-07-22
compositor: add a deprecation notice to .conf configs

vaxerski:
Adds a notification about .conf being removed starting 0.57


alba4k:
should I bother translating this? also, shouldn't this have been added in 0.56? or will it be in 0.56.1

vaxerski:
it will be in .1

PR #15543, opened 2026-07-22
desktop/windowRule: fix exec rules leaking

vaxerski:
ref #15488

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -1229,6 +1229,80 @@ TEST_CASE(execRulesTagMutation) {
     Tests::killAllWindows();
 }
 
+TEST_CASE(execRulesDescendantProcess) {
+    OK(getFromSocket("/dispatch hl.dsp.exec_cmd('kitty --class kitty_exec_workspace & wait', { workspace = '2 silent' })"));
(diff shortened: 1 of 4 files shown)


Dregu:
I think you fixed the lua style rules don't work on forks issue too. So possibly #14823, #15082, #15196, but I'm not testing flatpaks. This guy works now, didn't work before `hyprctl eval 'hl.exec_cmd("bash<<<foot", { float = true })'`

vaxerski:
sorry, accident

tannerellen:
> I think you fixed the lua style rules don't work on forks issue too. So possibly #14823, #15082, #15196, but I'm not testing flatpaks. This guy works now, didn't work before `hyprctl eval 'hl.exec_cmd("bash<<<foot", { float = true })'`

I think it fixed them partially. Most flatpaks seem to work now specifying things like float and size. Some still don't. For example the Cider music player flatpak doesn't and kdenlive does but only for the launch screen. So it appears apps like that are doing something more/different. Good progress though.

Dregu:
Well exec rules have only ever worked for the first window any process opens, that's what they're for, sadly that often means a launcher or a loading screen. Window rules are there to do the things exec rules can't...

PR #15555, opened 2026-07-22
Lua: support workspace selectors in `hl.get_workspace()`

LRitzdorf:

Rather than always treating string arguments to `hl.get_workspace()` as workspace names, this parses them as proper workspace selectors, so args like `e+1` work now. Names are still supported, but must be prefixed with `name:` as is standard for other spots where workspace selectors are used.

There's no wiki PR for this, because the wiki uh, *already* describes the argument as a "workspace selector" (which is what spawned this in the first place).

Nerp

Yerp

### DIFF
--- a/hyprtester/src/tests/main/workspaces.cpp
+++ b/hyprtester/src/tests/main/workspaces.cpp
@@ -985,3 +985,25 @@ TEST_CASE(workspaceRenameUpdatesRules) {
         ASSERT_CONTAINS(str, "at: 42,42");
     }
 }
+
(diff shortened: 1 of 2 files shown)


vaxerski (dismissed):

vaxerski:
can you add a test

LRitzdorf:
Aaaand yup, this is why we test. Old version would focus and then immediately destroy nonexistent workspaces; fixed and tested-for now

vaxerski:
:)

vaxerski (approved):

PR #15563, opened 2026-07-23
layout/drag: add binds:drag_center_window

LegendaryBanana37:
Currently when dragging a window from a tiled position or fullscreen it moves itself to center itself on the mouse, this PR adds an option to disable this behavior and instead just move based on the mouse cursors delta, via setting binds:drag_center_window to false. Defaults to true so current behavior is unchanged.

I also noticed during testing that while either gaps_in or border_size are non zero, when moving from the tiled to floating transition, the size of the window increases and jumps a bit, this was present in both the current version and in this PR. Also in the live version when you click a tiled window and don't move the mouse, it tweens to a smaller size, and then immediately jumps up again in size when dragged.

Ready for merging, tested both tiled and the fullscreen path with the option enabled and disabled.

Wiki PR: hyprwm/hyprland-wiki#1606

### DIFF
--- a/hyprtester/src/tests/main/drag.cpp
+++ b/hyprtester/src/tests/main/drag.cpp
@@ -1,0 +1,189 @@
+#include <cmath>
+#include <format>
+#include <string>
+#include <utility>
+
+#include "../../shared.hpp"
+#include "../../hyprctlCompat.hpp"
+#include "../shared.hpp"
+#include "tests.hpp"
+
+/// Parses an "X,Y" attribute (as returned by getAttribute for "at" and "size")
+/// into a pair of ints. Returns {0, 0} on a malformed value.
+static std::pair<int, int> parseXY(const std::string& value) {
+    const auto COMMA = value.find(',');
+    if (COMMA == std::string::npos)
+        return {0, 0};
+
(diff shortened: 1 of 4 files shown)


vaxerski (commented):
a better idea which would make sense for future development would be to store a hotspot at the beginning.

LegendaryBanana37:
i'm not sure what you mean by store a hotspot

vaxerski:
drag hotspot. Center is size / 2.

LegendaryBanana37:
I added a variable for drag hotspot, additionally, I changed the behavior of the non drag center version, so that when the window size changes on drag, the cursors position is based proportionally to where it was on the old window.

vaxerski (commented):
looks nice! Just write some tests for this please :)

You should have mouse and keyboard control via the plugin, so it shouldn't be hard.

LegendaryBanana37:
Added tests in a new file drag.cpp. It has 4 cases, tiled and fullscreen drags each with either drag center window on or off, the fullscreen tests floats and shrinks the window beforehand.

vaxerski (approved):
thanks!

PR #15568, opened 2026-07-23
keybinds: refactor, clean up code, move to new style

### LINKED ISSUE #15354: Allow binding left/right mods separately
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/15100

## Description

this used to be possible via a hack, but now is not anymore.

We should allow binds to specify which mod exactly the user wants to tackle via e.g. `ALT_L` or `CTRL_R`.

vaxerski:
What's up everyone it's BIGGIE CHEESE

image

Needs some testing because this is a chonker!!!


ImEdgaras:
SOMEONE SAY CHEESSE?
image

Dregu:
- MOD5 doesn't work at all. `hl.bind('MOD5+E', hl.dsp.exec_cmd'foot')` just makes a € (it's altgr, iso_level3_shift/mod5 in wev)
- `hl.bind('a+b', ...)` now makes it impossible to type a or b, even when non_consuming (which wasn't needed before anyway, but might make sense to set explicitly for modless binds)

I'll test more later, now I gotta revert this cause I can't do anything.

vaxerski:
@Dregu so did you get a chance to recheck?

src/keybinds/Manager.cpp
```diff
@@ -0,0 +1,898 @@
...
+}
+
+bool CKeybindManager::handleVT(xkb_keysym_t keysym) {
+    if (keysym < XKB_KEY_XF86Switch_VT_1 || keysym > XKB_KEY_XF86Switch_VT_12)
+        return false;
+
+    if (g_pCompositor->m_aqBackend->hasSession())
+        g_pCompositor->m_aqBackend->session->switchVT(keysym - XKB_KEY_XF86Switch_VT_1 + 1);
+    return true;
+}
```
gulafaran, line 873: this should guard to not VT switch to itself.

```
const auto CURRENT_TTY = g_pCompositor->getVTNr();
if (CURRENT_TTY.has_value() && *CURRENT_TTY == TTY)
    return true;
```

gulafaran (changes requested):

Dregu:
> @Dregu so did you get a chance to recheck?

Yeah nothing has changed.

vaxerski:
should be fixed. `a + b` requires non-consuming to not eat a, otherwise it strictly makes no sense because hyprland must consume a immediately because it doesn't know if b will be pressed. Re-sending a if you release a is just bad ux.

Dregu:
Well the previous things work. But still

- CTRL doesn't work at all now
- Modkey only `SUPER+Super_L` doesn't work
- Specifying _L/_R modifiers doesn't work
- Pressing and releasing modifiers in different order is weird, e.g.
  - Use combo `SUPER+SHIFT+2`, keep holding `SUPER`, then `SUPER+2` works, but after `SHIFT+SUPER+2` (different order) you have to release `SUPER` to use `SUPER+2` again

I can't keep testing all possible key combinations like this.

What seems better now is release triggers only when the triggering key gets released, not mods, that's good. Except it doesn't translate to mouse binds, e.g. dragging.

vaxerski:
thanks. I'll look into those. You don't need to find all issues immediately, I know this will take a while to work because it's a proper rewrite and we've accumulated tons of crust over the years

vaxerski:
@Dregu all of those should be fixed

Dregu:
Yes I think I have feature parity with all my old binds without fixing anything. L/R only mods also seem to work, haven't had those in a while. I don't plan to methodically test a bunch of stuff I don't already have in my config, but I can keep daily driving the pr since I can't find anything wrong with the stuff I do use.

> release triggers only when the triggering key gets released

This thing was short lived and already broke, but I think it should be added, i.e.

- `SUPER+X` shouldn't release on the release of SUPER, only on the release of X
- `SUPER+mouse:272` window drag should continue dragging without holding SUPER
- changing workspace during this drag should also be possible, it would be a pretty intuitive way to move window to different workspace

vaxerski:
those could be added later, this is mostly fixes and cleanups not features

vaxerski:
holy FUCK tests passed lfg

zjeffer:
I think this PR caused a bug with my Super + Super_L keybinds to open rofi: if I use any other keybind that contains super, rofi also opens at the same time. 

example:
```lua
hl.bind(mainMod .. " + SUPER_L",
    hl.dsp.exec_cmd("killall rofi || ~/.config/rofi/launchers/type-2/launcher.sh -show drun"),
    { release = true })
hl.bind(mainMod .. " + return", hl.dsp.exec_cmd("wezterm --config-file ~/.config/wezterm/wezterm.lua"))
```
mainMod is `SUPER`
Pressing `SUPER+return` opens wezterm, then releasing SUPER opens rofi. This is extremely annoying because after every workspace change, application launch, etc. I need to close rofi with Escape.

I just updated my whole system yesterday evening, and it's been a couple of weeks but I suspect it's caused by this PR.

gulafaran:
> I think this PR caused a bug with my Super + Super_L keybinds to open rofi: if I use any other keybind that contains super, rofi also opens at the same time.
> 
> example:
> 
> ```lua
> hl.bind(mainMod .. " + SUPER_L",
>     hl.dsp.exec_cmd("killall rofi || ~/.config/rofi/launchers/type-2/launcher.sh -show drun"),
>     { release = true })
> hl.bind(mainMod .. " + return", hl.dsp.exec_cmd("wezterm --config-file ~/.config/wezterm/wezterm.lua"))
> ```
> 
> mainMod is `SUPER` Pressing `SUPER+return` opens wezterm, then releasing SUPER opens rofi. This is extremely annoying because after every workspace change, application launch, etc. I need to close rofi with Escape.
> 
> I just updated my whole system yesterday evening, and it's been a couple of weeks but I suspect it's caused by this PR.

something like this then?

```
diff --git a/src/keybinds/Manager.cpp b/src/keybinds/Manager.cpp
index 9025dbfd..4e4983b4 100644
--- a/src/keybinds/Manager.cpp
+++ b/src/keybinds/Manager.cpp
@@ -765,6 +765,7 @@ void CKeybindManager::shadowBinds(const std::optional<SResolvedKey>& excluded, c
                                 [&bind](const auto& input) { return std::ranges::any_of(input.deferredBinds, [&bind](const auto& weak) { return weak.lock() == bind; }); }))
             continue;
 
+        const bool MOD_ONLY_RELEASE = bind->hasFlag(BIND_FLAG_RELEASE) && bind->chordSize() == 0;
         for (const auto& input : m_inputState.pressed()) {
             const auto device = input.device.lock();
             if (!device)
@@ -773,6 +774,11 @@ void CKeybindManager::shadowBinds(const std::optional<SResolvedKey>& excluded, c
             if (excluded && device == excludedDevice && input.key.code == excluded->code && input.key.event == excluded->event)
                 continue;
 
+            if (MOD_ONLY_RELEASE && bind->containsKey(input.key)) {
+                m_shadowed.emplace(bind);
+                break;
+            }
+
             if (bind->matches({
                     .heldKeys         = m_inputState.heldKeys(),
                     .trigger          = input.key,
```

vaxerski:
shouldnt that be just Super_L not super +?

zjeffer:
> shouldnt that be just Super_L not super +?

Has the same bug with just "Super_L" as the key. I don't think using just 'Super_L' did anything previously, I always had to do Super + Super_L, at least with the legacy config. Nice that that works now, but the same issue occurs.

@gulafaran I'll test that patch soon when I have more time.

fxzzi:
below still do not work, broken since lua as far as i remember

```lua
  bind({ mainMod, "CTRL" }, hl.dsp.window.drag(), { mouse = true })
  bind({ mainMod, "ALT" }, hl.dsp.window.resize(), { mouse = true })
```

relevant discussion https://github.com/hyprwm/Hyprland/discussions/14573

Dregu:
> below still do not work

But the examples in the wiki work fine, "CTRL" is not a specific key and shouldn't work.

moetayuko:
> > I think this PR caused a bug with my Super + Super_L keybinds to open rofi: if I use any other keybind that contains super, rofi also opens at the same time.
> > example:
> > ```lua
> > hl.bind(mainMod .. " + SUPER_L",
> >     hl.dsp.exec_cmd("killall rofi || ~/.config/rofi/launchers/type-2/launcher.sh -show drun"),
> >     { release = true })
> > hl.bind(mainMod .. " + return", hl.dsp.exec_cmd("wezterm --config-file ~/.config/wezterm/wezterm.lua"))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > mainMod is `SUPER` Pressing `SUPER+return` opens wezterm, then releasing SUPER opens rofi. This is extremely annoying because after every workspace change, application launch, etc. I need to close rofi with Escape.
> > I just updated my whole system yesterday evening, and it's been a couple of weeks but I suspect it's caused by this PR.
> 
> something like this then?
> 
> ```
> diff --git a/src/keybinds/Manager.cpp b/src/keybinds/Manager.cpp
> index 9025dbfd..4e4983b4 100644
> --- a/src/keybinds/Manager.cpp
> +++ b/src/keybinds/Manager.cpp
> @@ -765,6 +765,7 @@ void CKeybindManager::shadowBinds(const std::optional<SResolvedKey>& excluded, c
>                                  [&bind](const auto& input) { return std::ranges::any_of(input.deferredBinds, [&bind](const auto& weak) { return weak.lock() == bind; }); }))
>              continue;
>  
> +        const bool MOD_ONLY_RELEASE = bind->hasFlag(BIND_FLAG_RELEASE) && bind->chordSize() == 0;
>          for (const auto& input : m_inputState.pressed()) {
>              const auto device = input.device.lock();
>              if (!device)
> @@ -773,6 +774,11 @@ void CKeybindManager::shadowBinds(const std::optional<SResolvedKey>& excluded, c
>              if (excluded && device == excludedDevice && input.key.code == excluded->code && input.key.event == excluded->event)
>                  continue;
>  
> +            if (MOD_ONLY_RELEASE && bind->containsKey(input.key)) {
> +                m_shadowed.emplace(bind);
> +                break;
> +            }
> +
>              if (bind->matches({
>                      .heldKeys         = m_inputState.heldKeys(),
>                      .trigger          = input.key,
> ```

the patch doesn't appear to work for me

PR #15579, opened 2026-07-24
feat: add missing lua stub

nath1416:
Adds missing hl. functions in stub.

Ready to merge


meta/generateLuaStubs.py
```diff
@@ -728,6 +734,19 @@ def generate_stub(root: Path) -> str:
     )
     lines.append("")
 
+    lines.extend(
+        emit_class_block(
+            "HL.PLugin",
```
vaxerski: Plugin?

nath1416: Maybe `PluginInformation` would be a better name?

Dregu: Smaller L would be a better name...

vaxerski: that is what I meant.

nath1416: Oups, I miss it, just pushed the fix

vaxerski (approved):

PR #15598, opened 2026-07-25
hdr: reset surface metadata on change

gulafaran:
if we dont reset surface metadata, the monitor is stuck in using the surfaces metadata, and doesnt restore back ours. meaning after DS and gaming and returning, colors stay stuck fringed until a new metadata modeset.

however this causes a modeset on returning that isnt wanted if a notification temporarly shows and vanishes. but a future PR in perhaps using a timer if DS reenters in a short bit we dont care about the notif fringing?

better then leaving the monitor permanently wrong colors.

### DIFF
--- a/src/output/Monitor.hpp
+++ b/src/output/Monitor.hpp
@@ -358,7 +358,8 @@ namespace Monitor {
         bool                                                               m_renderingInitPassed = false;
 
         PHLWINDOWREF                                                       m_previousFSWindow;
-        bool                                                               m_needsHDRupdate = false;
+        bool                                                               m_needsHDRupdate         = false;
+        bool                                                               m_hdrMetadataFromSurface = false;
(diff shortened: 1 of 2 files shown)


gulafaran:
@UjinT34 this needs your input, it makes modesets happend on notifs showing momentarly, but it restores colors atleast when leaving.

vaxerski:
@codex review

UjinT34:
lgtm
Users might find some weird combo that doesn't work as expected. If this happens then it's probably better to redo the whole instead of adding extra flags.

vaxerski (approved):
👍

fxzzi:
Can we have a config option to not apply DS in HDR situations for now then? Since this makes DS + HDR close to unusable for me as I get notifications often :)

PR #15607, opened 2026-07-26
WIP: fixing mouse focus on window map/unmap/regroup

kolayne:

Fixes problems with lost mouse focus when adding new window to a group or closing a fullscreen window. #12192 

See comments.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -188,6 +188,7 @@ void CInputManager::simulateMouseMovement() {
     mouseMoveUnified(Time::millis(Time::steadyNow()));
 }
 
+// TODO: :/
 void CInputManager::sendMotionEventsToFocused() {
     if (!Desktop::focusState()->surface() || isConstrained())
         return;
--- a/src/desktop/view/Group.cpp
+++ b/src/desktop/view/Group.cpp
@@ -161,6 +161,7 @@ void CGroup::add(PHLWINDOW w, std::optional<size_t> index) {
     }
 
     m_target->recalc();
+    g_pInputManager->sendMotionEventsToFocused();
 }
 
 void CGroup::remove(PHLWINDOW w, Math::eDirection dir, eRemoveFromGroupReason reason) {
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -2675,6 +2675,7 @@ void CWindow::unmapWindow() {
         if (!candidate && m_workspace && m_workspace->getWindowCount() == 0)
             g_pInputManager->refocus();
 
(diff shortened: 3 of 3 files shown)


src/managers/input/InputManager.cpp
kolayne, line 224: When `CInputManager::sendMotionEventsToFocused` is called from either `CGroup::add` (added in this PR) or from `CWindow::unmapWindow` (where I added a TODO comment) for a fullscreen/maximized/groupped window, each of these two conditions holds, causing this function to return. Notably, `CWindow::unmapWindow` exits fullscreenmode for fullscreen windows, but this still doesn't help.

I'm confused and wonder what the write way to fix it is. Who controls the aliveAndVisibility and the box? Should the caller of `CInputManager::sendMotionEventsToFocused` first make a call to some compositor function to get windows focus states straight, and then call this one?

@vaxerski, your help is welcome.

vaxerski: Ah yes. I think visible() there makes little sense. We could refactor that to something like valid() which just checks if it's mapped and good.

kolayne: But what about the box? It should be there, since it is used in calculations later, but it is null. How does the box come and go?

vaxerski: visible() is false because window is obstructed by fullscreen thus box returns nullopt

kolayne: Hey. I'm back on this one.

I still don't get what the right way to fix this would be. Some parts of code seem to expect `getSurfaceBoxGlobal` to return `std::nullopt` when the window is not visible or does not accept input. Should I change all the places that expect that?.. Should I add a new method to `WLSurface`, and then to `IView` and all the classes that implement it?.. (is it even possible for all kinds of views?)

Or is there somewhere else we can get the information that we need from  `BOX`, other than this inactive view?

vaxerski: that sounds wrong. A window can still have a box and not accept input. You should fix those places yes

src/desktop/view/window/Window.cpp
```diff
@@ -239,7 +239,7 @@ bool CWindow::desktopComponent() const {
 }
 
 std::optional<CBox> CWindow::surfaceLogicalBox() const {
-    if (!mapped() || !acceptsInput() || !alphaNonZero())
+    if (!mapped() || !acceptsInput())
         return std::nullopt;
```
kolayne, line 243: I'm not really sure what I'm doing here 😅

Does this make sense?

vaxerski: it does yes

vaxerski:
can we get tests?

PR #15637, opened 2026-07-28
screenshare: fix toplevel capture missing software cursor

Dregu:

Fixes software cursors missing from toplevel capture. The function that was supposed to be *forced* to draw cursors still failed when it thought forcing was unnecessary. It thought wrong, drawing cursor on toplevel capture is always necessary.

#15604

Also tests if vouch bot is case-sensitive...

It's kinda silly, but someone designed the first force flag to not always work, and I think they also have a point, what can you do...

Ready.


Dregu:
Well :D

vaxerski:
HASHIMOTOOOO

PR #15638, opened 2026-07-28
screenshare: fix toplevel capture missing software cursor

Dregu:

Fixes software cursors missing from toplevel capture. The function that was supposed to be *forced* to draw cursors still failed when it thought forcing was unnecessary. It thought wrong, drawing cursor on toplevel capture is always necessary.

#15604

Fingers crossed...

It's kinda silly, but someone designed the first force flag to not always work, and I think they also have a point, what can you do...

Ready.

### DIFF
--- a/src/managers/screenshare/ScreenshareFrame.cpp
+++ b/src/managers/screenshare/ScreenshareFrame.cpp
@@ -353,7 +353,7 @@ void CScreenshareFrame::renderWindow() {
 
     CRegion fakeDamage = {0, 0, INT16_MAX, INT16_MAX};
     Pointer::mgr()->renderSoftwareCursorsFor(PMONITOR->m_self.lock(), NOW, fakeDamage,
-                                             g_pInputManager->getMouseCoordsInternal() - PWINDOW->position(Desktop::View::IGeometric::GEOMETRIC_CURRENT), true);
+                                             g_pInputManager->getMouseCoordsInternal() - PWINDOW->position(Desktop::View::IGeometric::GEOMETRIC_CURRENT), true, true);
 }
 
 void CScreenshareFrame::render() {
--- a/src/pointer/PointerManager.cpp
+++ b/src/pointer/PointerManager.cpp
@@ -638,23 +638,23 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
     return buf;
 }
 
-void CPointerManager::renderSoftwareCursorsFor(PHLMONITOR pMonitor, const Time::steady_tp& now, CRegion& damage, std::optional<Vector2D> overridePos, bool forceRender,
-                                               bool actuallyForceRender) {
+void CPointerManager::renderSoftwareCursorsFor(PHLMONITOR pMonitor, const Time::steady_tp& now, CRegion& damage, std::optional<Vector2D> overridePos, bool screencopy,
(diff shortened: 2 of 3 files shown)


vaxerski:
sorry @Dregu you aint on the vouch list dam so sad

src/pointer/PointerManager.cpp
```diff
@@ -638,7 +638,8 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
     return buf;
 }
 
-void CPointerManager::renderSoftwareCursorsFor(PHLMONITOR pMonitor, const Time::steady_tp& now, CRegion& damage, std::optional<Vector2D> overridePos, bool forceRender) {
+void CPointerManager::renderSoftwareCursorsFor(PHLMONITOR pMonitor, const Time::steady_tp& now, CRegion& damage, std::optional<Vector2D> overridePos, bool forceRender,
+                                               bool actuallyForceRender) {
```
vaxerski: what the fuck is this naming bro

Dregu: Yeah we can laugh at that now that the pr is open, who made the first forceRender not actually force anything, what should I call that then :D

umbrageodotus: why don't you make it one bool that actually forces it, and if the consuming fn wants the old behavior (which I assume atleast some do, as you didn't fully replace the var), you copy the condition and use it as force.

for example (in pseudocode):

instead of:
```rs
fn example(bool force, bool actuallyForce) {
    if (force && xyz || actuallyForce) { doCoolShit(); }
}

example(true, false);
```

do:

```rs
fn example(bool force) {
    if (force) { doCoolShit(); }
}

example(xyz);
```

or at most extract xyz into a fn if it's used that much, prolly not useful tho

ignore that the pseudocode is rusty plz

umbrageodotus: or, if cpp lets you (idk cpp), make the default value be the evaluation of, in the example, `xyz`, and consumers can set it to force false (previous behavior of it being false) or force true (new behavior)

Dregu: I guess technically the purpose of forceRender is actually "this is for screencopy" so I could rename them to `screencopy` and `forceRender` to look less stupid.

vaxerski (approved):
better. Can the affected test?

vaxerski:
@codex review

src/managers/screenshare/ScreenshareFrame.cpp
```diff
@@ -353,7 +353,7 @@ void CScreenshareFrame::renderWindow() {
 
     CRegion fakeDamage = {0, 0, INT16_MAX, INT16_MAX};
     Pointer::mgr()->renderSoftwareCursorsFor(PMONITOR->m_self.lock(), NOW, fakeDamage,
-                                             g_pInputManager->getMouseCoordsInternal() - PWINDOW->position(Desktop::View::IGeometric::GEOMETRIC_CURRENT), true);
+                                             g_pInputManager->getMouseCoordsInternal() - PWINDOW->position(Desktop::View::IGeometric::GEOMETRIC_CURRENT), true, true);
```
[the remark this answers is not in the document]
vaxerski: yeah no need

vaxerski:
I guess they dont care, classic.

code looks good.

PR #15645, opened 2026-07-29
compositor: double buffered states needs their own bit

gulafaran:
6484f43 introduced state merging, which showed a flaw with mako. a second commit with just changing geometry was handled as a "empty" commit and ignored. all protocols that listen to commit and is double buffered as in state applied on next commit needs their own updated.bits.flag to ensure the statequeue knows something is pending.

synced subsurface also needs to set the parent flag, because its waiting for it to commit.

we also have to widen the union to uint32_t, we are running out of bits.

fixes: https://github.com/hyprwm/Hyprland/discussions/15641


tekstryder:
This fixed mako notifications. Thx!

vaxerski (approved):

PR #15646, opened 2026-07-29
desktop: render IME above all desktop components

vaxerski:
Put IME popups above everything in the rendering and input stack.

Even on lockscreens, IME could be requested. Maybe the user wants to input some username. Etc.

### DIFF
--- a/src/managers/input/InputManager.cpp
+++ b/src/managers/input/InputManager.cpp
@@ -357,9 +357,11 @@ void CInputManager::mouseMoveUnified(uint32_t time, bool refocus, bool mouse, st
     }
 
     // check for windows that have focus priority like our permission popups
-    pFoundWindow = Desktop::viewState()->hitTest().windowAt(mouseCoords, Desktop::View::FOCUS_PRIORITY);
-    if (pFoundWindow)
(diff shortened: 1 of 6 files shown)


vaxerski:
@codex review please

Dregu:
I would still argue that it's weird to draw popups on top of completely unrelated overlay layers, popups should just draw on top of whatever their parent surface is. If that parent *is* overlay, then the ime should be on top of it of course.

vaxerski:
correct, but at the moment it's not exactly possible to detemine that other than hacks like "keyboard focus" which doesn't _have to_ be true technically I think

vaxerski:
@codex review pleae

shinnkka:
Thank You! Btw, I initially thought the screen locker I was using was broken under Hyprland, since it completely ignored all my input. It turned out that the IME popup was being covered by another layer.

vaxerski:
> correct, but at the moment it's not exactly possible to detemine that other than hacks like "keyboard focus" which doesn't have to be true technically I think

nvm I was wrong, we have 1-to-1 mapping to a wl_surface, but at any rate the current hyprland input stack is not made for this so no, not in this mr

vaxerski:
@codex review please

PR #15651, opened 2026-07-29
config/lua: fix forwarding of details to lua values

vaxerski:
astaghfirullah

### DIFF
--- a/src/config/lua/types/LuaConfigUtils.cpp
+++ b/src/config/lua/types/LuaConfigUtils.cpp
@@ -29,19 +29,19 @@ UP<ILuaConfigValue> Lua::fromGenericValue(SP<Config::Values::IValue> v) {
     };
 
     if (auto p = dc<Config::Values::CIntValue*>(v.get()))
-        return withRefresh(makeUnique<CLuaConfigInt>(p->defaultVal()));
+        return withRefresh(makeUnique<CLuaConfigInt>(p->defaultVal(), p->m_min, p->m_max, p->m_map));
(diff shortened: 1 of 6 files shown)


wabarTM:
it uses wrong values :/
rounding_power 1 aka triangle corners are erroring out and with them "enabled" rounding cap of 20 makes no sense. imo rounding cap of none would be better

[src/config/values/ConfigValues.cpp](https://github.com/hyprwm/Hyprland/blob/2b1723adda7da821bb503ef4de7abb42d746ac51/src/config/values/ConfigValues.cpp#L202)

[src/config/lua/bindings/LuaBindingsInternal.hpp](https://github.com/hyprwm/Hyprland/blob/2b1723adda7da821bb503ef4de7abb42d746ac51/src/config/lua/bindings/LuaBindingsInternal.hpp#L70)

and triangle corners were removed fom wiki recently because of that inconsistancy sad_face_emoji

Dregu:
Yeah this feature seems very endangered for some reason, worked for a long time even though the limits in config descriptions weren't updated when this was last fixed in https://github.com/hyprwm/Hyprland/commit/a209f9911ceb83e17a3fc0269ebea1a7d8587784, cause the limits were never enforced in the first place. Now the wrong limits are enforced.

Dregu:
Also as wabar pointed out with the two links, window rules have completely different limits for overriding the same global setting, how about fetching the limits from the same point of truth...

vaxerski:
I mean it doesn't look wrong to me, just looks like I set a suboptimal range.

PR #15654, opened 2026-07-29
config/values: adjust requirements for values

vaxerski:
anything else you wanna complain about? @wabartm


wabarTM:
thats only a groupbar rounding, decoration.rounding_power needs to be changed too
and rounding max cap is still at 20

in the else category: "special" gesture action works only once and some damaging issues with zoom/direct scanout
but i will report them in discussion properly later this day

tekstryder:
I used groupbar `indicator_height  = 0` until https://github.com/hyprwm/Hyprland/commit/2b1723adda7da821bb503ef4de7abb42d746ac51.

Any chance getting that back?

vaxerski:
adjusted. Anything else?

wabarTM:
in gestures action can be a table, but lua stub doesnt have that

nothing else i can think off

vaxerski:
thats not here

PR #15657, opened 2026-07-30
opengl: invalidate depth too

gulafaran:
previously we tried using stencil only GL_STENCIL_INDEX8 but some drivers just simply sucks and gave us incomplete framebuffers. so we are actually creating GL_DEPTH24_STENCIL8 but disabling the depth with glDisable(GL_DEPTH_TEST);

but this means we need to invalidate the depth too, even if we really didnt use it.

see : https://github.com/hyprwm/Hyprland/pull/13067

### DIFF
--- a/src/render/OpenGL.cpp
+++ b/src/render/OpenGL.cpp
@@ -849,7 +849,7 @@ void CHyprOpenGLImpl::end() {
         g_pHyprRenderer->m_renderData.pMonitor->resources()->forEachUnusedFB(
             [](const auto& fb) {
                 fb->bind();
-                GLFB(fb)->invalidate({GL_STENCIL_ATTACHMENT, GL_COLOR_ATTACHMENT0});
+                GLFB(fb)->invalidate({GL_DEPTH_STENCIL_ATTACHMENT, GL_COLOR_ATTACHMENT0});
             },
             false);
     }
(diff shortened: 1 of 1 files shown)


gulafaran:
some unrelated keybind submap test fails

src/render/OpenGL.cpp
```diff
@@ -849,7 +849,7 @@ void CHyprOpenGLImpl::end() {
         g_pHyprRenderer->m_renderData.pMonitor->resources()->forEachUnusedFB(
             [](const auto& fb) {
                 fb->bind();
-                GLFB(fb)->invalidate({GL_STENCIL_ATTACHMENT, GL_COLOR_ATTACHMENT0});
+                GLFB(fb)->invalidate({GL_DEPTH_STENCIL_ATTACHMENT, GL_COLOR_ATTACHMENT0});
```
vaxerski, line 852: didnt you mean to invalidate 3 instead of replacing?

vaxerski (approved):

src/render/OpenGL.cpp
gulafaran: `then attachments may contain GL_COLOR_ATTACHMENTi, GL_DEPTH_ATTACHMENT, GL_STENCIL_ATTACHMENT, and/or GL_DEPTH_STENCIL_ATTACHMENT.`

i assume GL_DEPTH_STENCIL_ATTACHMENT means both?

vaxerski: I have no idea. Check khronos docs.

gulafaran: https://wikis.khronos.org/opengl/index.php?title=Framebuffer_Object

`GL_DEPTH_STENCIL_ATTACHMENT: This is shorthand for "both depth and stencil". The image attached becomes both the depth and stencil buffers.`

must be both for invalidate too

vaxerski: ok

vaxerski (approved):

PR #15660, opened 2026-07-30
Minor fullscreen fixes for 15564

vaxerski:
Fixes movement in fullscreen
Fixes scrolling fullscreen full behavior
Adds test

@erstarr

### DIFF
--- a/hyprtester/src/tests/main/scroll.cpp
+++ b/hyprtester/src/tests/main/scroll.cpp
@@ -344,6 +344,67 @@ TEST_CASE(scroll_LAYOUT_HANDLED_maximized) {
     }
 }
 
+TEST_CASE(scroll_LAYOUT_HANDLED_fullscreenRetainsGeometryWhileScrolling) {
+    OK(getFromSocket("/eval hl.config({ general = { layout = 'scrolling' } })"));
+
+    SPAWN_KITTY("kitty_scroll_A");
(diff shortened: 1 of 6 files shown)


erstarr:
Also if this MR is to fix the `window.focus({direction})` in scrolling FS in addition to window geometry thing, either copy over all the fixes from https://github.com/hyprwm/Hyprland/pull/15569 or let this MR solely fix the window geometry for non-covering FS windows in scrolling

vaxerski:
why? I don't see any regressions

erstarr:
15569 is more comprehensive in fixing window.focus({direction/window}) related bugs.

It includes comprehensive testing of both default and layout handled FS with window.focus({direction}) (inc. testing with the 2 config opts it also fixes), fixes inconsistencies with 2 focus related config opts that interacts with window.focus({direction and window}), and iirc the method to fix focus direction you used here and I used there isn't the same for some reason - this I can only investigate as to why later today.

It's less about regressions but methodically fixing both bugs comprehensively

vaxerski:
okay. Address the review there and we'll get that one merged, then I'll rebase this to fix the scroll bug.

erstarr:
> okay. Address the review there and we'll get that one merged, then I'll rebase this to fix the scroll bug.

15569 is gtg if you and the oracle are happy with it.

Just need to remove the line changed in `WindowQuery.cpp` for this one - other than the review changes anyway

PR #15661, opened 2026-07-30
renderer/gl: implement swappable blur variants

vaxerski:
Adds multiple variants to the blur shader.

- kawase (default)
- frost (ice type thing)
- ripple (click to ripple)
- drops (rain)
- water (similar to ripple, kinda ass)
- fluid_jar (super cool)
- prism (triangular refraction)
- heat_shimmer (wavy thing)
- acrylic (liquid ass type shit)
- aurora (similar to heat shimmer, but more vertical distortion plus gradients)
- haze (pearlescent boost)


Dregu:
the piss jar is sideways on rotated monitor

wabarTM:
ripple can go from one workspace to another, unless this is intended

https://github.com/user-attachments/assets/4d728985-09e5-443d-b2f8-1743b8afb5e1

with long ripple duration its trail disappears

https://github.com/user-attachments/assets/89ea7163-addc-48ed-86a6-09be722781d2


add an options that makes jar water hit top border too?

can variant/whole blur be added to windowrules?
e.g. for terminals you use jarate, but default to ripple

Dregu:
If you shake the piss jar enough it collapses into a singularity could be the weirdest bug report I've written.

gulafaran:
your tests fails because you expect different values then what the default config options are set at.

drops:speed, expect 0.F , default 3.F etc

src/render/gl/blur/FluidJar.cpp
```diff
@@ -0,0 +1,651 @@
...
+void CFluidJarBlurProvider::scheduleNextFrame(const SState& state) const {
+    const auto window = state.window.lock();
+    if (!window)
+        return;
+    g_pHyprRenderer->damageWindow(window);
+}
+
+void CFluidJarBlurProvider::pruneStates() {
+    std::erase_if(m_states, [](const auto& state) { return state.window.expired(); });
+}
```
gulafaran: this might technically "leak" if the window is blurred once, the states are kept no matter if it stops being blurred, fullscreen/moved off-screen, some fancy noblur rule doing things.

something like CWaterBlurProvider::pruneStates() does.

src/render/GLRenderer.cpp
```diff
@@ -304,9 +316,95 @@ void CHyprGLRenderer::drawGlow(const CBox& box, int round, float roundingPower,
...
+    if (m_blur && m_blur->type() == type)
+        return;
+
+    m_blur = createBlurProvider(type, *g_pHyprOpenGL);
+}
+
+void CHyprGLRenderer::expandBlurDamage(CRegion& damage, float multiplier) const {
+    RASSERT(m_blur, "Cannot expand blur damage without a blur provider");
+    m_blur->expandDamage(damage, multiplier);
+}
```
gulafaran, line 342: shouldnt this clamp like old function did?

````
float CRenderPass::oneBlurRadius() {
    // TODO: is this exact range correct?
    static auto PBLURSIZE   = CConfigValue<Config::INTEGER>("decoration:blur:size");
    static auto PBLURPASSES = CConfigValue<Config::INTEGER>("decoration:blur:passes");

    const auto  BLUR_PASSES = std::clamp(*PBLURPASSES, sc<int64_t>(1), sc<int64_t>(8));

    return std::clamp(*PBLURSIZE, sc<int64_t>(1), sc<int64_t>(40)) * pow(2, BLUR_PASSES); // is this 2^pass? I don't know but it works... I think.
}
```

vaxerski:
@wabarTM ripple intended. Jar locked as per request. I also think it's better.

Rest of the issues fixed

vaxerski:
added prism, heat_shimmer and acrylic

vaxerski:
added aurora

vaxerski:
added haze

ryzendew:
image
acrylic doesn't look very good at all nothing like liquid glass maybe these can help https://github.com/4v3ngR/kwin-effects-glass/tree/main/src/shaders

vaxerski:
there are knobs you can twist. The defaults are a bit meh

ryzendew:
> there are knobs you can twist. The defaults are a bit meh

i tried em all nothing looks very good.

vaxerski:
picky ass mofo fine

ryzendew:
> picky ass mofo fine
image
with a bit of coding i got to here! it's not the best but could be used as a reference! https://github.com/Event-Horizon-desktop-environment/Hyprland/tree/more-blur-XDDDDDDDDDDDDDDDDDDDDDD i love the fact you added more blur! thank you!!

Dregu:
> the piss jar is sideways on rotated monitor

All the other animated things are also sideways.

> If you shake the piss jar enough it collapses into a singularity

This still happens, as in the amount of water seemingly reduces until it's all combined to a single small blob.

vaxerski:
image

this should be better.

My settings here:

```
        blur = {
            enabled   = true,
            size      = 5,
            passes    = 3,
            vibrancy  = 0.1696,
	   popups    = true,
	   variant   = "acrylic",
	   acrylic = { refraction = 48, bulb = 20, clarity = 0.5, tint = 0x0aDDEEFF, aberration = 0.2 },
        },
```

vaxerski:
any comments?

aurelleb:
Default `acyrlic` looks almost transparent, it is a bit better with @vaxerski config above but it doesn't look very glassy. I get that the goal is not to replicate liquid glass fully ([it's very overengineered anyways](https://developer.apple.com/videos/play/wwdc2025/219/)) but I think it feels kinda wrong, especially the default config.

Also if the goal is to have a glassy look we should probably not call it acyrlic I guess

Also worth noting that liquid glass on macOS behaves quite differently depending on whether the app declares  light or dark mode.



macOS liquid ass (0.6 opacity)

image

Hyprland liquid ass with @vaxerski config above (0.9 opacity, 0.6 is not legible)

image

Lowering `clarity` helps, but it just looks like kawase blur

@ryzendew version looks better

vaxerski:
I cant just copy his version cuz its gpl

aurelleb:
just call it inspiration

ryzendew:
> I cant just copy his version cuz its gpl

you 100% can take it lol enjoy

vaxerski:
it aint yours either lolmao

vaxerski:
I think we can merge this as is and worry about polishing it later, since most other effects are fine imo

fxzzi:
im using regular kawase blur and my blur effects keep randomly disappearing on foot, might be ext background effect related

aurelleb:
following changes in this PR or unrelated @fxzzi ?

fxzzi:
After this merged

vaxerski:
repro case please

fxzzi:
if i had repro case i would share but its just random, something to do with damage i think because trying to screenshot it immediately corrects it before i have a chance to capture it

just have some foot windows open with blur on in its config

```ini
[colors-dark]
alpha=0.850000
blur=yes
```

and just use your computer and you'll notice them randomly losing blur and then coming back at points

fxzzi:
https://github.com/user-attachments/assets/6aa8d6a5-7092-4e5f-b63b-2fc50371b763

was finally able to catch it on video you can see here the blur just randomly disappearing then coming back

vaxerski:
the more I look at foot the more I think its just a terminal like wezterm, made to annoy people

vaxerski:
I can't repro in a nest. anime boobs remain blurry. Can you bisect this MR?

vaxerski:
@fxzzi check the mr

Aqa-Ib:
xray is broken for acrylic.

vaxerski:
?

Aqa-Ib:
yeah, also broken for fluid_jar. For the rest of variants it works.

vaxerski:
both need a geometry. Xray doesnt have one. This is intended

kelian-4:
hey 
I read through the discussion. You might want to take a look at this repository for Acrylic — it could be helpful: https://github.com/OverShifted/LiquidGlass

PR #15665, opened 2026-07-31
input-capture: implement the session destroy request

tyvsmith:
This implements a destroy lifecycle on the compositor's InputCapture sessions.

Part 3 of 3 needed for hyprwm/xdg-desktop-portal-hyprland#419, leaking file descriptors
from InputCapture.
Needs hyprland-protocols ≥ 0.8.0 (<protocols PR link>).

Not visible in the diff:

- `active` is a strong `SP<>`, so the erase never dropped the last reference and
  `~CInputCaptureResource` couldn't run. Releasing first also fixes the pre-existing
  disconnect-while-capturing case.
- Doesn't fix the residual `+1.00/session`. Not the EIS client connection. It leaks the
  same with `ConnectToEIS` never called. Suspect is `eis_backend_fd_add_client()`'s fd
  never being closed after `sendEisFd()`, but ownership after send isn't clear to me.
  Can file separately.

Second commit makes `protocolnew()` depend on the XML it generates from; without it a
stale build tree keeps the old glue after a protocols bump and fails to build. Happy to
split out.

### Validation

`hyprland_gtests` 271/271. Built clean and installed all 3 PRs in an Arch VM as well as the latest main/master and then verified the leaking flow from hyprwm/xdg-desktop-portal-hyprland#419.

| | before | after |
|---|---|---|
| `eis-*.lock` | one per session, never freed | stays at 0 |
| compositor fds | +6/session | +1/session |
| teardown traffic | `disable` only | `destroy` only |
| protocol errors | none | none |

Object IDs get reused afterwards, so the resources are freed.

### Disclosure

Agent implemented and verified in VM, with oversight and review from me before publishing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -503,6 +503,7 @@ function(protocolnew protoPath protoName external)
            ${CMAKE_SOURCE_DIR}/protocols/${protoName}.hpp
     COMMAND hyprwayland-scanner ${path}/${protoName}.xml
             ${CMAKE_SOURCE_DIR}/protocols/
+    DEPENDS ${path}/${protoName}.xml
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
   target_sources(hyprland_lib PRIVATE protocols/${protoName}.cpp
                                   protocols/${protoName}.hpp)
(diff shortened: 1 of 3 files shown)


tyvsmith:
Pending [Vouch request](https://github.com/hyprwm/.github/discussions/9)

PR #15706, opened 2026-08-02
popup: make popup snapshot render their own size

gulafaran:
instead of fullscreen snapshots and essentially full monitor blits on dual gpu setups, make snapshots snapshot their own size, and damage their own size. but this means we have to consider transformed monitors.

add a Math::transformedSize and replace all "PMONITOR->m_transform % 2 == 0 ?" uses with it.

### DIFF
--- a/src/desktop/state/PopupFadeout.cpp
+++ b/src/desktop/state/PopupFadeout.cpp
@@ -21,15 +21,10 @@ static bool shouldBlurPopup() {
     return *PBLURPOPUPS && *PBLUR;
 }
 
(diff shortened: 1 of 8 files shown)


vaxerski:
The popup is allowed to extend, morph, etc, especially with transformers. I don't think this is a good idea.

gulafaran:
> The popup is allowed to extend, morph, etc, especially with transformers. I don't think this is a good idea.

Might require some bigger refactor in the future then, its currently very noticeable when i measure the rendertimes on dgpu blitting, dragging mouse over like 5+ tabs, almost 2 vblanks off heh.

PR #15709, opened 2026-08-02
renderer/wobble: fix artifacts on transformed

vaxerski:
Further fixes for wobble on transformed.


dim-ghub:
you missed one, if the window is wider than the display, the edges on both sides get sliced and extended

https://github.com/user-attachments/assets/2ea4c523-2679-493c-8c95-bb930926ddb1

(also noticed that xray stops... xraying while windows are being moved around, then snaps back into existence after sitting still)

PR #15714, opened 2026-08-02
render: always render in normal transform

vaxerski:
fixes a lot of effects needing to know about the output transform, screencopy, etc.

This needs to be tested. Also worth testing screencopy.

### DIFF
--- a/src/errorOverlay/Overlay.cpp
+++ b/src/errorOverlay/Overlay.cpp
@@ -108,7 +108,7 @@ void COverlay::createQueued() {
         return;
 
     const float       SCALE    = PMONITOR->m_scale;
-    const int         FONTSIZE = std::clamp(sc<int>(10.f * ((PMONITOR->m_pixelSize.x * SCALE) / 1920.f)), 8, 40);
+    const int         FONTSIZE = std::clamp(sc<int>(10.f * ((PMONITOR->m_transformedSize.x * SCALE) / 1920.f)), 8, 40);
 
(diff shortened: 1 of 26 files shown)


Dregu:
A lot of smart screen capture tools are broken, they still think the frames from a rotated monitor are also rotated, even though they are not. Some protocol is feeding unnecessary transform data, I don't know what. This includes wf-recorder, wl-mirror, obs through xdph...

Xdph is still using wlr-screencopy, which doesn't even support transform metadata, so xdph fetches the transform by other means and gets it wrong. I don't know how the others work.

Dumber tools like ye olde obs-wlroots-screencopy, browsers, discord that don't even try to rotate anything work better though since the actual image *is* always right way up now.

vaxerski:
xdph fix linked.

vaxerski:
I checked wf-recorder's source. It seems it applies the transform because it feels like it.

Problem with screencopy is that the old protocol does not clarify what's the expected behavior here, so it's kinda an FFA.

As for anything using ext_ICC and ext_ICS they will get NORMAL and if they ignore it, it's on them.

Dregu:
Oh yeah I forgot wf-recorder isn't even ported to icc yet, it's a pr I'm running but even in that the transform event is just a noop. Same with wl-mirror, they just log it and actually get the wrong transform from wl_output probably. Pity, but you're right, it's all on them. And screencopy can just go and die.

vaxerski:
any other issues?

neonvoidx:
this also broke hyprpicker

PR #15718, opened 2026-08-02
buffer: dont drop damage until actually read

gulafaran:
we cant drop dmabuf damage until we have actually hit a frame that reads it, otherwise clients like firefox attaches a buffer with damage and commits, eventually a bit later sends a .frame() and .commit() then we clear the damage and the render has nothing to use when pageflip hits

i tried checking the acummulatedamage paths, and it should work as is, but you never know. we also restore the the damage if direct scanout fails to commit and returns to normal rendering.

### DIFF
--- a/src/desktop/view/WLSurface.cpp
+++ b/src/desktop/view/WLSurface.cpp
@@ -85,51 +85,51 @@ CRegion CWLSurface::computeDamage(const std::optional<CBox>& box) const {
     if (!m_resource->m_current.texture)
         return {};
 
-    CRegion damage = m_resource->m_current.accumulateBufferDamage();
-    damage.transform(Math::wlTransformToHyprutils(m_resource->m_current.transform), m_resource->m_current.bufferSize.x, m_resource->m_current.bufferSize.y);
-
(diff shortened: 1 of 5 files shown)


vaxerski:
what about apps on two displays at once

gulafaran:
> what about apps on two displays at once

no difference, computeDamage is only called via damageSurface.

```
    CRegion damageBox = WLSURF->computeDamage(SURFACE_BOX);
    if (damageBox.empty())
        return;

    if (scale != 1.0)
        damageBox.scale(scale);

    damageBox.translate({x, y});

    const auto EXTENTS = damageBox.getExtents();

    CRegion    damageBoxForEach;

    for (auto const& m : State::monitorState()->monitors()) {
        if (!m->m_output)
            continue;
        if (!EXTENTS.overlaps(m->logicalBox()))
            continue;

        damageBoxForEach.set(damageBox);
        damageBoxForEach.translate({-m->m_position.x, -m->m_position.y}).scale(m->m_scale);

        m->addDamage(damageBoxForEach);
    }
```

vaxerski (approved):

PR #15721, opened 2026-08-02
protocols: relax vicinae-hotkey-v1 binding policy

aurelleb:

After [further discussion](https://github.com/niri-wm/niri/pull/4145#issuecomment-5151787448) it appears that the default binding policy we apply is too restrictive:

- There are use cases in which many applications could want to bind the same key. For instance, to use the same shortcut to trigger a microphone mute across all running voip applications. The single binding policy prevents this which is annoying and probably over restricting. To fix this, we simply disable this rule.
- Bare numpad keys and other non printable keys such as `Insert`, `End`, `Page Up/Down`  can't be bound without a modifier, which often goes against what users want. We now allow these, and only require a modifier for "text" keys. 

In order to keep the user in control, it is important that they are able to audit what application is reserving what shortcut in order to debug any potential problem. To that end, `hyprctl globalshortcuts` now prints out all reserved shortcuts, from the portal and from the `vicinae-hotkey-v1` reservations.

This is the output of `hyprctl globalshortcuts -j` on my system, where OBS and vicinae are reserving shortcuts:

```json
[{
    "name": "vicinae:ALT+SHIFT+c",
    "description": "clipboard:history",
    "appid": "vicinae",
    "trigger": "ALT+SHIFT+c",
    "bound": true
},
{
    "name": "vicinae:SUPER+space",
    "description": "Toggle Vicinae",
    "appid": "vicinae",
    "trigger": "SUPER+space",
    "bound": true
},
{
    "name": "com.obsproject.Studio:CTRL+KP_Home",
    "description": "OBS Studio hotkey",
    "appid": "com.obsproject.Studio",
    "trigger": "CTRL+KP_Home",
    "bound": true
}]
```

The changes are only compositor policy, the protocol itself is unaffected

ready

### DIFF
--- a/protocols/vicinae-hotkey-v1.xml
+++ b/protocols/vicinae-hotkey-v1.xml
@@ -59,11 +59,20 @@
 
     A global hotkey lets an unfocused client observe that a specific key
     combination was pressed. Compositors are the security boundary and SHOULD
-    apply policy when deciding whether to accept a request. In particular,
-    implementations SHOULD reject combinations that do not include at least one
-    non-latching modifier among Ctrl, Alt or Super, unless the trigger is a
-    function key, because unmodified (and Shift-only) keys carry ordinary text
-    entry and grabbing them turns this protocol into a keylogger.
+    apply policy when deciding whether to accept a request. As a default, we
+    suggest the following pseudo-code:
+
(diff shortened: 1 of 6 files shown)


aurelleb:
image

PR #15722, opened 2026-08-02
Lua: display dispatcher names in binds list and when stringifying

LRitzdorf:

At the moment, `hyprctl binds` just lists `dispatcher: __lua` for every bind. Users can still create their own bind *descriptions* in the config, but that requires effort, and is annoying when the old Hyprlang binds would just tell you what their dispatchers were.

This PR updates the `__tostring` metamethod for dispatchers (making them stringify to `HL.Dispatcher(<name>)`), and uses that result when printing them in `hyprctl binds`.

Some example output, for reference:
```
bind
	modmask: 8
	submap: clean
	key: Q
	keycode: 0
	catchall: false
	description: 
	dispatcher: HL.Dispatcher(submap)
	arg: 65

bind
	modmask: 9
	submap: 
	key: E
	keycode: 0
	catchall: false
	description: 
	dispatcher: HL.Dispatcher(exit)
	arg: 70

bind
	modmask: 8
	submap: 
	key: A
	keycode: 0
	catchall: false
	description: 
	dispatcher: function: 0x5648de2c0c60
	arg: 71
```

A few details I wish were more doable:
- Printing full dispatcher names (e.g. `hl.dsp.exit`), but the dispatcher object that `hl.bind()` receives only knows the last part of its name (`exit`), so I don't think this is reasonably possible
- Dispatcher args would also be nice to show, but unless we just want to print `table: 0xDEADBEEF` or whatever, that requires fancy table-traversal logic and probably isn't worth the effort

Nothing major to note. This is really just a string formatting change, so the test isn't super thorough; let me know if more are desired.

Ready, except that this should wait until #15568 merges

### DIFF
--- a/hyprtester/src/tests/main/keybinds.cpp
+++ b/hyprtester/src/tests/main/keybinds.cpp
@@ -830,3 +830,17 @@ TEST_CASE(keybinds) {
     CALL_SUBTEST(perDeviceKeybind);
     CALL_SUBTEST(unbind);
 }
+
+TEST_CASE(luaDispatcherStrings) {
+    OK(getFromSocket("/eval B = hl.bind('SUPER + F24', hl.dsp.exec_cmd('true'))"));
+    EXPECT(getFromSocket("/repl return B.handler"), "HL.Dispatcher(exec_cmd)");
+    OK(getFromSocket("/eval B:unbind()"));
+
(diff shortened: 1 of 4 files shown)


LRitzdorf:
Oop, should hold this until #15568 gets merged

vaxerski (approved):

PR #15729, opened 2026-08-03
fix: popup event handling with keyboard exclusive layers

aurelleb:

Native wayland popups attached to layers with keyboard interactivity set to exclusive didn't receive any mouse event, the layer was eating it all.

In practice this resulted in many popups outright closing on click, since very often, clicking outside of the popup makes the application close the popup.

ready


vaxerski (approved):
tnx

PR #15737, opened 2026-08-04
protocols: ext-background-effect for native popups/subsurfaces

aurelleb:

I forgot to add native popup support when I implemented ext-background-effect. This fixes it

Obviously that's only applicable to native popups, Electron apps which emulate popups can't benefit from this.

ready

### DIFF
--- a/src/desktop/state/PopupFadeout.cpp
+++ b/src/desktop/state/PopupFadeout.cpp
@@ -38,7 +38,7 @@ SP<CPopupFadeout> CPopupFadeout::create(SP<CPopup> popup, SP<Render::IFramebuffe
     fadeout->m_framebuffer = snapshot;
 
     static CConfigValue PBLURIGNOREA = CConfigValue<Config::FLOAT>("decoration:blur:popups_ignorealpha");
-    if (g_pHyprRenderer->shouldBlur(popup)) {
+    if (popup->shouldBlur()) {
         fadeout->m_effects.textureBlur.enabled               = true;
         fadeout->m_effects.textureBlur.blockBlurOptimization = true;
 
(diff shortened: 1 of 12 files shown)


src/desktop/state/PopupFadeout.cpp
```diff
@@ -45,7 +38,7 @@ SP<CPopupFadeout> CPopupFadeout::create(SP<CPopup> popup, SP<Render::IFramebuffe
     fadeout->m_framebuffer = snapshot;
 
     static CConfigValue PBLURIGNOREA = CConfigValue<Config::FLOAT>("decoration:blur:popups_ignorealpha");
-    if (shouldBlurPopup()) {
+    if (g_pHyprRenderer->shouldBlur(popup)) {
```
vaxerski: please mimic the approach in https://github.com/hyprwm/Hyprland/pull/15661 for this

vaxerski:
is it impl'd for subsurfaces? Maybe in general we should add glue for CWLSurface for this so we can have it be generic?

aurelleb:
I shamefully admit I just found out about the concept of subsurface, and it looks like it may solve some of my own problems
And no they are not implemented, I'm gonna do that, I agree that this should be generalized since they are all wl_surface in the end

aurelleb:
ok I think it's in a pretty good state now, both popup and subsurfaces are supported and I did a bit of refactoring.

There is maybe one case that is not handled: subsurfaces inside a popup. Idk if it's worth implementing given how rare this is going to be in practice, maybe we can do it later if we find it useful?

vaxerski:
subsurfaces in a popup would be quite the cursed setup but I'm sure it will happen at some point lmao

vaxerski (commented):
hold off with this for https://github.com/hyprwm/Hyprland/pull/15779 I think we could do some good things here with the window refactors for blurpolicy stuff

PR #15744, opened 2026-08-04
committiming: release for intersected monitors

gulafaran:
some clients never binds an output, so check for intersection like fifo does aswell.


vaxerski (approved):
👍

PR #15745, opened 2026-08-04
remove duplicate call to fullscreen function

Aqa-Ib:
The same call is done just before, here: https://github.com/hyprwm/Hyprland/blob/5c6f0aa2df6d1698ee7174c69d3123cb236c6fde/src/desktop/view/Group.cpp#L156
The code in question was added here https://github.com/hyprwm/Hyprland/commit/abffe75088e2d776e14e5dbd726a835fa157df9a

No.

Everything seems good.


erstarr (approved):
This should already be done by `CGroup` so this should be fine. Thanks.

vaxerski (approved):

PR #15746, opened 2026-08-04
layout/monocle/fullscreen: Fix alpha and input being incorrectly set when a Maximised window is moved from a normal workspace w/o monocle to a special workspace with monocle

erstarr:

- Title.

- also renames `moveToWorkspace()` -> `moveWindowToWorkspace()` to be less confusing (moving to workspace may also mean you are switching to it, though that has its own function so it's confusing)

<details> <summary> Repo steps: </summary>

Normal Workspace: Dwindle
Special Workspace: Monocle

Already have a window in the special workspace

Maximise a window in the normal workspace, move it to special with `on_focus_under_fullscreen = 1` and `follow = true`

It'll be invisible.

</details>

- If other layouts are to also use custom input blocking reasons and `WINDOW_ALPHA_LAYOUT`, the same exception also has to be made for them

Otherwise nothing

- [ ] Test

### DIFF
--- a/src/config/lua/bindings/LuaBindingsDispatchers.cpp
+++ b/src/config/lua/bindings/LuaBindingsDispatchers.cpp
@@ -483,7 +483,7 @@ static int dsp_moveToWorkspace(lua_State* L) {
         return Internal::dispatcherError(L, "Invalid workspace", ERR, C_INVARG);
 
     bool silent = lua_toboolean(L, lua_upvalueindex(2));
-    return Internal::checkResult(L, CA::moveToWorkspace(ws, silent, Internal::windowFromUpval(L, 3)));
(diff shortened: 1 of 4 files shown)


erstarr:
While I'm at it, I'll codify how moving windows to other workspaces while either or both workspaces have FS windows is to behave in all layouts.

To this end, I'll some MRs, all based on one another linearly (ideally only this and one other but it depends if/when i find erroneous behaviour).

I'll also add comprehensive testing for all this for all layouts tested independently (since this MR fixes a bug that only happened on monocle, there is a need for this), which will be its own (huge) MR, based on the last MR that fixes the move behaviour.

I'll provide the merging order for all this after they're all done.

PR #15751, opened 2026-08-05
monitor: fix UB reading Config::INTEGER from bool

willmtemple:

`misc:screencopy_force_8b` is declared `MS<Bool>` and then read as `Config::INTEGER`, which is UB at least two ways.

I think the option is inert as is, since `true` is default. THe option is likely always surrounded by nonzero bytes making the truthiness of the option guaranteed-ish when it reads 8 bytes from a pointer to bool. Most captures probably break if screencopies aren't 8 bit.

ready, 1-liner

### DIFF
--- a/src/output/Monitor.cpp
+++ b/src/output/Monitor.cpp
@@ -2495,7 +2495,7 @@ NColorManagement::SImageDescription::SPCMasteringLuminances CMonitor::getMasteri
 }
 
 uint32_t CMonitor::getPreferredReadFormat() {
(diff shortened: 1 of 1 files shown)


vaxerski:
this is not UB? 

image


Besides it's used across the codebase like this

willmtemple:
Apologies, getting used to the code and I'm not extremely skilled with C++, I didn't notice the template specialization and just noticed the type mismatch while I was trying to debug some issues with screencopies.

vaxerski:
nw

PR #15755, opened 2026-08-05
frame time scheduling

gulafaran:
the start of frame time scheduling, removes the old triple buffer logic that is sound in theory but we dont have the code to do it, yet. so currently its broken anyways.

lets make new_render_scheduling do something useful instead.

we also filter out render jitter from software clock garbage drivers.


vaxerski:
if you are removing triple buffering, new render scheduling should not just now do something else. Make it deprecated and a no-op and either make this the default or hide it under another config opt. Honestly if you want actual testing (and believe in this) thne I'd make it on by default

gulafaran:
> if you are removing triple buffering, new render scheduling should not just now do something else. Make it deprecated and a no-op and either make this the default or hide it under another config opt. Honestly if you want actual testing (and believe in this) thne I'd make it on by default

yes sir, yeah i mostly added it as draft already because some users wanted to test it already but im not satisfied yet. are a few things i want to rework first regarding sampling the rendering etc, and what to actually do on dgpu blits that changes the infence. so not calling for broadwide testing, yet :p

PR #15759, opened 2026-08-05
renderer/transformer: make window transformers adapt to viewport constraints

vaxerski:
ref https://github.com/hyprwm/Hyprland/discussions/15734

Transformers used to run in monitor space, which means windows sized above that would get their textures clipped.


vaxerski:
@codex review please

PR #15761, opened 2026-08-05
cm: re-send HDR metadata when its content changes, not only on a transition

willmtemple:

If the monitor wants to be in HDR, compares the HDR output metadata we would send with what aquamarine already has and set the update if there is any change. This is a more generalized rule for sending HDR metadata in addition to the changes that were merged yesterday (@gulafaran).

The issue is that changes to the HDR metadata in config do not propagate until a transition from sdr -> hdr, or something else forces resending the HDR metadata. The bug particularly affects changes to min/max/max_avg luminance which don't update if you change them in the config while HDR is active until you turn HDR off and back on again.

The implementation depends on a couple of invariants not changing:
- createHDRMetadata's result is deterministic for the image description.
- Aquamarine always holds hdrMetdata byte-identical to what we gave it.

Both of these invariants are trivially true I think, but if they ever become falsified we will see spurious commit churn.

ready pending feedback

CC @UjinT34

### DIFF
--- a/src/render/Renderer.cpp
+++ b/src/render/Renderer.cpp
@@ -4,6 +4,8 @@
 #include <algorithm>
 #include <aquamarine/output/Output.hpp>
 #include <cmath>
+#include <cstring>
+#include <drm_mode.h>
 #include <filesystem>
 #include "../config/ConfigValue.hpp"
 #include "../config/ConfigManager.hpp"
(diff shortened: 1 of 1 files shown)


umbrageodotus:
LGTM

vaxerski (approved):
jesus fuck I had to look up that static assert check

PR #15763, opened 2026-08-05
desktop/view: fix special workspace hiding on window unmap fallback

Meehir5907:
#### What does it fix/add?
This fixes the bug where an active special workspace hides when you close a pinned floating window (like a PiP video) inside it.
 
in depth explanation at: https://github.com/hyprwm/Hyprland/discussions/15691

 (this part was updated after first review)
 Refactored to use native refocusing. If a special workspace is active and a fallback candidate from a different workspace is selected, we now explicitly reject it (`candidate = nullptr`). This automatically triggers `g_pInputManager->refocus()` to find the correct window, preventing the special workspace from incorrectly hiding.
##### Pre-Patch

https://github.com/user-attachments/assets/3b576381-fb1c-466c-8dad-c7cfa04c69c5

##### Post-Patch

https://github.com/user-attachments/assets/d228c1e9-18ed-431b-842e-ecf6eb4bb84c

Ready for merging.

### DIFF
--- a/src/desktop/view/Window.cpp
+++ b/src/desktop/view/Window.cpp
@@ -2692,6 +2692,9 @@ void CWindow::unmapWindow() {
             }
         }
 
+        if (candidate && PMONITOR && PMONITOR->m_activeSpecialWorkspace && candidate->m_workspace != PMONITOR->m_activeSpecialWorkspace)
+            candidate = nullptr;
+
         Log::logger->log(Log::DEBUG, "On closed window, new focused candidate is {}", candidate);
 
         if (candidate != Desktop::focusState()->window() && candidate) {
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
I think this is kind of like trying to solve homelessness in seattle by shitting into a homeless guy's donation can.

Why not just reject candidates on a different workspace? Soudns way more reasonable to me?

Meehir5907:
I guess now it's properly handled.

src/desktop/view/Window.cpp
```diff
@@ -2679,6 +2692,13 @@ void CWindow::unmapWindow() {
             }
         }
 
+        const auto PMONITOR = m_monitor.lock();
+        if (PMONITOR && PMONITOR->m_activeSpecialWorkspace) {
+            if (candidate && candidate->m_workspace != PMONITOR->m_activeSpecialWorkspace && !candidate->m_pinned) {
```
vaxerski: why the pinned check? Shouldn't pinned windows automatically update their workspace on a switch? Or is that bugged?

Meehir5907: I added it because pinned windows keep the underlying workspace id, so the filter rejected them. I tested without the check and `refocus()` handles it natively anyway. So now I stripped it out and also flattened the logic.

vaxerski: ah, right, they do that.

vaxerski (approved):

vaxerski:
please add a test though

PR #15765, opened 2026-08-06
seat: don't block pointer focus for when no mouse is attached

navidmafi:
Stylus and touchscreen are libinput devices who don't register as IPointer. On tablets a typecover or some external pointer would be the only device in the system that could expose an IPointer. This makes it so that detaching TypeCover empties m_pointers, so setMouse(nullptr) runs and m_mouse becomes null. From that point, every setPointerFocus() call initiated from touch or tablet-tool proximity is no-op'ed by the guard.
This causes proximity_in, motion, pressure, and down events to not be delivered to clients. This PR removes the guard.

Shouldn't cause regressions or unintended side-effects. 

326/326 gtests passed

### DIFF
--- a/src/managers/SeatManager.cpp
+++ b/src/managers/SeatManager.cpp
@@ -301,11 +301,6 @@ void CSeatManager::setPointerFocus(SP<CWLSurfaceResource> surf, const Vector2D& 
         return;
     }
 
-    if (!m_mouse) {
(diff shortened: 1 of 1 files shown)


umbrageodotus:
@navidmafi plz apply for vouch :3

navidmafi:
https://github.com/hyprwm/.github/discussions/19 :D

PR #15767, opened 2026-08-06
hyprctl: random Lua-adjacent fixes

LRitzdorf:
Two separate things:
- Fixes `hyprctl repl`'s interactive-mode detection, so it triggers properly even when a specific instance is selected (`-i 1` or whatever)
- Updates the S1 socket's batch-command parsing, to require escaping for all semicolons inside of batched commands
  - This also eliminates some square-bracket matching that was intended to avoid needing escapes, but just makes things worse now that we have Lua and the user might need to include arbitrary strings

Some IPC batch examples, before and after this change:
```diff
-[[BATCH]][float;pin] kitty;activewindow
+[[BATCH]][float\;pin] kitty;activewindow

-[[BATCH]]repl local v = "asdf"; print(v);activewindow  // would over-split into 3 commands
+[[BATCH]]repl local v = "asdf"\; print(v);activewindow

-[[BATCH]]repl print("DEBUG]");activewindow  // wouldn't be split into 2 commands
+[[BATCH]]repl print("DEBUG]");activewindow
```

This changes the IPC format *specifically for batch execution*, so minor things relying on that might break. The main change is for the old-style `[rule;rule] command` exec-with-rules syntax; the semicolons between rules in there will now have to be escaped as shown above. Escaping is also needed for any backslashes (e.g. in Lua code that might `print("\n")`).

Probably ready? The only other change I thought about that'd be neat is modifying `hyprctl --batch` to treat each arg as a separate command, and then having it do the escaping and semicolon-joining *for* the user. The logic for that is kinda very incompatible with how hyprctl currently parses arguments, though, so I haven't done it and probably won't unless we decide it's important.

### DIFF
--- a/src/ipc/s1/S1.cpp
+++ b/src/ipc/s1/S1.cpp
@@ -142,9 +142,9 @@ SResponse CSocket1::dispatchBatch(std::string request, pid_t pid) {
         }
 
         if (request[i] == '\\') {
-            if (i < request.size() && (request[i + 1] == '\\' || request[i + 1] == ';')) {
+            if (i < request.size() && (request[i + 1] == '\\' || request[i + 1] == ';'))
                 ++i;
-            } else
+            else
                 Log::logger->log(Log::ERR, "Malformed socket1 request: invalid escape sequence {} at position {}, using it verbatim", request.subview(i, 2), i);
         }
         parsedCommand << request[i];
--- a/src/ipc/s1/S1.cpp
+++ b/src/ipc/s1/S1.cpp
@@ -7,6 +7,7 @@
 #include <algorithm>
 #include <optional>
 #include <ranges>
+#include <sstream>
 #include <hyprutils/string/String.hpp>
 
 using namespace IPC::Socket1;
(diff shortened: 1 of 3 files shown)


src/ipc/s1/S1.cpp
```diff
@@ -129,25 +130,24 @@ SResponse CSocket1::dispatchBatch(std::string request, pid_t pid) {
...
+            parsedCommand.str("");
+            parsedCommand.clear();
             continue;
+        }
 
-        if (commandStart < i)
-            commands.emplace_back(Hyprutils::String::trim(request.substr(commandStart, i - commandStart)));
-        commandStart = i + 1;
+        if (request[i] == '\\') {
+            if (i < request.size() && (request[i + 1] == '\\' || request[i + 1] == ';')) {
```
vaxerski: u me when the bruh no {} rule XD

LRitzdorf: I am an idiot sandwich

vaxerski (approved):
egg mcmuffin

PR #15776, opened 2026-08-06
Session: add disable-able systemd session target

LRitzdorf:

Back in June, xdg-desktop-portal broke stuff for any environments that didn't start `graphical-session.target` ([relevant discussion](https://github.com/hyprwm/Hyprland/discussions/15072)). Distros have applied some workarounds since then (notably, Arch just undoes the change that made x-d-p require graphical-session), but the actual proper fix is still on our end. Rather than making every systemd user (which is most of them) go find the wiki section on starting graphical-session, we might as well automate those steps ourselves.

Note that any non-systemd users can easily disable the logic this PR adds, by setting `HYPRLAND_NO_SD_TARGET`. Also respects the `NO_SYSTEMD` build flag if that's set.

Also just fyi, the original branch for this PR isn't actually mine, but @Dregu's. He didn't want to PR it, so I'll take the blame if this goes horribly wrong or whatever.

Prior discussion of this change: https://github.com/hyprwm/Hyprland/discussions/15072#discussioncomment-17345023.

Not really. This pretty much just automates the [steps currently on the wiki](https://wiki.hypr.land/Useful-Utilities/Systemd-start/#hyprland-sessiontarget).

Should be good. Also, @tekstryder's [been using this patch](https://github.com/hyprwm/Hyprland/discussions/15072#discussioncomment-17352822) for a few months now, so maybe they have more feedback?

### DIFF
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -543,7 +543,7 @@ void CCompositor::cleanEnvironment() {
 
     if (m_aqBackend->hasSession() && !Env::envEnabled("HYPRLAND_NO_SD_VARS") && !getenv("MANAGERPID")) {
 #ifdef USES_SYSTEMD
-        if (!Env::envEnabled("HYPRLAND_NO_SD_TARGET"))
+        if (m_sdSessionTarget)
             // stopping hyprland-session doesn't wait for dependent services; this does
             Config::Supplementary::executor()->spawn("systemctl --user stop graphical-session.target");
 #endif
@@ -804,8 +804,10 @@ void CCompositor::startCompositor() {
             "dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP HYPRLAND_INSTANCE_SIGNATURE QT_QPA_PLATFORMTHEME PATH XDG_DATA_DIRS";
         Config::Supplementary::executor()->spawn(CMD);
 #ifdef USES_SYSTEMD
-        if (!Env::envEnabled("HYPRLAND_NO_SD_TARGET"))
+        if (!Env::envEnabled("HYPRLAND_NO_SD_TARGET")) {
+            m_sdSessionTarget = true;
             Config::Supplementary::executor()->spawn("systemctl --user start hyprland-session.target");
+        }
 #endif
     }
 
--- a/src/Compositor.cpp
+++ b/src/Compositor.cpp
@@ -541,7 +541,7 @@ void CCompositor::cleanEnvironment() {
     if (m_desktopEnvSet)
         unsetenv("XDG_CURRENT_DESKTOP");
 
-    if (m_aqBackend->hasSession() && !Env::envEnabled("HYPRLAND_NO_SD_VARS")) {
+    if (m_aqBackend->hasSession() && !Env::envEnabled("HYPRLAND_NO_SD_VARS") && !getenv("MANAGERPID")) {
 #ifdef USES_SYSTEMD
         if (!Env::envEnabled("HYPRLAND_NO_SD_TARGET"))
             // stopping hyprland-session doesn't wait for dependent services; this does
@@ -793,7 +793,9 @@ void CCompositor::startCompositor() {
         /* Session-less Hyprland usually means a nest, don't update the env in that case */
         m_aqBackend->hasSession() &&
         /* Activation environment management is not disabled */
-        !Env::envEnabled("HYPRLAND_NO_SD_VARS")) {
+        !Env::envEnabled("HYPRLAND_NO_SD_VARS") &&
+        /* Being executed under systemd */
+        !getenv("MANAGERPID")) {
         const auto CMD =
 #ifdef USES_SYSTEMD
             "systemctl --user import-environment DISPLAY WAYLAND_DISPLAY HYPRLAND_INSTANCE_SIGNATURE XDG_CURRENT_DESKTOP QT_QPA_PLATFORMTHEME PATH XDG_DATA_DIRS && hash "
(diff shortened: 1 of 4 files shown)


tekstryder:
>Also, @tekstryder's https://github.com/hyprwm/Hyprland/discussions/15072#discussioncomment-17352822 for a few months now

I did eventually stop using this. However, it does feel correct that any compositor creating a graphical session should be doing this (guarded behind an env/build toggle).

Rare for Arch to go against upstream (`x-d-p` in this case), but as a result, the Sway packaging patch accomplishing this same thing was [also dropped](https://gitlab.archlinux.org/archlinux/packaging/packages/sway/-/commit/5d3b06eb71cfba1f46390008cfc7828d726e356a) in favor of the [project's wiki](https://github.com/swaywm/sway/wiki/systemd-integration) for users who want it.

Thanks for following up regardless of the direction this project takes.

LRitzdorf:
Actions [died today](https://www.githubstatus.com/incidents/qcvjkzcs7j74), in case a maintainer wants to retrigger workflows once status is green again

fufexan:
Have you run hyprland with these changes?

I'm unsure how it plays with uwsm.

LRitzdorf:
Just tested, and it does explicitly start `hyprland-session.target` even under UWSM. That doesn't seem to break anything, but I'm not familiar enough with UWSM to know if it's actually a problem.

UWSM does start `graphical-session.target` on its own, even without this patch, so it's certainly not *necessary*. Are there UWSM-specific env vars or something we can use to conditionally not do this?

src/Compositor.cpp
```diff
@@ -576,6 +560,10 @@ void CCompositor::cleanup() {
 #ifdef USES_SYSTEMD
     if (NSystemd::sdBooted() > 0 && !Env::envEnabled("HYPRLAND_NO_SD_NOTIFY"))
         NSystemd::sdNotify(0, "STOPPING=1");
+    if (m_aqBackend->hasSession() && !Env::envEnabled("HYPRLAND_NO_SD_TARGET")) {
+        Config::Supplementary::executor()->spawn("systemctl --user stop hyprland-session.target");
+        usleep(50000); // services blow up if there's no time to react
```
vaxerski: uhm what? There has to be a better way, this is :skull:

LRitzdorf: Yeahhh. So, graphical-session properly waits for dependent services to stop — but for some reason, even with `BindsTo` and `After`, hyprland-session always deactivates immediately. Systemd has the assumption of one graphical session per user anyway, so we should be safe to change this command to stop `graphical-session.target` (making it properly block until everything's had time to exit).

Some users might like to run multiple sessions at once from different TTYs or whatever, in which case exiting one session could break the other. I'll add a note about this in the associated wiki PR, they can just set `HYPRLAND_NO_SD_TARGET` to keep the behavior they had before now

LRitzdorf:
Should be significantly nicer now. Tested this under UWSM and vanilla, and we no longer have that janky sleep while services stop

fufexan (dismissed):

vaxerski:
fucked rn will check later

src/Compositor.cpp
```diff
@@ -541,7 +541,12 @@ void CCompositor::cleanEnvironment() {
     if (m_desktopEnvSet)
         unsetenv("XDG_CURRENT_DESKTOP");
 
-    if (m_aqBackend->hasSession() && !Env::envEnabled("HYPRLAND_NO_SD_VARS")) {
+    if (m_aqBackend->hasSession() && !Env::envEnabled("HYPRLAND_NO_SD_VARS") && !getenv("MANAGERPID")) {
+#ifdef USES_SYSTEMD
+        if (!Env::envEnabled("HYPRLAND_NO_SD_TARGET"))
+            // stopping hyprland-session doesn't wait for dependent services; this does
+            Config::Supplementary::executor()->spawn("systemctl --user stop graphical-session.target");
```
vaxerski, line 548: environment can change. This env needs to be checked once at start, then stored.

LRitzdorf:
Test failure is keybinds again, because of course it is

src/Compositor.hpp
```diff
@@ -75,6 +75,10 @@ class CCompositor {
     bool                     m_wantsXwayland          = true;
     bool                     m_onlyConfigVerification = false;
 
+#ifdef USES_SYSTEMD
+    bool m_sdSessionTarget = false;
```
vaxerski: don't put ABI behind ifdefs. Plugins.

Just keep the bool.

LRitzdorf: Ah okay, I was wondering about that. Didn't consider plugins though

vaxerski (approved):

tekstryder:
Thanks for landing this.

Fwiw, I've replaced Arch-patched x-d-p with vanilla upstream build and all is happy.

Tested session startup and teardown with no issue.

```
~ ❯ systemctl --user list-dependencies --reverse graphical-session.target
graphical-session.target
● ├─xdg-desktop-portal.service
● └─hyprland-session.target
```

```
~ ❯ systemctl --user status hyprland-session.target                 
● hyprland-session.target - Hyprland session
     Loaded: loaded (/usr/lib/systemd/user/hyprland-session.target; static)
     Active: active since Sun 2026-08-23 09:02:07 EDT; 4min 11s ago
 Invocation: f9a31f4f13e74c109348f71e64d5433b

Aug 23 09:02:07 boomshanka systemd[309731]: Reached target Hyprland session.
```

```
Aug 23 09:01:50 systemd[302209]: Stopped target Hyprland session.
Aug 23 09:02:07 systemd[309731]: Reached target Session services which should run early before the graphical session is brought u>
Aug 23 09:02:07 systemd[309731]: Reached target Current graphical user session.
Aug 23 09:02:07 systemd[309731]: Reached target Hyprland session.
```

vaxerski:
@fufexan does this need any nix changes? I've been hitting a similar problem on nix but havent rebuilt yet

fxzzi:
shouldn't need anything just chekc `systemctl status --user hyprland-session.target` and if it's being loaded from `/run/current-system/sw/share/systemd/user/hyprland-session.target` then you're done

vaxerski:
tnx

PR #15779, opened 2026-08-07
desktop: refactor views, major cleanups

vaxerski:
Tons of stuff cleaned up, etc.

Making a MR already because this is growing very very very large


src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
```diff
@@ -336,14 +340,20 @@ void CScrollingFullscreenHandler::setNoMembersAboveFullscreen() {
         const auto MONITOR   = WORKSPACE->m_monitor;
 
         for (auto const& w : Desktop::windowState()->windows()) {
-            if (w && w->m_workspace == getSpace()->workspace() && !isFullscreen(w->m_target) && !w->m_pinned) {
-                w->m_allowedOverFullscreen = !SET;
+            if (w && w->m_workspace == getSpace()->workspace() && !isFullscreen(w->windowTarget()) &&
+                (w->m_state & Desktop::View::WINDOW_STATE_PINNED) == Desktop::View::WINDOW_STATE_NONE) {
```
erstarr:
```suggestion
                !(w->m_state & Desktop::View::WINDOW_STATE_PINNED)) {
```
Reads more naturally as "must not be pinned"

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
```diff
@@ -336,14 +340,20 @@ void CScrollingFullscreenHandler::setNoMembersAboveFullscreen() {
...
+                w->fullscreenPolicy().setAllowedOverFullscreen(!SET);
                 w->updateFullscreenInputState();
             }
         }
         for (auto const& ls : Desktop::layerState()->layers()) {
-            if (ls->m_monitor == MONITOR)
-                ls->m_aboveFullscreen = !SET;
+            if (ls->m_monitor != MONITOR)
```
erstarr, line 344: This is more inefficient than

```suggestion
            if (ls->m_monitor == MONITOR)
```

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
```diff
@@ -427,15 +437,15 @@ void CScrollingFullscreenHandler::setNoMembersAboveFullscreen() {
                 if (!w || w->m_workspace != getSpace()->workspace())
                     continue;
 
-                if (w != COVERING_FULLSCREEN_WINDOW && !w->m_pinned) {
+                if (w != COVERING_FULLSCREEN_WINDOW && (w->m_state & Desktop::View::WINDOW_STATE_PINNED) == Desktop::View::WINDOW_STATE_NONE) {
```
erstarr:
```suggestion
                if (w != COVERING_FULLSCREEN_WINDOW && !(w->m_state & Desktop::View::WINDOW_STATE_PINNED)) {
```

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
```diff
@@ -658,7 +668,7 @@ void CScrollingFullscreenHandler::saveCurrentFsAndAllHiddenFloatingWindows(PHLWI
         if (!w || w->m_workspace != getSpace()->workspace())
             continue;
 
-        if (w != fullscreenWindow && !w->m_pinned && w->m_isFloating)
+        if (w != fullscreenWindow && (w->m_state & Desktop::View::WINDOW_STATE_PINNED) == Desktop::View::WINDOW_STATE_NONE && w->isFloating())
```
erstarr:
```suggestion
        if (w != fullscreenWindow && !(w->m_state & Desktop::View::WINDOW_STATE_PINNED) && w->isFloating())
```

src/managers/fullscreen/handler/FullscreenHandler.cpp
```diff
@@ -279,14 +284,20 @@ void IFullscreenHandler::setNoMembersAboveFullscreen() {
     const bool SET = hasFullscreen(true);
 
     for (auto const& w : Desktop::windowState()->windows()) {
-        if (w && w->m_workspace == getSpace()->workspace() && !isFullscreen(w->m_target) && !w->m_pinned) {
-            w->m_allowedOverFullscreen = !SET;
+        if (w && w->m_workspace == getSpace()->workspace() && !isFullscreen(w->windowTarget()) &&
+            (w->m_state & Desktop::View::WINDOW_STATE_PINNED) == Desktop::View::WINDOW_STATE_NONE) {
```
erstarr:
```suggestion
            !(w->m_state & Desktop::View::WINDOW_STATE_PINNED)) {
```

src/managers/fullscreen/handler/FullscreenHandler.cpp
```diff
@@ -279,14 +284,20 @@ void IFullscreenHandler::setNoMembersAboveFullscreen() {
...
+            w->fullscreenPolicy().setAllowedOverFullscreen(!SET);
             w->updateFullscreenInputState();
         }
     }
     for (auto const& ls : Desktop::layerState()->layers()) {
-        if (ls->m_monitor == MONITOR)
-            ls->m_aboveFullscreen = !SET;
+        if (ls->m_monitor != MONITOR)
```
erstarr, line 293: This is more inefficient than

```suggestion
            if (ls->m_monitor == MONITOR)
```

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -36,29 +39,29 @@ bool CFullscreenController::isFullscreen(const PHLWINDOW window, const std::opti
...
         const auto FS_TARGET     = FS_HANDLER->getFullscreen(covering);
-        const auto INTERNAL_MODE = FS_HANDLER->getFullscreenModes(window->m_target).internal;
+        const auto INTERNAL_MODE = FS_HANDLER->getFullscreenModes(window->windowTarget()).internal;
 
         if (FS_TARGET && FS_WINDOW == FS_TARGET->window() && INTERNAL_MODE != FSMODE_NONE && (!mode.has_value() || INTERNAL_MODE == mode.value()))
             return true;
         else {
             FS_HANDLER->syncFullscreenTargets();
-            return FS_HANDLER->isFullscreen(window->m_target, mode, covering);
+            return FS_HANDLER->isFullscreen(window->windowTarget(), mode, covering);
```
erstarr, line 52: General Suggestion:

target->window() gets group's current window, but this get's m_target of window. Might be confusing.

Maybe renamde window() in ITarget to `layoutWindow()`?

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -289,10 +292,32 @@ std::string CFullscreenController::getFullscreenHandlerNameAsString(const PHLWIN
 
 void CFullscreenController::setFullscreenMode(const PHLWINDOW window, std::optional<eFullscreenMode> internal, std::optional<eFullscreenMode> client,
                                               std::optional<bool> layoutAware) {
+    setFullscreenModeImpl(window, internal, client, layoutAware, false);
```
erstarr: Please keep internal/client mode settings as private methods. Outside should only have access to one unified method for setting both modes. There's no reason to have a unified `setFullscreenMode()` for setting both and also indiv functions for setting each. It used to be like that, and it was supremely confusing. Please just add code in these functions as conditions to the unified function (`setFullscreenMode`)

`setFullscreenModeImpl()` addition is fine, if redundant - I don't see the benefit in not having the code in `setFullscreenMode()`.

as far as i can tell, maximiseEcho is mainly a tool for tests, which is fine but why not just add them to the start and end of `setFullscreenMode()` instead then.

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -289,10 +292,32 @@ std::string CFullscreenController::getFullscreenHandlerNameAsString(const PHLWIN
...
+void CFullscreenController::setInternalFullscreenMode(const PHLWINDOW window, eFullscreenMode internal, bool layoutAware) {
+    if (!window)
+        return;
+
+    window->fullscreenPolicy().clearExpectedMaximizeEcho();
+    setFullscreenModeImpl(window, internal, std::nullopt, layoutAware, true);
+    window->fullscreenPolicy().clearExpectedMaximizeEcho();
+}
+
+void CFullscreenController::clearClientFullscreenMode(const PHLWINDOW window) {
```
erstarr: This is already done by setting the client mode to `FSMODE_NONE`. Explicit function for this is undesirable.

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -289,10 +292,32 @@ std::string CFullscreenController::getFullscreenHandlerNameAsString(const PHLWIN
 
 void CFullscreenController::setFullscreenMode(const PHLWINDOW window, std::optional<eFullscreenMode> internal, std::optional<eFullscreenMode> client,
                                               std::optional<bool> layoutAware) {
+    setFullscreenModeImpl(window, internal, client, layoutAware, false);
+}
+
+void CFullscreenController::setInternalFullscreenMode(const PHLWINDOW window, eFullscreenMode internal, bool layoutAware) {
```
erstarr: Same as above comment, exposing these publicly is not desirable.

Please keep the previous "one public method to set them all" approach.

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -310,46 +335,30 @@ void CFullscreenController::setFullscreenMode(const PHLWINDOW window, std::optio
...
+    const auto OLD_FS_MODES        = ORIGINAL_FS_HANDLER->getFullscreenModes(window->windowTarget());
 
     const auto TO_BE_USED_FS_HANDLER = getFsHandler(window, layoutAware.value_or(WAS_LAYOUT_HANDLED));
     if (!TO_BE_USED_FS_HANDLER) {
-        Log::logger->log(Log::ERR, "window {} doesn't have FS handler assinged. This should never happen", window->m_title);
+        Log::logger->log(Log::ERR, "window {} doesn't have FS handler assinged. This should never happen", window->metadata().title());
         return;
     }
 
     // handles FSMODE_MAX case. If nothing to handle on that front, saves the provided old values
```
erstarr, line 327: I advocate for storing all FS state information in one place - FS handlers and controllers. This was one of the main points of the FS rework after all. This spreads FS state information back to windows.

Please keep the old behaviour of storing FSMODE_MAX states in controller, or otherwise use a method to keep record of FSMODE_MAX window information inside fullscreen related files (a custom handler for just this case for example, though this seems redundant to me when we can just store this exceptional case in the controller).

I don't feel too strongly about this, unless this sets a precedent of spreading FS state information away from FS related files.

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -398,24 +407,26 @@ void CFullscreenController::setFullscreenMode(const PHLWINDOW window, std::optio
         Handling Pinned windows - allow_pin_fullscreen
         Pinned windows can only be floating, therefore it is guaranteed that they will use the same FS handler within the workspace
     */
-    const bool          WINDOW_IS_ALREADY_INTERNAL_FS_HANDLER_AGNOSTIC = OLD_FS_MODES.internal != FSMODE_NONE;
-    const bool          HANDLE_PINNED_WINDOW                           = window->m_pinned || window->m_pinFullscreened;
-    std::optional<bool> pinnedWindowRequetsInternalFS                  = std::nullopt;
+    const bool WINDOW_IS_ALREADY_INTERNAL_FS_HANDLER_AGNOSTIC = OLD_FS_MODES.internal != FSMODE_NONE;
+    const bool HANDLE_PINNED_WINDOW = (window->m_state & Desktop::View::WINDOW_STATE_PINNED) != Desktop::View::WINDOW_STATE_NONE || window->fullscreenPolicy().pinFullscreened();
```
erstarr:
```suggestion
    const bool HANDLE_PINNED_WINDOW = (window->m_state & Desktop::View::WINDOW_STATE_PINNED) || window->fullscreenPolicy().pinFullscreened();
```

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -398,24 +407,26 @@ void CFullscreenController::setFullscreenMode(const PHLWINDOW window, std::optio
...
+    std::optional<bool> pinnedWindowRequetsInternalFS = std::nullopt;
 
     static auto         PALLOWPINFULLSCREEN = CConfigValue<Config::INTEGER>("binds:allow_pin_fullscreen");
-    if (*PALLOWPINFULLSCREEN && !window->m_pinFullscreened && window->m_pinned && !WINDOW_IS_ALREADY_INTERNAL_FS_HANDLER_AGNOSTIC)
+    if (*PALLOWPINFULLSCREEN && !window->fullscreenPolicy().pinFullscreened() && (window->m_state & Desktop::View::WINDOW_STATE_PINNED) != Desktop::View::WINDOW_STATE_NONE &&
```
erstarr:
```suggestion
    if (*PALLOWPINFULLSCREEN && !window->fullscreenPolicy().pinFullscreened() && (window->m_state & Desktop::View::WINDOW_STATE_PINNED)  &&
```

src/managers/fullscreen/FullscreenController.cpp
```diff
@@ -462,15 +474,15 @@ void CFullscreenController::setWindowFullscreenModeInternal(const PHLWINDOW wind
             setFullscreenMode(COVERING_FS_WINDOW, FSMODE_NONE);
     }
 
-    // arm m_suppressNextMaximize to swallow the set_maximized echo on fullscreen exit
-    if (INTERNAL_FS_MODE_CHANGED && !window->m_isFloating && (getFullscreenModes(window).internal == FSMODE_FULLSCREEN) && mode != FSMODE_FULLSCREEN)
-        window->m_suppressNextMaximize = true;
+    // Swallow the set_maximized echo clients send after fullscreen exit.
```
erstarr, line 455: Comment explaining what this practically implies is helpful for posterity

erstarr (changes requested):

src/config/lua/objects/LuaLayerSurface.cpp
```diff
@@ -64,13 +64,13 @@ static int layerSurfaceIndex(lua_State* L) {
...
     } else if (key == "mapped")
-        lua_pushboolean(L, ls->m_mapped);
+        lua_pushboolean(L, ls->mapped());
     else if (key == "layer")
         lua_pushinteger(L, sc<lua_Integer>(ls->m_layer));
     else if (key == "interactivity")
         lua_pushinteger(L, sc<lua_Integer>(ls->m_interactivity));
     else if (key == "above_fullscreen")
-        lua_pushboolean(L, ls->m_aboveFullscreen);
+        lua_pushboolean(L, (ls->m_flags & Desktop::View::LAYER_FLAG_ABOVE_FULLSCREEN) != Desktop::View::LAYER_FLAG_NONE);
```
erstarr: General comment: this doesn't read as "must not be x" naturally. It's preference but i prefer.

Except FS stuff which i explicitly commented on each occurrence, I'll only leave this one comment here so you can easily dismiss if you disagree.

```suggestion
        lua_pushboolean(L, (ls->m_flags & Desktop::View::LAYER_FLAG_ABOVE_FULLSCREEN);
```

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
vaxerski: yes but I need to add boolean comparison to maskable enum types in the macro, this wont compile atm

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
vaxerski: what?

src/managers/fullscreen/FullscreenController.cpp
vaxerski: nah

vaxerski:
any further problems?

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
erstarr: Pedantic but `==` version is slightly faster in debug builds: https://godbolt.org/z/q6ob571r5

erstarr:
Unresolved reviews:

- FSMODE_MAX - either your way or my way either is fine
- that loop thing you can just ignore ;)

Otherwise,

<details>
<summary> ignoring (or rather considering) the parts that'll change once you rebase this on main </summary>

<br>
layer hiding no longer sets them above FS, just sets them below in all cases like before --- observed no breakages with this in the MR it was changed and this was the way it was before the FS rework (makes little sense, why set them below FS if you're unFSing and why does this not really matter) so there's merit consolidating or refactoring if need be the FS attributes layers use to stay ontop of fullscreens

<br>
</details>

I've given the FS parts another look over and it looks good to me.

I'll give it another go-over tomorrow and, since i'll hopefully be on nix tomorrow, can test it baremetal.

It's a good idea to expand the echo framework to be a bitmap of information that includes layout/default handled and fullscreen/maximised so some parts of the code, especially inter-handler communication, can be considerably simplified. If you don't wanna do it here, i can do this in another MR. I was floating the idea of doing it with events so this is very helpful

Thanks for the work

src/managers/fullscreen/FullscreenController.cpp
vaxerski: I dont understand how that relates to the line at hand. Are you talking about SPendingClientFullscreenRequest?

src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
vaxerski: and in release it makes no difference, but is much cleaner to read.

src/managers/fullscreen/FullscreenController.cpp
erstarr: correct. But as I said, I've come around to your way of doing this, so it's fine

erstarr (dismissed):
Noot Noot

vaxerski:
at last

non-bin:
Not sure why it wasn't automatically linked here, but this PR caused most apps to launch maximised. I'd love some input from someone more knowledgeable, so here's the discussion regarding the issue #15901

non-bin:
While I was diagnosing the issue, I also noticed that the logic in `CWindowFullscreenPolicy::consumeExpectedMaximizeEcho` is a little weird

https://github.com/hyprwm/Hyprland/blob/8d50be06aa7d84283ac364a03df74bd834cab0ee/src/desktop/view/window/WindowFullscreenPolicy.cpp#L58-L64

Rearranging it to make my tiny brain less confused, we get

```cpp
bool CWindowFullscreenPolicy::consumeExpectedMaximizeEcho(bool maximizeRequested) {
    if (m_expectsMaximizeEcho) {
        m_expectsMaximizeEcho = false;
        return maximizeRequested;
    }

    return false;
}
```

But I would expect something closer to 

```cpp
bool CWindowFullscreenPolicy::consumeExpectedMaximizeEcho(bool maximizeRequested) {
    if (m_expectsMaximizeEcho) {
        m_expectsMaximizeEcho = false;
        return false;
    }
    
    return maximizeRequested;
}
```

And for what it's used for in [src/desktop/view/window/Window.cpp#L850](https://github.com/hyprwm/Hyprland/blob/8d50be06aa7d84283ac364a03df74bd834cab0ee/src/desktop/view/window/Window.cpp#L850) I think this would make more sense

```cpp
// src/desktop/view/window/WindowFullscreenPolicy.cpp#L58
bool CWindowFullscreenPolicy::consumeExpectedMaximizeEcho() {
    if (m_expectsMaximizeEcho) {
        m_expectsMaximizeEcho = false;
        return false; // Ignore the request
    }
    
    return true; // Use the request
}

// src/desktop/view/window/Window.cpp#L850
if (!BACKEND_REQUEST || m_fullscreenPolicy->consumeExpectedMaximizeEcho()) {
```

Sorry if I'm way off, I'm new to the project :)

vaxerski:
that thing is there because we initially set all windows to maximized and it should consume the apps echo'ing it back. I don't seem to get the issue you're experiencing myself. Any rules?

non-bin:
> we initially set all windows to maximized

Is this a new behaviour? I thought it was a bug

I understand why `consumeExpectedMaximizeEcho` is there, just a little confused by the logic but if it looks good to you then don't worry :P

vaxerski:
no, we always send MAXIMIZED to apps, and we have been doing this forever. It's to force them to not draw non-window CSD.

non-bin:
Oh interesting, that's clever. I wonder if they are sending an echo in response that's not being properly ignored, or maybe requesting not to be maximised? Doesn't matter why the request is happening, but we were ignoring them until this PR.

Do you have any objections to #15929?

vaxerski:
yes, I do, it worked fine beforehand without the var :P

non-bin:
Do you have this in your config? (It's in the example so many people do)

```lua
hl.window_rule({
  match = {
    class = ".*"
  },
  suppress_event = "maximize"
});
```

I don't, and don't want to because I want some programs to be able to request maximize, just not immediatly on startup

vaxerski:
ah, I might.

PR #15780, opened 2026-08-07
input: do not enforce keyboard exclusivity during dnd

aurelleb:

Keyboard exclusivity is enforced even during drag and drop, so other surfaces never have an opportunity to act on it.

Basically this breaks drag and drop for layer surfaces with keyboard interactivity set to exclusive.

ready


vaxerski (approved):
at some point we need to make some better priority-aware code there this shit sucks balls

PR #15784, opened 2026-08-07
layout: fix pinned floating windows not reverting to last postion after exiting fullscreen

epsilonshmepsilon:
Having a pinned floating window and changing workspaces doesn't call `updateTarget(target);` in `void CDefaultFloatingAlgorithm::movedTarget()` because of an early return for pinned windows. `.lastBox` is erased for our pinned floating window in `m_datas` on workspace change. So after exiting fullscreen after a workspace change the pinned floating window is getting centered in the middle of the monitor in `void CDefaultFloatingAlgorithm::recenter()`.

Used LLM to explore codebase and find the problem.

Should be ready. Expanded the test that checks the restoring of floating window positions after cycling through fullscreen.


vaxerski (approved):
thanks

PR #15792, opened 2026-08-08
shaders: make each SRC/DST TF a variant

gulafaran:
by making each SRC/DST transferunction a variant we reduce the branching and instruction count by up to 60%, measured with INTEL_DEBUG=fs,gs,vs

while we are touching the same code paths, just cache 9% of redundant apitrace churn.


gulafaran:
yeah broken keybind test, unrelated.

vaxerski (approved):
huh, neat. I guess the shader compiler can optimize the ifs out when the condition is a constexpr.

Thanks!

PR #15796, opened 2026-08-08
Vouch: Allow vouched people to open issues

erstarr:

Title

Companion: https://github.com/hyprwm/.github/pull/23

WIP


erstarr:
wanna change the issue message and the like?

I don't think it matters much but parts about 'only maintainers are allowed to open issues' in the wiki/discord are no longer necessarily true with this -- though as vax said it's not like anyone who's in the MR vouch list can so it's rather more strongly gated so changing the phrasing there is neither here nor there.

vaxerski (approved):

PR #15803, opened 2026-08-09
config: enable skip_non_kms_dmabuf_formats by default

fxzzi:

fixes stuff for nvidia users, don't see why we can't enable it by default for us plebians
or maybe @UjinT34 has an objection

https://github.com/hyprwm/Hyprland/discussions/15781
https://github.com/hyprwm/Hyprland/discussions/14843

naw

ye

### DIFF
--- a/src/config/values/ConfigValues.cpp
+++ b/src/config/values/ConfigValues.cpp
@@ -752,7 +752,7 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          */
 
         MS<Int>("quirks:prefer_hdr", "Prefer HDR mode.", 0, {.min = 0, .max = 2, .map = OptionMap{{"disable", 0}, {"enable", 1}, {"gamescope_only", 2}}}),
-        MS<Bool>("quirks:skip_non_kms_dmabuf_formats", "Do not report dmabuf formats which cannot be imported into KMS", false),
+        MS<Bool>("quirks:skip_non_kms_dmabuf_formats", "Do not report dmabuf formats which cannot be imported into KMS", true),
     };
 
 #undef MS


src/config/values/ConfigValues.cpp
```diff
@@ -747,7 +747,8 @@ std::vector<SP<IValue>> Values::getConfigValues() {
          */
 
         MS<Int>("quirks:prefer_hdr", "Prefer HDR mode.", 0, {.min = 0, .max = 2, .map = OptionMap{{"disable", 0}, {"enable", 1}, {"gamescope_only", 2}}}),
-        MS<Bool>("quirks:skip_non_kms_dmabuf_formats", "Do not report dmabuf formats which cannot be imported into KMS", false),
+        MS<Int>("quirks:skip_non_kms_dmabuf_formats", "Do not report dmabuf formats which cannot be imported into KMS. Auto enables this on NVIDIA.", 2,
```
vaxerski: "auto" enables this...

vaxerski (dismissed):

vaxerski (dismissed):

vaxerski:
is there any reason not to skip KMS-unimportable buffer types overall?

fxzzi:
honestly yes it might be best to just set this as a default overall, seems like it's helped some AMD users in specific instances as well, lmk if you want me to just do that instead. and maybe move the option into debug

https://github.com/hyprwm/Hyprland/discussions/14843#discussioncomment-17968499

vaxerski:
let's make it default to auto then and auto sets this to yes as long as there are both 8b and 10b KMS formats available

fxzzi:
not exactly sure how i should do that :P do you want me to check if there's at least one monitor in a 10 bit format?

or i could maybe check if both DRM_FORMAT_XRGB8888 and DRM_FORMAT_XRGB2101010 are available kms formats for any monitor, there isn't really a list of 8bit formats and 10bit formats separately anywhere

UjinT34:
> is there any reason not to skip KMS-unimportable buffer types overall?

those might work faster or take less vram. probably won't make that much of a difference. default is off to keep the old behaviour. no reason to prefer any specific default over the other. feel free to change it to something that looks more convenient

gulafaran:
> > is there any reason not to skip KMS-unimportable buffer types overall?
> 
> those might work faster or take less vram. probably won't make that much of a difference. default is off to keep the old behaviour. no reason to prefer any specific default over the other. feel free to change it to something that looks more convenient

technically this might just be a workaround for something the tranche is doing wrong, if im not reading this wrong weston sends two tranches, one all render formats, then another with ZWP_LINUX_DMABUF_FEEDBACK_V1_TRANCHE_FLAGS_SCANOUT and only kms importable ones.

UjinT34:
> technically this might just be a workaround for something the tranche is doing wrong, if im not reading this wrong weston sends two tranches, one all render formats, then another with ZWP_LINUX_DMABUF_FEEDBACK_V1_TRANCHE_FLAGS_SCANOUT and only kms importable ones.

iirc HL should do the same. mb with some minor sorting differences.
A client is expected to change its format when it receives scanout tranches. wine-wayland doesn't even listen for this event (at least I couldn't find any proto mentions). That workaround just forces an importable format from the start (and that part is probably handled on vulkan driver side so no wine support is needed)

vaxerski:
I think we can default this to true for now. I don't see any reason why not to as long as filtering doesn't leave empty format lists.

UjinT34:
> I think we can default this to true for now. I don't see any reason why not to as long as filtering doesn't leave empty format lists.

Needs testing with headless. Otherwise at least formats used by HL itself will remain in the list.

vaxerski:
tester runs in headless so it should be fine if it passes

PR #15805, opened 2026-08-09
fix a lot of expand_undersized resize artifact/smear/issues

gulafaran:
we have to check against animation target size, otherwise the undersized
checks misinterprets it as undersized and smears and artifacts the UV

a client can ack a configure and attach the buffer for it in separate
commits, and we may have configured it again in between. the buffer it
still presents then matches an older configure, small() fires and the box
shrinks below the window, leaving the rest of it unpainted for a frame.

Before:

https://github.com/user-attachments/assets/2f5a8ae8-5033-4c0a-9cdd-6311aeab5022

https://github.com/user-attachments/assets/9834349e-72ad-4e78-bce4-0b05c7514c95

After:

https://github.com/user-attachments/assets/7782726a-aa1a-461c-90f1-f8b6c68d974c

### DIFF
--- a/src/desktop/view/WLSurface.cpp
+++ b/src/desktop/view/WLSurface.cpp
@@ -51,9 +51,16 @@ bool CWLSurface::small() const {
     if (!O || !O->mapped() || !O->acceptsInput() || !O->alphaNonZero())
         return false;
 
-    const auto REPORTED_SIZE = O->backend().reportedSize();
+    const auto  REPORTED_SIZE = O->backend().reportedSize();
+    const auto& CURRENT_SIZE  = m_resource->m_current.size;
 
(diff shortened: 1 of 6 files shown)


vaxerski:
conveniently forgot to show the terminals. I reaaaallly wonder what kind of magic you wish to do on those.

gulafaran:
> conveniently forgot to show the terminals. I reaaaallly wonder what kind of magic you wish to do on those.


https://github.com/user-attachments/assets/1ccaa39e-e27f-4842-97b3-b78e9581dcea

gulafaran:
i have over 800 pngs, of frame per frame from before and after each change. recorded and tested the changes throughly what it did change, and as i said on discord. its claude that hinted it here and reasoned why. as i said there, i dont trust it, go over it because it is claude slop. dont just merge it blindly. but it certainly is an issue with the UV

vaxerski:
well what you did is just reverting back to what we used to do ages ago and what happens with expand false.

PR #15810, opened 2026-08-09
dispatch/center: `hl.dsp.window.center()` doesn't save the new pos of the floating window

erstarr:

Fixes bug where hl.dsp.window.center() doesn't save the new pos of the floating window

Fixes case:

1. spawn floating kitty
2. Move it to somewhere not centre
3. hl.dsp.window.center()
4. FS and unFS
5. The window won't return to the centre but to where you moved it before

No

- [x] Test (also incidentally tests `hl.dsp.window.center()` since there wasn't test for this before)

### DIFF
--- a/hyprtester/src/tests/main/dwindle.cpp
+++ b/hyprtester/src/tests/main/dwindle.cpp
@@ -1180,3 +1180,46 @@ TEST_CASE(defaultHandledFsfocusInDirection) {
         EXPECT_CONTAINS(str, "fullscreenClient: 0");
     }
 }
+
+TEST_CASE(dwindleFullsreenCenterDispatchSavesPos) {
+    /*
+        This test serves as a test for all layouts that use deafult FS behaviour
+
+        Also tests hl.dsp.window.center() incidentally
+
+    */
+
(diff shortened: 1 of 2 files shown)


erstarr:
THIS IS A PATCH! A fix of the root cause would involve: Refactoring floating window size/pos setting and integrate logic of floating window pos/size setting, integrating the logic of saving/updating past window data, integrating calculation of adjusted CBOX as a result of `float_force_onscreen` into `CWindowTarget::setPositionGlobal` and getting rid of the default floating algo's own `setPositionGlobal`, automatically updating `m_datas` of the def floating also when size/pos set, adding explicit setting of unFSed floating window's size in defaultFullscreenHandler, and possible more.

vaxerski (approved):
looks fine to me

erstarr:
benign but that if is redundant since it's checked above. Was trying to get rid of default floating algo having its own functions that must be called to save pos properly and forgot to clean that up

https://github.com/hyprwm/Hyprland/pull/15814

vaxerski:
banana

PR #15842, opened 2026-08-12
monitor: fix layer arrangement on layout changes

vaxerski:
halal code


leiserfg:
It works, thanks, that was driving me crazy.

vaxerski:
mcdouble

PR #15863, opened 2026-08-14
desktop/windowRule: add no_glow window rule

Aqa-Ib:
Add no_glow window and workspace rules

Wiki MR: https://github.com/hyprwm/hyprland-wiki/pull/1631

ready


Aqa-Ib:
ready

Dregu:
Ah yes the thing I wanted for my laptop and said I would look at this Monday. I don't think we should add any more *workspace* rules for things that are essentially window rules. I tried to block the last one too. Unless someone can tell me why a workspace rule is specifically needed to change properties of windows, all the 5ish legacy workspace rules that just affect window properties and also have window rules should just be removed...

As for the general functionality, yeah innie decos should have a way to do "smart gaps" just like outie decos, I'll test soon.

Dregu:
Tested, yes, the window rule can be used to disable glow for a lone window like smart gaps can disable gaps for a lone window. Good enough for me.

Aqa-Ib:
I bet that there are weird people out there who wants a specific workspace to not have glow, wobble, or any of that window stuff.

Dregu:
Yes you can match specific workspaces in window rules too.

Aqa-Ib:
I didn`t know that. Then I don't know why is that stuff there in the first place.

Dregu:
Well like I said, legacy, it's old. But nevertheless, `hl.window_rule{ match = { workspace = 1 }, no_glow = true }` is very much the same thing as `hl.workspace_rule{ workspace = 1, no_glow = true }`, they just have a slightly different code path to do the same thing.

Aqa-Ib:
In case all of that old code may be removed, it should be in a different MR anyway.

Dregu:
Yeah for sure, just don't add to the pile.

Aqa-Ib:
I can remove it if you wish, Do I remove no_wobble workspace rule also?

src/desktop/rule/windowRule/WindowRuleApplicator.cpp
```diff
@@ -61,7 +62,7 @@ std::unordered_set<CWindowRuleEffectContainer::storageType> CWindowRuleApplicato
             std::pair{std::ref(m_scrollMouse), [this] { return scrollMouseEffect(); }}, std::pair{std::ref(m_scrollTouchpad), [this] { return scrollTouchpadEffect(); }},
             std::pair{std::ref(m_animationStyle), [this] { return animationStyleEffect(); }}, std::pair{std::ref(m_maxSize), [this] { return maxSizeEffect(); }},
             std::pair{std::ref(m_minSize), [this] { return minSizeEffect(); }}, std::pair{std::ref(m_activeBorderColor), [this] { return activeBorderColorEffect(); }},
-            std::pair{std::ref(m_inactiveBorderColor), [this] { return inactiveBorderColorEffect(); }}, std::pair{std::ref(m_noWobble), [this] { return noWobbleEffect(); }}));
```
vaxerski, line 64: akhem.

Aqa-Ib: just reordered it to follow the same order as in other places.

Aqa-Ib:
Removed workspace rules no_glow and no_wobble as per discussion above.

vaxerski (approved):

PR #15869, opened 2026-08-15
lua(window): make `stable_id` field a string

LRitzdorf:

Makes window objects' `stable_id` field a hex string (it was a decimal number before). That's literally it.

This matches the format shown by `hyprctl activewindow/clients`, and the format that the `stableid:` window selector wants. Without that, users would have to e.g.
```lua
local stable_id = string.format("%x", hl.get_active_window().stable_id)
hl.dsp.whatever({ window = "stableid:"..stable_id })
```
Which is just unnecessary.

Also, the Wayland protocol that stable_id exists for, passes it as a string too.

Nerp

Ready


vaxerski (approved):
I'm at a payphone trying to call home

PR #15882, opened 2026-08-17
scrolling: fix incorrect viewport movement when closing a group member

Aqa-Ib:
This fixes a bug in scrolling layout: when closing a group member, the viewport would move to the window in the next column instead of staying in place.

The call was missing the reason, so the function would assign the window being unmapped to a new space, and in consequence bugging the layout.

Tested and ready.
Cheers


erstarr:
Write a test

Aqa-Ib:
Can we test where is the scrolling viewport at?

erstarr:
Yes. 3 kitty, 2 grouped. make the grouped ones 0.8 col width. close one of them.

I believe the bug moved the viewport to the right, so keep the grouped tabs on the left. Check the current active window to make sure it's the grouped one, and check its at: and size: ==> closing a grouped window must NOT move the viewport (https://github.com/hyprwm/Hyprland/pull/13974 established this - I recall writing a test for this but proabbly didn't fail since the grouped windows were on the right), so test that the size: and at: of the current window match the prev (now closed) window - this confirms that the viewport did not move


Excuse the wall of text

Aqa-Ib:
fixed on main by af0d014cb26f536d8cb7cab2b9d5784f69767c8a

PR #15891, opened 2026-08-18
layout/scrolling: defer click-triggered view movement to button release

corrm:
Clicking a window that is not fully in view in the scrolling layout moved
the view while the button was still held. The window slides under the cursor
mid-press, so the app receives pointer motion during the press and treats
the click as a drag (text selection, file drag, ...). With
`input:follow_mouse = 2` this happens on every click, with `1` it happens
occasionally (fast clicks, follow_mouse_threshold misses).

The view move now happens on button release instead of on focus: clicks are
skipped in the focus callback, and the existing release-time callback (added
for #13351) performs the same focusOnInput call after the button is up, where
the motion is harmless. The clicked window still ends up in view, just one
beat later. Side effect: starting a real drag (text selection, DnD) from a
side window no longer yanks the view under the cursor mid-drag.

One behavior change: with `scrolling:follow_focus = 0`, clicking a side
window no longer moves the view at all (before it moved on press, since a
click is a hard input reason). I believe this matches the intent of the
option - in #13351 users expected follow_focus=0 to mean mouse input has no
impact on the layout.

Tested on v0.56.2 with this change, scrolling layout:
- click on a side window: clean click, no selection made in the app (checked
  via kitty's primary selection before/after), window slides in after release
- keyboard movefocus and wheel over unfocused windows: unchanged
- follow_mouse 1 and 2, follow_focus on and off: all as above

Related to #13351 and #13558.

The patch was drafted with AI assistance (I used it to walk through the
layout code); I reviewed the diff and tested each of the cases above myself.

First PR here, would appreciate a vouch.

### DIFF
--- a/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
+++ b/src/layout/algorithm/tiled/scrolling/ScrollingAlgorithm.cpp
@@ -637,6 +637,12 @@ CScrollingAlgorithm::CScrollingAlgorithm() : m_scrollingFullscreenHandler(makeUn
         if (!pWindow)
             return;
 
+        // Defer click-triggered view movement to button release (see m_mouseButtonCallback):
+        // moving the view while the button is still held slides the window under the cursor
+        // and the app receives pointer-motion mid-press, which it interprets as a drag.
+        if (reason == Desktop::FOCUS_REASON_CLICK)
+            return;
+
         static const auto PFOLLOW_FOCUS = CConfigValue<Config::INTEGER>("scrolling:follow_focus");
 
         if (!*PFOLLOW_FOCUS && !Desktop::isHardInputFocusReason(reason))


erstarr:
I'll respond here regarding your issue, but you should consider opening a discussion if this ends up being a long thread.

So the problem is, when you mouse-down, it moves the viewport; which is a problem as that moves the mouse relative to the application you are currently selecting text is.

The obvious solution is to not consider mouse-down to be able to move the viewport, but that might be a problem because, in the strict definition of the config options, this is working as intended.

See this is only a problem when `follow_mouse = 2` and `follow_focus = false`. `follow_focus = 2` in particular considers focus to move on mouse-down and all follow_focus cares about is that focus moved; so it is working as intended from their viewpoints.

The problem with making the obvious change is; this would be an exception because everywhere else, mouse down is focus move. I made the change and tried, and it feels off that a full click is needed to move the focus and it's not the case anywhere else.

Now i don't have a horse in this particular race, i always use `follow_focus = false`, but others may find the change to be problematic.

So i'd prefer to introduce a config option instead. viewport_focus_move_method or smt? It frankly makes more sense to add an option to `follow_mouse` as focus change on mouse-down or on mouse-up, but i don't know if anyone would ever find it useful to require a full click to change focus?

Anyway if you have a preference, we can discuss it. If not, i'll pick one and implement that.

corrm:
I would vote for full click to move view as option, and mouse up that would be good options

PR #15898, opened 2026-08-19
overview: add overview

vaxerski:
overview.

plugins can override the default overview via one api entry point.

以上です

todo:
- [x] window matches should focus the candidate upon close
- [x] input box is weird when hidden I think
- [x] workspace suggestions from typing numbers could be better
- [ ] wiki
- [x] gesture
- [x] expand filters into filter modes
- [x] allow to pass a default filter in the dispatcher
- [x] allow binding keys to nav in overview
- [x] find a better way to hook keys
- [x] allow creating a new workspace if it makes sense

future:
 - integrate hyprlauncher?
 
image

### DIFF
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -2,7 +2,6 @@
 
 #include "../Overview.hpp"
 #include "../../helpers/AnimatedVariable.hpp"
-#include "../../helpers/time/Timer.hpp"
 #include "../../managers/eventLoop/EventLoopTimer.hpp"
 
 #include <utility>
@@ -10,6 +9,9 @@
 
 namespace Monitor {
     class CMonitorResources;
+}
+namespace Layout {
+    class ITarget;
 }
 class IKeyboard;
 
@@ -31,6 +33,8 @@ namespace Overview::Hyprland {
         void                           closeImmediately();
         void                           installListeners();
         void                           recheckDrag();
+        void                           applyDragHoverTarget();
+        void                           resetDragHover();
         bool                           handleSearchKey(uint32_t keycode, SP<IKeyboard> keyboard, bool repeat = false);
         void                           startKeyRepeat(uint32_t keycode, SP<IKeyboard> keyboard);
         void                           stopKeyRepeat(uint32_t keycode);
@@ -50,12 +54,21 @@ namespace Overview::Hyprland {
             CHyprSignalListener mouseMove;
             CHyprSignalListener sessionLock;
             CHyprSignalListener keyboardKey;
+            CHyprSignalListener dragMotion;
+            CHyprSignalListener dragEnded;
         } m_listeners;
 
+        enum class eDragHoverTarget : uint8_t {
+            NONE,
+            LEFT_EDGE,
+            RIGHT_EDGE,
+            MINI_TILE,
+        };
+
         struct {
-            CTimer debouncer;
-            // met criteria
-            bool                isWithin = false;
+            eDragHoverTarget    target = eDragHoverTarget::NONE;
+            PHLWORKSPACEREF     workspace;
+            WP<Layout::ITarget> dragTarget;
             SP<CEventLoopTimer> eventLoopTimer;
         } m_drag;
 
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -95,6 +95,7 @@ namespace Overview::Hyprland {
             PHLWORKSPACEREF     workspace;
             WP<Layout::ITarget> dragTarget;
             SP<CEventLoopTimer> eventLoopTimer;
+            bool                createdWorkspace = false;
         } m_drag;
 
         struct {
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -31,6 +31,9 @@ namespace Overview::Hyprland {
         void                           closeImmediately();
         void                           installListeners();
         void                           recheckDrag();
+        bool                           handleSearchKey(uint32_t keycode, SP<IKeyboard> keyboard, bool repeat = false);
+        void                           startKeyRepeat(uint32_t keycode, SP<IKeyboard> keyboard);
+        void                           stopKeyRepeat(uint32_t keycode);
 
         bool                           m_isOpen         = false;
         bool                           m_sceneInstalled = false;
@@ -38,6 +41,8 @@ namespace Overview::Hyprland {
         WP<Monitor::CMonitorResources> m_resources;
         PHLANIMVAR<float>              m_progress;
         SP<COverviewScene>             m_scene;
+        struct SXKBComposeState;
+        UP<SXKBComposeState> m_composeState;
 
         struct {
             CHyprSignalListener monitorDisconnect;
@@ -56,7 +61,19 @@ namespace Overview::Hyprland {
             SP<CEventLoopTimer> eventLoopTimer;
         } m_drag;
 
+        struct {
+            WP<IKeyboard>       keyboard;
+            uint32_t            keycode = 0;
+            SP<CEventLoopTimer> timer;
+        } m_keyRepeat;
+
+        enum class eInputMode : uint8_t {
+            NAVIGATION,
+            TEXT
+        } m_inputMode = COverview::eInputMode::NAVIGATION;
+
         std::vector<std::pair<WP<IKeyboard>, uint32_t>> m_interceptedKeys;
+        std::vector<uint32_t>                           m_interceptedButtons;
 
         friend class COverviewScene;
     };
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -41,8 +41,6 @@ namespace Overview::Hyprland {
         WP<Monitor::CMonitorResources> m_resources;
         PHLANIMVAR<float>              m_progress;
         SP<COverviewScene>             m_scene;
-        struct SXKBComposeState;
-        UP<SXKBComposeState> m_composeState;
 
         struct {
             CHyprSignalListener monitorDisconnect;
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -1,0 +1,42 @@
+#pragma once
+
+#include "../Overview.hpp"
+#include "../../helpers/AnimatedVariable.hpp"
+
+namespace Monitor {
+    class CMonitorResources;
+}
+
+namespace Overview::Hyprland {
+    class COverviewScene;
+
+    class COverview : public Overview::IOverview {
+      public:
+        COverview();
+        virtual ~COverview() override;
+
+        virtual void open(PHLMONITOR monitor) override;
+        virtual void close() override;
+        virtual bool isOpen() const override;
+
+      private:
+        void                           finishClose(bool emitEvent = true);
+        void                           closeImmediately();
+
+        bool                           m_isOpen         = false;
+        bool                           m_sceneInstalled = false;
+        PHLMONITORREF                  m_monitor;
+        WP<Monitor::CMonitorResources> m_resources;
+        PHLANIMVAR<float>              m_progress;
+        SP<COverviewScene>             m_scene;
+
+        struct {
+            CHyprSignalListener monitorDisconnect;
+            CHyprSignalListener monitorModeChanged;
+            CHyprSignalListener monitorPreRender;
+            CHyprSignalListener sessionLock;
+        } m_listeners;
+
+        friend class COverviewScene;
+    };
+};
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -28,8 +28,10 @@ namespace Overview::Hyprland {
         virtual ~COverview() override;
 
         virtual void         open(PHLMONITOR monitor) override;
+        void                 open(PHLMONITOR monitor, const std::string& query);
         virtual void         close() override;
         virtual bool         isOpen() const override;
+        SOverviewState       state() const;
         virtual bool         shouldRenderWorkspace(PHLWORKSPACE workspace) const override;
         virtual PHLWORKSPACE inputWorkspace() const override;
         virtual bool         beginOpenGesture(PHLMONITOR monitor) override;
@@ -38,6 +40,8 @@ namespace Overview::Hyprland {
         virtual bool         beginMoveGesture() override;
         virtual void         updateMoveGesture(float Δ) override;
         virtual void         endMoveGesture() override;
+        bool                 moveLeft();
+        bool                 moveRight();
         virtual void         onKeyboardKey(const IKeyboard::SKeyEvent& event, SP<IKeyboard> keyboard) override;
 
         SP<COverviewScene>   scene() const;
--- a/src/overview/hyprland/Overview.hpp
+++ b/src/overview/hyprland/Overview.hpp
@@ -2,6 +2,7 @@
 
 #include "../Overview.hpp"
 #include "../../helpers/AnimatedVariable.hpp"
+#include "../../managers/SeatManager.hpp"
 #include "../../managers/eventLoop/EventLoopTimer.hpp"
 
 #include <utility>
@@ -16,13 +17,12 @@ namespace Layout {
 namespace Pointer {
     class CPointerTransformer;
 }
-class IKeyboard;
 struct SEventLoopDoLaterLock;
 
 namespace Overview::Hyprland {
     class COverviewScene;
 
-    class COverview : public Overview::IOverview, public Overview::IOverviewGestureOpenable, public Overview::IOverviewGestureMovable {
+    class COverview : public Overview::IOverview, public Overview::IOverviewGestureOpenable, public Overview::IOverviewGestureMovable, public IKeyboardEventHandler {
       public:
         COverview();
         virtual ~COverview() override;
@@ -38,6 +38,7 @@ namespace Overview::Hyprland {
         virtual bool         beginMoveGesture() override;
         virtual void         updateMoveGesture(float Δ) override;
         virtual void         endMoveGesture() override;
+        virtual void         onKeyboardKey(const IKeyboard::SKeyEvent& event, SP<IKeyboard> keyboard) override;
 
         SP<COverviewScene>   scene() const;
 
@@ -56,7 +57,7 @@ namespace Overview::Hyprland {
         void                             resetDragHover();
         bool                             handleSearchKey(uint32_t keycode, SP<IKeyboard> keyboard, bool repeat = false);
         void                             startKeyRepeat(uint32_t keycode, SP<IKeyboard> keyboard);
-        void                             stopKeyRepeat(uint32_t keycode);
+        void                             stopKeyRepeat(uint32_t keycode, SP<IKeyboard> keyboard = nullptr);
 
         bool                             m_isOpen            = false;
         bool                             m_sceneInstalled    = false;
@@ -78,7 +79,6 @@ namespace Overview::Hyprland {
             CHyprSignalListener mouseButton;
             CHyprSignalListener mouseMove;
             CHyprSignalListener sessionLock;
-            CHyprSignalListener keyboardKey;
             CHyprSignalListener dragMotion;
             CHyprSignalListener dragEnded;
         } m_listeners;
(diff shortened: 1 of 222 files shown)


src/overview/hyprland/Overview.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Overview::Hyprland {
+    class COverviewScene;
+
+    class COverview : public Overview::IOverview {
+      public:
+        COverview();
+        virtual ~COverview() override;
+
+        virtual void open(PHLMONITOR monitor) override;
```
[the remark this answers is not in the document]
vaxerski: are we deadass

Tom-De-Santa-FOSS:
can you send the anime girl wallpapers to me? it is a legit request

ferrreo:
Could we get like a mini workspace switcher at the bottom of the overview please? 
image
Something like this that when you click it, it moves you (and the overview) to the workspace you facesmashed?

vaxerski:
no why

ferrreo:
Because then the overview works like an actual overview and not just like you could already do with keybinds? It's not really an overview if it isn't giving you a view over all the things

vaxerski:
whats the point of that? You should never need to scour with your eyes over miniature images.

ferrreo:
So you can look for a specific window and jump right to that workspace? It's a pretty standard overview feature

vaxerski:
> You should never need to scour with your eyes over miniature images.

it's slow, inefficient and stupid. I literally gave you a search feature so you can type "fire" and hit enter.

dinhokusanagi:
It would be great if Hyprland had an Overview like this, with thumbnails of each workspace at the top, making it easy to see what’s open and move windows between workspaces by dragging them. Niri falls short in this regard.
Screenshot from 2026-08-19 13-52-48

ferrreo:
> > You should never need to scour with your eyes over miniature images.
> 
> it's slow, inefficient and stupid. I literally gave you a search feature so you can type "fire" and hit enter.

Typing a bunch of stuff in a search box that might collide with other window names and go to the wrong place when you hit enter is not less efficient or slower than just clicking once (or using the hotkey) on the thing you can see right in front of you

KristijanZic:
Personally I'd like to be able to have Niri-style overview without third party plugins.

vaxerski:
> Typing a bunch of stuff in a search box that might collide with other window names and go to the wrong place when you hit enter is not less efficient or slower than just clicking once (or using the hotkey) on the thing you can see right in front of you

disagree considering it works as a filter but ok I'll add that.

> Personally I'd like to be able to have Niri-style overview without third party plugins.

I'd rather not ruin a better overview with a worse one tnx

phucisstupid:
is this like niri overview?

falkkyburz:
Nice I like it. (don't need miniature WS preview)
Maybe it would be nice if the matched window would be visualized before exiting. For WS name matching I am not sure what should be highlighted.
Switching to a different workspace with the keybind when the overview is open just moves the window focus, wouldn't it make sense to switch to that workspace? Not an actual use case but more expected behavior.
~~Also I am not sure if we want window highlight following the mouse when entering overview.~~
I would expect that I can exit the search mode with Escape, or at least exit search mode when i empty the search bar.

Edit: 
With animations = {enabled = false} the window dragging from one WS to the other is broken until i let go of the drag (in nested Hyprland instance).

rachalaraj:
What about something like [this](https://youtu.be/RPwovTInagE?t=20)?

vaxerski:
I've implemented the above things.

definitelynotrazu:
> > Typing a bunch of stuff in a search box that might collide with other window names and go to the wrong place when you hit enter is not less efficient or slower than just clicking once (or using the hotkey) on the thing you can see right in front of you
> 
> disagree considering it works as a filter but ok I'll add that.

Outstanding. It's less useful from your POV since you mostly work in code, but imagine if you have several windows with different images open, you can snipe the right window almost instantly even with a tiny thumbnail.
Thanks for working on it

definitelynotrazu:
In the overview it "zooms out" a little bit, letting you see one workspace, and the edges of of two other workspaces. Can we get an option to control how far it zooms out? On very large monitors I would like to be able to zoom wayyyy out. (Unless this is already an option)

alba4k:
text cursor position is reset when pressing esc out of the text box, which is annoying

* start typing something
* press esc (you can even see the text cursor going back to the start)
* press backspace

would expect it to start deleting smth but instead it does nothing cuz cursor was reset

definitelynotrazu:
Another detail worth considering, is a way to activate it by pushing the cursor into one of the corners of the screen (with the corner being configurable). Most WMs/DEs trigger it via the top left corner, GNOME, Plasma, Niri. But having it configurable would be a nice bonus. If you add this I don't think you should make it exclusive to the overview, but make it something that can be used as a general bind people can configure freely.

vaxerski:
> text cursor position is reset when pressing esc out of the text box, which is annoying

kinda expected, escape unfocuses the text box so you can navigate the workspaces instead.

vaxerski:
> Another detail worth considering, is a way to activate it by pushing the cursor into one of the corners of the screen (with the corner being configurable). Most WMs/DEs trigger it via the top left corner, GNOME, Plasma, Niri. But having it configurable would be a nice bonus. If you add this I don't think you should make it exclusive to the overview, but make it something that can be used as a general bind people can configure freely.

out of scope for this mr

Vtorrealba:
> whats the point of that? You should never need to scour with your eyes over miniature images.

Bro look if you make an “over-view” you should be able to trigger it and, well…  “view over”. Otherwise is not really and overview and you’re wasting your time building it. 

Just make something like gnome and you’re good, although I can see how that feels uncomfortable for your philosophy. I personally thought that a overview would be a weird fit for hyprland fwiw

vaxerski:
ummm... why are you complaining when I literally agreed to it and already implemented it?

Vtorrealba:
> ummm... why are you complaining when I literally agreed to it and already implemented it?

My bad I replied to an old comment. All good

vaxerski:
I've updated the description to have an up-to-date screenshot

alba4k:
Some things I think are missing / would be cool:
* config option to toggle whether the search bar should be hidden when empty. It looks weird atm with the space being just empty when it's not there.
* Can't configure the color of the border around the previews, unless I missed smth
* Can't configure the shadow around the workspaces in the overview
* navigate using scrolling (horizontal swipe with two fingers, or dedicated mouse action. or maybe even just remapping normal scrolling), with it then snapping to the closest workspace or smth
* Switching to a new empty workspace while in the overview should probably focus it, not just create it while staying in the old one (?) maybe also deserves a config option

ryzendew:
Thanks for all the hard work i was just building my own overview in my DE but this is 100X better so far!

Dregu:
- Multi-monitor usage is weird, I can visually focus another window on a different monitor while overview is open, but all input still goes to overview
- Multi-monitor especially could use some initial workspace filters to only target workspaces from current monitor, only target non-visible workspaces etc. This could be generalized to accept any workspace selector, although we don't exactly have workspace selectors for the things I just said atm... Something like this anyway `overview.toggle({workspace = 'm[DP-1]'})`
- Option to follow monitor focus around and refresh the list of workspaces from aforementioned filter
- Follow focus and mouse cursor acts like it's still moving around the full-size desktop when e.g. hovering the thumbnail strip
- I use `mods+wasd` for all other directional navigation, need a way to bind nav keys for overview too, probably with a way to detect if overview is open to reuse some binds

CaptaiNiveau:
https://github.com/raybbian/hyprtasking

I use the above plugin with a 2D workspace layout, which also already implements an overview. Would it be possible to support a use case like that?

The 2D layout really makes a lot of sense imo

CaptaiNiveau:
https://private-user-images.githubusercontent.com/110075680/399425522-8d6cdfd2-2b17-4240-a117-1dbd2231ed4e.mp4?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODczMTg0MzMsIm5iZiI6MTc4NzMxODEzMywicGF0aCI6Ii8xMTAwNzU2ODAvMzk5NDI1NTIyLThkNmNkZmQyLTJiMTctNDI0MC1hMTE3LTFkYmQyMjMxZWQ0ZS5tcDQ_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwODIxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDgyMVQxMzE1MzNaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0xN2UxMTMwMTNlYWJiZTgxMmFlN2U2YTc4N2E5NzQwNzMwYTNlOWY2MTYzYzRlZjJiOWRlOGQ1ZTdhM2Q1ZWI3JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9dmlkZW8lMkZtcDQifQ.xZqE761OVPDU59uqx1MNQzF9oI9dijgkwjhSxKRoQRs

Linked the hyprtasking demo

rachalaraj:
> demo_small.mp4 
> Linked the hyprtasking demo

Whoa this is actually amazing and it's very useful imo

vaxerski:
most issues fixed

falkkyburz:
It is working very nicely now. 

I have some mouse pointer rendering issues. 
- Is it normal that the pointer hides after some seconds, my config doesn't have inactive timeout? sometimes it stays hidden after i start moving the mouse.
- There is some weird stuff happening when hovering/moving over the hidden text box.
- Pointer gets stuck in wrong icon, for example drag mode after dragging.

Bisect points to a8af239ab8bbf94b1ed42b54e609d4df9ffa3922

kolayne:
1. When in overview, I see my current workspace in the center and the two adjacent ones on the edges of the screen. If I mouseclick one of the adjacent ones, the overview exits and keeps me on the same workspace. I'd expect it to instead shift to the adjacent workspace.

2. When in overview, I can still use my touchpad gesture for switching between workspaces, but this only affects the window focus and not the overview state. I think it would make more sense if in the overview, my workspace-switching gesture was moving the observed overview workspace instead.

kolayne:
Also, as far as I can see, when I navigate between workspaces in the overview, the workspaces do not get focused until I exit the overview. Maybe this should be configurable: some users might want to use the overview mode to see what is going on in a window on another workspace, but some windows (e.g., the Telegram Desktop window with the chat list) won't update themselves when they know they are invisible.

falkkyburz:
Can confirm, cursor issue fixed.
Small cosmetic issue with more mini-previews than fit in width:
image

douglasfunster:
I think each layout should have its own overview cause for example this overview will not work on monocle or scrolling for example

vaxerski:
sometimes I wish I could punch people via TCP/IP

vaxerski:
rebased, most issues fixed. lmk

alba4k:
Gesture works great. Things I'd change:
* Using the `workspace` swipe gesture while in the overview moves them as it should but it's impossible to change workspace using that, you have to navigate using arrows or by clicking (see @kolayne's comment)
* using a keybind to switch to an other workspace while in the overview changes the window focus but doesn't affect the overview (would expect it to be focused there too). Why make the actually active workspace and the one currently previewed two different things? (see @kolayne's comment)
* Animation freaks out when switching to other workspace while the overview opening animation is playing
* ~Since we can show a neighbouring workspace by clicking on it, why can't we close the overview by clicking the currently focused one? (both with the big workspace and the previews on top). currently this is possible by right clicking anywhere on the overview~

vaxerski:
> Since we can show a neighbouring workspace by clicking on it, why can't we close the overview by clicking the currently focused one? (both with the big workspace and the previews on top). currently this is possible by right clicking anywhere on the overview

I thought "small fix" fixed that.

alba4k:
`build/Hyprland --version` says `edb7159ffe6d42e2915267933162227903635167 clean (small fix)` so idk what to tell you

vaxerski:
my bad, I'm not ai

alba4k:
thank fucking god you're not ai.

will test this evening and lyk

alba4k:
clicking works now btw

is this a "no" on the other ones?

vaxerski:
I fix select issues I find easier to fix, then I'll work on others. I'm already getting fucked by notifications in the past few days, yall apparently decided to just bombard me recently.

vaxerski:
so what issues are remaining so I can add them to the description?

alba4k:
> Things I'd change:
> * Using the `workspace` swipe gesture while in the overview moves them as it should but it's impossible to change workspace using that, you have to navigate using arrows or by clicking (see @kolayne's comment)
> * using a keybind to switch to an other workspace while in the overview changes the window focus but doesn't affect the overview (would expect it to be focused there too). Why make the actually active workspace and the one currently previewed two different things? (see @kolayne's comment)
> * Animation freaks out when switching to other workspace while the overview opening animation is playing

Dregu:
- Multi-monitor usage is weird, I can visually focus another window on a different monitor while overview is open, but all input still goes to overview
- Multi-monitor especially could use some initial workspace filters to only target workspaces from current monitor, only target non-visible workspaces etc. This could be generalized to accept any workspace selector, although we don't exactly have workspace selectors for the things I just said atm... Something like this anyway `overview.toggle({workspace = 'm[DP-1]'})`
  - The only_current_monitor global option is a start, but just weird, why is this not an option *in the dispatcher itself* so I could just bind both modes, now I have to make super weird binds
- Option to follow monitor focus around and refresh the list of workspaces from aforementioned filter
- I use `mods+wasd` for all other directional navigation, need a way to bind nav keys for overview too, probably with a way to detect if overview is open to reuse some binds. **I don't even have arrow keys**
- Very aggressively hooks a bunch of keys and blocks using them in all binds like `SUPER + Backspace/Escape`
  - Even breaks single key binds that are dont_inhibit, like my whole macropad just spams letters and numbers in the search box instead of running binds
- I can see through the corners of the thumbnails
- Shouldn't this thing use my decoration.rounding settings anyway

justkrysteq:
Also it would be great if dragging after the last workspace, like below, created a new one

Example

erstarr:
### Bugs

- Focus doesn't change to the workspace currently being viewed
This overview is already not very useful for scrolling as is but the ability to dispatch binds to the workspace while in overview mode is a good idea. It's just that the focus is held hostage by the workspace we open the overview in so i cannot scroll through other workspaces

### Suggestions


- Piggybacking on the prev point: Make the key to switch to the selected workspace configurable so we can actually use the workspace without having to switch to it, click on one thing, then re-open overview. For example: make only right click, or SUPER + right click, enlarge the selected workspace. All other binds, except left/right traversing on the overview  binds, pass through to the selected workspace

- Special workspaces. I get that they are not one workspace and layer over and whatnot, but an option to still show them? or at least make the window switching work for them. The first point is something i'm more or less just throwing against the wall, but the second point is important imo since it's degrading the capacity of the overview in acting like a fuzzy window finder/switcher

vaxerski:
> Focus doesn't change to the workspace currently being viewed

that's intentional. Especially on multimon, it wouldnt make much sense

erstarr:
> > Focus doesn't change to the workspace currently being viewed
> 
> that's intentional. Especially on multimon, it wouldnt make much sense

Than it'd be quite useless for scrolling workspaces, except to use as a fuzzy window switcher.

Consider that a very good amount of the people asking for an overview are scrolling users. Your comment about a niri-like overview is sitting at 12 thumbs down for example ;)

<br>

But well if we're being honest this gnome-like overview is never really going to be very useful for scrolling. 

I don't think there is such a thing other than a niri-like overview at this time.

CaptaiNiveau:
> But well if we're being honest this gnome-like overview is never really going to be very useful for scrolling.
> 
> I don't think there is such a thing other than a niri-like overview at this time.

Yeah, I agree. Maybe this can just be a separate overview, either built in or as plugin. It's similar to the 2D hyprtasking overview that already exists (linked above), though that plugin also has some wonkiness.

falkkyburz:
Have a crash with 6 or more workspaces in overview when navigating left/right quickly.


Crash Report: [hyprlandCrashReport2018187.txt](https://github.com/user-attachments/files/31388913/hyprlandCrashReport2018187.txt)
AI Summary: [hyprland-crash-summary.md](https://github.com/user-attachments/files/31389481/hyprland-crash-summary.md)

falkkyburz:
Two small visual issues:

Preview frame extends to screen edge:
screenshot-2026-08-24_23:00:48

The screenfreeze program shows nested overview somehow:
screenshot-2026-08-24_23:31:46

inifitrah:
**Suggestion: Jump mode**

It might be nice to have a "jump mode" for the overview, similar to MangoWC.

When entering jump mode, each visible window could get a small label above it, such as `H`, `J`, `K`, `L` (or configurable keys). Pressing a key would then immediately focus that window.

This would make it really quick to jump directly to a specific window without having to navigate through the overview or use the mouse, especially when there are many windows open.


screenshot_20260824_164830-region

vaxerski:
most requests done, anything else?

`hl.get_overview()` gives you state, move_left and right for moving. default query can also be passed.

erstarr:
consider giving overview the fullscreen treatment -- make it modular: make it layout-specific and make this the default one, so it's feasible/trivial to extend  it to include scrolling specific overview in the future without a massive rewrite.

I think in this form it'll be much easier/necessary to make a scrolling specific overview a layoutmsg, which is a lot of repeated code and is very ugly.

Dregu:
- Using predefined filter like `m[current]` is totally incompatible with also wanting to search things among the filtered results, since it just uses the same field but hidden. I end up searching for `foom[current]` when I want to find foot on active monitor
  - Not that being able to provide a predefined query as if it was typed in the search field is a bad feature, but add an additional `filter` field that can't be edited like a `query` can and results just go through both
  - Text cursor needs to default to the end of the input field in any case
  - Input field shouldn't be hidden by default if query was used
- When opening with query, things that got filtered out get animated out instead of just not being there in the first place, and everything else always slides in place weirdly

src/overview/hyprland/mode/WorkspaceQueryMode.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (query.starts_with("name:") || query.starts_with("special"))
+        return true;
+
+    if (!query.empty() && std::ranges::all_of(query, [](unsigned char c) { return std::isdigit(c); }))
+        return true;
+
+    if (query.size() < 2)
+        return false;
+
+    constexpr std::string_view SELECTORS = "rsnmwf";
```
Dregu, line 27: I don't think there's any reason to hardcode these specific characters. I was making some new workspace selectors and was a bit confused why they don't work on overview. I think checking `[1] == '['` is enough to either know it's a selector or user wrote some random garbage anyway.

vaxerski: I'll be doing a funny anyways

KristijanZic:
Does this allow for vertical workspaces?
I have one of those ultra wide monitors and having horizontal workspaces feels kinda off and mentally weird because the monitor is already so wide. So then I like to have vertical workspaces and shift between them with `alt+u` for down and `alt+i` for up.

falkkyburz:
I had a question regarding performance issues: 

- Is it possible to improve the performance or add a setting that reduces visual fidelity to improve performance?

On the internal display I have 0% GPU in idle with 10x WS (kitty) and overview open. When running sunshine/moonlight it jumps to around 41% with builtin display off and rendering to a virtual monitor at 60 FPS permanently. Then when I go to the overview the GPU usage goes to 100% and 14FPS (debug:overlay).
image

vaxerski:
it is a tad bit expensive but I'll see what I can do

unatried:
would it be possible to have the overview centered? like the one from windows but with workspaces instead of windows obv

falkkyburz:
> it is a tad bit expensive but I'll see what I can do

Disabling blur makes a big difference. This solved my issue:
```lua
function M.overview_toggle()
    hl.config({decoration = {blur = {enabled = hl.get_overview().open}}})
    hl.dispatch(hl.dsp.overview.toggle())
end
```

HeitorAugustoLN:
I was wondering if we could remove the minimap, like GNOME is [planning to do](https://gitlab.gnome.org/Teams/Design/os-mockups/-/raw/master/window-dnd/window-dnd.png)

PR #15899, opened 2026-08-19
Decouple pointer focus from layer-surface `keyboard_interactivity`

iciclejj:

[zwlr_layer_surface_v1::set_keyboard_interactivity](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_keyboard_interactivity)'s description states:

> Layer surfaces receive pointer, touch, and tablet events normally. If you do not want to receive them, set the input region on your surface to an empty region.

Layer surfaces currently treat [`keyboard_interactivity`](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity)  also as *pointer*-interactivity. Mainly:
- `onMap()` also forces pointer focus, not just keyboard focus, onto newly mapped `exclusive` (`keyboard_interactivity==exclusive`) layer-surfaces
- `mouseMoveUnified()` overrides pointer focus of any normal-focus-priority surfaces in favor of `exclusive` layer-surfaces, including workarounds for said behavior

Before this patch, that resulted in things like:
- If multiple `exclusive` surfaces were mapped, e.g. multi-output [slurp](https://github.com/emersion/slurp), pointer focus was always forced onto the most-recently-mapped surface, along with the keyboard. In other words, you might have to move your mouse after slurp launches to refresh the focus to the layer-surface your mouse is actually placed on, if the last-mapped surface didn't happen to be the one your cursor is currently on.
  - *Keyboard* focus target is explicitly ["implementation-defined"](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#:~:text=implementation%2D%20defined) in that scenario, so that part is fine.
- If there are **any** `exclusive` layer-surfaces mapped **anywhere**, on any output, your pointer will never be able to click or focus anything else.
  - Can be seen with basically any launcher (tofi, rofi, ...)

I also added a `simulateMouseMovement` call in onMap() to refocus the pointer. Programs like slurp were previously only accidentally grabbing pointer focus automatically at launch (i.e. without moving the mouse), due to having `exclusive` keyboard-interactivity. So this lets them grab focus automatically again, and at the same time fixing the pointer focus bug mentioned earlier with multiple `exclusive` surfaces.

I did not 100% finish investigating all the possible edge-case side-effects that the `simulateMouseMovement` call might introduce, but I was looking through a lot of the focus-related code in general, and it seems to already used it other places for virtually the exact same purpose, including the layer-surface's `onUnmap()`.

Someone who is more familiar with it should probably sanity-check the usage either way, though.

Other things:

- Lua config entry updated from `interactivity` to `keyboard_interactivity`
- https://github.com/hyprwm/Hyprland/commit/4072160677b574772a72d54445dbbca7fb4dcbc1: bugfix for some of the removed code, just for the record, in case the PR doesn't go through as is

Ready.

### DIFF
--- a/src/config/lua/objects/LuaLayerSurface.cpp
+++ b/src/config/lua/objects/LuaLayerSurface.cpp
@@ -67,8 +67,8 @@ static int layerSurfaceIndex(lua_State* L) {
         lua_pushboolean(L, ls->mapped());
     else if (key == "layer")
         lua_pushinteger(L, sc<lua_Integer>(ls->m_layer));
-    else if (key == "interactivity")
-        lua_pushinteger(L, sc<lua_Integer>(ls->m_interactivity));
+    else if (key == "keyboard_interactivity")
+        lua_pushinteger(L, sc<lua_Integer>(ls->m_keyboardInteractivity));
     else if (key == "above_fullscreen")
         lua_pushboolean(L, sc<bool>(ls->m_flags & Desktop::View::LAYER_FLAG_ABOVE_FULLSCREEN));
     else
--- a/src/desktop/state/FocusState.cpp
+++ b/src/desktop/state/FocusState.cpp
@@ -108,7 +108,7 @@ void CFocusState::rawWindowFocus(PHLWINDOW pWindow, eFocusReason reason, SP<CWLS
             return;
         }
 
-        if (!g_pInputManager->m_exclusiveLSes.empty()) {
+        if (!g_pInputManager->m_exclusiveKeyboardLSes.empty()) {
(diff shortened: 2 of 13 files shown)


iciclejj:
Sorry, forgot to repeat the AI disclosure. Same as in https://github.com/hyprwm/.github/discussions/31:
> I use AI for code review, and to speed up navigating/learning the codebase, finding state/code paths might be causing my issues, and similar.

vaxerski:
@codex review

vaxerski (approved):

PR #15916, opened 2026-08-20
input-capture: use libeis fd backend

argoneus:

CEis set up a socket backend, then getFileDescriptor() called
eis_backend_fd_add_client(), which asserts eis->backend. If socket
setup failed, that abort killed the compositor (SIGABRT) while
creating an input-capture session.

Use eis_setup_backend_fd() to match sendEisFd, and drop the context
on setup failure so getFileDescriptor() returns -1 instead of aborting.

This fixes a compatibility issue with lan-mouse that was resulting in a
consistent crash.

An AI assistant (Grok Build) was used to generate the fix, but was reviewed prior to
submission.  Tested locally to confirm the patch was successful.

Ready for merging.

### DIFF
--- a/src/managers/input/Eis.cpp
+++ b/src/managers/input/Eis.cpp
@@ -19,29 +19,21 @@ static const char* getClientName(eis_client* client) {
     return name ? name : "<unnamed>";
 }
 
-CEis::CEis(std::string socketName) {
-    Log::logger->log(Log::INFO, "[EIS] Init socket: {}", socketName);
-
-    const char* xdg = getenv("XDG_RUNTIME_DIR");
-    if (xdg)
-        m_socketPath = std::format("{}/{}", xdg, socketName);
-
-    if (m_socketPath.empty()) {
-        Log::logger->log(Log::ERR, "[EIS] Socket path is empty");
-        return;
-    }
-
+CEis::CEis() {
     m_eisCtx = eis_new(nullptr);
     if (!m_eisCtx) {
         Log::logger->log(Log::ERR, "[EIS] Cannot create eis context");
         return;
     }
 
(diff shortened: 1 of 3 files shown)


markallisongit:
Independent reproduction of this, in case a second data point helps it land.

Hyprland 0.56.2 (`efb5099`), libei 1.6.0, xdg-desktop-portal-hyprland 1.4.1, Arch, Intel/Mesa.

Same assert, same path:

```
Hyprland: ../libei/src/libeis-fd.c:78: eis_backend_fd_add_client: Assertion `eis->backend' failed.
```

```
#7  __assert_fail
#8  <libeis.so.1 + 0x47a5>
#9  CInputCaptureResource::CInputCaptureResource(...)
#10 CInputCaptureProtocol::onCreateSession(...)
#18 wl_event_loop_dispatch
#19 wl_display_run
#20 CEventLoopManager::enterLoop()
```

Trigger was lan-mouse 0.11.0 on `capture_backend = "input-capture-portal"` with an
intermittently unreachable peer, which churns sessions in a loop (create →
ConnectToEIS → GetZones → SetPointerBarriers → Enable → Disable → destroy → repeat).
It aborted while creating session `input-capture-758`, after 759 sessions over ~9h of
uptime — so it's probabilistic per session creation rather than deterministic, which
matches a setup failure that's only sometimes hit.

Ruled out resource exhaustion on my end: no OOM kills, `RLIMIT_NOFILE` was 524288 and
nowhere near exhausted.

Workaround for anyone else hitting this — `capture_backend = "layer-shell"` in the
lan-mouse config avoids `hyprland-input-capture-v1` entirely, and emulation uses a
separate backend so incoming input is unaffected.

Caveat: this is on the 0.56.2 release. I haven't built `main`, so I can't confirm
first-hand that the path is unchanged there.

PR #15929, opened 2026-08-21
config: Add `misc.suppress_initial_maximize_requests`

non-bin:
Fixes https://github.com/hyprwm/Hyprland/discussions/15901
Allows restoring previous behaviour that was changed by https://github.com/hyprwm/Hyprland/pull/15779

I chose true as the default value because that was the behaviour until recently

An alternative could be a window rule targeting unmapped windows which I got sort of working here #15931, but `suppress_event` is a static rule so that won't work unfortunately, and I can't think of any other dynamic rules we would want to apply only to mapped or unmapped windows, so I don't think it's worth adding.


wabarTM:
imo it would be better to move suppress_event from static to dynamic, instead of creating an option-that-works-when-mars-is-24.3-degrees-away-from-second-palm-tree-in-the-sunday-morning

non-bin:
I agree, do you want to give that a try? I'm not familiar enough with the project

wabarTM:
i:
1 - dont have a vouch
2 - 'm too lazy
3 - 'm working on wiki already

PR #15937, opened 2026-08-22
core/layers: Top layers go below fullscreen windows, fix DS solitary

erstarr:

Title
Fixes https://github.com/hyprwm/Hyprland/discussions/15881

Adds the config opt: `misc:allow_new_top_layers_over_existing_fullscreen`, which is self descriptive.

No

WIP

- [x] Notifications
- [ ] Test


vaxerski:
top layers over fs is expected. New ones should.

erstarr:
> top layers over fs is expected. New ones should.

https://github.com/hyprwm/Hyprland/discussions/15881 disagrees with you.

Besides, dark, ritz, etc... also think so. The reason being there isn't a layer that's below fullscreen but above normal windows. bars etc... use top in noctalia wayle etc.. and when you cycle them in fullscreen, they render over fullscreen

Dregu:
Yeah but nobody likes that. The protocol is stupid and there should be *an option* to not put any top layers above fullscreen. That's what overlay is for. So make this optional and everyone is happy.

LRitzdorf:
Yeah. Having timing-dependent behavior like we currently do just *feels* buggy, even if the spec says it's technically okay. Users go "oh not this bug again" and toggle fullscreen, which they really shouldn't have to do

darkwater:
The protocol doesn’t even prescribe that that should happen, it just describes that that’s what tends to happen.

It also means that if you don’t want notifications to pop up over fullscreen windows, there is no layer to place them.

Also when I ran into this I also saw it as a bug and went “welp gotta un-fs and fs my game again, which is annoying because it’s in hdr so it’ll blank out for a while”

e: As it stands now, there is no layer between tiled/floating windows and fullscreen windows.

erstarr:
Added a config opt to control this. Defaults to true, which is the old behaviour
 
`misc:allow_new_top_layers_over_existing_fullscreen`

erstarr:
I have absolutely no idea why the tests are failing in github CI.

Wait until layer spawn, wait until alpha is correctly set - all for naught. Layer tests are just not working on CI. Always passes in local.

vaxerski:
I don't like it. But, if we're gonna change this, just make all `top`s below fs. Don't add a knob.

Dregu:
Bug: Top layers that were never even visible still seem to block DS with "fadeout in progress" when closing under the fullscreen app. e.g. my bar normally shows up only when holding SUPER, this pr prevents that perfectly fine, but game goes out of ds for a fraction of a second when releasing SUPER and the bar closes.

> Don't add a knob.

Bit of a conflict here, if "new top layers go above fullscreen" is indeed how the protocol should be interpreted, why should there not even be an option to follow the protocol?

erstarr:
> Bug: Top layers that were never even visible still seem to block DS with "fadeout in progress" when closing under the fullscreen app. e.g. my bar normally shows up only when holding SUPER, this pr prevents that perfectly fine, but game goes out of ds for a fraction of a second when releasing SUPER and the bar closes.

From your other comments in other discussions, the regressor is https://github.com/hyprwm/Hyprland/pull/15991 - this MR or anything it changes have nothing to do with that.

Dregu:
My other comments elsewhere have nothing to do with layers fading. Without this PR the new top layer would obviously be visible on top of the game already, and the fadeout *should* block DS. It's definitely something this PR needs to account for when the fading layer isn't even visible.

LRitzdorf:
> > Don't add a knob.
>
> Bit of a conflict here, if "new top layers go above fullscreen" is indeed how the protocol should be interpreted, why should there not even be an option to follow the protocol?

FWIW, the protocol doesn't actually say which way this should go. All it really does is comment on how some compositors worked at the time of writing. Not to say that what other compositors do is totally irrelevant, but we certainly aren't violating the spec by saying "FS goes above the `top` layer always"

LRitzdorf:
> Bug: Top layers that were never even visible still seem to block DS with "fadeout in progress" when closing under the fullscreen app. e.g. my bar normally shows up only when holding SUPER, this pr prevents that perfectly fine, but game goes out of ds for a fraction of a second when releasing SUPER and the bar closes.

@Dregu Got a fix for this. Beforehand, *any* fadeouts whatsoever would block solitary (and thus DS), even ones on `background` or `bottom` which would never be above the FS window. Have sent patch to erstarr, will probably appear here soonish

Dregu:
If you say so, I did specifically only test an invisible (i.e. opened before fullscreen) top layer on main though and its fadeout did not block DS.

And yes the protocol specifically says "order is undefined" so we can do whatever we want with FS, screw all the shells that assume top might go above FS.

LRitzdorf:
> an invisible (i.e. opened before fullscreen) top layer on main though and its fadeout did not block DS.

Huh, I just checked that specific case on main, and it *did* block DS:

image

Well, in any case, it's fixed on this branch now!

FakeMichau:
Can confirm that this PR fixes the issue I was seeing in https://github.com/hyprwm/Hyprland/discussions/15881. 
But seems like allow_new_top_layers_over_existing_fullscreen has no effect. The fix seems to be always active regardless of the setting.

erstarr:
Really busy nowadays and don't have the time for this. I'll get to this as soon as I can (prob on the 18th or that weekend)

PR #15956, opened 2026-08-22
Fix some master layout window-drag bugs

LRitzdorf:

Fixes two bugs with window dragging in the master layout:
- `new_status = "master"` is incorrectly respected even while dragging, so windows can only be dropped as the new master
- With `new_on_top = true`, `drop_at_cursor = true` is ignored and always places the new window on top of the stack

Refs: #15920, #14830

Are these too edge-case-y, or do we want tests?

Ready, except for maybe those tests

### DIFF
--- a/hyprtester/src/tests/main/master.cpp
+++ b/hyprtester/src/tests/main/master.cpp
@@ -912,3 +912,104 @@ TEST_CASE(masterCenterDropAtCursor) {
     CALL_SUBTEST(expectCenterDrop, "R", false, "R", "M", "L");
     CALL_SUBTEST(expectCenterDrop, "R", true, "L", "M", "R");
 }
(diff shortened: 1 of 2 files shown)


vaxerski:
tests would be nice

LRitzdorf:
Yay, found more bugs from my tests, because master `addTarget` is weird

LRitzdorf:
Never mind, what I thought were bugs was in fact the layout being fancy about where it restores windows to in the stack. Tests fixed

LRitzdorf:
...or not; `layerPointerFocusPreservedOnKeyboardRefocus` failed. I don't think that's because of me, since it passes locally?

vaxerski (approved):

PR #15997, opened 2026-08-25
Cleaner incremental builds with the Makefile

LRitzdorf:

The Makefile now does semi-intelligent saving and checking of the options used during CMake's generation phase, so it only regenerates when options have actually changed. Previously, it'd regenerate no matter what, which took some extra time and led to unnecessary recompilation during the build phase.

Also cleaned up some minor stuff, mainly using a customizable `$(BUILDDIR)` rather than hard-coding `./build` everywhere.

Thanks @Aqa-Ib for pointing out that this was building extra stuff compared to VSCode/raw CMake!

Not really. Should just result in cleaner compiles via `make debug`/`make release`/etc.

Should be ready, but testing is appreciated since this would be a pretty bad thing to break

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,16 +3,24 @@ BUILDDIR ?= ./build
 CMAKE_BUILD_TYPE := Release
 CMAKE_ARGS = -DCMAKE_BUILD_TYPE:STRING=$(CMAKE_BUILD_TYPE) -DCMAKE_INSTALL_PREFIX:STRING=$(PREFIX)
 CMAKE_BUILDTYPE_FILE = $(BUILDDIR)/cmake_last_build_type
+CMAKE_GENERATE_CMD = cmake -Wno-unused-cli $(CMAKE_ARGS) -S . -B $(BUILDDIR)
 
+.DEFAULT: stub
 .PHONY: stub cmake_smartbuild release debug nopch clear all install uninstall pluginenv installheaders man asan format-check format-fix test
 
 stub:
 	@echo "Do not run $(MAKE) directly without any arguments. Please refer to the wiki on how to compile Hyprland."
 
-cmake_smartbuild:
+# Regenerate when CMakeLists have changed
+$(CMAKE_BUILDTYPE_FILE): $(shell find -type f -name CMakeLists.txt -not -path '*/_deps/*')
+	echo "$(CMAKE_BUILD_TYPE)+$(CMAKE_ARGS)" > $(CMAKE_BUILDTYPE_FILE)
+	$(CMAKE_GENERATE_CMD)
+
+# Regenerate when CMake options have changed, then build
+cmake_smartbuild: $(CMAKE_BUILDTYPE_FILE)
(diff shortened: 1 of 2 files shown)


LRitzdorf:
Actually, I think this also enables you to keep multiple builds cached, in different BUILDDIRs? I didn't really consider that while writing, but yeah, you can `make BUILDDIR=./build-rel release` without wiping out your debug build that lives elsewhere

erstarr:
No diff on vscode (using cmake directly). lgtm if generic make build works ok.

LRitzdorf:
Yeah, this literally only touches the Makefile, whose job is to drive CMake for us. CMake itself is totally unchanged, so testing in VSCode doesn't accomplish much

erstarr:
> Yeah, this literally only touches the Makefile, whose job is to drive CMake for us. CMake itself is totally unchanged, so testing in VSCode doesn't accomplish much

you want all cases tested just in case or risk getting shouted at if, by some miracle, you break building hyprland?

LRitzdorf:
Heh, fair point :)

LRitzdorf:
Even before this change, we do get the following error on specifically the *second* CMake build:
```
-- GLOB mismatch!
The following files were added:
  +<repo>/hyprctl/hw-protocols/hyprpaper_core-client.cpp
```

Not sure what that's about, other than stuff generated during the first build that's being caught by a wildcard during the second? I'm not a CMake guy, so I'm going to assume this is either fine, or somebody else's problem

LRitzdorf:
Never mind, this turned out to be an easy fix. The `hyprprotocol()` function already adds the files it generates as dependencies of the `hyprctl` target, so we can just... not glob for them. Confirmed that touching `hyprctl/hw-protocols/hyprpaper_core-client.cpp` still causes a rebuild and relink of hyprctl, so removing that glob didn't make us miss any dependencies

Aqa-Ib:
This works perfectly. Thank you for the work. Cheers

LRitzdorf:
A couple more changes:
- `make clean`, rather than `clear`, because the former is more standard
- We don't actually need the "scan for all CMakeLists" logic I'd added previously, CMake does that itself as long as `CMAKE_SUPPRESS_REGENERATION` isn't set
- Better test targets:
  - `make gtest` runs, well, gtests
  - `make hyprtest` runs hyrptester (what `make test` used to do)
  - `make test` runs both of the above

Also, a wiki PR to go with those new test targets, because they're contributor-facing: hyprwm/hyprland-wiki#1656

falkkyburz:
FYI: The `make uninstall` recipe leaves a hyprland symlink and empty directories behind. I noticed it when installing/uninstalling in /usr/local/. Not sure if it's worth fixing here.

LRitzdorf:
Not sure we can do much about empty dirs, but the symlink can be handled more elegantly. I want to check some stuff before I push, and cmake.org is currently down, but I'll have a thing soonish

LRitzdorf:
Okay, we now actually install the `hyprland` symlink via CMake so it's tracked in the manifest (and therefore uninstalled properly), and allow for uninstallation from `$DESTDIR`. Also rebased because why not

PR #16015, opened 2026-08-26
renderer: clear blur FBs on disabled blur

vaxerski:
A bit of a hack but I cba to have a proper self-allocating fb and then do shotgun surgery


tekstryder:
This is a really nice win even if hacky.

Dropped GPU mem from 239MiB to 171MiB.

On my trusty old Pascal chip with 4GB vram every MB counts.

PR #16018, opened 2026-08-26
keybinds: Per-device submaps

### LINKED ISSUE #15640: Per-device submaps
### Discussion link
https://github.com/hyprwm/Hyprland/discussions/14516

## Description

Add an extension to setting submap to set a submap per-device(s)

slopesque:
Closes #15640 

An implementation for per-device submap definitions. This PR thus proposes :

- A new optional parameter to `hl.define_submap` to accept metadata similar to what we do with `hl.bind`. So far, only the `device` key is handled.
- An architecture change promoting submaps to have their own data structure,  stored alongside `Keybinds::CBind`s in `Keybinds::CRegister`.

Just like bindings, if the submap has no device mapped to it, it becomes global. Otherwise, the submap will be set if and only if the last device used to press a key (in practice, the one that has the key pressed to trigger `hl.dsp.submap()`) is mapped by the submap.

Below is an example code snippet that can be added to an Hyprland test configuration to test the feature :

```lua
local mainMod = "ALT"

hl.on("keybinds.submap",
    function(submap)
        text = nil

        if submap == "" then
            text = "Returning to default"
        else
            text = "Enabling submap " .. submap
        end

        hl.notification.create({ text = text, timeout = 2500 })
    end
)

hl.bind(mainMod .. " + M", hl.dsp.submap("test"))

hl.define_submap("test",
    function()
        hl.bind(mainMod .. " + P",
            function()
                hl.notification.create({ text = "A special notification", timeout = 2000 })
            end
        )

        hl.bind(mainMod .. " + M", hl.dsp.submap("reset"))
    end,
    {
        device = {
            inclusive = true,
            list = { "wl_keyboard" }  -- or whatever is returned when running `hyprctl devices`
        }
    }
)

```

**`Keybinds::CSubset` Data Structure**

Right now (and if I understand correctly), we handle binds directly using two strings in `Keybinds::SBindMetadata` : `submap` and `submapReset`. When we want to check the submaps, we look for them in `CBind` objects directly (ex: [`Keybinds::CRegistry::hasSubmap`](https://github.com/hyprwm/Hyprland/blob/d10194c63b323e37c2ba5a4b954a7bbb12901e3b/src/keybinds/Registry.cpp#L54) ).

As we now allow metadata values for submaps which can be lead to evolve along with Hyprland's features, such an approach may not scale well on the long run. In order to handle device mappings, the PR thus introduces a dedicated data structure `Keybinds::CSubset` for submaps. The idea is to store them along with `Keybinds::CBind`s in the keybinds registry, where they are also reset when the keybinds are reset.

**`m_lastDevice` in `Config::Actions::CActionState`**

Changes I saved as commit `ddb4f88cbce700360f79f2bd2f1bf4dc49465604` gathers everything related to my current approach to pass on the current input device to the submap match search in `Config::Actions::setSubmap` : so far, I added a `SP<IHID> m_lastDevice` field to the global config action state, which I read during `setSubmap` .

I think there might be way, way better ways to do this other than straight up tracking the last device in the global action state. That being siad, I'm still very new to the codebase and this is the first option I have found. I kept this commit to at least have a functional first result, but please do tell if they are more efficient ways to pass around the input device across `Keybinds` callbacks.

**AI Usage**

As it is my first time working in Hyprland, I only relied on an AI model (Claude Sonnet in this case) for analysis purposes, in order to grasp an initial understanding of Hyprland's global structure and the project filesystem. Beyond that, I no longer used AI and favored understanding Hyprland's modules by reading the source code, testing its behavior using a debugger and checking out the documentation for the Lua C API. Moreover, I haven't used any AI while working on the feature itself.

The API and backend logic is ready and tested. The only missing part from the original feature request is to remove the global lock on submaps to allow multiple ones being active at the same time. Whether we implement it here or in a second PR is up to discussion.

The architecture change regarding `CSubmap` should also be approved before merging. If there is no problem with it, then we can change the existing logic to align it with the new API.

### DIFF
--- a/src/config/lua/bindings/LuaBindingsToplevel.cpp
+++ b/src/config/lua/bindings/LuaBindingsToplevel.cpp
@@ -209,24 +209,15 @@ static int hlBind(lua_State* L) {
         flags |= drag ? Keybinds::BIND_FLAG_DRAG : 0;
 
         lua_getfield(L, optsIdx, "device");
-        if (lua_istable(L, -1)) {
-            lua_getfield(L, -1, "inclusive");
-            const bool INCLUSIVE = lua_isnil(L, -1) ? true : lua_toboolean(L, -1);
-            flags |= INCLUSIVE ? Keybinds::BIND_FLAG_DEVICE_INCLUSIVE : 0;
-            lua_pop(L, 1);
 
-            lua_getfield(L, -1, "list");
-            if (lua_istable(L, -1)) {
-                lua_pushnil(L);
-                while (lua_next(L, -2)) {
-                    if (lua_isstring(L, -1))
-                        args.devices.emplace(lua_tostring(L, -1));
-                    lua_pop(L, 1);
-                }
-            }
-            lua_pop(L, 1);
-        }
+        auto devices = parseDeviceList(L, -1);
+        if (!devices)
+            return Internal::configError(L, std::format("hl.bind: {}", devices.error()));
+
+        args.devices = std::move(*devices);
+        flags |= args.devices.inclusive() ? Keybinds::BIND_FLAG_DEVICE_INCLUSIVE : 0;
         flags |= getBool("allow_input_capture") ? Keybinds::BIND_FLAG_ALLOW_INPUT_CAPTURE : 0;
+
         lua_pop(L, 1);
     }
 
--- a/src/config/lua/bindings/LuaBindingsToplevel.cpp
+++ b/src/config/lua/bindings/LuaBindingsToplevel.cpp
@@ -48,6 +48,28 @@ static std::expected<std::vector<std::string>, std::string> parseKeyString(std::
     return keys;
 }
 
+static std::expected<Keybinds::CDeviceList, std::string> parseDeviceList(lua_State* L, int idx) {
+    bool                 inclusive = false;
+    Keybinds::DeviceList devices;
+
+    if (lua_istable(L, idx)) {
+        lua_getfield(L, idx, "inclusive");
+        inclusive = lua_isnil(L, -1) || lua_toboolean(L, -1);
+        lua_pop(L, 1);
+
+        lua_getfield(L, idx, "list");
+        lua_pushnil(L);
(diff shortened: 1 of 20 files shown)


src/config/lua/bindings/LuaBindingsToplevel.cpp
```diff
@@ -251,11 +248,41 @@ static int hlDefineSubmap(lua_State* L) {
         fnIdx = 3;
     }
 
+    const int             optsIdx = fnIdx + 1;
+    Keybinds::SSubmapArgs args;
+
+    if (lua_gettop(L) >= optsIdx) {
```
vaxerski, line 254: since we are reusing this (Also for binds) it would make sense to unify those parsers to produce some shared CDeviceList from lua (CDeviceList::fromLua() or sumn)

src/config/shared/actions/ConfigActions.hpp
```diff
@@ -119,6 +120,7 @@ namespace Config::Actions {
         uint32_t    m_lastMouseCode       = 0;  // last mouse button code, 0 if last was keyboard
         uint32_t    m_timeLastMs          = 0;  // timestamp of last key/mouse event
         std::string m_currentSubmap       = ""; // current keybind submap name
+        SP<IHID>    m_lastDevice;               // device used to press last keycode
```
vaxerski, line 123: don't lock the device

tests/keybinds/Submap.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <gtest/gtest.h>
+
+using namespace Keybinds;
+
+static CSubmap makeSubmap(std::string name, std::unordered_set<std::string> devices, const bool inclusive) {
+    return CSubmap(name, {.devices = devices, .inclusive = inclusive});
+};
+
+// testing with actual devices requires being able to instantiate them
+// virtually :(
```
vaxerski, line 13: and you can, check the plugin in hyprtester. You can test it with integration tests

PR #16027, opened 2026-08-26
gestures/scrolling: add move invert option

rogervn:

Add a flag to invert the scroll-move gesture on the scrolling layout. 

This is a follow-up on https://github.com/hyprwm/Hyprland/pull/14952 where there was suggestions on how to proceed and the PR owner didn't have time to work on it for now, so I'm publishing an alternative PR pushing the invert flag.

Happy to do something else. This is a very small change and it was AI-assisted, but it does make sense and fit well in the original PR recommendation.

This was manually tested on a nested hyprland session with the invert config applyed and the scroll did invert as expected.

Ready for merging


erstarr:
one small problems with this, + a pre-existing bug here. I can send you the correct version and you can use that after you get your vouch, or i can fix it now. Which one would you like?

Also needs a wiki MR and a test.

rogervn:
> one small problems with this, + a pre-existing bug here. I can send you the correct version and you can use that after you get your vouch, or i can fix it now. Which one would you like?
> 
> Also needs a wiki MR and a test.

Sure, if you can send me the bug and problem I can have a look. For the test I can look into it, I've skipped because I found no test for move scroll, but IIUC it should be fine to add.

I can add it to the wiki too in another PR, as soon as I get the vouch.

PR #16036, opened 2026-08-27
setcursor applies immediately now

lukasz-rek:
Adds a call in updateTheme/CursorManager to immediately set the cursor after changing the buffer. Also saves previously set cursor name since CursorManager doesn't store it but it's needed to draw same cursor again. 

Issue I mean

https://github.com/user-attachments/assets/630d2527-5ea2-4d5f-9269-5b2508d7e991

I don't like that I duplicated where current cursor name is stored but it's a simple enough fix and since it's set every time it's changed it should follow 1:1. Might drift if someone also changes default cursor shape everywhere else in code, though I don't see why that might happen. I don't think rewriting lots of stuff for this small bandaid is worth it.

I've checked this works:
- when cursor is not moving (especially on startup)
- when cursor is not default state
- when cursor is hijacked by some other app (krita)

Ready

### DIFF
--- a/src/pointer/cursor/CursorManager.cpp
+++ b/src/pointer/cursor/CursorManager.cpp
@@ -170,8 +170,9 @@ void CCursorManager::setAnimationTimer(const int& frame, const int& delay) {
 void CCursorManager::setCursorFromName(const std::string& name) {
 
     static auto PUSEHYPRCURSOR = CConfigValue<Config::INTEGER>("cursor:enable_hyprcursor");
+    m_lastCursorName           = name;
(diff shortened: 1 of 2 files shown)


lukasz-rek:
LLM usage: I asked AI at the beginning to tell me if I was missing some cursor code when i started reading and then asked it at the end to tell me if I missed some side-effects with my fix (As AI does it glazed me but it nicely suggested I also check if it works when our buffer is not connected)

vaxerski (approved):
lgtm

PR #16037, opened 2026-08-27
groups: fixes for groupbar:disable_when_only

Aqa-Ib:
Fixes for `groupbar:disable_when_only`:
1. Fixes https://github.com/hyprwm/Hyprland/discussions/15820
2. Fixes the first member by giving it its groupbar after adding the second member, instead of giving it after being focused.
3. Fixes the remaining group member' size after removing other members from the group.

AI disclosure: I used the clanker for debugging and for C++ stuff.

Tested and everything seems good now.
CC @feelamee @tekstryder
Cheers

### DIFF
--- a/hyprtester/src/tests/main/groups.cpp
+++ b/hyprtester/src/tests/main/groups.cpp
@@ -590,6 +590,49 @@ TEST_CASE(groups_disable_when_only) {
     ASSERT(Tests::windowCount(), 0);
 }
 
+TEST_CASE(groupbarDisableWhenOnlyDimensions) {
+    OK(getFromSocket("/eval hl.config({ group = { groupbar = { disable_when_only = true } } })"));
+    SPAWN_KITTY("kittenA");
+    OK(getFromSocket("/dispatch hl.dsp.group.toggle()"));
+    {
+        auto str = getFromSocket("/activewindow");
+        EXPECT_CONTAINS(str, "at: 22,22");
+        EXPECT_CONTAINS(str, "size: 1876,1036");
+    }
+
+    SPAWN_KITTY("kittenB");
+    // when `groupbar:disable_when_only = true` and adding a second member to the group, check if the first member got its groupbar by checking its dimensions.
+    {
+        auto clients  = getFromSocket("/clients");
+        auto classPos = clients.find("class: kittenA");
+        if (classPos == std::string::npos) {
+            FAIL_TEST("Could not find kittenA in clients output");
+        } else {
+            auto entryStart  = clients.rfind("Window ", classPos);
+            auto entryEnd    = clients.find("\n\n", classPos);
+            auto windowEntry = clients.substr(entryStart, entryEnd - entryStart);
+            EXPECT_CONTAINS(windowEntry, "at: 22,43");
+            EXPECT_CONTAINS(windowEntry, "size: 1876,1015");
+        }
+    }
+
+    // also check the dimensions of the only member of the group after moving out the other member. It should have lost its groupbar and resized accordingly.
+    OK(getFromSocket("/dispatch hl.dsp.window.move({ out_of_group = true })"));
+    {
+        auto clients  = getFromSocket("/clients");
+        auto classPos = clients.find("class: kittenA");
+        if (classPos == std::string::npos) {
+            FAIL_TEST("Could not find kittenA in clients output");
+        } else {
+            auto entryStart  = clients.rfind("Window ", classPos);
+            auto entryEnd    = clients.find("\n\n", classPos);
+            auto windowEntry = clients.substr(entryStart, entryEnd - entryStart);
+            EXPECT_CONTAINS(windowEntry, "at: 22,22");
+            EXPECT_CONTAINS(windowEntry, "size: 931,1036");
+        }
+    }
+}
+
 TEST_CASE(autoGroupFloatedIntoTiled) {
     NLog::log("{}Test that we deny a new floated window getting auto-grouped into the focused tiled group.", Colors::GREEN);
 
--- a/src/desktop/view/Group.cpp
+++ b/src/desktop/view/Group.cpp
@@ -263,13 +263,13 @@ void CGroup::remove(PHLWINDOW w, Math::eDirection dir, eRemoveFromGroupReason re
     // we do it after the above because switchTargets expects this to be a valid group
     m_windows.erase(m_windows.begin() + *idx);
 
-    // recalculate the remaning group member for groupbar:disable_when_only
+    // if groupbar:disable_when_only is enabled and there is only one group member left, we need to fix its size because it will lose its groupbar.
     static auto PDISABLE = CConfigValue<Config::BOOL>("group:groupbar:disable_when_only");
     if (*PDISABLE && m_windows.size() == 1) {
-        const auto REMAININGMEMBER = m_windows.at(0).lock();
-        const auto GROUPBAR        = REMAININGMEMBER->presentation().decoration(DECORATION_GROUPBAR);
-        GROUPBAR->updateWindow(REMAININGMEMBER);
-        g_pDecorationPositioner->forceRecalcFor(REMAININGMEMBER);
+        const auto REMAINING_MEMBER = m_windows.at(0).lock();
+        const auto GROUPBAR         = REMAINING_MEMBER->presentation().decoration(DECORATION_GROUPBAR);
+        GROUPBAR->updateWindow(REMAINING_MEMBER);
+        g_pDecorationPositioner->forceRecalcFor(REMAINING_MEMBER);
     }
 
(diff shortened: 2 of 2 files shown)


tekstryder:
Thanks! Nice work. 

You solved the bigger issue I reported in https://github.com/hyprwm/Hyprland/discussions/15820.

No more overlap, and thus no more corrupted rendering therein.

One little niggle that still occurs (https://github.com/hyprwm/Hyprland/discussions/15820#discussioncomment-18117619):

>Notice the wack-ass tile sizes toward the end.

1) Open multiple windows on a workspace
2) Toggle group on and add a window to the group
3) `hl.dsp.window.move({ out_of_group = true })` on the original window which created the group
4) notice the tile size offset is still ~height of the now-gone groupbar



https://github.com/user-attachments/assets/9d844544-3b93-484f-a58d-8753ef724ce7

Aqa-Ib:
fixed

tekstryder:
Bingo. Can no longer repro either issue.

tekstryder:
Still all good after https://github.com/hyprwm/Hyprland/pull/16037/commits/b7265830d9565fe7666f95997bf0e06fd5c2d7cb.

And, on that note, also tested default grouping/un-grouping behavior without `disable_when_only` option.

Aqa-Ib:
Ready.

The test failing (`layerPointerFocusPreservedOnKeyboardRefocus`) is not related to this. It is also failing on other PRs.

src/desktop/view/Group.cpp
```diff
@@ -256,6 +263,15 @@ void CGroup::remove(PHLWINDOW w, Math::eDirection dir, eRemoveFromGroupReason re
     // we do it after the above because switchTargets expects this to be a valid group
     m_windows.erase(m_windows.begin() + *idx);
 
+    // if groupbar:disable_when_only is enabled and there is only one group member left, we need to fix its size because it will lose its groupbar.
+    static auto PDISABLE = CConfigValue<Config::BOOL>("group:groupbar:disable_when_only");
+    if (*PDISABLE && m_windows.size() == 1) {
+        const auto REMAINING_MEMBER = m_windows.at(0).lock();
+        const auto GROUPBAR         = REMAINING_MEMBER->presentation().decoration(DECORATION_GROUPBAR);
+        GROUPBAR->updateWindow(REMAINING_MEMBER);
```
Aqa-Ib: do we need to check here `if (REMAINING_MEMBER)` for any edge case?

erstarr: Should. `m_windows` holds `WP<>` so window can expire in the meantime. size counts the smart pointer, not its window reference.

Aqa-Ib: yeah in theory, but in practice i don't see where the edge case could come from.
We could take the overkill way just to be safe from weird edge cases. Do we also want to check `if (GROUPBAR)` just in case ? I have tested this without guards and no crashes, but the rare edge cases are beyond me.

erstarr: one if is small price to prevent a possible hard crash, and you don't know what might happen in the future with codebase changes. At least I think so.

Add remaining member exists check to the enclosing if

src/desktop/view/Group.cpp
```diff
@@ -256,6 +263,16 @@ void CGroup::remove(PHLWINDOW w, Math::eDirection dir, eRemoveFromGroupReason re
     // we do it after the above because switchTargets expects this to be a valid group
     m_windows.erase(m_windows.begin() + *idx);
 
+    // if groupbar:disable_when_only is enabled and there is only one group member left, we need to fix its size because it will lose its groupbar.
+    static auto PDISABLE = CConfigValue<Config::BOOL>("group:groupbar:disable_when_only");
+    if (*PDISABLE && m_windows.size() == 1) {
+        if (const auto REMAINING_MEMBER = m_windows.at(0).lock()) {
```
erstarr:
```suggestion
    if (const auto REMAINING_MEMBER = (*PDISABLE && m_windows.size() == 1) ?  m_windows.at(0).lock() : nullptr) {
```

Shorter

src/desktop/view/Group.cpp
```diff
@@ -95,7 +95,9 @@ bool CGroup::has(PHLWINDOW w) const {
 }
 
 void CGroup::add(PHLWINDOW w, std::optional<size_t> index) {
-    static auto INSERT_AFTER_CURRENT = CConfigValue<Config::INTEGER>("group:insert_after_current");
+    static auto INSERT_AFTER_CURRENT          = CConfigValue<Config::INTEGER>("group:insert_after_current");
+    static auto PDISABLE                      = CConfigValue<Config::BOOL>("group:groupbar:disable_when_only");
```
feelamee, line 99: shouldn't it also test `group:enable`? and is decorations visible at all?
Maybe replace this check with `CHyprGroupBarDecoration::visible`?
Same question with changes in `CGroup::remove` and `CGroup::applyWindowDecosAndUpdates`

src/desktop/view/Group.cpp
```diff
@@ -152,6 +154,11 @@ void CGroup::add(PHLWINDOW w, std::optional<size_t> index) {
     }
 
     applyWindowDecosAndUpdates(w);
+
+    // when groupbar:disable_when_only = true, give the only member of the group its groupbar after adding the second member.
+    if (GROUPBAR_DISABLED_ONLY_MEMBER)
```
feelamee, line 159: why declaring `GROUPBAR_DISABLED_ONLY_MEMBER` so far from its usage?
general practice is to define variables as close as possible to its usage

src/desktop/view/Group.cpp
```diff
@@ -256,6 +263,16 @@ void CGroup::remove(PHLWINDOW w, Math::eDirection dir, eRemoveFromGroupReason re
     // we do it after the above because switchTargets expects this to be a valid group
     m_windows.erase(m_windows.begin() + *idx);
 
+    // if groupbar:disable_when_only is enabled and there is only one group member left, we need to fix its size because it will lose its groupbar.
+    static auto PDISABLE = CConfigValue<Config::BOOL>("group:groupbar:disable_when_only");
```
feelamee, line 267: I dont really like how `disable_when_only` logic spreads across code base in different places.
Consider making something like: `onSizeChanged(size_t oldSize, size_t newSize)`. And do this checks and updates in it.

Also, if I understand correctly, problem appears at the scenario when groupbar changes visibility. So, why this is not handled by `CHyprGroupBarDecoration::draw`? It has similar checks and update.

hyprtester/src/tests/main/groups.cpp
```diff
@@ -590,6 +590,49 @@ TEST_CASE(groups_disable_when_only) {
     ASSERT(Tests::windowCount(), 0);
 }
 
+TEST_CASE(groupbarDisableWhenOnlyDimensions) {
```
feelamee, line 593: there is very similar test `groups_disable_when_only`, which I added initially.
Consider updating it, instead of creating new (and renaming it to camelCase)

Aqa-Ib: come on, this is fine...

src/desktop/view/Group.cpp
Aqa-Ib: > shouldn't it also test group:enable? and is decorations visible at all?
Maybe replace this check with CHyprGroupBarDecoration::visible?
Same question with changes in CGroup::remove and CGroup::applyWindowDecosAndUpdates

`CHyprGroupBarDecoration::visible` will be checked afterwards by `onWindowUpdate()` because we are calling `g_pDecorationPositioner->forceRecalcFor(GROUPBAR_DISABLED_ONLY_MEMBER);`.

src/desktop/view/Group.cpp
Aqa-Ib: Because we need to check if the member is the only one before adding the second member.

src/desktop/view/Group.cpp
Aqa-Ib: > I dont really like how disable_when_only logic spreads across code base in different places.

All of this stuff is only needed for `disable_when_only`.

> Consider making something like: onSizeChanged(size_t oldSize, size_t newSize). And do this checks and updates in it.

Feel free to do it that way if you wish.

> Also, if I understand correctly, problem appears at the scenario when groupbar changes visibility. So, why this is not handled by CHyprGroupBarDecoration::draw? It has similar checks and update.

There are 3 bugs fixed here:
1º) happens on adding a group member (thats why code is in CGroup::add).
2º) happens because we need to update the groupbar visibility before adding the groupbar, thats why it is in CGroup::applyWindowDecosAndUpdates.
3º) happens on removing a group member (thats why code is in CGroup::remove)

vaxerski (approved):

PR #16053, opened 2026-08-28
allow for negative scale on scroll move gesture

rogervn:

Allows for a negative scale in the gesture scroll. this will allow to invert the scrolling by passing a negative scale which today is restricted to 0.1 to 10.

This is a follow-up on the rejected https://github.com/hyprwm/Hyprland/pull/16027 which was a more complicated solution for something that we could already do with a negative scale.

This has been AI-assisted but I manually made changes I thought necessary and made sure the test added didn't break any existing test. It's a simple change anyway.

Ready for merging

### DIFF
--- a/hyprtester/src/tests/main/gestures.cpp
+++ b/hyprtester/src/tests/main/gestures.cpp
@@ -94,6 +94,27 @@ TEST_CASE(live_gesture_callbacks) {
     )"));
 
     OK(evalLua(R"(
+        __liveGesture.negative_scale_delta = nil
+
+        hl.gesture({
+            fingers = 9,
+            direction = 'left',
+            action = {
+                update = function(e)
+                    __liveGesture.negative_scale_delta = e.delta.x
+                end,
+            },
+            scale = -1,
+        })
+
+        hl.plugin.test.gesture('left', 9)
+
+        if __liveGesture.negative_scale_delta ~= 300 then
+            error('unexpected negatively scaled delta: ' .. tostring(__liveGesture.negative_scale_delta))
+        end
+    )"));
+
+    OK(evalLua(R"(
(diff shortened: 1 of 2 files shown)


erstarr:
AI disclosure... Anyway i can just make this fix if you can wait till tomorrow -- which you'll have to for the potential vouch in any case.

rogervn:
> AI disclosure... Anyway i can just make this fix if you can wait till tomorrow -- which you'll have to for the potential vouch in any case.

Yeah, sorry about that. I'm stil asking for the vouch as I intend to do similar contributions later, but if I don't get it this time I'm OK with you doing it if you don't mind.

erstarr:
You really have to read what the LLM writes before you plaster it on there tho; this is in the AI policy you agreed to twice by now:

`        if (deltaScale > -0.1F && deltaScale < 0.1F) {` only checks against (-0.1, 0.1) gap, the user can still give it -1000 or +1000

rogervn:
> You really have to read what the LLM writes before you plaster it on there tho; this is in the AI policy you agreed to twice by now:
> 
> ` if (deltaScale > -0.1F && deltaScale < 0.1F) {` only checks against (-0.1, 0.1) gap, the user can still give it -1000 or +1000

That was actually an error message I wrote on top of the error the LLM wrote :)

I believe the < -10 and > 10 ranges are already covered by `scaleParser(1.F, -10.F, 10.F)` . The added check is just because that checks the whole interval and we need to remove the -0.1 <-> 0.1 range

erstarr:
The error message and your check don't match then. Also you should print out the current delta value in the error message.

rogervn:
> The error message and your check don't match then. Also you should print out the current delta value in the error message.

No worries, let me publish a fixed message on my fork. If the vouch gets accepted I can reopen/create a new PR.

PR #16069, opened 2026-08-30
Core/Fullscreen: Improve `syncFullscreenTargets()`, account for an edge case, misc. improvements

erstarr:

Title.

The edge case is the possible non-determinism if there is more than one invalid entry in the FS targets list. At best this corrects itself, at worst this corrupts the list.

Also fixes a bug where the self correction lambda of isFullscreen() keeps triggering when you have 2+ FS windows on a scrolling workspace cuz it didn't account for isFullscreen() getting one and getFullscreen() getting another. 

Realistically neither of these has been a problem in both testing and in prod. but it's best to fix nonetheless.

Thanks @gulafaran for the suggestion of swap. That's a rotate, indeed. I'll need to profile to see if it's a good use case here.

Bugfix, refinement. Nothing breaking.

Ready

~~Profile new vs old sync to see perf gap is measurable.~~

### DIFF
--- a/src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
+++ b/src/layout/algorithm/tiled/scrolling/ScrollingFullscreenHandler.cpp
@@ -30,6 +30,12 @@ CScrollingFullscreenHandler::CScrollingFullscreenHandler(Layout::Tiled::CScrolli
 
 CScrollingFullscreenHandler::~CScrollingFullscreenHandler() {
 
+
+    // sift through any expired entries in case the following loop fucks us - read CScrollingFullscreenHandler::removeFsTarget()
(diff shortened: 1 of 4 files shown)


vaxerski:
conflict

erstarr:
> conflict

Done

vaxerski:
please add regression tests for the bugs this fixes

erstarr:
> please add regression tests for the bugs this fixes

fixes no specific bugs. Theoretically the edge case shouldn't happen to begin with as the entry is removed from the list upon unFS. This is only a 'self-correction edge-case'. If i could reproduce the need for the error correction, that would be an error itself and i'd have fixed that instead.

And the other part is performance improvement.

vaxerski (approved):
ok

PR #16072, opened 2026-08-30
allow for negative scale on scroll move gesture

rogervn:

Allows for a negative scale in the gesture scroll. this will allow to invert the scrolling by passing a negative scale which today is restricted to 0.1 to 10. The new range will be from -10 to -0.1 and from 0.1 to 10.

This is a follow-up for https://github.com/hyprwm/Hyprland/pull/16053 after vouch.

This has been AI-assisted but I manually made changes and applied follow-ups from the feedback from the other PR. I've also tested this manually on a nested hyprland calling `hl.plugin.test.gesture` with hyprctl with both a positive and a negative scale.

I'm still happy to take any further feedback if needed.

Ready for merging.

### DIFF
--- a/hyprtester/src/tests/main/gestures.cpp
+++ b/hyprtester/src/tests/main/gestures.cpp
@@ -180,6 +180,8 @@ TEST_CASE(live_gesture_callbacks) {
 
     EXPECT_CONTAINS(evalLua("hl.gesture({ fingers = 8, direction = 'up', action = { start = 1 } })"), "action.start must be a function");
     EXPECT_CONTAINS(evalLua("hl.gesture({ fingers = 8, direction = 'up', action = {} })"), "must define at least one of start, update, end, or finish");
+    EXPECT_CONTAINS(evalLua("hl.gesture({ fingers = 8, direction = 'up', action = function() end, scale = 0 })"),
+                    "value must be between -10 and -0.1 or between 0.1 and 10 - It's currently: 0");
 }
 
 // TODO: decompose this into multiple test cases
--- a/src/config/lua/bindings/LuaBindingsConfigRules.cpp
+++ b/src/config/lua/bindings/LuaBindingsConfigRules.cpp
@@ -898,13 +898,19 @@ static int hlGesture(lua_State* L) {
     float deltaScale = 1.F;
     lua_getfield(L, 1, "scale");
     if (!lua_isnil(L, -1)) {
-        CLuaConfigFloat scaleParser(1.F, 0.1F, 10.F);
+        CLuaConfigFloat scaleParser(1.F, -10.F, 10.F);
         auto            scaleErr = scaleParser.parse(L);
         if (scaleErr.errorCode != PARSE_ERROR_OK) {
             lua_pop(L, 1);
             return Internal::configError(L, std::format("hl.gesture: field \"scale\": {}", scaleErr.message));
         }
         deltaScale = scaleParser.parsed();
+        // must be [-10.F, -0.1F] ∪ [0.1F, 10.F]
+        // < -10 and >10 checks are made above
+        if (deltaScale > -0.1F && deltaScale < 0.1F) {
+            lua_pop(L, 1);
+            return Internal::configError(L, "hl.gesture: field \"scale\": value must be between -10 and -0.1 or between 0.1 and 10 - It's currently: {}", deltaScale);
+        }
     }
     lua_pop(L, 1);
 
--- a/hyprtester/src/tests/main/scroll.cpp
+++ b/hyprtester/src/tests/main/scroll.cpp
@@ -2391,6 +2391,50 @@ TEST_CASE(scroll_DEFAULT_HANDLED_FullscreenNonInterference) {
 
 /* Scroll viewport tests */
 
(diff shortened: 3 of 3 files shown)


src/config/lua/bindings/LuaBindingsConfigRules.cpp
```diff
@@ -898,13 +898,17 @@ static int hlGesture(lua_State* L) {
...
     if (!lua_isnil(L, -1)) {
-        CLuaConfigFloat scaleParser(1.F, 0.1F, 10.F);
+        CLuaConfigFloat scaleParser(1.F, -10.F, 10.F);
         auto            scaleErr = scaleParser.parse(L);
         if (scaleErr.errorCode != PARSE_ERROR_OK) {
             lua_pop(L, 1);
             return Internal::configError(L, std::format("hl.gesture: field \"scale\": {}", scaleErr.message));
         }
         deltaScale = scaleParser.parsed();
+        if (deltaScale > -0.1F && deltaScale < 0.1F) {
```
erstarr, line 910:
```suggestion
        // must be [-10.F, -1.F] ∪ [1.F, 10.F]
        // < -10 and >10 checks are made above
        if (deltaScale > -0.1F && deltaScale < 0.1F) {
```

hyprtester/src/tests/main/gestures.cpp
```diff
@@ -93,6 +93,27 @@ TEST_CASE(live_gesture_callbacks) {
...
+            scale = -1,
+        })
+
+        hl.plugin.test.gesture('left', 9)
+
+        if __liveGesture.negative_scale_delta ~= -300 then
+            error('unexpected negatively scaled delta: ' .. tostring(__liveGesture.negative_scale_delta))
+        end
+    )"));
+
```
erstarr:
```suggestion

    // test negative delta when moving viewport in scrolling via gesture -- negative delta works at all
    OK(evalLua(R"(
        __liveGesture.negative_scale_delta = nil

        hl.gesture({
            fingers = 9,
            direction = 'left',
            action = {
                update = function(e)
                    __liveGesture.negative_scale_delta = e.delta -- might as well check box axes
                end,
            },
            scale = -1,
        })

        hl.plugin.test.gesture('left', 9)

        if __liveGesture.negative_scale_delta.x ~= 300 then
            error('unexpected negatively scaled delta: ' .. tostring(__liveGesture.negative_scale_delta.x))
        end
        if __liveGesture.negative_scale_delta.y ~= 0 then
            error('unexpected negatively scaled delta: ' .. tostring(__liveGesture.negative_scale_delta.y) .. ' - It moves the y axis even though it should not')
        end
    )"));

    // test negative delta when moving viewport in scrolling via gesture -- test scaling
    OK(evalLua(R"(

        __liveGesture.scale2_delta = nil

        hl.gesture({
            fingers = 10,
            direction = 'left',
            action = {
                update = function(e)
                    __liveGesture.scale2_delta = e.delta
                end,
            },
            scale = -2,
        })

        hl.plugin.test.gesture('left', 10)


        if __liveGesture.scale2_delta.x ~= 600 then
            error('unexpected negatively scaled delta: ' .. tostring(__liveGesture.scale2_delta.x))
        end
        if __liveGesture.scale2_delta.y ~= 0 then
            error('unexpected negatively scaled delta: ' .. tostring(__liveGesture.scale2_delta.y) .. ' - It moves the y axis even though it should not')
        end


    )"));

```

src/config/lua/bindings/LuaBindingsConfigRules.cpp
```diff
@@ -898,13 +898,17 @@ static int hlGesture(lua_State* L) {
...
+            lua_pop(L, 1);
+            return Internal::configError(L, "hl.gesture: field \"scale\": {} can't be between -0.1 and 0.1", deltaScale);
```
erstarr:
```suggestion
            return Internal::configError(L, "hl.gesture: field \"scale\": value must be between -10 and -0.1 or between 0.1 and 10 - It's currently: {}", deltaScale);
```

erstarr (changes requested):

rogervn:
Sorry for the dealy, I found out that the tests were not covering the scale as it was checking the gestures and not the scroll, so I had to go back to the drawing board for the tests and had AI help me on creating another test on the scroll tests. I believe we've covered both the scale and size.

It's not as simple, but I think that it does covers it now and it makes sense, but I'm happy to revisit it if I was misled into that by the LLM.

erstarr:
> tests were not covering the scale as it was checking the gestures and not the scroll

what does this mean?

rogervn:
> > tests were not covering the scale as it was checking the gestures and not the scroll
> 
> what does this mean?

It means that the value it was measuring was due to the raw gestures regardless of the scale (positive or negative) which applies after that point. I was confused that the value was not changing although the manual test did mirror it.

The alternative was creating a new test on the scroll tests.

rogervn:
The test had a small issue with the configured gesture. Fixed it and made it more legible and made sure it's succeeding now.

erstarr (approved):
lgtm

vaxerski (approved):

PR #16108, opened 2026-09-02
Add `reload_config()` dispatcher

LRitzdorf:
An `hl.dsp.reload_config()` dispatcher, because shelling out to `hyprctl reload` only to then do IPC back to Hyprland feels awkward.

Wiki PR: hyprwm/hyprland-wiki#1646

Nothing super complicated here. The only thing is, we use `doLater` to not tear down config state while it's in active use, but that's pretty routine.

Tested locally, and it reloads properly (wiping out extra Lua variables, for example) without crashing or anything.

### DIFF
--- a/src/config/shared/actions/ConfigActions.cpp
+++ b/src/config/shared/actions/ConfigActions.cpp
@@ -1221,6 +1221,8 @@ ActionResult Actions::exit() {
 }
 
 ActionResult Actions::reloadConfig() {
+    if (!Config::mgr()->configLoaded())
+        return std::unexpected(std::string("Cannot trigger a reload while the config is already loading!"));
     // probably don't tear down state while stuff is actively running
     g_pEventLoopManager->doLater([] { Config::mgr()->reload(); });
     return {};
--- a/src/config/shared/actions/ConfigActions.cpp
+++ b/src/config/shared/actions/ConfigActions.cpp
@@ -1,5 +1,6 @@
 #include "ConfigActions.hpp"
 #include "../parserUtils/ParserUtils.hpp"
+#include "../../ConfigManager.hpp"
 #include "../../../desktop/state/FocusState.hpp"
 #include "../../../desktop/state/GlobalWindowController.hpp"
 #include "../../../desktop/state/WindowState.hpp"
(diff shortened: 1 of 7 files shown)


src/config/shared/actions/ConfigActions.cpp
```diff
@@ -1218,6 +1220,12 @@ ActionResult Actions::exit() {
     return {};
 }
 
+ActionResult Actions::reloadConfig() {
+    // probably don't tear down state while stuff is actively running
```
vaxerski, line 1226: very good.

vaxerski (commented):
- Needs a test
- Needs to throw an error if reload_config() is called from the global scope (otherwise it will reload forever) - exceptions apply with repl / eval

LRitzdorf:
For the infinite reload thing, I don't thiiiink we can really do Lua scope detection, but a better way IMO is to just make the dispatch fail if the config isn't done being loaded yet. That was easy enough to implement, didn't need carveouts for eval/repl, and even saves you from reloading in the `config.reloaded` handler if you somehow decide that's a good life choice.

What were you thinking in the way of tests? A basic "set var, reload, make sure var got wiped," or something fancier to hit edge cases like the above? If the latter, we'd kinda need a new config file for just that test, which sounds like pain to set up with hyprtester...

LRitzdorf:
Dang layer-pointer-focus test...

vaxerski:
> For the infinite reload thing, I don't thiiiink we can really do Lua scope detection, but a better way IMO is to just make the dispatch fail if the config isn't done being loaded yet. That was easy enough to implement, didn't need carveouts for eval/repl, and even saves you from reloading in the config.reloaded handler if you somehow decide that's a good life choice.

you dont need to, simply pop an error if it's called while the config is being parsed from the main ::reload()

> What were you thinking in the way of tests? A basic "set var, reload, make sure var got wiped," or something fancier to hit edge cases like the above? If the latter, we'd kinda need a new config file for just that test, which sounds like pain to set up with hyprtester...

the first is enough I think

LRitzdorf:
> simply pop an error if it's called while the config is being parsed from the main ::reload()

That's basically what I've done. The ConfigManager interface now includes a `configLoaded()` method, which for Lua just returns false if it's currently parsing (i.e. we're in `::reload()`). Dispatching `reload_config()` checks that and pops an error if it's false.

Also, the test is in here now, so this should be mergeable whenever.

vaxerski (approved):

PR #15958, opened 2026-08-22
monitor: solitary cannot have an underlying subsurface

vaxerski:
ref #15923

PR #16131, opened 2026-09-03
monitor: add check for empty buffer before denying Solitary

LionHeartP:
After #15958 and the release of GE-Proton 11-6 which fixes Steam overlay in wine-wayland, Solitary is denied for Proton because of an empty surface that is created.

This PR adds a check for the subsurface buffer being empty before denying Solitary and allows Direct Scanout to engage. When the, now working, Steam Overlay is brought up, Hyprland correctly blocks Solitary and Direct Scanout until the Overlay is closed.

Also setup a quick Plex server to test with Plezy, and verified Solitary is blocked when playing a video so there's no regression there.

It's possible this doesn't deny Solitary in an edge case I can't think of atm...

Ready.

### DIFF
--- a/src/render/shaders/glsl/blurFinish.glsl
+++ b/src/render/shaders/glsl/blurFinish.glsl
@@ -21,16 +21,25 @@ vec4 blurFinish(vec4 pixColor, vec2 v_texcoord, float noise, float brightness
                 int sourceTF, int targetTF, mat3 convertMatrix, vec2 srcTFRange, vec2 dstTFRange
 #endif
 ) {
+    float brightnessMultiplier = min(1.0, brightness);
+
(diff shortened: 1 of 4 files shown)


Dregu:
This works super randomly for some proton games with *DS on*, randomly getting black screen and blocked by "surface transformations", but even worse fullscreen proton games *don't work with DS off at all*?

LionHeartP:
> This works super randomly for some proton games with _DS on_, randomly getting black screen and blocked by "surface transformations", but even worse fullscreen proton games _don't work with DS off at all_?

Bummer, didn't think of testing with DS=0 :P. Let me draft this and look at it again over the weekend. Thanks

PR #16140, opened 2026-09-04
workspace: refactor workspace handling

vaxerski:
Major refactors for workspace handling, allowing for more unit tests and less clusterfuck overall.

Should also fix a few bugs.

Removes IDs as globally exposed properties in favor of addressables.

name: no longer necessary for named workspaces.

Assisted by codex.

NEEDS WIKIIIII
- remove `name:` requirement
- workspaces are always referred to as stable identities e.g. `hello` renamed to `world` is still `hello` in rules (renaming is only the display name doesnt change the identity)


Kam1k4dze:
\>named workspace
\>looks inside
\>no name

image

vaxerski:
wut? They still have an addressable and a display name

spikespaz:
I don't understand whatever your intent is, but all of my workspaces are indexed and I don't use names.

vaxerski:
is that a comment to me or kamikadze?

Kam1k4dze:
My comment was mostly a joke about `name: no longer necessary for named workspaces.`
After looking at the code, I see why it's now like this, it just sounded funny to me.

vaxerski:
I mean it's more user friendly to just call a workspace `sex` instead of `name:sex`

spikespaz:
It was a comment on the change.
How do I refer to workspaces by numeric index in order they're populated with windows?

vaxerski:
lua

spikespaz:
> lua

No thanks, please let me keep using static config

You went to all that effort to make hyprlang when nobody asked

vaxerski:
since when was that possible with a static config?

kolayne:
Waybar does not see any workspaces besides 1.

vaxerski:
@kolayne does seem so, but I can't see a difference in s2 output from git vs this MR.

`ext/workspaces` works fine on waybar.

LRitzdorf:
> How do I refer to workspaces by numeric index in order they're populated with windows?

@spikespaz You can use a number as the name. Like, if you windowrule some program over to workspace 3, you can still focus it with `hl.dsp.focus({ workspace = 3 })` (presumably ints get `tostring`'d for compatibility). Glancing at the diff, workspace selectors like `next` and `previous` look like they've been updated to traverse numerically-named workspaces in order too. So you should be good, I think?

Edit: also, you do know hyprlang is going away in 0.57, right?

vaxerski:
nice

PR #16150, opened 2026-09-05
layout/floating: fix "move" rule not using the size provided by the "size" rule

sitiyou:

Discussed in #13409

fixed a regression introduced by 723870337f299d4c80aa0048144f3a1ca1c885bd where the computed size was used to update realPosition

```cpp
// before src/desktop/view/Window.cpp
const auto COMPUTED = calculateExpression(m_ruleApplicator->static_.size);
// ...
*m_realSize = *COMPUTED;

// regression src/layout/algorithm/floating/default/DefaultFloatingAlgorithm.cpp
const auto COMPUTED = WINDOW->calculateExpression(*WINDOW->m_ruleApplicator->static_.size);
// ...
WINDOW->m_realPosition->setValueAndWarp(*COMPUTED);
```

e7996dc9 renames `m_realPosition/m_realSize` to `positionAnimation()/sizeAnimation()` with no functional change

### DIFF
--- a/hyprtester/src/tests/main/window.cpp
+++ b/hyprtester/src/tests/main/window.cpp
@@ -2182,3 +2182,26 @@ TEST_CASE(windowRuleBorderColorsFocus) {
 
     Tests::killAllWindows();
 }
+
+TEST_CASE(floatingMoveExpression) {
+    OK(getFromSocket("/eval hl.window_rule({ match = { class = 'kitty_floatingmove' }, "
+                     "float = true, size = { 800, 600 }, "
+                     "move = { '(monitor_w-window_w-10)', '(monitor_h-window_h-10)' } })"));
+
+    SPAWN_KITTY("kitty_floatingmove");
+
+    Tests::waitUntilWindowsN(1);
+
+    const auto EXPECTED_AT = getFromSocket("/repl local m = hl.get_active_monitor();"
+                                           "return (m.x + m.width - 800 - 10) .. ',' "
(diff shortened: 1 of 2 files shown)


vaxerski (dismissed):

vaxerski:
can we get a regression test here?

sitiyou:
> can we get a regression test here?

Done.

Without a final `/reload` the CI randomly fails, idk why.

ItsOhen:
> > can we get a regression test here?
> 
> Done.
> 
> Without a final `/reload` the CI randomly fails, idk why.

May i ask why you need the new headless output when you don't check what output the window is on?

sitiyou:
> > > can we get a regression test here?
> > 
> > 
> > Done.
> > Without a final `/reload` the CI randomly fails, idk why.
> 
> May i ask why you need the new headless output when you don't check what output the window is on?

I need to verify the window's position and size, so I need an output with a deterministic position and resolution to test against.

vaxerski:
you can query the monitor's pos and size and parse that.

Also conflict

sitiyou:
> you can query the monitor's pos and size and parse that.
> 
> Also conflict

Done.

the ci still fails without a `/reload`: https://github.com/sitiyou/Hyprland/actions/runs/34621430520/job/103336092342
I see the same failure pattern in another PR: https://github.com/hyprwm/Hyprland/actions/runs/34618154450/job/103325139920?pr=16227

seems to be caused by some unkown side effects.

vaxerski (approved):
lgtv

PR #16151, opened 2026-09-05
gestures: add support for using mouse  as a delta for gestures

### LINKED ISSUE #14509: Allow trackpad gestures to take button holds
### Discussion 

https://github.com/hyprwm/Hyprland/discussions/14438

## Description

Gestures could accept e.g. mouse or keyboard buttons, then mouse movement as the source of the delta.

rogervn:

Adds support for a "button" configuration for gestures to allow for holding a mouse button and use the mouse movement (or other pointers) for the gesture delta.

This is supposed to close https://github.com/hyprwm/Hyprland/issues/14509 

AI was used to write the majority of code, but always with human-in-the-loop:
1. I presented the problem and discussed potential solutions with the LLM, asking it to explain me the codebase and why that's a good solution.
2. I asked it to prototype the solution I understand best, at which I do a brief reading of the code and test it E2E just to validate it worked.
3. I did a deep read of the code, asking clarification on the codebase and using the LLM to the level of telling it what functions to write and variable names so the code is legible and removing potential hallucinations. I might do manual changes here as well, but it's usually easier to tell the LLM exactly what to write so it runs the formatter, testing and update cross-file references.
4. I did the same with the unit test, making sure it validates my understanding of the code and that it works.

I'm flagging that this needs feedback before being ready for merging because:

1. The configuration of having button mutually exclusive with finger is a design decision that would benefit from feedback by someone with a broader design vision for the configs.
5. I'm still learning the hyprland codebase and although AI has helped quite a bit understand how to make this possible, it could be short-sighted on some parts of the code where it's not the optimal place to put that and I don't have the full context yet to tell it apart.

This is the currently proposed design config:
```
hl.gesture({
    button = "mouse:274",
    mods = "SUPER",
    direction = "horizontal",
    action = "scroll_move",
})
```

It does need feedback before considered ready for merging due to the explained above about the configuration design and the layers where the implementation was done. It was tested manually in a nested hyprland session and it does work both on horizontal and vertical scrolling.

If that's all OK I'll publish a wiki PR as well and it will be ready for merging. I'm happy to get feedback here and redesign if necessary.


erstarr:
To what extent was AI involved? i.e. did it write practically all the code? did it also construct the logic?

AI-Assisted is too broad.

rogervn:
> To what extent was AI involved? i.e. did it write practically all the code? did it also construct the logic?
> 
> AI-Assisted is too broad.

Sorry, I used it in a few iterations:
1. I presented the problem and discussed potential solutions with the LLM, asking it to explain me the codebase and why that's a good solution.
2. I asked it to prototype the solution I understand best, at which I do a brief reading of the code and test it E2E just to validate it worked.
3. I did a deep read of the code, asking clarification on the codebase and using the LLM to the level of telling it what functions to write and variable names so the code is legible and removing potential hallucinations. I might do manual changes here as well, but it's usually easier to tell the LLM exactly what to write so it runs the formatter, testing and update cross-file references.
4. I did the same with the unit test, making sure it validates my understanding of the code and that it works.

3 and 4 are usually where I spend the most time (especially 4 as for some reason I didn't manage to catch yet I have many times cases the hyprtester VM tests work locally but fail in CI).

So yes, the AI proposed the logic and wrote the majority of the code, but nothing was done without a human-in-the-loop that did not understand the changes, asked explanations and ordered deep code changes where the LLM was acting as a glorified IDE.

As I ramp up on the code I'll even be more intentional with the changes and the LLM becomes much more of a glorified IDE and prototyping tool. I also expect to do another round of polishing after the design is agreed, so the review at this stage can be very coarse just to make sure I didn't overfocus on the wrong side of the codebase.

hyprtester/plugin/src/main.cpp
```diff
@@ -612,6 +612,45 @@ static SDispatchResult click(std::string in) {
     return {};
 }
 
+static SDispatchResult mouseMove(std::string in) {
+    CVarList2 data(std::move(in));
+
+    double    x, y;
+    try {
+        x = std::stod(std::string{data[0]});
+        y = std::stod(std::string{data[1]});
```
vaxerski, line 621: normally I wouldn't berate because these are tests, but if you want to contribute in the future, strToNumber from hu/string/numeric.

src/config/lua/bindings/LuaBindingsConfigRules.cpp
```diff
@@ -743,12 +743,37 @@ static int hlGesture(lua_State* L) {
...
+        if (fingersErr.errorCode != PARSE_ERROR_OK)
+            return Internal::configError(L, std::format("hl.gesture: {}", fingersErr.message));
+        fingerCount = fingersParser.parsed();
+    } else {
+        CLuaConfigString buttonParser("");
+        auto             buttonErr = Internal::parseTableField(L, 1, "button", buttonParser);
+        if (buttonErr.errorCode != PARSE_ERROR_OK)
+            return Internal::configError(L, std::format("hl.gesture: {}", buttonErr.message));
+        const auto& button = buttonParser.parsed();
+        if (!button.starts_with("mouse:") || button.size() == 6 || !std::ranges::all_of(button.substr(6), [](unsigned char c) { return std::isdigit(c); }))
```
vaxerski, line 769: if we're already doing this, it makes no sense to restrict this to `mouse:`. You could instead require a `source = ` for button cases (e.g. `source = "mouse"`) in which case mouse is the source of the delta.

E.g. I press SUPER + X and can drag a window around.

This is already the case with mouse binds (which would kinda become obsolete with this since we have resize and move gestures)

rogervn: Thanks, that makes sense!

What other options do you thing we could put on source? Like "finger" and "mouse" with "finger" as the default for compatibility, or split pointers types?  That might make it confusing though as I do have a few trackpads that identify themselves as "mouse". That's why I was using pointers on the implementation instead of mouse.

In any way this is currently working as a more informational/validation setting as for the code it's just a matter of branching and storing the number of fingers or the keybind.

Dregu: It has also been requested and it would make sense to support `finger` (and `continuous`) *scroll* events in live gestures and/or binds, currently there's no way to bind touchpad scroll, only `wheel` scroll. Two finger scroll is essentially a two finger gesture, but the driver always turns it into scroll events. Two finger scroll *can* be disabled in hyprland, but that still doesn't bring us two finger gestures.

https://wayland.app/protocols/wayland#wl_pointer:enum:axis_source

rogervn: Nice! I've expanded it a bit more then to allow for the axis sources.

As the first version I used AI to write the code, but I did the same process of discussing the architecture, testing it manually on a nested session then carefully reviewing the code for better improvements and iterating on simplifications/corrections on the code level.

I've also pushed changes to the wiki on https://github.com/hyprwm/hyprland-wiki/pull/1659 as well.

Happy to take any feedback on the code and rewrite as needed or rethink part of the design, but I like now how it reflects the axis source types.

kolayne:
Seems like something went wrong with PR atomicity. Not sure at which point that happened, but with commits pushed now by vaxerski (?), this PR includes changes to the gesture system, changes to the workspace addressing, and even changes to the github pull request description template (that's what I saw at the first glance).

This complicates the review, as well as identifying sources of regressions with `git bisect`.

erstarr:
Yeah something git related was done wrong
image

rogervn:
Oh, sorry, I did feel something wrong whilst pushing and rebasing this, let me try to fix it.

rogervn:
I'm still running the tests, will be ready for review in a bit.

erstarr:
Ok so some considerations about this MR as a whole:

> I'm still running the tests, will be ready for review in a bit.

Generally, you don't need to let us know for these updates. Mark as draft, do your work, then mark as ready when it's gtg. Otherwise, it won't be merged unless the checks run anyway so it's fine.

> On the point of AI use
image

You started this MR before i made the clarification changes to the AI Policy document and you've been exceedingly upfront about it, so I think this MR is clean to proceed - provided the code is sound ofc - despite the fact that this AI use case scenario in fact does not clear the (new[not really]) AI Policy criteria. 

Since you said you wanna contribute in the future, please keep the [new/clarified policy](https://github.com/hyprwm/.github/blob/main/policies/AI_USAGE.md#understand-what-you-are-doing) in mind. This use case fits under the "AI does all but i was there so it's fine (it's not fine)" scenario.

I skimmed over the code, and this sure does look like it's all written by AI.


I reiterate - this is particularly relevant for your future contributions; your honestly and upfront-ness as well as the timing of the creation of this MR clears this one I'd say.

umbrageodotus:
> your honestly

minor spelling mistake. burn him in fire.
(myf for the unrelated discussion)

rogervn:
It's still work in progress but I thought I might ask before finishing.

I went deeper on the understanding now, the change ended being bigger than I thought or I'm lacking some understanding.

IIUC I need to create a different type of gesture as basically we need to start it with a keybind and have the source for the input delta update the gesture on movement and stopping when the keybind is stopped.

If that's the case, code-wide my initial thought is to split the gestures into finger and triggered gestures and have the triggered gesture register keybind callbacks that start/stop the gesture whilst the input manager updates the gesture with the delta.

And to keep being honest, I have used AI to understand point me the layers of the codebase we needed to make the changes, then I read the code on those parts myself and used the LLM again as a code generator to the design that I made myself with my understanding and making smaller changes manually.  Of course this is not 100% different from what the AI had written before, but I do own the changes that I'm doing and I'm using the LLM more of a code search, pattern comparison and code generator.

I'm still working on my understanding on the binding callbacks. There is code written for that which was written by the AI, but I'm using that more of a prototype rather what I'm going to end up with. That and validating this design is why this PR is still a draft. Also feel free to tell me that this is still not enough and I'm still relying too much on the LLM still.

PR #16174, opened 2026-09-07
docs: update links to wiki of Hyprland manpages

carlo711:
point to external, up-to-date wiki instead of github's one


wabarTM:
wait, no, what? wiki.hypr.land is made from repo, why would you do that?

repo is the upstream for site and both links are not external

carlo711:
it is external in the sense of externally hosted at wiki.hypr.land instead of the wiki tab provided by github.

old link points to the "internal" github wiki tab, see below
image

wabarTM:
oh, you mean from pages? i guess that works then

PR #16221, opened 2026-09-11
Workspace events: `created` to `created_early`, new `created` after setup is done

LRitzdorf:

Renames the existing `workspace.created` event to `workspace.created_early`, and then adds a new `workspace.created` that fires when the new workspace has been fully set up (i.e. moved to its relevant monitor, focused if appropriate, populated with the window sent to it if appropriate...).

This means the workspace object that event callbacks receive is fully initialized, and you can do stuff like this, which eliminates the need for the `on_created_empty` workspace rule:
```lua
hl.on("workspace.created", function(ws)
    if ws.empty then
        hl.exec_cmd("kitty", { workspace = ws.name.." silent" })
    end
end)
```

(Once this merges, we can kill that rule, and with it the legacy `[rules] command` exec-with-rules syntax.)

The only slight complication here is that now, anyone who creates a new workspace is responsible for calling `ready()` on it to fire `workspace.created`. That's noted in a comment, though, and is kinda necessary if we want proper behavior from this event.

Unrelated, but I've added a comment regex to disable clang-format on just one line at a time, so we don't have to wrap single lines in `// clang-format off`/`on`.

Should be good, has tests as well.

### DIFF
--- a/src/config/lua/bindings/LuaBindingsDispatchers.cpp
+++ b/src/config/lua/bindings/LuaBindingsDispatchers.cpp
@@ -1,6 +1,7 @@
 #include "LuaBindingsInternal.hpp"
 
 #include <lua.h>
+#include <hyprutils/utils/ScopeGuard.hpp>
 
 #include "Check.hpp"
 
@@ -443,7 +444,9 @@ static int dsp_moveToWorkspace(lua_State* L) {
     auto ws = Internal::resolveWorkspaceStr(lua_tostring(L, lua_upvalueindex(1)));
     if (!ws)
         return Internal::dispatcherError(L, "Invalid workspace", ERR, C_INVARG);
+    Hyprutils::Utils::CScopeGuard x([&]() { ws->ready(); });
 
+    //
     bool silent = lua_toboolean(L, lua_upvalueindex(2));
     return Internal::checkResult(L, CA::moveToWorkspace(ws, silent, Internal::windowFromUpval(L, 3)));
 }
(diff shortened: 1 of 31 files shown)


src/config/lua/bindings/LuaBindingsDispatchers.cpp
```diff
@@ -443,7 +444,9 @@ static int dsp_moveToWorkspace(lua_State* L) {
     auto ws = Internal::resolveWorkspaceStr(lua_tostring(L, lua_upvalueindex(1)));
     if (!ws)
         return Internal::dispatcherError(L, "Invalid workspace", ERR, C_INVARG);
+    Hyprutils::Utils::CScopeGuard x([&]() { ws->ready(); });
```
vaxerski, line 447: can we do this without 200 callsites of ws->ready(); because this **will** break

LRitzdorf: I was really hoping we could, but I don't think so. `workspace.created` should fire after the new workspace has been "set up," and all these callsites are needed because there are so many different places where new workspaces are created and initialized. We can't really keep it where it was before (where `workspace.created_early` is now), because at that point a lot of setup hasn't been done, like moving the workspace to the proper monitor for example.

The scope guards are an attempt to make it less hellish (especially for functions with multiple return points), but I'm not sure we can do much better than that. Ideas would be welcome though

vaxerski: well then maybe the process of initialization should be centralized?

LRitzdorf: Maybe. Like, have a workspace-creation function that takes... a new WS name, the monitor to place it on, whether to initially focus it, and an optional window to place there? I think that covers all the stuff that needs to happen before the event fires

vaxerski: if it works it works

LRitzdorf: Hm. Making workspace init accept an initial window, gets... messy, particularly when mapping a new window and sending it to a new workspace. The workspace wants a window, but the window isn't ready yet.

More practical idea: when init'ing a workspace, the caller can optionally tell it to *not* mark the workspace as ready, if it needs to do other stuff (like putting windows there) first. That won't eliminate *all* the extra `ready()` callsites, but should cut down on them significantly

PR #16231, opened 2026-09-12
Make group-lock dispatchers behave more sensibly

LRitzdorf:

Currently, `hl.dsp.group.lock()` locks *all* groups globally, but it's documented as taking a `window` table key. Also, `hl.dsp.group.lock_active()` exists.

Instead, this PR unifies those into a single dispatcher, which can be used as either `hl.dsp.group.lock({ window?, action? })` or `hl.dsp.group.lock({ all, action? })`. If specified, `all = true` gets priority. `window` defaults to the active window, as usual.

Wiki PR: hyprwm/hyprland-wiki#1658

Not really, this just follows the general format of `hl.dsp.focus`, and actually allows group-lock toggling on windows other than the active one.

Ready


hyprtester/src/tests/main/groups.cpp
```diff
@@ -308,7 +308,7 @@ TEST_CASE(groupLock) {
         }
         OK(getFromSocket("/dispatch hl.dsp.group.toggle()"));
         OK(getFromSocket(std::format("/dispatch hl.dsp.focus({{ window = 'pid:{}' }})", lockedWin->pid())));
-        OK(getFromSocket("/dispatch hl.dsp.group.lock_active({ action = 'set' })"));
```
vaxerski, line 311: keep compatibility to not break configs. Add to deprecated.

src/config/lua/bindings/LuaBindingsDispatchers.cpp
```diff
@@ -1315,7 +1327,6 @@ void Internal::registerDispatcherBindings(lua_State* L) {
         Internal::setFn(L, "active", hlGroupActive);
         Internal::setFn(L, "move_window", hlGroupMoveWindow);
         Internal::setFn(L, "lock", hlGroupLock);
-        Internal::setFn(L, "lock_active", hlGroupLockActive);
```
vaxerski, line 1318: deprecate. We have a deprecation framework.

LRitzdorf: Isn't that just for config values? I didn't see anything about deprecating Lua functions, and even the list of deprecated stuff is autogenerated so I don't think I can just add to that. Maybe we pop a notification, or something, when the deprecated dispatcher is called?

LRitzdorf:
The existing deprecation system is only for config options, not Lua API stuff. I've made `group.lock_active()` pop a notification when used, hopefully that's okay?

vaxerski:
no the current deprecation system can use arbitrary shit. Just because variables have a built-in method to append to deprecation notices doesnt mean a dispatcher cant. See `deprecationNotices`. You can add a small vec of strings cleared on reload that stores additional notices.

LRitzdorf:
Ah okay, there was only plumbing for config options. I've added what I think is a reasonable system for deprecated Lua bindings; there's a string vector that we append to when a deprecated feature is used, and that vector gets cleared on reload.

For dispatchers, the notice is added at construction time, so we do indeed get the usual "hey you're using deprecated stuff" notif on startup. Also, if you construct a deprecated dispatcher twice, you get two identical deprecation messages; that seems reasonable enough to me (e.g. if the user removes one usage, they see one less deprecation notice and it's clear that progress is being made).

hyprtester/src/tests/main/groups.cpp
vaxerski: this is still unresolved. Tests should cover both

src/config/lua/bindings/LuaBindingsDispatchers.cpp
```diff
@@ -138,18 +141,42 @@ static int hlGroupActive(lua_State* L) {
...
 
+// TODO: Remove this later
 static int hlGroupLockActive(lua_State* L) {
+    deprecated.emplace_back(std::pair("hl.dsp.group.lock_active()", "The group.lock_active() dispatcher is deprecated, use group.lock() instead"));
+
     const auto action = Internal::tableToggleAction(L, 1);
 
-    lua_pushnumber(L, (int)action);
-    lua_pushcclosure(L, dsp_lockActiveGroup, 1);
+    lua_pushnumber(L, static_cast<int>(action));
```
vaxerski: sc<> from hu

src/config/lua/bindings/LuaBindingsDispatchers.cpp
```diff
@@ -13,14 +14,16 @@
...
 #include "../../../keybinds/Resolver.hpp"
-#include "config/shared/actions/ConfigActions.hpp"
 
 using namespace Config;
 using namespace Config::Lua;
 using namespace Config::Lua::Bindings;
 
 namespace CA = Config::Actions;
 
+static std::vector<std::pair<std::string, std::string>> deprecated;
```
vaxerski, line 24: ?!

vaxerski:
> For dispatchers, the notice is added at construction time, so we do indeed get the usual "hey you're using deprecated stuff" notif on startup. Also, if you construct a deprecated dispatcher twice, you get two identical deprecation messages; that seems reasonable enough to me (e.g. if the user removes one usage, they see one less deprecation notice and it's clear that progress is being made).

it can be call time or construction time I dont care

src/config/lua/bindings/LuaBindingsDispatchers.cpp
LRitzdorf: This seemed reasonable because we can store the name of the deprecated function, and a message that doesn't need to restate the name of the function. Would a vector of single strings be better?

PR #16234, opened 2026-09-12
workspace: create named workspaces for bare named lookups

vaxerski:
makes `hl.dsp.focus({ workspace = 'PISS' })` work even when the workspace isnt opened yet

### DIFF
--- a/src/desktop/state/FocusState.cpp
+++ b/src/desktop/state/FocusState.cpp
@@ -277,7 +277,7 @@ void CFocusState::rawMonitorFocus(PHLMONITOR pMonitor) {
 
     const auto PWORKSPACE = pMonitor->m_activeWorkspace;
 
-    const auto WORKSPACE_ADDRESS = PWORKSPACE ? Workspace::selector(*PWORKSPACE) : "";
+    const auto WORKSPACE_ADDRESS = PWORKSPACE ? PWORKSPACE->addressableName() : "";
     const auto WORKSPACE_NAME    = PWORKSPACE ? PWORKSPACE->addressableName() : "?";
 
     IPC::Socket2::sock()->postEvent({.event = "focusedmon", .data = std::format("{},{}", pMonitor->m_name, WORKSPACE_NAME)});
--- a/src/desktop/state/FocusState.cpp
+++ b/src/desktop/state/FocusState.cpp
@@ -278,7 +278,7 @@ void CFocusState::rawMonitorFocus(PHLMONITOR pMonitor) {
     const auto PWORKSPACE = pMonitor->m_activeWorkspace;
 
     const auto WORKSPACE_ADDRESS = PWORKSPACE ? Workspace::selector(*PWORKSPACE) : "";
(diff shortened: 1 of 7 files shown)


Dregu:
Yeah this works for focus.

All socket2 events with WORKSPACEID are also sending "name:foo" instead of "foo", because the events use `Workspace::selector()` which appends name: to non-number workspaces. I guess this function is used for *matching* legacy workspace selectors, but for all socket2 events we should send `ws->addressableName()` to match socket1.

vaxerski:
true

vaxerski:
@Dregu yes?

src/desktop/state/FocusState.cpp
```diff
@@ -278,7 +278,7 @@ void CFocusState::rawMonitorFocus(PHLMONITOR pMonitor) {
     const auto PWORKSPACE = pMonitor->m_activeWorkspace;
 
     const auto WORKSPACE_ADDRESS = PWORKSPACE ? Workspace::selector(*PWORKSPACE) : "";
-    const auto WORKSPACE_NAME    = PWORKSPACE ? PWORKSPACE->displayName() : "?";
+    const auto WORKSPACE_NAME    = PWORKSPACE ? PWORKSPACE->addressableName() : "?";
```
Dregu, line 281: Fixed the wrong one, NAME = displayName, ADDRESS = addressableName... Also all v1 events should use displayName where spec says WORKSPACENAME, you changed some random v1s to addressableName too.

vaxerski: I dont think v1s should use display name?

vaxerski:
@Dregu yes?

Dregu:
No, surely WORKSPACENAME and WORKSPACEID are NOT supposed to be the same thing in the IPC. For example workspacev2 makes no sense, supposed to send two different things, id and name, you made it send the same thing twice. If I have a rule `hl.workspace_rule{ workspace = 4, default_name = "Games" }` it's supposed to send `[create]workspacev2>>4,Games` not `4,4`, what would be the point of that :D. v1 is supposed to be `workspace>>Games` not `4`.

Just go through the IPC specification and match WORKSPACENAME with displayName, WORKSPACEID with addressableName, everything else is wrong. Make a v3 if you want to change how everything works.

vaxerski:
yes, the thing is V2 makes no sense with this new system in general. the distinction between name and id was there precisely because IDs existed and you couldn't refer to `name:hello` as `-1632`.

Now with this system the distinction is pointless because you can refer to them as the addressable names report

Dregu:
Yeah but what is the point of removing the name for no reason at all *when address is already in the same fucking event*? "Address is better than id" is a stupid ass reason to intentionally break the name functionality :D. Previously it was possible to make e.g. a workspace plugin for a bar update solely on the socket2 data, now you have to ask socket1 what the name of this new workspace is cause someone decided to remove it from socket2 for no reason.

vaxerski:
ok

PR #16250, opened 2026-09-13
feature: Input swallow focus click

pedromiguelvc:
#### Pre requirements
First of all, the vouch was properly opened (and already closed) as it is a requirement for new users to open pr in this repo
https://github.com/hyprwm/.github/discussions/62

This pr adds a new boolean option to suppress forwarding of the mouse click on window focus, `input:swallow_focus_click`, being true means the first click won't be forwarded and just change the keyboard focus. It will be false by default so it wont affect any pre existing hyprland configuration. Of course, it only makes sense to be true if the `input:follow_mouse` is either 0 or 2, though it is safe to have it true while `input:follow_mouse` is 1 or 3 - won't have any practical effect.

To use it, just add 
```lua
input {
    swallow_focus_click = true
}
```

One edge case I did not test properly is if the `input:swallow_focus_click` change while pressing the mouse and before the release of it, although it seems that the behavior will honor whatever was defined in the pressing time since the release time only reads what was populated in the `m_swallowedFocusButton` list.

AI was used to implement this pr as a mentor, although I wrote the code myself, AI helped me to locate where in the code base I would need to change and explain it to me. Also I used it to make sure I was following the code base pattern (formatting, naming, conventions, etc). One specific case it supported me was debugging two mistakes I introduced before the feature actually worked, a computed but unused `swallowed` variable and duplicated `sendPointerButton` call

Yes, it is ready for merging. A follow-up documentation PR to hyprland-wiki will be opened once this one gets accepted.


src/managers/input/InputManager.cpp
```diff
@@ -860,13 +860,14 @@ void CInputManager::setClickMode(eClickBehaviorMode mode) {
...
-    const auto  PASS              = Keybinds::mgr()->onMouseEvent(e, mouse);
-    static auto PFOLLOWMOUSE      = CConfigValue<Config::INTEGER>("input:follow_mouse");
-    static auto PRESIZEONBORDER   = CConfigValue<Config::INTEGER>("general:resize_on_border");
-    static auto PBORDERSIZE       = CConfigValue<Config::INTEGER>("general:border_size");
-    static auto PBORDERGRABEXTEND = CConfigValue<Config::INTEGER>("general:extend_border_grab_area");
-    const auto  BORDER_GRAB_AREA  = *PRESIZEONBORDER ? *PBORDERSIZE + *PBORDERGRABEXTEND : 0;
+    static auto PPASSMOUSE         = CConfigValue<Config::INTEGER>("binds:pass_mouse_when_bound");
+    const auto  PASS               = Keybinds::mgr()->onMouseEvent(e, mouse);
+    static auto PFOLLOWMOUSE       = CConfigValue<Config::INTEGER>("input:follow_mouse");
+    static auto PSWALLOWFOCUSCLICK = CConfigValue<Config::BOOL>("input:swallow_focus_click");
```
pedromiguelvc: formatting shifted the whole column because of the new PSWALLOWFOCUSCLICK name

src/managers/input/InputManager.cpp
```diff
@@ -928,8 +932,16 @@ void CInputManager::processMouseDownNormal(const IPointer::SButtonEvent& e, SP<I
         case WL_POINTER_BUTTON_STATE_RELEASED: break;
     }
 
-    // notify app if we didn't handle it
-    g_pSeatManager->sendPointerButton(e.timeMs, e.button, e.state);
+    bool swallowed;
```
vaxerski: no uninitialized primitives

src/managers/input/InputManager.cpp
```diff
@@ -928,8 +932,16 @@ void CInputManager::processMouseDownNormal(const IPointer::SButtonEvent& e, SP<I
...
+    if (e.state == WL_POINTER_BUTTON_STATE_PRESSED) {
```
vaxerski: no {}

src/managers/input/InputManager.hpp
```diff
@@ -277,6 +277,9 @@ class CInputManager {
     // for releasing mouse buttons
     std::list<SHeldPointerButton> m_currentlyHeldButtons;
 
+    // for swallowing button presses that triggered a focus change
+    std::list<SHeldPointerButton> m_swallowedFocusButtons;
```
vaxerski: don't use lists

src/managers/input/InputManager.hpp
```diff
@@ -277,6 +277,9 @@ class CInputManager {
     // for releasing mouse buttons
     std::list<SHeldPointerButton> m_currentlyHeldButtons;
 
+    // for swallowing button presses that triggered a focus change
+    std::vector<SHeldPointerButton> m_swallowedFocusButtons;
```
vaxerski, line 281: this can get stuck. Shouldn't it be cleared or something?

pedromiguelvc: Reused the destroyPointer and releaseAllMouseButtons for this purpose since they already handle m_currentlyHeldButtons the same way.

Also left out the synthetic release event because m_swallowedFocusButtons never forwards presses, different from m_currentlyHeldButtons.

PR #16251, opened 2026-09-13
feature: fullscreen controller takeover & restore 

oclaw:
This PR adds to the FullscreenController fullscreen takeover feature support - in case if new window in the workspace was created by the backend + requests fullscreen at the moment of creation - controller remembers current workspace fullscreen state and re-applies it when the window that took FS over the existing one, gets closed
 
It might be useful for various Media / PDF / etc viewers that are spawned from inside of the application and often are fullscreened by default and their termination leads to the reset of the workspace FS state to the default one
 
Changes:
- is_backend bool flag added to the pending fullscreen request
- fullscreen requests with backend source call `beginFullscreenTakeover` in `mapWindow()`
- fullscreened windows try to rollback the FS state with `endFullscreenTakeover` on `unmapWindow()`
 
Initial impl of the feature is enabled by default and is disabled in case of existence of any config flags overriding FS behavior (like `exit_window_retains_fullscreen`). It is a question to discuss, maybe it should also be guarded by some config as well 

AI was used for the navigation through the codebase and exploring the corner cases.

Decision about config flag should be made. After agreement on the design of the feature the integration test should also be added

### DIFF
--- a/src/desktop/view/window/Window.cpp
+++ b/src/desktop/view/window/Window.cpp
@@ -5,6 +5,8 @@
 #include <hyprutils/animation/AnimatedVariable.hpp>
 
 #include "../Group.hpp"
+#include "debug/log/Logger.hpp"
+#include "managers/fullscreen/FullscreenTypes.hpp"
 
 #if defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__DragonFly__)
 #include <sys/types.h>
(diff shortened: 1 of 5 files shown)


erstarr:
1. MR body. You need to properly fill each field, this is not optional.

2. As i mentioned in the vouch req, consider expanding the already existing client FS request framework.

Save the fact that current window is an FS takeover, and upon closing you are to FS this window (WP<> to that widow).


Also an enum reason for the specifici client FS request special case (FSMODE_MAX, or fullscreen takeover).

This will allow you to chain these as well: 2 FS takeovers back to back will properly restore one by one as they are closed.

Consider how this might be done in scrolling FS --> newly created FSed window doesn't really take over. You probably need to make a method to check if this client FS request is applicable for the current FS handler and don't save the request if not. FSMODE_MAX is applicable, this is not. Passing the reason enum to see what's what is good.


I also did the FSMODE_MAX case with an unordered map before, but vax moved that to fullscreen policy of each window - which i think is indeed a better place to store these special cases.

Handlers should keep their lists as i want to make it possible for them to implement custom FS modes if they want to (or at least for now. In the future i might rework that again), but otherwise keeping em tied to window objects is better to avoid maintenance functions like `syncFullscreenTargets()` .

Also extending the framework will make it possible to slot future special cases like this one right in without making a list of each of em

<br><br>

Honestly as it is I don't really like it. The general idea is solid (save state, restore), but having separate handling/storage for each special case will get messy quickly

oclaw:
@erstarr Thanks for the review! MR body is updated and I will try to follow your approach soon

vaxerski:
I dont understand the transactionality of this. Why not simply have two methods, set and unset, and then a replace?

Set would automatically push to a fullscreen "stack"

PR #16252, opened 2026-09-13
protocols/WaylandProtocol: init destroy-listener link before global-create check

ziebamikolaj:
## Summary

`IWaylandProtocol`'s constructor returns early if `wl_global_create()` fails, before ever calling `wl_list_init()` on `m_liDisplayDestroy.listener.link`. `~IWaylandProtocol()` (via `onDisplayDestroy()`) unconditionally calls `wl_list_remove()` on that link regardless — a segfault on whatever uninitialized memory happens to be there.

`wl_global_create()` failing isn't hypothetical: on a system where the installed `wayland-protocols` implements an older version of an interface than a specific protocol subclass requests (e.g. `zxdg_decoration_manager_v1` v1 vs. a requested v2 — logged as a `WARN`, not treated as fatal elsewhere in Hyprland), this is exactly what happens. The result is a **100% deterministic crash on every graceful Hyprland exit** on such a system, for any protocol whose global creation fails this way — with or without any plugin loaded, unrelated to whatever triggered the exit.

I ran into this while chasing what looked at first like a plugin-specific, then a data-race, touchscreen-gesture crash. A real gdb backtrace from a real core dump nailed it down to this constructor/destructor pair:

```
#0 wl_list_remove() [libwayland-server.so]
#1 IWaylandProtocol::~IWaylandProtocol()
#2 CXDGDecorationProtocol::~CXDGDecorationProtocol()
#3 CUniquePointer<CXDGDecorationProtocol>::decrement()
#4 CProtocolManager::~CProtocolManager()
#5 CCompositor::cleanup()
#6 main()
```

Inspecting the crashing object's memory directly (via `gdb` on the core dump) confirmed the listener was never initialized at all: `m_liDisplayDestroy.parent = NULL`, `.listener.notify` garbage, `.listener.link.next = 0xf0` — none of which the constructor had set on the early-return path.

## Fix

- `wl_list_init(&m_liDisplayDestroy.listener.link)` now runs unconditionally, before the `!m_global` check, so a never-registered listener stays safely self-linked. `wl_list_remove()` on it is then a correct no-op instead of a use of uninitialized memory.
- Added a small guard (`m_displayDestroyed`) so `onDisplayDestroy()`'s body can't run twice for the same object either (once via the `wl_display` destroy-listener firing, once via the destructor calling it directly) — a smaller, separate hardening, not the root cause here, but correct regardless.

## Test plan

- [x] Reproduced reliably before the fix: a bare Hyprland instance (no plugins) crashing identically on every graceful exit (`hyprctl dispatch exit`) on a system with an older `wayland-protocols` package.
- [x] After the fix: 5+ repeated exit cycles, zero crashes, zero new core dumps, on the same system/config that crashed every single time before.
- [x] Verified via `nm`/`readelf` build-id matching that the deployed binary was actually the patched one before declaring it fixed.


erstarr:
Slopped. Don't bother with the Vouch request, you won't get it with this MR

PR #16254, opened 2026-09-13
input: touchpad scroll acceleration and inertial coasting

avillagran:
## Motivation

Touchpad scroll on Hyprland has speed (`scroll_factor`) but no acceleration and, more importantly, **no inertia**: the moment the finger lifts, scrolling stops dead. Every major desktop (macOS, Windows precision touchpads) accelerates scroll with finger velocity and keeps coasting smoothly after release. This adds both, compositor-side, so it works in every Wayland client regardless of toolkit.

## What

New options under `input:touchpad:`:

| Option | Default | Description |
|---|---|---|
| `scroll_accel_profile` | 1 | 0=off, 1=linear, 2=adaptive (smoothstep) |
| `scroll_accel_speed` | 1.0 | Velocity (delta/ms) at which acceleration reaches maximum |
| `scroll_accel_max` | 3.0 | Acceleration multiplier cap |
| `scroll_decel` | 600 | Inertial coasting time (ms) after finger lift, 0=off |
| `scroll_ignore_classes` | "" | Comma-separated window class substrings where accel+coast stay off |

- Acceleration applies only to `WL_POINTER_AXIS_SOURCE_FINGER` events; wheel mice are untouched.
- Velocity is tracked per axis (vertical/horizontal) with an exponential moving average over the gesture.
- Coasting: after a 24 ms silence watchdog, synthetic axis events are emitted for `scroll_decel` ms with a quadratic ease-out decay (120 Hz ticks), seeded from the gesture's smoothed velocity. Any real motion on an axis cancels its coast; a clean zero-delta stop event ends the gesture for clients.
- `scroll_ignore_classes`: apps with their own touchpad inertia (browsers like Chrome/Firefox, some terminals) keep their native behavior instead of stacking a second inertia on top — matching windows receive raw deltas, no acceleration and no coasting, and a running coast cancels immediately if the pointer moves over an ignored window. Empty (default) applies the patch everywhere; the check short-circuits on an empty list so the default path has zero cost.
- Values are read per event at runtime — no input device re-initialization on config change.

## Testing

Tested locally on an Apple Magic Trackpad (nested session) with a Quickshell test UI toggling all options live via `hyprctl eval`: acceleration curves behave as documented, coast distance matches the integral of the decay curve (~`seed * ms / 3` units), per-axis state handles diagonal flicks correctly, and the ignore list round-trips via `getoption`/`eval`.

Notes:
- Defaults are open to discussion; profiles default to linear-on which matches typical touchpad behavior, but 0 (opt-in) is a one-line change if preferred.
- While testing nested I noticed Aquamarine's Wayland backend drops `wl_pointer.axis_source` (everything arrives as `WHEEL`); separate PR: https://github.com/hyprwm/aquamarine/pull/405 — on real hardware the libinput backend reports `FINGER` correctly.


erstarr:
AI written, AI thought, AI filed. Don't bother making a vouch request, it will not be accepted with this MR.

avillagran:
Thanks! I'm happy to help. My touchpad works perfectly on Linux; I think I'm the first in the world ;)

Lots of love!

PR #16269, opened 2026-09-15
ipc: clean up workspace data

vaxerski:
this removes some redundant data from s1 and cleans up the jsons. Adds "id" where possible for compat

tests by slopex


Dregu:
Several projects have already migrated to using the type field you added 2 weeks ago and now you want to remove it... Dude.

vaxerski:
type was unintentional, it provides no real information

vaxerski:
I can keep it though but I will keep it to normal and special. Numbered is just normal.

vaxerski:
also it's on them to be chasing git lol

Dregu:
Well chasing is always a guessing game of supporting new features too soon or too late, nobody knows when release is...

"special" is actually where "type" is useful anyway, I agree that workspaces "7" and "foobar" are just the same type of workspace... Don't really see the usefulness of the id field though, probably confusing it's not always present and it's basically same as doing atoi(address).

vaxerski:
it is just for compat just in case™

src/workspace/AbstractWorkspace.hpp
```diff
@@ -33,13 +34,20 @@ namespace Workspace {
...
-
     enum class eWorkspaceType : uint8_t {
         NORMAL,
         SPECIAL,
     };
 
+    constexpr const char* workspaceTypeToString(eWorkspaceType t) {
+        switch (t) {
+            case eWorkspaceType::NORMAL: return "Normal";
+            case eWorkspaceType::SPECIAL: return "Special";
```
Dregu: I'd keep these lowercase, there's no precedent in the json for capitalizing like this, and most things that represent types of things seem to be all lowercase? Some are also all uppercase, but this one would add a third inconsistent way to do it...

vaxerski: kid named `toLower`

Dregu:
I guess it's fine. But #16234 still left some inconsistent v1 events, i.e. `activespecial`, `createworkspace` and `destroyworkspace` are still using names instead of addresses. Please don't touch anything else :D

vaxerski:
or what

Dregu:
Ah great you decided to not fix the problem AGAIN

PR #16284, opened 2026-09-17
Fix typo in damping parameter to dampening

vindictivebanana:
I tried to use this config, and it works just fine, except that 2 animations are set to use a curve "easy" that has a typo. it uses the option "dampening" but is mistyped as "damping", which causes a few errors.


wabarTM:
AHAHAHAHBAHAHAAHHABA

PR #15435, opened 2026-07-16
Allow pointer constraints to work on subsurface

Shiphan:
#### Describe your PR, what does it fix/add?

Add support for pointer constraints on subsurface.

Godot use subsurface to embed game window into editor and debug window.
With godotengine/godot#120647 merged + this patch, mouse capture can then work on embedded game window.

PR #16311, opened 2026-09-20
Allow pointer constraints to work on subsurface

Shiphan:

#15435 
Add support for pointer constraints on subsurface.
Godot use subsurface to embed game window into editor and debug window.

There was a regression issue (now solved) where pointer can still be locked after keyboard focus is already on a different window
The condition that I can reproduce this bug is, with two screens, and switch focus to a window that is already on the other screen, then pointer will still be locked on the original window.

Didn't find other similar issue, but I don't usually use a dual monitor setup, so some use cases may not have been tested.


vaxerski:
@Dregu

Dregu:
In hindsight I probably shouldn't have uninstalled the 200GB game where the repro was...

PR #16314, opened 2026-09-20
Fix: dumb tokens not updating when using autostart

ItsOhen:
Would only update the persistent tokens (initial_workspace_tracking = 2) when a window was moved between workspaces or put into a group. So windows started with a workspace exec rule would never update it and have all children spawn on ws1.

No.

No.

Yes.


vaxerski:
tests?

ItsOhen:
> tests?

Are you sure you want me to autostart a program with a daemon in the tester just to test this? (in hl.on("hyprland.start",))

Because that's the only way i can reliably prove that it works.

